PR #1, opened 2018-05-10
Create basic files and test package

deads2k:
This gets the repo started with a package we know we'll want to move.

/assign @liggitt


deads2k:
Starting here.

PR #2, opened 2018-05-11
add controller package

deads2k:
moves the controller package from openshift.


deads2k:
the kube package deps are reasonable, the fanout is unfortunate

PR #5, opened 2018-05-11
[WIP] let's talk about complexity

deads2k:
I've pulled this item as a real-life example that is holding up the split of the openshift oauthserver.  There is a helper shared between the openshift-apiserver and the openshift-authserver to determine the name of a client-authorization.  

The complexity here is low and this effectively cannot change or the API callers will not be able to predict how to use the API.  I'd claim that in this case we simply duplicate the functionality in both repos.

@openshift/api-review @openshift/sig-security Disagreement?  I suspect we'll have several like this over time.

/hold

### DIFF
--- a/pkg/oauth/helpers.go
+++ b/pkg/oauth/helpers.go
@@ -1,0 +1,7 @@
+package oauth
+
+// ClientAuthorizationName creates the computed name for a given username, clientName tuple
+// This cannot change or client authorizations will be have unpredictably
+func ClientAuthorizationName(userName, clientName string) string {
+	return userName + ":" + clientName
+}


pkg/oauth/helpers.go
```diff
@@ -0,0 +1,7 @@
+package oauth
+
+// ClientAuthorizationName creates the computed name for a given username, clientName tuple
+// This cannot change or client authorizations will be have unpredictably
```
enj: "behave"

enj:
I had planned on changing this when I get around to adding user management of these objects (which would require changing the ETCD layout), but I can simply drop support for "legacy" authorizations.

enj:
xref: https://github.com/openshift/origin/pull/11909

deads2k:
Great.  Hopefully we'll be able to find this link for all pulls that don't meet the bar.

liggitt:
> Hopefully we'll be able to find this link for all pulls that don't meet the bar.

hoisting it as an example into the repo readme might help

PR #6, opened 2018-05-14
Add config helpers

deads2k:
Adds config helpers for webconsole, service serving certs, and eventually origin

Adds serviceability for image-registry, webconsole, service serving cert, and origin.

@openshift/sig-master


pkg/config/helpers/general.go
```diff
@@ -0,0 +1,58 @@
...
+			// Don't resolve absolute paths
+			if !filepath.IsAbs(*ref) {
+				*ref = filepath.Join(base, *ref)
+			}
+		}
+	}
+	return nil
+}
+
+func MakeRelative(path, base string) (string, error) {
```
mfojtik: @deads2k i know you will hate me, but I think if we call this "library" we should document the exported library functions ;-)

pkg/config/helpers/general.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, ref := range refs {
+		// Don't relativize empty paths
+		if len(*ref) > 0 {
+			rel, err := MakeRelative(*ref, base)
+			if err != nil {
+				return err
+			}
+
+			// if we have a backstep, don't mess with the path
+			if strings.HasPrefix(rel, "../") {
```
mfojtik: will break on windows (if we care or ever plan to use this in CLI)

pkg/config/helpers/serving_info.go
```diff
@@ -0,0 +1,29 @@
+package helpers
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func GetHTTPServingInfoFileReferences(config *configv1.HTTPServingInfo) []*string {
```
mfojtik: `var GetHTTPServingInfoFileReferences = GetServingInfoFileReferences` :-)

pkg/config/helpers/general.go
deads2k: > @deads2k i know you will hate me, but I think if we call this "library" we should document the exported library functions ;-)

sure

pkg/config/helpers/general.go
deads2k: > will break on windows (if we care or ever plan to use this in CLI)

Hasn't come up.  I'm inclined to leave it.  I'm also of the opinion that I only like to support forward slashes and windows users can just deal with it.

pkg/config/helpers/serving_info.go
deads2k: > var GetHTTPServingInfoFileReferences = GetServingInfoFileReferences

They are different types

deads2k:
comments addressed

pkg/config/helpers/general.go
sttts: Can I quote "I'm also of the opinion that I only like to support forward slashes and windows users can just deal with it." ?

PR #10, opened 2018-05-23
easy resource manipulation

deads2k:
Adds functions to provide easy manipulation of common resources needed by operators.  These were pulled from real-life usage in webconsole and for applicability to SSCS.

@sttts @mfojtik


pkg/operator/resource/resourcecread/apps.go
```diff
@@ -0,0 +1,34 @@
...
+	appsCodecs = serializer.NewCodecFactory(appsScheme)
+)
+
+func init() {
+	if err := appsv1.AddToScheme(appsScheme); err != nil {
+		panic(err)
+	}
+}
+
+func ReadDeploymentV1OrDie(objBytes []byte) *appsv1.Deployment {
```
mfojtik: @deads2k as this is library, i would like to have godoc on these ;-) i'm fine with follow up after we move everything we need, not going to hold this on godoc...

mfojtik:
/lgtm

the content was already reviewed during webconsole operator work, this is a mechanical move...
one optional nit which can be addressed after we finish moving everything we need

@sttts ?

PR #11, opened 2018-05-25
helpers and tests

deads2k:
Adds helpers and tests for those helpers from https://github.com/openshift/service-serving-cert-signer/pull/5

/assign @mfojtik 

@mfojtik I'm looking for this before I merge that WIP.  Once I have it, we'll be able to snip a good amount of code.


pkg/crypto/crypto.go
```diff
@@ -819,29 +864,27 @@ func encodeKey(key crypto.PrivateKey) ([]byte, error) {
...
-	// ensure parent dir
-	if err := os.MkdirAll(filepath.Dir(path), os.FileMode(0755)); err != nil {
+	if _, err := f.Write(bytes); err != nil {
 		return err
 	}
 
-	b, err := encodeKey(key)
+	return nil
+}
+func writeKeyFile(f io.Writer, key crypto.PrivateKey) error {
```
mfojtik: 👍

pkg/operator/resource/resourceapply/core.go
```diff
@@ -101,3 +101,26 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, required *corev1.Confi
 	actual, err := client.ConfigMaps(required.Namespace).Update(existing)
 	return actual, true, err
 }
+
+// ApplySecret merges objectmeta, requires data
```
mfojtik: improvement :-))

pkg/operator/resource/resourceapply/core.go
```diff
@@ -101,3 +101,26 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, required *corev1.Confi
...
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	dataSame := equality.Semantic.DeepEqual(existing.Data, required.Data)
+	if dataSame && !*modified {
+		return existing, false, nil
+	}
+	existing.Data = required.Data
```
mfojtik: if somebody use this with lister, we will need a deep copy here, right?

mfojtik:
just one question

LGTM (I still don't have merge rights for this repo :-)

pkg/operator/resource/resourceapply/core.go
deads2k: > if somebody use this with lister, we will need a deep copy here, right?

We do the get immediately above.  It can't wire to a lister.

PR #21, opened 2018-07-18
v1.11.1

liggitt:
Depends on
* https://github.com/openshift/api/pull/75


liggitt:
@deads2k PTAL
pinned two deps to kube levels, `make verify build test` passes cleanly

deads2k:
lgtm

PR #22, opened 2018-08-07
 add network annotation helper

deads2k:
This appears to be an attempt to type have type safety on an annotation for a type we own.  It's a little weird, but it's used from oc and controllers.

/hold 

holding for real bump


liggitt:
lgtm other than the glide yaml switch back

deads2k:
> lgtm other than the glide yaml switch back

updated

/hold cancel

PR #23, opened 2018-08-23
move pkg/git to library-go

bparees:


pkg/git/repository_test.go
```diff
@@ -0,0 +1,149 @@
+package git
+
+import (
+	"testing"
+	"time"
+)
+
+func TestGetRootDir(t *testing.T) {
```
bparees: @mfojtik @deads2k do we run unit tests against library-go?

bparees: @mfojtik @deads2k bump

mfojtik: Yes we do, in Travis

deads2k:
merging per michal's lgtm

PR #25, opened 2018-09-10
add image ref parsing utilities

bparees:


smarterclayton:
@deads2k too

pkg/image/reference/reference.go
```diff
@@ -0,0 +1,607 @@
...
+	if len(match) != 3 {
+		return "", name
+	}
+	return match[1], match[2]
+}
+
+// ParseToReference parses s and returns a syntactically valid Reference.
+// If an error was encountered it is returned, along with a nil Reference.
+// NOTE: Parse will not handle short digests.
+func ParseToReference(s string) (Reference, error) {
```
bparees: this and the other Parse* function below are the notable changes.  I munged the two "reference.go" files from origin into a single file here, and they both declared a global Parse(string) function.

bparees:
@deads2k please sign off asap so https://github.com/openshift/origin/pull/20885 can move forward.

pkg/image/digest/digest.go
```diff
@@ -0,0 +1,138 @@
+package digest
```
deads2k: this used to be in an `internal` package that prevented anyone from depending on it. We did this because it was a copy of existing code from elsewhere.

pkg/image/reference/reference.go
deads2k: I'd like to understand why.  This entire package is a copy from elsewhere, forced internal so no one would ever depend on it directly, used to snip a link to docker.  The comments on the package indicate that we're strongly biased against change here.

pkg/util/util.go
```diff
@@ -0,0 +1,22 @@
+package util
+
+import (
+	"bufio"
+	"os"
+)
+
+// ReadLines reads the content of the given file into a string slice
+func ReadLines(fileName string) ([]string, error) {
```
deads2k: This appears to be a copy of a function only used by builders.  The rule for library go is that it must be of reasonable complexity (this may fail that) and must be used by multiple repos (this does fail that).

deads2k:
remove pkg/util/util.go and you're good.

deads2k:
/lgtm

plan to merge on green.

PR #26, opened 2018-09-13
add recommended defaults and client helpers

deads2k:
Adds recommended defaults and client helpers used by openshift and kube apiservers

/assign @mfojtik @sttts


pkg/config/configdefaults/config_default.go
```diff
@@ -0,0 +1,67 @@
+package configdefaults
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+func DefaultString(target *string, defaultVal string) {
+	if len(*target) == 0 {
```
soltysh: How about nil check as well?

pkg/config/configdefaults/config_default.go
```diff
@@ -0,0 +1,67 @@
...
+		*target = defaultVal
+	}
+}
+
+func DefaultStringSlice(target *[]string, defaultVal []string) {
+	if len(*target) == 0 {
```
soltysh: Ditto?

pkg/config/configdefaults/config_default.go
deads2k: > How about nil check as well?

There's nothing you can do to fix the call at that point.  We discussed it in the original pull and decided it was programmer error.

PR #27, opened 2018-09-14
add helper for processing and writing assets using go templates

mfojtik:
We will need this in every master operator repo to render the manifests for bootkube.
Instead of copying this, make it shared.

/cc @deads2k
/cc @sttts

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -1,0 +1,79 @@
+package assets
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+type Permission os.FileMode
+
+const (
+	PermissionDirectoryDefault Permission = 0755
+	PermissionFileDefault      Permission = 0644
+	PermissionFileRestricted   Permission = 0600
+)
+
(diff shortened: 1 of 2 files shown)


pkg/assets/assets.go
```diff
@@ -0,0 +1,71 @@
...
+	"text/template"
+)
+
+const (
+	defaultAssetDirPerm = 0755
+	defaultAssetPerm    = 0600
+)
+
+// Asset defines a single static asset.
+type Asset struct {
```
deads2k: I expected to see a permission option here.

pkg/assets/assets.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"text/template"
+)
+
+const (
+	defaultAssetDirPerm = 0755
+	defaultAssetPerm    = 0600
```
deads2k: not 0644 with an option to make it secret and be 0600?

pkg/assets/assets.go
deads2k: Or perhaps a `Confidential` bool to get it to be 0600

PR #28, opened 2018-09-17
add missing file ref

deads2k:
Missed a file reference when purging adding new types.

/assign @mfojtik

### DIFF
--- a/pkg/config/helpers/config_refs.go
+++ b/pkg/config/helpers/config_refs.go
@@ -95,6 +95,16 @@ func GetAuditConfigReferences(config *configv1.AuditConfig) []*string {
 	return refs
 }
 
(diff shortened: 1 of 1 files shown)


pkg/config/helpers/config_refs.go
```diff
@@ -95,6 +95,16 @@ func GetAuditConfigReferences(config *configv1.AuditConfig) []*string {
...
 }
 
+func GetKubeClientConfigReferences(config *configv1.KubeClientConfig) []*string {
+	if config == nil {
+		return []*string{}
+	}
+
+	refs := []*string{}
+	refs = append(refs, &config.KubeConfig)
+	return refs
```
mfojtik:
```
return []*string{&config.Kubeconfig} ?
```

mfojtik: actually nvmd, better to be consistent

PR #29, opened 2018-09-17
 add apiservice applying

deads2k:
This could/should be used to install the service catalog and metrics servers too.  If we ever split up the openshift apiserver, we'll have even more.

/assign @sttts @mfojtik


pkg/operator/resource/resourceapply/apiregistration.go
```diff
@@ -0,0 +1,36 @@
...
+		actual, err := client.APIServices().Create(required)
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(modified, &existing.ObjectMeta, required.ObjectMeta)
+	// there was no change to metadata, the service and priorities were right
```
sttts: misplaced?

sttts:
One nit. Otherwise lgtm.

deads2k:
fixed nit.  plan to merge on green.

PR #31, opened 2018-09-19
assets: add helpers for recursive directory walk

sttts:


pkg/assets/assets.go
```diff
@@ -67,13 +85,37 @@ func MustCreateAssetFromTemplate(name string, template []byte, config interface{
...
+	files := map[string][]byte{}
+	err := filepath.Walk(dir,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				return nil
+			}
+
```
mfojtik: do we want to load only .yaml files here? iow. does this load README or OWNERS ?

pkg/assets/assets.go
```diff
@@ -67,13 +85,37 @@ func MustCreateAssetFromTemplate(name string, template []byte, config interface{
...
+			bs, err := ioutil.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			files[strings.TrimLeft(strings.TrimPrefix(path, dir), "/")] = bs
```
mfojtik: nit: can you drop a comment that this preserves the relative path?

pkg/assets/assets.go
```diff
@@ -67,13 +85,37 @@ func MustCreateAssetFromTemplate(name string, template []byte, config interface{
 }
 
 func assetFromTemplate(name string, tb []byte, data interface{}) (Asset, error) {
```
mfojtik: nit: can we just return *Asset ? then in case of error we don't have to return empty struct...

pkg/assets/assets.go
```diff
@@ -27,6 +25,26 @@ type Asset struct {
...
+	files, err := LoadFilesRecursively(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	var as Assets
+	for path, bs := range files {
+		a, err := assetFromTemplate(path, bs, data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to render %q: %v", path, err)
```
mfojtik: i would probably just store the error in array and return errors at once... it will suck when I fix one error, rerun and get other error, etc... would be better to get all errors at once ;-)

sttts: good idea

pkg/assets/assets.go
```diff
@@ -27,6 +25,26 @@ type Asset struct {
 // Assets is a list of assets.
 type Assets []Asset
 
+// NewRecursiveAssets walks through a directory and render each file as asset.
+func NewRecursiveAssets(dir string, data interface{}) (Assets, error) {
```
deads2k: if there isn't another new function, `New` should do.  I don't think people should get an option against recursive.

pkg/assets/assets.go
sttts: that's your code ;-) But yes, we can :)

pkg/assets/assets.go
sttts: good question. This function is also used to load certs and keys though. We could add a predicate maybe?

pkg/assets/assets.go
```diff
@@ -1,20 +1,18 @@
...
-	"text/template"
+	"strings"
 )
 
 type Permission os.FileMode
 
 const (
 	PermissionDirectoryDefault Permission = 0755
 	PermissionFileDefault      Permission = 0644
-	PermissionFileRestricted   Permission = 0600
```
deads2k: I think this is important to make it really easy for people to indicate their secrets.  We don't want to render 0644 if we can help it.

pkg/assets/assets.go
mfojtik: bootkube code, but yeah, I was lazy :)

pkg/assets/assets.go
mfojtik: or a pass a filename filter function, if nil then no filtering... DefaultFilter can have OWNERS and README.md

pkg/assets/assets.go
sttts: `filepath.Rel` is even better.

pkg/assets/assets.go
sttts: done

pkg/assets/assets.go
sttts: done

pkg/assets/assets.go
sttts: done. Added predicates.

pkg/assets/assets.go
sttts: done

pkg/assets/assets.go
sttts: readded

pkg/assets/assets.go
sttts: done

deads2k:
looks fine to start.

PR #32, opened 2018-09-19
add ability to copy configmap/secret from another namespace

deads2k:
/assign @mfojtik @sttts

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -124,3 +124,49 @@ func ApplySecret(client coreclientv1.SecretsGetter, required *corev1.Secret) (*c
 	actual, err := client.Secrets(required.Namespace).Update(existing)
 	return actual, true, err
 }
+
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -124,3 +124,47 @@ func ApplySecret(client coreclientv1.SecretsGetter, required *corev1.Secret) (*c
 	actual, err := client.Secrets(required.Namespace).Update(existing)
 	return actual, true, err
 }
+
+func CopyConfigMap(client coreclientv1.ConfigMapsGetter, sourceNamespace, sourceName, targetNamespace, targetName string) (*corev1.ConfigMap, bool, error) {
+	source, err := client.ConfigMaps(sourceNamespace).Get(sourceName, metav1.GetOptions{})
+	switch {
+	case apierrors.IsNotFound(err):
+		deleteErr := client.ConfigMaps(targetNamespace).Delete(targetName, nil)
```
mfojtik: if we delete it is not Copy but Sync right?

PR #35, opened 2018-09-21
Rename resourcecread to resourceread

wongma7:
the package name doesn't match the import/directory name...it confused me briefly. Is it intentional? merging this will break existing importers


deads2k:
oops

PR #42, opened 2018-10-11
Add storage to resourceread

wongma7:
for reading StorageClasses

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -1,0 +1,38 @@
+package resourceapply
+
+import (
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


wongma7:
Almost every interesting field on storageclass is immutable so i didn't think `apply` very useful. however for consistency it will be good to have, will add it to this PR tomorrow

wongma7:
@deads2k added an `Apply`, PTAL

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -0,0 +1,38 @@
...
+	contentSame := equality.Semantic.DeepEqual(existing, required)
+	if contentSame && !*modified {
+		return existing, false, nil
+	}
+
+	objectMeta := existing.ObjectMeta.DeepCopy()
+	existing = required.DeepCopy()
+	existing.ObjectMeta = *objectMeta
+
+	// if provisioner, parameters, reclaimpolicy, or volumebindingmode are different the update will fail
```
deads2k: make a TODO that would cause this to delete and recreate the storage class.

wongma7: comment amended to TODO

PR #43, opened 2018-10-11
add default value functions

deads2k:
Add default value setters for some additional types.

/assign @mfojtik


deads2k:
@soltysh I think you added the linter.  Any idea what's up with it?

soltysh:
@deads2k they've changed the download URL: fix in https://github.com/openshift/library-go/pull/44

PR #48, opened 2018-10-16
add ClusterOperator status controller

deads2k:
lifted with comments addressed from https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/20


deads2k:
approved in https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/20

PR #52, opened 2018-10-17
Add pkg/operator/render as a generic render command

sttts:
Example usage in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/53.

### DIFF
--- a/pkg/operator/render/options/generic.go
+++ b/pkg/operator/render/options/generic.go
@@ -1,0 +1,151 @@
+package options
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"text/template"
+
+	"github.com/ghodss/yaml"
+	"github.com/openshift/library-go/pkg/assets"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GenericOptions contains the generic render command options.
+type GenericOptions struct {
+	DefaultFile                   string
+	BootstrapOverrideFile         string
+	PostBootstrapOverrideFile     string
+	AdditionalConfigOverrideFiles []string
+
+	ConfigOutputFile string
+
+	TemplatesDir   string
+	AssetInputDir  string
+	AssetOutputDir string
+}
+
+type Template struct {
+	FileName string
+	Content  []byte
+}
+
+// NewGenericOptions returns a default set of generic options.
+func NewGenericOptions() *GenericOptions {
+	return &GenericOptions{
+		TemplatesDir: "/usr/share/bootkube/manifests",
+	}
+}
+
+// AddFlags adds the generic flags to the flagset.
+func (o *GenericOptions) AddFlags(fs *pflag.FlagSet, configGVK schema.GroupVersionKind) {
+	fs.StringVar(&o.AssetOutputDir, "asset-output-dir", o.AssetOutputDir, "Output path for rendered manifests.")
+	fs.StringVar(&o.AssetInputDir, "asset-input-dir", o.AssetInputDir, "A path to directory with certificates and secrets.")
+	fs.StringVar(&o.TemplatesDir, "templates-input-dir", o.TemplatesDir, "A path to a directory with manifest templates.")
+	fs.StringSliceVar(&o.AdditionalConfigOverrideFiles, "config-override-files", o.AdditionalConfigOverrideFiles,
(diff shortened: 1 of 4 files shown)


pkg/operator/render/options/generic.go
```diff
@@ -0,0 +1,127 @@
...
+	// load and render templates
+	if cfg.Assets, err = assets.LoadFilesRecursively(o.AssetInputDir); err != nil {
+		return fmt.Errorf("failed loading assets from %q: %v", o.AssetInputDir, err)
+	}
+
+	return nil
+}
+
+func (o *GenericOptions) configFromDefaultsPlusOverride(templateData interface{}, tlsOverride, bootstrapVersion string, loadAsset LoadAsset) ([]byte, error) {
+	defaultConfig := loadAsset(filepath.Join(bootstrapVersion, "kube-controller-manager", "defaultconfig.yaml"))
```
deads2k: this looks specific, not generic.

pkg/operator/render/options/generic.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// LoadAsset loads asset files.
+type LoadAsset func(name string) []byte
+
+// AddFlags adds the generic flags to the flagset.
+func (o *GenericOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.AssetOutputDir, "asset-output-dir", "", "Output path for rendered manifests.")
+	fs.StringVar(&o.AssetInputDir, "asset-input-dir", "", "A path to directory with certificates and secrets.")
+	fs.StringVar(&o.TemplatesDir, "templates-input-dir", "/usr/share/bootkube/manifests", "A path to a directory with manifest templates.")
```
deads2k: don't do defaults like this.  Default values based on current values and have a `NewOptions` method.

pkg/operator/render/options/manifest.go
```diff
@@ -0,0 +1,81 @@
...
+	ImagePullPolicy       string
+	ConfigHostPath        string
+	ConfigFileName        string
+	CloudProviderHostPath string
+	SecretsHostPath       string
+}
+
+// AddfFlags adds the manifest related flags to the flagset.
+func (o *ManifestOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.Namespace, "manifest-namespace", "openshift-kube-controller-manager",
```
deads2k: same comment about defaults and this looks very specific.

deads2k:
I'm fine with it in concept, but it needs to be generalized and apply lessons about defaults values learned from `oc`.

pkg/operator/render/options/generic.go
sttts: indeed, will fix

pkg/operator/render/options/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+	ConfigFileName        string
+	CloudProviderHostPath string
+	SecretsHostPath       string
+}
+
+// NewManifestOptions return default values for ManifestOptions.
+func NewManifestOptions(componentName string) *ManifestOptions {
+	return &ManifestOptions{
+		Namespace:             fmt.Sprintf("openshift-%s", componentName)
+		Image:                 "openshift/origin-hyperkube:latest",
```
deads2k: this image is likely to be different.  An arg seems appropriate.

pkg/operator/render/options/manifest.go
```diff
@@ -0,0 +1,95 @@
...
+		ImagePullPolicy:       "IfNotPresent",
```
deads2k: I've found myself switching these to `Always`, now that I'm not building the source locally like I did for cluster up.

deads2k:
> @deads2k addressed your comments. ptal.

two minor comments and lgtm.

deads2k:
compile error though.

pkg/operator/render/options/manifest.go
sttts: Then let's leave it empty. wdyt?

sttts: flag is more explicit maybe. Having an image set in the flag makes certainly sense. Switching to func param.

pkg/operator/render/options/manifest.go
sttts: fixed

pkg/operator/render/options/manifest.go
sttts: fixed

sttts:
Also compile error fixed.

PR #53, opened 2018-10-17
Add EnsureOperatorConfigExists helper function

damemi:
Adds a generic helper that can be used by any operator to ensure that a config exists

### DIFF
--- a/pkg/operator/v1alpha1helpers/helper_test.go
+++ b/pkg/operator/v1alpha1helpers/helper_test.go
@@ -6,7 +6,11 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/client-go/dynamic/fake"
 
 	"time"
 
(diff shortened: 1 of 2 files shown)


damemi:
@deads2k moved my changes (with your feedback) from https://github.com/openshift/cluster-openshift-apiserver-operator/pull/36 to here, once this merges I'll clean out that PR to just use the helper

pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -133,3 +142,62 @@ func SetStatusFromAvailability(status *operatorsv1alpha1.OperatorStatus, specGen
...
+		panic(err)
+	}
+
+	requiredOperatorConfig, ok := operatorConfigObj.(*unstructured.Unstructured)
+	if !ok {
+		panic(fmt.Sprintf("unexpected object in %t", operatorConfigObj))
+	}
+
+	hasImageEnvVar := false
+	if imagePullSpecFromEnv := os.Getenv("IMAGE"); len(imagePullSpecFromEnv) > 0 {
```
deads2k: put `os.Getenv` into a private helper function you can override for unit tests.

deads2k:
Unit tests please and you've got it.

damemi:
@deads2k updated and added tests. I ended up putting `os.Getenv` into a public function, my thinking being if this is used by multiple operators they can all just pass the one in this library to `EnsureOperatorConfigExists` but it's still stubbable for tests. Let me know if that sounds right

pkg/operator/v1alpha1helpers/helper_test.go
```diff
@@ -196,3 +200,103 @@ func TestRemoveOperatorCondition(t *testing.T) {
...
+}
+
+func newUnstructured(apiVersion, kind, name, imagePullSpec string) *unstructured.Unstructured {
+	return &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": apiVersion,
+			"kind":       kind,
+			"metadata": map[string]interface{}{
+				"name":      name,
+				"namespace": "",
```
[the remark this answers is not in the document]
deads2k: probably just a default serialization.  That's fine.  Empty string is correct.

pkg/operator/v1alpha1helpers/helper_test.go
```diff
@@ -196,3 +200,103 @@ func TestRemoveOperatorCondition(t *testing.T) {
...
+		gvr         schema.GroupVersionResource
+		config      []byte
+		expected    *unstructured.Unstructured
+		existingObj *unstructured.Unstructured
+	}{
+		{
+			name:   "Update operator config from env variable",
+			getEnv: func() string { return "foo" },
+			gvr:    schema.GroupVersionResource{Group: "openshiftapiserver.operator.openshift.io", Version: "v1alpha1", Resource: "openshiftapiserveroperatorconfigs"},
+			config: []byte(`apiVersion: openshiftapiserver.operator.openshift.io/v1alpha1
```
deads2k: Are all three of these the same?  if so, give me one constant?

pkg/operator/v1alpha1helpers/helper_test.go
```diff
@@ -196,3 +200,103 @@ func TestRemoveOperatorCondition(t *testing.T) {
 		})
 	}
 }
+
+func getEnv(env string) string {
```
deads2k: dead?

deads2k:
minor comments, lgtm otherwise.

PR #56, opened 2018-10-23
Add tnozicka

tnozicka:
primarily as this is mirrored into our CI and for the shared build machinery.

/cc @deads2k @sttts


tnozicka:
also with most changes I modify the main makefile making the sub owners there useless to me

tnozicka:
going for labels

PR #57, opened 2018-10-23
add zero-diff check to status controller sync method

juanvallejo:
cc @deads2k

### DIFF
--- a/pkg/operator/status/controller.go
+++ b/pkg/operator/status/controller.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/golang/glog"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


pkg/operator/status/controller.go
```diff
@@ -122,6 +125,10 @@ func (c StatusSyncer) sync() error {
 	}
 	unstructured.SetNestedSlice(operatorConfig.Object, conditions, "status", "conditions")
 
+	if reflect.DeepEqual(originalConfig, operatorConfig) {
```
deads2k: try `equalities.Semantic.DeepEqual` (apimachinery).

pkg/operator/status/controller.go
```diff
@@ -79,6 +80,8 @@ func (c StatusSyncer) sync() error {
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
+	originalConfig := operatorConfig.DeepCopy()
```
deads2k: realizing I started this pattern and we actually want the copy to be the one we modify.

juanvallejo:
@deads2k thanks, review comments addressed

deads2k:
link to a fake bump pull you've done and tested loclaly that makes sure that this works.

once that's confirmed, this lgtm

deads2k:
travis runs go test. merging

PR #60, opened 2018-10-26
add installer pod command

deads2k:
Adds the installer pod we're already using in kube-apiserver-operator

/assign @sttts


deads2k:
lgtm'd before.

PR #63, opened 2018-10-30
Basic installer controller unit tests

mfojtik:


pkg/operator/staticpod/staticpodcontroller/installer_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+					NodeName:                    "test-node-1",
+					CurrentDeploymentGeneration: 0,
+					TargetDeploymentGeneration:  0,
+				},
+			},
+		},
+		t:               t,
+		resourceVersion: "0",
+	}
+	os.Setenv("OPERATOR_IMAGE", "docker.io/foo/bar")
```
deads2k: I make an injectable function to avoid doing this.  Looking at this test, it appears you'll need to make it a private member on the controller struct.

pkg/operator/staticpod/staticpodcontroller/installer_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+	c := NewInstallerController(
+		"test",
+		[]string{"test-secret"},
```
deads2k: did I invert these?

deads2k:
@mfojtik A couple comments I definitely think we should fix.

As a followup, a test for the multi-node case would probably be good.

pkg/operator/staticpod/staticpodcontroller/installer_controller_test.go
mfojtik: i did, fixed

mfojtik:
@deads2k comments fixed, agree with multi-node scenarios follow up.

deads2k:
lgtm

PR #65, opened 2018-10-30
Add multi-node tests and fix some logical errors

mfojtik:
/cc @deads2k


deads2k:
certainly an improvement.  The test client usage is clever/odd.  I can't decide which.  units are green.

PR #68, opened 2018-10-31
Make mac with prehistoric shell happy by using the old redirection syntax

tnozicka:
@mfojtik this issue you were having with bash (shell) < 4.x

/assign @sttts


mfojtik:
/lgtm

Thanks for caring about Mac users :-)

PR #69, opened 2018-11-01
Cleanup static pod controllers and add staticpod.NewControllers

mfojtik:
This change will move the static pod controllers into individual directories as we follow this pattern in Origin and in Kubernetes and it provides nice isolation of the controller packages. 

In addition to this change, this introduces `staticpod.NewControllers(...)` function which merges initialization of all controller loops into one call. This approach allow us to add more controller loops (eg. `installer-sa` is coming) in future, without iterating over all repos.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/staticpod/controller/common/interfaces.go
+++ b/pkg/operator/staticpod/controller/common/interfaces.go
@@ -1,8 +1,9 @@
-package controller
+package common
 
 import (
-	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"k8s.io/client-go/tools/cache"
+
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 )
 
 type OperatorClient interface {
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -1,0 +1,65 @@
+package staticpod
+
+import (
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/deployment"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/node"
+)
+
+type staticPodOperatorControllers struct {
+	deploymentController *deployment.DeploymentController
+	installerController  *installer.InstallerController
+	nodeController       *node.NodeController
+}
+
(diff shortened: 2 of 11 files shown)


pkg/operator/staticpod/controller/common/interfaces.go
```diff
@@ -1,8 +1,9 @@
-package staticpodcontroller
+package common
```
sttts: common is like util is like misc. What does it contain?

pkg/operator/staticpod/controller/controller.go
```diff
@@ -0,0 +1,61 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/deployment"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/node"
+)
+
+type StaticPodOperator struct {
```
sttts: controller or operator? package is controller.

pkg/operator/staticpod/controller/common/interfaces.go
mfojtik: @sttts common operator client and common test utilities... calling things with a name corresponding to what they are ;-)

pkg/operator/staticpod/controller/controller.go
mfojtik: isn't operator just a lightweight controller? ;-)

sttts: or is an operator a set of controllers?

sttts: Should this be one package level up?

mfojtik: per slack: move one level up, and rename to `NewControllers`... which makes it look nice on call `staticpod.NewControllers(...)`

sttts:
Lgtm.

pkg/operator/staticpod/controllers.go
```diff
@@ -0,0 +1,63 @@
...
+
+	controller.nodeController = node.NewNodeController(
+		staticPodOperatorClient,
+		kubeInformersClusterScoped,
+	)
+
+	return controller
+}
+
+func (o *staticPodOperatorControllers) Run(stopCh <-chan struct{}) {
```
deads2k: This call should block.  And please give glogs

mfojtik: glog info is given by all individual controller, do you need more glog?

sure I can make it block

PR #73, opened 2018-11-02
restart installer container after error

mfojtik:
In some cases the installer container will fail when it is not able to reach the API server (network glitch, endpoints charade, etc). In that case the installer pod will fail and never retry. A manual intervention is required in that case to set the ForceRedeploy field.

That approach seems to be clunky in regards to our e2e tests when the inability to recover the installer after it lost the connection to API server is fatal for test and likely cause a lot of flakes.

This change will allow the installer pod to backoff restart its failed container 5 times. After 5 failed attempts to run, the controller will delete the pod, scrape the last termination error from it and report it back to operator. A downside of deletion could be deleting a pod with "running" container, which I think is unlikely as after 5 restart, the back-off will be >1m and we likely fit into this window...

I think this is an edge case (the API server should be always available), but this will unstuck us from network flakes?

/cc @sttts 
/cc @deads2k 

TODO: This screams for a unit test before merging.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 
@@ -26,7 +27,10 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1alpha1helpers"
 )
 
-const installerControllerWorkQueueKey = "key"
+const (
+	installerControllerWorkQueueKey = "key"
+	maxInstallerPodRestarts         = 5
+)
 
 type InstallerController struct {
 	targetNamespace string
@@ -40,6 +44,7 @@ type InstallerController struct {
 
 	operatorConfigClient common.OperatorClient
 
+	podLister  corelisterv1.PodLister
 	kubeClient kubernetes.Interface
 
 	// queue only ever has one item, but it has nice error handling backoff/retry semantics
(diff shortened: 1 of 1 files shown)


mfojtik:
@deads2k if we want to do this, then I'm thinking we can probably turn this into a `installer_pod_controller` ? It could be a follow up perhaps, I guess this change is enough to unstick us for now.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -66,6 +71,7 @@ func NewInstallerController(
 
 		operatorConfigClient: operatorConfigClient,
 		kubeClient:           kubeClient,
+		podLister:            kubeInformersForTargetNamespace.Core().V1().Pods().Lister(),
```
mfojtik: @deads2k any reason why not to use lister if we already have the informer?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -157,16 +163,19 @@ func (c *InstallerController) createInstallerController(operatorSpec *operatorv1
 // newNodeStateForInstallInProgress returns the new NodeState or error
 func (c *InstallerController) newNodeStateForInstallInProgress(currNodeState *operatorv1alpha1.NodeStatus) (*operatorv1alpha1.NodeStatus, error) {
 	ret := currNodeState.DeepCopy()
-	installerPod, err := c.kubeClient.CoreV1().Pods(c.targetNamespace).Get(getInstallerPodName(currNodeState.TargetDeploymentGeneration, currNodeState.NodeName), metav1.GetOptions{})
+	installerPod, err := c.podLister.Pods(c.targetNamespace).Get(getInstallerPodName(currNodeState.TargetDeploymentGeneration, currNodeState.NodeName))
 	if apierrors.IsNotFound(err) {
-		ret.LastFailedDeploymentGeneration = currNodeState.TargetDeploymentGeneration
-		ret.TargetDeploymentGeneration = currNodeState.CurrentDeploymentGeneration
-		ret.LastFailedDeploymentErrors = []string{err.Error()}
+		if len(ret.LastFailedDeploymentErrors) == 0 {
```
mfojtik: this is somewhat controversial... this means that if we failed because of number of restart was exceeded, we deleted the pod and already set the last failed deployment errors. However, the pod won't exists anymore.

mfojtik:
@deads2k @sttts more I'm thinking about this approach, more I don't like it... IF we go with this, we will basically never get Failed pod, but rather no pod (we will delete it after 5 restarts)... It complicates the controller logic.

I think the only possible solution is to implement retry mechanism inside the installer container. The command have to be reentrant to API server failures and retry. We should set the activeDeadlineSeconds to the installer Pod to enforce an timeout which will limit the number of retries. WDYT?

PR #74, opened 2018-11-02
retry copying content in installer command

mfojtik:
An alternative approach to https://github.com/openshift/library-go/pull/73

This will add exponential back-off to installer command. In case of failure (API server down/etc.) the command will keep retry for ~2 minutes until it fails.


mfojtik:
@deads2k govet typo fixed, retag pls./

PR #75, opened 2018-11-02
add support for controllers to expose servers

deads2k:
Controllers need to expose metrics, this makes it pretty easy.

/assign @mfojtik @sttts


deads2k:
@mfojtik @sttts this works in the downstream pulls with both optional and non-optional serving certs.  Adding a listener that kills the server when the cert changes can be a later addition.

pkg/controller/prometheusmetrics/prometheus.go
```diff
@@ -0,0 +1,66 @@
+package prometheusmetrics
```
mfojtik: nit: would call this just "metrics"... not sure we are even going to have any other metrics

pkg/controller/controllercmd/builder.go
```diff
@@ -3,15 +3,22 @@ package controllercmd
 import (
 	"fmt"
 
+	"github.com/openshift/library-go/pkg/config/configdefaults"
+
```
sttts: import order

pkg/controller/controllercmd/builder.go
```diff
@@ -64,12 +89,38 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
...
+		serverConfig.HealthzChecks = append(serverConfig.HealthzChecks, b.healthChecks...)
+
+		server, err := serverConfig.Complete(nil).New(b.componentName, genericapiserver.NewEmptyDelegate())
+		if err != nil {
+			return err
+		}
+
+		go func() {
+			server.PrepareRun().Run(stopCh)
+			glog.Fatal("server exited")
```
sttts: print the error

pkg/controller/controllercmd/builder.go
```diff
@@ -64,12 +89,38 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
...
+		}()
+	} else {
```
sttts: else if

sttts: even better: check the error case `len(b.healthChecks) > 0 && b.servingInfo == nil` first.

pkg/controller/controllercmd/cmd.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
```
sttts: order

pkg/config/serving/options.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+func ToServingOptions(servingInfo configv1.HTTPServingInfo) (*genericapiserveroptions.SecureServingOptionsWithLoopback, error) {
+	host, portString, err := net.SplitHostPort(servingInfo.BindAddress)
+	if err != nil {
+		return nil, err
+	}
+	port, err := strconv.Atoi(portString)
+	if err != nil {
+		return nil, err
```
sttts: add some context

pkg/config/serving/options.go
```diff
@@ -0,0 +1,47 @@
...
+	genericapiserveroptions "k8s.io/apiserver/pkg/server/options"
+	utilflag "k8s.io/apiserver/pkg/util/flag"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func ToServingOptions(servingInfo configv1.HTTPServingInfo) (*genericapiserveroptions.SecureServingOptionsWithLoopback, error) {
+	host, portString, err := net.SplitHostPort(servingInfo.BindAddress)
+	if err != nil {
+		return nil, err
```
sttts: add some context

pkg/config/serving/options.go
```diff
@@ -0,0 +1,47 @@
...
+	"strconv"
+
+	genericapiserveroptions "k8s.io/apiserver/pkg/server/options"
+	utilflag "k8s.io/apiserver/pkg/util/flag"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func ToServingOptions(servingInfo configv1.HTTPServingInfo) (*genericapiserveroptions.SecureServingOptionsWithLoopback, error) {
+	host, portString, err := net.SplitHostPort(servingInfo.BindAddress)
```
sttts: do we validate that BindAddress is actually an IP? Otherwise, we get a nil IP in line 24 I believe, not what is intended.

pkg/config/serving/options.go
```diff
@@ -0,0 +1,47 @@
...
+			KeyFile:  namedCert.KeyFile,
+		}
+
+		servingOptions.SNICertKeys = append(servingOptions.SNICertKeys, genericNamedCertKey)
+	}
+
+	// TODO sort out what we should do here
+	//servingOptions.HTTP2MaxStreamsPerConnection = ??
+
+	servingOptionsWithLoopback := genericapiserveroptions.WithLoopback(servingOptions)
```
sttts: WithLoopback for future testing?

pkg/controller/controllercmd/builder.go
```diff
@@ -64,12 +89,38 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
...
 	clientConfig, err := b.getClientConfig()
 	if err != nil {
 		return err
 	}
 
+	if b.servingInfo != nil {
+		kubeConfig := ""
+		if b.kubeAPIServerConfigFile != nil {
+			kubeConfig = *b.kubeAPIServerConfigFile
+		}
```
mfojtik: check that kubeConfig != "" ?

pkg/config/serving/options.go
deads2k: > WithLoopback for future testing?

seemed potentially useful and easy to call.

pkg/controller/controllercmd/builder.go
deads2k: > check that kubeConfig != "" ?

shrugs.  it's allowed to be.

pkg/config/serving/options.go
deads2k: > do we validate that BindAddress is actually an IP? Otherwise, we get a nil IP in line 24 I believe, not what is intended.

it will eventuallly.  it does not today.

deads2k:
comments addressed.

Issue #71, opened 2018-11-01
Switch staticpod unit tests to use validateActions instead of messing with reactors


mfojtik:
A follow-up for https://github.com/openshift/library-go/pull/70

https://github.com/openshift/service-serving-cert-signer/blob/master/pkg/controller/apiservicecabundle/apiservice_cabundle_controller_test.go#L24

PR #76, opened 2018-11-05
refactor staticpod unit tests

mfojtik:
Refactors backing resource and deployment unit tests to evaluate test client actions instead of mocking with test client reactors. 

I left the installer test as it was, since it needs to do more trickery to kubeclient than other tests.

Fixes: https://github.com/openshift/library-go/issues/71

/cc @deads2k

### DIFF
--- a/pkg/operator/staticpod/controller/backingresource/backing_resource_controller.go
+++ b/pkg/operator/staticpod/controller/backingresource/backing_resource_controller.go
@@ -144,7 +144,7 @@ func (c BackingResourceController) sync() error {
 			glog.Error(updateError)
 		}
 	}
-	return err
+	return fmt.Errorf("synthetic requeue (errs: %q)", strings.Join(errors, ","))
 
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/backingresource/backing_resource_controller.go
```diff
@@ -144,7 +144,7 @@ func (c BackingResourceController) sync() error {
 			glog.Error(updateError)
 		}
 	}
-	return err
+	return fmt.Errorf("synthetic requeue (errs: %q)", strings.Join(errors, ","))
```
mfojtik: @deads2k unit tests are useful :-) this will report nil and cause the controller not to resync when the resource creation fail.

pkg/operator/staticpod/controller/deployment/deployment_controller_test.go
```diff
@@ -1,145 +1,173 @@
...
+				if !hasConfig {
+					t.Errorf("expected config to be created")
+					return
+				}
+				if config.Name != "test-config-1" {
+					t.Errorf("expected config to have name 'test-config-1', got %q", config.Name)
+				}
+				secret, hasSecret := createdObjects[1].(*v1.Secret)
+				if !hasSecret {
+					t.Errorf("expected secret to be created")
```
sttts: return

PR #77, opened 2018-11-05
Add helpers to read installconfig from file/cluster config

damemi:
Lifts code from https://github.com/openshift/machine-config-operator/pull/47 to provide a generalized approach for operators to access the cluster's installconfig from the cluster configmap file available on the system.

Will be helpful for fixing https://github.com/openshift/cluster-kube-apiserver-operator/issues/83, for example

### DIFF
--- a/pkg/operator/v1alpha1helpers/helpers.go
+++ b/pkg/operator/v1alpha1helpers/helpers.go
@@ -8,14 +8,17 @@ import (
 
 	"github.com/ghodss/yaml"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
 
 	operatorsv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	installertypes "github.com/openshift/installer/pkg/types"
 )
 
 func SetErrors(versionAvailability *operatorsv1alpha1.VersionAvailability, errors ...error) {
@@ -206,4 +209,27 @@ func EnsureOperatorConfigExists(client dynamic.Interface, operatorConfigBytes []
 			panic(err)
 		}
 	}
+}
+
+// InstallConfigFromFile parses the InstallConfig from the cluster ConfigMap file data
+func InstallConfigFromFile(configMapData []byte) (installertypes.InstallConfig, error) {
+	obji, err := runtime.Decode(scheme.Codecs.UniversalDecoder(corev1.SchemeGroupVersion), configMapData)
+	if err != nil {
+		return installertypes.InstallConfig{}, err
+	}
+	configMap, ok := obji.(*corev1.ConfigMap)
+	if !ok {
+		return installertypes.InstallConfig{}, fmt.Errorf("expected *corev1.ConfigMap found %T", obji)
+	}
+
+	icData, ok := configMap.Data["install-config"]
+	if !ok {
+		return installertypes.InstallConfig{}, fmt.Errorf("install-config doesn't exist")
+	}
+
+	var ic installertypes.InstallConfig
+	if err := yaml.Unmarshal([]byte(icData), &ic); err != nil {
+		return installertypes.InstallConfig{}, err
+	}
+	return ic, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -207,3 +210,38 @@ func EnsureOperatorConfigExists(client dynamic.Interface, operatorConfigBytes []
 		}
 	}
 }
+
+type InstallConfigGetter func() (installertypes.InstallConfig, error)
```
sttts: Do we really need this type abstraction?

pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -207,3 +210,38 @@ func EnsureOperatorConfigExists(client dynamic.Interface, operatorConfigBytes []
...
+
+		return InstallConfigFromClusterConfig(cm)
+	}
+}
+
+// InstallConfigFromClusterConfig gets the install config from the cluster configmap
+func InstallConfigFromClusterConfig(configMap *corev1.ConfigMap) (installertypes.InstallConfig, error) {
+	var (
+		icKey = "install-config"
+		ic    installertypes.InstallConfig
```
sttts: it's more Go-like to declare variable directly where they are needed instead of at the top of the func.

pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -207,3 +210,38 @@ func EnsureOperatorConfigExists(client dynamic.Interface, operatorConfigBytes []
...
+	)
+	icData, ok := configMap.Data[icKey]
+	if !ok {
+		return ic, fmt.Errorf("%s doesn't exist", icKey)
+	}
+
+	if err := yaml.Unmarshal([]byte(icData), &ic); err != nil {
+		return ic, err
```
sttts: return nil, err

pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -207,3 +210,38 @@ func EnsureOperatorConfigExists(client dynamic.Interface, operatorConfigBytes []
 		}
 	}
 }
+
+type InstallConfigGetter func() (installertypes.InstallConfig, error)
+
+// InstallConfigFromFile gets the install config from the cluster configmap file data
+func InstallConfigFromFile(configMapData []byte) InstallConfigGetter {
```
sttts: is this function really worth to be exposed? In an operator you will most probably use a typed ConfigMap client instead of moving around `[]byte`.

pkg/operator/v1alpha1helpers/helpers.go
damemi: I kept it straight from what was in MCO, it seems like it just serves for [test stubbing](https://github.com/abhinavdahiya/machine-config-operator/blob/55872cc024432b18d687a8010d30456fe5ad9778/pkg/operator/render_test.go#L86). Was hoping to carry it here to try not and totally break their existing test but I'm fine with getting rid of it

pkg/operator/v1alpha1helpers/helpers.go
damemi: This is for reading the ConfigMap from a file, can we do that with a client?

sttts: Won't you have the install config as yaml directly in a file, not the config map?

damemi: No, this is intending to use the cluster configmap (which has the install config and network config as nested data) which is already available on the system and generated here in the installer: https://github.com/openshift/installer/blob/master/pkg/asset/manifests/operators.go#L66

We could write the install config to be written as its own file that can be read directly, but imo that seems redundant when we have the data already in the form of the configmap which is going to be using it anyway.

pkg/operator/v1alpha1helpers/helpers.go
```diff
@@ -8,14 +8,17 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes/scheme"
 
 	operatorsv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	installertypes "github.com/openshift/installer/pkg/types"
```
deads2k: you may not depend on this rep.

damemi: Are you sure? It makes sense to avoid circular dependencies but it's already a dependency in [machine-config-operator](https://github.com/openshift/machine-config-operator/blob/master/pkg/operator/render.go#L11) and library-go doesn't seem to be imported anywhere in installer right now, is there another reason for not depending on installer? If so it should be noted in the readme next to kubernetes/origin

sttts: The installer is a possible consumer of library-go. We would get a circular dependency. This having said: we only need the API type, right? So we have two choices:

1. move the install config API type to github.com/openshift/api.
2. handle the install config as pure JSON here, i.e. without dependency on the Golang types from the installer repo.

sttts:
@damemi I think we can close this, right?

PR #79, opened 2018-11-09
build: fix GO_BUILD_PACKAGES to build binaries

mfojtik:
This fix will build all commands that are represented as directories under `./cmd` instead of doing `./cmd/...` which does not result into any binary.

/cc @tnozicka 
/cc @soltysh

### DIFF
--- a/alpha-build-machinery/make/lib/golang.mk
+++ b/alpha-build-machinery/make/lib/golang.mk
@@ -8,7 +8,7 @@ GO_PACKAGES ?=./...
 GO_PACKAGES_EXPANDED ?=$(GOLIST) $(GO_PACKAGES)
 GO_TEST_PACKAGES ?=$(GO_PACKAGES)
 
-GO_BUILD_PACKAGES ?=./cmd/...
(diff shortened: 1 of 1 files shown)


alpha-build-machinery/make/lib/golang.mk
```diff
@@ -8,7 +8,7 @@ GO_PACKAGES ?=./...
 GO_PACKAGES_EXPANDED ?=$(GOLIST) $(GO_PACKAGES)
 GO_TEST_PACKAGES ?=$(GO_PACKAGES)
 
-GO_BUILD_PACKAGES ?=./cmd/...
+GO_BUILD_PACKAGES ?=$(shell find ./cmd -mindepth 1 -maxdepth 1 -print)
```
sttts: `ls -1 ./cmd`

mfojtik: @sttts does it make difference? :)

mfojtik:
@deads2k @sttts thought we disabled verify-deps for now until we fix the diff in glide ?

sttts:
Merging manually.

PR #80, opened 2018-11-09
build: make verify-deps work on osx

mfojtik:
```
COMPATIBILITY
     Historic versions of the cp utility had a -r option.  This implementation supports that option; however, its use is strongly discouraged, as it does not cor-
     rectly copy special files, symbolic links, or fifo's.
```


sttts:
Merging manually.

PR #81, opened 2018-11-09
alpha-build-machinery: fix verify-deps

sttts:
- ~~use compatible `cp -R` instead of `cp -r`.~~ merged in @mfojtik's PR.
- gitignore `.hg_archival.txt` files as there content depends on the hg version installed
- make `verify-deps` make target ignore gitignored files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ origin.iml
 *.go~
 tags
 .envrc
+.hg_archival.txt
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -18,3 +18,4 @@ origin.iml
 *.go~
 tags
```
mfojtik: can you nuke this while you are here? caused pain in origin repo ;-)

PR #82, opened 2018-11-13
Implement operator installer revision history pruning controller

damemi:
JIRA: https://jira.coreos.com/browse/MSTR-221

~~Depends on https://github.com/openshift/api/pull/128~~
Now depends on https://github.com/openshift/library-go/pull/141

Prunes installer revision history based on a specified limit in the `OperatorSpec`, with `0` being unlimited.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang/glog"
-	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -25,6 +24,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/bindata"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -38,12 +38,15 @@ const (
 
 	hostResourceDirDir = "/etc/kubernetes/static-pod-resources"
 	hostPodManifestDir = "/etc/kubernetes/manifests"
+
+	revisionLabel       = "revision"
+	statusConfigMapName = "revision-status"
 )
 
 // InstallerController is a controller that watches the currentRevision and targetRevision fields for each node and spawn
 // installer pods to update the static pods on the master nodes.
 type InstallerController struct {
-	targetNamespace, staticPodName string
+	targetNamespace, staticPodName, podResourcePrefix string
 	// configMaps is the list of configmaps that are directly copied.A different actor/controller modifies these.
 	// the first element should be the configmap that contains the static pod manifest
 	configMaps []string
@@ -77,11 +80,9 @@ const (
 	staticPodStateFailed
 )
 
-const revisionLabel = "revision"
-
-// NewBackingResourceController creates a new installer controller.
+// NewInstallerController creates a new installer controller.
 func NewInstallerController(
-	targetNamespace, staticPodName string,
+	targetNamespace, staticPodName, podResourcePrefix string,
 	configMaps []string,
 	secrets []string,
 	command []string,
@@ -91,11 +92,12 @@ func NewInstallerController(
 	eventRecorder events.Recorder,
 ) *InstallerController {
 	c := &InstallerController{
-		targetNamespace: targetNamespace,
-		staticPodName:   staticPodName,
-		configMaps:      configMaps,
-		secrets:         secrets,
-		command:         command,
+		targetNamespace:   targetNamespace,
+		staticPodName:     staticPodName,
+		podResourcePrefix: podResourcePrefix,
+		configMaps:        configMaps,
+		secrets:           secrets,
+		command:           command,
 
 		operatorConfigClient: operatorConfigClient,
 		kubeClient:           kubeClient,
@@ -189,6 +191,23 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 		return false, nil
 	}
 
+	// Store success/fail status in the status configmap for revision history tracking
+	statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(statusConfigMapName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		statusConfigMapObject := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Namespace: c.targetNamespace, Name: statusConfigMapName},
+			Data: map[string]string{
+				"phase": "InProgress",
+			},
+		}
+		statusConfigMap, _, err = resourceapply.ApplyConfigMap(c.kubeClient.CoreV1(), c.eventRecorder, statusConfigMapObject)
+		if err != nil {
+			return true, err
+		}
+	}
+
+	phaseStatus := string(corev1.PodSucceeded)
+
 	// start with node which is in worst state (instead of terminating healthy pods first)
 	startNode, err := nodeToStartRevisionWith(c.getStaticPodState, operatorStatus.NodeStatuses)
 	if err != nil {
@@ -206,6 +225,9 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 			prevNodeState = &operatorStatus.NodeStatuses[prev]
 		}
 
+		statusConfigMap.Name = statusConfigMapNameForRevision(currNodeState.CurrentRevision)
+		statusConfigMap.Data["revision"] = fmt.Sprintf("%d", currNodeState.CurrentRevision)
+
 		// if we are in a transition, check to see if our installer pod completed
 		if currNodeState.TargetRevision > currNodeState.CurrentRevision {
 			if err := c.ensureInstallerPod(currNodeState.NodeName, operatorSpec, currNodeState.TargetRevision); err != nil {
@@ -218,6 +240,10 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 			newCurrNodeState, err := c.newNodeStateForInstallInProgress(currNodeState, pendingNewRevision)
 			if err != nil {
 				return true, err
+			}
+
+			if newCurrNodeState.LastFailedRevision != 0 {
+				phaseStatus = string(corev1.PodFailed)
 			}
 
 			// if we make a change to this status, we want to write it out to the API before we commence work on the next node.
@@ -263,6 +289,12 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 			return false, nil
 		}
 		break
+	}
+
+	statusConfigMap.Data["phase"] = phaseStatus
+	_, _, err = resourceapply.ApplyConfigMap(c.kubeClient.CoreV1(), c.eventRecorder, statusConfigMap)
+	if err != nil {
+		return true, err
 	}
 
 	return false, nil
@@ -549,4 +581,8 @@ func (c *InstallerController) eventHandler() cache.ResourceEventHandler {
 
 func mirrorPodNameForNode(staticPodName, nodeName string) string {
 	return staticPodName + "-" + nodeName
+}
+
+func statusConfigMapNameForRevision(revision int32) string {
+	return fmt.Sprintf("%s-%d", statusConfigMapName, revision)
 }
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -9,12 +9,14 @@ import (
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/node"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/prune"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/revision"
 )
 
 type staticPodOperatorControllers struct {
 	revisionController       *revision.RevisionController
 	installerController      *installer.InstallerController
+	pruneController          *prune.PruneController
 	nodeController           *node.NodeController
 	serviceAccountController *backingresource.BackingResourceController
 }
@@ -27,7 +29,7 @@ type staticPodOperatorControllers struct {
 //    appears that doesn't match the current latest for first kubeletStatus and the first kubeletStatus isn't already transitioning,
 //    it kicks off an installer pod.  If the next kubeletStatus doesn't match the immediate prior one, it kicks off that transition.
 // 3. NodeController - watches nodes for master nodes and keeps the operator status up to date
-func NewControllers(targetNamespaceName, staticPodName string, command, revisionConfigMaps, revisionSecrets []string,
+func NewControllers(targetNamespaceName, staticPodName, podResourcePrefix string, installerCommand, prunerCommand, revisionConfigMaps, revisionSecrets []string,
 	staticPodOperatorClient common.OperatorClient, kubeClient kubernetes.Interface, kubeInformersNamespaceScoped,
 	kubeInformersClusterScoped informers.SharedInformerFactory, eventRecorder events.Recorder) *staticPodOperatorControllers {
 	controller := &staticPodOperatorControllers{}
@@ -45,12 +47,22 @@ func NewControllers(targetNamespaceName, staticPodName string, command, revision
 	controller.installerController = installer.NewInstallerController(
 		targetNamespaceName,
 		staticPodName,
+		podResourcePrefix,
 		revisionConfigMaps,
 		revisionSecrets,
-		command,
+		installerCommand,
 		kubeInformersNamespaceScoped,
 		staticPodOperatorClient,
 		kubeClient,
+		eventRecorder,
+	)
+
+	controller.pruneController = prune.NewPruneController(
+		targetNamespaceName,
+		podResourcePrefix,
+		prunerCommand,
+		kubeClient,
+		staticPodOperatorClient,
 		eventRecorder,
 	)
 
@@ -75,6 +87,7 @@ func (o *staticPodOperatorControllers) Run(stopCh <-chan struct{}) {
 	go o.serviceAccountController.Run(1, stopCh)
 	go o.revisionController.Run(1, stopCh)
 	go o.installerController.Run(1, stopCh)
+	go o.pruneController.Run(1, stopCh)
 	go o.nodeController.Run(1, stopCh)
 
 	<-stopCh


damemi:
cc @mfojtik

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -244,6 +255,7 @@ func (c *InstallerController) createInstallerPod(currNodeState *operatorv1alpha1
 		fmt.Sprintf("--pod=%s", c.configMaps[0]),
 		fmt.Sprintf("--resource-dir=%s", "/etc/kubernetes/static-pod-resources"),
 		fmt.Sprintf("--pod-manifest-dir=%s", "/etc/kubernetes/manifests"),
+		fmt.Sprintf("--revision-history-limit=%d", operatorSpec.RevisionHistoryLimit),
```
deads2k: I wouldn't expect an installer pod to know this.  I would expect a separate binary to deal with pruning

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -77,6 +78,7 @@ func (o *InstallOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringSliceVar(&o.ConfigMapNamePrefixes, "configmaps", o.ConfigMapNamePrefixes, "list of configmaps to be included")
 	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
 	fs.StringVar(&o.PodManifestDir, "pod-manifest-dir", o.PodManifestDir, "directory for the static pod manifest")
+	fs.IntVar(&o.RevisionHistoryLimit, "revision-history-limit", 0, "limit on installer pod attempts to preserve, 0=unlimited")
```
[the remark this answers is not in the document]
deads2k: > "number of deployments to keep on disk" maybe?

I don't think the reaping pod should be making the choice. I think it should be given a revision number of `maxEligibleRevisionID` and a list of `protectedRevisionID` and it deletes every unprotected eligible revision.

pkg/operator/staticpod/controller/installer/installer_controller.go
damemi: Why can't the installer handle cleaning up old installations? It created them so I don't think it's unreasonable for it to prune them too

damemi:
@deads2k updated this, could you take another look?

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+		return err
+	}
+	return nil
+}
+
+func getPrunerPodName(podName, nodeName string) string {
+	return fmt.Sprintf("prune-%s-%s", podName, nodeName)
+}
+
+func (c *PruneController) pruneRevisionHistory(operatorStatus *operatorv1.StaticPodOperatorStatus) error {
```
deads2k: nit: move the action above the boilerplate

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, nodeStatus := range operatorStatus.NodeStatuses {
+		// Read status config map to figure which deployments to keep
+		statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(getStatusConfigMapName(nodeStatus.TargetRevision, nodeStatus.NodeName), metav1.GetOptions{})
```
deads2k: Yeah, only one of these.

deads2k: > Yeah, only one of these.

one per revision

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -266,6 +266,17 @@ func (c *InstallerController) newNodeStateForInstallInProgress(currNodeState *op
 	failed := false
 	errors := []string{}
 
+	// Store status of this deployment in the status configmap for revision history tracking
+	statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(getStatusConfigMapName(currNodeState.TargetRevision, currNodeState.NodeName), metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	statusConfigMap.Data[strconv.Itoa(int(currNodeState.TargetRevision))] = string(installerPod.Status.Phase)
```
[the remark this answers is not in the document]
deads2k: > Data["nodeName.phase"]

Actually, just `Data["phase"]`

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+
+		for revisionID, revisionStatus := range statusConfigMap.Data {
+			switch revisionStatus {
+			case string(corev1.PodSucceeded):
+				succeededRevisionIDs = append(succeededRevisionIDs, revisionID)
+			case string(corev1.PodFailed):
+				failedRevisionIDs = append(failedRevisionIDs, revisionID)
+			}
+		}
+		sort.Strings(succeededRevisionIDs)
```
deads2k: this looks likely to sort improperly.  Add test to expose it.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	required.Namespace = c.targetNamespace
+	required.Spec.NodeName = nodeName
+	required.Spec.Containers[0].Image = c.prunerPodImageFn()
+	required.Spec.Containers[0].Command = c.command
+	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
+		fmt.Sprintf("-v=%d", 4),
+		fmt.Sprintf("--max-eligible-id=%s", maxEligibleRevision),
+		fmt.Sprintf("--protected-ids=%s", strings.Join(protectedRevisions, ",")),
+		fmt.Sprintf("--resource-dir=%s", "/etc/kubernetes/static-pod-resources"),
+		fmt.Sprintf("--pod-manifest-dir=%s", "/etc/kubernetes/manifests"),
```
deads2k: didn't expect to see this.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+		fmt.Sprintf("--static-pod-name=%s", c.staticPodName),
```
deads2k: also looks odd

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	return true
+}
+
+func (c *PruneController) sync() error {
+	_, operatorStatus, _, err := c.operatorConfigClient.Get()
+	if err != nil {
+		return err
+	}
+
+	err = c.pruneRevisionHistory(operatorStatus)
```
deads2k: I don't see where you delete the configmaps from the API

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is a directory AND the revision is not protected
+		// AND it is a revision of the pod we want AND it is less than or equal to the max eligible revision
+		if _, protected := protectedIDs[file.Name()]; !protected &&
+			file.IsDir() &&
+			strings.HasPrefix(file.Name(), o.StaticPodName) &&
+			strings.Compare(file.Name(), o.nameForRevision(o.MaxEligibleRevisionID)) <= 0 {
```
deads2k: alpha sort on numbers again.

pkg/operator/staticpod/controller/installer/installer_controller.go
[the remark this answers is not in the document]
damemi: @deads2k I think we should clarify the structure of the configmap we're using. Are you saying that each revision's configmap should only hold the phase for that revision? It seems like that would take a lot more work to build the list of successful/failed revisions if we can only get one data point from each revision's configmap.

What I'm trying to do is have each revision's configmap hold every other revision's statuses. The data would be copied over and updated by the InstallerController each time a new revision is made, and the PruneController can also clean up the configmap so it doesn't get massive. Then it only takes one api call to get the info (but we still need to sort it)

I don't know how we could use one data point per configmap without going back and loading in each revision's configmap data (from file?) If this isn't actually that expensive of an operation I'm fine with it, and it would let us sort the statuses in-place as we read them. I just want to make sure I understand your approach. Thoughts?

pkg/operator/staticpod/controller/prune/prune_controller.go
damemi: @deads2k can we trust that these directories are consistent enough that they can be hardcoded into the library code?

pkg/operator/staticpod/controller/prune/prune_controller.go
damemi: @deads2k similar question then, if this is being used by multiple operators this was just how I thought we could pass the name in

pkg/operator/staticpod/controller/prune/prune_controller.go
soltysh: I'm inclined to say we should move the sorting operation into `pruneEligibaleIDs` to get rid of that requirement from callers. You don't care if it is sorted here, either.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	required := resourceread.ReadPodV1OrDie([]byte(prunerPod))
+	required.Name = getPrunerPodName(podName, nodeName)
+	required.Namespace = c.targetNamespace
+	required.Spec.NodeName = nodeName
+	required.Spec.Containers[0].Image = c.prunerPodImageFn()
+	required.Spec.Containers[0].Command = c.command
+	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
+		fmt.Sprintf("-v=%d", 4),
+		fmt.Sprintf("--max-eligible-id=%s", maxEligibleRevision),
+		fmt.Sprintf("--protected-ids=%s", strings.Join(protectedRevisions, ",")),
```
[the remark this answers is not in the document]
soltysh: > This looks at risk of exceeding the CLI limit.

What limit you're talking about, the OS one is pretty high and I don't see any in cobra library.

pkg/operator/staticpod/controller/prune/prune_controller.go
soltysh: Don't we use `0001` etc numbering?

pkg/operator/staticpod/controller/prune/prune_controller.go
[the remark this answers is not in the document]
damemi: I agree with @deads2k that this could probably be handled better, with very high history limits we could be passing a lot here (at least with the current approach which runs 2 pods: one to prune successful revisions, where *all* failed revisions are "protected" followed by a pod to prune failed revisions, where *all* successful revisions are "protected")

pkg/operator/staticpod/controller/prune/prune_controller.go
soltysh: In that case David's concert about sorting numbers with alpha is critical.

pkg/operator/staticpod/controller/prune/prune_controller.go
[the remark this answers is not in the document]
soltysh: How many protected you're imagining we'll have?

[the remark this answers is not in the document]
damemi: The max would be the history limit set by the operator, if the pruner is running successfully. So it could be 10, if the history limit is 10, but if the history limit is 50 then we're passing a lot more. I don't know what, if any, the CLI limit is though

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -66,6 +66,7 @@ const (
 )
 
 const revisionLabel = "revision"
+const statusConfigMapName = "revision-status"
```
mfojtik: nit: group consts

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&o.MaxEligibleRevisionID, "max-eligible-id", o.MaxEligibleRevisionID, "highest revision ID to be eligible for pruning")
+	fs.IntSliceVar(&o.ProtectedRevisionIDs, "protected-ids", o.ProtectedRevisionIDs, "list of revision IDs to reserve from being pruned")
+	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
+	fs.StringVar(&o.StaticPodName, "static-pod-name", o.StaticPodName, "name of the static pod")
+}
+
+func (o *PruneOptions) Validate() error {
```
mfojtik: do we need to validate o.MaxEligibleRevisionID ?

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		if err != nil {
+			return err
+		}
+		if info.IsDir() && strings.HasPrefix(info.Name(), c.staticPodName) {
+			// Read the phase configmap for this revision and sort into failed or succeeded
+			phase, err := ioutil.ReadFile(path.Join(resourceDir, configMapDir))
+			if err != nil {
+				return err
+			}
+			switch fmt.Sprintf("%s", phase) {
```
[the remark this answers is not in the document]
damemi: need to cast the `bytes[]` to a string though to switch on it right? what about `switch string(phase)`

damemi:
@mfojtik updated with all your feedback (and also added tests)

@deads2k could you double check that the logic for updating the configmap looks right?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -179,6 +183,25 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 		return false, nil
 	}
 
+	// Store success/fail status in the status configmap for revision history tracking
+	statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(statusConfigMapName, metav1.GetOptions{})
```
deads2k: this is one per revision, right?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -179,6 +183,25 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
...
+	if apierrors.IsNotFound(err) {
+		statusConfigMapObject := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Namespace: c.targetNamespace, Name: statusConfigMapName},
+			Data: map[string]string{
+				"phase": "pending",
```
deads2k: `InProgress`?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -264,6 +291,12 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 		break
 	}
 
+	statusConfigMap.Data["phase"] = phaseStatus
```
deads2k: I expected to see a helper for updating status for a revision that interpreted the status update and decided what the phase should be .

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+)
+
+type PruneController struct {
+	targetNamespace, staticPodName              string
```
deads2k: project standard is a new line for each member

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	return c
+}
+
+func (c *PruneController) pruneRevisionHistory(resourceDir, configMapDir string) error {
+	var succeededRevisionIDs, failedRevisionIDs []string
+
+	// Walk the resource directory looking for files that match our static pod name prefix
+	err := filepath.Walk(resourceDir, func(currentPath string, info os.FileInfo, err error) error {
```
deads2k: The controller has no access to the filesystem where these defintions live.

deads2k: > The controller has no access to the filesystem where these defintions live.

seems like this would come from inspecting the configmaps via the API

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Make sure the lists are sorted in a way that accounts for numbering
+	sort.Sort(sortorder.Natural(succeededRevisionIDs))
+	sort.Sort(sortorder.Natural(failedRevisionIDs))
+
+	// Get list of protected IDs
+	protectedSucceededRevisionIDs := succeededRevisionIDs[len(succeededRevisionIDs)-c.succeededRevisionLimit:]
```
deads2k: What do negatives do?

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	protectedFailedRevisionIDs := failedRevisionIDs[len(failedRevisionIDs)-c.failedRevisionLimit:]
+	excludedIDs := append(protectedSucceededRevisionIDs, protectedFailedRevisionIDs...)
```
deads2k: this does weird things in memory.  New slice and append both.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	required.Spec.Containers[0].Command = c.command
+	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
+		fmt.Sprintf("-v=%d", 4),
+		fmt.Sprintf("--max-eligible-id=%s", maxEligibleRevision),
+		fmt.Sprintf("--protected-ids=%s", strings.Join(protectedRevisions, ",")),
+		fmt.Sprintf("--resource-dir=%s", defaultResourceDir),
+		fmt.Sprintf("--static-pod-name=%s", c.staticPodName),
+	)
+
+	if _, err := c.kubeClient.CoreV1().Pods(c.targetNamespace).Create(required); err != nil && !apierrors.IsAlreadyExists(err) {
```
deads2k: I expected you to need to pin this to the "correct" node and run three of them.

deads2k: > I expected you to need to pin this to the "correct" node and run three of them.

and run one for every node

pkg/operator/staticpod/controllers.go
```diff
@@ -47,13 +50,23 @@ func NewControllers(targetNamespaceName, staticPodName string, command, revision
...
 		kubeClient,
 		eventRecorder,
 	)
 
+	controller.pruneController = prune.NewPruneController(
+		targetNamespaceName,
+		staticPodName,
+		prunerCommand,
+		kubeClient,
+		failedRevisionLimit,
```
deads2k: just hard code this in the operator for now.  Later you can pull it from the API

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+
+	files, err := ioutil.ReadDir(o.ResourceDir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is a directory AND the revision is not protected
+		// AND it is a revision of the pod we want AND it is less than or equal to the max eligible revision
+		if _, protected := protectedIDs[file.Name()]; !protected &&
```
deads2k: isn't the key for `protetedIDs` just a number?

deads2k:
Let's start breaking this into pieces.  Can you separate the code that creates the pruning command itself?

pkg/operator/staticpod/prune/cmd.go
damemi: > isn't the key for protetedIDs just a number?
it was originally, but now the list of protected IDs is passed to the command with their full name (eg `kube-apiserver-pod-1`) which should match the directory for the revision. I build this into a `map[string]struct{}` above for easier lookup

damemi:
@deads2k made some updates: moved everything regarding the new prune command to its own commit, and squashed down all prior feedback as well. 

I've kept the changes from the current round of review to their own commits to try and preserve the comments in a way that makes sense, but I can just squash those down as well if it's easier.

At this point I've gotten the pruner to work in manual testing (with 10+ successful revisions, seeing the right older ones deleted from `/etc/kubernetes/static-pod-resources`). However it seems that updating the `revision-status` configmap is itself kicking off a new revision, causing a constant loop. I'm not sure how to fix that, but it did make testing a lot easier. Probably not what we want though.

Still need to clean up a couple things including clearing out old resources from the namespace itself (like old revision-status configmaps) as well as testing with some failed revisions as well but this is progress

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -179,6 +184,23 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 		return false, nil
 	}
 
+	// Store success/fail status in the status configmap for revision history tracking
+	statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(statusConfigMapName, metav1.GetOptions{})
```
deads2k: still looks like the wrong multiplicity

damemi: This can probably be reworked, but I don't have the revision number until I start looking at the node statuses, which is where I update the configmap name to be unique for that revision: https://github.com/openshift/library-go/pull/82/files#diff-aea28fda19fe8a3c27c2cf6590e72d89R221

Then once outside the loop I save the configmap

pkg/operator/staticpod/controller/prune/prune_controller.go
damemi: added here: https://github.com/openshift/library-go/pull/82/files#diff-5c9aefda4344c1bce9b242c5c5f5d82eR114

damemi:
Added more changes to hopefully be able to use the dynamic client with unstructured to get revision limits from the operator config instead of hardcoding (https://github.com/openshift/library-go/pull/82/commits/d8fb037b2bc73b1adc6a56187e4660051c17c1fd)

and the matching changes in kube-apiserver-operator: https://github.com/openshift/cluster-kube-apiserver-operator/pull/154/commits/c4df34954e67b4775b200f997705793e31e1e1a5

pkg/operator/staticpod/controller/installer/installer_controller.go
damemi: We want to look at the status update for each node and set phase to `failed` if any nodes are failed, otherwise set it as `success`, right? If that's the approach then can't I just interpret it here while we're already working on each node's status update?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -77,11 +80,9 @@ const (
 	staticPodStateFailed
 )
 
-const revisionLabel = "revision"
-
-// NewBackingResourceController creates a new installer controller.
+// NewInstallerController creates a new installer controller.
 func NewInstallerController(
-	targetNamespace, staticPodName string,
+	targetNamespace, staticPodName, podResourcePrefix string,
```
deads2k: project standard is separate lines

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -38,12 +38,15 @@ const (
...
+
+	revisionLabel       = "revision"
+	statusConfigMapName = "revision-status"
 )
 
 // InstallerController is a controller that watches the currentRevision and targetRevision fields for each node and spawn
 // installer pods to update the static pods on the master nodes.
 type InstallerController struct {
-	targetNamespace, staticPodName string
+	targetNamespace, staticPodName, podResourcePrefix string
```
deads2k: project standard is separate lines

deads2k:
please open a new pull

PR #83, opened 2018-11-14
add generic way to observe changes in files and react to them

mfojtik:
This use sigma/go-inotify which is in pair with what kube is using. Only works on Linux and I don't think we need this to support other architectures.

The idea is to define a file observer and register "reaction" function associated with paths you want to react upon on. A simple example might be:

```go
o, _ := NewObserver()
o.AddReactor(func(string) error { os.Exit(0) }, "/foo/var/cert.crt")
// ...
go o.Run(stopCh)
```

Doing this means if the file is either created, modified or deleted this observer will react with exiting the process and picking up the updated cert.


pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	watcher   *inotify.Watcher
+	reactors  map[string][]reactorFn
+	isRunning bool
+}
+
+func init() {
+	NewObserver = NewLinuxObserver
+}
+
+func NewLinuxObserver() (Observer, error) {
```
sttts: make it private? Also the type.

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	w, err := inotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	return &ObserverLinux{watcher: w}, nil
+}
+
+func (o *ObserverLinux) AddReactor(reactor reactorFn, filenames ...string) Observer {
+	for _, f := range filenames {
+		glog.V(5).Infof("adding file %q", f)
```
sttts: observing file

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,68 @@
...
+		o.watcher.AddWatch(f, inotify.IN_CREATE|inotify.IN_DELETE|inotify.IN_MODIFY)
```
sttts: only call AddWatch if not yet registered.

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return o
+}
+
+func (o *ObserverLinux) Run(stopCh <-chan struct{}) {
+	if o.isRunning {
+		panic("an instance of file observer is already running")
+	}
+	o.isRunning = true
+	glog.Info("starting file observer")
```
sttts: capital case

pkg/controller/fileobserver/observer_linux.go
mfojtik: @sttts what if you want multiple reaction to a single file? ;-)

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,69 @@
...
+	go func() {
+		select {
+		case <-stopCh:
+			glog.Infof("Shutting down file observer")
+			o.watcher.Close()
+		}
+	}()
+
+	for event := range o.watcher.Event {
+		if reactorFns, hasReactor := o.reactors[event.Name]; hasReactor {
```
sttts: continue early

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,69 @@
...
+		o.reactors[f] = append(o.reactors[f], reactor)
+	}
+	return o
+}
+
+func (o *linuxObserver) Run(stopCh <-chan struct{}) {
+	if o.isRunning {
+		panic("an instance of file observer is already running")
+	}
+	o.isRunning = true
```
sttts: reset when stopCh is closed?

pkg/controller/fileobserver/observer_default.go
```diff
@@ -0,0 +1,23 @@
...
+type defaultObserver struct{}
+
+func init() {
+	NewObserver = func() (Observer, error) {
+		return &defaultObserver{}, nil
+	}
+}
+
+func (*defaultObserver) Run(stopChan <-chan struct{}) {
+	glog.Warning("file observer is not supported on this platform")
```
sttts: capital case. What about panic?

mfojtik: @sttts yeah, I had it there before and I think panic() is the right behavior here... noone should run this outside containers anyway.

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,72 @@
...
+		if !hasReactor {
+			continue
+		}
+		glog.Infof("Detected change to %q", event.Name)
+		for _, reactor := range reactorFns {
+			go func() {
+				if err := reactor(event.Name); err != nil {
+					glog.Infof("Error reacting to %q: %v", err)
+				}
+			}()
```
sttts: use a wait group. Otherwise you can get nested calls. And maybe even queue events for each reactor.

glide.yaml
```diff
@@ -15,6 +15,10 @@ import:
 - package: github.com/openshift/api
   version: master
 
+
+# sig-master - needed for file observer
+- package: github.com/sigma/go-inotify
+  version: master
```
deads2k: you can't pin to master.  choose a sha or a tag

pkg/controller/controllercmd/cmd.go
```diff
@@ -96,6 +97,11 @@ func (c *ControllerCommandConfig) StartController(stopCh <-chan struct{}) error
 		return fmt.Errorf("unexpected config: %T", uncastConfig)
 	}
 
+	observedFiles := []string{
+		// TODO: Do we want to exit when the config file changed on disk?
```
deads2k: yes, for now because the CVO won't do a rollout for us.

pkg/controller/controllercmd/flags.go
```diff
@@ -47,6 +49,7 @@ func (f *ControllerFlags) AddFlags(cmd *cobra.Command) {
 	cmd.MarkFlagRequired("config")
 	flags.StringVar(&f.KubeConfigFile, "kubeconfig", f.KubeConfigFile, "Location of the master configuration file to run from.")
 	cmd.MarkFlagFilename("kubeconfig", "kubeconfig")
+	flags.BoolVar(&f.ExitOnConfigChange, "exit-on-config-change", f.ExitOnConfigChange, "Exit when change in config or certificates is detected.")
```
deads2k: Not flaggable.  Let someone explain their use-case

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+
+func newLinuxObserver() (Observer, error) {
+	inotifyWatcher, err := inotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	return &linuxObserver{
+		watcher:  inotifyWatcher,
+		reactors: map[string][]reactorFn{},
+		queue:    workqueue.New(),
```
deads2k: always name these

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+	}, nil
+}
+
+func (o *linuxObserver) AddReactor(reactor reactorFn, filenames ...string) Observer {
+	for _, f := range filenames {
+		glog.V(5).Infof("Adding reactor for file %q", f)
```
deads2k: 3

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+		glog.Infof("Observed change %s", event.String())
+		o.queue.Add(&reactionImpl{action: maskToActionType(event.Mask), filename: event.Name})
+	}
+}
+
+func (o *linuxObserver) Run(stopCh <-chan struct{}) {
+	glog.Info("Starting file observer")
+
+	// Process all errors from the watcher in parallel
+	go o.reportWatcherErrors()
```
deads2k: go funcs need stop channels.  Also, you need to catch panics `utilruntime.somethin`

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Process all reaction from queue
+	go o.processQueueReactions()
+
+	// Process all watch events
+	go o.processWatchEvents()
+
+	// Wait for the stop signal to shut down watcher and queue
+	select {
```
deads2k: nit: no need for select

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+	case <-stopCh:
+		glog.Infof("Shutting down file observer")
+		o.watcher.Close()
```
deads2k: you can defer these

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+		watcher:  inotifyWatcher,
+		reactors: map[string][]reactorFn{},
+		queue:    workqueue.New(),
+	}, nil
+}
+
+func (o *linuxObserver) AddReactor(reactor reactorFn, filenames ...string) Observer {
+	for _, f := range filenames {
+		glog.V(5).Infof("Adding reactor for file %q", f)
+		o.watcher.AddWatch(filepath.Dir(f), inotify.IN_MODIFY|inotify.IN_CREATE|inotify.IN_DELETE)
```
deads2k: so this is watching the directory.  modifies come to the directory or to the file?

mfojtik: file, the event always carry the file name (have test to prove that)

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+func (o *linuxObserver) processWatchEvents() {
+	for event := range o.watcher.Event {
+		if _, hasReactor := o.reactors[event.Name]; !hasReactor {
+			continue
+		}
+		glog.Infof("Observed change %s", event.String())
+		o.queue.Add(&reactionImpl{action: maskToActionType(event.Mask), filename: event.Name})
```
deads2k: Why are we adding the queue here instead of processing right here?  It would block, but do we really expect much traffic?

mfojtik: depends on where everywhere we end up using this... wanted to build this to scale if we get some more heavier traffic use cases or more reactors... i don't think the queue adding too much overhead.

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,114 @@
...
+	for {
+		select {
+		case event := <-o.watcher.Event:
+			o.reactorsMutex.Lock()
+			if _, hasReactor := o.reactors[event.Name]; !hasReactor {
+				continue
+			}
+			glog.Infof("Observed change: %s", event.String())
+			for i := range o.reactors[event.Name] {
+				if err := o.reactors[event.Name][i](event.Name, maskToAction(event.Mask)); err != nil {
```
sttts: good enough

pkg/controller/fileobserver/observer_linux.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+			o.reactorsMutex.Unlock()
+		case <-stopCh:
+			o.watcher.Close()
+			return
+		}
+	}
+}
+
+func (o *linuxObserver) Run(stopCh <-chan struct{}) {
```
sttts: godoc

pkg/controller/fileobserver/observer_linux.go
mfojtik: yeah, if we get to a point where we will watch 100+ files or need some rate limiter for executing the actions we can re-add the queue, but I'm fine with sequential processing for now..

it also mean the reactors should not do anything blocking.

sttts:
Same here, lgtm.

PR #88, opened 2018-11-16
controllercmd: wire the service serving certs properly

mfojtik:
This is on the path to serve HTTPS using the secure certificates provided by service serving cert signer. 

This will remove observation for change for temporary self-signed, insecure certs.
Will assume we mounting serving certs to `/var/run/secrets/serving-certs` in all operators (PR to actually mount this secret will follow)
Will observe appearance of `tls.key` and `tls.crt` in this directory and when they exists, it will kill the operator, causing the pod to restart the container and pick up the trusted certificates.


pkg/controller/controllercmd/cmd.go
```diff
@@ -97,44 +107,55 @@ func (c *ControllerCommandConfig) StartController(stopCh <-chan struct{}) error
...
 	// TODO maybe this should be optional?  It's a little difficult to come up with a scenario where this is worse than nothing though.
 	if len(config.ServingInfo.CertFile) == 0 && len(config.ServingInfo.KeyFile) == 0 {
 		servingInfoCopy := config.ServingInfo.DeepCopy()
 		configdefaults.SetRecommendedHTTPServingInfoDefaults(servingInfoCopy)
-		_, keyErr := os.Stat(servingInfoCopy.KeyFile)
-		_, certErr := os.Stat(servingInfoCopy.CertFile)
-		if os.IsNotExist(keyErr) && os.IsNotExist(certErr) {
-			certDir, err := ioutil.TempDir("", "serving-cert-")
+
+		if hasServiceServingCerts(certDir) {
```
sttts: do we have a race here if the files are mounted between here and the observer start?

deads2k: > do we have a race here if the files are mounted between here and the observer start?

I would take  TODO if it becomes a practical problem.

deads2k:
looks ok.  I could live with the until it becomes a practical problem.

pkg/controller/controllercmd/cmd.go
sttts: yes, TODO is fine.

PR #93, opened 2018-11-16
fileobserver: use polling to observe file changes

mfojtik:
The inotify implementation does not work in containers for projected files (shame Kube!). This PR switched the default implementation to use polling and hash calculation which should be enough to detect changes in files we want to monitor (certs, configs)...

This might not be used to large files where the hash calculation might take more time...


pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,124 @@
...
+	reactorsMutex sync.RWMutex
+}
+
+func (o *pollingObserver) AddReactor(reaction reactorFn, files ...string) Observer {
+	o.reactorsMutex.Lock()
+	defer o.reactorsMutex.Unlock()
+	for _, f := range files {
+		glog.V(3).Infof("Adding reactor for file %q", f)
+		var err error
+		o.files[f], err = calculateFileHash(f)
```
sttts: nit: only if f is new

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
+	var done bool
+	go func() {
+		<-stopCh
+		done = true
+	}()
+	err := wait.PollImmediateInfinite(1*time.Second, func() (bool, error) {
+		if done {
```
sttts:
```
switch {
case <-stopCh
  return true, nil
}
```

No need for the go routine.

mfojtik:
/hold

(need unit test)

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,124 @@
...
+			o.reactorsMutex.Unlock()
+
+			for i := range reactors {
+				action := FileModified
+				switch {
+				case len(lastKnownHash) == 0:
+					action = FileCreated
+				case len(lastKnownHash) > 0:
+					action = FileModified
+				case len(currentHash) == 0:
```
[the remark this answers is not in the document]
mfojtik: @sttts there is not falltrough right?

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,124 @@
...
+					action = FileDeleted
+				default:
+					glog.Errorf("%s: current hash: %q, lastObserved: %q", filename, currentHash, lastKnownHash)
```
sttts: len is either zero or >0. The default is not needed.

pkg/controller/fileobserver/observer_polling.go
mfojtik: @sttts won't this block?

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+}
+
+func (o *pollingObserver) Run(stopChan <-chan struct{}) {
+	glog.Info("Starting file observer")
+	defer glog.Infof("Shutting down file observer")
+
+	go o.processReactors(stopChan)
+
+	<-stopChan
```
sttts: Call processReactors directly. Then this line can go.

pkg/controller/fileobserver/observer_polling.go
mfojtik: @sttts why? even if you re-add you probably want to calc the hash straight away, no?

sttts: then existing reactors lose events.

pkg/controller/fileobserver/observer_polling.go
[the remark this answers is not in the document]
sttts: there isn't. https://goplay.space/#cnclx9UobBm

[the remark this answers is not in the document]
mfojtik: and you are obviously right ;-)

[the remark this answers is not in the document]
sttts: move the computation of the action out of the loop.

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
+	err := wait.PollImmediateInfinite(o.interval, func() (bool, error) {
+		select {
+		case <-stopCh:
+			return true, nil
+		default:
+		}
+		for filename, reactors := range o.reactors {
```
sttts: you need a lock here

mfojtik: that will mean you won't be able to add new reactors after you run, right?

mfojtik: iow. it will block until all reactors are processed.

pkg/controller/fileobserver/observer_polling_test.go
```diff
@@ -0,0 +1,132 @@
...
+func (r *reactionRecorder) get(f string) []ActionType {
+	r.RLock()
+	defer r.RUnlock()
+	return r.reactions[f]
+}
+
+func (r *reactionRecorder) add(f string, action ActionType) {
+	r.Lock()
+	r.reactions[f] = append(r.reactions[f], action)
+	defer r.Unlock()
```
sttts: nit: move one line up

pkg/controller/fileobserver/observer_polling_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	o, err := NewObserver(1 * time.Second)
+	if err != nil {
+		t.Fatalf("observer: %v", err)
+	}
+
+	reactions := newReactionRecorder()
+
+	var reactionsMutex sync.Mutex
```
sttts: what does reactionsMutex protect?

mfojtik: leftover, will delete

pkg/controller/fileobserver/observer_polling_test.go
```diff
@@ -0,0 +1,125 @@
...
+	r.Lock()
+	defer r.Unlock()
+	r.reactions[f] = append(r.reactions[f], action)
+}
+
+func TestObserverSimple(t *testing.T) {
+	dir, err := ioutil.TempDir("", "observer-simple-")
+	if err != nil {
+		t.Fatalf("tempdir: %v", err)
+	}
```
sttts: defer deleteEverything()?

pkg/controller/fileobserver/observer_polling_test.go
```diff
@@ -0,0 +1,125 @@
...
+				return true, fmt.Errorf("expected FileDeleted, got: %#v", reactions.get(testFile))
+			}
+			t.Logf("recv: %#v", reactions.get(testFile))
+			return true, nil
+		}, stopCh); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	}()
+
+	os.Remove(testFile)
```
sttts: this should be a defer for the temp dir at the top.

sttts: oh, ic. it's part of the test. Nvmd.

PR #94, opened 2018-11-16
 make v1 standard and update static pod operator to revision

deads2k:
/assign @sttts 

I'm testing the bump in the kube-apiserver now.


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -150,7 +150,7 @@ func (o *InstallOptions) copyContent() error {
 		return err
 	}
 	podContent := podConfigMap.Data["pod.yaml"]
-	podContent = strings.Replace(podContent, "DEPLOYMENT_ID", o.DeploymentID, -1)
+	podContent = strings.Replace(podContent, "DEPLOYMENT_ID", o.Revision, -1)
```
sttts: replace DEPLOYMENT_ID as well?

pkg/operator/v1alpha1status/controller.go
```diff
@@ -92,22 +92,31 @@ func (c StatusSyncer) sync() error {
 	unstructured.SetNestedField(operatorConfig.Object, c.clusterOperatorNamespace, "metadata", "namespace")
 	unstructured.SetNestedField(operatorConfig.Object, c.clusterOperatorName, "metadata", "name")
 
+	errorMessages := []string{}
```
sttts: this is a non-mechanic change?

deads2k: > this is a non-mechanic change?

I think the diff is confused.

pkg/operator/staticpod/installerpod/cmd.go
deads2k: fixed

PR #97, opened 2018-11-19
config: add status helper funcs for v1.ClusterOperator

abhinavdahiya:
These functions currently exist in https://github.com/openshift/cluster-version-operator/blob/0eaaaa01e881933223f5afcb57de73dbaf793b54/lib/resourcemerge/os.go#L36-L109.
This removes dependency of operators from cvo completely.

@deads2k hopefully this import `github.com/library-go/pkg/config/clusteroperator/v1` is sane for getting these functions on `ClusterOperatorStatus`.

/cc @smarterclayton @deads2k


abhinavdahiya:
ping :)

pkg/config/clusteroperator/v1/status.go
```diff
@@ -0,0 +1,90 @@
+package v1
+
+import (
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
deads2k: nit, project standard is 


```
standard imports

non-kube, non-openshift

kube

openshift
```

deads2k:
You sure want this or want to simply update the `ClusterOperator` resource to use the `operator/v1` conditions?

abhinavdahiya:
> You sure want this or want to simply update the `ClusterOperator` resource to use the `operator/v1` conditions?

Either is fine. I don't have any strong prefs on that, whatever you think is more correct.
/cc @smarterclayton

deads2k:
> Either is fine. I don't have any strong prefs on that, whatever you think is more correct.
> /cc @smarterclayton

I think we should collapse our status conditions into a single type.  It's a quick change in openshift/api and doesn't affect serialization in any serious way.

abhinavdahiya:
>  Feel free to self-tag after that.

Adding the `lgtm` label manually as @deads2k allowed me to do it. :innocent:

PR #98, opened 2018-11-20
update generation handling for operator v1

deads2k:
Generation handling changed in operatorv1 and this updates us to use it.

/assign @sanchezl


mfojtik:
LGTM

PR #101, opened 2018-11-20
use protobuf when possible

deads2k:
/hold

holding until we prove it works.

### DIFF
--- a/pkg/config/client/client_config.go
+++ b/pkg/config/client/client_config.go
@@ -71,6 +71,15 @@ func applyClientConnectionOverrides(overrides *ClientConnectionOverrides, kubeCo
 	if len(overrides.ContentType) > 0 {
 		kubeConfig.ContentConfig.ContentType = overrides.ContentType
 	}
+
+	// if we have no preferences at this point, claim that we accept both proto and json.  We will get proto if the server supports it.
+	// this is a slightly niggly thing.  If the server has proto and our client does not (possible, but not super likely) then this fails.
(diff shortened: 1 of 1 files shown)


deads2k:
@smarterclayton spelled like this?

enj:
Does this work with the dynamic client?

pkg/config/client/client_config.go
```diff
@@ -71,6 +71,15 @@ func applyClientConnectionOverrides(overrides *ClientConnectionOverrides, kubeCo
 	if len(overrides.ContentType) > 0 {
 		kubeConfig.ContentConfig.ContentType = overrides.ContentType
 	}
+
+	// if we have no preferences at this point, claim that we accept both proto and json.  We will get proto if the server supports it.
+	// this is a slightly niggly thing.  If the server has proto and our client does not (possible, but not super likely) then this fails.
+	if len(kubeConfig.ContentConfig.AcceptContentTypes) == 0 {
+		kubeConfig.ContentConfig.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
+	}
+	if len(kubeConfig.ContentConfig.AcceptContentTypes) == 0 {
```
smarterclayton: This is duplication?

smarterclayton:
We need the dynamic client to bypass proto somehow.  If you are using unstructured you require json.  I can look at sollys watch bug (that should be easy to review), but we may want to make dynamic client clearly override / default to json even if we want to leave open that we may have a more efficient unstructured payload in the future.

PR #104, opened 2018-11-22
Add simple event recorder

mfojtik:
Adding event recorder interface for recording events from operators. Recorder should be wired to resourceapply package and observer package to. Events published by operator are for historical action debugging.

This implementation require the ObjectReference to be passed to NewRecorder in the factory when the operator is started. It means that the `EVENT_SOURCE_REPLICASET_NAME` will have to be passed via downward API into the container running the operator and the replica set is then used as the object we recording the events against. These events should be also propagated to the owning deployment.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/events/recorder.go
+++ b/pkg/operator/events/recorder.go
@@ -1,0 +1,78 @@
+package events
+
+import (
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+// Recorder is a simple event recording interface.
+type Recorder interface {
+	Event(reason, message string) error
+	Eventf(reason, messageFmt string, args ...interface{}) error
+	Warning(reason, message string) error
+	Warningf(reason, messageFmt string, args ...interface{}) error
+}
+
+// NewRecorder returns new event recorder.
+func NewRecorder(client corev1client.EventInterface, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
+	return &recorder{
+		eventClient:       client,
+		involvedObjectRef: involvedObjectRef,
+		sourceComponent:   sourceComponentName,
+	}
+}
+
+// recorder is an implementation of Recorder interface.
+type recorder struct {
+	eventClient       corev1client.EventInterface
+	involvedObjectRef *corev1.ObjectReference
+	sourceComponent   string
+}
+
+// Event emits the normal type event and allow formatting of message.
+func (r *recorder) Eventf(reason, messageFmt string, args ...interface{}) error {
+	return r.Event(reason, fmt.Sprintf(messageFmt, args...))
+}
+
+// Warning emits the warning type event and allow formatting of message.
+func (r *recorder) Warningf(reason, messageFmt string, args ...interface{}) error {
+	return r.Warning(reason, fmt.Sprintf(messageFmt, args...))
+}
+
+// Event emits the normal type event.
+func (r *recorder) Event(reason, message string) error {
+	_, err := r.eventClient.Create(r.makeEvent(r.involvedObjectRef, corev1.EventTypeNormal, reason, message))
+	return err
+}
+
+// Warning emits the warning type event.
+func (r *recorder) Warning(reason, message string) error {
+	_, err := r.eventClient.Create(r.makeEvent(r.involvedObjectRef, corev1.EventTypeWarning, reason, message))
+	return err
+}
+
+func (r recorder) makeEvent(involvedObjRef *corev1.ObjectReference, eventType, reason, message string) *corev1.Event {
+	currentTime := metav1.Time{Time: time.Now()}
+	event := &corev1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%v.%x", involvedObjRef.Name, currentTime.UnixNano()),
+			Namespace: involvedObjRef.Namespace,
+		},
+		InvolvedObject: *involvedObjRef,
+		Reason:         reason,
+		Message:        message,
+		Type:           eventType,
+		Count:          1,
+		FirstTimestamp: currentTime,
+		LastTimestamp:  currentTime,
+		EventTime:      metav1.MicroTime{Time: currentTime.Time},
+	}
+	if len(r.sourceComponent) > 0 {
+		event.Source.Component = r.sourceComponent
+	}
+	return event
+}
--- a/pkg/operator/events/recorder_deployment.go
+++ b/pkg/operator/events/recorder_deployment.go
@@ -1,0 +1,34 @@
+package events
+
+import (
+	"fmt"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+	"k8s.io/client-go/tools/reference"
+)
+
+// eventSourceReplicaSetNameEnv is a name of environment variable inside container that specifies the name of the current replica set.
+// This replica set name is then used as a source/involved object for operator events.
+const eventSourceReplicaSetNameEnv = "EVENT_SOURCE_REPLICASET_NAME"
+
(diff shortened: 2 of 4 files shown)


pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (r recorder) makeEvent(source RecorderSource, eventType, reason, message string) *corev1.Event {
+	currentTime := metav1.Time{Time: time.Now()}
+	gvk := source.GroupVersionKind()
+	apiVersion, kind := gvk.ToAPIVersionAndKind()
+
+	event := &corev1.Event{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%v.%x", source.GetObjectMeta().GetName(), currentTime.UnixNano()),
```
sttts: never looked into how events are named. this is how? 😱

mfojtik: yes, copied from deployment controller :)

pkg/operator/events/recorder_deployment.go
```diff
@@ -0,0 +1,32 @@
...
+var eventSourceReplicaSetNameEnvFunc = func() string {
+	return os.Getenv(eventSourceReplicaSetNameEnv)
+}
+
+// DeploymentFromEnvRecorderSource return event recorder source using the EVENT_SOURCE_REPLICASET_NAME environment variable
+// set inside the container created by deployment.
+func DeploymentFromEnvRecorderSource(client appsv1client.ReplicaSetInterface) (RecorderSourceFunc, error) {
+	replicaSetName := eventSourceReplicaSetNameEnvFunc()
+	if len(replicaSetName) == 0 {
+		return nil, fmt.Errorf("Unable to setup event recorder as '%s' env variable is not set.", eventSourceReplicaSetNameEnv)
```
sttts: lower case error

sttts: %q instead of quoted %s

pkg/operator/events/recorder_deployment.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	appsv1client "k8s.io/client-go/kubernetes/typed/apps/v1"
+)
+
+const eventSourceReplicaSetNameEnv = "EVENT_SOURCE_REPLICASET_NAME"
```
sttts: document what this does

pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,92 @@
...
+	CreateEvent(reason, message string) error
+	CreateEventWarning(reason, message string) error
+}
+
+type RecorderSource interface {
+	GroupVersionKind() schema.GroupVersionKind
+	metav1.ObjectMetaAccessor
+}
+
+func NewRecorder(client corev1client.EventInterface, sourceComponent string, sourceFunc RecorderSourceFunc) Recorder {
```
sttts: why the RecorderSourceFunc and not RecorderSource directly? The func without arguments is like a constant for the record, isn't it?

pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type RecorderSourceFunc func() (RecorderSource, error)
```
soltysh: Why not use `k8s.io/client-go/tools/record` package instead of writing your own, we already vendor it

mfojtik: Because there is no way to only use the EventRecorder without adding EventBroadcaster. Because the EventRecorder does not set all the events fields we want to set and I really want to be in control what fields are set and how. Because we will have to create an instance of this recorder everything we creating an event to fill the EventSource (as you can't dynamically change the involved object, etc..)

In summary, I'm not a fan of re-using things if they don't fit this use case. Especially if this recorder does not need to be super sophisticated and we don't really need/want the sophistication of EventSink.

pkg/operator/events/recorder.go
mfojtik: yeah, good point

pkg/operator/events/recorder.go
soltysh: Discussed this in length on Slack, I'm ok with it for now :wink:

pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,92 @@
...
+
+func NewRecorder(client corev1client.EventInterface, sourceComponent string, sourceFunc RecorderSourceFunc) Recorder {
+	return &recorder{
+		eventClient:     client,
+		sourceFunc:      sourceFunc,
+		sourceComponent: sourceComponent,
+	}
+}
+
+type recorderSourceObject struct {
```
soltysh: Why this? Why not just using `ObjectReference`, all you care is there already? This also applies to the interface `RecordSource`, why not just use `runtime.Object`?

pkg/operator/events/recorder.go
mfojtik: @sttts actually, `RecorderSourceFunc` returns the RecorderSource and error and it allows things like `DeploymentFromEnvRecorderSource` to be used as argument. I don't see how to refactor this to loose the RecorderSourceFunc ?

pkg/operator/events/recorder.go
soltysh: How about instead having your own implementation of `EventRecorder` interface, this will save us some work in the future, if we need to switch?

pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,87 @@
...
+	eventClient     corev1client.EventInterface
+	sourceFunc      RecorderSourceFunc
+	sourceComponent string
+}
+
+// Event emits the normal type event.
+func (r *recorder) Event(reason, message string) error {
+	source, err := r.sourceFunc()
+	if err != nil {
+		return fmt.Errorf("enable to get event source: %v", err)
```
soltysh: s/enable/unable

soltysh: or failed

pkg/operator/events/recorder.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// This will likely never happen.
+	return fmt.Errorf("unable to create event as the source object is not specified")
+}
+
+// EventWarning emits the warning type event.
+func (r *recorder) EventWarning(reason, message string) error {
+	source, err := r.sourceFunc()
+	if err != nil {
+		return fmt.Errorf("enable to get event source: %v", err)
```
soltysh: ditto

soltysh: Why not move this block into `makeEvent`?

pkg/operator/events/recorder_deployment.go
```diff
@@ -0,0 +1,44 @@
...
+	if len(replicaSetName) == 0 {
+		return nil, fmt.Errorf("unable to setup event recorder as %q env variable is not set", eventSourceReplicaSetNameEnv)
+	}
+
+	return func() (*corev1.ObjectReference, error) {
+		rs, err := client.Get(replicaSetName, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		return &corev1.ObjectReference{
```
soltysh: Use `k8s.io/client-go/tools/reference#GetReference` instead

pkg/operator/events/recorder_deployment.go
```diff
@@ -0,0 +1,35 @@
...
+const eventSourceReplicaSetNameEnv = "EVENT_SOURCE_REPLICASET_NAME"
+
+// eventSourceReplicaSetNameEnvFunc allows to override the way we get the environment variable value (for unit tests).
+var eventSourceReplicaSetNameEnvFunc = func() string {
+	return os.Getenv(eventSourceReplicaSetNameEnv)
+}
+
+// DeploymentFromEnvRecorderSource return event recorder source using the EVENT_SOURCE_REPLICASET_NAME environment variable
+// set inside the container created by deployment.
+// TODO: Switching to a lister will make this faster.
```
sttts: it's called once per process, isn't  it?

mfojtik: yeah, this todo is not relevant

PR #105, opened 2018-11-22
Wire event recorder

mfojtik:
Wire events into generic controllers and to controller command builder (so operators can use it).

Note that with this we have to bump and fix every operator using this framework as the startFunc will have different signature.

Do not merge this until we agree where and what events we want to have.

/cc @deads2k 
/cc @sttts


pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -365,7 +371,7 @@ func TestCreateInstallerPodMultiNode(t *testing.T) {
 			expectedUpgradeOrder: []int{0, 1, 2},
 		},
 		{
-			name: "three nodes with current revision, second static pods unread",
+			name:                    "three nodes with current revision, second static pods unread",
```
mfojtik: wtf gofmt

pkg/controller/controllercmd/builder.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"k8s.io/client-go/kubernetes"
```
soltysh: Move this to k8s imports

pkg/controller/controllercmd/builder.go
```diff
@@ -70,6 +73,21 @@ func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, actio
 	return b
 }
 
+func (b *ControllerBuilder) WithEventRecorder() *ControllerBuilder {
+	clientConfig, err := b.getClientConfig()
+	if err != nil {
+		panic(fmt.Sprintf("error getting client config: %v", err))
+	}
+	kubeClient := kubernetes.NewForConfigOrDie(clientConfig)
+	replicaSetRef, err := events.GetReplicaSetOwnerReference(kubeClient.AppsV1().ReplicaSets(b.componentName))
```
soltysh: I think we should rename that `componentName` to `targetNamespaceName` or something alike, currently it's confusing.

pkg/operator/staticpod/controller/backingresource/backing_resource_controller.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/operator/events"
```
soltysh: Move down to origin imports.

pkg/operator/staticpod/controller/installer/installer_controller_test.go
soltysh: lol

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/operator/events"
```
soltysh: Ditto

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
```
soltysh: Ditto

pkg/operator/staticpod/controller/revision/deployment_controller_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
```
soltysh: I think goland hates you :stuck_out_tongue:

pkg/controller/controllercmd/builder.go
```diff
@@ -19,10 +20,11 @@ import (
 	leaderelectionconverter "github.com/openshift/library-go/pkg/config/leaderelection"
 	"github.com/openshift/library-go/pkg/config/serving"
 	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
 )
 
 // StartFunc is the function to call on leader election start
-type StartFunc func(config *rest.Config, stop <-chan struct{}) error
+type StartFunc func(config *rest.Config, eventRecorder events.Recorder, stop <-chan struct{}) error
```
enj: This signature may start getting out of hand with https://github.com/openshift/library-go/pull/103 and future requirements.  Maybe refactor into a single config struct.

pkg/controller/controllercmd/builder.go
enj: These probably need to be separate things?

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -78,5 +80,22 @@ func ApplyDirectly(client kubernetes.Interface, manifests AssetFunc, files ...st
 		ret = append(ret, result)
 	}
 
+	if recorder != nil {
```
enj: Maybe provide a noop recorder to avoid nil checks?

pkg/controller/controllercmd/builder.go
mfojtik: the componentName is currently set to namespace name as we are trying to match those (remember @deads2k talking about us moving namespaces? ;-)

I can add targetNamespace as argument to this func and we can set it to componentName and others to whatever they need :-)

pkg/operator/staticpod/controller/revision/deployment_controller_test.go
mfojtik: has nothing to do with goland... it is goimports weakness ;-)

pkg/controller/controllercmd/builder.go
mfojtik: /cc @deads2k 

i'm open to refactor as follow up

pkg/operator/v1alpha1staticpod/controller/installer/installer_controller_test.go
```diff
@@ -327,7 +327,7 @@ func TestCreateInstallerPodMultiNode(t *testing.T) {
 		updateStatusErrors                  []error
 	}{
 		{
-			name: "three fresh nodes",
+			name:                                "three fresh nodes",
```
mfojtik:
```
~/g/s/g/o/library-go ❯❯❯ GOROOT=~/go/1.10 make update-gofmt                                                                                                                                                                                                                                                                                             wire-event-recorder
Running `gofmt -s -l -w` on 108 file(s).
~/g/s/g/o/library-go ❯❯❯
```

pkg/operator/resource/resourceapply/generic.go
mfojtik: added NewFakeRecorder() that records events to a slice (handy for unit tests)...

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/imdario/mergo"
+	"github.com/openshift/library-go/pkg/operator/events"
```
soltysh: Yet another cherry :wink:

pkg/operator/events/recorder.go
```diff
@@ -34,28 +35,34 @@ type recorder struct {
...
 
 // Event emits the normal type event.
-func (r *recorder) Event(reason, message string) error {
-	_, err := r.eventClient.Create(r.makeEvent(r.involvedObjectRef, corev1.EventTypeNormal, reason, message))
-	return err
+func (r *recorder) Event(reason, message string) {
+	event := makeEvent(r.involvedObjectRef, r.sourceComponent, corev1.EventTypeNormal, reason, message)
+	_, err := r.eventClient.Create(event)
+	if err != nil {
+		glog.Warningf("Error creating event %+v: %v", event, err)
```
soltysh: Why you decided to go with this?

pkg/operator/events/recorder_fake.go
```diff
@@ -0,0 +1,50 @@
...
+	Name:       "test",
+	APIVersion: "v1",
+}
+
+type FakeRecorder interface {
+	Events() []*corev1.Event
+	Recorder
+}
+
+// NewFakeRecorder provides an event recorder that
```
soltysh: that... what?

pkg/operator/resource/resourceapply/generic.go
soltysh: I don't think that's what Mo meant, NoopRecorder is different from FakeRecorder. The former can be used by folks who don't want to have events recorded, but ensures that one is always set and you don't have to perform this nil check, something like `ioutil.Discard`. The latter is **only** for tests.

mfojtik: @soltysh @enj i don't think we should allow events to not be recorded at all. If a function is publishing events, they should be recorded and not ignored. This function is used in operators to apply changes to resources the operator manages. I think the point of events here is to always have timeline based on events for debugging of why the resource XYZ changed.

enj: That would still mean removing the nil check right?  As long as it is easy to get an event recorder, I can live without the noop (one can always create one if needed...).

mfojtik: it is super easy ;-)

soltysh: With Michal we went as far as always requiring that recorder so that nil won't be necessary at all.

pkg/controller/controllercmd/builder.go
sttts: doesn't this force us to be managed by ReplicaSet i.e. Deployments? What about DaemonSets? This does not look generic.

sttts: Would expect a way to inject the way the target object is derived, with a default implementation for replicasets.

pkg/controller/controllercmd/cmd.go
```diff
@@ -163,6 +163,7 @@ func (c *ControllerCommandConfig) StartController(stopCh <-chan struct{}) error
 		WithLeaderElection(config.LeaderElection, "", c.componentName+"-lock").
 		WithServer(config.ServingInfo, config.Authentication, config.Authorization).
 		WithFileObserver(fileobserver.ExitOnChangeReactor, observedFiles...).
+		WithEventRecorder().
```
sttts: Calling this is optional? No nil pointer panic? This should be documented in Godoc of WithEventRecorder.

pkg/operator/events/recorder.go
```diff
@@ -34,28 +35,34 @@ type recorder struct {
...
 }
 
 // Event emits the normal type event.
-func (r *recorder) Event(reason, message string) error {
-	_, err := r.eventClient.Create(r.makeEvent(r.involvedObjectRef, corev1.EventTypeNormal, reason, message))
-	return err
+func (r *recorder) Event(reason, message string) {
+	event := makeEvent(r.involvedObjectRef, r.sourceComponent, corev1.EventTypeNormal, reason, message)
+	_, err := r.eventClient.Create(event)
+	if err != nil {
```
sttts: nit: `if err := ...; err != nil {`

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -87,7 +91,9 @@ func (c ConfigObserver) sync() error {
 	spec := originalSpec.DeepCopy()
 	// don't worry about errors.  If we can't decode, we'll simply stomp over the field.
 	existingConfig := map[string]interface{}{}
-	json.NewDecoder(bytes.NewBuffer(spec.ObservedConfig.Raw)).Decode(&existingConfig)
+	if err := json.NewDecoder(bytes.NewBuffer(spec.ObservedConfig.Raw)).Decode(&existingConfig); err != nil {
+		glog.V(4).Info("decode failed with error: %v", err)
```
sttts: decode of what?

pkg/operator/events/recorder.go
```diff
@@ -34,28 +35,34 @@ type recorder struct {
...
+func (r *recorder) Warningf(reason, messageFmt string, args ...interface{}) {
+	r.Warning(reason, fmt.Sprintf(messageFmt, args...))
 }
 
 // Event emits the normal type event.
-func (r *recorder) Event(reason, message string) error {
-	_, err := r.eventClient.Create(r.makeEvent(r.involvedObjectRef, corev1.EventTypeNormal, reason, message))
-	return err
+func (r *recorder) Event(reason, message string) {
+	event := makeEvent(r.involvedObjectRef, r.sourceComponent, corev1.EventTypeNormal, reason, message)
```
sttts: return early if r is nil

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -78,5 +80,18 @@ func ApplyDirectly(client kubernetes.Interface, manifests AssetFunc, files ...st
...
```
sttts: return early if it is nil

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -78,5 +80,18 @@ func ApplyDirectly(client kubernetes.Interface, manifests AssetFunc, files ...st
...
+		for _, result := range ret {
+			if result.Error != nil {
+				recorder.Warningf("ResourceApplyFailed", "Failed to apply updated to %s: %v", result.Type, result.Error)
```
sttts: Failed to apply update to

pkg/operator/staticpod/controller/backingresource/backing_resource_controller_test.go
```diff
@@ -119,7 +120,7 @@ func TestBackingResourceController(t *testing.T) {
 			),
 			validateActions: func(t *testing.T, actions []clienttesting.Action) {
 				createdObjects := filterCreateActions(actions)
-				if createdObjectCount := len(createdObjects); createdObjectCount != 1 {
+				if createdObjectCount := len(createdObjects); createdObjectCount != 2 {
```
sttts: would prefer to make the tests of the logic and the events orthogonal. This way this gets more and more complicated.

pkg/controller/controllercmd/builder.go
mfojtik: @stts to support daemonsets we will have to bake that logic here, otherwise we will be bloating the NewCommand? Since we only will support deployments and daemonsets, I think we can do a simple if here based on the existence of environment variable? (if REPLICASET_POD or DAEMONSET_POD ?)

pkg/operator/events/recorder.go
mfojtik: how can be 'r' nil ?

pkg/operator/staticpod/controller/backingresource/backing_resource_controller_test.go
mfojtik: i have fake recorder now, so this is not needed, good catch

pkg/controller/controllercmd/builder.go
mfojtik: The current implementation is generic, this was just bad name. It handles every pod that has controller owner ref (it follow that owner ref to get the controller object (replicaset/daemonset)

pkg/controller/controllercmd/builder.go
```diff
@@ -58,6 +58,8 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
 	}
 }
 
+// WithRestartOnChange will enable a file observer controller loop that observes changes into specified files. If a change to file is detected,
```
sttts: to a file

pkg/controller/controllercmd/builder.go
```diff
@@ -58,6 +58,8 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
...
+// the operator container will restart to reload the resources.
```
sttts: better: the process will be terminated with return code zero. The restart is not done by this logic, but by the pod around the binary.

pkg/controller/controllercmd/builder.go
```diff
@@ -82,6 +84,8 @@ func (b *ControllerBuilder) WithRestartOnChange(stopChan chan struct{}, files ..
 	return b
 }
 
+// WithEventRecorder will enable event recording for this controller. To make this work, the 'EVENT_SOURCE_POD_NAME' environment variable has
+// to be present in system with a name of the pod that is currently running the container.
```
sttts: in the system

sttts: Why is the EVENT_SOURCE_POD_NAME var not just called POD_NAME?

pkg/controller/controllercmd/builder.go
```diff
@@ -82,6 +84,8 @@ func (b *ControllerBuilder) WithRestartOnChange(stopChan chan struct{}, files ..
...
 func (b *ControllerBuilder) WithEventRecorder(targetNamespace string) *ControllerBuilder {
```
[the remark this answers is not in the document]
mfojtik: will read the namespace from container (/var/run/secrets..)... we don't support running outside container anyway.

pkg/operator/events/recorder.go
sttts: you use the Recorder interface in the plumbing. So yes, it can't be nil here, but the interface can.

pkg/controller/controllercmd/builder.go
```diff
@@ -55,7 +59,9 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
 	}
```
sttts: We should set a noop-recorder here

pkg/controller/controllercmd/cmd.go
```diff
@@ -75,7 +73,8 @@ func (c *ControllerCommandConfig) NewCommand() *cobra.Command {
 				glog.Fatal(err)
 			}
 
-			if err := c.StartController(wait.NeverStop); err != nil {
+			stopChan := make(chan struct{})
```
sttts: use wait.NeverStops

pkg/controller/controllercmd/cmd.go
```diff
@@ -97,7 +96,7 @@ func hasServiceServingCerts(certDir string) bool {
 }
 
 // StartController runs the controller
-func (c *ControllerCommandConfig) StartController(stopCh <-chan struct{}) error {
+func (c *ControllerCommandConfig) StartController(stopCh chan struct{}) error {
```
sttts: I don't like the change to `chan struct{}`. It makes the ownership unclear and leads to panic if both, caller and controller close the channel. Better use your own channel `exitOnChangeReactorCh` in addition locally here and combine it with `stopCh`. Pass the result to `Run`.

sttts: Or even better: hide this extra channel from here by moving it into `staticPodOperatorControllers` and let `Run` stop if it is closed.

pkg/operator/events/recorder.go
```diff
@@ -69,10 +106,9 @@ func (r recorder) makeEvent(involvedObjRef *corev1.ObjectReference, eventType, r
 		Count:          1,
 		FirstTimestamp: currentTime,
 		LastTimestamp:  currentTime,
-		EventTime:      metav1.MicroTime{Time: currentTime.Time},
 	}
-	if len(r.sourceComponent) > 0 {
-		event.Source.Component = r.sourceComponent
+	if len(sourceComponent) > 0 {
+		event.Source.Component = sourceComponent
```
sttts: isn't `event.Source.Component` empty by default? No need for the if clause.

pkg/operator/v1alpha1staticpod/controller/backingresource/backing_resource_controller_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
```
sttts: order

pkg/controller/controllercmd/builder.go
```diff
@@ -51,11 +55,14 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
...
 		componentName:    componentName,
+		eventRecorder:    events.NewDummyRecorder(componentName),
 		observerInterval: defaultObserverInterval,
 	}
 }
 
-func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, action fileobserver.ActionType) error, files ...string) *ControllerBuilder {
+// WithRestartOnChange will enable a file observer controller loop that observes changes into specified files. If a change to a file is detected,
+// the process will be terminated with return code zero.
+func (b *ControllerBuilder) WithRestartOnChange(stopChan chan struct{}, files ...string) *ControllerBuilder {
```
sttts: nit: stopCh by convention

pkg/controller/controllercmd/builder.go
```diff
@@ -66,10 +73,40 @@ func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, actio
 		}
 		b.fileObserver = observer
 	}
+	reactorFunc := func(filename string, action fileobserver.ActionType) error {
+		message := fmt.Sprintf("Restart triggered because %s", action.String(filename))
```
sttts: nit: "because %s changed" or "because of %s"

pkg/controller/controllercmd/builder.go
```diff
@@ -66,10 +73,40 @@ func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, actio
...
+		if b.eventRecorder != nil {
```
sttts: Do we need this nil check?

pkg/controller/controllercmd/builder.go
mfojtik: fixed.

pkg/controller/controllercmd/builder.go
mfojtik: fixed.

pkg/controller/controllercmd/builder.go
mfojtik: no, fixed

pkg/controller/controllercmd/cmd.go
```diff
@@ -10,16 +10,15 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/util/wait"
```
sttts: nit: order

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
```
sttts: nit: order

pkg/operator/events/recorder_logging.go
```diff
@@ -1 +1,33 @@
...
+	event := makeEvent(&inMemoryDummyObjectReference, "", corev1.EventTypeNormal, reason, message)
+	glog.Info(event.String())
+}
+
+func (r *LoggingEventRecorder) Eventf(reason, messageFmt string, args ...interface{}) {
+	r.Event(reason, fmt.Sprintf(messageFmt, args...))
+}
+
+func (r *LoggingEventRecorder) Warning(reason, message string) {
+	event := makeEvent(&inMemoryDummyObjectReference, "", corev1.EventTypeNormal, reason, message)
```
sttts: event type

pkg/controller/controllercmd/builder.go
```diff
@@ -51,11 +55,14 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
 	return &ControllerBuilder{
 		startFunc:        startFunc,
 		componentName:    componentName,
+		eventRecorder:    events.NewInMemoryRecorder(componentName),
```
sttts: NewLoggingEventRecorder

pkg/operator/events/recorder.go
```diff
@@ -2,19 +2,52 @@ package events
...
+const podNameEnv = "POD_NAME"
+
+// podNameEnvFunc allows to override the way we get the environment variable value (for unit tests).
+var podNameEnvFunc = func() string {
+	return os.Getenv(podNameEnv)
+}
+
+// GetControllerReferenceForCurrentPod provides an object reference to a controller managing the pod/container where this process runs.
+// The pod name must be provided via the POD_NAME name.
+func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*corev1.ObjectReference, error) {
```
sttts: Make it private

mfojtik: used in builder, must stay external.

pkg/controller/controllercmd/builder.go
```diff
@@ -51,11 +55,14 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
...
 		componentName:    componentName,
+		eventRecorder:    events.NewLoggingEventRecorder(),
 		observerInterval: defaultObserverInterval,
 	}
 }
 
-func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, action fileobserver.ActionType) error, files ...string) *ControllerBuilder {
+// WithRestartOnChange will enable a file observer controller loop that observes changes into specified files. If a change to a file is detected,
+// the process will be terminated with return code zero.
+func (b *ControllerBuilder) WithRestartOnChange(stopCh chan struct{}, files ...string) *ControllerBuilder {
```
sttts: Godoc is outdated.

Make stopCh a `chan<- struct{}`.

pkg/controller/controllercmd/builder.go
```diff
@@ -66,10 +73,38 @@ func (b *ControllerBuilder) WithFileObserver(reactorFunc func(file string, actio
...
+	namespace, err := b.getNamespace()
+	if err != nil {
+		panic("unable to read the namespace")
+	}
+	controllerRef, err := events.GetControllerReferenceForCurrentPod(kubeClient.CoreV1().Pods(namespace))
+	if err != nil {
+		panic(fmt.Sprintf("unable to obtain replicaset reference for events: %v", err))
+	}
+
+	b.eventRecorder = events.NewRecorder(kubeClient.CoreV1().Events(namespace), b.componentName, controllerRef)
```
deads2k: Why save it like this?  Why not simply build it directly in `Run` and always pass it through?

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -100,7 +106,9 @@ func (c ConfigObserver) sync() error {
 
 	mergedObservedConfig := map[string]interface{}{}
 	for _, observedConfig := range observedConfigs {
-		mergo.Merge(&mergedObservedConfig, observedConfig)
+		if err := mergo.Merge(&mergedObservedConfig, observedConfig); err != nil {
+			glog.Warningf("merging observed config failed: %v", err)
```
deads2k: event and status worthy.  Can be followup for @sanchezl

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -87,7 +91,9 @@ func (c ConfigObserver) sync() error {
 	spec := originalSpec.DeepCopy()
 	// don't worry about errors.  If we can't decode, we'll simply stomp over the field.
 	existingConfig := map[string]interface{}{}
-	json.NewDecoder(bytes.NewBuffer(spec.ObservedConfig.Raw)).Decode(&existingConfig)
+	if err := json.NewDecoder(bytes.NewBuffer(spec.ObservedConfig.Raw)).Decode(&existingConfig); err != nil {
+		glog.V(4).Infof("decode of existing config failed with error: %v", err)
```
deads2k: We can make this noiser if you like.  Just don't fail on it.

pkg/controller/controllercmd/builder.go
mfojtik: fixed

pkg/operator/configobserver/config_observer_controller.go
mfojtik: @deads2k i don't think we need, new goland version highlight unhandled errors and it triggered my OCD ;-)

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/operator/events"
```
deads2k: order

pkg/operator/staticpod/controller/backingresource/backing_resource_controller_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
```
deads2k: order

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/operator/events"
```
deads2k: order

pkg/controller/controllercmd/builder.go
```diff
@@ -163,24 +197,23 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, stopCh <-chan
...
 	leaderelection.RunOrDie(leaderElection)
 	return fmt.Errorf("exited")
 }
 
-func (b *ControllerBuilder) getClientConfig() (*rest.Config, error) {
-	kubeconfig := ""
-	if b.kubeAPIServerConfigFile != nil {
-		kubeconfig = *b.kubeAPIServerConfigFile
+func (b *ControllerBuilder) getNamespace() (string, error) {
+	nsBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
deads2k: this will eventually need to respect the kubeconfig file, but I suspect that @enj will fix it up.

mfojtik: @deads2k off topic, but I was surprised there is no method in client-go that return this...

PR #110, opened 2018-11-26
Pass controller context struct to StartFunc

enj:
StartFunc was refactored to take a ControllerContext struct.  This prevents the function signature from expanding indefinitely and makes is easier to add new fields without breaking every caller.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -6,12 +6,11 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"k8s.io/client-go/kubernetes"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/wait"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/server/healthz"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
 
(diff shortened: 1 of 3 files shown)


pkg/controller/controllercmd/builder.go
```diff
@@ -26,7 +25,14 @@ import (
 )
 
 // StartFunc is the function to call on leader election start
-type StartFunc func(config *unstructured.Unstructured, kubeconfig *rest.Config, eventRecorder events.Recorder, stop <-chan struct{}) error
+type StartFunc func(*ControllerConfig) error
+
+type ControllerConfig struct {
```
deads2k: ControllerContext

pkg/controller/controllercmd/builder.go
```diff
@@ -26,7 +25,14 @@ import (
...
+	ComponentConfig runtime.Object
```
deads2k: I think this should be unstructured.

pkg/controller/controllercmd/builder.go
```diff
@@ -117,6 +124,12 @@ func (b *ControllerBuilder) WithKubeConfigFile(kubeConfigFilename string, defaul
 	return b
 }
 
+// WithComponentConfig sets an optional config object which is passed to startFunc
+func (b *ControllerBuilder) WithComponentConfig(componentConfig runtime.Object) *ControllerBuilder {
```
deads2k: This looks inverted to me.

pkg/controller/controllercmd/cmd.go
```diff
@@ -57,6 +53,14 @@ func NewControllerCommandConfig(componentName string, version version.Info, star
 	}
 }
 
+// WithComponentConfig sets an optional config object which is used as a decoding target for the config command line flag.
+// After successful decoding, this config object is passed to startFunc.
+func (c *ControllerCommandConfig) WithComponentConfig(componentConfig runtime.Object, gv schema.GroupVersion) *ControllerCommandConfig {
```
deads2k: this doesn't look any easier than decoding when you receive the unstructured.

pkg/controller/controllercmd/flags.go
```diff
@@ -32,10 +31,7 @@ func NewControllerFlags() *ControllerFlags {
 
 // Validate makes sure the required flags are specified and no illegal combinations are found
 func (o *ControllerFlags) Validate() error {
-	if len(o.ConfigFile) == 0 {
```
deads2k: oops, this we need.

enj:
@deads2k comments addressed.

PR #111, opened 2018-11-27
installer: fix revision update events

sttts:


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -199,8 +199,10 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.Op
 				if !reflect.DeepEqual(originalOperatorStatus, operatorStatus) {
 					_, updateError := c.operatorConfigClient.UpdateStatus(resourceVersion, operatorStatus)
 					if updateError == nil {
-						c.eventRecorder.Eventf("NodeTargetRevisionChanged", "Moving node %q from revision %d to %d", currNodeState.NodeName,
-							currNodeState.CurrentRevision, currNodeState.TargetRevision)
+						if currNodeState.CurrentRevision != newCurrNodeState.CurrentRevision {
```
deads2k: seems like we may want to know more about the change, but I'm fine starting here.

sttts: This fixes the "updating from 1 to 0" event.

PR #112, opened 2018-11-27
tolerate imperfect pod specs

deads2k:
If we are missing an envVar, we should not straight out fail.  Instead we can take a best guess for the owner.

/assign @mfojtik


deads2k:
This appears to work locally, but the actual owner is wrong.

PR #114, opened 2018-11-27
events: wire events directly to resourceapply

mfojtik:
This wire event recorder to all resourceapply functions. Any operator using these should pass an event recorder.


deads2k:
/lgtm

unit test failure

mfojtik:
@deads2k fixed.

PR #116, opened 2018-11-27
Support namespaced operator config

csrwng:
Adds support for namespaced configs to `EnsureOperatorConfigExists`


deads2k:
I don't think this is a goal.  All the use-cases we have for our SLOs so far deal with either single items or highly privileged items.  If we go to unprivileged or many items, we won't be auto-creating anyway.

PR #117, opened 2018-11-28
wire events to config observer functions

mfojtik:
This is needed so we can wire events into https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/etcd/observe_etcd.go#L17 

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/configobserver/config_observer_controller.go
+++ b/pkg/operator/configobserver/config_observer_controller.go
@@ -45,7 +45,7 @@ type Listers interface {
 // observedConfig that would cause the service being managed by the operator to crash. For example, if a required
 // configuration key cannot be observed, consider reusing the configuration key's previous value. Errors that occur
 // while attempting to generate the observedConfig should be returned in the errs slice.
-type ObserveConfigFunc func(listers Listers, existingConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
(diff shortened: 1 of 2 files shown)


pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -45,7 +45,7 @@ type Listers interface {
 // observedConfig that would cause the service being managed by the operator to crash. For example, if a required
 // configuration key cannot be observed, consider reusing the configuration key's previous value. Errors that occur
 // while attempting to generate the observedConfig should be returned in the errs slice.
-type ObserveConfigFunc func(listers Listers, existingConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
+type ObserveConfigFunc func(listers Listers, recorder events.Recorder, existingConfig map[string]interface{}) (observedConfig map[string]interface{}, errs []error)
```
sttts: alternative: return slice of events.

mfojtik: @sttts why you think that is better? ;)

mfojtik: (it will mean every observe config func will have to use inmemory recorder and return events...), why not just create events directly?

PR #119, opened 2018-11-28
events: use deployment as ownerRef involved object

mfojtik:


pkg/operator/events/recorder.go
```diff
@@ -40,13 +41,18 @@ func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*cor
 	if err != nil {
 		return nil, err
 	}
-	ownerRef := metav1.GetControllerOf(pod)
+	replicaSetOwnerRef := metav1.GetControllerOf(pod)
+	rs, err := rsClient.Get(replicaSetOwnerRef.Name, metav1.GetOptions{})
```
deads2k: only do this if the podOwnerRef is of replicaset.

mfojtik: good catch, fixed.

pkg/operator/events/recorder.go
```diff
@@ -40,13 +41,24 @@ func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*cor
 	if err != nil {
 		return nil, err
 	}
-	ownerRef := metav1.GetControllerOf(pod)
+	replicaSetOwnerRef := metav1.GetControllerOf(pod)
```
deads2k: `podOwnerRef =`

pkg/operator/events/recorder.go
```diff
@@ -40,13 +41,24 @@ func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*cor
...
+	if replicaSetOwnerRef == nil || replicaSetOwnerRef.Kind != "ReplicaSet" {
+		return nil, fmt.Errorf("pod %s is not controlled by replicaSet", podName)
```
deads2k: this isn't an error, just return what you have.  We have operators that daemonsets for instance

pkg/operator/events/recorder.go
```diff
@@ -40,13 +41,24 @@ func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*cor
...
+	}
+	rs, err := rsClient.Get(replicaSetOwnerRef.Name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	deploymentOwnerRef := metav1.GetControllerOf(rs)
```
deads2k: `replicaSetOwnerRef`

pkg/operator/events/recorder.go
mfojtik: @deads2k oh, I forgot about daemonsets... will need to handle that case as well

pkg/operator/events/recorder.go
```diff
@@ -40,14 +41,35 @@ func GetControllerReferenceForCurrentPod(client corev1client.PodInterface) (*cor
 	if err != nil {
 		return nil, err
 	}
-	ownerRef := metav1.GetControllerOf(pod)
+	// get replica set owner reference from pod, if we fail use the pod itself as involved object
+	replicaSetOwnerRef := metav1.GetControllerOf(pod)
+	if replicaSetOwnerRef == nil || replicaSetOwnerRef.Kind != "ReplicaSet" {
+		return makeObjectReference(&metav1.OwnerReference{Kind: "Pod", Name: pod.Name, UID: pod.UID, APIVersion: "v1"}, pod.Namespace), nil
```
deads2k: so if the pod were owned by a daemonset, this would return an indication that it was linked to a pod?

Perhaps it will be easier to express as a recursive function that takes an object reference as input and returns the most root owner.  Each stage calling back into itself.  A switch case for handled types.

pkg/operator/events/recorder.go
```diff
@@ -31,23 +32,60 @@ var podNameEnvFunc = func() string {
...
+			return GetControllerReferenceForCurrentPod(client, targetNamespace, makeObjectReference(rsController, targetNamespace))
+		}
+		// This is a replicaSet without any ownerReference
+		return reference, nil
+	case "DaemonSet":
+		return reference, nil
+	case "Deployment":
+		return reference, nil
+	default:
+		return nil, fmt.Errorf("unknown reference kind: %v", reference.Kind)
```
deads2k: if you have a reference, you can return that reference instead of failing, right?

deads2k:
simple comment, but I think it's worth fixing.

/approve

lgtm otherwise, feel free to self-tag

PR #121, opened 2018-11-28
events: add events to operator status controller

mfojtik:
Add events to operator status controllers.

/cc @deads2k


deads2k:
I like the idea, the code itself has changed quite a bit though.

PR #125, opened 2018-11-29
resourceapply: fix event helpers messages

mfojtik:


mfojtik:
@sttts @deads2k noticed i returned wrong order, this is just fixing that... seft-lgtm so I can pick it up in bumps.

PR #127, opened 2018-12-02
Add proc reaper for PID 1 containerized processes

smarterclayton:
On Linux pid 1 processes (container processes) must reap their children.
This library copied directly from openshift/origin.


smarterclayton:
No dependencies and trivial.  Prerequisite for moving router out of origin.

PR #133, opened 2018-12-04
wip: add revision metrics

mfojtik:


pkg/operator/staticpod/controller/revision/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+func (p prometheusMetricsProvider) NewRevisionMetrics() *prometheus.GaugeVec {
+	revisionMetrics := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Subsystem: p.componentName,
+			Name:      "latestRevision",
+			Help:      fmt.Sprintf("Current %s operator revision.", p.componentName),
+		},
+		[]string{"observedGeneration"},
+	)
+	if err := prometheus.Register(revisionMetrics); err != nil {
```
sttts: note that this must be a singleton in the process. If we call this from tests, we have to wrap that into some `sync.Once` or similar.

PR #135, opened 2018-12-04
configobserver: move install-config CIDR readers here

sttts:
This code is used by multiple operators.

### DIFF
--- a/pkg/operator/configobserver/network/observe_network.go
+++ b/pkg/operator/configobserver/network/observe_network.go
@@ -1,0 +1,112 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/ghodss/yaml"
+	"github.com/golang/glog"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/listers/core/v1"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const (
+	clusterConfigNamespace = "kube-system"
+	clusterConfigName      = "cluster-config-v1"
+)
+
+// GetClusterCIDRs reads the cluster CIDRs from the install-config ConfigMap in the cluster.
+func GetClusterCIDRs(lister v1.ConfigMapLister, recorder events.Recorder) ([]string, error) {
+	clusterConfig, err := lister.ConfigMaps(clusterConfigNamespace).Get(clusterConfigName)
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveClusterCIDRFailed", "Required %s/%s config map not found", clusterConfigNamespace, clusterConfigName)
+		glog.Warning("configmap/cluster-config-v1.kube-system: not found")
+		return nil, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	installConfigYaml, ok := clusterConfig.Data["install-config"]
+	if !ok {
+		glog.Warning("configmap/cluster-config-v1.kube-system: install-config not found")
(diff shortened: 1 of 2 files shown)


pkg/operator/configobserver/network/observe_network.go
```diff
@@ -0,0 +1,111 @@
+package network
+
+import (
+	"fmt"
+
+	"github.com/ghodss/yaml"
+	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/operator/events"
```
mfojtik: nit: imports

pkg/operator/configobserver/network/observe_networking_test.go
```diff
@@ -0,0 +1,160 @@
+package network
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/openshift/library-go/pkg/operator/events"
```
mfojtik: nit: move this down

pkg/operator/configobserver/network/observe_network.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/client-go/listers/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
```
mfojtik: @sttts @deads2k we hardcode this in so many places, I almost wish we have some const package in library-go to control this from one place?

pkg/operator/configobserver/network/observe_network.go
```diff
@@ -0,0 +1,111 @@
...
+	clusterConfigNamespace = "kube-system"
+	clusterConfigName      = "cluster-config-v1"
+)
+
+// GetClusterCIDRs reads the cluster CIDRs from the install-config ConfigMap in the cluster.
+func GetClusterCIDRs(lister v1.ConfigMapLister, recorder events.Recorder) ([]string, error) {
+	clusterConfig, err := lister.ConfigMaps(clusterConfigNamespace).Get(clusterConfigName)
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveClusterCIDRFailed", "Required %s/%s config map not found", clusterConfigNamespace, clusterConfigName)
+		glog.Warning("configmap/cluster-config-v1.kube-system: not found")
```
mfojtik: @deads2k @sttts every event also generate a glog message when created... I wonder if we can deduplicate these to get cleaner logs?

pkg/operator/configobserver/network/observe_network.go
sttts: fixed

pkg/operator/configobserver/network/observe_network.go
sttts: a configobserver/log package. Same for events.

mfojtik:
/lgtm

follow-up 1: cleanup the logging/event reporting code
follow-up 2: have some pkg/consts package with common namespaces/names?

pkg/operator/configobserver/network/observe_network.go
deads2k: > @sttts @deads2k we hardcode this in so many places, I almost wish we have some const package in library-go to control this from one place?

no const package

pkg/operator/configobserver/network/observe_network.go
deads2k: > @deads2k @sttts every event also generate a glog message when created... I wonder if we can deduplicate these to get cleaner logs?

you mean just delete the glog?

deads2k:
> follow-up 2: have some pkg/consts package with common namespaces/names?

No.  We should not do that.  People own their own location.  For example, we know these are moving.

PR #139, opened 2018-12-05
WIP: Add generic operator status metrics collector

mfojtik:
This will add prometheus metrics collector that observe changes in operator status (both normal operator and static pod operator) and converts the status into prometheus metrics.

There are few example metrics registered, we can add more (anything that we can derive from status).

### DIFF
--- a/pkg/operator/metrics/collector.go
+++ b/pkg/operator/metrics/collector.go
@@ -1,0 +1,215 @@
+package metrics
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+const (
+	openshiftOperatorNamePrefix = "operator_v1_openshift_io"
+	workkey                     = "key"
+)
+
+// OperatorStateGetter knows how to get the current operator state and provides operator client informer.
+type OperatorStateGetter interface {
+	GetOperatorState() (spec *operatorv1.OperatorSpec, status *operatorv1.OperatorStatus, staticPodStatus *operatorv1.StaticPodOperatorStatus, resourceVersion string, err error)
+	Informer() cache.SharedIndexInformer
+}
+
+// operatorMetricUpdater knows how to update a single tracked metric.
+type operatorMetricUpdater interface {
+	Update(*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, *operatorv1.StaticPodOperatorStatus) error
+}
+
+type operatorMetricUpdateFn func(*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, *operatorv1.StaticPodOperatorStatus) ([]metricValue, error)
+
+type Collector struct {
+	operatorClient  OperatorStateGetter
+	operatorMetrics []operatorMetricUpdater
+	targetNamespace string
+
+	queue          workqueue.RateLimitingInterface
+	hasSynced      cache.InformerSynced
+	collectorMutex sync.RWMutex
+}
+
+// metricValue represents a single metric value sample
+type metricValue struct {
+	value  float64
+	labels []string
+}
+
+// operatorMetrics represents a single metric definition
+type operatorMetric struct {
+	prometheusDesc *prometheus.Desc
+	valueType      prometheus.ValueType
+
+	// values are the recently observed metric values provided by the updateFn
+	values []metricValue
+
+	// updateFn know how to gather the values for this metric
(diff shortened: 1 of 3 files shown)


mfojtik:
/hold

Will require proof that it works in real operator.

mfojtik:
@deads2k @sttts this is ready for review now, I tested it in kube apiserver operator and it works as expected.

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// operatorMetricUpdater knows how to update a single tracked metric.
+type operatorMetricUpdater interface {
+	Update(*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, *operatorv1.StaticPodOperatorStatus) error
+}
+
+type operatorMetricUpdateFn func(*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, *operatorv1.StaticPodOperatorStatus) ([]metricValue, error)
+
+type Collector struct {
```
sttts: godoc

sttts: What is this thing doing?

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+
+const (
+	openshiftOperatorNamePrefix = "operator_v1_openshift_io"
+	workkey                     = "key"
+)
+
+// OperatorStateGetter knows how to get the current operator state and provides operator client informer.
+type OperatorStateGetter interface {
+	GetOperatorState() (spec *operatorv1.OperatorSpec, status *operatorv1.OperatorStatus, staticPodStatus *operatorv1.StaticPodOperatorStatus, resourceVersion string, err error)
+	Informer() cache.SharedIndexInformer
```
sttts: why is `Informer()` not just a param to NewOperatorStatusMetricsCollector?

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+	defer m.valueMutex.Unlock()
+	currentValues, err := m.updateFn(spec, status, staticPodStatus)
+	if err != nil {
+		return err
+	}
+	m.values = currentValues
+	return nil
+}
+
+// NewOperatorStatusMetricsCollector returns a metrics collector that automatically collect prometheus metrics from given operator status.
```
sttts: nit: collects

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// register default set of metrics we track
+	addStatusConditionsMetrics(collector)
+	addStaticPodStatusLatestRevisionMetrics(collector)
+	addStaticPodNodesMetrics(collector)
+
+	client.Informer().AddEventHandler(collector.eventHandler())
+
+	// NOTE: Only call this once and this must be called after all metrics are registered otherwise prometheus will panic.
+	prometheus.MustRegister(collector)
```
sttts: can and should we wrap this somehow in a `Once` struct? Not being able to call a constructor twice is odd.

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+	valueType      prometheus.ValueType
+
+	// values are the recently observed metric values provided by the updateFn
+	values []metricValue
+
+	// updateFn know how to gather the values for this metric
+	updateFn operatorMetricUpdateFn
+
+	// valueMutex protects from values changes
+	valueMutex sync.RWMutex
```
sttts: just `values`? By convention, but locks directly above (without empty lines) the objects they protect.

pkg/operator/metrics/collector.go
```diff
@@ -0,0 +1,215 @@
...
+	defer c.collectorMutex.RUnlock()
+	for _, m := range c.operatorMetrics {
+		ch <- m.(*operatorMetric).prometheusDesc
+	}
+}
+
+// Collect implements Prometheus Collector interface
+func (c *Collector) Collect(ch chan<- prometheus.Metric) {
+	c.collectorMutex.RLock()
+	c.collectorMutex.RUnlock()
```
sttts: no defer?

pkg/operator/metrics/status_metrics.go
```diff
@@ -0,0 +1,118 @@
...
+			for _, c := range conditions {
+				values = append(values, metricValue{value: 1, labels: []string{c.Type, string(c.Status)}})
+			}
+			return values, nil
+		})
+	c.register(
+		"version",
+		"Tracks operator version for given observed generation",
+		prometheus.GaugeValue,
+		[]string{"observed_generation", "version"},
```
sttts: version as value and as label at the same time?

pkg/operator/metrics/status_metrics.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+func addStatusConditionsMetrics(c *Collector) {
+	c.register(
```
sttts: none of these uses `c` other than for calling `register`. Why not make them independent funcs and call `c.register` at a central place. Would make the code easier and maybe help with solving this "prometheus panic on second call" problem.

mfojtik:
will rework this and reopen.

PR #141, opened 2018-12-06
Static pod revision history prune command

damemi:
Contains just the `Prune` command for static pod installer revisions first created in https://github.com/openshift/library-go/pull/82

part of https://jira.coreos.com/browse/MSTR-221

### DIFF
--- a/pkg/operator/staticpod/prune/cmd.go
+++ b/pkg/operator/staticpod/prune/cmd.go
@@ -1,0 +1,116 @@
+package prune
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"strconv"
+	"strings"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/golang/glog"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type PruneOptions struct {
+	MaxEligibleRevisionID int
+	ProtectedRevisionIDs  []int
+
+	ResourceDir   string
+	StaticPodName string
+}
+
+func NewPruneOptions() *PruneOptions {
+	return &PruneOptions{}
+}
+
+func NewPrune() *cobra.Command {
+	o := NewPruneOptions()
+
+	cmd := &cobra.Command{
+		Use:   "prune",
+		Short: "Prune static pod installer revisions",
+		Run: func(cmd *cobra.Command, args []string) {
+			glog.V(1).Info(cmd.Flags())
+			glog.V(1).Info(spew.Sdump(o))
+
+			if err := o.Validate(); err != nil {
+				glog.Fatal(err)
+			}
+			if err := o.Run(); err != nil {
+				glog.Fatal(err)
+			}
+		},
+	}
+
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&o.MaxEligibleRevisionID, "max-eligible-id", o.MaxEligibleRevisionID, "highest revision ID to be eligible for pruning")
+	fs.IntSliceVar(&o.ProtectedRevisionIDs, "protected-ids", o.ProtectedRevisionIDs, "list of revision IDs to preserve (not delete)")
+	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
+	fs.StringVar(&o.StaticPodName, "static-pod-name", o.StaticPodName, "name of the static pod")
+}
+
+func (o *PruneOptions) Validate() error {
+	if len(o.ResourceDir) == 0 {
+		return fmt.Errorf("--resource-dir is required")
+	}
+	if o.MaxEligibleRevisionID == 0 {
+		return fmt.Errorf("--max-eligible-id is required")
+	}
+	if len(o.StaticPodName) == 0 {
+		return fmt.Errorf("--static-pod-name is required")
+	}
+
+	return nil
+}
+
+func (o *PruneOptions) Run() error {
+	protectedIDs := sets.NewInt(o.ProtectedRevisionIDs...)
+
+	files, err := ioutil.ReadDir(o.ResourceDir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is not a resource directory...
+		if !file.IsDir() {
+			continue
+		}
+		// And doesn't match our static pod prefix...
+		if !strings.HasPrefix(file.Name(), o.StaticPodName) {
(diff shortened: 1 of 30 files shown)


pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.MaxEligibleRevisionID, "max-eligible-id", o.MaxEligibleRevisionID, "highest revision ID to be eligible for pruning")
+	fs.StringSliceVar(&o.ProtectedRevisionIDs, "protected-ids", o.ProtectedRevisionIDs, "list of revision IDs to reserve from being pruned")
```
deads2k: list of revisionIDs to preserve (not delete)

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	if len(o.StaticPodName) == 0 {
+		return fmt.Errorf("--static-pod-name is required")
+	}
+
+	return nil
+}
+
+func (o *PruneOptions) Run() error {
+	if err := o.pruneRevisions(); err != nil {
```
deads2k: move content of this function right here and don't wrap the error.

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+		return fmt.Errorf("error pruning installerpod resource directory revisions: %v", err)
+	}
+	return nil
+}
+
+func (o *PruneOptions) pruneRevisions() error {
+	protectedIDs := make(map[string]struct{}, len(o.ProtectedRevisionIDs))
```
deads2k: sets.String

deads2k: `protectedDirectories`

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+	for _, id := range o.ProtectedRevisionIDs {
+		protectedIDs[id] = struct{}{}
```
deads2k: the the substitution here: `protectedDirectories.Insert(prefix+"-" + revisionId)`

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	files, err := ioutil.ReadDir(o.ResourceDir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is a directory AND the revision is not protected
```
deads2k: invert and create separate if for each one if you please.

pkg/operator/staticpod/prune/cmd.go
deads2k: > protectedDirectories

actually, looking at the logic, you should have `protectedRevisionsId` and then do the separate below in the directory check

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+		},
+	}
+
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.MaxEligibleRevisionID, "max-eligible-id", o.MaxEligibleRevisionID, "highest revision ID to be eligible for pruning")
```
deads2k: because we have a max, these are ints

pkg/operator/staticpod/prune/cmd.go
deads2k: > sets.String

sets.integer

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is a directory AND the revision is not protected
+		// AND it is a revision of the pod we want AND it is less than or equal to the max eligible revision
+		if _, protected := protectedIDs[file.Name()]; !protected &&
+			file.IsDir() &&
+			strings.HasPrefix(file.Name(), o.StaticPodName) &&
```
deads2k: please split the name after doing this to show the integer part

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,107 @@
...
+
+	files, err := ioutil.ReadDir(o.ResourceDir)
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		// If the file is a directory AND the revision is not protected
+		// AND it is a revision of the pod we want AND it is less than or equal to the max eligible revision
+		if _, protected := protectedIDs[file.Name()]; !protected &&
```
deads2k: this detection here wont' work when you pass in the revisionID which won't match full filenames.  Please make an explicit test that demonstrates this failure.

pkg/operator/staticpod/prune/cmd_test.go
```diff
@@ -0,0 +1,93 @@
...
+	tests := []struct {
+		name     string
+		o        PruneOptions
+		files    []string
+		expected []string
+	}{
+		{
+			name: "only deletes non-protected revisions of the specified pod",
+			o: PruneOptions{
+				MaxEligibleRevisionID: "test-3",
```
deads2k: revisions are ints.  integers have Max, strings may technically have them but logically they dont'

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+		if !file.IsDir() {
+			continue
+		}
+		// And doesn't match our static pod prefix...
+		if !strings.HasPrefix(file.Name(), o.StaticPodName) {
+			continue
+		}
+
+		// Split file name to get just the integer revision ID
+		fileSplit := strings.Split(file.Name(), fmt.Sprintf("%s-", o.StaticPodName))
```
deads2k: seems like ensuring the remainder after the index of the last `-` is an integer will be more accurate.  Ones that aren't ints are not failures, they are just like `mypod-isspecial-1` when you wanted `mypod-1`

deads2k: > seems like ensuring the remainder after the index of the last `-` is an integer will be more accurate. Ones that aren't ints are not failures, they are just like `mypod-isspecial-1` when you wanted `mypod-1`

I'll take a followup

deads2k: > I'll take a followup

along with a test that ensure it doesn't slip.  prefix checking is the bane of testing.

deads2k:
solid starting point.

/lgtm

PR #142, opened 2018-12-06
add functionality required to rotate certificates

deads2k:
1. filter expired certificates

@openshift/sig-auth fyi

WIP while I work through what is needed.


pkg/crypto/crypto.go
```diff
@@ -454,6 +476,18 @@ func GetCA(certFile, keyFile, serialFile string) (*CA, error) {
 	}, nil
 }
 
+func GetCAFromBytes(certBytes, keyBytes []byte) (*CA, error) {
+	caConfig, err := GetTLSCertificateConfigFromBytes(certBytes, keyBytes)
```
ericavonb: It seems unnecessary to have both `GetCAFromBytes` and `GetTLSCertificateConfigFromBytes`. Can you simplify it before we merge?

mrogers950:
@deads2k is this still a WIP? It looks OK to me at the moment.

PR #145, opened 2018-12-10
staticpod: make installer pod yaml file

mfojtik:
We should move embedded objects into YAML files and use them as assets.. Will help to see what is parametrized and what not.

/cc @sttts


pkg/operator/resource/resourceapply/core.go
```diff
@@ -71,6 +71,29 @@ func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder,
 	return actual, true, err
 }
 
+// ApplyPod merges objectmeta, does not worry about anything else
+func ApplyPod(client coreclientv1.PodsGetter, recorder events.Recorder, required *corev1.Pod) (*corev1.Pod, bool, error) {
```
mfojtik: @deads2k not sure if we want this here, but I don't see any problem with it.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -417,35 +425,66 @@ func getInstallerPodName(revision int32, nodeName string) string {
...
-	required.Name = getInstallerPodName(revision, nodeName)
-	required.Namespace = c.targetNamespace
-	required.Spec.NodeName = nodeName
-	required.Spec.Containers[0].Image = c.installerPodImageFn()
-	required.Spec.Containers[0].Command = c.command
-	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
-		fmt.Sprintf("-v=%d", 4),
+// arrToYAML converts a string array to YAML array
+// Example: []string{"1", "2"} will convert to: '["1","2"]'
+func arrToYAML(arr []string) string {
```
mfojtik: probably should be some string helper somewhere?

pkg/operator/resource/resourceapply/core.go
sttts: shouldn't this at least be its own commit?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -3,11 +3,15 @@ package installer
...
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/golang/glog"
+	"github.com/openshift/library-go/pkg/assets"
```
sttts: order

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: json.Marshal would do, no?

mfojtik: good idea

mfojtik:
@sttts made this two commits, switched to use json.Marshal and added some more tests to verify that it produces what is expected from it to produce.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -417,35 +425,66 @@ func getInstallerPodName(revision int32, nodeName string) string {
...
-	required.Name = getInstallerPodName(revision, nodeName)
-	required.Namespace = c.targetNamespace
-	required.Spec.NodeName = nodeName
-	required.Spec.Containers[0].Image = c.installerPodImageFn()
-	required.Spec.Containers[0].Command = c.command
-	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
-		fmt.Sprintf("-v=%d", 4),
+// toJSONArray converts a string array into JSON format that can be used in array field in YAML (args, command, etc.)
+// Example: []string{"1", "2"} will convert to: '["1","2"]'
+func toJSONArray(arr []string) string {
```
damemi: This would still be nice to have as a helper somewhere if we want to switch away from the inline pod definitions

mfojtik: agree, probably something in assets package or new "string" package.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -22,13 +24,21 @@ import (
...
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/bindata"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
-const operatorStatusInstallerControllerFailing = "InstallerControllerFailing"
-const installerControllerWorkQueueKey = "key"
+const (
+	operatorStatusInstallerControllerFailing = "InstallerControllerFailing"
+	installerControllerWorkQueueKey          = "key"
+	manifestDir                              = "pkg/operator/staticpod/controller/installer"
```
sttts: nit: empty line above

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -419,33 +429,32 @@ func getInstallerPodName(revision int32, nodeName string) string {
...
 // ensureInstallerPod creates the installer pod with the secrets required to if it does not exist already
 func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *operatorv1.OperatorSpec, revision int32) error {
-	required := resourceread.ReadPodV1OrDie([]byte(installerPod))
-	required.Name = getInstallerPodName(revision, nodeName)
-	required.Namespace = c.targetNamespace
-	required.Spec.NodeName = nodeName
-	required.Spec.Containers[0].Image = c.installerPodImageFn()
-	required.Spec.Containers[0].Command = c.command
-	required.Spec.Containers[0].Args = append(required.Spec.Containers[0].Args,
-		fmt.Sprintf("-v=%d", 4),
```
sttts: this was lost

PR #146, opened 2018-12-10
Pruning controller for static pod installer revisions

damemi:
Follow up to https://github.com/openshift/library-go/pull/141

Based off of https://github.com/openshift/library-go/pull/82

Creates just the controller which interprets revision statuses and makes the call to the pruner pod


pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+const (
+	pruneControllerWorkQueueKey = "key"
+	statusConfigMapName         = "revision-status"
+	defaultResourceDir          = "/etc/kubernetes/static-pod-resources"
+	prunerPodName               = "revision-pruner"
+)
+
+func NewPruneController(
```
mfojtik: nit: godoc

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	config, err := c.dynamicClient.Resource(c.operatorConfigGVR).Get(c.operatorConfigName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	revisionLimits, ok, err := unstructured.NestedMap(config.Object, "spec", "operatorRevisionLimit")
+	if err != nil {
+		return err
+	}
+	if ok {
```
mfojtik: nit: i would flip this... `if !ok { return nil }`. reason is to reduce the levels of nesting in code

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+		if val, set := revisionLimits["failed"]; set {
+			c.failedRevisionLimit = val.(int)
```
mfojtik: we probably want to check if the value is really int? in case it is not, this will probably return 0 and will have weird side-effects?

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (c *PruneController) pruneRevisionHistory(operatorStatus *operatorv1.StaticPodOperatorStatus, resourceDir string) error {
+	var succeededRevisionIDs, failedRevisionIDs []int
+
+	configMaps, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	for _, configMap := range configMaps.Items {
+		if strings.HasPrefix(configMap.Name, statusConfigMapName) {
```
mfojtik: nit: flip this (`if !strings.HasPrefix() { continue }`)

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	protectedFailedRevisionIDs := failedRevisionIDs[revisionKeyToStart(len(failedRevisionIDs), c.failedRevisionLimit):]
+
+	excludedIDs := make([]int, 0, len(protectedSucceededRevisionIDs)+len(protectedFailedRevisionIDs))
+	excludedIDs = append(excludedIDs, protectedSucceededRevisionIDs...)
+	excludedIDs = append(excludedIDs, protectedFailedRevisionIDs...)
+	sort.Ints(excludedIDs)
+
+	// Run pruning pod on each node and pin it to that node
+	for _, nodeStatus := range operatorStatus.NodeStatuses {
+		err = c.ensurePrunePod(prunerPodName, nodeStatus.NodeName, excludedIDs[len(excludedIDs)-1], excludedIDs, nodeStatus.TargetRevision)
```
mfojtik: nit: `if err:=` (do not shadow err)

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+
+func revisionKeyToStart(length, limit int) int {
+	if length > limit {
+		return length - limit
+	}
+	return 0
+}
+
+func (c *PruneController) ensurePrunePod(podName, nodeName string, maxEligibleRevision int, protectedRevisions []int, revision int32) error {
+	required := resourceread.ReadPodV1OrDie([]byte(prunerPod))
```
mfojtik: i think we will prefer to use bindata for manifests instead of embedding the manifests in code via const.

see https://github.com/openshift/library-go/pull/145 for example

/cc @sttts

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	required.Name = fmt.Sprintf("%s-%d-%s", podName, revision, nodeName)
```
mfojtik: nit: make this function with 3 parameters

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	if _, err := c.kubeClient.CoreV1().Pods(c.targetNamespace).Create(required); err != nil && !apierrors.IsAlreadyExists(err) {
+		c.eventRecorder.Eventf("PrunePodFailed", "Failed to create pod for %s: %v", resourceread.WritePodV1OrDie(required), err)
+		return err
+	}
+	return nil
+}
+
+func revisionsToString(revisions []int) string {
+	revisionString := ""
+	if len(revisions) > 0 {
```
mfojtik: nit: `if len(revisions) == 0 { return "" }`

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+		for i := 0; i < len(revisions)-1; i++ {
+			revisionString = fmt.Sprintf("%s,%d", revisionString, revisions[i])
```
mfojtik: this almost looks like you want to use `strings.Join` ? ;-)

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,296 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, operatorStatus, _, err := c.operatorConfigClient.Get()
+	if err != nil {
+		return err
+	}
+
+	err = c.pruneRevisionHistory(operatorStatus, defaultResourceDir)
```
mfojtik: nit: just `return c.pruneRevisionHistory(operatorStatus, defaultResourceDir)`

pkg/operator/staticpod/controller/prune/prune_controller.go
damemi: I do! but I have a slice of ints and `strings.Join` takes a slice of strings, so I figured this hack was about as many steps as converting the ints to strings which would still have to be one-by-one I think

damemi: can probably come up with a better way to do this though

damemi:
@mfojtik updated with your feedback, will need to see what happens with https://github.com/openshift/library-go/pull/145 to depend on the bindata stuff so I've got that as a separate commit for now

pkg/operator/staticpod/controller/prune/manifests/pruner-pod.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: {{ .TargetNamespace }}
```
deads2k: strongly dislike

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+		if revision, ok := configMap.Data["revision"]; ok {
+			revisionID, err := strconv.Atoi(revision)
+			if err != nil {
+				return err
+			}
+			switch configMap.Data["phase"] {
+			case string(corev1.PodSucceeded):
+				succeededRevisionIDs = append(succeededRevisionIDs, revisionID)
+			case string(corev1.PodFailed):
+				failedRevisionIDs = append(failedRevisionIDs, revisionID)
```
deads2k: add a default that errors

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+
+func revisionKeyToStart(length, limit int) int {
+	if length > limit {
+		return length - limit
+	}
+	return 0
+}
+
+// Example: []string{"1", "2"} will convert to: '["1","2"]'
+func toJSONArray(arr []string) string {
```
deads2k: blech. remove

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+			switch configMap.Data["phase"] {
+			case string(corev1.PodSucceeded):
+				succeededRevisionIDs = append(succeededRevisionIDs, revisionID)
+			case string(corev1.PodFailed):
+				failedRevisionIDs = append(failedRevisionIDs, revisionID)
+			}
+		}
+	}
+
+	sort.Ints(succeededRevisionIDs)
```
deads2k: check length.  If you have none, you don't need to prune anything and you should return early.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+	eventRecorder events.Recorder
+
+	dynamicClient      dynamic.Interface
+	operatorConfigGVR  schema.GroupVersionResource
+	operatorConfigName string
+}
+
+const (
+	pruneControllerWorkQueueKey = "key"
+	statusConfigMapName         = "revision-status"
```
deads2k: add a dash.  `revision-status-`

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+	defaultResourceDir          = "/etc/kubernetes/static-pod-resources"
+
+	manifestDir           = "pkg/operator/staticpod/controller/prune"
```
deads2k: just inline this

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+	manifestPrunerPodPath = "manifests/pruner-pod.yaml"
```
deads2k: and this

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+		fmt.Sprintf("--protected-ids=%s", revisionsToString(protectedRevisions)),
+		fmt.Sprintf("--resource-dir=%s", defaultResourceDir),
+		fmt.Sprintf("--static-pod-name=%s", c.podResourcePrefix),
+	}
+	config.Args = toJSONArray(args)
+
+	return assets.MustCreateAssetFromTemplate(manifestPrunerPodPath, bindata.MustAsset(filepath.Join(manifestDir, manifestPrunerPodPath)), config).Data, nil
+}
+
+func (c *PruneController) ensurePrunePod(podName, nodeName string, maxEligibleRevision int, protectedRevisions []int, revision int32) error {
```
deads2k: podname is dead

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+const (
+	pruneControllerWorkQueueKey = "key"
+	statusConfigMapName         = "revision-status"
+	defaultResourceDir          = "/etc/kubernetes/static-pod-resources"
+
+	manifestDir           = "pkg/operator/staticpod/controller/prune"
+	manifestPrunerPodPath = "manifests/pruner-pod.yaml"
+	prunerPodName         = "revision-pruner"
```
deads2k: and this

deads2k:
needs unit tests.  This matches my expectations.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,304 @@
...
+
+	dynamicClient      dynamic.Interface
+	operatorConfigGVR  schema.GroupVersionResource
+	operatorConfigName string
+}
+
+const (
+	pruneControllerWorkQueueKey = "key"
+	statusConfigMapName         = "revision-status-"
+	defaultResourceDir          = "/etc/kubernetes/static-pod-resources"
```
deads2k: making magic values into constants does not improve code quality.  do this if it is used in many places or possible to change. this is neither.  Just inline

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	eventRecorder events.Recorder,
+	dynamicClient dynamic.Interface,
+	operatorConfigGVR schema.GroupVersionResource,
+	operatorConfigName string,
+) *PruneController {
+	c := &PruneController{
+		targetNamespace:        targetNamespace,
+		podResourcePrefix:      podResourcePrefix,
+		command:                command,
+		failedRevisionLimit:    0,
```
deads2k: unsafe default, choose 5

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,304 @@
...
+		queue:            workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "PruneController"),
+		prunerPodImageFn: getPrunerPodImageFromEnv,
+	}
+
+	operatorConfigClient.Informer().AddEventHandler(c.eventHandler())
+
+	return c
+}
+
+func (c *PruneController) syncRevisionLimits() error {
```
deads2k: remove this for now.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	limit, err := strconv.Atoi(val.(string))
+	if err != nil {
+		return 0, fmt.Errorf("error converting %s revision limit %v to int: %v", key, val, err)
+	}
+	return limit, nil
+}
+
+func (c *PruneController) pruneRevisionHistory(operatorStatus *operatorv1.StaticPodOperatorStatus) error {
```
deads2k: followup to remove the resources from the API server.

deads2k:
after removing that method and setting a good default, lgtm

/approve

PR #150, opened 2019-01-02
staticpod: add monitoring resources controller

mfojtik:
This change adds a controller loop that manages the lifecycle of resources needed to make service monitoring work. Making this generic means that all operator which currently use library-go will automatically get this.

A non-generic attempt + history: https://github.com/openshift/cluster-kube-apiserver-operator/pull/172

/cc @deads2k 
/cc @sttts


pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ .ServiceMonitorName }}
```
mfojtik: question: do we want the name match the namespace? if yes, I can drop this parameter, but it felt weird.

mfojtik: (also I dunno if the name of the service monitor plays any role as this is namespaces resource, we can just call this "monitor" ;-)

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
```diff
@@ -0,0 +1,204 @@
...
+		clusterRoleBindingLister:       kubeInformersForTargetNamespace.Rbac().V1().ClusterRoleBindings().Lister(),
+
+		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "MonitoringResourceController"),
+		kubeClient:    kubeClient,
+		dynamicClient: dynamicClient,
+	}
+
+	operatorConfigClient.Informer().AddEventHandler(c.eventHandler())
+
+	// TODO: We need a dynamic informer here to observe changes to ServiceMonitor resource.
```
mfojtik: @sttts @deads2k right now, when something stomp the service monitor or delete it, we won't notice right away, but in next reconcile cycle for this controller (or when resources we monitor change)... 

To fix this, we will need dynamic informers which will be available in 1.13. Alternatively, as @sttts suggested, we can copy dynamicinformer package into library-go and then replace it. However, I don't think it is worth the effort for now and I can create follow up issue to add the informer when we get 1.13 ?

pkg/operator/staticpod/controller/monitoring/manifests/prometheus-role.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  # TODO this should be a clusterrole
```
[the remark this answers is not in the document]
mfojtik: /cc @openshift/sig-monitoring

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
```diff
@@ -0,0 +1,204 @@
...
+
+func (c MonitoringResourceController) mustTemplateAsset(name string) ([]byte, error) {
+	config := struct {
+		TargetNamespace     string
+		MonitoringNamespace string
+		ServiceMonitorName  string
+	}{
+		TargetNamespace:     c.targetNamespace,
+		ServiceMonitorName:  c.serviceMonitorName,
+		MonitoringNamespace: defaultMonitoringNamespace,
```
sttts: wouldn't mind to have this hard-coded in the template. We don't support customization anyway.

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// eventHandler queues the operator to check spec and status
+func (c *MonitoringResourceController) eventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(controllerWorkQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(controllerWorkQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(controllerWorkQueueKey) },
+	}
+}
```
sttts: anything in here that is not copied from the other controllers?

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
mfojtik: agree, if this change we can change in manifest

pkg/operator/staticpod/controllers.go
```diff
@@ -60,19 +66,30 @@ func NewControllers(targetNamespaceName, staticPodName string, command, revision
...
+		dynamicClient,
+		eventRecorder,
+	)
+
 	return controller
 }
 
 func (o *staticPodOperatorControllers) Run(stopCh <-chan struct{}) {
-	go o.serviceAccountController.Run(1, stopCh)
+	go o.backingResourceController.Run(1, stopCh)
```
sttts: nit: order as in the struct

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
mfojtik: this is basically a backing resource controller with some extra logic around service monitor (resourceapply.ApplyDirect won't work)... Also failure to create monitoring resources should not set the operator to failed state.

mfojtik:
Comments addressed, waiting for @deads2k feedback.

pkg/operator/staticpod/controller/monitoring/manifests/prometheus-role.yaml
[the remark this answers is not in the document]
s-urbaniak: @sttts @mfojtik yes yes, aware of this one. Give me a bit latency to recheck the necessity and appropriate settings here, I'll submit a PR then.

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+      interval: 30s
+      metricRelabelings:
+        - action: drop
+          regex: etcd_(debugging|disk|request|server).*
+          sourceLabels:
+            - __name__
+      port: https
+      scheme: https
+      tlsConfig:
+        # TODO where do you mount the service-ca.crt?
```
s-urbaniak:
```
    tlsConfig:
      caFile: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
```

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+        insecureSkipVerify: true
```
s-urbaniak: with `caFile` this can be dropped.

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
mfojtik: thanks!

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
mfojtik: done, thanks!

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,26 @@
...
+      interval: 30s
+      metricRelabelings:
+        - action: drop
+          regex: etcd_(debugging|disk|request|server).*
+          sourceLabels:
+            - __name__
+      port: https
+      scheme: https
+      tlsConfig:
+        caFile: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
```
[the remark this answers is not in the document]
s-urbaniak: @mfojtik @deads2k ack, my bad, I had an "old" 3.11 branch checked out, the new location is `/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt`

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: monitor
+  namespace: {{ .TargetNamespace }}
```
deads2k: Because we only have unstructured for now, I'm ok using a golang template.  With real objects, I prefer real assignment in golang code.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	requiredSpec, _, err := unstructured.NestedFieldNoCopy(required.UnstructuredContent(), "spec")
+	if err != nil {
+		return false, err
+	}
+
+	if equality.Semantic.DeepEqual(existingSpec, requiredSpec) {
```
deads2k: We want something different.  We want to make sure everything specified in our spec matches, but we cannot ensure nothing else is set.   Give us a helper that makes sure that every value we specify matches.

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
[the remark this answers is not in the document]
s-urbaniak: just fyi, cluster monitoring operator changed to the new location quite some time ago: https://github.com/openshift/cluster-monitoring-operator/pull/137

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
deads2k: > @sttts @deads2k right now, when something stomp the service monitor or delete it, we won't notice right away, but in next reconcile cycle for this controller (or when resources we monitor change)...

Just set a poll every few minutes for this controller.

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
```diff
@@ -0,0 +1,199 @@
...
+
+	serviceMonitorBytes, err := c.mustTemplateAsset("manifests/service-monitor.yaml")
+	if err != nil {
+		errs = append(errs, fmt.Errorf("manifests/service-monitor.yaml: %v", err))
+	} else {
+		_, serviceMonitorErr := resourceapply.ApplyServiceMonitor(c.dynamicClient, c.eventRecorder, serviceMonitorBytes)
+		errs = append(errs, serviceMonitorErr)
+	}
+
+	err = common.NewMultiLineAggregate(errs)
```
deads2k: TODO could we refactor this actually just be a `common.ErrorToConditionMessage()` that handles the "normal" aggregate?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -0,0 +1,188 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+var serviceMonitorGVR = schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors"}
+
+func mergeUnstructuredSlice(required, existing []interface{}) ([]interface{}, bool, error) {
```
deads2k: @sanchezl can we just use mergo?

@sanchezl did you pin mergo in glide.yaml to the level used in kube?  if not, please do so.

pkg/operator/staticpod/controller/monitoring/manifests/service-monitor.yaml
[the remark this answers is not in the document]
deads2k: I still see this as outstanding.

pkg/operator/resource/resourceapply/monitoring.go
sanchezl: mergo worked for me. The use of mergo got [moved into library-go](https://github.com/openshift/library-go/blob/ed90b992b08815b36c2b71c80218ea2135d8e4bd/pkg/operator/configobserver/config_observer_controller.go#L103).

pkg/operator/staticpod/controller/monitoring/monitoring_resource_controller.go
mfojtik: https://github.com/openshift/library-go/issues/155

mfojtik:
@deads2k refactored to `mergo.Merge`, included unit test to proof it works as expected.

PR #157, opened 2019-01-04
Set revision status ownerrefs when creating new revision

damemi:
When creating a new static pod deployment, sets the ownerrefs for copied resources to be the new revision-status configmap (see https://github.com/openshift/cluster-kube-apiserver-operator/pull/179), to be used for garbage collection as part of https://jira.coreos.com/browse/MSTR-221

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -167,7 +167,7 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 	return actual, true, err
 }
 
-func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string) (*corev1.ConfigMap, bool, error) {
+func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string, ownerRefs []metav1.OwnerReference) (*corev1.ConfigMap, bool, error) {
 	source, err := client.ConfigMaps(sourceNamespace).Get(sourceName, metav1.GetOptions{})
 	switch {
 	case apierrors.IsNotFound(err):
@@ -186,12 +186,12 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 		source.Namespace = targetNamespace
 		source.Name = targetName
 		source.ResourceVersion = ""
-		source.OwnerReferences = []metav1.OwnerReference{}
+		source.OwnerReferences = ownerRefs
 		return ApplyConfigMap(client, recorder, source)
 	}
 }
 
-func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string) (*corev1.Secret, bool, error) {
+func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string, ownerRefs []metav1.OwnerReference) (*corev1.Secret, bool, error) {
 	source, err := client.Secrets(sourceNamespace).Get(sourceName, metav1.GetOptions{})
 	switch {
 	case apierrors.IsNotFound(err):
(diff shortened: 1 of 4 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -186,12 +186,12 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 		source.Namespace = targetNamespace
 		source.Name = targetName
 		source.ResourceVersion = ""
-		source.OwnerReferences = []metav1.OwnerReference{}
+		source.OwnerReferences = ownerRefs
```
mfojtik: can you do `source.SetOwnerReferences` ? (I believe that is part of `runtime.Object`)

mfojtik:
@damemi do we have any unit test exercising the owner refs and proving they are set properly?

pkg/operator/resource/resourceapply/core.go
```diff
@@ -183,15 +183,15 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 	case err != nil:
 		return nil, false, err
 	default:
-		source.Namespace = targetNamespace
-		source.Name = targetName
-		source.ResourceVersion = ""
-		source.OwnerReferences = []metav1.OwnerReference{}
+		source.SetNamespace(targetNamespace)
+		source.SetName(targetName)
+		source.SetResourceVersion("")
```
mfojtik: drop

pkg/operator/resource/resourceapply/core.go
```diff
@@ -183,15 +183,15 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 	case err != nil:
 		return nil, false, err
 	default:
-		source.Namespace = targetNamespace
-		source.Name = targetName
-		source.ResourceVersion = ""
-		source.OwnerReferences = []metav1.OwnerReference{}
+		source.SetNamespace(targetNamespace)
```
deads2k: use straight equals unless you cannot

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -159,8 +159,22 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 }
 
 func (c RevisionController) createNewRevision(revision int32) error {
+	obj, _, err := resourceapply.SyncConfigMap(c.kubeClient.CoreV1(), c.eventRecorder, c.targetNamespace, "revision-status", c.targetNamespace, nameFor("revision-status", revision), []metav1.OwnerReference{})
```
deads2k: Create your configmap struct from scratch since it's mostly empty

deads2k: not `obj`.  Give me a real name

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -159,8 +159,22 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
...
+	if err != nil {
+		return err
+	}
+	ownerRefs := []metav1.OwnerReference{}
```
deads2k: inline the creation

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -159,8 +159,22 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
...
+	if obj != nil {
```
deads2k: this is never nil

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -159,8 +159,22 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
...
+		ownerRefs = append(ownerRefs, metav1.OwnerReference{
```
deads2k: check client-go and apimachinery for a helper for this

deads2k:
correct in concept.  Needs some spelling changes

damemi:
@deads2k I updated the revision controller tests to 1) actually validate the actions we expect and 2) exclude `Event`s from the expected actions because it increased the action count 2x+1 in this scenario. If we really want to test the number of events I think it would be better as a separate check, but I'm assuming events are less predictable than objects. This is ready for more review along with 151

PR #158, opened 2019-01-07
segregate and simplify the cert rotation config

deads2k:
per @mfojtik's suggestion, this switches us to use structs for each bit of a cert rotation logic

/assign @sttts


mfojtik:
LGTM, it looks much better now! :-)

will leave it to @sttts to tag this.

deads2k:
> LGTM, it looks much better now! :-)
> 
> will leave it to @sttts to tag this.

It's a shuffle. tagging

PR #159, opened 2019-01-07
Update revision status configmap based on node statuses

damemi:
Part of https://jira.coreos.com/browse/MSTR-221

Depends on https://github.com/openshift/library-go/pull/157 and https://github.com/openshift/cluster-kube-apiserver-operator/pull/179

Adds a function to the pruning controller that updates the most recent revision status configmap by looping through all node statuses and returning failure if there are any failed nodes, otherwise returning success

### DIFF
--- a/pkg/operator/staticpod/controller/prune/prune_controller.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller.go
@@ -131,6 +131,32 @@ func (c *PruneController) pruneRevisionHistory(operatorStatus *operatorv1.Static
 	return nil
 }
 
+func (c *PruneController) updateRevisionStatus(operatorStatus *operatorv1.StaticPodOperatorStatus) error {
+	if len(operatorStatus.NodeStatuses) == 0 {
+		return nil
+	}
+	latestRevision := operatorStatus.LatestAvailableRevision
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -227,6 +253,11 @@ func (c *PruneController) sync() error {
 		return err
 	}
 
+	err = c.updateRevisionStatus(operatorStatus)
```
mfojtik: nit: `if err := ...` (as you don't need the `err` below.

mfojtik:
/approve

One small nit, LGTM otherwise.

deads2k:
/hold

Thanks for putting this together to compare against https://github.com/openshift/library-go/pull/151/files .  I can see that we still end up edge triggered in this and are vulnerable to losing status.  After seeing it here, I can see that it is fundamental to our status.

In light of that, the creation in the revision controller makes a lot of sense (https://github.com/openshift/library-go/pull/157 almost lgtm).  I think you end up moving this function (perhaps with a few tweaks) to https://github.com/openshift/library-go/pull/151/files to be updated as an non-failing update about here: https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L233

PR #162, opened 2019-01-08
(wip) Prune API resources owned by revision status configmap

damemi:
Part of https://jira.coreos.com/browse/MSTR-221

Depends on:

1. https://github.com/openshift/cluster-kube-apiserver-operator/pull/179
2. https://github.com/openshift/library-go/pull/157
3. https://github.com/openshift/library-go/pull/159


deads2k:
Ok, I think I see where you got here.  What about doing it in the prune control loop instead.  Perhaps a queue that is separate from the pruning on disk.

PR #165, opened 2019-01-10
installer: QoS class guaranteed with 100MB

sttts:
We measured 25MB RSS. So 100MB should be safe to have some room to breath and not too big.


deads2k:
that's v1alpha1, you also need to get the bindata one I linked in the other pull (the v1 flavor)

deads2k:
we left the old for an orderly transition.  I'm ok deleting that now too.

sttts:
@deads2k removed v1alpha1staticpods and updated v1 variant

PR #173, opened 2019-01-14
add better event information and glog info on diff

deads2k:
Makes sure we always have a message about what is changing.  If we don't see an explicit change, we will serialize it to json.  If we end up with a cardinality problem on events, we can deal with it, but this is a fallthrough to find our ghost writes.

/assign @sanchezl

### DIFF
--- a/pkg/config/clusteroperator/v1helpers/status.go
+++ b/pkg/config/clusteroperator/v1helpers/status.go
@@ -1,11 +1,15 @@
 package v1
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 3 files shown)


mfojtik:
i believe this fixes this case:

`
I0115 10:27:14.661859       1 event.go:221] Event(v1.ObjectReference{Kind:"Deployment", Namespace:"openshift-kube-apiserver-operator", Name:"openshift-kube-apiserver-operator", UID:"5bd237d4-18aa-11e9-a483-1268e4ea135c", APIVersion:"apps/v1", ResourceVersion:"", FieldPath:""}): type: 'Normal' reason: 'OperatorStatusChanged' Status for operator openshift-kube-apiserver-operator changed:
`

PR #177, opened 2019-01-15
Add optional configMaps to revision controller

mrogers950:
This lets us pass optional configMaps to the controller that will not halt progress if they don't exist.
@deads2k @openshift/sig-security 
re: https://github.com/openshift/cluster-kube-apiserver-operator/pull/185

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	"github.com/openshift/library-go/pkg/operator/staticpod/controller/installer/bindata"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/revision"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
@@ -48,9 +49,9 @@ type InstallerController struct {
 	targetNamespace, staticPodName string
 	// configMaps is the list of configmaps that are directly copied.A different actor/controller modifies these.
 	// the first element should be the configmap that contains the static pod manifest
-	configMaps []string
+	configMaps []revision.RevisionResource
 	// secrets is a list of secrets that are directly copied for the current values.  A different actor/controller modifies these.
-	secrets []string
+	secrets []revision.RevisionResource
 	// command is the string to use for the installer pod command
 	command []string
 
@@ -82,8 +83,8 @@ const (
 // NewInstallerController creates a new installer controller.
 func NewInstallerController(
 	targetNamespace, staticPodName string,
-	configMaps []string,
-	secrets []string,
+	configMaps []revision.RevisionResource,
+	secrets []revision.RevisionResource,
 	command []string,
 	kubeInformersForTargetNamespace informers.SharedInformerFactory,
 	operatorConfigClient v1helpers.StaticPodOperatorClient,
@@ -478,19 +479,30 @@ func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *
 	pod.Spec.Containers[0].Image = c.installerPodImageFn()
 	pod.Spec.Containers[0].Command = c.command
 
+	if c.configMaps[0].Optional {
+		return fmt.Errorf("pod configmap %s is required, cannot be optional", c.configMaps[0].Name)
+	}
 	args := []string{
 		"-v=4", // TODO: Make this configurable?
 		fmt.Sprintf("--revision=%d", revision),
 		fmt.Sprintf("--namespace=%s", pod.Namespace),
-		fmt.Sprintf("--pod=%s", c.configMaps[0]),
+		fmt.Sprintf("--pod=%s", c.configMaps[0].Name),
 		fmt.Sprintf("--resource-dir=%s", hostResourceDirDir),
 		fmt.Sprintf("--pod-manifest-dir=%s", hostPodManifestDir),
 	}
-	for _, name := range c.configMaps {
-		args = append(args, fmt.Sprintf("--configmaps=%s", name))
+	for _, cm := range c.configMaps {
+		if cm.Optional {
+			args = append(args, fmt.Sprintf("--optional-configmaps=%s", cm.Name))
+		} else {
+			args = append(args, fmt.Sprintf("--configmaps=%s", cm.Name))
+		}
 	}
-	for _, name := range c.secrets {
-		args = append(args, fmt.Sprintf("--secrets=%s", name))
+	for _, s := range c.secrets {
+		if s.Optional {
+			args = append(args, fmt.Sprintf("--optional-secrets=%s", s.Name))
+		} else {
+			args = append(args, fmt.Sprintf("--secrets=%s", s.Name))
+		}
 	}
 	pod.Spec.Containers[0].Args = args
 
--- a/pkg/operator/staticpod/controller/revision/revision_controller.go
+++ b/pkg/operator/staticpod/controller/revision/revision_controller.go
@@ -33,9 +33,9 @@ type RevisionController struct {
 	targetNamespace string
 	// configMaps is the list of configmaps that are directly copied.A different actor/controller modifies these.
 	// the first element should be the configmap that contains the static pod manifest
-	configMaps []string
+	configMaps []RevisionResource
 	// secrets is a list of secrets that are directly copied for the current values.  A different actor/controller modifies these.
-	secrets []string
+	secrets []RevisionResource
 
 	operatorConfigClient v1helpers.StaticPodOperatorClient
 	configMapGetter      corev1client.ConfigMapsGetter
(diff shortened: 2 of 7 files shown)


pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -180,6 +184,12 @@ func (c RevisionController) createNewRevision(revision int32) error {
 		UID:        statusConfigMap.UID,
 	}}
 
+	for _, name := range c.optionalConfigMaps {
+		_, _, err := resourceapply.SyncConfigMap(c.kubeClient.CoreV1(), c.eventRecorder, c.targetNamespace, name, c.targetNamespace, nameFor(name, revision), ownerRefs)
+		if err != nil {
+			return err
```
enj: uuh, how is optional?

mrogers950: does not return err if obj == nil

enj: Ah sorry I was thinking this was `ApplyConfigMap`.

pkg/operator/staticpod/controller/revision/deployment_controller.go
```diff
@@ -52,16 +54,18 @@ type RevisionController struct {
 func NewRevisionController(
 	targetNamespace string,
 	configMaps []string,
+	optionalConfigMaps []string,
```
deads2k: no.  Make the normal configmaps and secrets become a struct with an `optional` member.

mrogers950:
Updated.

pkg/operator/staticpod/controllers.go
```diff
@@ -33,15 +33,34 @@ type staticPodOperatorControllers struct {
 // 4. BackingResourceController - this creates the backing resources needed for the operand, such as cluster rolebindings and installer service
 //    account.
 // 5. MonitoringResourceController - this creates the service monitor used by prometheus to scrape metrics.
-func NewControllers(targetNamespaceName, staticPodName string, command, revisionConfigMaps, revisionSecrets []string,
+func NewControllers(targetNamespaceName, staticPodName string, command, revisionConfigMaps, optionalRevisionConfigMaps, revisionSecrets []string,
```
deads2k: why isn't this the same revision resource?  I expect that ultimately everything that took the list of resources needs to be updated to handle the idea of an optional resource.  The compile breaks will help you.

mrogers950: fixed this.

mrogers950:
@deads2k updated.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -482,15 +483,15 @@ func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *
...
 		fmt.Sprintf("--namespace=%s", pod.Namespace),
-		fmt.Sprintf("--pod=%s", c.configMaps[0]),
+		fmt.Sprintf("--pod=%s", c.configMaps[0].Name),
 		fmt.Sprintf("--resource-dir=%s", hostResourceDirDir),
 		fmt.Sprintf("--pod-manifest-dir=%s", hostPodManifestDir),
 	}
-	for _, name := range c.configMaps {
-		args = append(args, fmt.Sprintf("--configmaps=%s", name))
+	for _, cm := range c.configMaps {
+		args = append(args, fmt.Sprintf("--configmaps=%s", cm.Name))
```
deads2k: You're losing information here.  The fact that your signature changed above suggests that the information needs to preserved all the way to the command this is running.  In this case, `installerpod/cmd.go` needs to know that some items can 404 or you'll just permafail the cluster.

deads2k: Also, add a unit test to that command so you can see it work.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -478,19 +479,26 @@ func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *
...
-		fmt.Sprintf("--pod=%s", c.configMaps[0]),
+		fmt.Sprintf("--pod=%s", c.configMaps[0].Name),
 		fmt.Sprintf("--resource-dir=%s", hostResourceDirDir),
 		fmt.Sprintf("--pod-manifest-dir=%s", hostPodManifestDir),
 	}
-	for _, name := range c.configMaps {
-		args = append(args, fmt.Sprintf("--configmaps=%s", name))
+	for _, cm := range c.configMaps {
+		if !cm.Optional {
+			args = append(args, fmt.Sprintf("--configmaps=%s", cm.Name))
```
deads2k: These are args to https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/installerpod/cmd.go#L131 which does not tolerate the idea of optional resources.  You need optional resources, so that information must be plumbed through and those which are *optional* should be tolerated as missing.

mrogers950: Ah, I see now. Will fix.

mrogers950:
Updated to allow optional configmaps/secrets in the installer pod command, as well as fixing isLatestRevisionCurrent() so it does not always create a new revision when an optional configmap/secret is missing. I added more tests as well.
@deads2k PTAL

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -129,29 +134,47 @@ func nameFor(name string, revision int32) string {
...
+
+		required, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(cm.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) && !cm.Optional {
 			return false, err.Error()
 		}
-		existing, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(nameFor(name, revision), metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
+		existing, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(nameFor(cm.Name, revision), metav1.GetOptions{})
+		if apierrors.IsNotFound(err) && !cm.Optional {
 			return false, err.Error()
```
deads2k: not new, but this should probably be `return false, nil`, right?  otherwise I can't easily add anything.

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -129,29 +134,47 @@ func nameFor(name string, revision int32) string {
...
+
+		required, err := c.secretGetter.Secrets(c.targetNamespace).Get(s.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) && !s.Optional {
 			return false, err.Error()
 		}
-		existing, err := c.secretGetter.Secrets(c.targetNamespace).Get(nameFor(name, revision), metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
+		existing, err := c.secretGetter.Secrets(c.targetNamespace).Get(nameFor(s.Name, revision), metav1.GetOptions{})
+		if apierrors.IsNotFound(err) && !s.Optional {
 			return false, err.Error()
```
deads2k: same comment

pkg/operator/staticpod/controller/revision/revision_controller.go
deads2k: > t new, but this should probably be `return false, nil`, right? otherwise I can't easily add anything.

never mind.  reason, not error. ignore me

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -19,6 +19,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/library-go/pkg/config/client"
+	"k8s.io/apimachinery/pkg/api/errors"
```
deads2k: nit order

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -133,6 +138,15 @@ func (o *InstallOptions) copyContent() error {
 		}
 		secrets = append(secrets, val)
 	}
+	for _, currPrefix := range o.OptionalSecretNamePrefixes {
+		val, err := o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(currPrefix), metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		if !errors.IsNotFound(err) {
```
deads2k: don't need this condition.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -133,6 +138,15 @@ func (o *InstallOptions) copyContent() error {
 		}
 		secrets = append(secrets, val)
 	}
+	for _, currPrefix := range o.OptionalSecretNamePrefixes {
+		val, err := o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(currPrefix), metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
```
deads2k:
```
if errors.IsNotFound{ 
continue
} 
if err != nil{
 return err
}
secrets = append(
```

deads2k:
/lgtm

given where we are, at the slightest hint of trouble, this will be reverted.  Sorry.

PR #179, opened 2019-01-16
make rotation more robust

deads2k:
First I capped validity based on min of desired and expiry of the signer.  This doesn't affect verification, but does affect the annotation.  I'm willing to split that out separately if it's easier.

Second, I updated the target regeneration to look for matching signer names in any of the certs in the ca bundle.  This avoids instant and unnecessary rotation of target certs when signers are rotated.

/assign @mrogers950 @sttts 

This needs unit test updates, but ptal and let me know about the concept

/hold


pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -75,11 +75,12 @@ func (c CertRotationController) sync() error {
...
 	}
 
-	if err := c.CABundleRotation.ensureConfigMapCABundle(signingCertKeyPair); err != nil {
+	cabundleCerts, err := c.CABundleRotation.ensureConfigMapCABundle(signingCertKeyPair)
+	if err != nil {
 		return err
 	}
 
-	if err := c.TargetRotation.ensureTargetCertKeyPair(signingCertKeyPair); err != nil {
+	if err := c.TargetRotation.ensureTargetCertKeyPair(signingCertKeyPair, cabundleCerts); err != nil {
```
sttts: good, I feared this plumbing is more complicated

sttts:
Looks good. Unit tests would be good indeed.

deads2k:
> Looks good. Unit tests would be good indeed.

modified the existing one to plumb in the correct spot.

PR #180, opened 2019-01-16
Return current service in ApplyService

enj:
Return the existing service when an update is not required.

@openshift/sig-master


deads2k:
just an oversight

/lgtm

PR #182, opened 2019-01-16
don't clear status conditions or we get update churn

deads2k:
clearing the conditions like this loses the last transition time

/assign @mfojtik @sanchezl

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -89,7 +89,6 @@ func (c StatusSyncer) sync() error {
 			ObjectMeta: metav1.ObjectMeta{Name: c.clusterOperatorName},
 		}
 	}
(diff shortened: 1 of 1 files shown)


deads2k:
green.  need to bump

pkg/operator/status/status_controller.go
```diff
@@ -89,7 +89,6 @@ func (c StatusSyncer) sync() error {
 			ObjectMeta: metav1.ObjectMeta{Name: c.clusterOperatorName},
 		}
 	}
-	clusterOperatorObj.Status.Conditions = nil
```
soltysh: I agree that clearing always was bad, but shouldn't we ensure that we clean the conditions eventually for operator? We clean properly available and progressing conditions, but I don't think we are clearing failing, shouldn't we?

PR #183, opened 2019-01-16
fix ownerref to have values, not empty

deads2k:
/assign @damemi


damemi:
/lgtm

question though shouldn't these have values filled out after the call to `ApplyConfigMap`?

PR #187, opened 2019-01-17
Reduce qps

deads2k:
Use listers to compute hashes 
/assign @juanvallejo


deads2k:
juan lgtm'd via slack

PR #192, opened 2019-01-18
rotate only after the signer has had some time to propogate

deads2k:
This makes us wait for 10% of available overlap for the signer to age before we sign.  Assuming we refresh the target a lot more often than the signer, this works.

Also adds a time based trigger every minute to check if we need to do work.

/assign @sttts

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/golang/glog"
+
 	"github.com/openshift/library-go/pkg/operator/events"
 	corev1informers "k8s.io/client-go/informers/core/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -61,7 +63,7 @@ func (c TargetRotation) ensureTargetCertKeyPair(signingCertKeyPair *crypto.CA, c
 	}
 	targetCertKeyPairSecret.Type = corev1.SecretTypeTLS
 
-	if needNewTargetCertKeyPair(targetCertKeyPairSecret.Annotations, caBundleCerts, c.Validity, c.RefreshPercentage) {
+	if needNewTargetCertKeyPair(targetCertKeyPairSecret.Annotations, signingCertKeyPair, caBundleCerts, c.Validity, c.RefreshPercentage) {
 		c.EventRecorder.Eventf("TargetUpdateRequired", "%q in %q requires a new target cert/key pair", c.Name, c.Namespace)
 		if err := setTargetCertKeyPairSecret(targetCertKeyPairSecret, c.Validity, signingCertKeyPair, c.ClientRotation, c.ServingRotation); err != nil {
 			return err
@@ -83,8 +85,8 @@ func (c TargetRotation) ensureTargetCertKeyPair(signingCertKeyPair *crypto.CA, c
 	return nil
 }
 
-func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x509.Certificate, validity time.Duration, renewalPercentage float32) bool {
-	if needNewCertKeyPairForTime(annotations, validity, renewalPercentage) {
+func needNewTargetCertKeyPair(annotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, validity time.Duration, renewalPercentage float32) bool {
+	if needNewTargetCertKeyPairForTime(annotations, signer, validity, renewalPercentage) {
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -102,6 +104,36 @@ func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x5
 	return true
 }
 
+func needNewTargetCertKeyPairForTime(annotations map[string]string, signer *crypto.CA, validity time.Duration, renewalPercentage float32) bool {
```
sttts: go doc summing up the general idea would help a lot

pkg/operator/certrotation/target.go
```diff
@@ -102,6 +104,36 @@ func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x5
...
+
+	// If Certificate is past its validity, we may must generate new.
+	if time.Now().After(certExpiry) {
+		return true
+	}
+
+	// If Certificate is past its validity*renewpercent, we may have action to take. if the signer is old enough
+	renewalDuration := -1 * float32(validity) * (1 - renewalPercentage)
+	if time.Now().After(certExpiry.Add(time.Duration(renewalDuration))) {
+		// make sure the signer is at 10% or renewal overlap old so we had time to rotate trust. Before signing new.
```
sttts: grammar

pkg/operator/certrotation/target.go
```diff
@@ -102,6 +104,36 @@ func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x5
...
+	certExpiry, err := time.Parse(time.RFC3339, targetExpiry)
+	if err != nil {
+		glog.Infof("bad expiry: %q", targetExpiry)
+		// just create a new one
+		return true
+	}
+
+	// If Certificate is past its validity, we may must generate new.
+	if time.Now().After(certExpiry) {
+		return true
```
sttts: shouldn't we start at 90% or so?

pkg/operator/certrotation/target.go
```diff
@@ -102,6 +104,36 @@ func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x5
...
+	}
+
+	// If Certificate is past its validity*renewpercent, we may have action to take. if the signer is old enough
+	renewalDuration := -1 * float32(validity) * (1 - renewalPercentage)
+	if time.Now().After(certExpiry.Add(time.Duration(renewalDuration))) {
+		// make sure the signer is at 10% or renewal overlap old so we had time to rotate trust. Before signing new.
+		timeToWaitForTrustRotation := -1 * renewalDuration / 10
+		if time.Now().After(signer.Config.Certs[0].NotBefore.Add(time.Duration(timeToWaitForTrustRotation))) {
```
sttts: this requires that the signer renewal period is > 10% of cert validity. Should we document that? Maybe even check in the constructor (or whereever it makes sense)?

pkg/operator/certrotation/target.go
deads2k: done

pkg/operator/certrotation/target.go
deads2k: > shouldn't we start at 90% or so?

caught below.  this is when it's straight out expired

pkg/operator/certrotation/target.go
```diff
@@ -83,8 +85,8 @@ func (c TargetRotation) ensureTargetCertKeyPair(signingCertKeyPair *crypto.CA, c
 	return nil
 }
 
-func needNewTargetCertKeyPair(annotations map[string]string, caBundleCerts []*x509.Certificate, validity time.Duration, renewalPercentage float32) bool {
-	if needNewCertKeyPairForTime(annotations, validity, renewalPercentage) {
+func needNewTargetCertKeyPair(annotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, validity time.Duration, renewalPercentage float32) bool {
```
sttts: in my own words, we rotate if
- our old CA is gone from the bundle (then we are pretty late to the renewal party)
- or the cert expired (then we are also pretty late)
- or we are over the renewal percentage of the validity, but only if the new CA at least 10% into its age. 

Maybe worth a go doc.

So in general we need to see a signing CA at least aged 10% within 1-percentage of the cert validity.

Hence, if the CAs are rotated too fast (like CA percentage around 10% or smaller), we will not hit the time to make use of the CA. Or if the cert renewal percentage is at 90%, there is not much time either.

So with a cert percentage of 75% and equally long CA and cert validities at the worst case we start at 85% of the cert to renew, trying again every minute.

pkg/operator/certrotation/target.go
sttts: One could do a last try before 100%, even without the 10% check for the CA. but it's not clear it makes it much better.

PR #193, opened 2019-01-21
status: report cluster operator version

mfojtik:
This is pre-requirement for operators to report status in ClusterOperator CR.

/cc @sttts 
/cc @deads2k


deads2k:
Oh, I see.  I was too late

PR #194, opened 2019-01-21
Revert "status: report cluster operator version"

mfojtik:
@deads2k @sttts guess we don't want this as the field is going away.


soltysh:
Yup, see https://github.com/openshift/api/pull/176

PR #175, opened 2019-01-15
Wire prune controller to run and sync limits from operator config

damemi:

PR #195, opened 2019-01-21
Change revision configmap from phase to status

damemi:
At some point we switched from using `phase` as the configmap value to using `status` (see https://github.com/openshift/library-go/blob/ab277dcc0384e180c6bb5e1f2cd2120143b2093c/pkg/operator/staticpod/controller/revision/revision_controller.go#L169), but some important parts never got updated. This updates them.


damemi:
disregard, rolled these changes into https://github.com/openshift/library-go/pull/175

PR #197, opened 2019-01-21
set version for clusteroperator

deads2k:
The version for clusteroperator operands are not know at compile time and need to be set asynchronously.  This is a notifying thread-safe impl to use to set these values.

I think this works, but I have not tried it.

/assign @mfojtik


deads2k:
@damemi turns out we need this or we can't bump library-go for the other :(

deads2k:
@damemi @sanchezl I've got a fake bump trying this, but if it works, any objections?

deads2k:
ok, it did no harm here https://github.com/openshift/cluster-kube-apiserver-operator/pull/215

I'm going for this to clean up bumps that fix the cert rotation status and caching problems across the other operators

PR #200, opened 2019-01-22
Fix building multiple binaries

tnozicka:
`go build` will produce a binary only with one package, otherwise it just compiles the code and no binary is produced, so we now call `go build` for each binary.

/cc @juanvallejo @soltysh @sttts @mfojtik

### DIFF
--- a/alpha-build-machinery/make/default.example.mk.help.log
+++ b/alpha-build-machinery/make/default.example.mk.help.log
@@ -1,6 +1,8 @@
 The following make targets are available:
 all
 build
+clean
+clean-binaries
 help
 image-origin-cluster-openshift-apiserver-operator
 images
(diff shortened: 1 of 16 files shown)


alpha-build-machinery/make/default.example.mk.help.log
```diff
@@ -1,6 +1,8 @@
 The following make targets are available:
 all
 build
+clean
```
tnozicka: we also go auto-target for cleaning binaries, that's why the fixtures changed

juanvallejo:
lgtm

juanvallejo:
Tried this locally in `openshift/must-gather` on a branch with https://github.com/openshift/must-gather/pull/29. Can confirm this fix works as expected

PR #201, opened 2019-01-22
set transition time on conditions properly

bparees:


pkg/operator/v1helpers/helpers.go
```diff
@@ -58,7 +58,7 @@ func SetOperatorCondition(conditions *[]operatorv1.OperatorCondition, newConditi
 
 	if existingCondition.Status != newCondition.Status {
 		existingCondition.Status = newCondition.Status
-		existingCondition.LastTransitionTime = newCondition.LastTransitionTime
+		existingCondition.LastTransitionTime = metav1.NewTime(time.Now())
```
bparees: @deads2k this was resulting in a nil transition time when updating an existing condition.

bparees:
/assign @deads2k @mfojtik @sttts 

any one of you can probably merge this

mfojtik:
@deads2k i remember we had a problem when we kept updating conditions on every sync because the timestamp changed. I saw today that we have `null` timestamps today:

```
               "conditions": [
                    {
                        "lastTransitionTime": null,
                        "status": "False",
                        "type": "Failing"
                    },
                    {
                        "lastTransitionTime": null,
                        "message": "2 of 3 nodes are at revision 3",
                        "status": "True",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": null,
                        "message": "2 of 3 nodes are at revision 3, 1 are not",
                        "status": "True",
                        "type": "Progressing"
                    }
                ],
```

Means I have to inspect events to know when we become available ;-)

PR #205, opened 2019-01-23
Add mutex and trace profiles to serviceability, handle unique

smarterclayton:
The current behavior of writing the file to the current directory is convenient but doesn't allow parallel profiles to be gathered for things like `openshift-tests` which uses child processes, or
short lived commands that you want to profile in a loop.

Callers may now append `-tmp` to the mode and profiling will write to the temp directory which will create a `profile-RANDOM/mode.pprof` file that is unique across runs.

### DIFF
--- a/pkg/serviceability/serviceability.go
+++ b/pkg/serviceability/serviceability.go
@@ -3,6 +3,7 @@ package serviceability
 import (
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 
 	"github.com/pkg/profile"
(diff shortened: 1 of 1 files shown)


pkg/serviceability/serviceability.go
```diff
@@ -18,15 +19,26 @@ type stopper struct{}
 func (stopper) Stop() {}
 
 // Profile returns an interface to defer for a profile: `defer serviceability.Profile(os.Getenv("OPENSHIFT_PROFILE")).Stop()` is common.
+// Prefixing the mode with `-tmp` will have the profiler write the run to a temporary directory with a unique name, which
```
soltysh: suffixing

smarterclayton: Damnit

PR #207, opened 2019-01-24
set failing status for failing installer pod

deads2k:
When the installer pods fail, we are failing and our status should reflect that.  This makes a new condition (so we stay single writer) for `NodeInstallerFailing` and sets it appropriately with a summary.

@eparis your "kubelet never starts" clusters will now fail in the CVO quickly.

/assign @mfojtik @damemi 

@damemi please bump our repos to fix this status.


damemi:
/hold
until rebase lands

Issue #208, opened 2019-01-28
Static pod installer pods are being OOM killed


mfojtik:
On pretty poor machine, in libvirt I saw two out of four installer pods OutOfMemory. These pods seems to be never retried.
We already set resource requests/limits (100mb), but I guess when the master node is really under fire, we can still get scheduled and fail. 

I think we should set the` podPriorityClass` for the installer pods as minimum, but we should also discuss additional ways how to make sure the installer pod can be retried. (can we re-create the pod under some circumstances? should we automatically start a new revision? etc)

PR #209, opened 2019-01-28
staticpod: set priority class and tolerations for installer pod

mfojtik:
This PR sets the podPriorityClass to for the static pod installer pods as these pods should be scheduled first in all cases. AFAIK this does not guarantee the memory for them.

xref: https://github.com/openshift/library-go/issues/208

### DIFF
--- a/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml
+++ b/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml
@@ -27,6 +27,9 @@ spec:
         limits:
           memory: 100M
   restartPolicy: Never
+  priorityClassName: system-node-critical
+  tolerations:
+    - operator: "Exists"
   securityContext:
     runAsUser: 0
   volumes:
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml
```diff
@@ -27,6 +27,9 @@ spec:
         limits:
           memory: 100M
   restartPolicy: Never
+  priorityClassName: system-node-critical
+  tolerations:
+    - operator: "Exists"
```
sttts: is this correct? Does it apply to any taint?

mfojtik: @sttts this is the toleration our operators are using (copied from kube-apiserver-operator).

sttts: Didn't we have many more tolerations there?

mfojtik: https://github.com/openshift/cluster-kube-apiserver-operator/commit/e7af9bd097321b1f70368568000e8e44b43adb30#diff-08a6124d2d9be934074b6601cf61dac5

I thought so as well, but no, we only added this one.

PR #213, opened 2019-01-29
wait to trigger new revision until all nodes finish current

damemi:
we were seeing cases were 3 nodes would be at 3 different revisions for a staticpod operator, so this backs off on creating a new revision if not all nodes are currently at the same revision

### DIFF
--- a/pkg/operator/staticpod/controller/revision/revision_controller.go
+++ b/pkg/operator/staticpod/controller/revision/revision_controller.go
@@ -2,6 +2,8 @@ package revision
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


mfojtik:
hrm, I talked to @deads2k the other day and he suggested that this can be okey-ish behavior.

deads2k:
/hold

rolling an update always results in skew.  I don't see a reason why skewing forward is any worse than skewing backward.

smarterclayton:
If you don't think there is ever a case of the following:

4.0.1 <- broken on upgrade
4.0.2 <- works on upgrade
4.1.0 <- needs 4.0.2 to upgrade safely

happening because of this, I'm less worried.

smarterclayton:
I will say it doubles the interactions of things you have to debug - it's not just version 1 and 2, it's version 1 with version 3 and version 2 with version 3.

PR #214, opened 2019-01-30
add pruner pod priorityclass and toleration

damemi:
copied from https://github.com/openshift/library-go/pull/209


pkg/operator/staticpod/controller/prune/manifests/pruner-pod.yaml
```diff
@@ -22,6 +22,9 @@ spec:
     - mountPath: /etc/kubernetes/
       name: kubelet-dir
   restartPolicy: Never
+  priorityClassName: system-node-critical
+    tolerations:
+      - operator: "Exists"
```
deads2k: looks indented wrong, please test

damemi: @deads2k thanks for catching, updated

PR #215, opened 2019-01-30
add pruner pod ownerrefs for revision

damemi:
sets ownerrefs for pruning pods for revision pruning (and refactors previously merged installerpod ownerrefs, see https://github.com/openshift/library-go/pull/211)

### DIFF
--- a/pkg/operator/staticpod/controller/prune/prune_controller.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller.go
@@ -39,6 +39,8 @@ type PruneController struct {
 
 	// prunerPodImageFn returns the image name for the pruning pod
 	prunerPodImageFn func() string
+	// ownerRefsFn sets the ownerrefs on the pruner pod
+	ownerRefsFn func(revision int32) ([]metav1.OwnerReference, error)
 
 	operatorConfigClient v1helpers.StaticPodOperatorClient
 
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -223,10 +232,30 @@ func (c *PruneController) ensurePrunePod(nodeName string, maxEligibleRevision in
...
+	}
+	pod.OwnerReferences = ownerRefs
+
+	_, _, err = resourceapply.ApplyPod(c.podGetter, c.eventRecorder, pod)
 	return err
 }
 
+func (c *PruneController) setOwnerRefs(revision int32) ([]metav1.OwnerReference, error) {
+	ownerReferences := []metav1.OwnerReference{}
+	statusConfigMap, err := c.kubeClient.CoreV1().ConfigMaps(c.targetNamespace).Get(fmt.Sprintf("revision-status-%d", revision), metav1.GetOptions{})
```
deads2k: use the configmap getter

pkg/operator/staticpod/controllers.go
```diff
@@ -73,6 +73,7 @@ func NewControllers(targetNamespaceName, staticPodName, podResourcePrefix string
 		secretGetter,
 		podsGetter,
 		staticPodOperatorClient,
+		kubeClient,
```
deads2k: this is no longer needed, rigth?

damemi: right

PR #216, opened 2019-02-01
add signer target for rotation

deads2k:
to rotate CSR signers we need to produce signing certs as a target.  We want it as a target so that we can rotate the overall trust slower than the individual signers themselves. This gives confidence that rotation works without forcing all consumers to rotate at the same time.

/assign @sttts

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -29,8 +29,11 @@ type TargetRotation struct {
 	Validity          time.Duration
 	RefreshPercentage float32
 
+	// Only one of client, serving, or signer rotation may be specified.
+	// TODO refactor with an interface for actually signing and move the one-of check higher in the stack.
 	ClientRotation  *ClientRotation
 	ServingRotation *ServingRotation
+	SignerRotation  *SignerRotation
 
 	Informer      corev1informers.SecretInformer
 	Lister        corev1listers.SecretLister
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -47,6 +48,10 @@ type ServingRotation struct {
 	CertificateExtensionFn []crypto.CertificateExtensionFunc
 }
 
+type SignerRotation struct {
```
sttts: godocs

pkg/operator/certrotation/target.go
```diff
@@ -154,10 +159,21 @@ func needNewTargetCertKeyPairForTime(annotations map[string]string, signer *cryp
 }
 
 // setTargetCertKeyPairSecret creates a new cert/key pair and sets them in the secret
-func setTargetCertKeyPairSecret(targetCertKeyPairSecret *corev1.Secret, validity time.Duration, signer *crypto.CA, clientRotation *ClientRotation, servingRotation *ServingRotation) error {
-	if (servingRotation != nil) == (clientRotation != nil) {
-		return fmt.Errorf("must be one of server or client cert")
+func setTargetCertKeyPairSecret(targetCertKeyPairSecret *corev1.Secret, validity time.Duration, signer *crypto.CA, clientRotation *ClientRotation, servingRotation *ServingRotation, signerRotation *SignerRotation) error {
```
sttts: godocs for the one-out-of-three non-nil requirements. Why is that?

pkg/operator/certrotation/target.go
```diff
@@ -31,6 +31,7 @@ type TargetRotation struct {
 
 	ClientRotation  *ClientRotation
 	ServingRotation *ServingRotation
+	SignerRotation  *SignerRotation
```
sttts: godoc for all three

sttts: You are building a sum-type here. One could put the actual signer call into a method of each and implement and interface here, hence only have one rotation object here.

deads2k:
documented and I agree with the assertion.  I know the auth team is has been working on a more comprehensive replacement

PR #217, opened 2019-02-04
installer: add test to exercise revision skew

mfojtik:
This add two more tests to installer to exercise the behavior when nodes are on different revision levels.


sttts:
/lgtm
/approve

This is not enough (it only tests incomplete logic), but the right direction.

PR #218, opened 2019-02-04
Fix diffing symlinks for `make verify-deps`

tnozicka:
/cc @sttts @deads2k @openshift/sig-master


alpha-build-machinery/make/targets/openshift/deps.mk
```diff
@@ -2,7 +2,7 @@ self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
 scripts_dir :=$(self_dir)/../../../scripts
 
 # We need to force localle so different envs sort files the same way for recursive traversals
-DIFF :=LC_COLLATE=C diff
+deps_diff :=LC_COLLATE=C diff --no-dereference -N
```
tnozicka: the `--no-dereference` makes the difference

sttts: links in repos are always fun

PR #219, opened 2019-02-04
Fix updating deps for mercurial

tnozicka:
/cc @sttts


alpha-build-machinery/scripts/update-deps.sh
```diff
@@ -21,3 +21,7 @@ function verify_glide_version() {
 verify_glide_version
 
 glide update --strip-vendor
+
+# glide doesn't handle mercurial properly and leaves internal files (equivalent of .git/) laying around
+# Given those files differ by mercurial version it was cloned with, verify-deps would break
```
sttts: dejavu, but don't know where from.

PR #220, opened 2019-02-04
assets: add creater based on dynamic client

mfojtik:
This PR add `pkg/assets/create` package that contain helper to create multiple resources from disk and wait for them to be created.

The rationale here is to move this create and wait logic away from [cluster-bootstrap](https://github.com/openshift/cluster-bootstrap/pull/7) (former bootkube) and make it available for multiple components as needed.

### DIFF
--- a/pkg/assets/create/create_test.go
+++ b/pkg/assets/create/create_test.go
@@ -1,0 +1,269 @@
+package create
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	ktesting "k8s.io/client-go/testing"
+
+	"github.com/openshift/library-go/pkg/assets"
+)
+
+func init() {
+	fetchLatestDiscoveryInfoFn = func(dc *discovery.DiscoveryClient) (meta.RESTMapper, error) {
+		resourcesForEnsureMutex.Lock()
+		defer resourcesForEnsureMutex.Unlock()
+		return restmapper.NewDiscoveryRESTMapper(resourcesForEnsure), nil
+	}
+	newClientsFn = func(config *rest.Config) (dynamic.Interface, *discovery.DiscoveryClient, error) {
+		fakeScheme := runtime.NewScheme()
+		// TODO: This is a workaround for dynamic fake client bug where the List kind is enforced and duplicated in object reactor.
+		fakeScheme.AddKnownTypeWithName(schema.GroupVersionKind{Version: "v1", Kind: "ListList"}, &unstructured.UnstructuredList{})
+		dynamicClient := dynamicfake.NewSimpleDynamicClient(fakeScheme)
+		return dynamicClient, nil, nil
+	}
+}
+
+var (
+	resources = []*restmapper.APIGroupResources{
+		{
+			Group: metav1.APIGroup{
+				Name: "kubeapiserver.operator.openshift.io",
+				Versions: []metav1.GroupVersionForDiscovery{
+					{Version: "v1alpha1"},
+				},
+				PreferredVersion: metav1.GroupVersionForDiscovery{Version: "v1alpha1"},
+			},
+			VersionedResources: map[string][]metav1.APIResource{
+				"v1alpha1": {
+					{Name: "kubeapiserveroperatorconfigs", Namespaced: false, Kind: "KubeAPIServerOperatorConfig"},
+				},
+			},
+		},
+		{
+			Group: metav1.APIGroup{
+				Name: "apiextensions.k8s.io",
+				Versions: []metav1.GroupVersionForDiscovery{
+					{Version: "v1beta1"},
+				},
+				PreferredVersion: metav1.GroupVersionForDiscovery{Version: "v1beta1"},
+			},
+			VersionedResources: map[string][]metav1.APIResource{
+				"v1beta1": {
+					{Name: "customresourcedefinitions", Namespaced: false, Kind: "CustomResourceDefinition"},
+				},
+			},
+		},
+		{
+			Group: metav1.APIGroup{
+				Name: "",
+				Versions: []metav1.GroupVersionForDiscovery{
+					{Version: "v1"},
+				},
+				PreferredVersion: metav1.GroupVersionForDiscovery{Version: "v1"},
+			},
+			VersionedResources: map[string][]metav1.APIResource{
+				"v1": {
+					{Name: "namespaces", Namespaced: false, Kind: "Namespace"},
+					{Name: "configmaps", Namespaced: true, Kind: "ConfigMap"},
+					{Name: "secrets", Namespaced: true, Kind: "Secret"},
+				},
+			},
+		},
+	}
+
+	// Copy this to not overlap with other tests if ran in parallel
+	resourcesForEnsure      = resources
+	resourcesForEnsureMutex sync.Mutex
+)
+
+func TestEnsureManifestsCreated(t *testing.T) {
+	// Success
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	err := EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Missing discovery info for kubeapiserverconfig
+	out := &bytes.Buffer{}
+	operatorResource := resourcesForEnsure[0]
+	resourcesForEnsure = resourcesForEnsure[1:]
+	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true, StdErr: out})
+	if err == nil {
+		t.Fatal("expected error creating kubeapiserverconfig resource, got none")
+	}
+	if !strings.Contains(out.String(), "unable to get REST mapping") {
+		t.Fatalf("expected error logged to output when verbose is on, got: %s\n", out.String())
+	}
+
+	// Should succeed on updated discovery info
+	go func() {
+		time.Sleep(2 * time.Second)
+		resourcesForEnsureMutex.Lock()
+		defer resourcesForEnsureMutex.Unlock()
+		resourcesForEnsure = append(resourcesForEnsure, operatorResource)
+	}()
+	out = &bytes.Buffer{}
+	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true, StdErr: out})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !strings.Contains(out.String(), `no matches for kind "KubeAPIServerOperatorConfig"`) {
+		t.Fatalf("expected error logged to output when verbose is on, got: %s\n", out.String())
+	}
+	if !strings.Contains(out.String(), `Creating kubeapiserver.operator.openshift.io/v1alpha1`) {
+		t.Fatalf("expected success logged to output when verbose is on, got: %s\n", out.String())
+	}
+}
+
+func TestCreate(t *testing.T) {
+	resourcesWithoutKubeAPIServer := resources[1:]
+	testConfigMap := &unstructured.Unstructured{}
+	testConfigMap.SetGroupVersionKind(schema.GroupVersionKind{
+		Version: "v1",
+		Kind:    "ConfigMap",
+	})
+	testConfigMap.SetName("aggregator-client-ca")
+	testConfigMap.SetNamespace("openshift-kube-apiserver")
+
+	tests := []struct {
+		name              string
+		discovery         []*restmapper.APIGroupResources
+		expectError       bool
+		expectFailedCount int
+		expectReload      bool
+		existingObjects   []runtime.Object
+		evalActions       func(*testing.T, []ktesting.Action)
+	}{
+		{
+			name:      "create all resources",
+			discovery: resources,
+		},
+		{
+			name:              "fail to create kube apiserver operator config",
+			discovery:         resourcesWithoutKubeAPIServer,
+			expectFailedCount: 1,
+			expectError:       true,
+			expectReload:      true,
+		},
+		{
+			name:            "create all resources",
+			discovery:       resources,
+			existingObjects: []runtime.Object{testConfigMap},
+		},
+	}
+
+	fakeScheme := runtime.NewScheme()
+	// TODO: This is a workaround for dynamic fake client bug where the List kind is enforced and duplicated in object reactor.
+	fakeScheme.AddKnownTypeWithName(schema.GroupVersionKind{Version: "v1", Kind: "ListList"}, &unstructured.UnstructuredList{})
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			manifests, err := load("testdata", CreateOptions{})
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			dynamicClient := dynamicfake.NewSimpleDynamicClient(fakeScheme, tc.existingObjects...)
+			restMapper := restmapper.NewDiscoveryRESTMapper(tc.discovery)
+
+			err, reload := create(manifests, dynamicClient, restMapper, CreateOptions{Verbose: true, StdErr: os.Stderr})
+			if tc.expectError && err == nil {
+				t.Errorf("expected error, got no error")
+				return
+			}
+			if !tc.expectError && err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+			if tc.expectReload && !reload {
+				t.Errorf("expected reload, got none")
+				return
+			}
+			if !tc.expectReload && reload {
+				t.Errorf("unexpected reload, got one")
+				return
+			}
+			if len(manifests) != tc.expectFailedCount {
+				t.Errorf("expected %d failed manifests, got %d", tc.expectFailedCount, len(manifests))
+				return
+			}
+			if tc.evalActions != nil {
+				tc.evalActions(t, dynamicClient.Actions())
+			}
+		})
+
+	}
+}
+
+func TestLoad(t *testing.T) {
+	tests := []struct {
+		name                  string
+		options               CreateOptions
+		assetDir              string
+		expectedManifestCount int
+		expectError           bool
+	}{
+		{
+			name:                  "read all manifests",
+			assetDir:              "testdata",
+			expectedManifestCount: 5,
+		},
+		{
+			name:        "handle missing dir",
+			assetDir:    "foo",
+			expectError: true,
+		},
+		{
+			name: "read only 00_ prefixed files",
+			options: CreateOptions{
+				Filters: []assets.FileInfoPredicate{
+					func(info os.FileInfo) bool {
+						return strings.HasPrefix(info.Name(), "00")
+					},
+				},
+			},
+			assetDir:              "testdata",
+			expectedManifestCount: 2,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			result, err := load(tc.assetDir, tc.options)
+			if tc.expectError && err == nil {
+				t.Errorf("expected error, got no error")
+				return
+			}
+			if !tc.expectError && err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+			if len(result) != tc.expectedManifestCount {
+				t.Errorf("expected %d manifests loaded, got %d", tc.expectedManifestCount, len(result))
+				return
+			}
+		})
+	}
+}
--- a/pkg/assets/create/creater.go
+++ b/pkg/assets/create/creater.go
@@ -39,7 +39,8 @@ type CreateOptions struct {
 }
 
 // EnsureManifestsCreated ensures that all resource manifests from the specified directory are created.
-// This function will keep retrying creation until no errors are reported.
+// This function will try to create remaining resources in the manifest list after error is occurred.
+// This function will keep retrying creation until no errors are reported or the timeout is hit.
 // Pass the context to indicate how much time you are willing to wait until all resources are created.
 func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig *rest.Config, options CreateOptions) error {
 	client, dc, err := newClientsFn(restConfig)
@@ -73,7 +74,10 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig 
 		if needDiscoveryRefresh {
 			mapper, err = fetchLatestDiscoveryInfoFn(dc)
 			if err != nil {
-				return false, err
+				if options.Verbose {
+					fmt.Fprintf(options.StdErr, "[#%d] failed to fetch discovery: %s\n", retryCount, err)
+				}
+				return false, nil
 			}
 		}
 		lastCreateError, needDiscoveryRefresh = create(manifests, client, mapper, options)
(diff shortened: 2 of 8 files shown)


mfojtik:
TODO: need unit test

/hold

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,181 @@
...
+	errors := map[string][]error{}
+
+	for _, path := range sortedManifestPaths {
+		gvk := manifests[path].GetObjectKind().GroupVersionKind()
+		mappings, err := options.mapper.RESTMapping(schema.GroupKind{
+			Group: gvk.Group,
+			Kind:  gvk.Kind,
+		}, gvk.Version)
+		if err != nil {
+			errors[path] = append(errors[path], fmt.Errorf("unable to get REST mapping: %v", err))
```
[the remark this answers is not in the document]
sttts: continue ?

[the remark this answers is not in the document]
mfojtik: i wanted to give the errors context as they will be logging from different place so you can track them down :-)

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+	// the chances for the create loop to succeed faster.
+	var (
+		sortedManifestPaths []string
+	)
+	for key := range manifests {
+		sortedManifestPaths = append(sortedManifestPaths, key)
+	}
+	sort.Strings(sortedManifestPaths)
+
+	errors := map[string][]error{}
```
sttts: errs

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/ghodss/yaml"
+	"k8s.io/apimachinery/pkg/util/wait"
```
sttts: order

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/openshift/library-go/pkg/assets"
+)
+
+type CreateOptions struct {
+	// Filters allows to filter which files we will read from disk.
+	Filters []assets.FileInfoPredicate
+
+	// Context specifies time to terminate creation and report last error
+	Context context.Context
```
sttts: context should be the first argument to  `EnsureManifestsCreated` by convention.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+
+	client dynamic.Interface
```
sttts: why are some lower and some upper case?

mfojtik: didn't want to expose them as they are set in ensure

mfojtik: without context, this will be just place to specify the filters

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+// of observed errors are returned to caller.
+func EnsureManifestsCreated(manifestDir string, restConfig *rest.Config, options CreateOptions) error {
+	var err error
+
+	options.client, err = dynamic.NewForConfig(restConfig)
+	if err != nil {
+		return err
+	}
+
+	dc := discovery.NewDiscoveryClientForConfigOrDie(restConfig)
```
sttts: no orDor, return an error

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,177 @@
...
+	gr, err := restmapper.GetAPIGroupResources(dc)
+	if err != nil {
+		return err
+	}
+	options.mapper = restmapper.NewDiscoveryRESTMapper(gr)
```
sttts: how ugly to use that as state. Just pass the mapper down.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		// Creation succeed lets remove the manifest from the list to avoid creating it second time
+		delete(manifests, path)
+	}
+
+	return reportErrors(errs)
+}
+
+func reportErrors(errors map[string][]error) error {
```
sttts: report? Didn't we have an aggregated error formatter in library-go already?

pkg/assets/create/creater.go
sttts: see comment below. Using this as state is ugly.

sttts: either pass them down manually or wrap this.

pkg/assets/create/creater.go
mfojtik: this is special, it use paths

pkg/assets/create/creater.go
sttts: Not that this mapper does not update automatically in case of the CRDs.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,179 @@
...
+	Filters []assets.FileInfoPredicate
+
+	client dynamic.Interface
+	mapper meta.RESTMapper
+}
+
+// EnsureManifestsCreated ensures that all resource manifests from the specified directory are created.
+// This function will keep retrying creation until no errors are reported. After specified timeout, the last set
+// of observed errors are returned to caller.
+func EnsureManifestsCreated(manifestDir string, restConfig *rest.Config, ctx context.Context, options CreateOptions) error {
```
sttts: ctx should be first

sttts: "Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:"

pkg/assets/create/creater.go
sttts: ok, then it's fine

sttts: maybe formatError is better. This does not report anything.

pkg/assets/create/creater.go
mfojtik: yeah, i removed client and context from the struct, cleaned up the code.

pkg/assets/create/creater.go
mfojtik: @sttts @soltysh added logic to handle this case. If I get rest mapping error (no mapping), i will reload the discovery to get it updated...

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+
+	"github.com/openshift/library-go/pkg/assets"
+)
+
+type CreateOptions struct {
+	// Filters allows to filter which files we will read from disk.
```
wking: Document this as a logical or?

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+
+		if mappings.Scope.Name() == meta.RESTScopeNameRoot {
+			_, err = client.Resource(mappings.Resource).Create(manifests[path], metav1.CreateOptions{})
+		} else {
+			_, err = client.Resource(mappings.Resource).Namespace(manifests[path].GetNamespace()).Create(manifests[path], metav1.CreateOptions{})
+		}
+
+		// Resource already exists means we already succeeded
+		// This should never happen as we remove already created items from the manifest list, unless the resource existed beforehand.
+		if kerrors.IsAlreadyExists(err) {
```
wking: Is this an exact match, or just a type(/namespace)/name match?  If the latter, should an inexact match be an error?  Seems like we'd only want to claim success on exact matches.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// In case we fail to find a rest-mapping for the resource, force to fetch the updated discovery on next run.
+	reloadDiscovery := false
+
+	for _, path := range sortedManifestPaths {
+		gvk := manifests[path].GetObjectKind().GroupVersionKind()
+		mappings, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		if err != nil {
+			errs[path] = append(errs[path], fmt.Errorf("unable to get REST mapping: %v", err))
+			reloadDiscovery = true
```
wking: And `continue`?

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		// Creation succeeded lets remove the manifest from the list to avoid creating it second time
+		delete(manifests, path)
+	}
+
+	return formatErrors(errs), reloadDiscovery
+}
+
+func formatErrors(errors map[string][]error) error {
```
wking: Use [go-multierror][1] instead of rolling our own?

[1]: https://godoc.org/github.com/hashicorp/go-multierror

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	mapper, err := getRESTMapper()
+	if err != nil {
+		return err
+	}
+
+	// Retry creation until no errors are returned or the timeout is hit.
+	var lastCreateError error
+	err = wait.PollImmediateUntil(500*time.Millisecond, func() (bool, error) {
```
wking: Half a second is not a big deal, but if we are worried about flooding the API server it seems like we'd want to sleep between API requests, and not just between `manifests` iterations.  That would be dropping `PollImmediateUntil` with a raw loop:

```go
while len(manifests) > 0 {
  err, refresh := create(ctx, manifests, client, mapper, options)
  ...
}
```

possibly with a sleep inside `create`'s loop.  If you have any manifests at all, it's hard to see the iteration over those manifests completing in less than half a second.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// Return the last observed set of errors from the create process instead of timeout error.
+	if lastCreateError != nil {
+		return lastCreateError
+	}
+
+	return err
+}
+
+func create(manifests map[string]*unstructured.Unstructured, client dynamic.Interface, mapper meta.RESTMapper, options CreateOptions) (error, bool) {
```
wking: This should take a `Context` argument so you can cancel mid-`manifests`.  When you have a large map, callers may not want to wait until the end of `manifests` before we check their context for cancellation.

pkg/assets/create/creater.go
mfojtik: This matches the return code from API server that indicates the resource you're trying to create already exists, iow. there is zero outcome on re-creating the resource.

pkg/assets/create/creater.go
mfojtik: missing continue here is a logical bug, fixed

pkg/assets/create/creater.go
mfojtik: I would be concerned with flooding API server, it should handle creating few resources just fine (right @sttts ?)

I would like to keep `wait.PollImmediateUntil` because it support context and timeouts. If we wait after each create call, calculating timeout will be more complex (that we need).

pkg/assets/create/creater.go
mfojtik: A larger map to me is 1000+ keys (i don't think we will ever have that case ;-) I think we want to cancel when we finish processing entire map (all manifests) and hit the timeout. I think in that case, we will return error anyway so the outcome will be error.

pkg/assets/create/creater.go
mfojtik: sure

pkg/assets/create/creater.go
mfojtik: i re-use formatErrors in "load"

pkg/assets/create/creater.go
mfojtik: i will have to format each error with prefix when appending it and we will introduce new thing to vendor, is it worth? :-)

pkg/assets/create/creater.go
wking: > A larger map to me is 1000+ keys (i don't think we will ever have that case ;-)...

How fast do you expect a single `Resource(...)...Create(...)` to be?  1000+ keys each going through that seems like a big chunk of work to not allow a cancel before you finish all the keys.

mfojtik: The create will be hundred of milliseconds and I don't expect us to ever create 1000+ files trough this. The poller in `PollImmediateUntil` spawns a go routine and terminate it when the timeout is reached anyway, so I don't think we want to add some complex logic here and wrap every create call with its own timeout or portion the timeout for each create call?

pkg/assets/create/creater.go
wking: > I would like to keep wait.PollImmediateUntil because it support context and timeouts. If we wait after each create call, calculating timeout will be more complex (that we need).

I've pushed git://github.com/wking/openshift-library-go.git mid-create-cancel 8e509de6f1de4b68b59c1f5097b9a61bbcd8a53f with a stab at dropping `PollImmediateUntil` and moving the wait to per-manifest instead of per-`create` call.  If you're comfortable without a per-manifest rate-limit, I'd recommend just dropping the ticker and looping over manifests as fast as we get responses from the server.  Thoughts?

pkg/assets/create/creater.go
wking: I've pushed https://github.com/openshift/library-go/compare/master...wking:format-errors-drop-slice a4bb2915a9, dropping the errors slice, to address the "never have more than one error per path" portion of this thread.

wking: > I've pushed https://github.com/wking/openshift-library-go/tree/format-errors-drop-slice a4bb291...

Oops, missed one line here.  Fixed with a4bb291 -> c009e95.

pkg/assets/create/creater.go
wking: > I've pushed git://github.com/wking/openshift-library-go.git mid-create-cancel 8e509de...

Rebased onto your 0069198 with 8e509de -> fd93f4f.

pkg/assets/create/creater.go
wking: > The create will be hundred of milliseconds and I don't expect us to ever create 1000+ files trough this.

100 milliseconds seems overly optimistic to me.  What if the user is calling this from some location far from their cluster?  And I think mid-`create` cancels [are fairly straightforward][1], so I don't see a need to make assumptions about input manifest counts.

[1]: https://github.com/openshift/library-go/pull/220#discussion_r253639228

pkg/assets/create/creater.go
mfojtik: thanks! fixed up to the PR.

pkg/assets/create/creater.go
mfojtik: @wking the client itself can be configured to do QPS and rate limiting:

https://github.com/openshift/library-go/blob/e0f627346ad39aa32ada2ed1f17d268ed4b0c3f2/vendor/k8s.io/client-go/rest/config.go#L106

If we are concerned about DOSing the API server, we can set this up in the `*rest.Config` the caller pass to this helper? I can drop the ticker to 10ms or so just to avoid hot-loop, but each individual create call can be limited by using these parameters IMHO.

@sttts @deads2k agree?

pkg/assets/create/creater.go
mfojtik: see https://github.com/openshift/library-go/pull/220#discussion_r253787937

pkg/assets/create/creater.go
sttts: Does the kubeconfig have a QPS setting as well? I don't think this belongs hardcoded here, but should be settable from the outside. If the kubeconfig knows qps, there is nothing to do.

soltysh: I don't think kubeconfig has that option, we always default qps to 5 in https://github.com/kubernetes/kubernetes/blob/4992402fa5bee9f7ac6fe931d101d362a4295cea/staging/src/k8s.io/client-go/rest/config.go#L276-L278

pkg/assets/create/create_test.go
```diff
@@ -0,0 +1,269 @@
...
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	ktesting "k8s.io/client-go/testing"
+
+	"github.com/openshift/library-go/pkg/assets"
+)
+
+func init() {
+	fetchLatestDiscoveryInfoFn = func(dc *discovery.DiscoveryClient) (meta.RESTMapper, error) {
```
sttts: why not a var?

mfojtik: it is a var?

mfojtik: (also I need mutex to prevent data race when testing discovery refresh in-flight)

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Retry creation until no errors are returned or the timeout is hit.
+	var (
+		lastCreateError      error
+		retryCount           int
+		mapper               meta.RESTMapper
+		needDiscoveryRefresh bool = true
+	)
+	err = wait.PollImmediateUntil(interval, func() (bool, error) {
+		retryCount += 1
```
sttts: ++

mfojtik: sure

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+		return nil, nil, err
+	}
+
+	return client, dc, nil
+}
+
+// allow to override in unit test
+var fetchLatestDiscoveryInfoFn = fetchLatestDiscoveryInfo
+
+func fetchLatestDiscoveryInfo(dc *discovery.DiscoveryClient) (meta.RESTMapper, error) {
```
sttts: there is `CachedDiscoveryInterface` with an `Invalidate()` func, created via `NewMemCacheClient`.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return err
+	}
+
+	if options.Verbose && options.StdErr == nil {
+		options.StdErr = os.Stderr
+	}
+
+	// Default QPS in client (when not specified) is 5 requests/per second
+	// This specify the interval between "create-all-resources", no need to make this configurable.
```
sttts: specifies

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+		if needDiscoveryRefresh {
+			mapper, err = fetchLatestDiscoveryInfoFn(dc)
+			if err != nil {
+				return false, err
+			}
+		}
+		err, needDiscoveryRefresh = create(manifests, client, mapper, options)
+		if err == nil {
+			// No errors means all resources were created successfully, reset lastCreateError and exit the loop.
+			lastCreateError = nil
```
sttts: why not assign directly, `lastCreateError, needDiscoveryRefresh = create(...)` ?

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+		return lastCreateError
+	}
+
+	return err
+}
+
+// allow to override in unit test
+var getClientsFn = getClients
+
+func getClients(config *rest.Config) (dynamic.Interface, *discovery.DiscoveryClient, error) {
```
sttts: this is not a getter, but a constructor, i.e. `newClients`.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			errs[path] = fmt.Errorf("failed to create: %v", err)
+			continue
+		}
+
+		// Creation succeeded lets remove the manifest from the list to avoid creating it second time
+		delete(manifests, path)
+	}
+
+	return formatErrors("some manifests failed to create", errs), reloadDiscovery
```
sttts: failed to create some manifests

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,232 @@
...
+
+	for _, k := range keys {
+		aggregatedErrMessages = append(aggregatedErrMessages, fmt.Sprintf("%q: %v", k, errors[k]))
+	}
+	return fmt.Errorf("%s:\n%s", prefix, strings.Join(aggregatedErrMessages, "\n"))
+}
+
+func load(assetsDir string, options CreateOptions) (map[string]*unstructured.Unstructured, error) {
+	if _, err := os.Stat(assetsDir); os.IsNotExist(err) {
+		return nil, fmt.Errorf("directory %q does not exists", assetsDir)
```
sttts: what about different errors?

mfojtik: wanted to fail fast when the dir does not exists (i think file walk will probably explode with its own error?)

mfojtik: iow. i think i can remove this

pkg/assets/create/creater.go
mfojtik: settled via IRC, added TODO as this does not have to be elegant on first iteration. Also CachedDiscoveryInterface might complicate the unit test and it might not be worth the benefits.

pkg/assets/create/create_test.go
sttts: oh, we are in test here.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,224 @@
...
+		if mappings.Scope.Name() == meta.RESTScopeNameRoot {
+			_, err = client.Resource(mappings.Resource).Create(manifests[path], metav1.CreateOptions{})
+		} else {
+			_, err = client.Resource(mappings.Resource).Namespace(manifests[path].GetNamespace()).Create(manifests[path], metav1.CreateOptions{})
+		}
+
+		// Resource already exists means we already succeeded
+		// This should never happen as we remove already created items from the manifest list, unless the resource existed beforehand.
+		if kerrors.IsAlreadyExists(err) {
+			delete(manifests, path)
```
sttts: why is it deleted only now and not when the create succeeds?

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,227 @@
...
+		mapper               meta.RESTMapper
+		needDiscoveryRefresh bool = true
+	)
+	err = wait.PollImmediateUntil(interval, func() (bool, error) {
+		retryCount++
+		// If we get rest mapper error, we need to pull updated discovery info from API server
+		if needDiscoveryRefresh {
+			mapper, err = fetchLatestDiscoveryInfoFn(dc)
+			if err != nil {
+				return false, err
```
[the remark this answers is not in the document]
mfojtik: @sttts i think we should survive that and probably skip create() when we see the error here and just retry?

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,230 @@
...
+		// Resource already exists means we already succeeded
+		// This should never happen as we remove already created items from the manifest list, unless the resource existed beforehand.
+		if kerrors.IsAlreadyExists(err) {
+			delete(manifests, path)
+			continue
+		}
+
+		if err != nil {
+			errs[path] = fmt.Errorf("failed to create: %v", err)
+			continue
```
sttts: so on error we stop the current loop. Should this be configurable? It might be that some (alphabetically) later resources are necessary for earlier resources to succeed. In CVO we probably want `continue`, in unsorted manifests we don't (not sure which case the rendered manifests are).

sttts: oops, no. it's continue

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,235 @@
...
+	// Verbose if true will print out extra messages for debugging
+	Verbose bool
+
+	// StdErr allows to override the standard error output for printing verbose messages.
+	// If not set, os.StdErr is used.
+	StdErr io.Writer
+
+	// SortFilenames allows to sort all files alphabetically when creating them.
+	// By default, the manifests creation order is not guaranteed.
+	SortFilenames bool
```
sttts: not sure this is what we want. It's most important we continue creation of remaining resources even after an error. Random order or not, I don't care much. I would always sort, so it's deterministic and easier to read.

mfojtik: ok :-)

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,229 @@
...
+	// Filters allows to filter which files we will read from disk.
+	// Multiple filters can be specified, in that case only files matching all filters will be returned.
+	Filters []assets.FileInfoPredicate
+
+	// Verbose if true will print out extra messages for debugging
+	Verbose bool
+
+	// StdErr allows to override the standard error output for printing verbose messages.
+	// If not set, os.StdErr is used.
+	StdErr io.Writer
```
wking: Seems like `Verbose` and `StdErr` should be replaced with a slot for a logrus logger or generic logging interface.  I can file a follow-up for that.

pkg/assets/create/creater.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+
+	if options.Verbose && options.StdErr == nil {
+		options.StdErr = os.Stderr
+	}
+
+	// Default QPS in client (when not specified) is 5 requests/per second
+	// This specifies the interval between "create-all-resources", no need to make this configurable.
+	interval := 200 * time.Millisecond
```
wking: Still not clear to me why we need more sleeping on top of `restConfig`'s limits.

pkg/assets/create/creater.go
wking: > I can file a follow-up for that.

Filed as #225.

pkg/assets/create/creater.go
wking: > Still not clear to me why we need more sleeping on top of `restConfig`'s limits.

I've filed #226 dropping this additional sleep.

PR #222, opened 2019-02-04
installer: handle installer failures better

sttts:
- do not proceed with other nodes if one node installation fails
- retry on OOM of the installer

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -212,7 +212,7 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 			prevNodeState = &operatorStatus.NodeStatuses[prev]
 		}
 
-		// if we are in a transition, check to see if our installer pod completed
+		// if we are in a transition, check to see whether our installer pod completed
 		if currNodeState.TargetRevision > currNodeState.CurrentRevision {
 			if err := c.ensureInstallerPod(currNodeState.NodeName, operatorSpec, currNodeState.TargetRevision); err != nil {
 				c.eventRecorder.Warningf("InstallerPodFailed", "Failed to create installer pod for revision %d on node %q: %v",
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -230,6 +230,13 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 		return false, nil
 	}
 
+	// stop on first deployment failure of the latest revision (excluding OOM, that never sets LatestAvailableRevision).
+	for _, s := range operatorStatus.NodeStatuses {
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -156,31 +157,65 @@ func nodeToStartRevisionWith(getStaticPodState func(nodeName string) (state stat
 		}
 	}
 
-	// otherwise try to find a node that is not ready
+	// otherwise try to find a node that is not ready. Take the oldest one.
+	oldestNotReadyRevisionNode := -1
+	oldestNotReadyRevision := math.MaxInt32
 	for i := range nodes {
```
mfojtik: nit: can you change `i` to `nodeID` ? makes the code more readable imho.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -156,31 +157,65 @@ func nodeToStartRevisionWith(getStaticPodState func(nodeName string) (state stat
 		}
 	}
 
-	// otherwise try to find a node that is not ready
+	// otherwise try to find a node that is not ready. Take the oldest one.
```
mfojtik: what does it mean "oldest" ? the one that has the lowest revision?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -156,31 +157,65 @@ func nodeToStartRevisionWith(getStaticPodState func(nodeName string) (state stat
...
+		}
+	}
+	if oldestNotReadyRevisionNode >= 0 {
+		return oldestNotReadyRevisionNode, nil
 	}
 
-	// last but not least, find a node that is has the wrong revision
+	// find a node that is has the wrong revision. Take the oldest one.
+	oldestPodRevisionNode := -1
+	oldestPodRevision := math.MaxInt32
```
mfojtik: we better not have more than 2147483647 revisions :D

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -156,31 +157,65 @@ func nodeToStartRevisionWith(getStaticPodState func(nodeName string) (state stat
...
+			oldestNotReadyRevisionNode = i
+			oldestNotReadyRevision = revisionNum
+		}
+	}
+	if oldestNotReadyRevisionNode >= 0 {
+		return oldestNotReadyRevisionNode, nil
 	}
 
-	// last but not least, find a node that is has the wrong revision
+	// find a node that is has the wrong revision. Take the oldest one.
```
mfojtik: what means "wrong" revision? you looking for a node that is on the oldest revision?

mfojtik:
/lgtm

We will add markdown explaining the logic of picking nodes and rolling revisions as a follow up.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: Not before OpenShift 5.0

PR #224, opened 2019-02-05
make it possible for an installer pod to have extra args

deads2k:
etcd needs to be able to change the static pod on a per node basis.  This makes it possible to register pod mutation functions in both the installer command (to adjust the static pod) and the installer controller (to tell the installer command what to adjust).

@hexfusion @mfojtik @sttts

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -70,6 +70,16 @@ type InstallerController struct {
 	installerPodImageFn func() string
 	// ownerRefsFn sets the ownerrefs on the pruner pod
 	ownerRefsFn func(revision int32) ([]metav1.OwnerReference, error)
+
+	installerPodMutationFns []InstallerPodMutationFunc
+}
+
+// InstallerPodMutationFunc is a function that has a chance at changing the installer pod before it is created
+type InstallerPodMutationFunc func(pod *corev1.Pod, nodeName string, operatorSpec *operatorv1.StaticPodOperatorSpec, revision int32) error
+
+func (o *InstallerController) WithInstallerPodMutationFn(installerPodMutationFn InstallerPodMutationFunc) *InstallerController {
+	o.installerPodMutationFns = append(o.installerPodMutationFns, installerPodMutationFn)
+	return o
 }
 
 // staticPodState is the status of a static pod that has been installed to a node.
@@ -562,6 +572,13 @@ func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *
 		}
 	}
 	pod.Spec.Containers[0].Args = args
+
+	// Some owners need to change aspects of the pod.  Things like arguments for instance
+	for _, fn := range c.installerPodMutationFns {
+		if err := fn(pod, nodeName, operatorSpec, revision); err != nil {
+			return err
+		}
+	}
 
 	_, _, err = resourceapply.ApplyPod(c.kubeClient.CoreV1(), c.eventRecorder, pod)
 	return err
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -36,10 +36,19 @@ type staticPodOperatorControllers struct {
 // 4. BackingResourceController - this creates the backing resources needed for the operand, such as cluster rolebindings and installer service
 //    account.
 // 5. MonitoringResourceController - this creates the service monitor used by prometheus to scrape metrics.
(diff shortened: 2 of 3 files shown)


deads2k:
see usage in https://github.com/openshift/cluster-etcd-operator/pull/8

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -70,6 +70,16 @@ type InstallerController struct {
...
 	// ownerRefsFn sets the ownerrefs on the pruner pod
 	ownerRefsFn func(revision int32) ([]metav1.OwnerReference, error)
+
+	installerPodMutationFn InstallerPodMutationFunc
+}
+
+// InstallerPodMutationFunc is a function that has a chance at changing the installer pod before it is created
+type InstallerPodMutationFunc func(pod *corev1.Pod, nodeName string, operatorSpec *operatorv1.StaticPodOperatorSpec, revision int32) error
+
+func (o *InstallerController) WithInstallerPodMutationFn(installerPodMutationFn InstallerPodMutationFunc) *InstallerController {
```
sttts: func sounds like those funcs can be chained by calling it multiple times.

deads2k: sure, you prefer an appending slice of mutators?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -227,7 +235,17 @@ func (o *InstallOptions) copyContent() error {
 	if err := os.MkdirAll(o.PodManifestDir, 0755); err != nil {
 		return err
 	}
+
+	if o.PodMutationFn != nil {
+		glog.Infof("mutating static pod")
```
mfojtik: V(2)?

pkg/operator/staticpod/controllers.go
```diff
@@ -63,7 +73,7 @@ func NewControllers(targetNamespaceName, staticPodName, podResourcePrefix string
 		staticPodOperatorClient,
 		kubeClient,
 		eventRecorder,
-	)
+	).WithInstallerPodMutationFn(installerPodMutationFn)
```
sttts: alternatively turn this into another `WithInstallerPodMutationFn` func against `staticPodOperatorControllers`. We are clearly getting too many parameters. `With*` funcs will help to keep the usual case shorter.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -227,7 +235,17 @@ func (o *InstallOptions) copyContent() error {
...
+		pod := resourceread.ReadPodV1OrDie([]byte(podContent))
+		if err := o.PodMutationFn(pod); err != nil {
+			return err
+		}
+		podContent = resourceread.WritePodV1OrDie(pod)
+	}
 	glog.Infof("writing static pod %q", path.Join(o.PodManifestDir, podFileName))
+	glog.Infof("writing static pod\n%s", podContent)
```
mfojtik: V(2) probably as well? or higher?

pkg/operator/staticpod/installerpod/cmd.go
sttts: this sounds dangerous, like a warning. "Customizing" or "Adapting" is a better word maybe.

deads2k: sure

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: wonder if we can do objectdiff here to make the mutating changes clear in log (what changed)

pkg/operator/staticpod/controllers.go
deads2k: > alternatively turn this into another `WithInstallerPodMutationFn` func against `staticPodOperatorControllers`. We are clearly getting too many parameters. `With*` funcs will help to keep the usual case shorter.

hrm, no struct.  Can I get an IOU?

pkg/operator/staticpod/installerpod/cmd.go
deads2k: > V(2) probably as well? or higher?

small log.  You want this when it fails.

pkg/operator/staticpod/installerpod/cmd.go
deads2k: > this sounds dangerous, like a warning. "Customizing" or "Adapting" is a better word maybe.

ok

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: but i probably want to see what changed (i see only updated pod right?)

deads2k:
comments addressed

PR #225, opened 2019-02-05
pkg/assets/create: Replace Verbose and StdErr with Logger

wking:
To allow consumers to plug in their preferred logger instead of requring an output stream.  I'm using [go-log/log][1] as an unopinionated interface, which callers can feed with [the stdlib's log package][2], [logrus][3], etc. depending on their logging preferences.

CC @mfojtik, @sttts

[1]: https://github.com/go-log/log
[2]: https://golang.org/pkg/log/
[3]: https://godoc.org/github.com/sirupsen/logrus


mfojtik:
We should not import new package here, instead we can create the interface here and plug whatever we need there (the interfaces are exchangeable). I would just copy https://github.com/go-log/log/blob/master/log.go#L5 at the top of the create.go

pkg/assets/create/creater.go
```diff
@@ -53,8 +48,8 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
 		return err
 	}
 
-	if options.Verbose && options.StdErr == nil {
-		options.StdErr = os.Stderr
+	if options.Logger == nil {
```
mfojtik: If the options.Logger is not set, just don't log anything or create some dummy implementation of Logger interface?

tnozicka:
/hold

The whole org uses glog for very clear reason and because the whole k8s apimachinery and client-go is tightly coupled to it. Using more then 1 logger is a lost of time and **will not work** given the hardcoded glog in k8s. You just don't use 2 loggers in 1 project.

sttts:
Decoupling the code from glog is fine, but not by pulling in another package. Let's just abstract with a local interface and every consumer can plug a wrapper in for his/her favourite logger.

tnozicka:
Making more loggers configurable is work that could have been better spent on important things, same goes for reviews. 

In the end the same statement applies for logging in the end component, if it doesn't use glog.

(note: I personally don't like glog at all, but I understand the necessity of using it implied by being hardcoded in k8s.)

pkg/assets/create/creater.go
wking: > If the options.Logger is not set, just don't log anything or create some dummy implementation of Logger interface?

The default I'm plugging in in this block [is a no-op logger][1] unless the user set that package-level global.

[1]: https://github.com/go-log/log/blob/a514cf01a3eb56ec6aff9ae5663453d990e68ed3/log.go#L17

wking:
> The whole org uses glog for very clear reason...

The installer inherited logrus.  I don't see it consuming this code directly, but I also don't see a need for opinionated logging in libraries.  However, I can reroll to take glog (or [klog][1]?) types if folks don't like the interface approach.

> We should not import new package here, instead we can create the interface here and plug whatever we need there (the interfaces are exchangeable).

Why copy/paste this (and the no-op implementation), when they're already written and can be vendored?  Seems easier to maintain that way, and if your vendor tooling pruned unused packages it would be about the same amount of code either way.

[1]: https://github.com/kubernetes/klog

wking:
Also, lots of Kubernetes library log handling discussion in kubernetes/kubernetes#61006, where interface-based approaches were raised but not accepted (although I'm still not clear on why not).

soltysh:
> The installer inherited logrus. I don't see it consuming this code directly, but I also don't see a need for opinionated logging in libraries. However, I can reroll to take glog (or [klog](https://github.com/kubernetes/klog)?) types if folks don't like the interface approach.

Klog will be brought up with 1.13 kubernetes, so I wouldn't bother with that, yet.

> Also, lots of Kubernetes library log handling discussion in [kubernetes/kubernetes#61006](https://github.com/kubernetes/kubernetes/issues/61006), where interface-based approaches were raised but not accepted (although I'm still not clear on why not).

That's one of the reasons klog happened, it might be a road towards just that, we'll see.

pkg/assets/create/creater.go
```diff
@@ -3,13 +3,12 @@ package create
...
 	"context"
 	"fmt"
-	"io"
-	"os"
 	"sort"
 	"strings"
 	"time"
 
 	"github.com/ghodss/yaml"
+	"github.com/go-log/log"
```
deads2k: no new logging dep.  Project standard is glog and we'll switch to klog with kube.

pkg/assets/create/creater.go
```diff
@@ -74,19 +69,15 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
 		if needDiscoveryRefresh {
 			mapper, err = fetchLatestDiscoveryInfoFn(dc)
 			if err != nil {
-				if options.Verbose {
-					fmt.Fprintf(options.StdErr, "[#%d] failed to fetch discovery: %s\n", retryCount, err)
-				}
+				options.Logger.Logf("[#%d] failed to fetch discovery: %s", retryCount, err)
```
deads2k: project standard upstream and downstream is fmt.printf for CLI and glog for everything else.  The previous code is in keeping.

wking: > project standard upstream and downstream is fmt.printf for CLI and glog for everything else. The previous code is in keeping.

But this is library code, so I'd expect glog over fmt, right?

mfojtik: @wking depends on the caller of this code. So far it is cluster-bootstrap which is a command we run during the bootstrap (so it is CLI ;-). I'm fine switching this to glog, was trying to avoid importing glog because I know that some components/cli have problem with parsing glog flags (so wanted to avoid that if not necessary).

wking: > depends on the caller of this code. So far it is cluster-bootstrap which is a command we run during the bootstrap (so it is CLI ;-).

I don't think code in library-go should have API choices hinge on "we know all of our consumers, and they're not libraries themselves".  Is there a reason that approach is not as brittle as it seems?

> I'm fine switching this to glog, was trying to avoid importing glog because I know that some components/cli have problem with parsing glog flags (so wanted to avoid that if not necessary).

Like openshift/installer#1181.  This is exactly why I prefer logging interfaces, but it sounds like that's a no-go?

wking:
Hrm.  I went to reroll this to use glog as requested by @tnozicka [here][1] @deads2k [here][2].  But I'm stuck with collecting the logs for unit tests.  Should I drop those tests?  Should I have `init` code set `-log_dir` and collect the test logs there?  Should I add an interface after all so the test suite can plug in a capture implementation (but with copies to local code, and not the vendored capture implementation I'm currently using as of 673e3d38ede?

[1]: https://github.com/openshift/library-go/pull/225#issuecomment-460963484
[2]: https://github.com/openshift/library-go/pull/225#discussion_r254264855

PR #226, opened 2019-02-05
pkg/assets/create: Allow mid-create cancels

wking:
As mentioned in the `interval` comment I'm removing, the REST client configuration includes its own requests-per-second configuration. There's no need for an additional 200ms sleep between loops on top of that.

I'm also passing the context into `create()` so we can cancel without waiting for all of the manifests to be attempted.  At the default 5 requests per second, the installer's 30+ manifests would take 6+ seconds in the "negligable network time" best-case, and we can be more responsive to cancels than that.

I've also shifted `retryCount` into the `for` loop, because that's where I'm used to seeing iteration counters.  And I've dropped logging on returned errors, because the caller will have the error and can make
that logging decision itself.  Logging should be reserved for internal progress updates, discarded errors, and other activity which the caller cannot access.

CC @mfojtik, @sttts

### DIFF
--- a/pkg/assets/create/create_test.go
+++ b/pkg/assets/create/create_test.go
@@ -138,6 +138,8 @@ func TestEnsureManifestsCreated(t *testing.T) {
 }
 
 func TestCreate(t *testing.T) {
+	ctx := context.Background()
+
 	resourcesWithoutKubeAPIServer := resources[1:]
 	testConfigMap := &unstructured.Unstructured{}
 	testConfigMap.SetGroupVersionKind(schema.GroupVersionKind{
@@ -188,7 +190,7 @@ func TestCreate(t *testing.T) {
 			dynamicClient := dynamicfake.NewSimpleDynamicClient(fakeScheme, tc.existingObjects...)
 			restMapper := restmapper.NewDiscoveryRESTMapper(tc.discovery)
 
-			err, reload := create(manifests, dynamicClient, restMapper, CreateOptions{Verbose: true, StdErr: os.Stderr})
+			err, reload := create(ctx, manifests, dynamicClient, restMapper, CreateOptions{Verbose: true, StdErr: os.Stderr})
 			if tc.expectError && err == nil {
 				t.Errorf("expected error, got no error")
 				return
--- a/pkg/assets/create/creater.go
+++ b/pkg/assets/create/creater.go
@@ -80,12 +80,16 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig 
 				return false, nil
 			}
 		}
-		lastCreateError, needDiscoveryRefresh = create(manifests, client, mapper, options)
-		if lastCreateError == nil {
+		err, needDiscoveryRefresh = create(ctx, manifests, client, mapper, options)
+		if err == nil {
+			lastCreateError = nil
 			return true, nil
 		}
+		if ctx.Err() == nil || lastCreateError == nil {
+			lastCreateError = err
+		}
 		if options.Verbose {
-			fmt.Fprintf(options.StdErr, "[#%d] %s\n", retryCount, lastCreateError)
+			fmt.Fprintf(options.StdErr, "[#%d] %s\n", retryCount, err)
 		}
 		return false, nil
 	}, ctx.Done())
@@ -131,7 +135,7 @@ func fetchLatestDiscoveryInfo(dc *discovery.DiscoveryClient) (meta.RESTMapper, e
 // create will attempt to create all manifests provided using dynamic client.
 // It will mutate the manifests argument in case the create succeeded for given manifest. When all manifests are successfully created the resulting
 // manifests argument should be empty.
-func create(manifests map[string]*unstructured.Unstructured, client dynamic.Interface, mapper meta.RESTMapper, options CreateOptions) (error, bool) {
+func create(ctx context.Context, manifests map[string]*unstructured.Unstructured, client dynamic.Interface, mapper meta.RESTMapper, options CreateOptions) (error, bool) {
 	sortedManifestPaths := []string{}
 	for key := range manifests {
 		sortedManifestPaths = append(sortedManifestPaths, key)
@@ -145,6 +149,12 @@ func create(manifests map[string]*unstructured.Unstructured, client dynamic.Inte
 	reloadDiscovery := false
 
 	for _, path := range sortedManifestPaths {
+		select {
+		case <-ctx.Done():
+			return ctx.Err(), false
+		default:
+		}
+
 		gvk := manifests[path].GetObjectKind().GroupVersionKind()
 		mappings, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
 		if err != nil {


pkg/assets/create/creater.go
```diff
@@ -57,45 +55,47 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
...
 		mapper               meta.RESTMapper
 		needDiscoveryRefresh bool = true
 	)
-	err = wait.PollImmediateUntil(interval, func() (bool, error) {
-		retryCount++
+	for retryCount := 1; ; retryCount++ {
+		if retryCount > 1 {
+			select {
+			case <-ctx.Done():
+				if lastCreateError == nil {
```
abhinavdahiya: nit: completely ignorable ;)
```go
if lastCreateError != nil {
    return lastCreateError
}
return ctx.Err()
```
IMO that is more `Return the last observed set of errors from the create process instead of timeout error.` ...

pkg/assets/create/creater.go
```diff
@@ -57,45 +55,47 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
...
+				// Return the last observed set of errors from the create process instead of timeout error.
+				return lastCreateError
+			default:
+			}
+		}
 		// If we get rest mapper error, we need to pull updated discovery info from API server
 		if needDiscoveryRefresh {
 			mapper, err = fetchLatestDiscoveryInfoFn(dc)
 			if err != nil {
-				if options.Verbose {
```
abhinavdahiya: Okay this is changing the beahvior that discovery errors would cause retry after interval and not exit the function...

pkg/assets/create/creater.go
wking: > nit: completely ignorable ;)
>
> ```go
> if lastCreateError != nil {
>     return lastCreateError
> }
> return ctx.Err()
> ```

Isn't that just reversing the order (but not changing the behavior) of my current:

```go
if lastCreateError == nil {
  return ctx.Err()
}
return lastCreateError
```

?  I'm fine updating if you like, but when I have both `if` and `else` cases (with a flattened `else` here) with short bodies, I prefer the positive `==` to the negative `!=` so you don't have a double-negative in your head for the `else` case.

pkg/assets/create/creater.go
wking: > Okay this is changing the beahvior that discovery errors would cause retry after interval and not exit the function...

Ah, good catch.  For some reason I'd read the old behavior as a fail-fast, but it wasn't.  Fixed to retry these too with 7d5d9ca -> bfedebc.

pkg/assets/create/create_test.go
```diff
@@ -102,17 +102,16 @@ func TestEnsureManifestsCreated(t *testing.T) {
...
 	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
-	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true, StdErr: out})
+	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true})
 	if err == nil {
 		t.Fatal("expected error creating kubeapiserverconfig resource, got none")
 	}
-	if !strings.Contains(out.String(), "unable to get REST mapping") {
-		t.Fatalf("expected error logged to output when verbose is on, got: %s\n", out.String())
+	if !strings.Contains(err.Error(), "unable to get REST mapping") {
```
sttts: @mfojtik I should commented earlier: we need a test for this that the error string did not change in apimachinery.

pkg/assets/create/creater.go
```diff
@@ -57,45 +55,50 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
...
-	// Retry creation until no errors are returned or the timeout is hit.
 	var (
 		lastCreateError      error
-		retryCount           int
 		mapper               meta.RESTMapper
 		needDiscoveryRefresh bool = true
 	)
-	err = wait.PollImmediateUntil(interval, func() (bool, error) {
-		retryCount++
+	for retryCount := 1; ; retryCount++ {
```
sttts: `wait.PollImmediateUntil` is the canonical kubernetes way of such a loop. I would stay with it.

mfojtik: Yeah, it is used and well tested in many kube components and packages.

pkg/assets/create/create_test.go
```diff
@@ -102,17 +102,16 @@ func TestEnsureManifestsCreated(t *testing.T) {
...
 	operatorResource := resourcesForEnsure[0]
 	resourcesForEnsure = resourcesForEnsure[1:]
 	ctx, cancel = context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
-	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true, StdErr: out})
+	err = EnsureManifestsCreated(ctx, "testdata", nil, CreateOptions{Verbose: true})
 	if err == nil {
 		t.Fatal("expected error creating kubeapiserverconfig resource, got none")
 	}
-	if !strings.Contains(out.String(), "unable to get REST mapping") {
```
mfojtik: this was in fact testing that the messages are seen with verbose:true ;-) the error is just aggregation on those messages.

pkg/assets/create/creater.go
```diff
@@ -57,45 +55,50 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
 		options.StdErr = os.Stderr
 	}
 
-	// Default QPS in client (when not specified) is 5 requests/per second
-	// This specifies the interval between "create-all-resources", no need to make this configurable.
-	interval := 200 * time.Millisecond
```
mfojtik: This interval just match the client QPS, to the `PollImmediateUntil(200*time.Millisecond)` just does what client does (there is no delay).

pkg/assets/create/creater.go
```diff
@@ -144,7 +147,15 @@ func create(manifests map[string]*unstructured.Unstructured, client dynamic.Inte
 	// In case we fail to find a rest-mapping for the resource, force to fetch the updated discovery on next run.
 	reloadDiscovery := false
 
-	for _, path := range sortedManifestPaths {
+	for i, path := range sortedManifestPaths {
+		if i > 0 {
+			select {
+			case <-ctx.Done():
+				return ctx.Err(), false
```
mfojtik: i'm not fan of cancelling in middle of the bulk run. i would rather give the loop the extra time to potentially get positive outcome (maybe it will succeed this run) instead of cancelling and error out.

pkg/assets/create/creater.go
```diff
@@ -57,45 +55,50 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
...
 			}
 		}
-		lastCreateError, needDiscoveryRefresh = create(manifests, client, mapper, options)
-		if lastCreateError == nil {
-			return true, nil
+		err, needDiscoveryRefresh = create(ctx, manifests, client, mapper, options)
+		if err == nil {
+			return nil
+		}
+		if err == context.Canceled || err == context.DeadlineExceeded {
```
mfojtik: i believe this is handled in `context.Done()` channel already.

pkg/assets/create/create_test.go
mfojtik: @sttts if it change, this test will fail (i don't this we rely on this message in the code).

sttts:
I like the overall change of adding the context handling in `create`. But let's strip the change down to the actual behaviour change, and not mix it with coding style changes (especially the `wait. PollImmediateUntil` change). This makes the PR twice as big as necessary, an 10 times harder to review.

pkg/assets/create/creater.go
wking: > `wait.PollImmediateUntil` is the canonical kubernetes way of such a loop. I would stay with it.

And just set the delay to zero?  It doesn't seem useful enough if all it does is wrap an initial `Context` check.

pkg/assets/create/creater.go
wking: > i would rather give the loop the extra time to potentially get positive outcome (maybe it will succeed this run)...

Then don't cancel it?  I don't think we want to re-interpret "please give up and gently shut down" as "I'm starting to get bored, but go ahead and run for another several minutes if you think it might work" ;).

pkg/assets/create/create_test.go
wking: > this was in fact testing that the messages are seen with verbose:true ;-) the error is just aggregation on those messages.

As I discuss in the commit message, I think the returned error is more important.  But I can update this to test both.

pkg/assets/create/creater.go
```diff
@@ -57,45 +55,50 @@ func EnsureManifestsCreated(ctx context.Context, manifestDir string, restConfig
...
-	interval := 200 * time.Millisecond
-
-	// Retry creation until no errors are returned or the timeout is hit.
 	var (
 		lastCreateError      error
-		retryCount           int
 		mapper               meta.RESTMapper
 		needDiscoveryRefresh bool = true
 	)
-	err = wait.PollImmediateUntil(interval, func() (bool, error) {
```
deads2k: `PollImmediate` is project standard.

pkg/assets/create/creater.go
deads2k: > i believe this is handled in `context.Done()` channel already.

Yeah, this appears to just duplicate the existing code.

pkg/assets/create/creater.go
deads2k: > i'm not fan of cancelling in middle of the bulk run. i would rather give the loop the extra time to potentially get positive outcome (maybe it will succeed this run) instead of cancelling and error out.

I may let this play out.  I wouldn't gate it on that if above though.

pkg/assets/create/creater.go
wking: > `PollImmediate` is project standard

Even when [you don't want to sleep between iterations][1]?

[1]: https://github.com/openshift/library-go/pull/226#discussion_r254237203

pkg/assets/create/creater.go
wking: > I wouldn't gate it on that if above though.

So cancel immediately if requested by the `Context`, even if we've done nothing?

pkg/assets/create/creater.go
mfojtik: I can make that sleep 10ms ;-) But i wanted it to match the default client QpS (5 which means 200ms between requests). So a retry will wait the same time as client will do. Again, I don't have problem changing that interval to 10ms (I doubt we ever will have problems when we DoS API server with create calls).

wking: > I can make that sleep 10ms ;-)

It's still a wasted ticker goroutine ;).

> But i wanted it to match the default client QpS (5 which means 200ms between requests). So a retry will wait the same time as client will do.

But if we'd made a client request, we'd already get the client's internal limiting, right?  If you dump logs for this PR's tests as they stand, you see I'm not flooding fake client requests even without an external sleep.

mfojtik: In real life, that interval as it is now, will likely never cause any problem or extensive delay (unless we plan to use this helper to create >500 manifests). In real life, pulling updated discovery from the server will probably take seconds already and every non-CR resource will be created at the first run. So whether we remove this interval or set it to 10ms or 400ms won't make that much difference ;-)

wking: > So instead of writing custom polling and cancellation logic, I would prefer to follow the k8s style.

A `for` loop is not custom polling logic ;).  The `select` for cancelation checking is also a standard approach, and seems simpler to me to inline when you don't need a wait vs. calling `PollImmediateUntil` just to wrap this one `Context` `select`.  But I'm not a maintainer ;).  You're all sure you prefer `PollImmediateUntil` over `for`?

mfojtik: yeah, PollImmediateUntil is k8s standard as @deads2k already mentioned.

wking:
I still think that `PollImmediateUntil` is more complicated than `for` here, but since that seems contentious I've pushed bfedebc -> 85e4130 with a minimal pivot to a cancel-able `create` as [requested][1] by @sttts.  There's a bit more of a dance around `lastCreateError` now that `create` can also return boring canceled and timed-out errors, while we still want to prefer the last actual manifest-creation failure.

[1]: https://github.com/openshift/library-go/pull/226#issuecomment-460971213

PR #227, opened 2019-02-06
installerpod: fix retry handling and report events

mfojtik:
* Fix the retry logic to only retry failed gets from API server
* Use `context.Context` to handle timeout
* Report events back to API server (success/failure) (it seems like the ownerRefs for the installer pod is the ConfigMap so the events will be recorded for that object... but at least we get something ;-)

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -1,6 +1,7 @@
 package installerpod
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -16,10 +18,14 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/openshift/library-go/pkg/config/client"
+	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 )
 
@@ -39,6 +45,9 @@ type InstallOptions struct {
 
 	ResourceDir    string
 	PodManifestDir string
+
+	Timeout  time.Duration
+	NodeName string
 
 	PodMutationFns []PodMutationFunc
 }
@@ -71,7 +80,10 @@ func NewInstaller() *cobra.Command {
 			if err := o.Validate(); err != nil {
 				glog.Fatal(err)
 			}
-			if err := o.Run(); err != nil {
+
+			ctx, cancel := context.WithTimeout(context.TODO(), time.Duration(o.Timeout)*time.Second)
+			defer cancel()
+			if err := o.Run(ctx); err != nil {
 				glog.Fatal(err)
 			}
 		},
@@ -93,6 +105,7 @@ func (o *InstallOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringSliceVar(&o.OptionalConfigMapNamePrefixes, "optional-configmaps", o.OptionalConfigMapNamePrefixes, "list of optional configmaps to be included")
 	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
 	fs.StringVar(&o.PodManifestDir, "pod-manifest-dir", o.PodManifestDir, "directory for the static pod manifest")
+	fs.DurationVar(&o.Timeout, "timeout-duration", 120*time.Second, "maximum time in seconds to wait for the copying to complete (default: 2m)")
 }
 
 func (o *InstallOptions) Complete() error {
@@ -120,6 +133,9 @@ func (o *InstallOptions) Validate() error {
 	if len(o.ConfigMapNamePrefixes) == 0 {
 		return fmt.Errorf("--configmaps is required")
 	}
+	if o.Timeout == 0 {
+		return fmt.Errorf("--timeout-duration cannot be 0")
+	}
 
 	if o.KubeClient == nil {
 		return fmt.Errorf("missing client")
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -40,6 +44,9 @@ type InstallOptions struct {
 	ResourceDir    string
 	PodManifestDir string
 
+	TimeoutSeconds int32
```
sttts: time.Duration

sttts: Don't we have an nice flag type for that?

mfojtik: `(type *time.Duration) as type *int32 in argument to fs.Int32Var`

mfojtik: DurationVar! didn't know that exist

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +150,163 @@ func (o *InstallOptions) prefixFor(name string) string {
...
 	secrets := []*corev1.Secret{}
-	for _, currPrefix := range o.SecretNamePrefixes {
-		glog.Infof("getting secrets/%s -n %s", o.nameFor(currPrefix), o.Namespace)
-		val, err := o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(currPrefix), metav1.GetOptions{})
-		if err != nil {
-			return err
+	err := utilwait.PollImmediateUntil(200*time.Millisecond, func() (bool, error) {
+		for _, prefix := range append(secretPrefixes.List(), optionalSecretPrefixes.List()...) {
+			secret, err := o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(prefix), metav1.GetOptions{})
+			if err != nil {
```
sttts: are you sure we want to repeat on NotFound?

mfojtik: why not? it might be that the secret is being recreated by other controller or we waiting for some process to add that secret.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +150,164 @@ func (o *InstallOptions) prefixFor(name string) string {
...
-		if err != nil {
-			return err
+
+	// Gather all config maps
+	configs := []*corev1.ConfigMap{}
+	err = utilwait.PollImmediateUntil(200*time.Millisecond, func() (bool, error) {
+		for _, prefix := range append(configPrefixes.List(), optionalConfigPrefixes.List()...) {
+			config, err := o.KubeClient.CoreV1().ConfigMaps(o.Namespace).Get(o.nameFor(prefix), metav1.GetOptions{})
+			if err != nil {
+				glog.Infof("Failed to get config map %s/%s: %v", o.Namespace, o.nameFor(prefix), err)
```
sttts: now this retries OptionalConfigMapNamePrefixes, but probably should ignore NotFounds.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -9,17 +10,22 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/api/errors"
```
sttts: order

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -93,6 +105,7 @@ func (o *InstallOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringSliceVar(&o.OptionalConfigMapNamePrefixes, "optional-configmaps", o.OptionalConfigMapNamePrefixes, "list of optional configmaps to be included")
 	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
 	fs.StringVar(&o.PodManifestDir, "pod-manifest-dir", o.PodManifestDir, "directory for the static pod manifest")
+	fs.DurationVar(&o.TimeoutSeconds, "timeout-seconds", 120*time.Second, "maximum time in seconds to wait for the copying to complete (default: 2m)")
```
sttts: timeout-duration. Now it is not seconds anymore, but any unit can be passed.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
-			return err
+	err := utilwait.PollImmediateUntil(200*time.Millisecond, func() (bool, error) {
+		for _, prefix := range append(secretPrefixes.List(), optionalSecretPrefixes.List()...) {
+			secret, err := o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(prefix), metav1.GetOptions{})
+			// When we encounter not found for required secret, return immediately
+			if errors.IsNotFound(err) {
+				if optionalSecretPrefixes.Has(prefix) {
+					optionalSecretPrefixes.Delete(prefix)
+					continue
+				} else {
```
sttts: drop the else

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
 
-	// gather all configmaps
-	configmaps := []*corev1.ConfigMap{}
-	for _, currPrefix := range o.ConfigMapNamePrefixes {
-		glog.Infof("getting configmaps/%s -n %s", o.nameFor(currPrefix), o.Namespace)
-		val, err := o.KubeClient.CoreV1().ConfigMaps(o.Namespace).Get(o.nameFor(currPrefix), metav1.GetOptions{})
+	// Gather pod yaml from config map
+	var podContent string
+	err = utilwait.PollImmediateUntil(200*time.Millisecond, func() (bool, error) {
+		glog.Infof("getting pod configmaps/%s -n %s", o.nameFor(o.PodConfigMapNamePrefix), o.Namespace)
```
sttts: nit: capital `Get` to be consistent.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
-		}
-		configmaps = append(configmaps, val)
-	}
-	for _, currPrefix := range o.OptionalConfigMapNamePrefixes {
-		glog.Infof("getting optional configmaps/%s -n %s", o.nameFor(currPrefix), o.Namespace)
-		val, err := o.KubeClient.CoreV1().ConfigMaps(o.Namespace).Get(o.nameFor(currPrefix), metav1.GetOptions{})
-		if errors.IsNotFound(err) {
-			glog.Infof("missing optional configmaps/%s -n %s", o.nameFor(currPrefix), o.Namespace)
-			continue
+			glog.Infof("Failed to get pod %s/%s: %v", o.Namespace, o.nameFor(o.PodConfigMapNamePrefix), err)
```
sttts: NotFound should be fatal here as fell

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
+			return false, nil
 		}
-		if err != nil {
-			return err
+		if podData, exists := podConfigMap.Data["pod.yaml"]; !exists {
+			return false, fmt.Errorf("required 'pod.yaml' key does not exists in configmap")
```
sttts: nit: does not exist

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
+		} else {
```
sttts: drop else

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -136,144 +152,188 @@ func (o *InstallOptions) prefixFor(name string) string {
...
-			return false, nil
-		}
-		return true, nil
-	})
+	unstructuredPodObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, podJSONBytes)
 	if err != nil {
-		return fmt.Errorf("error: %v", err)
+		return "", err
 	}
+	nodeName, _, err := unstructured.NestedString(unstructuredPodObj.(*unstructured.Unstructured).Object, "spec", "nodeName")
```
sttts: nit: the 2nd return value tells you whether the fields existed. Might be worth to check for `len(nodeName) != 0`.

mfojtik: i think we can return "<unknown>" in that case?

sttts: This won't happen anyway.

mfojtik:
/retest

seems like we are short on memory in unit test pod...

PR #228, opened 2019-02-06
target signers should be signed by the original signer

deads2k:
Signers that were rotation targets were actually getting self-signed, defeating the purpose of a longer trust cycle.  This fixes that signer and includes it in the list of certs.

/assign @sjenning @enj


enj:
/lgtm

@mrogers950 PTAL

PR #229, opened 2019-02-08
set failing when static pods are not all healthy

deads2k:
When static pods fail to be ready or crashloop, we should update our status. This sets failing because we are failing and it is a union condition.  I thing we should consider making them all union conditions.

@openshift/sig-master

### DIFF
--- a/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
+++ b/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
@@ -1,0 +1,168 @@
+package staticpodstate
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+var (
+	staticPodStateControllerFailing      = "StaticPodsFailing"
+	staticPodStateControllerWorkQueueKey = "key"
+)
+
+// StaticPodStateController is a controller that watches static pods and will produce a failing status if the
+//// static pods start crashing for some reason.
+type StaticPodStateController struct {
+	targetNamespace, staticPodName string
+
+	operatorConfigClient v1helpers.StaticPodOperatorClient
+	kubeClient           kubernetes.Interface
+	eventRecorder        events.Recorder
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+}
+
+// NewStaticPodStateController creates a controller that watches static pods and will produce a failing status if the
+// static pods start crashing for some reason.
+func NewStaticPodStateController(
+	targetNamespace, staticPodName string,
+	kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	operatorConfigClient v1helpers.StaticPodOperatorClient,
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder,
+) *StaticPodStateController {
+	c := &StaticPodStateController{
+		targetNamespace: targetNamespace,
+		staticPodName:   staticPodName,
+
+		operatorConfigClient: operatorConfigClient,
+		kubeClient:           kubeClient,
+		eventRecorder:        eventRecorder,
+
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
+package staticpodstate
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
```
mfojtik: nit: order

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
...
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+var (
+	staticPodStateControllerFailing      = "StaticPodsFailing"
```
mfojtik: nit: align with the value?

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
...
+	staticPodStateControllerWorkQueueKey = "key"
+)
+
+// StaticPodStateController is a controller that watches the currentRevision and targetRevision fields for each node and spawn
```
mfojtik: nit: this godoc seems off ;-)

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// NewStaticPodStateController creates a controller that watches static pods and will produce a failing status if the
+// static pods start crashing for some reason.
+func NewStaticPodStateController(
+	targetNamespace, staticPodName string,
+	kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	operatorConfigClient v1helpers.StaticPodOperatorClient,
+	kubeClient kubernetes.Interface,
+	eventRecorder events.Recorder,
```
mfojtik: do you record any events?

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
...
+
+	switch operatorSpec.ManagementState {
+	case operatorv1.Unmanaged:
+		return nil
+	case operatorv1.Removed:
+		// TODO probably just fail.  Static pod managers can't be removed.
+		return nil
+	}
+
+	errs := []error{}
```
mfojtik: hrm, we still don't have good lib to deal with error aggregation?

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -0,0 +1,166 @@
...
+		pod, err := c.kubeClient.CoreV1().Pods(c.targetNamespace).Get(mirrorPodNameForNode(c.staticPodName, node.NodeName), metav1.GetOptions{})
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		for _, containerStatus := range pod.Status.ContainerStatuses {
+			if !containerStatus.Ready {
+				errs = append(errs, fmt.Errorf("nodes/%s pods/%s container=%q is not ready", node.NodeName, pod.Name, containerStatus.Name))
+			}
+			if containerStatus.State.Waiting != nil {
```
mfojtik: not ready && waiting? do we want to report both?

pkg/operator/staticpod/controllers.go
```diff
@@ -1,6 +1,7 @@
 package staticpod
 
 import (
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/staticpodstate"
```
mfojtik: nit: order

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
deads2k: > do you record any events?

not yet.  I like the option though.

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
deads2k: > not ready && waiting? do we want to report both? (at the same time)

I think so, they have different levels of detail and I'm not sure they're quite the same thing.

deads2k:
comments addressed

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
soltysh: There's https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/errors/errors.go, but that's useful only when you want to filter through them. What's wrong with an array?

PR #232, opened 2019-02-11
operator: stop logging not found as errors

mfojtik:
Fixes pruning controller that logs no-op as error and status controller that logs not found as error when deleting cluster operator.


pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -138,7 +138,8 @@ func (c *PruneController) excludedRevisionHistory(operatorStatus *operatorv1.Sta
 
 	// Return early if nothing to prune
 	if len(succeededRevisionIDs)+len(failedRevisionIDs) == 0 {
-		return []int{}, fmt.Errorf("no revision IDs currently eligible to prune")
+		glog.V(2).Info("no revision IDs currently eligible to prune")
```
sttts: how often is this printed?

mfojtik: @sttts rarely

PR #233, opened 2019-02-12
Prepare to use protobuf client-go

mfojtik:
This change will prepare controllers to use protobuf enabled client-go. Via the `ControllerContext` we are going to provide two REST config (`KubeConfig` and `ProtoKubeConfig`). The later is a copy with accept set to prefer protobuf protocol.

We need to be careful to not use the protobuf for CR resources (@smarterclayton is fixing it?), so I renamed the `kubeClient` in controllers to `protoKubeClient`.

This change also remove usage of `kubernetes.Interface` and replace it with corresponding clients (pods/configmaps).

Lastly, this removes the `context.Context` from the controller context as storing the context in struct is discouraged in Go. Instead we just expose the `Done()` function.

### DIFF
--- a/pkg/config/client/client_config.go
+++ b/pkg/config/client/client_config.go
@@ -21,8 +21,11 @@ func GetKubeConfigOrInClusterConfig(kubeConfigFile string, overrides *ClientConn
 	if err != nil {
 		return nil, err
 	}
+
 	applyClientConnectionOverrides(overrides, clientConfig)
-	clientConfig.WrapTransport = defaultClientTransport
+
+	t := clientTransportOverrides{maxIdleConnsPerHost: overrides.MaxIdleConnsPerHost}
+	clientConfig.WrapTransport = t.defaultClientTransport
 
 	return clientConfig, nil
 }
@@ -33,23 +36,18 @@ func GetClientConfig(kubeConfigFile string, overrides *ClientConnectionOverrides
 	if err != nil {
 		return nil, err
 	}
-	// TODO after 1.11 rebase, restore this
-	//kubeConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfigBytes)
-	//if err != nil {
-	//	return nil, err
-	//}
-	config, err := clientcmd.Load(kubeConfigBytes)
+	kubeConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfigBytes)
 	if err != nil {
 		return nil, err
 	}
-	kubeConfig := clientcmd.NewNonInteractiveClientConfig(*config, "", &clientcmd.ConfigOverrides{}, nil)
-
 	clientConfig, err := kubeConfig.ClientConfig()
 	if err != nil {
 		return nil, err
 	}
 	applyClientConnectionOverrides(overrides, clientConfig)
-	clientConfig.WrapTransport = defaultClientTransport
+
+	t := clientTransportOverrides{maxIdleConnsPerHost: overrides.MaxIdleConnsPerHost}
+	clientConfig.WrapTransport = t.defaultClientTransport
 
 	return clientConfig, nil
 }
@@ -82,22 +80,28 @@ func applyClientConnectionOverrides(overrides *ClientConnectionOverrides, kubeCo
 	}
 }
 
+type clientTransportOverrides struct {
+	maxIdleConnsPerHost int
+}
+
 // defaultClientTransport sets defaults for a client Transport that are suitable for use by infrastructure components.
-func defaultClientTransport(rt http.RoundTripper) http.RoundTripper {
+func (c *clientTransportOverrides) defaultClientTransport(rt http.RoundTripper) http.RoundTripper {
 	transport, ok := rt.(*http.Transport)
 	if !ok {
 		return rt
 	}
 
-	// TODO: this should be configured by the caller, not in this method.
-	dialer := &net.Dialer{
+	transport.DialContext = (&net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
-	}
-	transport.Dial = dialer.Dial
+	}).DialContext
+
 	// Hold open more internal idle connections
-	// TODO: this should be configured by the caller, not in this method.
 	transport.MaxIdleConnsPerHost = 100
+	if c.maxIdleConnsPerHost > 0 {
+		transport.MaxIdleConnsPerHost = c.maxIdleConnsPerHost
+	}
+
 	return transport
 }
 
@@ -115,4 +119,8 @@ type ClientConnectionOverrides struct {
 	QPS float32
 	// Burst allows extra queries to accumulate when a client is exceeding its rate.
 	Burst int32
+
+	// MaxIdleConnsPerHost, if non-zero, controls the maximum idle (keep-alive) connections to keep per-host:port.
+	// If zero, DefaultMaxIdleConnsPerHost is used.
+	MaxIdleConnsPerHost int
 }
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -31,9 +31,17 @@ type StartFunc func(*ControllerContext) error
 
 type ControllerContext struct {
 	ComponentConfig *unstructured.Unstructured
-	KubeConfig      *rest.Config
-	EventRecorder   events.Recorder
-	Context         context.Context
+
+	// KubeConfig provides the REST config with no content type (it will default to JSON).
+	// Use this config for CR resources.
+	KubeConfig *rest.Config
+
+	// ProtoKubeConfig provides the REST config with "application/vnd.kubernetes.protobuf,application/json" content type.
+	// Note that this config might not be safe for CR resources, instead it should be used for other resources.
+	ProtoKubeConfig *rest.Config
+
+	EventRecorder events.Recorder
+	Context       context.Context
 }
 
 // defaultObserverInterval specifies the default interval that file observer will do rehash the files it watches and react to any changes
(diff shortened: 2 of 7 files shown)


pkg/config/client/client_config.go
```diff
@@ -33,23 +36,18 @@ func GetClientConfig(kubeConfigFile string, overrides *ClientConnectionOverrides
 	if err != nil {
 		return nil, err
 	}
-	// TODO after 1.11 rebase, restore this
-	//kubeConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfigBytes)
-	//if err != nil {
-	//	return nil, err
-	//}
-	config, err := clientcmd.Load(kubeConfigBytes)
+	kubeConfig, err := clientcmd.NewClientConfigFromBytes(kubeConfigBytes)
```
mfojtik: @soltysh PTAL

pkg/config/client/client_config.go
```diff
@@ -82,22 +80,27 @@ func applyClientConnectionOverrides(overrides *ClientConnectionOverrides, kubeCo
...
-	dialer := &net.Dialer{
+	transport.DialContext = (&net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
-	}
-	transport.Dial = dialer.Dial
+	}).DialContext
+
 	// Hold open more internal idle connections
-	// TODO: this should be configured by the caller, not in this method.
```
mfojtik: @deads2k callers can configure via overrides now, ptal

pkg/controller/controllercmd/builder.go
```diff
@@ -31,9 +32,20 @@ type StartFunc func(*ControllerContext) error
...
-	EventRecorder   events.Recorder
-	Context         context.Context
+
+	// KubeConfig provides the REST config with no content type (it will default to JSON).
+	// Use this config for CR resources.
+	KubeConfig *rest.Config
+
+	// ProtoKubeConfig provides the REST config with "application/vnd.kubernetes.protobuf,application/json" content type.
+	// Note that this config might not be safe for CR resources, instead it should be used for other resources.
+	ProtoKubeConfig *rest.Config
```
mfojtik: @deads2k the `GetKubeConfigOrInClusterConfig` seems to already have some logic to prefer protobuf via overrides if the `Accept` is not set.

However, in prometheus we see the operator making a lot of requests for secrets/configs via JSON, which might suggest the default is set to JSON in client-go?

In any case, I like to separate proto vs. json kube configs at least until @smarterclayton fix problems with CRs.

sttts: afaik default is still json in client-go. Each binary has to override that manually.

pkg/controller/controllercmd/builder.go
```diff
@@ -40,8 +41,11 @@ type ControllerContext struct {
 	// Note that this config might not be safe for CR resources, instead it should be used for other resources.
 	ProtoKubeConfig *rest.Config
 
+	// EventRecorder is used to record events in controllers.
 	EventRecorder events.Recorder
-	Context       context.Context
+
+	// Done returns a channel which will close on termination.
+	Done func() <-chan struct{}
```
sttts: make the channel a private field and add a method?

pkg/config/client/client_config.go
```diff
@@ -21,8 +21,11 @@ func GetKubeConfigOrInClusterConfig(kubeConfigFile string, overrides *ClientConn
 	if err != nil {
 		return nil, err
 	}
+
 	applyClientConnectionOverrides(overrides, clientConfig)
-	clientConfig.WrapTransport = defaultClientTransport
+
+	t := clientTransportOverrides{maxIdleConnsPerHost: overrides.MaxIdleConnsPerHost}
```
sttts: why this line? why not assign directly?

pkg/config/client/client_config.go
```diff
@@ -33,23 +36,18 @@ func GetClientConfig(kubeConfigFile string, overrides *ClientConnectionOverrides
...
-	kubeConfig := clientcmd.NewNonInteractiveClientConfig(*config, "", &clientcmd.ConfigOverrides{}, nil)
-
 	clientConfig, err := kubeConfig.ClientConfig()
 	if err != nil {
 		return nil, err
 	}
 	applyClientConnectionOverrides(overrides, clientConfig)
-	clientConfig.WrapTransport = defaultClientTransport
+
+	t := clientTransportOverrides{maxIdleConnsPerHost: overrides.MaxIdleConnsPerHost}
```
sttts: same here, why this `t`?

pkg/config/client/client_config.go
sttts: ic, defaultClientTransport is a method. I prefer the functional style `func defaultClientTransport(maxIdleConnsPerHost) func (...) {...}`. But @deads2k hates it. Nvmd.

pkg/controller/controllercmd/builder.go
mfojtik: sure, refactored.

mfojtik: as we discussed, this is not perfect and we should probably pass the context as the first arg in RunControllers, but i don't want to groom the scope for this PR.

pkg/config/client/client_config.go
mfojtik: funny, i had using functional style and I hated it because it was pretty hard to read ;-)

pkg/controller/controllercmd/builder.go
mfojtik: @sttts in that case this good change then. the callers can pick which client they prefer depending on what resources are they going to deal with.

pkg/config/client/client_config.go
sttts: I am used to scheme, haskell, scala and friends. Every method looks hard to read ;-)

pkg/config/client/client_config.go
```diff
@@ -82,22 +80,27 @@ func applyClientConnectionOverrides(overrides *ClientConnectionOverrides, kubeCo
...
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
-	}
-	transport.Dial = dialer.Dial
+	}).DialContext
+
 	// Hold open more internal idle connections
-	// TODO: this should be configured by the caller, not in this method.
-	transport.MaxIdleConnsPerHost = 100
+	if transport.MaxIdleConnsPerHost = c.maxIdleConnsPerHost; transport.MaxIdleConnsPerHost == 0 {
```
sttts: that's an odd style to write an assignment. Usually, the assignee is temporary.

pkg/config/client/client_config.go
```diff
@@ -115,4 +118,8 @@ type ClientConnectionOverrides struct {
 	QPS float32
 	// Burst allows extra queries to accumulate when a client is exceeding its rate.
 	Burst int32
+
+	// MaxIdleConnsPerHost, if non-zero, controls the maximum idle (keep-alive) connections to keep per-host.
```
sttts: per-host+port

mfojtik: @sttts copy&paste from upstream godoc, but sure :)

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -34,7 +34,9 @@ type ApplyResult struct {
 	Error   error
 }
 
-func ApplyDirectly(client kubernetes.Interface, recorder events.Recorder, manifests AssetFunc, files ...string) []ApplyResult {
+// ApplyDirectly applies the given manifest files to API server.
+// NOTE: Do not pass CR to this function as it will break if the client is using protobuf.
```
sttts: isn't the switch excluding them anyway?

mfojtik: @sttts wrong wording, I wanted to discourage people to add CR inside that switch.

pkg/config/client/client_config.go
soltysh: lgtm

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -53,27 +54,28 @@ func ApplyDirectly(client kubernetes.Interface, recorder events.Recorder, manife
 		}
 		result.Type = fmt.Sprintf("%T", requiredObj)
 
+		// NOTE: Do not add CR resources into this switch otherwise the protobuf client can cause problems.
```
soltysh: or explicitly use `kubeClient` instead

mfojtik: i don't think we ever want to add CR's here, I think this message is good enough to scare people from doing it

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -34,7 +34,8 @@ type ApplyResult struct {
 	Error   error
 }
 
-func ApplyDirectly(client kubernetes.Interface, recorder events.Recorder, manifests AssetFunc, files ...string) []ApplyResult {
+// ApplyDirectly applies the given manifest files to API server.
+func ApplyDirectly(protoKubeClient kubernetes.Interface, recorder events.Recorder, manifests AssetFunc, files ...string) []ApplyResult {
```
deads2k: no rename please.  We don't care which kind of serialization this client uses.

pkg/operator/staticpod/controller/backingresource/backing_resource_controller.go
```diff
@@ -47,8 +47,8 @@ type BackingResourceController struct {
 	// queue only ever has one item, but it has nice error handling backoff/retry semantics
 	queue workqueue.RateLimitingInterface
 
-	kubeClient    kubernetes.Interface
-	eventRecorder events.Recorder
+	protoKubeClient kubernetes.Interface
```
deads2k: no rename

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -60,7 +61,8 @@ type InstallerController struct {
 
 	operatorConfigClient v1helpers.StaticPodOperatorClient
 
-	kubeClient kubernetes.Interface
+	configMapsGetter corev1client.ConfigMapsGetter
```
deads2k: good update

mfojtik:
Fixed the renames, tagging per slack LGTM from @deads2k

PR #235, opened 2019-02-12
fix version reporting in status controller

deads2k:
we need to report version updates during updates, not just create.

/assign @mfojtik


deads2k:
memory

/retest

mfojtik:
seems like i need to add even more memory....

/retest

PR #238, opened 2019-02-13
add events for why a revision happens

deads2k:
This is just fixing debt from before we had an event recorder

/assign @mfojtik


pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -97,7 +97,7 @@ func (c RevisionController) createRevisionIfNeeded(operatorSpec *operatorv1.Stat
 	}
 
 	nextRevision := latestRevision + 1
-	glog.Infof("new revision %d triggered by %q", nextRevision, reason)
+	c.eventRecorder.Eventf("RevisionNeeded", "new revision %d triggered by %q", nextRevision, reason)
```
mfojtik: nit: s/RevisionNeeded/RevisionTriggered/

deads2k:
fixed

PR #240, opened 2019-02-13
add reasons about why a cert was rotated

deads2k:
We need this to figure out why a particular cert is getting rotated. We know that it is, but we don't know why.

/assign @mfojtik


deads2k:
this worked well.

PR #242, opened 2019-02-13
use the resourceapply methods in the cert rotation

deads2k:
Our resourceapply methods emit events and handle the update/create logic.  This updates the cert rotator to use it.

/assign @mfojtik

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 )
 
 // CABundleRotation maintains a CA bundle config map, but adding new CA certs and removing expired old ones.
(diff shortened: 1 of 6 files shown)


pkg/operator/certrotation/cabundle.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+
```
mfojtik: nit: order

pkg/operator/certrotation/signer.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
```
mfojtik: nit: same nit.

pkg/operator/certrotation/target.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
```
mfojtik: we really should work on our own import formatter :-)

mfojtik:
LGTM, good changes, fix the import order and feel free to tag.

PR #244, opened 2019-02-18
resourceapply: log changes when objects are modified

mfojtik:
This change will cause logging all changes to resources that we apply updates to. We already log diffs with CR's but we don't see why we changing things like pods/role bindings/etc.

I think this will improve our ability to debug changes done by operators and identify changes that are not desired which is worth the extra `DeepCopy()` ?

Test bump: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/160

### DIFF
--- a/pkg/operator/resource/resourceapply/apiextensions.go
+++ b/pkg/operator/resource/resourceapply/apiextensions.go
@@ -1,6 +1,8 @@
 package resourceapply
 
 import (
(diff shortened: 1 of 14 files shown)


mfojtik:
/hold

will add more changes

mfojtik:
/hold cancel

Change the status controller to spit json patch diff instead of object diff (it reads better and it is more compact).

PR #248, opened 2019-02-18
add methods for rendering a graph describing operator behavior

deads2k:
This is based on the work done in kube-apiserver.  We want this for our other control-plane operators at the very least.

/assign @mfojtik


deads2k:
oom'ed maybe?

/retest

PR #261, opened 2019-02-20
operator: stop managing on unknown management state

mfojtik:
When the management state value is set to `foo`, we should stop managing and additionally emit a warning. We don't want to emit a warning per controller loop, so I added one to target config reconciler in each operator.

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -135,10 +135,13 @@ func (c *ResourceSyncController) sync() error {
 	}
 
 	switch operatorSpec.ManagementState {
+	case operatorv1.Managed:
 	case operatorv1.Unmanaged:
 		return nil
 	case operatorv1.Removed:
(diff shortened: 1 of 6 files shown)


pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -135,10 +135,13 @@ func (c *ResourceSyncController) sync() error {
...
 
 	switch operatorSpec.ManagementState {
+	case operatorv1.Managed:
 	case operatorv1.Unmanaged:
 		return nil
 	case operatorv1.Removed:
-		// TODO: Should we try to actively remove the resources created by this controller here?
+		// TODO probably just fail
+		return nil
+	default:
```
sttts: no warning?

mfojtik: Warning is here https://github.com/openshift/cluster-openshift-apiserver-operator/pull/147

I don't want to spam events/logs (it will create same warning for every controller loop ;-)

PR #266, opened 2019-02-20
make functions for easy adoption of rendered certificates

deads2k:
This will let us adopt the certs from https://github.com/openshift/installer/pull/1232 into the "normal" rotator.  We need to do this the kubelet client cert.

proven in https://github.com/openshift/cluster-kube-apiserver-operator/pull/276

/assign @sttts

### DIFF
--- a/pkg/assets/template.go
+++ b/pkg/assets/template.go
@@ -5,12 +5,18 @@ import (
 	"encoding/base64"
 	"strings"
 	"text/template"
+	"time"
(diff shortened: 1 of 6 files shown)


deads2k:
/hold

until I have proof of rendering.

deads2k:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/276 looks good.

PR #267, opened 2019-02-20
eliminate the race to read the serving certs in operator filepath change detection

deads2k:
We knew we had a theoretical race, but we didn't expect to see it happen in any practical timeframe.  This is a good candidate fix for https://bugzilla.redhat.com/show_bug.cgi?id=1678929 which appears to happen sometimes.  We just force the initial values to hash.

/hold

hold for proof in another pull.

### DIFF
--- a/pkg/controller/controllercmd/cmd.go
+++ b/pkg/controller/controllercmd/cmd.go
@@ -1,6 +1,7 @@
 package controllercmd
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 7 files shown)


pkg/controller/controllercmd/cmd.go
```diff
@@ -146,10 +153,23 @@ func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
...
+			if err := servingCert.WriteCertConfigFile(config.ServingInfo.CertFile, config.ServingInfo.KeyFile); err != nil {
+				return err
+			}
+			crtContent := &bytes.Buffer{}
+			keyContent := &bytes.Buffer{}
+			if err := servingCert.WriteCertConfig(crtContent, keyContent); err != nil {
+				return err
+			}
+
+			// If we generate our own certificates, then we want to specify empty content to avoid a starting race.  This way,
```
bparees: comment says "specify empty content" but this doesn't look like empty content, what am i missing?

deads2k: > comment says "specify empty content" but this doesn't look like empty content, what am i missing?

stale comments.  clearly I should not write comments :)

bparees:
@adambkaplan we need to make sure the openshift controller manager operator picks up a bump w/ this change after it merges.  (my other bump PR does not include this).

deads2k:
it's no worse anyway.

PR #268, opened 2019-02-21
assets: add more sense to creater verbose messages

mfojtik:
1) in case the asset already exists, it is skipped (not created again), but user still see:

```
Feb 21 09:54:23 ip-10-0-13-88 bootkube.sh[13308]: Creating /v1, Resource=namespaces ...
Feb 21 09:54:23 ip-10-0-13-88 bootkube.sh[13308]: Creating /v1, Resource=namespaces ...
Feb 21 09:54:23 ip-10-0-13-88 bootkube.sh[13308]: Creating /v1, Resource=namespaces ...
```

2) in case of error, the user will see summary of errors at the end, but we should print error immediately in verbose mode

3) move the created message when the resource is really created


mfojtik:
(fixed message in unit tests, re-tagging)

mfojtik:
/retest

memory

PR #270, opened 2019-02-21
allow hostnames to be specified and changed after initial creation

deads2k:
This updates rotation to have server hostnames settable after the fact using a function and provides a way to wire a channel to tickle the controller when there are changes

/hold

holding until I proof of working

/assign @sttts


deads2k:
/hold cancel

works locally, approach verified, pushing forward.

PR #273, opened 2019-02-22
Installer pod fixes

mfojtik:
Fixes this case: https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.0/4546

Also switch the installer client to protobuf when fetching secrets/configs.

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 
 	"github.com/openshift/library-go/pkg/config/client"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -302,8 +302,17 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
 
 func (o *InstallOptions) Run(ctx context.Context) error {
-	eventTarget, err := events.GetControllerReferenceForCurrentPod(o.KubeClient, o.Namespace, nil)
-	if err != nil {
+	var eventTarget *corev1.ObjectReference
+
+	// Poll for the pod here to prevent flakes when the API server is temporary down or connection refused errors.
+	if err := utilwait.PollImmediateUntil(200*time.Millisecond, func() (bool, error) {
+		var getReferenceErr error
+		eventTarget, getReferenceErr = events.GetControllerReferenceForCurrentPod(o.KubeClient, o.Namespace, nil)
```
sttts: why not err

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -302,8 +302,17 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
+		if getReferenceErr != nil {
```
sttts: are there any errors that we want to pass through? NotFound maybe.

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: don't like shadowing

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: sure

PR #276, opened 2019-02-25
revision: skip syncing configs and secrets that were already created

mfojtik:
This is a bit complex to explain but:

1) Check https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-4.0/4659/artifacts/release-e2e-aws/pods - you will see `openshift-kube-apiserver_installer-2-ip-10-0-131-91.ec2.internal_installer.log.gz` which indicates "success" for installing the revision `2`, including all secrets/configmaps copied successfully.

2) However, you won't see the other 2 master nodes there, because the installer failed on the second node with:
```
F0224 06:44:04.845704       1 cmd.go:84] failed to copy: configmaps \"kubelet-serving-ca-2\" not found\n
```

3) When I `cat operator.log| grep 'RevisionTriggered' | grep -o 'new revision .* triggered'`: 
```
new revision 2 triggered
new revision 2 triggered
new revision 3 triggered
new revision 3 triggered
new revision 3 triggered
new revision 4 triggered
new revision 4 triggered
new revision 5 triggered
new revision 6 triggered
new revision 7 triggered
new revision 8 triggered
```

This suggests that the revision 2, 3 and 4 were triggered twice. It means, we ran `createNewRevision` twice for them. The first run caused the `kubelet-serving-ca-2` to be created and the installer started on the first node and succeeded. The second run seems to remove `kubelet-serving-ca-2` inside `SyncConfigMap` as the "source" for this config map was deleted (?).

So this change, will prevent deletion of resources we created using the content when we observed the content the first time and we saw the content exists. It will make sure, we don't delete the content when we already committed the revision and fired the installer pods.

### DIFF
--- a/pkg/operator/staticpod/controller/revision/revision_controller.go
+++ b/pkg/operator/staticpod/controller/revision/revision_controller.go
@@ -205,6 +205,12 @@ func (c RevisionController) createNewRevision(revision int32) error {
 	}}
 
 	for _, cm := range c.configMaps {
+		// If the target config map was already created, do not remove it as there might be installer pods in flight and we don't
+		// want the installer pod to fail. The installer should finish installing the revision and if the source configmap changed
+		// the new revision will replace the old revision afterwards.
+		if _, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(nameFor(cm.Name, revision), metav1.GetOptions{}); err == nil {
+			continue
+		}
 		obj, _, err := resourceapply.SyncConfigMap(c.configMapGetter, c.eventRecorder, c.targetNamespace, cm.Name, c.targetNamespace, nameFor(cm.Name, revision), ownerRefs)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -205,6 +205,12 @@ func (c RevisionController) createNewRevision(revision int32) error {
 	}}
 
 	for _, cm := range c.configMaps {
+		// If the target config map was already created, do not remove it as there might be installer pods in flight and we don't
+		// want the installer pod to fail. The installer should finish installing the revision and if the source configmap changed
+		// the new revision will replace the old revision afterwards.
+		if _, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(nameFor(cm.Name, revision), metav1.GetOptions{}); err == nil {
```
sttts: IMO this is wrong: it leaves old data and does not deployment of new data.

mfojtik: new data should trigger new revisions, not update existing "in-flight" revisions and cause the installer pods to fail?

mfojtik:
```
$ cat operator.log | grep 'RevisionCreate' | grep -o 'Revision .* created'
Revision 1 created
Revision 2 created
Revision 3 created
Revision 4 created
Revision 5 created
Revision 6 created
Revision 7 created
```

It looks like, we wrote all 7 revisions in order, however we triggered revision 2, 3 and 4 twice once with existing configmap (the one installer pod that ran successfully and second time because the `isLatestRevisionCurrent` found the configmap was deleted... Which is very odd because `LatestAvailableRevision` is being wrote when we send `RevisionCreate` event...

Status update has built in retry on conflict mechanism, so it is unlikely we fail to update so many times we have to retry the reconciliation. However, in that case, I would probably skip the "already created" resources for the given revision?

PR #279, opened 2019-02-25
add controller for setting upgradeable conditions to false if unsupprtedConfigOverrides is set

deads2k:
proven in https://github.com/openshift/cluster-kube-apiserver-operator/pull/304

/assign @mfojtik


deads2k:
lgtm'd in proof pull

PR #282, opened 2019-02-26
resourceapply: enforce apiGroup for rolebinding subjects

mfojtik:
We don't have the apiGroup in manifest but the server will default it for `User` kind which mean our operator constantly stomping the rolebinding and removing the group...

/cc @deads2k


sttts:
/lgtm
/approve

I somehow dislike that we have to replicate the defaulting logic here. We need something better.

PR #284, opened 2019-02-26
make no matching conditions result in unknown, not success

deads2k:
If there are no conditions to union, you are unknown, not whatever your default is.


deads2k:
@enj @bparees @benjaminapetersen I think you all need this.

PR #287, opened 2019-02-27
staticpod: retry on connection errors for optional secrets

mfojtik:


pkg/operator/staticpod/installerpod/copy.go
```diff
@@ -16,54 +16,52 @@ import (
 // It will retry on API server connection errors except not found error which is fatal for non-optional secrets.
 // In case the secret is optional and we fail to get it, no error is returned and the secret returning is nil.
 func (o *InstallOptions) getSecretWithRetry(ctx context.Context, secretNamePrefix string, isOptional bool) (*v1.Secret, error) {
-	var secret *v1.Secret
-	retryErr := wait.PollImmediateUntil(200*time.Millisecond, func() (done bool, err error) {
-		secret, err = o.KubeClient.CoreV1().Secrets(o.Namespace).Get(o.nameFor(secretNamePrefix), metav1.GetOptions{})
+	var resultSecret *v1.Secret = nil
```
sttts: that's better

mfojtik: i did whatever I can to make this error proof, so if we still getting empty secret I blame client-go

pkg/operator/staticpod/installerpod/copy.go
```diff
@@ -16,54 +16,52 @@ import (
...
+			glog.Warningf("Failed to get secret %s/%s: %v (will retry)", o.Namespace, o.nameFor(secretNamePrefix), err)
+			return false, nil
 		default:
+			resultSecret = secret
 			return true, nil
 		}
 	}, ctx.Done())
-
-	return secret, retryErr
+	if retryErr != nil {
```
sttts: if clause is not needed, is it?

mfojtik: no, but i'm in paranoid mode now.

sttts: it's logic, nothing to be paranoid about ;-)

mfojtik: just wanted to make sure I never return secret when there is an error.

mfojtik: ok, removed this.

PR #289, opened 2019-02-28
retry: add package to help retrying connection errors

mfojtik:
This is something I miss in the client-go `retry` package and I think we should add it there.

Retrying on connection errors or (transient) server errors become a common problem as all operators have to talk to a restarting API server and the clients in operators need to handle retries in that case gracefully. While in the operator control loops this problem is not urgent (and it is usually resolved by reconcile), in components like installer pods or pruning pods this problem is more imminent and require some plumbing.

The retry mechanism here use exponential backoff (similar to retry on conflict) and it retry on all errors except the HTTP client errors (not found, auth, etc). The context passed to this function should specify the timeout and the user function should observe the context cancellation (if the function only do client call, it is no-op until client-go get context support).

upstream PR: https://github.com/kubernetes/kubernetes/pull/74722

### DIFF
--- a/pkg/operator/resource/retry/retry.go
+++ b/pkg/operator/resource/retry/retry.go
@@ -1,0 +1,59 @@
+package retry
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+)
+
+// ignoreConnectionErrors is a wrapper for condition function that will cause to retry on all errors like
+// connection refused, EOF, no route to host, etc. but also all 50x API server errors.
+// This wrapper will return immediately on HTTP 40x client errors and those will not be retried.
+func ignoreConnectionErrors(lastError *error, fn ConditionWithContextFunc) ConditionWithContextFunc {
+	return func(ctx context.Context) (bool, error) {
+		done, err := fn(ctx)
+		switch {
+		case done:
+			return true, err
+		case err == nil:
+			return true, nil
+		case IsHTTPClientError(err):
+			return true, err
+		default:
+			*lastError = err
+			return false, nil
+		}
+	}
+}
+
+// RetryOnConnectionErrors will take context and condition function and retry the condition function until:
+// 1) no error is returned
+// 2) a client (4xx) HTTP error is returned
+// 3) the context passed to the condition function is done
+// 4) numbers of steps in the exponential backoff are met
+// In case of 3) or 4) the error returned will be the last observed error from the condition function.
+func RetryOnConnectionErrors(ctx context.Context, fn ConditionWithContextFunc) error {
+	var lastRetryErr error
+	err := ExponentialBackoffWithContext(ctx, retry.DefaultBackoff, ignoreConnectionErrors(&lastRetryErr, fn))
+	switch err {
+	case wait.ErrWaitTimeout:
+		if lastRetryErr != nil {
+			return lastRetryErr
+		}
+		return err
+	default:
+		return err
+	}
+}
+
+// IsHTTPClientError indicates whether the error passes is an 4xx API server error (client error).
+func IsHTTPClientError(err error) bool {
+	switch t := err.(type) {
+	case errors.APIStatus:
+		return t.Status().Code >= 400 && t.Status().Code < 500
+	default:
+		return false
+	}
+}
--- a/pkg/operator/resource/retry/retry_test.go
+++ b/pkg/operator/resource/retry/retry_test.go
@@ -1,0 +1,122 @@
+package retry
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestRetryOnConnectionErrors(t *testing.T) {
+	tests := []struct {
+		name           string
+		contextTimeout time.Duration
+		jobDuration    time.Duration
+		jobError       error
+		jobDone        bool
+		evalError      func(*testing.T, error)
+		evalAttempts   func(*testing.T, int)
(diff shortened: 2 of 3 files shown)


pkg/operator/resource/retry/retry_test.go
```diff
@@ -0,0 +1,105 @@
...
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.TODO(), test.contextTimeout)
+			defer cancel()
+			attempts := 0
+			err := RetryOnConnectionErrors(ctx, func(c context.Context) error {
```
sttts: s/c/ctx/

pkg/operator/resource/retry/retry.go
```diff
@@ -0,0 +1,72 @@
...
+	errors.IsInvalid,
+	errors.IsResourceExpired,
+	errors.IsMethodNotSupported,
+	errors.IsInvalid,
+}
+
+// RetryOnErrorsWithContext will take the function and retry the function on any errors except the not retriable errors which
+// in case such error is observed this function return this error.
+// This function will stop retry when the context is done. The context is passed to user function to allow user to terminate faster.
+func RetryOnErrorsWithContext(ctx context.Context, backoff wait.Backoff, noRetriableErrorsList []isErrorFn, fn func(context.Context) error) error {
```
sttts: Would also expect a bool for `fn`, to exit early.

pkg/operator/resource/retry/retry_test.go
mfojtik: name conflict, but sure.

pkg/operator/resource/retry/retry.go
```diff
@@ -0,0 +1,77 @@
...
+// defaultNoRetryErrors is a list of client errors that should not be retries by default.
+var defaultNoRetryErrors = []isErrorFn{
+	// all 4xx errors are no retry
+	IsHTTPClientError,
+}
+
+// ignoreConnectionErrorWrapper is a wrapper for condition function that will cause to retry on all errors like
+// connection refused, EOF, no route to host, etc. but also all 50x API server errors.
+// This wrapper will return immediately on HTTP 40x client errors and those will not be retried.
+func ignoreConnectionErrorWrapper(ctx context.Context, lastError *error, fn ConditionWithContextFunc) ConditionWithContextFunc {
```
sttts: s/Wrapper/s/

pkg/operator/resource/retry/retry.go
```diff
@@ -0,0 +1,77 @@
...
+	switch t := err.(type) {
+	case errors.APIStatus:
+		return t.Status().Code >= 400 && t.Status().Code < 500
+	default:
+		return false
+	}
+}
+
+// isNoRetryError return true if the error passed appear in the no retry errors list.
+func isNoRetryError(err error, noRetryErrorsList []isErrorFn) bool {
```
sttts: either embed the default list or rename this to something generic. It's not about retry errors with this signature.

mfojtik: embedded to a wrapper... this exists just to make the switch reading better

pkg/operator/resource/retry/retry.go
```diff
@@ -0,0 +1,75 @@
...
+// defaultNoRetryErrors is a list of client errors that should not be retries by default.
+var defaultNoRetryErrors = []isErrorFn{
+	// all 4xx errors are no retry
+	IsHTTPClientError,
+}
+
+// ignoreConnectionErrors is a wrapper for condition function that will cause to retry on all errors like
+// connection refused, EOF, no route to host, etc. but also all 50x API server errors.
+// This wrapper will return immediately on HTTP 40x client errors and those will not be retried.
+func ignoreConnectionErrors(ctx context.Context, lastError *error, fn ConditionWithContextFunc) ConditionWithContextFunc {
```
sttts: ctx is not used.

pkg/operator/resource/retry/retry.go
```diff
@@ -0,0 +1,75 @@
...
+		return false
+	}
+	return func(context.Context) (bool, error) {
+		done, err := fn(ctx)
+		switch {
+		case done:
+			return true, err
+		case err == nil:
+			return true, nil
+		case isNoRetryError(err, defaultNoRetryErrors):
```
sttts: move the default list directly above into the func

sttts: I meant to drop the `defaultNoRetryErrors` param here.

mfojtik:
@sttts all green, ok merging this and wiring this to installer and pruner pods in follow up PR?

pkg/operator/resource/retry/wait.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// TODO: This should be added to k8s.io/client-go/util/retry
+
+// ConditionWithContextFunc returns true if the condition is satisfied, or an error
+// if the loop should be aborted. The context passed to condition function allow function body
+// to return faster on context.Done().
+type ConditionWithContextFunc func(ctx context.Context) (done bool, err error)
+
+// ExponentialBackoff repeats a condition check with exponential backoff and stop repeating
```
sttts: ExponentialBackoffWithContext

pkg/operator/resource/retry/wait.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// TODO: This should be added to k8s.io/client-go/util/retry
+
+// ConditionWithContextFunc returns true if the condition is satisfied, or an error
+// if the loop should be aborted. The context passed to condition function allow function body
+// to return faster on context.Done().
```
sttts: faster than context.Done().

mfojtik:
Will merge together with wiring here: https://github.com/openshift/library-go/pull/290

PR #290, opened 2019-02-28
wire updated retry on connection logic to installer pod

mfojtik:
This is something I miss in the client-go `retry` package and I think we should add it there.

Retrying on connection errors or (transient) server errors become a common problem as all operators have to talk to a restarting API server and the clients in operators need to handle retries in that case gracefully. While in the operator control loops this problem is not urgent (and it is usually resolved by reconcile), in components like installer pods or pruning pods this problem is more imminent and require some plumbing.

The retry mechanism here use exponential backoff (similar to retry on conflict) and it retry on all errors except the HTTP client errors (not found, auth, etc). The context passed to this function should specify the timeout and the user function should observe the context cancellation (if the function only do client call, it is no-op until client-go get context support).

upstream PR: [kubernetes/kubernetes#74722](https://github.com/kubernetes/kubernetes/pull/74722)

I will create a proof PR.

### DIFF
--- a/pkg/operator/resource/retry/retry.go
+++ b/pkg/operator/resource/retry/retry.go
@@ -1,0 +1,59 @@
+package retry
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+)
+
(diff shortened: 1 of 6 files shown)


mfojtik:
Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/310

PR #291, opened 2019-02-28
Compare LastFailed and Current revisions for pruner pod

damemi:
Using `TargetRevision` for the revision of the pruner pod fails when the pruner takes too long to start the pod and the TargetRevision gets set back to 0 when the install completes (successfully or unsuccessfully). However, `CurrentRevision` only updates when we get a successful install and the pruner kicks off on every install no matter the status. So this compares the max of CurrentRevision and `LastFailedRevision`, the latter which updates whenever we get a failed install.

I also added a couple logging messages that might help debugging in the future

### DIFF
--- a/pkg/operator/staticpod/controller/prune/prune_controller.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller.go
@@ -102,7 +102,7 @@ func getRevisionLimits(operatorSpec *operatorv1.StaticPodOperatorSpec) (int32, i
 }
 
 func (c *PruneController) excludedRevisionHistory(operatorStatus *operatorv1.StaticPodOperatorStatus, failedRevisionLimit, succeededRevisionLimit int32) ([]int, error) {
-	var succeededRevisionIDs, failedRevisionIDs, inProgressRevisionIDs, unknownStatusRevisionIDs []int
+	var succeededRevisions, failedRevisions, inProgressRevisions, unknownStatusRevisions []int
 
 	configMaps, err := c.configMapGetter.ConfigMaps(c.targetNamespace).List(metav1.ListOptions{})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


damemi:
cc @mfojtik

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -163,7 +164,8 @@ func (c *PruneController) excludedRevisionHistory(operatorStatus *operatorv1.Sta
 func (c *PruneController) pruneDiskResources(operatorStatus *operatorv1.StaticPodOperatorStatus, excludedIDs []int, maxEligibleRevisionID int) error {
 	// Run pruning pod on each node and pin it to that node
 	for _, nodeStatus := range operatorStatus.NodeStatuses {
-		if err := c.ensurePrunePod(nodeStatus.NodeName, maxEligibleRevisionID, excludedIDs, nodeStatus.TargetRevision); err != nil {
+		currentRevision := math.Max(float64(nodeStatus.LastFailedRevision), float64(nodeStatus.CurrentRevision))
```
sttts: define a local max func for ints.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -163,7 +164,8 @@ func (c *PruneController) excludedRevisionHistory(operatorStatus *operatorv1.Sta
 func (c *PruneController) pruneDiskResources(operatorStatus *operatorv1.StaticPodOperatorStatus, excludedIDs []int, maxEligibleRevisionID int) error {
```
sttts: what are IDs? We don't have this term. Rename them to revisions please.

mfojtik:
LGTM (please address @sttts comments)

We need an e2e test to prove the pruning is working properly, probably in 4.1

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -102,7 +103,7 @@ func getRevisionLimits(operatorSpec *operatorv1.StaticPodOperatorSpec) (int32, i
 }
 
 func (c *PruneController) excludedRevisionHistory(operatorStatus *operatorv1.StaticPodOperatorStatus, failedRevisionLimit, succeededRevisionLimit int32) ([]int, error) {
-	var succeededRevisionIDs, failedRevisionIDs, inProgressRevisionIDs, unknownStatusRevisionIDs []int
+	var succeededRevisions, failedRevisions, inProgressRevisions, unknownStatusRevisions []int
```
sttts: as much as I prefer this of the ID everywhere, please split those refactorings and the actual logic change into two commits. It's hard to review now.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -340,3 +345,7 @@ func (c *PruneController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(pruneControllerWorkQueueKey) },
 	}
 }
+
+func currentHighestRevision(a, b int32) int32 {
```
sttts: rename it to max.

And:
```
func max(a, b int32) int32 {
  if a > b {
    return a
  }
  return b
}
```

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -340,3 +345,7 @@ func (c *PruneController) eventHandler() cache.ResourceEventHandler {
...
+	return int32(math.Max(float64(a), float64(b)))
```
sttts: no need for floating point arithmetic for integers.

PR #292, opened 2019-03-01
revision: log all changes for the latest revision

mfojtik:
With this change we will log the JSON diff of the config/secret that caused the new revision rollout.
This should make us understand better the reason for the rollouts without spamming events to much.

Additionally, this will report **all** changed secrets/configs instead of reporting just one.

### DIFF
--- a/pkg/operator/staticpod/controller/revision/revision_controller.go
+++ b/pkg/operator/staticpod/controller/revision/revision_controller.go
@@ -2,6 +2,7 @@ package revision
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -153,9 +155,14 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 			existingData = existing.Data
 		}
 		if !equality.Semantic.DeepEqual(existingData, requiredData) {
-			return false, fmt.Sprintf("configmap/%s has changed", cm.Name)
+			if glog.V(4) {
+				glog.Infof("configmap %q changes for revision %d: %s", revision, resourceapply.JSONPatch(existing, required))
```
sttts: output misses the name

mfojtik: thanks, fixed.

PR #294, opened 2019-03-04
Add helpers for copy 

mfojtik:
This change just moves the logic behind copying secret/configmap content into local disk to `resourcecopy` helper so additional commands can re-use this logic without too much copy&paste.

Additionally, this adding helper that will retry getting the self-reference on object for events for callers that use context (installer pods for example).

David, I thought you will find this handy when making the cert installer command and it will reduce the boilerplate for copying.

/cc @deads2k


pkg/operator/staticpod/installerpod/copy_test.go
```diff
@@ -1,135 +0,0 @@
...
-)
-
-func addFakeSecret(name string) *v1.Secret {
-	secret := &v1.Secret{}
-	secret.Name = name
-	secret.Namespace = "test"
-	return secret
-}
-
-func TestGetSecretWithRetry(t *testing.T) {
```
deads2k: removing this instead of moving it looks suspicious.

mfojtik: i promise to add unit tests to the new helpers :-)

mfojtik: (also the retry is heavily tested already)

deads2k:
How about only extracting as far as something under static pods to start.  I think I have a slightly different vision for the model of composition.

mfojtik:
@deads2k having this under static pods makes sense (as this will be only used by static pod installers)

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"k8s.io/api/core/v1"
```
sttts: double import

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -160,64 +162,22 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
-	// This does not need timeout, instead we should fail hard when we are not able to write.
 	resourceDir := path.Join(o.ResourceDir, o.nameFor(o.PodConfigMapNamePrefix))
 	glog.Infof("Creating target resource directory %q ...", resourceDir)
 	if err := os.MkdirAll(resourceDir, 0755); err != nil {
 		return err
 	}
 
-	for _, secret := range secrets {
+	secretSources := []resourcecopy.Source{}
+	for _, secretPrefix := range append(secretPrefixes.List(), optionalSecretPrefixes.List()...) {
```
sttts: this is dangerous: https://github.com/kubernetes/kubernetes/pull/67815

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -160,64 +162,22 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
+		if optionalSecretPrefixes.Has(secretPrefix) {
```
sttts: do the two lists overlap? why not two loops?

pkg/operator/staticpod/resourcecopy/source.go
```diff
@@ -0,0 +1,45 @@
+package resourcecopy
+
+type Source interface {
```
sttts: docs

pkg/operator/staticpod/resourcecopy/source.go
```diff
@@ -0,0 +1,54 @@
+package resourcecopy
+
+// Source define a source resource to be copied. The resource must have name, namespace and indicate whether it is optional or not.
```
sttts: nit: defines

mfojtik: fixed.

pkg/operator/staticpod/resourcecopy/source.go
```diff
@@ -0,0 +1,54 @@
+package resourcecopy
+
+// Source defines a source resource to be copied. The resource must have name, namespace and indicate whether it is optional or not.
+type Source interface {
+	Name() string
+	Namespace() string
+	IsOptional() bool
+}
```
sttts: why does it need a source abstraction? It's equivalent to a struct with 3 fields.

pkg/operator/staticpod/resourcecopy/source.go
```diff
@@ -0,0 +1,54 @@
...
+func (s *DefaultSource) Namespace() string {
+	return s.namespace
+}
+
+// Namespace return the resource name (with mutation applied if mutation function is set).
+func (s *DefaultSource) Name() string {
+	if s.nameMutationFn == nil {
+		return s.name
+	}
+	return s.nameMutationFn(s.name)
```
sttts: why can a source change its name on every call?

PR #297, opened 2019-03-04
operator: ignore Unmanaged state in control plane operators

mfojtik:
Reasoning:
Making a core or control plane operator "Unmanaged" will likely lead to a bricked cluster (in case of cert rotation, the cluster will self-destruct in 5 days today). The "Unmanaged" should only be used for development or debugging and it makes sense on higher level operators, where the damage it can do can be recovered by reverting the management state to managed. For debugging core/control plane operators, people can use CVO overrides and deleting the operator deployment (not sure what debugging that can be).

Because the "unmanaged" is giving admins a supported way to brick the clusters, we should not support it.

UPDATED: `management.SetOperatorAlwaysManaged()` and `management.SetOperatorNotRemovable()` will make the operator ignore the "unmanaged" and "removed" state, so you can opt-in.

UPDATE2: Added `management.NewOperatorManagementStateController()` controller that observe the management state and report invalid/unsupported values back to cluster operator status via conditions.

/cc @bparees 
/cc @ravisantoshgudimetla 
/cc @adambkaplan 
/cc @ericavonb 

CC'ing the folks that will be affected by this change.

### DIFF
--- a/pkg/operator/management/management_state.go
+++ b/pkg/operator/management/management_state.go
@@ -1,0 +1,69 @@
+package management
+
+import (
+	"github.com/openshift/api/operator/v1"
+)
+
+var (
+	allowOperatorUnmanagedState = true
+	allowOperatorRemovedState   = true
+)
+
+// These are for unit testing
+var (
+	getAllowedOperatorUnmanaged = func() bool {
+		return allowOperatorUnmanagedState
+	}
+	getAllowedOperatorRemovedState = func() bool {
+		return allowOperatorRemovedState
+	}
+)
+
+// SetOperatorAlwaysManaged is one time choice when an operator want to opt-out from supporting the "unmanaged" state.
+// This is a case of control plane operators or operators that are required to always run otherwise the cluster will
+// get into unstable state or critical components will stop working.
+func SetOperatorAlwaysManaged() {
+	allowOperatorUnmanagedState = false
+}
+
+// SetOperatorNotRemovable is one time choice the operator author can make to indicate the operator does not support
+// removing of his operand. This makes sense for operators like kube-apiserver where removing operand will lead to a
+// bricked, non-automatically recoverable state.
+func SetOperatorNotRemovable() {
+	allowOperatorRemovedState = false
+}
+
+// IsOperatorAlwaysManaged means the operator can't be set to unmanaged state.
+func IsOperatorAlwaysManaged() bool {
+	return !getAllowedOperatorUnmanaged()
+}
+
+// IsOperatorNotRemovable means the operator can't bet set to removed state.
+func IsOperatorNotRemovable() bool {
+	return !getAllowedOperatorRemovedState()
+}
+
+func IsOperatorUnknownState(state v1.ManagementState) bool {
+	switch state {
+	case v1.Managed, v1.Removed, v1.Unmanaged:
+		return false
+	default:
+		return true
+	}
+}
+
+// IsOperatorManaged indicates whether the operator management state allows the control loop to proceed and manage the operand.
+func IsOperatorManaged(state v1.ManagementState) bool {
+	if IsOperatorAlwaysManaged() || IsOperatorNotRemovable() {
+		return true
+	}
+	switch state {
+	case v1.Managed:
+		return true
+	case v1.Removed:
+		return false
+	case v1.Unmanaged:
+		return false
+	}
+	return true
+}
--- a/pkg/operator/management/management_state_controller.go
+++ b/pkg/operator/management/management_state_controller.go
@@ -1,0 +1,138 @@
+package management
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/golang/glog"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var workQueueKey = "instance"
+
+// ManagementStateController watches changes of `managementState` field and react in case that field is set to an unsupported value.
+// As each operator can opt-out from supporting `unmanaged` or `removed` states, this controller will add failing condition when the
+// value for this field is set to this values for those operators.
+type ManagementStateController struct {
+	operatorName   string
+	operatorClient operatorv1helpers.OperatorClient
+	eventRecorder  events.Recorder
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+}
+
+func NewOperatorManagementStateController(
+	name string,
+	operatorStatusProvider operatorv1helpers.OperatorClient,
+	recorder events.Recorder,
+) *ManagementStateController {
+	c := &ManagementStateController{
+		operatorName:   name,
+		operatorClient: operatorStatusProvider,
+		eventRecorder:  recorder,
+
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ManagementStateController-"+name),
+	}
+
+	operatorStatusProvider.Informer().AddEventHandler(c.eventHandler())
+	// TODO watch clusterOperator.status changes when it moves to openshift/api
+
+	return c
+}
+
+func (c ManagementStateController) sync() error {
+	detailedSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if apierrors.IsNotFound(err) {
+		c.eventRecorder.Warningf("StatusNotFound", "Unable to determine current operator status for %s", c.operatorName)
+		return nil
+	}
+
+	cond := operatorv1.OperatorCondition{
+		Type:   "ManagementStateFailing",
+		Status: operatorv1.ConditionFalse,
+	}
+
+	if IsOperatorAlwaysManaged() && detailedSpec.ManagementState == operatorv1.Unmanaged {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "Unmanaged"
+		cond.Message = fmt.Sprintf("Unmanaged is not supported for %s operator", c.operatorName)
+	}
+
+	if IsOperatorNotRemovable() && detailedSpec.ManagementState == operatorv1.Removed {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "Removed"
+		cond.Message = fmt.Sprintf("Removed is not supported for %s operator", c.operatorName)
+	}
+
+	if IsOperatorUnknownState(detailedSpec.ManagementState) {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "Unknown"
+		cond.Message = fmt.Sprintf("Unsupported management state %q for %s operator", detailedSpec.ManagementState, c.operatorName)
+	}
+
+	if _, _, updateError := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(cond)); updateError != nil {
+		if err == nil {
+			return updateError
+		}
+	}
+
+	return nil
+}
+
+func (c *ManagementStateController) Run(workers int, stopCh <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	glog.Infof("Starting management-state-controller-" + c.operatorName)
+	defer glog.Infof("Shutting down management-state-controller-" + c.operatorName)
+
+	// doesn't matter what workers say, only start one.
+	go wait.Until(c.runWorker, time.Second, stopCh)
+
+	<-stopCh
+}
+
+func (c *ManagementStateController) runWorker() {
+	for c.processNextWorkItem() {
+	}
+}
+
+func (c *ManagementStateController) processNextWorkItem() bool {
+	dsKey, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	defer c.queue.Done(dsKey)
+
+	err := c.sync()
+	if err == nil {
+		c.queue.Forget(dsKey)
+		return true
+	}
+
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %v", dsKey, err))
+	c.queue.AddRateLimited(dsKey)
+
+	return true
+}
+
+// eventHandler queues the operator to check spec and status
+func (c *ManagementStateController) eventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(workQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(workQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
+	}
+}
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -19,6 +19,7 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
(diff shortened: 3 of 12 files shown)


pkg/operator/status/status_controller.go
```diff
@@ -113,22 +113,8 @@ func (c StatusSyncer) sync() error {
...
-
-		if equality.Semantic.DeepEqual(clusterOperatorObj, originalClusterOperatorObj) {
-			return nil
-		}
-		if _, updateErr := c.clusterOperatorClient.ClusterOperators().UpdateStatus(clusterOperatorObj); err != nil {
-			return updateErr
-		}
-		c.eventRecorder.Eventf("OperatorStatusChanged", "Status for operator %s changed: %s", c.clusterOperatorName, configv1helpers.GetStatusDiff(originalClusterOperatorObj.Status, clusterOperatorObj.Status))
-		return nil
+		c.eventRecorder.Eventf("OperatorStateInvalid", "Unmanaged state is not supported for operator %s")
```
mfojtik: @deads2k i hope this will be enough warning ?

mfojtik:
@deads2k with this in place, I wonder if we can just remove management state from the API (before we ship with it and never be able to take it back).

pkg/operator/status/status_controller.go
```diff
@@ -113,22 +113,8 @@ func (c StatusSyncer) sync() error {
...
-
-		if equality.Semantic.DeepEqual(clusterOperatorObj, originalClusterOperatorObj) {
-			return nil
-		}
-		if _, updateErr := c.clusterOperatorClient.ClusterOperators().UpdateStatus(clusterOperatorObj); err != nil {
-			return updateErr
-		}
-		c.eventRecorder.Eventf("OperatorStatusChanged", "Status for operator %s changed: %s", c.clusterOperatorName, configv1helpers.GetStatusDiff(originalClusterOperatorObj.Status, clusterOperatorObj.Status))
-		return nil
+		c.eventRecorder.Eventf("OperatorStateInvalid", "Unmanaged state is not supported for operator %s", c.clusterOperatorName)
```
mfojtik: @deads2k i hope this will be enough warning ?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -137,7 +137,6 @@ func (c *ResourceSyncController) sync() error {
 	switch operatorSpec.ManagementState {
 	case operatorv1.Managed:
 	case operatorv1.Unmanaged:
-		return nil
 	case operatorv1.Removed:
```
enj: Does `Removed` and the `default` case make sense?  In general it seems like we always want to be in "managed" state for certain operators (and not others).

enj: Also `""` needs to be tolerated better IMO since the field is optional.

mfojtik: I can't see "Remove" working for control plane :-)

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -137,7 +137,6 @@ func (c *ResourceSyncController) sync() error {
 	switch operatorSpec.ManagementState {
 	case operatorv1.Managed:
 	case operatorv1.Unmanaged:
-		return nil
```
bparees: doesn't this affect more than control plane operators?

pkg/operator/status/status_controller.go
bparees: this isn't just used for control plane operators....library-go is a utility library that any operator can use.  So how is this safe?

pkg/operator/unsupportedconfigoverridescontroller/unsupportedconfigoverrides_controller.go
```diff
@@ -67,7 +67,6 @@ func (c *UnsupportedConfigOverridesController) sync() error {
 	switch operatorSpec.ManagementState {
 	case operatorv1.Managed:
 	case operatorv1.Unmanaged:
-		return nil
```
bparees: not necessarily control plane specific.

bparees:
@mfojtik @deads2k wrote library-go as a framework for all operators, not just control plane operators, and there are several non-control plane operators using it, so I don't think these changes are safe.

mfojtik:
@bparees i can make some flag that will allow to opt-out from supporting unmanaged (so control plane will use it).

pkg/operator/resourcesynccontroller/resourcesync_controller.go
ericavonb: Should we emit a warning or something? At the least we should add debug logs around what state the operator observes here.

mfojtik: added event with warning

pkg/operator/status/status_controller.go
mfojtik: you can opt-in with your operator with `status.SetOperatorAlwaysManaged()`

pkg/operator/status/status_controller.go
```diff
@@ -131,6 +130,19 @@ func (c StatusSyncer) sync() error {
 		return nil
 	}
 
+	// When operator is not allowed to be unmanagable, but the state is set to unmanagable, prevent upgrades done by this operators (but do not prevent managing operand).
+	if IsOperatorAlwaysManaged() && detailedSpec.ManagementState == operatorv1.Unmanaged {
+		clusterOperatorObj.Status = configv1.ClusterOperatorStatus{}
+		configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, configv1.ClusterOperatorStatusCondition{Type: configv1.OperatorUpgradeable, Status: configv1.ConditionUnknown, Reason: "Unmanaged"})
+
+		if !equality.Semantic.DeepEqual(clusterOperatorObj, originalClusterOperatorObj) {
+			if _, updateErr := c.clusterOperatorClient.ClusterOperators().UpdateStatus(clusterOperatorObj); err != nil {
```
deads2k: This will be overwritten below by the union upgradeable, right?

mfojtik: good catch, missed return nil, fixed.

pkg/operator/status/management_state.go
```diff
@@ -0,0 +1,40 @@
...
+
+func IsOperatorAlwaysManaged() bool {
+	return !allowOperatorUnmanagedState
+}
+
+// IsOperatorManaged indicates whether the operator management state allows the control loop to proceed and manage
+// the operand.
+func IsOperatorManaged(state v1.ManagementState) bool {
+	switch {
+	case state == v1.Managed || IsOperatorAlwaysManaged():
```
deads2k: nit

```
if IsOperatorAlwaysManaged(){
return true
}

switch case....
`

File: pkg/operator/status/management_state.go [hunk 8]
```diff
@@ -0,0 +1,40 @@
...
+	case state == v1.Removed:
+		// If state is "removed" it means operator is still managing operand uninstall.
+		// Some operators (control plane) ignore this state.
+		return true
+	case state == v1.Unmanaged:
+		return false
+	}
+
+	// In case users set wrong value, we default to managed.
+	glog.Warningf("Unknown %q value used for operator management state", state)
```
deads2k: A real control loop will allow you to set the `ManagementStateFailing` condition consistently and it will ripple up to clusteroperator status.

mfojtik: ok, convinced, I turned this to a small controller loop observing the management state.

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	configv1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var workQueueKey = "instance"
+
+type ManagementStateController struct {
+	clusterOperatorName string
+	relatedObjects      []configv1.ObjectReference
```
deads2k: this looks naughty. What are you doing with it?

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,170 @@
...
+
+	configv1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var workQueueKey = "instance"
+
+type ManagementStateController struct {
+	clusterOperatorName string
```
deads2k: Who isn't `cluster`?

pkg/operator/status/status_controller.go
```diff
@@ -113,24 +114,6 @@ func (c StatusSyncer) sync() error {
 	}
 	clusterOperatorObj := originalClusterOperatorObj.DeepCopy()
 
-	// if we are unmanaged, force everything to Unknown.
-	if detailedSpec.ManagementState == operatorv1.Unmanaged {
```
mfojtik: @deads2k this moved to new controller.

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	return c
+}
+
+// sync reacts to a change in prereqs by finding information that is required to match another value in the cluster. This
+// must be information that is logically "owned" by another component.
+func (c ManagementStateController) sync() error {
+	detailedSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if apierrors.IsNotFound(err) {
+		c.eventRecorder.Warningf("StatusNotFound", "Unable to determine current operator status for %s", c.clusterOperatorName)
+		if err := c.clusterOperatorClient.ClusterOperators().Delete(c.clusterOperatorName, nil); err != nil && !apierrors.IsNotFound(err) {
```
deads2k: let's not delete the clusteroperator.  this was probably a msitake before

pkg/operator/management/management_state_controller.go
mfojtik: copy & pasta

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		// Wait for the status controller to create status for us
+		glog.Infof("unable to get operator status: %v", err)
+		return err
+	}
+
+	clusterOperatorObj := originalClusterOperatorObj.DeepCopy()
+
+	clusterOperatorObj.Status = configv1.ClusterOperatorStatus{}
+
```
deads2k:
```go
if IsOperatorAlwaysManaged() && detailedSpec.ManagementState == operatorv1.Unmanaged{
    setStatus(c.operatorClient, condition{"managementStateFailing", Message: "illegal state"})
}
if IsOperatorUnknownState{
    setStatus(c.operatorClient, condition{"managementStateFailing", Message: "unknown state"})
}
```

pkg/operator/management/management_state_controller.go
mfojtik: sure

pkg/operator/management/management_state_controller.go
mfojtik: fixed

pkg/operator/status/status_controller.go
deads2k: skip this block if the operator is "AlwaysOn"

mfojtik: why? lets set this in management state controller?

pkg/operator/management/management_state.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// IsOperatorNotRemovable means the operator can't bet set to removed state.
+func IsOperatorNotRemovable() bool {
+	return !getAllowedOperatorRemovedState()
+}
+
+func IsOperatorUnknownState(state v1.ManagementState) bool {
+	switch state {
+	case v1.Managed, v1.Removed, v1.Unmanaged:
```
enj: Is not `""` meant to say "give me the default" - i.e. it is a valid state that means the same as `Managed`?

pkg/operator/management/management_state.go
```diff
@@ -0,0 +1,69 @@
...
+		return true
+	}
+}
+
+// IsOperatorManaged indicates whether the operator management state allows the control loop to proceed and manage the operand.
+func IsOperatorManaged(state v1.ManagementState) bool {
+	if IsOperatorAlwaysManaged() || IsOperatorNotRemovable() {
+		return true
+	}
+	switch {
```
enj: `switch state`?

pkg/operator/management/management_state_controller.go
mfojtik: @deads2k is it worth? it will make the code more complex (I need deep equal and unmanage is setting more conditions)

pkg/operator/management/management_state.go
```diff
@@ -0,0 +1,69 @@
...
+	case state == v1.Managed:
```
enj: Same empty string question.

pkg/operator/management/management_state.go
mfojtik: there is not default and "" means unknown.

pkg/operator/management/management_state.go
mfojtik: same answer :-)

pkg/operator/management/management_state.go
mfojtik: sure

pkg/operator/management/management_state.go
enj: @deads2k I thought you wanted to support rendering an empty object as valid?

pkg/operator/status/status_controller.go
mfojtik: nvmd.

pkg/operator/management/management_state.go
mfojtik: i would assume that for `ManagementState` we would always want to have an non-zero value and the field present in operator spec and this won't be "optional" field.

enj: https://github.com/openshift/api/blob/master/operator/v1/types.go#L46-L47

mfojtik: @sttts ^

mfojtik: @enj according to @sttts that has no effect, just docs and we should remove that. 

Using https://github.com/openshift/kubernetes-sigs-controller-tools/pull/4 we can specify pattern and only allow valid values to be passed.

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,157 @@
...
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+
+	configv1helpers "github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+var workQueueKey = "instance"
+
+type ManagementStateController struct {
```
sttts: high level doc

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,157 @@
...
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ManagementStateController-"+name),
+	}
+
+	operatorStatusProvider.Informer().AddEventHandler(c.eventHandler())
+	// TODO watch clusterOperator.status changes when it moves to openshift/api
+
+	return c
+}
+
+// sync reacts to a change in prereqs by finding information that is required to match another value in the cluster. This
```
sttts: ?

mfojtik: copy & pasta from status_controller, I will refit this

pkg/operator/management/management_state.go
```diff
@@ -0,0 +1,69 @@
+package management
+
+import (
+	"github.com/openshift/api/operator/v1"
+)
+
+var (
+	allowOperatorUnmanagedState = true
```
sttts: why this global state?

mfojtik: debated via slack, basically we can plumb this trough each individual controller as a bool field in struct, which won't make it less ugly and will require everybody that use library-go to update their controllers once they bump.

pkg/operator/management/management_state.go
mfojtik: https://github.com/openshift/api/pull/232

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,158 @@
...
+			Type:   "ManagementStateFailing",
+			Status: configv1.ConditionTrue,
+			Reason: fmt.Sprintf("Unsupported management state %q for %s operator", detailedSpec.ManagementState, c.operatorName),
+		})
+	}
+
+	if equality.Semantic.DeepEqual(operatorObj, originalOperatorObj) {
+		return nil
+	}
+	if _, updateErr := c.clusterOperatorGetter.ClusterOperators().UpdateStatus(operatorObj); err != nil {
```
deads2k: by doing this, you're created a case of two writers to the clusteroperator resource.  Two writers can fight, a single writer cannot.

pkg/operator/management/management_state_controller.go
```diff
@@ -0,0 +1,142 @@
...
+
+var workQueueKey = "instance"
+
+// ManagementStateController watches changes of `managementState` field and react in case that field is set to an unsupported value.
+// As each operator can opt-out from supporting `unmanaged` or `removed` states, this controller will add failing condition when the
+// value for this field is set to this values for those operators.
+type ManagementStateController struct {
+	operatorName          string
+	operatorClient        operatorv1helpers.OperatorClient
+	clusterOperatorGetter configv1client.ClusterOperatorsGetter
```
deads2k: this can be deleted now

deads2k:
remove the cruft and lgtm

/approve

PR #299, opened 2019-03-06
crd-schema-gen: allow a *-yaml-merge-patch file to post-kubebuilder-gen patch validation schemata

sttts:
Put a `<crd>.yaml-merge-patch` merge patch (https://tools.ietf.org/html/rfc7386) next to a CRD manifest `<crd>.yaml` and it will be applied to the CRD after the kubebuilder openapi generator has been run (only `spec.validation` changes are supported for now). This is useful to customize behaviour of shared types for specific types, e.g. ObjectMeta or OperatorSpec.

### DIFF
--- a/cmd/crd-schema-gen/generator/generator.go
+++ b/cmd/crd-schema-gen/generator/generator.go
@@ -10,11 +10,13 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/evanphx/json-patch"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


sttts:
/cc @DirectXMan12

For inspiration. That use-case might also exist for kubebuilder. Wdyt?

DirectXMan12:
What's the main usecase here?  That we don't support generation of validation for out-of-tree types?

PR #301, opened 2019-03-06
events: allow to set component per controller

mfojtik:
Based on https://github.com/openshift/library-go/blob/master/vendor/k8s.io/client-go/tools/record/events_cache.go#L414 the events are correlated and that also means the "spam" is filtered out. Based on the "key" for counting the events, if we keep the component the same for entire operator and all controller loops inside, we might get rate limited and some events might be dropped as "spam".

This change will allow to set custom component per-controller which should mitigate this issue.

### DIFF
--- a/pkg/operator/events/recorder_upstream.go
+++ b/pkg/operator/events/recorder_upstream.go
@@ -13,21 +13,36 @@ import (
 
 // NewKubeRecorder returns new event recorder.
 func NewKubeRecorder(client corev1client.EventInterface, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
-	broadcaster := record.NewBroadcaster()
-	broadcaster.StartLogging(glog.Infof)
-	broadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: client})
-	eventRecorder := broadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: sourceComponentName})
-
-	return &upstreamRecorder{
-		eventRecorder:     eventRecorder,
+	return (&upstreamRecorder{
+		client:            client,
+		component:         sourceComponentName,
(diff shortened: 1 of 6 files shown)


pkg/operator/events/recorder_upstream.go
```diff
@@ -13,23 +13,38 @@ import (
...
-	eventRecorder := broadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: sourceComponentName})
-
-	return &upstreamRecorder{
-		eventRecorder:     eventRecorder,
+	return (&upstreamRecorder{
+		client:            client,
+		component:         sourceComponentName,
 		involvedObjectRef: involvedObjectRef,
-	}
+	}).ForComponent(sourceComponentName)
```
sttts: just set it

pkg/operator/events/recorder_upstream.go
```diff
@@ -13,23 +13,38 @@ import (
...
+	broadcaster       record.EventBroadcaster
 	eventRecorder     record.EventRecorder
 	involvedObjectRef *corev1.ObjectReference
 }
 
+func (r *upstreamRecorder) ForComponent(componentName string) Recorder {
+	newRecorderForComponent := *r
+	broadcaster := record.NewBroadcaster()
+	broadcaster.StartLogging(glog.Infof)
+	broadcaster.StartRecordingToSink(&corev1client.EventSinkImpl{Interface: newRecorderForComponent.client})
```
sttts: what's the cost? A go routine?

mfojtik: yeah, per recorder

pkg/operator/events/recorder_upstream.go
mfojtik: it will duplicate the code from ForComponent()

mfojtik:
(fixed govet)

PR #302, opened 2019-03-06
events: wire component name for each controller

mfojtik:
Builds on https://github.com/openshift/library-go/pull/301

This will add `-controller-name` suffix to event recorder component name for each controller we own.

### DIFF
--- a/pkg/operator/configobserver/config_observer_controller.go
+++ b/pkg/operator/configobserver/config_observer_controller.go
@@ -63,7 +63,7 @@ func NewConfigObserver(
 ) *ConfigObserver {
 	return &ConfigObserver{
 		operatorConfigClient: operatorClient,
-		eventRecorder:        eventRecorder,
+		eventRecorder:        eventRecorder.WithComponentSuffix("config-observer"),
 
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ConfigObserver"),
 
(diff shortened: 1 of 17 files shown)


pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -63,7 +63,7 @@ func NewConfigObserver(
 ) *ConfigObserver {
 	return &ConfigObserver{
 		operatorConfigClient: operatorClient,
-		eventRecorder:        eventRecorder,
+		eventRecorder:        eventRecorder.ForComponent(eventRecorder.ComponentName() + "-network-observer"),
```
sttts: is this network only?

mfojtik: good catch, fixed.

pkg/operator/unsupportedconfigoverridescontroller/unsupportedconfigoverrides_controller.go
```diff
@@ -46,7 +46,7 @@ func NewUnsupportedConfigOverridesController(
 ) *UnsupportedConfigOverridesController {
 	c := &UnsupportedConfigOverridesController{
 		operatorClient: operatorClient,
-		eventRecorder:  eventRecorder,
+		eventRecorder:  eventRecorder.ForComponent(eventRecorder.ComponentName() + "-unsupportedconfigoverrides-controller"),
```
sttts: with dashes like static-pod-state ?

PR #303, opened 2019-03-06
don't create pruner pod with revision 0

damemi:
We see errors like `E0306 19:58:57.817107       1 prune_controller.go:303] key failed with : unable to set pruner pod ownerrefs: configmap "revision-status-0" not found`, but there is no reason to try to make or prune a "revision 0" nor any reason one of our revisions should be at that number (that I know of)


mfojtik:
@damemi looks like the unit test failures are real.

damemi:
@mfojtik yeah, had some tests with currentRevision==0. updated

PR #305, opened 2019-03-12
Leader election: increase LeaseDuration,RenewDeadline,RetryPeriod

sallyom:
@smarterclayton 
This PR increases leader election default periods. So, instead of pings every 2s, it's every 20s with this change.
* **k8s default** --> **thisPR**          
* lease-duration 15s  --> 120s
* renew-deadline 10s --> 90s
* retry-period 2s --> 20s
[k8s defaults](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/)

### DIFF
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -75,13 +75,13 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace, 
 	ret := *(&config).DeepCopy()
 
 	if ret.LeaseDuration.Duration == 0 {
-		ret.LeaseDuration.Duration = 15 * time.Second
+		ret.LeaseDuration.Duration = 120 * time.Second
 	}
 	if ret.RenewDeadline.Duration == 0 {
-		ret.RenewDeadline.Duration = 10 * time.Second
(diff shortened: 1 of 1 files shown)


smarterclayton:
What components are impacted by this on vendor that might need to now set faster defaults?  Just KAO/KSM/KCM/OAS?

The default should be tuned for the vast majority of operators, which this seems correct.

That said, 150s is long, let's make the numbers:

20s retry-period, 90s renew deadline, 120s lease duration

Lease duration and renew deadline are a measure of clock skew in the cluster and generally we have no need to keep their ratios the same.

sallyom:
I've modified based on your comment, and I'm not sure which operators will need to modify their defaults. apiservers, scheduler, and controller-manager sound right
@enj ?

enj:
also service-ca operator and console operator.  machine approver does not use this code today.

enj:
There is no distinction for this code in between an operator and a controller though.

smarterclayton:
There are only a small subset of core controllers that need aggressive election.  Everyone else can be a bit more relaxed.

smarterclayton:
Please make sure impacted controllers are aware if they want more aggressive defaults they are responsible for setting them (and also communicating any rules of thumb into docs in library-go)

PR #307, opened 2019-03-14
Use CRD kind and group as the key

soltysh:
/assign @sttts

### DIFF
--- a/cmd/crd-schema-gen/generator/generator.go
+++ b/cmd/crd-schema-gen/generator/generator.go
@@ -67,7 +67,7 @@ func Run() error {
 		return err
 	}
 
-	// generate kubebuilder NamedYaml manifests into temp dir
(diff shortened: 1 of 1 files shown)


cmd/crd-schema-gen/generator/generator.go
```diff
@@ -343,7 +343,8 @@ func crdsFromDirectory(dir string) (map[string]NamedYaml, error) {
 			fmt.Printf("Warning: failed to unmarshal %q, skipping\n", info.Name())
 			continue
 		}
-		ret[filepath.Join(dir, info.Name())] = NamedYaml{crd.Name, y}
+		key := crd.Spec.Names.Kind + "." + crd.Spec.Group
+		ret[filepath.Join(dir, info.Name())] = NamedYaml{key, y}
```
sttts: change NamedYaml above as well: s/Name/KindGroup/

PR #309, opened 2019-03-17
web profiler should take implicit dependency on net/http/pprof

smarterclayton:
Otherwise components that don't pull in net/http/pprof can't use
it automatically.

### DIFF
--- a/pkg/serviceability/profiler.go
+++ b/pkg/serviceability/profiler.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"runtime"
 
+	_ "net/http/pprof" // include the default Go profiler mux
+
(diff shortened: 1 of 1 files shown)


pkg/serviceability/profiler.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 	"runtime"
 
+	_ "net/http/pprof" // include the default Go profiler mux
```
sttts: hmm, does this even have any effect? We don't use the default mux I believe.

sttts: nvmd, default server is started below. lgtm

PR #310, opened 2019-03-18
certrotation: don't start rotation with less than 30 min cert validity

sttts:
Before this PR the minDeploymentDuration was 20% of the cert validity. With 30+ days validity, this is a lot, far too much to be on the safe side, and surprising for the admin.

### DIFF
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -69,7 +69,7 @@ func needNewSigningCertKeyPair(annotations map[string]string, refresh time.Durat
 		return reason
 	}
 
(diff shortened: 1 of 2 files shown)


sttts:
I misunderstood the logic in the cert rotation controller. These 20% were not to avoid late rotation, but to enforce it.

Background: we wait for 10% validity of the (potentially new CA) before we update old certs after their refresh time. If for whatever reason we had no chance to update a cert (maybe because the cluster was shut down or we are restoring from a backup or the operator was down), a newly created CA might not be 10% old, but a cert might be at 80% or higher. This 20% extra check was there to enforce the cert update nevertheless in that case.

PR #312, opened 2019-03-19
installer: do no report errors for nodes that are not failing

mfojtik:
`0 nodes are failing on revision 6:\nNodeInstallerFailing: static pod has been installed, but is not ready while new revision is pending` indicates we report errors for nodes that are not actually failing (their failing count == 0).


bparees:
@mfojtik who do you expect to review/merge this?

PR #315, opened 2019-03-19
certs: add helpers to unify cert format

mfojtik:
We use this in must-gather and we should make the output consistent across all tools.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/certs/util.go
+++ b/pkg/certs/util.go
@@ -1,0 +1,55 @@
+package certs
+
+import (
+	"crypto/x509"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// CertificateToString parse provided certificate to human-readable form.
+// This function should guarantee consistent output format for must-gather tooling and any code
+// that prints the certificate details.
+func CertificateToString(certificate *x509.Certificate) string {
+	humanName := certificate.Subject.CommonName
+	signerHumanName := certificate.Issuer.CommonName
+
+	if certificate.Subject.CommonName == certificate.Issuer.CommonName {
+		signerHumanName = "<self-signed>"
+	}
+
+	usages := []string{}
+	for _, curr := range certificate.ExtKeyUsage {
+		if curr == x509.ExtKeyUsageClientAuth {
+			usages = append(usages, "client")
+			continue
+		}
+		if curr == x509.ExtKeyUsageServerAuth {
+			usages = append(usages, "serving")
+			continue
+		}
+
+		usages = append(usages, fmt.Sprintf("%d", curr))
+	}
+
+	validServingNames := []string{}
+	for _, ip := range certificate.IPAddresses {
+		validServingNames = append(validServingNames, ip.String())
+	}
+	for _, dnsName := range certificate.DNSNames {
+		validServingNames = append(validServingNames, dnsName)
+	}
+
+	servingString := ""
+	if len(validServingNames) > 0 {
+		servingString = fmt.Sprintf(" validServingFor=[%s]", strings.Join(validServingNames, ","))
+	}
+
+	groupString := ""
+	if len(certificate.Subject.Organization) > 0 {
+		groupString = fmt.Sprintf(" groups=[%s]", strings.Join(certificate.Subject.Organization, ","))
+	}
+
+	return fmt.Sprintf("%q [%s]%s%s issuer=%q (%v to %v (now=%v))", humanName, strings.Join(usages, ","), groupString,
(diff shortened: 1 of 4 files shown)


pkg/certs/util.go
```diff
@@ -0,0 +1,55 @@
...
+	"crypto/x509"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// CertificateToString parse provided certificate to human-readable form.
+// This function should guarantee consistent output format for must-gather tooling and any code
+// that prints the certificate details.
+func CertificateToString(certificate *x509.Certificate) string {
```
sttts: a unit test with examples?

deads2k: bundles are really common, I suspect you'll want a helper for that too

mfojtik: yeah, tired of writing `for` loops... good idea.

pkg/operator/certrotation/cabundle.go
```diff
@@ -52,6 +55,8 @@ func (c CABundleRotation) ensureConfigMapCABundle(signingCertKeyPair *crypto.CA)
 		if err != nil {
 			return nil, err
 		}
+		glog.V(2).Infof("Updated ca-bundle.crt configmap %s/%s with:\n%s", certs.CertificateBundleToString(updatedCerts), caBundleConfigMap.Namespace, caBundleConfigMap.Name)
```
mfojtik: @deads2k @sttts do you think this is a good spot to print out the bundle? i hope this will allow is to inspect what we had/have in the bundle during e2e runs.

pkg/certs/util_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestCertificateToString(t *testing.T) {
+	tests := []struct {
+		name     string
+		cert     *x509.Certificate
+		expected string
+	}{
+		{
+			name:     "empty cert",
+			cert:     &x509.Certificate{},
+			expected: `"" [] issuer="<self-signed>" (0001-01-01 00:00:00 +0000 UTC to 0001-01-01 00:00:00 +0000 UTC (now=2019-01-01 00:00:00 +0000 UTC))`,
```
sttts: can we remove the timeshift and timezone?

mfojtik: i'm not sure why @deads2k added this (i'm just copying his code)

pkg/operator/certrotation/cabundle.go
sttts: make it dependent on the second return value of ApplyConfigMap

mfojtik: yes, updated.

pkg/certs/util.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+)
+
+const defaultOutputTimeFormat = "Jan 2 15:04:05 2006"
+
+var nowFn = func() time.Time {
+	return time.Now().UTC()
+}
+
+// CertificateToString convert a certificate into a human readable string.
```
sttts: nit: converts

pkg/certs/util.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"crypto/x509"
+	"fmt"
+	"strings"
+	"time"
+)
+
+const defaultOutputTimeFormat = "Jan 2 15:04:05 2006"
+
+var nowFn = func() time.Time {
```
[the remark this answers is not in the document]
sttts: `nowFn = time.Now().UTC`

I guess this is used for test overrides?

pkg/certs/util.go
```diff
@@ -0,0 +1,71 @@
...
+	return fmt.Sprintf("%q [%s]%s%s issuer=%q (%v to %v (now=%v))", humanName, strings.Join(usages, ","), groupString,
+		servingString, signerHumanName, certificate.NotBefore.UTC().Format(defaultOutputTimeFormat),
+		certificate.NotAfter.UTC().Format(defaultOutputTimeFormat), nowFn().Format(defaultOutputTimeFormat))
+}
+
+// CertificateBundleToString convert a certificate bundle into a human readable string.
+func CertificateBundleToString(bundle []*x509.Certificate) string {
+	output := []string{}
+	for i, cert := range bundle {
+		output = append(output, fmt.Sprintf("[#%d]: %s", i, CertificateToString(cert)))
```
sttts: you seem to like `[ ]` 😀

mfojtik: yeah :-)

pkg/certs/util.go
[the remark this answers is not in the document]
mfojtik: yes, added comment.

PR #317, opened 2019-03-20
Correctly run UnsupportedConfigOverridesController when Managed

enj:

@openshift/sig-master


pkg/operator/unsupportedconfigoverridescontroller/unsupportedconfigoverrides_controller.go
```diff
@@ -65,7 +65,7 @@ func (c *UnsupportedConfigOverridesController) sync() error {
 		return err
 	}
 
-	if management.IsOperatorManaged(operatorSpec.ManagementState) {
+	if !management.IsOperatorManaged(operatorSpec.ManagementState) {
```
mfojtik: @enj good catch!

PR #318, opened 2019-03-22
Concat condition reason to it's type on condition union

jhadvig:
Per discussion with @deads2k 

/assign @deads2k

### DIFF
--- a/pkg/operator/status/condition.go
+++ b/pkg/operator/status/condition.go
@@ -95,6 +95,9 @@ func unionMessage(conditions []operatorv1.OperatorCondition) string {
 
 func unionReason(conditions []operatorv1.OperatorCondition) string {
 	if len(conditions) == 1 {
+		if len(conditions[0].Reason) != 0 {
+			return conditions[0].Type + conditions[0].Reason
+		}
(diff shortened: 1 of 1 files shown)


pkg/operator/status/condition.go
```diff
@@ -95,6 +95,9 @@ func unionMessage(conditions []operatorv1.OperatorCondition) string {
 
 func unionReason(conditions []operatorv1.OperatorCondition) string {
 	if len(conditions) == 1 {
+		if len(conditions[0].Reason) != 0 {
+			return fmt.Sprintf("%s: %s", conditions[0].Type, conditions[0].Reason)
+		}
 		return conditions[0].Type
```
deads2k: `return conditions[0].Type+conditions[0].Reason` I htink

jhadvig: Thought we want it in the same format as https://github.com/openshift/library-go/blob/master/pkg/operator/status/condition.go#L90
Not sure if just joining Type and Reason without any additional punctuation right.

jhadvig:
@deads2k PR updated

PR #319, opened 2019-03-25
ClusterOperatorStatusController: Wait for cache sync

sallyom:
Without this PR, during upgrades, components that utilize the ClusterOperatorStatusController
(authentication and service-ca) have a glitch in reporting cluster operator version. This PR adds a wait for the operatorClient informer to catch up.

(testing now, this fixes the glitch in version reporting for auth, service-ca COs)

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -45,6 +45,7 @@ type StatusSyncer struct {
 	operatorClient        operatorv1helpers.OperatorClient
 	clusterOperatorClient configv1client.ClusterOperatorsGetter
 	clusterOperatorLister configv1listers.ClusterOperatorLister
+	clusterOperatorSynced cache.InformerSynced
 	eventRecorder         events.Recorder
 
 	// queue only ever has one item, but it has nice error handling backoff/retry semantics
@@ -66,6 +67,7 @@ func NewClusterOperatorStatusController(
 		versionGetter:         versionGetter,
 		clusterOperatorClient: clusterOperatorClient,
 		clusterOperatorLister: clusterOperatorInformer.Lister(),
+		clusterOperatorSynced: clusterOperatorInformer.Informer().HasSynced,
 		operatorClient:        operatorClient,
 		eventRecorder:         recorder.WithComponentSuffix("status-controller"),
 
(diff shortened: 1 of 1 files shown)


pkg/operator/status/status_controller.go
```diff
@@ -83,10 +83,16 @@ func NewClusterOperatorStatusController(
 func (c StatusSyncer) sync() error {
 	detailedSpec, currentDetailedStatus, _, err := c.operatorClient.GetOperatorState()
 	if apierrors.IsNotFound(err) {
-		c.eventRecorder.Warningf("StatusNotFound", "Unable to determine current operator status for %s", c.clusterOperatorName)
-		if err := c.clusterOperatorClient.ClusterOperators().Delete(c.clusterOperatorName, nil); err != nil && !apierrors.IsNotFound(err) {
```
enj: While I am not sure this delete makes sense in general, I believe we just need to wait for informer.HasSynced before attempting to use it.

sallyom: I did not know of this HasSynced, looking.

sallyom: I've updated to add the HasSynced, testing now.

sttts: If operator state is not found, does this mean the ClusterOperator object will stay with invalid contents?

sallyom: I'll look closer, test with this.  It should be fine to keep the delete ifNotFound, as long as we waitForCacheSync.  What I see now is the statusController is up, checks for serviceca CO, doesn't find it bc the informer is not caught up,  that causes a glitch for a few seconds in version reporting (while it recreates?).  The glitch is most likely fixed just by adding the waitForCacheSync though.  I'll report back.

sallyom: is there a reason one would _not_ want the waitForCacheSync? I could be missing something.  : )

sallyom: @sttts @soltysh I confirmed with an upgrade that the WaitForCacheSync resolves the issue I'm seeing with service-ca, authentication COs (last I checked they are only operators other than kube-apiserer making use of this ClusterOperatorStatusController).  I've restored the Delete ifNotFound, I'm still not sure it's necessary but it's not causing the glitch, it's the COInformer not being synced that is.  : )

sallyom:
@soltysh after this merges, I'll open a PR to the temporary 1-12 branch, that's what cluster-auth-operator is pinning (temporarily). @sttts ready for a merge? : )

PR #321, opened 2019-03-26
Add ResourceSync controller debug endpoint

damemi:
Adds an http handler to serve resource sync rules info

See https://github.com/openshift/cluster-openshift-apiserver-operator/pull/172

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -44,6 +44,9 @@ type ControllerContext struct {
 	// EventRecorder is used to record events in controllers.
 	EventRecorder events.Recorder
 
+	// Server is the GenericAPIServer serving healthz checks and debug info
+	Server *genericapiserver.GenericAPIServer
+
 	stopChan <-chan struct{}
 }
 
@@ -182,33 +185,31 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, ctx context.C
 		}
 	}
 
-	switch {
-	case b.servingInfo == nil && len(b.healthChecks) > 0:
-		return fmt.Errorf("healthchecks without server config won't work")
-
-	default:
-		kubeConfig := ""
-		if b.kubeAPIServerConfigFile != nil {
-			kubeConfig = *b.kubeAPIServerConfigFile
-		}
-		serverConfig, err := serving.ToServerConfig(*b.servingInfo, *b.authenticationConfig, *b.authorizationConfig, kubeConfig)
-		if err != nil {
-			return err
-		}
-		serverConfig.HealthzChecks = append(serverConfig.HealthzChecks, b.healthChecks...)
-
-		server, err := serverConfig.Complete(nil).New(b.componentName, genericapiserver.NewEmptyDelegate())
-		if err != nil {
-			return err
-		}
-
-		go func() {
-			if err := server.PrepareRun().Run(ctx.Done()); err != nil {
-				glog.Error(err)
-			}
-			glog.Fatal("server exited")
-		}()
+	if b.servingInfo == nil {
+		return fmt.Errorf("server config required for health checks and debugging endpoints")
 	}
+
+	kubeConfig := ""
+	if b.kubeAPIServerConfigFile != nil {
+		kubeConfig = *b.kubeAPIServerConfigFile
+	}
+	serverConfig, err := serving.ToServerConfig(*b.servingInfo, *b.authenticationConfig, *b.authorizationConfig, kubeConfig)
+	if err != nil {
+		return err
+	}
+	serverConfig.HealthzChecks = append(serverConfig.HealthzChecks, b.healthChecks...)
+
+	server, err := serverConfig.Complete(nil).New(b.componentName, genericapiserver.NewEmptyDelegate())
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		if err := server.PrepareRun().Run(ctx.Done()); err != nil {
+			glog.Error(err)
+		}
+		glog.Fatal("server exited")
+	}()
 
 	protoConfig := rest.CopyConfig(clientConfig)
 	protoConfig.AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json"
@@ -219,6 +220,7 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, ctx context.C
 		KubeConfig:      clientConfig,
 		ProtoKubeConfig: protoConfig,
 		EventRecorder:   eventRecorder,
+		Server:          server,
 		stopChan:        ctx.Done(),
 	}
 
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -1,12 +1,15 @@
 package resourcesynccontroller
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/http"
+	"sort"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/golang/glog"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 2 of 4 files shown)


pkg/controller/controllercmd/builder.go
```diff
@@ -197,7 +201,7 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, ctx context.C
 		}
 		serverConfig.HealthzChecks = append(serverConfig.HealthzChecks, b.healthChecks...)
 
-		server, err := serverConfig.Complete(nil).New(b.componentName, genericapiserver.NewEmptyDelegate())
+		server, err = serverConfig.Complete(nil).New(b.componentName, genericapiserver.NewEmptyDelegate())
```
sttts: turn the first case into an if clause instead.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,24 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(controllerWorkQueueKey) },
 	}
 }
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (c *ResourceSyncController) ServeSyncRules(w http.ResponseWriter, req *http.Request) {
+	if req.URL.Path != "/syncRules" {
```
sttts: these is a mux for that

pkg/controller/controllercmd/builder.go
```diff
@@ -182,34 +185,32 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, ctx context.C
 		}
 	}
 
-	switch {
-	case b.servingInfo == nil && len(b.healthChecks) > 0:
+	if b.servingInfo == nil && len(b.healthChecks) > 0 {
```
deads2k: I think just require serving info overall.  I see no reason to have an option without it.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,19 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
 		DeleteFunc: func(obj interface{}) { c.queue.Add(controllerWorkQueueKey) },
 	}
 }
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (c *ResourceSyncController) ServeSyncRules(w http.ResponseWriter, req *http.Request) {
+	syncRules := make(map[string]map[ResourceLocation]ResourceLocation, 2)
+	syncRules["secrets"] = c.secretSyncRules
```
deads2k: references to these require a lock, right?

deads2k:
I prefer a construct similar to the GC one where a NewResourceSyncer gives you the control loop *and* the debughandler which is added opaquely.  Such a construct will wire nicely into the starter.go

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,30 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+
+type debugHTTPHandler struct {
+	controller *ResourceSyncController
+}
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (h *debugHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	syncRules := make(map[string]map[ResourceLocation]ResourceLocation, 2)
+
+	h.controller.syncRuleLock.Lock()
```
deads2k: RLock and RUnlock

deads2k:
/lgtm
/hold

hold for correct lock, then self-fix-up the tags

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,47 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+}
+
+type ResourceSyncRule struct {
+	Source      ResourceLocation `json:"source"`
+	Destination ResourceLocation `json:"destination"`
+}
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (h *debugHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	syncRules := make(map[string][]ResourceSyncRule, 2)
```
deads2k: make this a struct

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,47 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+
+	h.controller.syncRuleLock.RLock()
+	defer h.controller.syncRuleLock.RUnlock()
+	syncRules["secrets"] = mappedSyncRules(h.controller.secretSyncRules)
```
deads2k: struct, not map

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,47 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+	controller *ResourceSyncController
+}
+
+type ResourceSyncRule struct {
+	Source      ResourceLocation `json:"source"`
+	Destination ResourceLocation `json:"destination"`
+}
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (h *debugHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
deads2k: if this failed (I think it did), unit test

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,47 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+func NewDebugHandler(controller *ResourceSyncController) http.Handler {
+	return &debugHTTPHandler{controller: controller}
+}
+
+type debugHTTPHandler struct {
+	controller *ResourceSyncController
+}
+
+type ResourceSyncRule struct {
+	Source      ResourceLocation `json:"source"`
```
deads2k: add json tags to ResourceLocation to serialize lowercase

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,53 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+	Destination ResourceLocation `json:"destination"`
+}
+
+// ServeSyncRules provides a handler function to return the sync rules of the controller
+func (h *debugHTTPHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	syncRules := struct {
```
deads2k: don't use anonymous structs outside of tests.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +244,53 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+	if err != nil {
+		w.Write([]byte(err.Error()))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Write(data)
+	w.WriteHeader(http.StatusOK)
+}
+
+func mappedSyncRules(syncRules map[ResourceLocation]ResourceLocation) []ResourceSyncRule {
```
deads2k: we should have a stable serialization, so sort the keys by namespace then by name.  Use a "normal" sort function.

pkg/operator/resourcesynccontroller/resourcesync_controller_test.go
```diff
@@ -111,3 +113,30 @@ func TestSyncConfigMap(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestServeHTTP(t *testing.T) {
+	c := &ResourceSyncController{
+		secretSyncRules: map[ResourceLocation]ResourceLocation{
+			{Namespace: "foo", Name: "cat"}: {Namespace: "bar", Name: "dog"},
```
deads2k: looks like your test needs multiple rules so that you would catch the unstable serialization.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +245,67 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+	Destination ResourceLocation `json:"destination"`
+}
+
+type ResourceSyncRuleList []ResourceSyncRule
+
+func (l ResourceSyncRuleList) Len() int      { return len(l) }
+func (l ResourceSyncRuleList) Swap(i, j int) { l[i], l[j] = l[j], l[i] }
+func (l ResourceSyncRuleList) Less(i, j int) bool {
+	if sortorder.NaturalLess(l[i].Source.Namespace, l[j].Source.Namespace) {
+		return true
```
deads2k: why not false?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +245,67 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+	Source      ResourceLocation `json:"source"`
+	Destination ResourceLocation `json:"destination"`
+}
+
+type ResourceSyncRuleList []ResourceSyncRule
+
+func (l ResourceSyncRuleList) Len() int      { return len(l) }
+func (l ResourceSyncRuleList) Swap(i, j int) { l[i], l[j] = l[j], l[i] }
+func (l ResourceSyncRuleList) Less(i, j int) bool {
+	if sortorder.NaturalLess(l[i].Source.Namespace, l[j].Source.Namespace) {
```
deads2k: strings.compare and avoid the dep

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -242,3 +245,67 @@ func (c *ResourceSyncController) eventHandler() cache.ResourceEventHandler {
...
+		return true
+	}
+	if sortorder.NaturalLess(l[i].Source.Namespace, l[j].Source.Namespace) {
```
deads2k: repeat?

pkg/operator/resourcesynccontroller/resourcesync_controller_test.go
```diff
@@ -111,3 +113,34 @@ func TestSyncConfigMap(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestServeHTTP(t *testing.T) {
+	c := &ResourceSyncController{
+		secretSyncRules: map[ResourceLocation]ResourceLocation{
+			{Namespace: "foo", Name: "cat"}: {Namespace: "bar", Name: "cat"},
+			{Namespace: "foo", Name: "dog"}: {Namespace: "bar", Name: "dog"},
```
deads2k: this test didn't catch the namespace comparison problem.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
damemi: `Less` should return true if `i < j` right?

PR #322, opened 2019-03-28
Get unnsuported cert rotation config reading

soltysh:
This was requested by David in https://github.com/openshift/cluster-kube-apiserver-operator/pull/359#issuecomment-477623400

/assign @deads2k

### DIFF
--- a/pkg/operator/certrotation/config.go
+++ b/pkg/operator/certrotation/config.go
@@ -1,0 +1,40 @@
+package certrotation
+
+import (
+	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+// GetCertRotationScale  The normal scale is based on a day.  The value returned by this function
+// is used to scale rotation durations instead of a day, so you can set it shorter.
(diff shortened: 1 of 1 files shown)


pkg/operator/unsupportedcertrotation/config.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+// GetCertRotationBase is helper method to read the unsupported cert rotation configuration
```
deads2k:
```
// GetCertRotationScale  The normal scale is based on a day.  The value returned by this function is used to scale rotation durations instead of a day, so you can set it shorter.

```

pkg/operator/unsupportedcertrotation/config.go
```diff
@@ -0,0 +1,36 @@
...
+			if err != nil {
+				return false, err
+			}
+		}
+		return true, nil
+	})
+	if err != nil {
+		return 0, err
+	}
+	return certRotationBase, nil
```
deads2k: fail if the value is longer than a day.

pkg/operator/unsupportedcertrotation/config.go
```diff
@@ -0,0 +1,36 @@
+package unsupportedcertrotation
```
deads2k: put in the normal rotation package.

PR #323, opened 2019-03-29
Add prometheus metrics for rotated certificates

mfojtik:
This change will label all secrets and configmaps we rotate, collect certificate metadata from them and publish them to prometheus as:

`certificates_ca_bundle_expire_hours`
`certificates_signer_expire_hours`
`certificates_target_expire_hours`

The gauge values indicates number of hours until each certificate expire. This should help us better monitor the certificate expiration in live cluster and add alerts when certificates are about to expire without replacement.

Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/360

/cc @soltysh 
/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/certs"
 	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/certrotation/metrics"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 )
@@ -39,6 +40,7 @@ func (c CABundleRotation) ensureConfigMapCABundle(signingCertKeyPair *crypto.CA)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
+
 	caBundleConfigMap := originalCABundleConfigMap.DeepCopy()
 	if apierrors.IsNotFound(err) {
 		// create an empty one
@@ -50,6 +52,7 @@ func (c CABundleRotation) ensureConfigMapCABundle(signingCertKeyPair *crypto.CA)
 	}
 	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
 		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert", c.Name, c.Namespace)
+		metrics.LabelAsManagedConfigMap(caBundleConfigMap, metrics.CertificateTypeCABundle)
 
 		actualCABundleConfigMap, modified, err := resourceapply.ApplyConfigMap(c.Client, c.EventRecorder, caBundleConfigMap)
 		if err != nil {
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -7,11 +7,13 @@ import (
 	"k8s.io/klog"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+
+	"github.com/openshift/library-go/pkg/operator/certrotation/metrics"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 const (
@@ -128,6 +130,9 @@ func (c *CertRotationController) Run(workers int, stopCh <-chan struct{}) {
 		utilruntime.HandleError(fmt.Errorf("caches did not sync"))
 		return
 	}
+
+	// register prometheus metrics for rotated certificates
+	metrics.Register(c.CABundleRotation.Informer.Lister(), c.SigningRotation.Informer.Lister())
 
 	// doesn't matter what workers say, only start one.
 	go wait.Until(c.runWorker, time.Second, stopCh)
--- a/pkg/operator/certrotation/metrics/monitor.go
+++ b/pkg/operator/certrotation/metrics/monitor.go
@@ -1,0 +1,187 @@
+package metrics
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+	"k8s.io/klog"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+const (
+	// ManagedCertificateTypeLabelName marks config map or secret as object that contains managed certificates.
+	// This groups all objects that store certs and allow easy query to get them all.
+	// Also any object with this label will be collected by prometheus and transformed to metrics.
+	// The value of this label should be set to "true".
+	ManagedCertificateTypeLabelName = "auth.openshift.io/managed-certificate-type"
+)
+
+type CertificateType string
+
+var (
+	CertificateTypeCABundle CertificateType = "ca-bundle"
+	CertificateTypeSigner   CertificateType = "signer"
+	CertificateTypeTarget   CertificateType = "target"
+	CertificateTypeUnknown  CertificateType = "unknown"
+
+	timeNowFn = time.Now
+)
+
+var (
+	caBundleExpireHoursDesc = prometheus.NewDesc(
+		"certificates_ca_bundle_expire_hours",
+		"Number of hours until certificates in given CA bundle expire",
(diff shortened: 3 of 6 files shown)


pkg/operator/certmonitor/util.go
```diff
@@ -0,0 +1,29 @@
+package certmonitor
+
+import (
+	"k8s.io/api/core/v1"
+)
+
+func SetCertificateMonitoredConfigMap(config *v1.ConfigMap, certificateType CertificateType) {
```
soltysh: Why you want to make them public, if the certrotation is the only one using them let's make those private and live inside certrotation controller?

mrogers950: If left public, service-ca could use these since ATM it does not utilize the certrotation controller for its certs.

soltysh: Is there any reason service-ca can't use the certrotation mechanism, I'd be more interested in aligning the functionality rather than multiplying options.

mrogers950: Although it looks like that would need a public metrics collector controller as well. @mfojtik would it be possible to make it so service-ca can use this for cert metrics outside of a certrotation controller?

mrogers950: > Is there any reason service-ca can't use the certrotation mechanism, I'd be more interested in aligning the functionality rather than multiplying options.

I'll look and see if there might be any problems using it. The service-ca operation is a single CA with N number of server certs, with no client certs. So, I'm not sure if the usage pattern is an exact fit.

soltysh: Honestly I'd prefer we share as much code as possible rather than having similarities in two places. It's better for all of us if we have the same foundation that we're working on :smiley:

mfojtik: There is nothing preventing people for using this to report their cert metrics if they follow the annotations/labels.

mfojtik:
As a follow up, when we do bump we can add this label manually to installer provided certs.

pkg/operator/certmonitor/monitor.go
```diff
@@ -0,0 +1,196 @@
...
+
+const (
+	// CertificateManagedLabelName marks config map or secret as object that contains managed certificates.
+	// This groups all objects that store certs and allow easy query to get them all.
+	// Also any object with this label will be collected by prometheus and transformed to metrics.
+	// The value of this label should be set to "true".
+	CertificateManagedLabelName = "auth.openshift.io/certificate-managed"
+
+	// CertificateType indicates what kind of certificate the object contain. Can be of of "signer", "target" or "ca-bundle".
+	CertificateTypeAnnotationName = "auth.openshift.io/certificate-type"
```
deads2k: seems like the presence of this implies the one above

soltysh: True, but you can't have selectors over labels.

mfojtik: can you make label selector match everything just having the label regardless of its value?

deads2k: Yeah, I think so.  have a look a the syntax

pkg/operator/certmonitor/monitor.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		signingCertKeyPair, err := crypto.GetCAFromBytes(secret.Data["tls.crt"], secret.Data["tls.key"])
+		if err != nil {
+			continue
+		}
+
+		var targetDescType *prometheus.Desc
+
+		switch GetCertificateTypeFromObject(secret) {
```
deads2k: do this before your expensive parse

pkg/operator/certrotation/cabundle.go
```diff
@@ -39,17 +40,20 @@ func (c CABundleRotation) ensureConfigMapCABundle(signingCertKeyPair *crypto.CA)
...
+	} else {
+		caBundleConfigMap = originalCABundleConfigMap.DeepCopy()
 	}
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
 	if err != nil {
 		return nil, err
 	}
 	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
 		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert", c.Name, c.Namespace)
+		certmonitor.LabelAsManagedConfigMap(caBundleConfigMap, certmonitor.CertificateTypeCABundle)
```
deads2k: don't make this depend on its monitoring package. invert it.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -129,6 +130,9 @@ func (c *CertRotationController) Run(workers int, stopCh <-chan struct{}) {
 		return
 	}
 
+	// register prometheus metrics for rotated certificates
+	certmonitor.Register(c.CABundleRotation.Informer.Lister(), c.SigningRotation.Informer.Lister())
```
deads2k: oh, I see.

pkg/operator/certmonitor/monitor.go
```diff
@@ -0,0 +1,196 @@
+package certmonitor
```
deads2k: if it's going to be tightly coupled, move it as a subpackage of the main controller

deads2k:
fairly minor comments.  This doesn't tell you who uses what though.

pkg/operator/certrotation/cabundle.go
```diff
@@ -39,17 +40,20 @@ func (c CABundleRotation) ensureConfigMapCABundle(signingCertKeyPair *crypto.CA)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
-	caBundleConfigMap := originalCABundleConfigMap.DeepCopy()
```
deads2k: let's just keep this line. it avoids special cases.

pkg/operator/certrotation/metrics/monitor.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+const (
+	// CertificateManagedLabelName marks config map or secret as object that contains managed certificates.
+	// This groups all objects that store certs and allow easy query to get them all.
+	// Also any object with this label will be collected by prometheus and transformed to metrics.
+	// The value of this label should be set to "true".
+	CertificateManagedLabelName = "auth.openshift.io/certificate-managed"
```
deads2k: remove this.  You can query on exists: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement

pkg/operator/certrotation/metrics/monitor.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// CertificateType indicates what kind of certificate the object contain. Can be of of "signer", "target" or "ca-bundle".
+	CertificateTypeAnnotationName = "auth.openshift.io/certificate-type"
```
deads2k: `/managed-certificate-type` I'd think.

pkg/operator/certrotation/cabundle.go
mfojtik: that line just moved... why deepcopy when we not found and we don't need to deep copy?

pkg/operator/certmonitor/monitor.go
mfojtik: yeah, labels.Parse() and `cert-type in (foo, bar)`

pkg/operator/certrotation/cabundle.go
deads2k: > that line just moved... why deepcopy when we not found and we don't need to deep copy?

to avoid an unnecessary branching condition

pkg/operator/certrotation/metrics/monitor.go
```diff
@@ -0,0 +1,187 @@
...
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+const (
+	// CertificateManagedTypeLabelName marks config map or secret as object that contains managed certificates.
+	// This groups all objects that store certs and allow easy query to get them all.
+	// Also any object with this label will be collected by prometheus and transformed to metrics.
+	// The value of this label should be set to "true".
+	CertificateManagedTypeLabelName = "auth.openshift.io/certificate-managed-type"
```
deads2k: I think these are `managed-certificates`, so the prefix would be related to managed-certificates for sorting.  As distinct from serving-certificate

pkg/operator/certrotation/metrics/monitor.go
```diff
@@ -0,0 +1,187 @@
...
+			expireHours := certificate.NotAfter.UTC().Sub(c.nowFn().UTC()).Hours()
+			labelValues = append(labelValues, []string{
+				secret.Namespace,
+				secret.Name,
+				certificate.Subject.CommonName,
+				certificate.Issuer.CommonName,
+				fmt.Sprintf("%s", certificate.NotBefore.UTC()),
+			}...)
+
+			ch <- prometheus.MustNewConstMetric(
```
deads2k: I don't actually know what this is doing.  You sure this isn't a loop?

mfojtik: we do the same in deployment configs metrics, we publishing single cert to prometheus

mfojtik:
proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/380

will unhold this if the tests will be green

PR #328, opened 2019-04-04
Remove artificial cert rotation scale limitation on unsupported config

tnozicka:
We already limit the available choices on the supported config. There is no reason to limit the unsupported choices where we already claim it UNSUPPORTED** to force **artificial** limitations on users agreeing to the terms of being unsupported and changing it on their own.

**The supported config is where we express our opinions on what the setting should be**. The unsupported config is where user or developer expresses his opinions and gets to do stuff that may not work but only because there is reason to, not because a developer though he is smarter.

I think the above is a sufficient reason on it's own but:

This could easily unblock the whole CodeReady Containers (used to be CDK) efforts in minutes time by letting them override the value to a large duration for now.

It also let developers to change the value to test cert rotation.

/assign @deads2k 

(fyi @soltysh since I know we really disagree on this topic, but don't want to go behind your back in secret)

### DIFF
--- a/pkg/operator/certrotation/config.go
+++ b/pkg/operator/certrotation/config.go
@@ -1,7 +1,6 @@
 package certrotation
 
 import (
-	"fmt"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -32,9 +31,6 @@ func GetCertRotationScale(client kubernetes.Interface, namespace string) (time.D
 	})
 	if err != nil {
 		return 0, err
-	}
-	if certRotationScale > 24*time.Hour {
-		return 0, fmt.Errorf("scale longer than 24h is not allowed: %v", certRotationScale)
 	}
 	return certRotationScale, nil
 }


mfojtik:
nack, we are very opinionated about cert rotation configuration and we won't allow users to extended the intervals even as unsupported feature. we are going to allow shorten the intervals artificially to support our CI and QE so we can make sure that rotation works as it should.

Users that want to pause their clusters for longer than 31 days, should have a tool to revive that cluster when they are booted up. That script should replace expired certificates and allow the cluster to restore and then the cluster should rotate using default, baked in cadence.

tnozicka:
> nack, we are very opinionated about cert rotation configuration and we won't allow users to extended t

We are opinionated and our opinion is expressed as "no config for it" (means 30 days) and we are properly limiting that by not even having an option for it, meaning 30 days.

Limiting the unsupported config where you are changing the value on your own risk is purely artificial.

deads2k:
> Limiting the unsupported config where you are changing the value on your own risk is purely artificial.

If we make it longer, someone will use it.  We do not want anyone to use this.

/close

LalatenduMohanty:
> . you can ONLY be allowed to shorten the default cert rotation which is 30 days and for that 24h is a reasonable maximum to ensure this value is not overused.

@soltysh @mfojtik There are two challenged for CodeReady Containers (CRC) team. Before going there, let me explain how we build CRC as it will help you guys to understand it better. 


Currently we run installer with modified config (for making it work in single node master) and then package the disk image for users , so that users can just start the VM and get a working single node OpenShift 4.0 working for them.  So the current build process is as follows

1. Run the installer
2. Wait for 24 hours, so the first cert rotation happens. 
3. Then package the disk image for CRC. Now the disk image might be valid for 31 days from the day of build. Otherwise the disk images will be not usable for users 

We want to fix step-1 and step-3 from above as the build process can not take more than 24hours and the cluster should be valid for at least 90days. 


> we are going to allow shorten the intervals artificially to support our CI and QE so we can make sure that rotation works as it should.

@mfojtik how this can achieved? (or how soon) As this will greatly help us as we do not need to wait for 24 hours for the build process.  This is at this point is a bigger blocker for us.

PR #331, opened 2019-04-08
Allow controllercmd to reuse default config and rotation logic

smarterclayton:
A controller that needs to slightly tweak the builder (to provide
either extra API serving, health checks, or alter authorization)
must currently clone the StartController() method. Instead, split
config and default rotation into their own methods to make it
it easier to turn the default command into config, then start
default rotation.

@deads2k when I tried to reuse had to copy the rotation stuff
to tweak the builder

### DIFF
--- a/pkg/controller/controllercmd/cmd.go
+++ b/pkg/controller/controllercmd/cmd.go
@@ -13,6 +13,7 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/version"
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold

Doesn't do exactly what I need.

smarterclayton:
/hold cancel

Had a bug, I was calling SetDefaultRotation after the builder, but SetDefaultRotation actually defaults the config when dealing with self-signed certs.  New order is correct.

pkg/controller/controllercmd/cmd.go
```diff
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
deads2k: nit order

pkg/controller/controllercmd/cmd.go
```diff
@@ -10,6 +10,8 @@ import (
...
+
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
```
deads2k: since you're here, nit, order

pkg/controller/controllercmd/cmd.go
```diff
@@ -99,10 +91,25 @@ func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
...
+	if _, err := os.Stat(filepath.Join(certDir, "tls.crt")); os.IsNotExist(err) {
+		return false
+	}
+	if _, err := os.Stat(filepath.Join(certDir, "tls.key")); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+// StartWithRotation starts the provided builder with the default rotation set (config + serving info). Use StartController if
```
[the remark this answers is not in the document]
deads2k: outstanding

pkg/controller/controllercmd/cmd.go
```diff
@@ -76,21 +78,11 @@ func (c *ControllerCommandConfig) NewCommand() *cobra.Command {
...
-	}
-	if _, err := os.Stat(filepath.Join(certDir, "tls.key")); os.IsNotExist(err) {
-		return false
-	}
-	return true
-}
-
-// StartController runs the controller
-func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
+// Config returns the configuration of this command. Use StartController if you don't need to customize the default operator.
```
deads2k: indicate whether this mutates method receiver

pkg/controller/controllercmd/cmd.go
```diff
@@ -99,10 +91,25 @@ func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
...
+	}
+	if _, err := os.Stat(filepath.Join(certDir, "tls.key")); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+// StartWithRotation starts the provided builder with the default rotation set (config + serving info). Use StartController if
+// you do not need to customize the controller builder.
+func (c *ControllerCommandConfig) SetDefaultRotation(config *operatorv1alpha1.GenericOperatorConfig, configContent []byte) (map[string][]byte, []string, error) {
```
deads2k: indicate whether this mutates method receiver.  reading top to bottom it is unclear.  The method is called `Set`, but returns values.

deads2k: `ObservedFiles`?

deads2k:
Minor comments.  The names are actually confusing and I think for a Setting that doesn't set, the doc is important.

lgtm otherwise, self-tag after fixing.

pkg/controller/controllercmd/cmd.go
smarterclayton: How about AddDefaultRotationToConfig?

PR #335, opened 2019-04-09
events: fallback to use namespace as controller if we are not able to get pod

mfojtik:
With this change, when we are unable to get the pod running the operator or we are not able to guess the pod by listing all pods in namespace, instead of failing hard use the provided namespace as a controller.

This will prevent operator failure to start in some cases.

/cc @deads2k

### DIFF
--- a/pkg/operator/events/recorder.go
+++ b/pkg/operator/events/recorder.go
@@ -68,7 +68,6 @@ func GetControllerReferenceForCurrentPod(client kubernetes.Interface, targetName
 	case "Pod":
 		pod, err := client.CoreV1().Pods(reference.Namespace).Get(reference.Name, metav1.GetOptions{})
 		if err != nil {
-			klog.Warningf("Unable to get owner reference for pod %q (will use namespace instead): %v", podNameEnvFunc(), err)
 			return getControllerReferenceForNamespace(reference.Namespace), err
 		}
 		if podController := metav1.GetControllerOf(pod); podController != nil {
@@ -79,7 +78,6 @@ func GetControllerReferenceForCurrentPod(client kubernetes.Interface, targetName
 	case "ReplicaSet":
 		rs, err := client.AppsV1().ReplicaSets(reference.Namespace).Get(reference.Name, metav1.GetOptions{})
 		if err != nil {
-			klog.Warningf("Unable to get owner reference for replica set %q (will use namespace instead): %v", reference.Name, err)
 			return getControllerReferenceForNamespace(reference.Namespace), err
 		}
 		if rsController := metav1.GetControllerOf(rs); rsController != nil {
--- a/pkg/operator/events/recorder.go
+++ b/pkg/operator/events/recorder.go
@@ -44,6 +44,8 @@ var podNameEnvFunc = func() string {
 
 // GetControllerReferenceForCurrentPod provides an object reference to a controller managing the pod/container where this process runs.
 // The pod name must be provided via the POD_NAME name.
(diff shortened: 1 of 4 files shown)


pkg/operator/events/recorder.go
```diff
@@ -54,7 +54,11 @@ func GetControllerReferenceForCurrentPod(client kubernetes.Interface, targetName
...
 		reference, err := guessControllerReferenceForNamespace(client.CoreV1().Pods(targetNamespace))
 		if err != nil {
-			return nil, err
+			// If this fails, do not give up with error but instead use the namespace as controller reference for the pod
+			// NOTE: This is last resort, if we see this often it might indicate something is wrong in the cluster.
+			//       In some cases this might help with flakes.
+			klog.Warningf("Unable to guess owner reference in namespace %q (will use namespace instead): %v", targetNamespace, err)
+			return getControllerReferenceForNamespace(targetNamespace)
 		}
 		return GetControllerReferenceForCurrentPod(client, targetNamespace, reference)
```
deads2k: change signature of this method to never fail or return the error and a valid reference.

smarterclayton: I hate having to set /var/run/.../namespace to test locally.  It would be better if that is absent to report to openshift-config or something.

mfojtik: @deads2k ok, I changed the signature to return error but also fallback reference to namespace

@smarterclayton you don't have to create any file, the namespace is provided as argument to the function, so it can be `foo`.

mfojtik:
@deads2k updated installer pod to retry on connection errors (refused/etc.) but if it fails accept the namespace as a fallback instead of dying hard.

pkg/operator/events/recorder.go
```diff
@@ -44,6 +44,8 @@ var podNameEnvFunc = func() string {
 
 // GetControllerReferenceForCurrentPod provides an object reference to a controller managing the pod/container where this process runs.
 // The pod name must be provided via the POD_NAME name.
+// Even if this method returns an error, it always return valid reference to the namespace. It allows the callers to control the logging
+// and decide to fail or accept the namespace.
 func GetControllerReferenceForCurrentPod(client kubernetes.Interface, targetNamespace string, reference *corev1.ObjectReference) (*corev1.ObjectReference, error) {
```
deads2k: need to update callsites to log error and continue instead of panicing.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -336,14 +336,15 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 func (o *InstallOptions) Run(ctx context.Context) error {
 	var eventTarget *corev1.ObjectReference
 
-	if err := retry.RetryOnConnectionErrors(ctx, func(context.Context) (bool, error) {
+	err := retry.RetryOnConnectionErrors(ctx, func(context.Context) (bool, error) {
 		var clientErr error
 		eventTarget, clientErr = events.GetControllerReferenceForCurrentPod(o.KubeClient, o.Namespace, nil)
 		if clientErr != nil {
 			return false, clientErr
```
deads2k: you shouldn't fail on this

pkg/operator/events/recorder.go
mfojtik: updated builder and installer pod, my goland don't see any more call sites in this repo.

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: i'm trying to get it a chance to get real reference in case this is really connection error...

PR #336, opened 2019-04-09
Match dependency levels with operators

mfojtik:
See the glide commit for details. This brings us on pair with operator repos.


glide.yaml
```diff
@@ -6,7 +6,8 @@ import:
 - package: k8s.io/apiserver
   version: kubernetes-1.13.4
 - package: k8s.io/apiextensions-apiserver
-  version: kubernetes-1.13.4
+  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
+  version: 6e2b3eb1b401beaf836f640f73f730f6da7aff2f # pinned to get nullable patch
```
deads2k: I would track a branch instead.  Also, why do this when the ugly should be hidden in the new repo?

glide.yaml
```diff
@@ -18,10 +19,10 @@ import:
 
 # crd-schema-gen
 - package: sigs.k8s.io/controller-tools
-  version: 43466124052c1a1aa7d8fd33624b00bc111fb7cf
-  repo: https://github.com/openshift/kubernetes-sigs-controller-tools.git
+  repo: https://github.com/openshift/kubernetes-sigs-controller-tools
+  version: master
 - package: k8s.io/gengo
-  version: 4242d8e6c5dba56827bb7bcf14ad11cda38f3991
+  version: master
```
deads2k: this is wrong.  Choose a specific level.

glide.yaml
```diff
@@ -18,10 +19,10 @@ import:
 
 # crd-schema-gen
 - package: sigs.k8s.io/controller-tools
-  version: 43466124052c1a1aa7d8fd33624b00bc111fb7cf
-  repo: https://github.com/openshift/kubernetes-sigs-controller-tools.git
+  repo: https://github.com/openshift/kubernetes-sigs-controller-tools
+  version: master
```
deads2k: this looks unlikely to keep working well.  If this is a fork, you need a branch.  If we really want to track a master, pin to sha.

sttts: https://github.com/sttts/controller-tools/tree/master-1.13-rebase – how should we call it?

glide.yaml
mfojtik: Tracking branch now with TODO about moving this to new repo once @damemi is ready.

glide.yaml
mfojtik: added specific commit.

glide.yaml
mfojtik: Tracking branch now with standard name.

mfojtik:
(re-adding label manually as I just removed extra new line in glide.yaml ;-)

PR #340, opened 2019-04-10
wrap the existing wrap transport to handle 1.13

deads2k:
This avoids losing the important wrap transport for in-cluster config. There is more work to collapse the two paths we have.

/assign @mfojtik 

wait for proof before merging


deads2k:
proof here: https://github.com/openshift/cluster-kube-apiserver-operator/pull/379

PR #350, opened 2019-04-12
Lock installerpod manifests

tnozicka:
This avoids racing with cert recovery tool by using advisory POSIX file locks. These locks are identified by [i_node,pid] pair and kernel will release them if we are terminated (ungracefully).

/cc @deads2k


deads2k:
I'm -1 on this.  I think we have a general need for a cluster-admin to control whether installer pods are allowed to progress on a particular node.  

we spoke in depth at scrum and this is on hold

/hold

PR #353, opened 2019-04-16
domain is not optional, repo is optional

### LINKED ISSUE #347: --domain says it is optional, but it is not
https://github.com/openshift/library-go/blob/master/cmd/crd-schema-gen/generator/generator.go

https://github.com/openshift/library-go/blob/master/cmd/crd-schema-gen/generator/generator.go#L38 lists the domain flag as  optional, but https://github.com/openshift/library-go/blob/master/cmd/crd-schema-gen/generator/generator.go#L73-L75 throws an error if --domain is not specified.

So, is it optional, or not?

coreydaley:
Fixes #347


coreydaley:
Looks like I still need an approval for this pull request.

PR #354, opened 2019-04-17
Change ClusterOperatorStatus Failed to Degraded

eparis:
This is just a rename, but it should take on a slightly newer definition over time.

https://github.com/openshift/cluster-version-operator/pull/161


glide.yaml
```diff
@@ -1,6 +1,7 @@
 package: github.com/openshift/library-go
 import:
 - package: k8s.io/apimachinery
+  version: kubernetes-1.13.4
```
eparis: how the heck was this not already pinned? do you know how long it took me to figure out this was my problem? I've gotten so dumb.

eparis:
based on https://github.com/openshift/api/pull/287

glide.yaml
deads2k: > how the heck was this not already pinned? do you know how long it took me to figure out this was my problem? I've gotten so dumb.

Clearly that guy should be fired.

deads2k:
everybody jump!

/lgtm

eparis:
/hold
probably shouldn't jump until:
https://github.com/openshift/cluster-version-operator/pull/165

eparis:
see: https://github.com/openshift/api/pull/287

deads2k:
/hold cancel

running for API freeze, the CVO will take the update eventually, releasing this

PR #355, opened 2019-04-17
Add UpdateStaticPodOperatorSpec function

damemi:
My use case: writing an e2e test for revision pruning, want to trigger a bunch of new revisions by updating the `ForceRedeploymentReason` in `StaticPodOperatorSpec`. Currently we can get the spec but I don't know of any way to update it.


deads2k:
squash, then lgtm

You must wait until after https://github.com/openshift/library-go/pull/354

PR #356, opened 2019-04-17
Set an arbitrary event namespace when not running inside a cluster

smarterclayton:
openshift-config is a reasonable place for operators to report events,
although they may not have access. Allow a builder to set it so we can
both test locally and override it for some operators.

At worst, we will attempt to log events to "openshift-config"

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -67,10 +67,11 @@ type ControllerBuilder struct {
 	fileObserver            fileobserver.Observer
 	fileObserverReactorFn   func(file string, action fileobserver.ActionType) error
 
-	startFunc        StartFunc
-	componentName    string
-	instanceIdentity string
-	observerInterval time.Duration
+	startFunc          StartFunc
+	componentName      string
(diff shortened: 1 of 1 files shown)


pkg/controller/controllercmd/builder.go
```diff
@@ -246,12 +252,15 @@ func (b *ControllerBuilder) Run(config *unstructured.Unstructured, ctx context.C
...
 
-func (b *ControllerBuilder) getNamespace() (string, error) {
+func (b *ControllerBuilder) getComponentNamespace() (string, error) {
+	if len(b.componentNamespace) > 0 {
+		return b.componentNamespace, nil
+	}
 	nsBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
 	if err != nil {
-		return "", err
+		return "openshift-config", err
```
deads2k: this is a user owned namespace.  Can we get `openshift-config-managed` isntead?

deads2k:
request for the operator driven namespace.  lgtm otherwise.  Self-tag after fix

/approve

PR #357, opened 2019-04-18
resourcesynccontroller: use cached get to confirm resource exists before deleting target

mfojtik:
This change will make us check cache before making a DELETE call that hits the API server.
This will reduce the amount of DELETE requests we issue when the resource (secret/configmap) does not exists in the target namespace (and therefore the DELETE call always returned 404).

/cc @smarterclayton


mfojtik:
@deads2k this time I added unit test that actually run the controller and confirm the expected behavior 

I also found that my original PR missed the `resourceapply` change and worked only when resource location was empty...

mfojtik:
/hold

I'm going to open a proof this time.

PR #359, opened 2019-04-18
loglevel: fix reporting via events

mfojtik:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1701272

/cc @deads2k


pkg/operator/loglevel/logging_controller.go
```diff
@@ -54,12 +54,20 @@ func (c LogLevelController) sync() error {
 	logLevel := fmt.Sprintf("%d", LogLevelToKlog(detailedSpec.OperatorLogLevel))
 
 	var level klog.Level
+
+	oldLevel, ok := level.Get().(klog.Level)
```
sjenning: this is not a valid way to read the current loglevel; it will always return 0

PR #360, opened 2019-04-18
wait for caches to synchronize before working

deads2k:
Upgrades are flapping on revisions because the operator is getting redeployed and we don't wait for cache syncing. The specific instance is https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade-4.0/1077/artifacts/e2e-aws-upgrade/events.json in the revision controller, but we had many that were missing because they all used to be live look ups.

@openshift/sig-master


deads2k:
I also refactored a few names that weren't consistent after so many refactors

PR #363, opened 2019-04-22
Remove EnsureOperatorConfigExists helper

damemi:
Removes this helper which should no longer be used in favor of CVO managed operator resources


damemi:
/hold cancel

most operators have this now

PR #364, opened 2019-04-23
controller: retry connection errors when getting extension-apiserver-authentication 

mfojtik:
Add retry to prevent being fatal:

`Get https://172.30.0.1:443/api/v1/namespaces/kube-system/configmaps/extension-apiserver-authentication: dial tcp 172.30.0.1:443: connect: connection refused`

Best I can think of to address https://bugzilla.redhat.com/show_bug.cgi?id=1702104 (but I suspect there might be other components failing due to races).

/cc @deads2k

### DIFF
--- a/pkg/config/serving/server.go
+++ b/pkg/config/serving/server.go
@@ -1,17 +1,23 @@
 package serving
 
 import (
+	"context"
+	"time"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/wait"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	genericapiserveroptions "k8s.io/apiserver/pkg/server/options"
+	"k8s.io/klog"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 )
 
-func ToServerConfig(servingInfo configv1.HTTPServingInfo, authenticationConfig operatorv1alpha1.DelegatedAuthentication, authorizationConfig operatorv1alpha1.DelegatedAuthorization, kubeConfigFile string) (*genericapiserver.Config, error) {
+func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, authenticationConfig operatorv1alpha1.DelegatedAuthentication, authorizationConfig operatorv1alpha1.DelegatedAuthorization,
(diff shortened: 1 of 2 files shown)


pkg/config/serving/server.go
```diff
@@ -20,23 +25,46 @@ func ToServerConfig(servingInfo configv1.HTTPServingInfo, authenticationConfig o
...
 	if !authenticationConfig.Disabled {
 		authenticationOptions := genericapiserveroptions.NewDelegatingAuthenticationOptions()
 		authenticationOptions.RemoteKubeConfigFile = kubeConfigFile
-		if err := authenticationOptions.ApplyTo(&config.Authentication, config.SecureServing, config.OpenAPIConfig); err != nil {
-			return nil, err
+
+		// In some cases the API server can return connection refused when getting the "extension-apiserver-authentication"
+		// config map.
+		err := wait.PollImmediateUntil(20*time.Second, func() (done bool, err error) {
+			lastApplyErr = authenticationOptions.ApplyTo(&config.Authentication, config.SecureServing, config.OpenAPIConfig)
```
soltysh: I'm not sure that wrapping `ApplyTo` into retry loop is a good idea here. Why not re-trying those operation that actually fail due to that error?

mfojtik: @soltysh that goes deep into kube code, i'm not sure it will be accepted in upstream

soltysh: Gotcha, I started looking into the code to nail down those places and haven't found it easily, so that makes sense.

pkg/config/serving/server.go
```diff
@@ -20,23 +25,46 @@ func ToServerConfig(servingInfo configv1.HTTPServingInfo, authenticationConfig o
...
+
 	if !authenticationConfig.Disabled {
 		authenticationOptions := genericapiserveroptions.NewDelegatingAuthenticationOptions()
 		authenticationOptions.RemoteKubeConfigFile = kubeConfigFile
-		if err := authenticationOptions.ApplyTo(&config.Authentication, config.SecureServing, config.OpenAPIConfig); err != nil {
-			return nil, err
+
+		// In some cases the API server can return connection refused when getting the "extension-apiserver-authentication"
+		// config map.
+		err := wait.PollImmediateUntil(20*time.Second, func() (done bool, err error) {
```
deads2k: don't wait forever.

mfojtik: i don't AFAIK, there is 20s timeout on this + stopCh

deads2k: yeah, the stop channel won't close.

deads2k: We prefer crash looping to waiting forever.  PollImmediate, every one second for 15 seconds, then crashloop

PR #365, opened 2019-04-23
wait for certs to be present before installing

deads2k:
for https://bugzilla.redhat.com/show_bug.cgi?id=1702104

This waits for unrevisioned certs to be present before continuing.

/hold

hold for proof and unit tests

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -69,6 +69,7 @@ type InstallerController struct {
 	operatorClient v1helpers.StaticPodOperatorClient
 
 	configMapsGetter corev1client.ConfigMapsGetter
+	secretsGetter    corev1client.SecretsGetter
 	podsGetter       corev1client.PodsGetter
 
 	cachesToSync  []cache.InformerSynced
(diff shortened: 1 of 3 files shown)


deads2k:
proof pull https://github.com/openshift/cluster-kube-apiserver-operator/pull/431

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -1320,3 +1324,9 @@ func TestSetConditions(t *testing.T) {
 	}
 
 }
+
+func TestEnsureCert(t *testing.T) {
```
mfojtik: nit: implement :-)

deads2k: > nit: implement :-)

hence the hold

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -711,6 +751,10 @@ func (c InstallerController) sync() error {
 		return nil
 	}
 
+	if err := c.ensureCerts(); err != nil {
+		return err
```
mfojtik: can we log the error here? (V(4)) ?

mfojtik: iow. "why my installer pods are not starting"

deads2k: > iow. "why my installer pods are not starting"

level above logs

PR #368, opened 2019-04-24
Revision recovery from operator object deletion

damemi:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1701316

If an operator resource is deleted, it comes back with latest revision number 0 which will cause conflicts with what the actual latest revision is. This uses the revision status configmaps to get the revision history and update the operator status to the actual latest revision number

### DIFF
--- a/pkg/operator/staticpod/controller/revision/revision_controller.go
+++ b/pkg/operator/staticpod/controller/revision/revision_controller.go
@@ -2,6 +2,7 @@ package revision
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


damemi:
Right now this only briefly sets the LatestAvailableRevision to the correct value upon recreation of the resource, then it's immediately set back to 1. Trying to track down why

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -259,6 +291,20 @@ func (c RevisionController) sync() error {
 		return nil
 	}
 
+	// Check to see if current revision is accurate and if not, search through configmaps for latest revision
+	latestRevision, err := c.getLatestAvailableRevision(operatorStatus)
+	if err != nil {
+		return err
+	}
+	// Then make sure that revision number is what's in the operator status
+	operatorStatus, _, err = v1helpers.UpdateStaticPodStatus(c.operatorClient, func(status *operatorv1.StaticPodOperatorStatus) error {
```
deads2k: only do this if the value isn't zero

deads2k: if you make a change, just return and let yourself be rekicked.

deads2k:
You need a proof pull.

damemi:
@deads2k updated and opened a PR to kas https://github.com/openshift/cluster-kube-apiserver-operator/pull/438

mfojtik:
/lgtm
/hold

(cancel hold if the proof is successful)

PR #370, opened 2019-04-26
loglevel: rework setting the log level using flags

mfojtik:
The recommended way (from klog) to set the loglevel is go trough flags and use the flag.Value to mutate the current verbosity. I added that before we fallback to mutate the global logging var directly.

Added helper to determine what is the current logging verbosity and added a test to prove it works.

/cc @deads2k

### DIFF
--- a/pkg/operator/loglevel/logging_controller.go
+++ b/pkg/operator/loglevel/logging_controller.go
@@ -51,23 +51,20 @@ func (c LogLevelController) sync() error {
 		return err
 	}
 
-	logLevel := fmt.Sprintf("%d", LogLevelToKlog(detailedSpec.OperatorLogLevel))
+	currentLogLevel := CurrentLogLevel()
 
-	var level klog.Level
-
-	oldLevel, ok := level.Get().(klog.Level)
-	if !ok {
-		oldLevel = level
+	// When the current loglevel is the desired one, do nothing
+	if currentLogLevel == detailedSpec.OperatorLogLevel {
+		return nil
 	}
 
-	if err := level.Set(logLevel); err != nil {
-		c.eventRecorder.Warningf("LoglevelChangeFailed", "Unable to set loglevel level %v", err)
+	// Set the new loglevel if the operator spec changed
+	if err := SetVerbosityValue(detailedSpec.OperatorLogLevel); err != nil {
+		c.eventRecorder.Warningf("OperatorLoglevelChangeFailed", "Unable to change operator log level from %q to %q: %v", currentLogLevel, detailedSpec.OperatorLogLevel, err)
 		return err
 	}
 
-	if oldLevel.String() != logLevel {
-		c.eventRecorder.Eventf("LoglevelChange", "Changed loglevel level to %q", logLevel)
-	}
+	c.eventRecorder.Eventf("OperatorLoglevelChange", "Operator log level changed from %q to %q", currentLogLevel, detailedSpec.OperatorLogLevel)
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/operator/loglevel/logging_controller.go
```diff
@@ -51,23 +51,20 @@ func (c LogLevelController) sync() error {
...
+	// Set the new loglevel if the operator spec changed
+	if err := SetVerbosityFlagValue(detailedSpec.OperatorLogLevel); err != nil {
 		c.eventRecorder.Warningf("LoglevelChangeFailed", "Unable to set loglevel level %v", err)
 		return err
 	}
 
-	if oldLevel.String() != logLevel {
-		c.eventRecorder.Eventf("LoglevelChange", "Changed loglevel level to %q", logLevel)
-	}
+	c.eventRecorder.Eventf("LoglevelChange", "Changed loglevel level to %q", detailedSpec.OperatorLogLevel)
```
soltysh: `Changed loglevel to %q`

pkg/operator/loglevel/logging_controller.go
```diff
@@ -51,23 +51,20 @@ func (c LogLevelController) sync() error {
...
-		oldLevel = level
+	// When the current loglevel is the desired one, do nothing
+	if currentLogLevel == detailedSpec.OperatorLogLevel {
+		return nil
 	}
 
-	if err := level.Set(logLevel); err != nil {
+	// Set the new loglevel if the operator spec changed
+	if err := SetVerbosityFlagValue(detailedSpec.OperatorLogLevel); err != nil {
 		c.eventRecorder.Warningf("LoglevelChangeFailed", "Unable to set loglevel level %v", err)
```
soltysh: Why we're such fans of repeating the level word :wink:

mfojtik:
/hold

until I prove this works manually

pkg/operator/loglevel/util.go
```diff
@@ -16,3 +24,56 @@ func LogLevelToKlog(logLevel operatorv1.LogLevel) int {
...
+	switch {
+	case klog.V(8) == true:
+		return operatorv1.TraceAll
+	case klog.V(6) == true:
+		return operatorv1.Trace
+	case klog.V(4) == true:
+		return operatorv1.Debug
+	case klog.V(2) == true:
+		return operatorv1.Normal
+	default:
```
soltysh: I'm not sure this is reachable at all, given how `V(x)` performs its checks. Basically every level up to 2 (inclusively) will pick the previous case.

pkg/operator/loglevel/logging_controller.go
mfojtik: i fixed these

pkg/operator/loglevel/util.go
mfojtik: i think this means we won't allow V(1) or V(0) and defer to V(2) always.

soltysh: We do allow both, it's just everything <2 is Normal.

mfojtik: @soltysh we don't allow that for operatorv1.LogLevel is what I meant (there is no string value representing those, so if the current is 0 or 1, we report `Normal` (and won't change it).

mfojtik:
/hold cancel

I tested this manually and switching log level works as expected. Also I don't see the event repeating anymore.

pkg/operator/loglevel/util.go
soltysh: :+1:

PR #371, opened 2019-04-26
installer: fire event on missing required certs

mfojtik:
The "missing: secrets/serving-cert should be logged by event.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -741,8 +741,9 @@ func (c InstallerController) ensureCerts() error {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -741,7 +741,9 @@ func (c InstallerController) ensureCerts() error {
 		return nil
 	}
 
-	return fmt.Errorf("missing: %v", strings.Join(missing, ","))
+	c.eventRecorder.Eventf("RequiredCertsMissing", strings.Join(missing, ","))
```
deads2k: warning

deads2k:
nit on warning. lgtm otherwise

/approve

Self tag after fix up.

PR #372, opened 2019-04-26
always retry failed installer pods

deads2k:
We currently only retry on OOMs, but we can retry on anything. The backoff should keep us from storming and the last few lines of the installer logs are always included in our operator log.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -297,7 +297,7 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 			}
 
 			pendingNewRevision := operatorStatus.LatestAvailableRevision > currNodeState.TargetRevision
-			newCurrNodeState, oom, err := c.newNodeStateForInstallInProgress(currNodeState, pendingNewRevision)
+			newCurrNodeState, installerPodFailed, err := c.newNodeStateForInstallInProgress(currNodeState, pendingNewRevision)
 			if err != nil {
 				return true, err
 			}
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -212,16 +212,13 @@ func TestNewNodeStateForInstallInProgress(t *testing.T) {
 	}
 
 	_, currStatus, _, _ = fakeStaticPodOperatorClient.GetStaticPodOperatorState()
-	if generation := currStatus.NodeStatuses[0].LastFailedRevision; generation != 2 {
-		t.Errorf("expected last failed revision generation for node to be 2, got %d", generation)
+	if generation := currStatus.NodeStatuses[0].LastFailedRevision; generation != 0 {
```
deads2k: this is an interesting side effect.  We don't mark the revision failed because we constantly retry it.  Objections/concerns?

mfojtik: i'm ok with this as this might blew in unexpected ways if we start setting that property right?

also technically, if we still trying to run installer pod, retries means we are not "failed" (maybe we should after N retries?)

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -560,10 +560,9 @@ func (c *InstallerController) newNodeStateForInstallInProgress(currNodeState *op
 		for _, containerStatus := range installerPod.Status.ContainerStatuses {
 			if containerStatus.State.Terminated != nil && len(containerStatus.State.Terminated.Message) > 0 {
 				errors = append(errors, fmt.Sprintf("%s: %s", containerStatus.Name, containerStatus.State.Terminated.Message))
-				if containerStatus.State.Terminated.Reason == "OOMKilled" {
-					// do not set LastFailedRevision
-					return currNodeState, true, nil
-				}
+				klog.Errorf("installer errors: %v", strings.Join(errors, "\n"))
```
mfojtik: event worthy?

PR #373, opened 2019-04-26
set operator version to match other operators

deads2k:
/hold

hold for proof

This could resolve our operator version if the static pod operators set a different key.

proof in https://github.com/openshift/cluster-kube-apiserver-operator/pull/440


mfojtik:
/lgtm

the proof seems to work.

PR #375, opened 2019-04-29
logging: fix case when logging is set to empty string

mfojtik:
In case the logging is set to empty string, we should assume Normal loglevel and do not change it.


mfojtik:
self-approve, this fixed @sjenning problem and I added case into unit test to not regress here.

PR #377, opened 2019-04-30
WIP: service availability controller

mfojtik:
Just a concept (for 4.2). During 4.1 we encountered the "connection refused" and other API availability issues related to connectivity either via NLB, service network or external hostnames.
In some cases these were caused by networking (OVS upgrade), in other cases these were caused by handing shutdown (systemd SIGTERM).

This PR introduces controller that will perform monitoring of service availability of static pod operands from the operator. It allows to define "monitors" that will receive the node definition (can be just status) where they can read external/internal/hostname and perform HTTP(S) request to report connectivity issues continuously. If the NLB hostname is passed here, we can almost monitor the NLB connectivity issues.

If a node is failing to respond to monitor, we can set the state for that node to "degraded" (in this PR I just fire warning events).

This is up to discussion in architectural call.

/cc @openshift/sig-master

### DIFF
--- a/pkg/operator/staticpod/controller/availability/availability_controller.go
+++ b/pkg/operator/staticpod/controller/availability/availability_controller.go
@@ -1,0 +1,355 @@
+package availability
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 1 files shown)


mfojtik:
/close

Not needed, we should rather use alerts and prometheus.

PR #380, opened 2019-05-01
fix updatestatus return value

deads2k:
See 20 lines down where this was fixed for static pods


sanchezl:
Do we need to do the same to `UpdateSpec`?

deads2k:
> Do we need to do the same to `UpdateSpec`?

oops, your turn

PR #376, opened 2019-04-30
installer: ensure the etcd-client secret is available

mfojtik:
A high rate of failed installer pods on revision `1` observed via CI. Caused by missing `etcd-client-1` secret which is part of the "required" secrets.

```
I0429 20:43:28.093507       1 cmd.go:246] Creating target resource directory "/etc/kubernetes/static-pod-resources/kube-apiserver-pod-1" ...
I0429 20:43:28.093861       1 cmd.go:171] Creating target resource directory "/etc/kubernetes/static-pod-resources/kube-apiserver-pod-1" ...
I0429 20:43:28.093951       1 cmd.go:179] Getting secrets ...
I0429 20:43:28.096810       1 copy.go:24] Failed to get secret openshift-kube-apiserver/etcd-client-1: secrets "etcd-client-1" not found
F0429 20:43:28.101133       1 cmd.go:89] failed to copy: secrets "etcd-client-1" not found
```

We already have mechanism to ensure some (cert) required secrets and configs are available at the time we run the pod, this add all required secrets and configs to the check as the installer pod will always fail when those does not exists and we should rather wait in the operator than creating and failing the installer pod?

Also improving the error message to see the namespace (not always obvious in tests).

Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/446

/cc @deads2k 
/cc @damemi

PR #381, opened 2019-05-01
properly wait for revision resources

deads2k:
ugly version of waiting for the correct names.  I don't care if it's cleaned up later, but this is what I have time for now.

@ironcladlou 
/assign @mfojtik @sttts


deads2k:
I think this is a correction of https://github.com/openshift/library-go/pull/376

mfojtik:
https://github.com/openshift/library-go/pull/382
https://github.com/openshift/library-go/pull/383

Will distribute bumps, closing.

PR #382, opened 2019-05-02
installer: add revision suffix to required resources

mfojtik:
Cleaner alternative to https://github.com/openshift/library-go/pull/381

/cc @deads2k

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -10,18 +10,19 @@ import (
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
-	"k8s.io/klog"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -702,35 +703,51 @@ func getInstallerPodImageFromEnv() string {
...
+	configMapsNames := sets.NewString()
+	for _, c := range c.configMaps {
+		configMapsNames.Insert(c.Name)
+	}
 	for _, cm := range append(append([]revision.RevisionResource{}, c.certConfigMaps...), c.configMaps...) {
 		if cm.Optional {
 			continue
 		}
-		_, err := c.configMapsGetter.ConfigMaps(c.targetNamespace).Get(cm.Name, metav1.GetOptions{})
+		name := cm.Name
```
sttts: I would rather split the for loop. This does not make the logic easier.

mfojtik: i don't like duplicating the logic just because the name in one case need suffix.

mfojtik: would adding comment here help?

sttts: We have that loop already two times. Having it 4 times does not make it worse.

sttts:
Not super nice code.</bikeshedding>

/approve

PR #384, opened 2019-05-02
Report detailed revision reason as installer pod annotation

mfojtik:
Knowing why revision N was triggered means somebody have to dig trough events or operator logs.
As we dump all pods in CI, if we store the detailed reason (what changed and how) inside the installer pod annotation, we can understand why given revision (and installer pod) was created.

/cc @deads2k 
/cc @sttts 

NOTE: This include illegal bump to openshift/api as we might need new API field if we want to do this,.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -290,7 +290,7 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 
 		// if we are in a transition, check to see whether our installer pod completed
 		if currNodeState.TargetRevision > currNodeState.CurrentRevision {
-			if err := c.ensureInstallerPod(currNodeState.NodeName, operatorSpec, currNodeState.TargetRevision); err != nil {
+			if err := c.ensureInstallerPod(currNodeState.NodeName, operatorSpec, currNodeState.TargetRevision, operatorStatus.LatestAvailableRevisionReason); err != nil {
 				c.eventRecorder.Warningf("InstallerPodFailed", "Failed to create installer pod for revision %d on node %q: %v",
 					currNodeState.TargetRevision, currNodeState.NodeName, err)
 				return true, err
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -611,11 +611,18 @@ func getInstallerPodName(revision int32, nodeName string) string {
...
 	pod := resourceread.ReadPodV1OrDie(bindata.MustAsset(filepath.Join(manifestDir, manifestInstallerPodPath)))
 
 	pod.Namespace = c.targetNamespace
 	pod.Name = getInstallerPodName(revision, nodeName)
+	if pod.Annotations == nil {
+		pod.Annotations = map[string]string{}
+	}
+
+	// Report the reason this installer pod was created.
+	pod.Annotations["operator.openshift.io/revision-reason"] = revisionReason
```
sttts: operator.openshift.io is pretty generic. staticpod.operator.openshift.io maybe?

mfojtik: sure

mfojtik:
@deads2k @sttts i think it is too late for this in 4.1, but I would like to have conversation on adding that extra field so we can debug better :-)

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -143,19 +146,20 @@ func nameFor(name string, revision int32) string {
...
 	configChanges := []string{}
+	configChangesDetailed := []string{}
 	for _, cm := range c.configMaps {
 		requiredData := map[string]string{}
 		existingData := map[string]string{}
 
 		required, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(cm.Name, metav1.GetOptions{})
 		if apierrors.IsNotFound(err) && !cm.Optional {
-			return false, err.Error()
+			return false, err.Error(), ""
```
sttts: so there is a chance that the annotation is empty?

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -191,18 +198,20 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 			existingData = existing.Data
 		}
 		if !equality.Semantic.DeepEqual(existingData, requiredData) {
+			resourceDiffString := resourceapply.JSONPatch(existing, required)
 			if klog.V(4) {
-				klog.Infof("secret %q changes for revision %d: %s", s.Name, revision, resourceapply.JSONPatch(existing, required))
+				klog.Infof("secret %q changes for revision %d: %s", s.Name, revision, resourceDiffString)
```
sttts: do we leak security critical information?

vendor/github.com/openshift/api/operator/v1/types.go
```diff
@@ -185,6 +185,10 @@ type StaticPodOperatorStatus struct {
 	// +optional
 	LatestAvailableRevision int32 `json:"latestAvailableRevision,omitEmpty"`
 
+	// latestAvailableRevisionReason describe the reason the latest available revision was updated
+	// +optional
+	LatestAvailableRevisionReason string `json:"latestAvailableRevisionReason,emitEmpty"`
```
sttts: is omitempty case insensitive?

pkg/operator/staticpod/controller/revision/revision_controller.go
mfojtik: yeah, was thinking the same... I think we might even now on loglevel 4 ;-)

vendor/github.com/openshift/api/operator/v1/types.go
mfojtik: and I have typo there

pkg/operator/staticpod/controller/revision/revision_controller.go
mfojtik: i believe in this case we won't create the installer pod at all, so that value is not used.

PR #387, opened 2019-05-02
controllercmd: handle TERM signals gracefully in operator

mfojtik:
This allows to wire `stopCh := genericapiserver.SetupSignalHandler()` inside each operator `main.go` and pass the stopCh to the operator command. When the stopCh closes, it cancels the context the command use and trigger shutdown for all controller loops and observers (if wired properly).

https://github.com/kubernetes/kubernetes/pull/77361 to make this more generic.

/cc @sttts 
/cc @deads2k

### DIFF
--- a/pkg/controller/controllercmd/cmd.go
+++ b/pkg/controller/controllercmd/cmd.go
@@ -11,15 +11,17 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/util/logs"
+	"k8s.io/klog"
 
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+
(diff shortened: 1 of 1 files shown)


pkg/controller/controllercmd/cmd.go
```diff
@@ -51,22 +53,42 @@ func NewControllerCommandConfig(componentName string, version version.Info, star
...
 			rand.Seed(time.Now().UTC().UnixNano())
 			logs.InitLogs()
+
+			// handle SIGTERM and SIGINT by cancelling the context.
+			shutdownCtx, cancel := context.WithCancel(ctx)
+			shutdownHandler := server.SetupSignalHandler()
+			go func() {
+				defer cancel()
+				<-shutdownHandler
+				klog.Infof("Received SIGTERM signal, shutting down controller.")
```
sttts: TERM or INT?

pkg/controller/controllercmd/cmd.go
```diff
@@ -51,22 +53,42 @@ func NewControllerCommandConfig(componentName string, version version.Info, star
 
 // NewCommand returns a new command that a caller must set the Use and Descriptions on.  It wires default log, profiling,
 // leader election and other "normal" behaviors.
+// TODO: use the NewCommandWithContext instead, this is here to be less disturbing for existing usages.
 func (c *ControllerCommandConfig) NewCommand() *cobra.Command {
```
sttts: there is a syntax for deprecation.

pkg/controller/controllercmd/cmd.go
mfojtik: either

PR #389, opened 2019-05-03
only acccept cloud config files for certain providers

deads2k:
/hold

PR #390, opened 2019-05-03
only acccept cloud config files for certain providers

sttts:
Replaces https://github.com/openshift/library-go/pull/389.


tnozicka:
["Change looks good otherwise and is needed to unblock work."](https://github.com/openshift/library-go/pull/389#issuecomment-488889343) (@derekwaynecarr)
@deads2k asked us to move it forward to unblock other teams

/approve
/lgtm

PR #392, opened 2019-05-03
file-watcher-watchdog: add command to kill the operand process on file changes

mfojtik:
The purpose of this package is to live aside to an operand container (side-car). If the operand pod runs with `shareProcessNamespace: true` the watchdog can send SIGTERM to the operand to trigger graceful shutdown.

The watchdog use the fileobserver to detect changes in files (`--files`) and in reaction to a change it will send TERM signal to a process specified by name `--process-name`.

### DIFF
--- a/pkg/operator/watchdog/cmd.go
+++ b/pkg/operator/watchdog/cmd.go
@@ -1,0 +1,345 @@
+package watchdog
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+
+	"github.com/openshift/library-go/pkg/config/client"
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/retry"
+)
+
+type FileWatcherOptions struct {
+	// ProcessName is the name of the process we will send SIGTERM
+	ProcessName string
+
+	// Files lists all files we want to monitor for changes
+	Files      []string
+	KubeConfig string
+
+	// Namespace to report events to
+	Namespace string
+	recorder  events.Recorder
+
+	// Interval specifies how aggressive we want to be in file checks
+	Interval time.Duration
+
+	// Time to give the process to terminate gracefully
+	TerminationGracePeriod time.Duration
+
+	// for unit-test to mock getting the process PID (unit-test)
+	findPidByNameFn func(name string) (int, bool, error)
+
+	// processExistsFn to mock checking the process PID (unit-test)
+	processExistsFn func(int) (bool, error)
+
+	// for unit-test to mock sending UNIX signals
+	handleTerminationFn func(pid int) error
+
+	handleKillFn func(pid int) error
+
+	// for unit-test to mock prefixing files (/proc/PID/root)
+	addProcPrefixToFilesFn func([]string, int) []string
+
+	// lastTerminatedPid is used track the value of a PID that we already terminated
+	lastTerminatedPid int
+}
+
+func NewFileWatcherOptions() *FileWatcherOptions {
+	return &FileWatcherOptions{
+		findPidByNameFn:        FindProcessByName,
+		processExistsFn:        ProcessExists,
+		addProcPrefixToFilesFn: addProcPrefixToFiles,
+		handleTerminationFn: func(pid int) error {
+			return syscall.Kill(pid, syscall.SIGTERM)
+		},
+		handleKillFn: func(pid int) error {
+			return syscall.Kill(pid, syscall.SIGKILL)
+		},
+	}
+}
+
+// NewFileWatcherWatchdog return the file watcher watchdog command.
+// This command should be used as a side-car to a container which will react to file changes in the main container
+// and terminate the main container process in case a change is observed.
+// TODO: If the main container start before the watchdog side-car container (image pull) there might be a case
+// 		 the watchdog won't react to a changed file (simply because it is not running yet). In that case the main process
+//       will not be reloaded. However, the operator image should be pulled on master node and therefore chances to hit this
+//       case are minimal.
+func NewFileWatcherWatchdog() *cobra.Command {
+	o := NewFileWatcherOptions()
+
+	cmd := &cobra.Command{
+		Use:   "file-watcher-watchdog",
+		Short: "Watch files on the disk and terminate the specified process on change",
+		Run: func(cmd *cobra.Command, args []string) {
+			klog.V(1).Info(cmd.Flags())
+			klog.V(1).Info(spew.Sdump(o))
+
+			// Handle shutdown
+			termHandler := server.SetupSignalHandler()
+			ctx, shutdown := context.WithCancel(context.TODO())
+			go func() {
+				defer shutdown()
+				<-termHandler
+			}()
+
+			if err := o.Complete(); err != nil {
+				klog.Fatal(err)
+			}
+			if err := o.Validate(); err != nil {
+				klog.Fatal(err)
+			}
+
+			if err := o.Run(ctx); err != nil {
+				klog.Fatal(err)
+			}
+		},
+	}
+
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *FileWatcherOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.ProcessName, "process-name", "", "name of the process to send TERM signal to on file change (eg. 'hyperkube').")
+	fs.StringSliceVar(&o.Files, "files", o.Files, "comma separated list of file names to monitor for changes")
+	fs.StringVar(&o.KubeConfig, "kubeconfig", o.KubeConfig, "kubeconfig file or empty")
+	fs.StringVar(&o.Namespace, "namespace", o.Namespace, "namespace to report the watchdog events")
+	fs.DurationVar(&o.Interval, "interval", 5*time.Second, "interval specifying how aggressive the file checks should be")
+	fs.DurationVar(&o.TerminationGracePeriod, "termination-grace-period", 30*time.Second, "interval specifying how long to wait until sending KILL signal to the process")
+}
+
+func (o *FileWatcherOptions) Complete() error {
+	clientConfig, err := client.GetKubeConfigOrInClusterConfig(o.KubeConfig, nil)
+	if err != nil {
+		return err
+	}
+	kubeClient, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 10*time.Second)
+	defer cancel()
+
+	// Get event recorder.
+	// Retry on connection errors for 10s, but don't error out, instead fallback to the namespace.
+	var eventTarget *v1.ObjectReference
+	err = retry.RetryOnConnectionErrors(ctx, func(context.Context) (bool, error) {
+		var clientErr error
+		eventTarget, clientErr = events.GetControllerReferenceForCurrentPod(kubeClient, o.Namespace, nil)
+		if clientErr != nil {
+			return false, clientErr
+		}
+		return true, nil
+	})
+	if err != nil {
+		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
+	}
+	o.recorder = events.NewRecorder(kubeClient.CoreV1().Events(o.Namespace), "file-change-watchdog", eventTarget)
+
+	return nil
+}
+
+func (o *FileWatcherOptions) Validate() error {
+	if len(o.ProcessName) == 0 {
+		return fmt.Errorf("process name must be specified")
+	}
+	if len(o.Files) == 0 {
+		return fmt.Errorf("at least one file to observe must be specified")
+	}
+	if len(o.Namespace) == 0 && len(os.Getenv("POD_NAMESPACE")) == 0 {
+		return fmt.Errorf("either namespace flag or POD_NAMESPACE environment variable must be specified")
+	}
+	return nil
+}
+
+// runPidObserver runs a loop that observes changes to the PID of the process we send signals after change is detected.
+func (o *FileWatcherOptions) runPidObserver(ctx context.Context, pidObservedCh chan int) {
+	defer close(pidObservedCh)
+	currentPID := 0
+	retries := 0
+	pollErr := wait.PollImmediateUntil(1*time.Second, func() (done bool, err error) {
+		retries++
+		// attempt to find the PID by process name via /proc
+		observedPID, found, err := o.findPidByNameFn(o.ProcessName)
+		if !found || err != nil {
+			klog.Warningf("Unable to determine PID for %q (retry: %d, err: %v)", o.ProcessName, retries, err)
+			return false, nil
+		}
+
+		if currentPID == 0 {
+			currentPID = observedPID
+			// notify runWatchdog when the PID is initially observed (we need the PID to mutate file paths).
+			pidObservedCh <- observedPID
+		}
+
+		// watch for PID changes, when observed restart the observer and wait for the new PID to appear.
+		if currentPID != observedPID {
+			return true, nil
+		}
+
+		return false, nil
+	}, ctx.Done())
+
+	// These are not fatal errors, but we still want to log them out
+	if pollErr != nil && pollErr != wait.ErrWaitTimeout {
+		klog.Warningf("Unexpected error: %v", pollErr)
+	}
+}
+
+// readInitialFileContent reads the content of files specified.
+// This is needed by file observer.
+func readInitialFileContent(files []string) (map[string][]byte, error) {
+	initialContent := map[string][]byte{}
+	for _, name := range files {
+		// skip files that does not exists (yet)
+		if _, err := os.Stat(name); os.IsNotExist(err) {
+			continue
+		}
+		content, err := ioutil.ReadFile(name)
+		if err != nil {
+			return nil, err
+		}
+		initialContent[name] = content
+	}
+	return initialContent, nil
+}
+
+// addProcPrefixToFiles mutates the file list and prefix every file with /proc/PID/root.
+// With shared pid namespace, we are able to access the target container filesystem via /proc.
+func addProcPrefixToFiles(oldFiles []string, pid int) []string {
+	files := []string{}
+	for _, file := range oldFiles {
+		files = append(files, filepath.Join("/proc", fmt.Sprintf("%d", pid), "root", file))
+	}
+	return files
+}
+
+// Run the main watchdog loop.
+func (o *FileWatcherOptions) Run(ctx context.Context) error {
+	for {
+		{
+			o.lastTerminatedPid = 0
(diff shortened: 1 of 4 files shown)


sttts:
PID namespaces are available as beta since 1.12 🎉

Following the naming discussion on slack: file-change-monitor or file-change-watchdog are better names. Also this is no suicide anymore as the sidecar kills another process.

mfojtik:
@sttts @deads2k updated.

I think I want to fully use the shared PID namespace here and use the `/proc/N/root/var/...` trick to check the files. That will reduce the copy&paste for container volume mounts.

I already adjusted the logic here, just need to add the proc path prefix in the sweet spot.

I also added unit tests to prove this works (at least in theory ;-).

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (o *FileWatcherOptions) Run(ctx context.Context) error {
+	runCtx, shutdown := context.WithCancel(ctx)
+
+	pidObservedCh := make(chan struct{})
+	go o.monitorProcessPID(shutdown, pidObservedCh)
+
+	// Wait while we get the initial PID for the process
+	klog.Infof("Waiting for process %q ...", o.ProcessName)
+	<-pidObservedCh
```
mfojtik: this is the sweet spot... we can prefix all Files with `/proc/getCurrentPID()/root/` :-)

mfojtik: when the PID change, the watchdog will shutdown (and container will restart)... it will get the new PID and prefix the files again.

sttts:
I like this, also with the /proc file access.

mfojtik:
@sttts implementation finished, i'm going to test this live tomorrow morning with kube apiserver

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *FileWatcherOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.ProcessName, "process-name", "", "name of the process to send signal to on file change (eg. 'hyperkube').")
+	fs.StringSliceVar(&o.Files, "files", o.Files, "comma separated list of file names to monitor for changes")
+	fs.StringVar(&o.KubeConfig, "kubeconfig", o.KubeConfig, "kubeconfig file or empty")
+	fs.StringVar(&o.Namespace, "namespace", o.Namespace, "namespace to report the watchdog events")
```
mfojtik: this is used only for events... as this will restart on every change, I think we **need** events to confirm this works as expected.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+// When the PID change is observed (after we got initial PID), the shutdown is called (to signal watchdog to restart).
+func (o *FileWatcherOptions) startPidObserver(shutdown context.CancelFunc, pidObservedCh chan int) {
+	defer shutdown()
+	defer close(pidObservedCh)
+	currentPID := 0
+	for {
+		// get the monitored process PID
+		observedPID, found, err := o.getPIDByNameFn(o.ProcessName)
+		if !found {
+			klog.Warningf("Process %q does not have PID", o.ProcessName)
```
mfojtik: (this means the watchdog will restart until it gets the process)

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/openshift/library-go/pkg/controller/fileobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/retry"
+)
+
+type FileWatcherOptions struct {
+	ProcessName string
+	Files       []string
+	KubeConfig  string
+	Namespace   string
```
sttts: which namespace is this about? godoc

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+	Interval    time.Duration
+
+	recorder       events.Recorder
+	initialContent map[string][]byte
+
+	// for unit-test
+	getPIDByNameFn       func(name string) (int, bool, error)
+	signalFn             func(pid int) error
```
sttts: what is this? godoc

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+
+	return cmd
+}
+
+func (o *FileWatcherOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.ProcessName, "process-name", "", "name of the process to send signal to on file change (eg. 'hyperkube').")
+	fs.StringSliceVar(&o.Files, "files", o.Files, "comma separated list of file names to monitor for changes")
+	fs.StringVar(&o.KubeConfig, "kubeconfig", o.KubeConfig, "kubeconfig file or empty")
+	fs.StringVar(&o.Namespace, "namespace", o.Namespace, "namespace to report the watchdog events")
+	fs.DurationVar(&o.Interval, "interval", 5*time.Second, "interval specify how aggressive the file checks should be")
```
sttts: specifying

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(o.Files) == 0 {
+		return fmt.Errorf("at least one file to observe must be specified")
+	}
+	if len(o.Namespace) == 0 && len(os.Getenv("POD_NAMESPACE")) == 0 {
+		return fmt.Errorf("either namespace flag or POD_NAMESPACE environment variable must be specified")
+	}
+	return nil
+}
+
+// startPidObserver runs a loop that observe changes to the PID of the process we send signals after change is detected.
```
sttts: observes

sttts: changes to the PID?

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+// When the PID is observed for the first time, the pidObservedCh is closed.
+// When the PID change is observed (after we got initial PID), the shutdown is called (to signal watchdog to restart).
```
sttts: why do we restart? Isn't there a chance this way that the watchdog misses file changes?

* operand starts
* sidecar start
* secret changes
* sidecar kills operand
* sidecar terminates, operand terminates
* operand start
* secret changes again
* sidecar start
=> boom. Latest change not noticed.

sttts: tl/dr: I don't like the sidecar to restart at all. It shouldn't.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+			return
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+}
+
+// readInitialFileContent reads the content of files specified.
+// This is needed by file observer.
+func (o *FileWatcherOptions) readInitialFileContent() error {
```
sttts: how do we make sure that this is the state the operand sees? We can be late, and miss an update.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			return err
+		}
+		o.initialContent[name] = content
+	}
+	return nil
+}
+
+// setFilesPathPrefix mutates the file list and prefix every file with /proc/PID/root.
+func setFilesPathPrefix(oldFiles []string, pid int) []string {
```
sttts: this is not a setter. Call it `addProcPrefix` or something like that.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+	currentPID := <-pidObservedCh
+
+	// Mutate path for specified files as '/proc/PID/root/<original path>'
+	// This means side-car container don't have to duplicate the mounts from main container.
+	// This require shared PID namespace feature.
+	o.Files = o.setFilesPathPrefixFn(o.Files, currentPID)
+
+	// Read initial file content. If shared PID namespace does not work, this will error.
+	if err := o.readInitialFileContent(); err != nil {
+		o.recorder.Warningf("FileChangeWatchdogFailed", "Reading initial file content failed: %v", err)
```
sttts: does this really need events? It's a misconfiguration. I think logs are enough.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+		return fmt.Errorf("unable to read initial file content: %v", err)
+	}
+
+	o.recorder.Eventf("FileChangeWatchdogStarted", "Started watching files for process %s[%d]", o.ProcessName, currentPID)
```
sttts: these is are too many events. I only see one event we need: "file xyz changed, restarting operand".

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,247 @@
...
+		o.recorder.Warningf("ObserverFailed", "Failed to start to file observer: %v", err)
+		klog.Fatalf("unable to start file observer: %v", err)
+	}
+
+	observer.AddReactor(func(file string, action fileobserver.ActionType) error {
+		o.recorder.Eventf("FileChangeObserved", "Observed change in file %q, sending SIGINT to %s[%d]", file, o.ProcessName, currentPID)
+		// When file change is observed, don't give up on errors when we fail to send signal, but retry sending
+		// and fail hard if we fail to deliver it.
+		retries := 0
+		return wait.PollImmediate(1*time.Second, 5*time.Second, func() (done bool, err error) {
```
sttts: what are these intervals? Why those numbers?

pkg/operator/watchdog/proc.go
```diff
@@ -0,0 +1,52 @@
...
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// GetPidByName find the process name specified by name and return the PID of that process.
+// If the process is not found, the bool is false.
+// NOTE: This require container with shared process namespace (if run as side-car).
+func GetPidByName(name string) (int, bool, error) {
```
sttts: FindPid?

pkg/operator/watchdog/cmd.go
mfojtik: thought the `for unit-test` applies to all below ;-)

mfojtik: i will add them

pkg/operator/watchdog/cmd.go
mfojtik: yeah, when the PID change this will shutdown() or send the pid to the control logic

pkg/operator/watchdog/cmd.go
mfojtik: I'm more worried about the watchdog to be "stucked" or reading the files from wrong directory on PID change. If we don't want to restart, it means we have to make the logic in the `Run()` more complex and less error prone.

Why you hate restarts?

pkg/operator/watchdog/cmd.go
mfojtik: If we restart, yes. Practically I don't see a big chance to miss a change (we rotate certs on hour basis). IF this will be monitoring configs, then yes.

We can discuss restart vs. monitor via Slack.

pkg/operator/watchdog/cmd.go
mfojtik: sure

pkg/operator/watchdog/cmd.go
mfojtik: if we restart/crash the log is gone, don't we learned our lessons ? :-)

pkg/operator/watchdog/cmd.go
mfojtik: Without aggregated logging, the events are the only way we can discover problems or breakages. If I can get logs from all watchdog containers (even the ones that died), I would remove events.

pkg/operator/watchdog/cmd.go
mfojtik: I will add godoc (make 5 attempts to SIGINT, wait 1s between them... This might not be needed, but i'm skeptical that things will always work and this is a critical path for watchdog, hence the retry.

pkg/operator/watchdog/proc.go
mfojtik: It is by using name... golang already have funcs to get pid, but not by process name... `FindPidByName` ?

sttts: FindPidByName is good

pkg/operator/watchdog/cmd.go
sttts: Feels like we misuse events as logs. Events should be user relevant. But our events are for us.

pkg/operator/watchdog/cmd.go
sttts: restarts mean races. I don't like races.

pkg/operator/watchdog/cmd.go
sttts: it's more like `handleFileChangeFn(pid int)`, right?

pkg/operator/watchdog/cmd.go
mfojtik: Who is the user in this case? I think the target audience for these events are "admins" who are trying to figure out why their operands are not restarting? :-)

pkg/operator/watchdog/cmd.go
mfojtik: lets go with the monitor for now (no crashing)

pkg/operator/watchdog/cmd.go
mfojtik: ok, lets go with no restart, I updated the code to run in loop, contexts are fun.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,295 @@
...
+	if len(o.Namespace) == 0 && len(os.Getenv("POD_NAMESPACE")) == 0 {
+		return fmt.Errorf("either namespace flag or POD_NAMESPACE environment variable must be specified")
+	}
+	return nil
+}
+
+// startPidObserver runs a loop that observes changes to the PID of the process we send signals after change is detected.
+// When the PID is observed for the first time, the pidObservedCh is closed.
+// When the PID change is observed (after we got initial PID), the shutdown is called (to signal watchdog to restart).
+func (o *FileWatcherOptions) startPidObserver(ctx context.Context, pidObservedCh chan int) {
```
sttts: "start" sounds like this will not block.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// Namespace to report events to
+	Namespace string
+	recorder  events.Recorder
+
+	// Interval specifies how aggressive we want to be in file checks
+	Interval time.Duration
+
+	// for unit-test to mock getting the process PID
+	handleFindPidByNameFn func(name string) (int, bool, error)
```
sttts: no need for the "handle" prefix.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// for unit-test to mock sending UNIX signals
+	handleSignalFn func(pid int) error
+
+	// for unit-test to mock prefixing files (/proc/PID/root)
+	handleAddProcPrefixToFilesFn func([]string, int) []string
```
sttts: no need for the "handle" prefix

pkg/operator/watchdog/cmd.go
mfojtik: will replace with `run`

pkg/operator/watchdog/proc.go
```diff
@@ -0,0 +1,57 @@
...
+		return pid, true, nil
+	}
+	return 0, false, nil
+}
+
+// readlink is copied from the os.Readlink() but does not return error when the target path does not exists.
+// This is used to read broken links as in case of share PID namespace, the /proc/1/exe points to a binary
+// that does not exists from the source container.
+func readlink(name string) string {
+	for l := 128; ; l *= 2 {
```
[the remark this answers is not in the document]
mfojtik: @soltysh will `os.Readlink` return the link target AND the error to ignore? i remember I tried several ways and all of them returned empty string :-(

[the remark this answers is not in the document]
mfojtik: added unit test to cover all cases we need readlink() for... if it breaks (on any platform) we will know.

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,316 @@
...
+		},
+	}
+
+	o.AddFlags(cmd.Flags())
+
+	return cmd
+}
+
+func (o *FileWatcherOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.ProcessName, "process-name", "", "name of the process to send signal to on file change (eg. 'hyperkube').")
```
sttts: which signal? And only once?

mfojtik: do we want to give user choice? it is TERM

pkg/operator/watchdog/proc.go
```diff
@@ -0,0 +1,57 @@
...
+			continue
+		}
+		return pid, true, nil
+	}
+	return 0, false, nil
+}
+
+// readlink is copied from the os.Readlink() but does not return error when the target path does not exists.
+// This is used to read broken links as in case of share PID namespace, the /proc/1/exe points to a binary
+// that does not exists from the source container.
```
sttts: omg

mfojtik: https://github.com/golang/go/pull/31994 :-)

soltysh: Yeah, I've had very long discussion with @mfojtik about it, but there's no way, for now. He opened a go issue, iirc.

pkg/operator/watchdog/cmd.go
sttts: no, but document :)

SIGTERM is pretty aggressive. I.e. not graceful shutdown. Intentional?

sttts: normally: SIGINT, wait seconds, SIGTERM. A configurable delay would be good (again, graceful shutdown).

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,320 @@
...
+		pollErr := wait.PollImmediate(1*time.Second, 5*time.Second, func() (done bool, err error) {
+			// we already signalled this PID to terminate and the process is being gracefully terminated now.
+			// do not send the signal to the same PID twice, but wait for the new PID to appear.
+			if currentPID == o.lastUsedPid {
+				return true, nil
+			}
+			defer func() {
+				shutdown()
+			}()
+			o.recorder.Eventf("FileChangeObserved", "Observed change in file %q, sending SIGINT to %s[%d]", file, o.ProcessName, currentPID)
```
sttts: INT or TERM?

sttts: not that the API server will terminate non-gracefully on the second signal. Does this mean it has only 1 second to terminate gracefully?

mfojtik: I added graceful shutdown sequence

pkg/operator/watchdog/cmd.go
```diff
@@ -0,0 +1,345 @@
...
+	observer, err := fileobserver.NewObserver(o.Interval)
+	if err != nil {
+		o.recorder.Warningf("ObserverFailed", "Failed to start to file observer: %v", err)
+		return fmt.Errorf("unable to start file observer: %v", err)
+	}
+
+	observer.AddReactor(func(file string, action fileobserver.ActionType) error {
+		// We already signalled this PID to terminate and the process is being gracefully terminated now.
+		// Do not duplicate termination process for PID we already terminated, but wait for the new PID to appear.
+		if currentPID == o.lastTerminatedPid {
```
sttts: add something here that PIDs change on restart.

sttts:
unit test

PR #393, opened 2019-05-06
expose run-once and wait-for-ready options on cert rotation

deads2k:
This is needed for cert recovery.

### DIFF
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -113,16 +113,29 @@ func (c CertRotationController) syncWorker() error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -113,16 +113,29 @@ func (c CertRotationController) syncWorker() error {
 	return nil
 }
 
+func (c *CertRotationController) WaitForReady(stopCh <-chan struct{}) {
+	klog.Infof("Waiting for CertRotationController - %q", c.name)
+	defer klog.Infof("Waiting for CertRotationController - %q", c.name)
```
tnozicka: finished waiting?

tnozicka: maybe mention caches in there somewhere

PR #395, opened 2019-05-13
controllercmd: restart operator when service account ca.crt change

mfojtik:
When kubelet change the ca.crt inside container, the operator file observer should see that change and automatically trigger restart to reload the certificate.

proof: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/253


mfojtik:
/hold

we probably don't need this as the CA is valid for 10y.

PR #396, opened 2019-05-14
operator/condition: add initial list of condition names

mfojtik:
Collect all conditions that are spread across different controller loops and concentrate them in one place with proper godoc.

### DIFF
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -4,14 +4,16 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/klog"
-
-	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/condition"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


pkg/operator/condition/condition.go
```diff
@@ -0,0 +1,56 @@
+package condition
+
+const (
+	// ManagementStateDegradedConditionType represents a condition when the operator management state is changed from Managed to other value.
```
sttts: is tue when ...

sttts: Would not formulate it in a differential way. Better: is true if the ManagementState is not "Managed".

pkg/operator/condition/condition.go
```diff
@@ -0,0 +1,56 @@
...
+	// Possible reasons are Unmanaged, Removed or Unknown. Any of these cases means the operator is not actively managing the operand.
+	ManagementStateDegradedConditionType = "ManagementStateDegraded"
+
+	// UnsupportedConfigOverridesUpgradeableConditionType represents a condition when operator unsupported config overrides is changed.
```
sttts: is true if UnsupportedConfigOverrides is set.

mfojtik: it applies to all, I will replace

pkg/operator/condition/condition.go
```diff
@@ -0,0 +1,56 @@
+package condition
+
+const (
```
sttts: Usually we introduce a condition type (string alias) and type the constant that way.

mfojtik: @sttts right, but it will require openshift/api type and change in the operator status? or type casting on usage with sucks.

pkg/operator/condition/condition.go
```diff
@@ -0,0 +1,56 @@
...
+	ResourceSyncControllerDegradedConditionType = "ResourceSyncControllerDegraded"
+
+	// CertRotationDegradedConditionType represents a condition when the operator failed to properly rotate one or more certificates required by the operand.
+	// The RotationError reason is given with message describing details of this failure. This condition can be fatal when ignored as the existing certificate(s)
+	// validity can expire and without rotating/renewing them manual recovery might be required to fix the cluster.
+	CertRotationDegradedConditionType = "CertRotation_%s_Degraded"
+
+	// InstallerControllerDegradedConditionType represents a condition when the operator is not able to create new installer pods so the new revisions
+	// cannot be rolled out. This might happen when one or more required secrets or config maps does not exists.
+	InstallerControllerDegradedConditionType = "InstallerControllerDegraded"
```
sttts: ... is not able to create ... is differential again, some event. Please describe also in those cases, when the condition is set back to false again.

mfojtik: added the False case to godoc (for others as well).

mfojtik:
@sttts thanks! updated godoc.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -79,16 +81,16 @@ func NewCertRotationController(
 func (c CertRotationController) sync() error {
 	syncErr := c.syncWorker()
 
-	condition := operatorv1.OperatorCondition{
-		Type:   "CertRotation_" + c.name + "_Degraded",
+	newCondition := operatorv1.OperatorCondition{
+		Type:   fmt.Sprintf(condition.CertRotationDegradedConditionType, c.name),
```
sttts: rename the constant to ...Fmt

PR #397, opened 2019-05-14
Rebase 1.14.0

soltysh:


glide.yaml
```diff
@@ -1,27 +1,26 @@
...
   version: master
 
 # crd-schema-gen
   # TODO: we need to this to get nullable patch, but we will replace this with new repo soon.
-- package: k8s.io/apiextensions-apiserver
-  repo: https://github.com/openshift/kubernetes-apiextensions-apiserver
-  version: origin-4.1-kubernetes-1.13.4
 - package: sigs.k8s.io/controller-tools
-  repo: https://github.com/openshift/kubernetes-sigs-controller-tools
+  repo:    https://github.com/openshift/kubernetes-sigs-controller-tools
```
sttts: nit: no alignment in yaml

pkg/controller/metrics/workqueue_metrics.go
```diff
@@ -13,78 +14,195 @@ func init() {
...
+		Name:        DepthKey,
+		Help:        "Current depth of workqueue",
+		ConstLabels: prometheus.Labels{"name": name},
 	})
 	prometheus.Register(depth)
 	return depth
 }
 
 func (prometheusMetricsProvider) NewAddsMetric(name string) workqueue.CounterMetric {
+	adds := prometheus.NewCounter(prometheus.CounterOpts{
```
sttts: Who owns these metrics? Surprising amount of changes.

soltysh: These changed upstream, the reason there are this many is that the old ones are being deprecated and we're switching to new ones right away.

glide.yaml
soltysh: I'll fix that in the next bump to newer 1.14.x

PR #398, opened 2019-05-15
Add GetDefaultMTU function

rcarrillocruz:
That function is currently in CNO tree, however we want to use it
on origin too, thus libifying it on library-go to avoid duplications
of code.


danwinship:
You need to pass `--strip-vendor` to `glide`. Or rather, just use `make update-deps`.

glide.yaml
```diff
@@ -57,3 +57,6 @@ import:
 # matches openshift/origin
 - package: github.com/gonum/graph
   version: 50b27dea7ebbfb052dfaf91681afc6fde28d8796
+
+- package: github.com/vishvananda/netlink
```
smarterclayton: This dependency really doesn’t belong in library-go.  Pulling this dependency into library go to reuse a 20 line function is overkill

smarterclayton:
/hold

I don’t think this belongs in library go.  A 20 line function is fine to copy

danwinship:
ok

PR #399, opened 2019-05-16
fileobserver: allow to specify start time for file observer

mfojtik:
In some cases, we need to trigger a modified reaction when the file modification timestamp is never than a process consuming the modification. Such case exists when a file watch dog is not running yet (image pull?) but the main process is already started. In that case, there is some time when we don't react to file changes (because we don't run).

If we are able to read the timestamp of the process and use that timestamp when initializing the observer, we might trigger when the files we observing are "newer" than that process.

### DIFF
--- a/pkg/controller/fileobserver/observer.go
+++ b/pkg/controller/fileobserver/observer.go
@@ -57,3 +57,12 @@ func NewObserver(interval time.Duration) (Observer, error) {
 		files:    map[string]string{},
 	}, nil
 }
+
+func NewObserverWithStartTime(interval time.Duration, startTime time.Time) (Observer, error) {
+	observer, err := NewObserver(interval)
+	if err != nil {
+		return nil, err
+	}
+	observer.(*pollingObserver).startTimestamp = &startTime
+	return observer, nil
+}
--- a/pkg/controller/fileobserver/observer_polling.go
+++ b/pkg/controller/fileobserver/observer_polling.go
@@ -19,6 +19,11 @@ type pollingObserver struct {
 	reactors map[string][]reactorFn
 	files    map[string]string
 
+	// If files are newer then startTimestamp the observer will report a change for them.
+	// This exists to mitigate issue when we start observing files but have to synchronize
+	// the observing with the process consuming the files.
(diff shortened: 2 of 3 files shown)


pkg/controller/fileobserver/observer.go
```diff
@@ -57,3 +57,12 @@ func NewObserver(interval time.Duration) (Observer, error) {
 		files:    map[string]string{},
 	}, nil
 }
+
+func NewObserverWithStartTime(interval time.Duration, startTime time.Time) (Observer, error) {
+	return &pollingObserver{
+		interval:       interval,
+		reactors:       map[string][]reactorFn{},
+		files:          map[string]string{},
+		startTimestamp: &startTime,
```
sttts: reuse the other constructor. Otherwise they risk to diverge.

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -19,6 +19,11 @@ type pollingObserver struct {
 	reactors map[string][]reactorFn
 	files    map[string]string
 
+	// If files are newer then startTimestamp the observer will report a change for them.
+	// This exists to mitigate issue when we start observing files but have to synchronize
+	// the observing with the process consuming the files.
+	startTimestamp *time.Time
```
sttts: you can avoid the pointer if you use the zero value instead. Isn't there `time.IsZero`?

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -70,8 +75,13 @@ func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
 			}
 			lastKnownHash := o.files[filename]
 
+			modifiedBeforeStartTime, err := o.isModifiedBeforeStartTimestamp(filename)
```
sttts: would use the inverse: modifiedAfterStartup

pkg/controller/fileobserver/observer_polling.go
mfojtik: there is no `IsZero`, i was thinking about that as well, but have not find a nice way to do this.

PR #404, opened 2019-05-26
staticpod: prune installer pods for removed masters

mfojtik:
This fixes an edge case when the master node is removed but some if the installer pods for that master were in "new" or "pending" state. In such case, the installer pod will never succeed because the node selector won't match any node.

The change here will cause all installer pods for the removed node being deleted when we observe master node removed (all pods, not just pending, as there is no reason keeping them around?).

/cc @deads2k 
/cc @sanchezl

### DIFF
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -4,16 +4,20 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -86,6 +97,12 @@ func (c NodeController) sync() error {
 		if found {
 			newTargetNodeStates = append(newTargetNodeStates, originalOperatorStatus.NodeStatuses[i])
 		} else {
+			// Prune installer pods we created for this node. In case these are "pending" they will never go to running
+			// as their nodeName won't exists.
+			if err := c.pruneInstallerPodsForRemovedNode(nodeState.NodeName); err != nil {
+				c.eventRecorder.Warningf("MasterNodePruningFailed", "Failed to prune pods for removed master node %s: %v", nodeState.NodeName, err)
+				return err
```
mfojtik: @deads2k not sure if this is worth erroring out or should be just best-effort.

deads2k:
Why don't care about a pending pod?  Seems like we can leave that clean up to a cluster admin if they wish.  Are we waiting on it somewhere?  A wait would indicate that our list of nodes isn't updating properly

pkg/operator/staticpod/controller/node/node_controller.go
sanchezl: I think this should be best effort.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -116,6 +133,27 @@ func (c NodeController) sync() error {
...
+	if err != nil {
+		return err
+	}
+	if len(pods.Items) == 0 {
+		return nil
+	}
+	for _, pod := range pods.Items {
+		err := c.podsGetter.Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
```
sanchezl: Just log the inability to delete the pod and continue.

PR #405, opened 2019-05-28
add authorization helper functions

deads2k:
move of pkg/authorization/util  used by CLI, controllers, and apiserver

### DIFF
--- a/pkg/authorization/authorizationutil/subject.go
+++ b/pkg/authorization/authorizationutil/subject.go
@@ -1,0 +1,56 @@
+package authorizationutil
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/authorization/authorizationutil/subject.go
```diff
@@ -0,0 +1,56 @@
+package authorizationutil
```
sttts: wow, authorizationutil. Why this name?

mfojtik: i would just call it pkg/authorization

deads2k: > wow, authorizationutil. Why this name?

auto-completes, like `http/httptest` (when they realized their advice was bad).  this particular one is often abbreviated in our codebase today.

PR #413, opened 2019-05-29
pkg/cmdutil: introduce cmdutil

mfojtik:
Used by `oc` and `openshift-apiserver`. Clean extraction from `pkg/cmd/util`.

/cc @deads2k 
/cc @soltysh


pkg/cmdutil/cmd.go
```diff
@@ -0,0 +1,80 @@
...
+	if absName, err := filepath.Abs(filename); err == nil {
+		return absName
+	}
+
+	return filename
+}
+
+// ResolveResource returns the resource type and name of the resourceString.
+// If the resource string has no specified type, defaultResource will be returned.
+func ResolveResource(defaultResource schema.GroupResource, resourceString string, mapper meta.RESTMapper) (schema.GroupResource, string, error) {
```
deads2k: this method is only used by `oc`

pkg/cmdutil/cmd.go
```diff
@@ -0,0 +1,80 @@
+package util
```
deads2k: looks like this whole file is only used in oc

pkg/cmdutil/crypto.go
```diff
@@ -0,0 +1,73 @@
...
+	"bytes"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+
+	"k8s.io/client-go/util/cert"
+)
+
+func CertPoolFromFile(filename string) (*x509.CertPool, error) {
```
deads2k: this is only used by an integration test

pkg/cmdutil/cmd.go
soltysh: There are a lot of stuff here, and majority is somewhat oc related, I'd not move it, just yet.

PR #418, opened 2019-05-29
add name generation methods

deads2k:
pkg/api/apihelpers  used by controllers and newapp

/assign @soltysh

### DIFF
--- a/pkg/build/naming/namer.go
+++ b/pkg/build/naming/namer.go
@@ -1,0 +1,73 @@
+package naming
+
+import (
+	"fmt"
+	"hash/fnv"
+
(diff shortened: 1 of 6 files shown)


pkg/build/naming/namer.go
```diff
@@ -0,0 +1,73 @@
+package naming
+
+import (
+	"fmt"
+	"hash/fnv"
+
+	kvalidation "k8s.io/apimachinery/pkg/util/validation"
+)
+
+// GetName returns a name given a base ("deployment-5") and a suffix ("deploy")
```
soltysh: Good, I'll be able to get rid of the nasty copy I've created temporarily in oc.

PR #419, opened 2019-05-30
Add image utils

soltysh:
This is very small fraction of image helpers that is being shared between apiserver, controllers and oc.

/assign @deads2k

### DIFF
--- a/pkg/image/imageutil/helpers.go
+++ b/pkg/image/imageutil/helpers.go
@@ -1,0 +1,49 @@
+package imageutil
+
+import (
+	"fmt"
+	"strings"
+)
+
+const (
+	// DefaultImageTag is used when an image tag is needed and the configuration does not specify a tag to use.
+	DefaultImageTag = "latest"
+)
+
+// SplitImageStreamTag turns the name of an ImageStreamTag into Name and Tag.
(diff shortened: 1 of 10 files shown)


pkg/image/helpers.go
```diff
@@ -0,0 +1,49 @@
+package image
```
deads2k: move to a subpackage.  even pkg/image/imageutil is swallowable.

soltysh: Done

pkg/image/trigger/pods.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return nil, errNoImageMutator
+}
+
+var errNoPodSpec = fmt.Errorf("No PodSpec available for this object")
+
+// GetPodSpecV1 returns a mutable pod spec out of the provided object, including a field path
+// to the field in the object, or an error if the object does not contain a pod spec.
+// This only returns pod specs for v1 compatible objects.
+func GetPodSpecV1(obj runtime.Object) (*corev1.PodSpec, *field.Path, error) {
```
deads2k: keep in package like imagereferencemutator.  We want these contained.

PR #425, opened 2019-05-31
Add deployment utils

soltysh:
/assign @mfojtik @deads2k


soltysh:
Better version in #426.

PR #430, opened 2019-05-31
start informers after the listers have been requested

deads2k:
also wait for the caches to fill

/assign @sanchezl 

/hold
@sanchezl please make a fake bump

### DIFF
--- a/pkg/operator/staticpod/certsyncpod/certsync_cmd.go
+++ b/pkg/operator/staticpod/certsyncpod/certsync_cmd.go
@@ -66,10 +66,8 @@ func (o *CertSyncControllerOptions) Run() error {
 	observer.AddReactor(fileobserver.ExitOnChangeReactor, map[string][]byte{o.KubeConfigFile: initialContent}, o.KubeConfigFile)
 
 	stopCh := make(chan struct{})
-	go observer.Run(stopCh)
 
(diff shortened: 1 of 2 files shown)


sanchezl:
/lgtm 

Tested in CI via https://github.com/openshift/cluster-kube-apiserver-operator/pull/490
Confirmed on patched install that this fixes https://bugzilla.redhat.com/show_bug.cgi?id=1714771

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
```diff
@@ -90,7 +88,11 @@ func (o *CertSyncControllerOptions) Run() error {
 	if err != nil {
 		return err
 	}
+
+	// start everything. Informers start after they have been requested.
 	go controller.Run(1, stopCh)
+	go observer.Run(stopCh)
+	go kubeInformers.Start(stopCh)
```
p0lyn0mial: nit: I think you don't have to start Kube informers in a separate "thread" because all informers will be run in a dedicated goroutine, see https://github.com/openshift/library-go/blob/5dbeda57c0d97b335dd66bff593d0f08bc8cadf3/vendor/k8s.io/client-go/informers/factory.go#L132

maybe we could change the name of the method to `StartAsync` :)

PR #436, opened 2019-06-04
Add missing build name helper

soltysh:
This helper function exists in 2 copies (sic!) and is heavily used across apiserver, controller manager and cli.
/assign @deads2k @mfojtik


soltysh:
Discussed on slack, we don't want that.

PR #441, opened 2019-06-06
Add image util used by cli and apiserver

soltysh:
/assign @mfojtik


soltysh:
I'm going to self-approve this one, since this is relatively small fix.

PR #444, opened 2019-06-06
staticpod-controller: introduce reason strings why nodes were chosen

sttts:


sttts:
`2019/06/06 15:32:04 Container test in pod unit failed, exit code 2, reason OOMKilled`

/retest

PR #448, opened 2019-06-07
 snip links between build controller and build apiserver

deads2k:
these helpers are shared between OAS and OCM

/hold

holding for a fake bump


deads2k:
Proof is green

PR #450, opened 2019-06-10
add helpers shared with e2e tests

mfojtik:
These two are needed by e2e tests and having them here avoids copying.

### DIFF
--- a/pkg/image/imageutil/helpers.go
+++ b/pkg/image/imageutil/helpers.go
@@ -1,6 +1,7 @@
 package imageutil
 
 import (
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/image/imageutil/helpers.go
```diff
@@ -200,3 +204,135 @@ func PrioritizeTags(tags []string) {
...
+		return &t.Items[0]
+	}
+
+	return nil
+}
+
+// ImageWithMetadata mutates the given image. It parses raw DockerImageManifest data stored in the image and
+// fills its DockerImageMetadata and other fields.
+// Copied from github.com/openshift/image-registry/pkg/origin-common/util/util.go
+func ImageWithMetadata(image *imagev1.Image) error {
```
soltysh: Can we just return that `dockerv10.DockerImage` instead of filing it in back? That'd save us from those nasty casts we do all over the place?

soltysh:
/hold
until I get my question answered

deads2k:
/hold cancel

@soltysh Add a parallel method and try to  switch our codebase.  You're right, but images are a scary place.

soltysh:
@damemi @sallyom either of you two please fix my comment and apply the changes in origin afterwards.

deads2k:
> @damemi @sallyom either of you two please fix my comment and apply the changes in origin afterwards.

I nominate @sallyom.  @damemi is in test-integration hell.

PR #453, opened 2019-06-10
reports `NotReady` status of master nodes

p0lyn0mial:
extends node controller to examine `Node.Status.Ready` field and to report unhealthy status.
the status is reported by the controller as a new condition of type `NodeControllerDegraded` that is added to `StaticPodOperatorStatus`

closes https://jira.coreos.com/browse/MSTR-404

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -35,6 +35,7 @@ var versions = map[string]uint16{
 	"VersionTLS10": tls.VersionTLS10,
 	"VersionTLS11": tls.VersionTLS11,
 	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
 }
 
 // TLSVersionToNameOrDie given a tls version as an int, return its readable name
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -2,13 +2,10 @@ package node
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
-	"github.com/openshift/library-go/pkg/operator/v1helpers"
-
-	"k8s.io/klog"
-
-	"k8s.io/apimachinery/pkg/api/equality"
+	coreapiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -17,9 +14,12 @@ import (
 	corelisterv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/condition"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 const nodeControllerWorkQueueKey = "key"
@@ -59,11 +59,10 @@ func NewNodeController(
 }
 
 func (c NodeController) sync() error {
-	_, originalOperatorStatus, resourceVersion, err := c.operatorClient.GetStaticPodOperatorState()
+	_, originalOperatorStatus, _, err := c.operatorClient.GetStaticPodOperatorState()
 	if err != nil {
 		return err
 	}
-	operatorStatus := originalOperatorStatus.DeepCopy()
 
 	selector, err := labels.NewRequirement("node-role.kubernetes.io/master", selection.Equals, []string{""})
 	if err != nil {
@@ -106,13 +105,52 @@ func (c NodeController) sync() error {
 		newTargetNodeStates = append(newTargetNodeStates, operatorv1.NodeStatus{NodeName: node.Name})
 	}
 
-	operatorStatus.NodeStatuses = newTargetNodeStates
-	if !equality.Semantic.DeepEqual(originalOperatorStatus, operatorStatus) {
-		if _, updateError := c.operatorClient.UpdateStaticPodOperatorStatus(resourceVersion, operatorStatus); updateError != nil {
-			return updateError
+	// detect and report master nodes that are not ready
+	notReadyNodes := []string{}
+	for _, node := range nodes {
+		for _, con := range node.Status.Conditions {
+			if con.Type == coreapiv1.NodeReady && con.Status != coreapiv1.ConditionTrue {
+				notReadyNodes = append(notReadyNodes, node.Name)
+			}
 		}
 	}
+	newCondition := operatorv1.OperatorCondition{
+		Type: condition.NodeControllerDegradedConditionType,
+	}
+	if len(notReadyNodes) > 0 {
+		newCondition.Status = operatorv1.ConditionTrue
+		newCondition.Reason = "MasterNodesReady"
+		newCondition.Message = fmt.Sprintf("The master node(s) %q not ready", strings.Join(notReadyNodes, ","))
+	} else {
+		newCondition.Status = operatorv1.ConditionFalse
+		newCondition.Reason = "MasterNodesReady"
+		newCondition.Message = "All master node(s) are ready"
+	}
 
+	oldStatus := &operatorv1.StaticPodOperatorStatus{}
+	_, updated, updateError := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition), func(status *operatorv1.StaticPodOperatorStatus) error {
+		status.NodeStatuses = newTargetNodeStates
+		return nil
+	}, func(status *operatorv1.StaticPodOperatorStatus) error {
+		//a hack for storing the old status (before the update)
+		oldStatus = status
+		return nil
+	})
+
+	if updateError != nil {
+		return updateError
+	}
+
+	if !updated {
+		return nil
+	}
+
+	for _, oldCondition := range oldStatus.Conditions {
+		if oldCondition.Type == condition.NodeControllerDegradedConditionType && oldCondition.Message != newCondition.Message {
+			c.eventRecorder.Eventf("MasterNodesReadyChanged", newCondition.Message)
+			break
+		}
+	}
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,14 +105,33 @@ func (c NodeController) sync() error {
 		newTargetNodeStates = append(newTargetNodeStates, operatorv1.NodeStatus{NodeName: node.Name})
 	}
 
-	operatorStatus.NodeStatuses = newTargetNodeStates
-	if !equality.Semantic.DeepEqual(originalOperatorStatus, operatorStatus) {
-		if _, updateError := c.operatorClient.UpdateStaticPodOperatorStatus(resourceVersion, operatorStatus); updateError != nil {
-			return updateError
+	// detect and report master nodes that are not ready
+	notReadyNodes := []string{}
+	for _, node := range nodes {
```
p0lyn0mial: not sure if this is the best place but it plugs nicely into an existing flow.

p0lyn0mial:
unhealthy nodes set `Degraded` of `clusteroperator`  to `True`, for example:

```
kind: ClusterOperator
...
status:
  conditions:
  - lastTransitionTime: "2019-06-10T21:21:07Z"
    message: 'NodeControllerDegraded: The master node(s) [ip-10-0-151-95.ec2.internal]
      not ready'
    reason: NodeControllerDegradedNotReady
    status: "True"
    type: Degraded
...
```

p0lyn0mial:
I will add a missing unit test later.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,14 +105,33 @@ func (c NodeController) sync() error {
...
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "NotReady"
+	    cond.Message = fmt.Sprintf("The master node(s) %v not ready", notReadyNodes)
+	} else {
+	    cond.Status = operatorv1.ConditionFalse
+		cond.Reason = "Ready"
+		cond.Message = "All master nodes are ready"
+	}
+	_, _, updateError := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(cond), func(status *operatorv1.StaticPodOperatorStatus) error{
+		status.NodeStatuses = newTargetNodeStates
```
p0lyn0mial: alternatively, we could extend `[]NodeStatus` and change the other controllers to respect it.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,14 +105,33 @@ func (c NodeController) sync() error {
...
 	}
-
-	return nil
+	cond := operatorv1.OperatorCondition{
+		Type:   condition.NodeControllerDegradedConditionType,
+	}
+	if len(notReadyNodes) > 0 {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "NotReady"
+	    cond.Message = fmt.Sprintf("The master node(s) %v not ready", notReadyNodes)
```
sttts: gofmt

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,14 +105,33 @@ func (c NodeController) sync() error {
...
-	if !equality.Semantic.DeepEqual(originalOperatorStatus, operatorStatus) {
-		if _, updateError := c.operatorClient.UpdateStaticPodOperatorStatus(resourceVersion, operatorStatus); updateError != nil {
-			return updateError
+	// detect and report master nodes that are not ready
+	notReadyNodes := []string{}
+	for _, node := range nodes {
+		for _, con := range node.Status.Conditions {
+			if con.Type == coreapiv1.NodeReady && con.Status != coreapiv1.ConditionTrue {
+				notReadyNodes = append(notReadyNodes, node.Name)
+				c.eventRecorder.Eventf("MasterNodeNotReady", "the master node %s is not ready", node.Name)
```
sttts: does this need comparison to an existing condition? We are creating an event every time we run this sync method.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,14 +105,33 @@ func (c NodeController) sync() error {
...
+	cond := operatorv1.OperatorCondition{
+		Type:   condition.NodeControllerDegradedConditionType,
+	}
+	if len(notReadyNodes) > 0 {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "NotReady"
+	    cond.Message = fmt.Sprintf("The master node(s) %v not ready", notReadyNodes)
+	} else {
+	    cond.Status = operatorv1.ConditionFalse
+		cond.Reason = "Ready"
```
sttts: shouldn't we use the same condition reason, but set status to false?

p0lyn0mial: yes, good idea.

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: alright, will remove it.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,10 +105,48 @@ func (c NodeController) sync() error {
...
+				notReadyNodes = append(notReadyNodes, node.Name)
+			}
+		}
+	}
+	newCondition := operatorv1.OperatorCondition{
+		Type: condition.NodeControllerDegradedConditionType,
+	}
+	if len(notReadyNodes) > 0 {
+		newCondition.Status = operatorv1.ConditionTrue
+		newCondition.Reason = "Ready"
```
mfojtik: @sttts `Ready` as reason sounds strange here... can we make it explicit? `MastersReady` perhaps?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,10 +105,48 @@ func (c NodeController) sync() error {
...
+		newCondition.Message = fmt.Sprintf("The master node(s) %v not ready", notReadyNodes)
```
mfojtik: can we pretty-print this? this might appear in higher level (console), the Go slice might look weird? `strings.Join(notReadyNodes, ",")` and `%q` ?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,10 +105,48 @@ func (c NodeController) sync() error {
...
+		//a hack for storing the old status (before the update)
+		oldStatus = status
+		return nil
+	})
+
+	if updateError != nil {
+		return updateError
+	}
+
+	if updated {
```
mfojtik:
```
if !updated {
  return nil
}
```

nit: decrease the level of nesting and makes it read better ;-)

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -106,10 +105,48 @@ func (c NodeController) sync() error {
...
+		for _, oldCondition := range oldStatus.Conditions {
+			if oldCondition.Type == condition.NodeControllerDegradedConditionType && oldCondition.Message != newCondition.Message {
+				c.eventRecorder.Eventf("MasterNodeReadyChanged", newCondition.Message)
```
mfojtik: `MasterNodeReady` sounds like a good choice for the "reason" comment above

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
mfojtik: nit: we use to call this `corev1` :-) also this alias is redundant

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -26,6 +26,129 @@ func fakeMasterNode(name string) *v1.Node {
...
+		{
+			name:        "scenario 1: one unhealthy master node is reported",
+			masterNodes: []runtime.Object{makeNodeNotReady(fakeMasterNode("test-node-1")), fakeMasterNode("test-node-2")},
+			evaluateNodeStatus: func(conditions []operatorv1.OperatorCondition) error {
+				if len(conditions) != 1 {
+					return fmt.Errorf("expected exaclty 1 condition, got %d", len(conditions))
+				}
+
+				condition := conditions[0]
+				if condition.Type != "NodeControllerDegraded" {
```
mfojtik: why not use `NodeControllerDegradedConditionType` ?

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: yes, I like that.

pkg/operator/staticpod/controller/node/node_controller_test.go
p0lyn0mial: because I don't like to use "production code" in the tests :) Anyway I'll change that :)

pkg/operator/staticpod/controller/node/node_controller_test.go
p0lyn0mial: will change to `corev1`

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: decided to change to `MasterNodesReady`

pkg/crypto/crypto.go
```diff
@@ -35,6 +35,7 @@ var versions = map[string]uint16{
 	"VersionTLS10": tls.VersionTLS10,
 	"VersionTLS11": tls.VersionTLS11,
 	"VersionTLS12": tls.VersionTLS12,
+	"VersionTLS13": tls.VersionTLS13,
```
p0lyn0mial: @deads2k this won't make you happy.

 I updated go version to `1.12` and that made `TestConstantMaps` test to fail. It seems like a similar test exists upstream https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag_test.go#L77 which is broken as well but only if you run that test manually. Running via `make tests` don't reveal anything. I'll try to fix that.

mfojtik:
/lgtm
/hold

(for @deads2k )

pkg/crypto/crypto.go
deads2k: @mrogers950 anything we should be worried about?

deads2k:
/lgtm
/hold cancel

@mrogers950 we'll revert if there's a need.

PR #454, opened 2019-06-11
Add console server config types

jhadvig:
We need to share these types in:
- https://github.com/openshift/console
- https://github.com/openshift/console-operator (@benjaminapetersen is adding it ATM in https://github.com/openshift/console-operator/pull/235)

Was thinking about putting it into `openshift/api` repo but not sure about that since it's a config type and may change over type.

@spadgett @benjaminapetersen PTAL

@mfojtik fyi

### DIFF
--- a/pkg/console/config/types.go
+++ b/pkg/console/config/types.go
@@ -1,0 +1,56 @@
+package config
+
+// Config is the top-level console configuration.
+type Config struct {
+	APIVersion    string `yaml:"apiVersion"`
+	Kind          string `yaml:"kind"`
+	ServingInfo   `yaml:"servingInfo"`
+	ClusterInfo   `yaml:"clusterInfo"`
+	Auth          `yaml:"auth"`
+	Customization `yaml:"customization"`
+	Providers     `yaml:"providers"`
+}
+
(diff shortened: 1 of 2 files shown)


pkg/console/config/types.go
```diff
@@ -0,0 +1,56 @@
+package consoleserver
```
jhadvig: Not sure if its the right name for this pkg ...

mfojtik: should be `config`

mfojtik:
@jhadvig is this only these types or also helpers associated with encode/decode ?

jhadvig:
@mfojtik makes sense to have it altogether .. updated the PR

benjaminapetersen:
Maybe just the structure / types, not the funds for parsing cli flags?

jhadvig:
Make sense to leave the function in the console repo. Thanks @benjaminapetersen 
Updated

enj:
I am not sure if CLI types belong here.

enj:
Copy the structs between the repos.  This does not warrant code in lib-go.

/close

PR #457, opened 2019-06-13
split valid and enabled TLS versions

mrogers950:
We need to discern between the package-valid TLS versions (with go1.12 this now includes tls1.3) and the server enabled versions (which only goes up to tls1.2). Currently some origin tests assume that the package-valid versions above the default are automatically supported, but tls1.3 is opt-in at build time.

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -31,11 +31,19 @@ import (
 	"k8s.io/client-go/util/cert"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/crypto/crypto.go
```diff
@@ -72,6 +80,14 @@ func TLSVersionOrDie(versionName string) uint16 {
 	}
 	return version
 }
+func SupportedTLSVersions() []string {
```
deads2k: Clarify the difference between supported and valid for me. I don't understand it.

PR #459, opened 2019-06-14
auth: add impersonating client

mfojtik:
Used by integration and openshift apiserver (apiserver and admission).

This is the last dependency in tests that point to openshift-apiserver.

/cc @deads2k

### DIFF
--- a/pkg/authentication/impersonatingclient/impersonate_rbac.go
+++ b/pkg/authentication/impersonatingclient/impersonate_rbac.go
@@ -1,0 +1,19 @@
+package impersonatingclient
+
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	rbacv1 "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+// NewImpersonatingRBACFromContext returns an impersonated RBAC client with the user of the context.
+func NewImpersonatingRBACFromContext(ctx context.Context, restclient rest.Interface) (rbacv1.RbacV1Interface, error) {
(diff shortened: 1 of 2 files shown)


pkg/authentication/impersonatingclient/impersonate_rbac.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	rbacv1 "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+func NewImpersonatingRBACFromContext(ctx context.Context, restclient rest.Interface) (rbacv1.RbacV1Interface, error) {
```
sttts: godoc

mfojtik: fixed

pkg/authentication/impersonatingclient/impersonate_rbac.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	rbacv1 "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+// NewImpersonatingRBACFromContext returns an RBAC client based on the current request context.
```
sttts: returns an impersonated RBAC client with the user of the context.

mfojtik:
/hold

@deads2k had objections

deads2k:
/close

This is a very advanced power that basically no one should be using.  Only the config creation was used elsewhere, we can copy the five lines.

PR #460, opened 2019-06-14
add usercache package

mfojtik:
Used by kube-apiserver and openshift-apiserver. Seems small and contained.

/cc @deads2k


deads2k:
/hold

The only actual usage of this is now in the kube-apiserver.  Remove the reference in openshfit-apiserver because that's a stale way to wire to an informer we don't use.

PR #461, opened 2019-06-14
add ability to control whether a secret is synced

deads2k:
/hold

holding for proof and for tests

Needed by kcm-o


deads2k:
found a leaner approach

/close

PR #463, opened 2019-06-20
allow more interia for degraded because installer pods take a long time for AWS

deads2k:
We chose 1 minute of inertia back when we only waited 30 seconds for AWS.  We now wait 70 seconds, so we blip to degraded.  This makes us longer than the 70 seconds.

/assign @derekwaynecarr


derekwaynecarr:
makes sense.

/lgtm

PR #466, opened 2019-06-24
certs: add prometheus metrics for CA, signer and target certs

mfojtik:
Attempt number two to introduce certificate expiration prometheus metrics. There are three metrics, all with low cardinality:

* `cert_ca_bundle_expire_in_hours` is a gauge that tracks the number of hours before the certs inside the CA bundle reach their expiration. 
* `cert_signer_expire_in_hours` is a gauge that tracks the number of hours before the signer cert reach its expiration.
* `cert_target_expire_in_hours` is a guage that tracks the number of hours before the target cert expire.

The collector tracks all secrets and config maps provided by the lister. The labels and types are already being set on secrets and configs we produce.

This version add unit test and fixes some logic errors inside the collector. Also fixes the "double" registration bug which caused this to explode before 4.1 release.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/certrotation/metrics.go
+++ b/pkg/operator/certrotation/metrics.go
@@ -1,0 +1,206 @@
+package certrotation
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/labels"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/cert"
+	"k8s.io/klog"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+var (
+	defaultTimeNowFn = time.Now
+)
+
+var (
+	caBundleExpireHoursDesc = prometheus.NewDesc(
+		"cert_ca_bundle_expire_in_hours",
+		"Number of hours until certificates in given CA bundle expire",
+		[]string{"namespace", "name", "common_name", "signer_name", "index", "valid_from"}, nil)
+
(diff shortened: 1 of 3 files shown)


pkg/operator/certrotation/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+		[]string{"namespace", "name", "common_name", "signer_name", "valid_from"}, nil)
+)
+
+type certExpirationMetricsCollector struct {
+	configLister corev1listers.ConfigMapLister
+	secretLister corev1listers.SecretLister
+
+	nowFn func() time.Time
+}
+
```
sttts: put in a type assertion

sttts: against the interface

pkg/operator/certrotation/metrics.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	return selector
+}
+
+// registerOnce guarantee that this prometheus metric collector will only be registered once.
+// TODO: Provide more reliable metric registration system in future.
+var registerOnce sync.Once
+
+// RegisterCertExpirationMetrics registers certificate monitoring metrics.
+func RegisterCertExpirationMetrics(configMaps corev1listers.ConfigMapLister, secrets corev1listers.SecretLister) {
```
sttts: why public?

mfojtik: fixed

pkg/operator/certrotation/metrics.go
mfojtik: i probably don't need this at all

sttts:
Would prefer to see sensible alerts before merging any new metrics. The chance is high that we add something not usable for alerts.

/hold

PR #467, opened 2019-06-26
crypto: add constant time equal and rand gen

stlaz:
Currently used in OAuth server and its operator, these are generally useful.

### DIFF
--- a/pkg/crypto/random.go
+++ b/pkg/crypto/random.go
@@ -1,0 +1,35 @@
+package crypto
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+)
+
+// RandomBitsString returns a random string with at least the requested bits of entropy.
+// It uses RawURLEncoding to ensure we do not get / characters or trailing ='s.
+// Callers should avoid using a value less than 256 unless they have a very good reason.
+func RandomBitsString(bits int) string {
+	size := (bits + 7) / 8
+
+	b := make([]byte, size)
+	if _, err := rand.Read(b); err != nil {
+		panic(err) // rand should never fail
+	}
+	return base64.RawURLEncoding.EncodeToString(b)
+}
+
+// Random256BitsString is a convenience function for calling RandomBitsString(256).
+// Callers that need a random string should use this function unless they have a
+// very good reason to need a different amount of entropy.
+func Random256BitsString() string {
+	// 32 bytes (256 bits) = 43 base64-encoded characters
+	return RandomBitsString(256)
+}
+
+// Random512BitsString is a convenience function for calling RandomBitsString(512).
+// Callers that need a random string should use this function unless they have a
+// very good reason to need a different amount of entropy.
+func Random512BitsString() string {
+	// 32 bytes (256 bits) = 43 base64-encoded characters
+	return RandomBitsString(512)
+}


enj:
/lgtm

Small but not code that I want repeated everywhere.

enj:
Note that this is also used by the console operator to generate its OAuth client secret.

pkg/crypto/compare.go
```diff
@@ -0,0 +1,7 @@
+package crypto
+
+import "crypto/subtle"
+
+func IsEqualConstantTime(s1, s2 string) bool {
```
sttts: docs. What is this?

deads2k: nak, no wrappers on one liners

pkg/crypto/random.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+)
+
+// RandomBits returns a random byte slice with at least the requested bits of entropy.
+// Callers should avoid using a value less than 256 unless they have a very good reason.
+func RandomBits(bits int) []byte {
+	size := bits / 8
```
sttts: `size := (bits + 7) / 8`

pkg/crypto/random.go
```diff
@@ -0,0 +1,34 @@
+package crypto
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+)
+
+// RandomBits returns a random byte slice with at least the requested bits of entropy.
+// Callers should avoid using a value less than 256 unless they have a very good reason.
+func RandomBits(bits int) []byte {
```
deads2k: nak, never used outside this. excessive flexibility

pkg/crypto/random.go
```diff
@@ -0,0 +1,34 @@
...
+	if _, err := rand.Read(b); err != nil {
+		panic(err) // rand should never fail
+	}
+	return b
+}
+
+// RandomBitsString returns a random string with at least the requested bits of entropy.
+// It uses RawURLEncoding to ensure we do not get / characters or trailing ='s.
+func RandomBitsString(bits int) string {
+	return base64.RawURLEncoding.EncodeToString(RandomBits(bits))
```
deads2k: nak, no one liners

pkg/crypto/random.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Random256BitsString is a convenience function for calling RandomBitsString(256).
+// Callers that need a random string should use this function unless they have a
+// very good reason to need a different amount of entropy.
+func Random256BitsString() string {
```
deads2k: leaving this function which becomes four lines long.  

nak, copy.

deads2k:
I think you can get a away with one liners (just embed them, wrappers aren't helpful) and a four line function (at most)  in two spots.

/close

deads2k:
If you can find three callers in different binaries, I can live with Random256 and Random512.

deads2k:
/hold

for evidence

enj:
> If you can find three callers in different binaries, I can live with Random256 and Random512.

1. OAuth server
1. Authn operator
1. Console operator

I am sure I can find others.

deads2k:
> > If you can find three callers in different binaries, I can live with Random256 and Random512.
> 
> 1. OAuth server
> 2. Authn operator
> 3. Console operator
> 
> I am sure I can find others.

if you're moving it here to be useful you should find them all.

> > If you can find three callers in different binaries, I can live with Random256 and Random512.
> 
> 1. OAuth server
> 2. Authn operator
> 3. Console operator
> 
> I am sure I can find others.

Then do so.

deads2k:
@tnozicka verify-deps again

enj:
1. oc
1. extended tests (I do not really count this though)

GitHub search is lacking in this regard.

stlaz:
Removed the constant-time-comparison, kept only the `RandomBitsString*` functions, while keeping their base still exporrted

enj:
/unassign
/uncc

@stlaz @sttts @mfojtik

PR #471, opened 2019-07-10
pkg/asset/create: Add Update options to patch existing resources

abhinavdahiya:
Allows consumers to switch between current behavior of skipping already existing objects in cluster
and updating the objects in cluster to desired specification in the manifest file

Also extends the default create to PUT status subresource when status field is set for a manifest.

### DIFF
--- a/pkg/assets/create/creater.go
+++ b/pkg/assets/create/creater.go
@@ -15,6 +15,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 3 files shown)


pkg/assets/create/creater.go
```diff
@@ -200,6 +204,75 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			}
+			modData, err := manifest.MarshalJSON()
+			if err != nil {
+				return nil, unknown, err
+			}
+			patch, err := strategicpatch.CreateTwoWayMergePatch(origData, modData, manifest)
+			if err != nil {
+				return nil, unknown, err
+			}
+			incluster, err = resource.Patch(manifest.GetName(), types.MergePatchType, patch, metav1.PatchOptions{})
```
deads2k: As I recall, patching objects in this way doesn't apply the removals correctly.  What is the use-case for this?

deads2k:
UpdateStatus is something we need to add, but I would suggest keying that off of the presence of a status block in the input and it would be done once: create, updatestatus.  I don't see a need for trying to issue a general update and I don't think that the patch logic works in general.

Can we scale this back to just adding the UpdateStatus we need for proxy information?

Also, I think the fake bump needs to happen here.: https://github.com/openshift/cluster-bootstrap

/assign @sttts

sttts:
Is this a feature CVO is going to use or is it for cluster-bootstrap? The latter only creates, never tries to replace.

sttts:
/hold

We discussed this at length in Slack. Instead of solving the update problem here is to remove the possible race between CVO and cluster-bootstrap. The CVO should start only after all other manifests have been created. `cluster-config-operator render` is supposed to render all manifests needed for that.

PR #472, opened 2019-07-12
resourceapply: don't log secret data

sttts:
Log OLD and NEW instead if a data value changed.

Reject `stringData` because we don't support that right now (could be added if needed, but not in this PR).

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -196,6 +196,10 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
 
 // ApplySecret merges objectmeta, requires data
 func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
+	if len(required.StringData) > 0 {
+		return nil, false, fmt.Errorf("Secret.stringData is not supported")
+	}
+
 	existing, err := client.Secrets(required.Namespace).Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		actual, err := client.Secrets(required.Namespace).Create(required)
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -196,6 +196,10 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
 
 // ApplySecret merges objectmeta, requires data
 func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
+	if len(required.StringData) > 0 {
+		return nil, false, fmt.Errorf("Secret.stringData is not supported")
```
deads2k: why not?

sttts: Because it's normalized by the server to `Data`, and we would have to add that normalization here before the deepcopy. Not impossible, just not done.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -210,14 +214,29 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
...
 	}
 	existingCopy.Data = required.Data
 
 	if klog.V(4) {
-		klog.Infof("Secret %q changes: %v", required.Namespace+"/"+required.Name, JSONPatch(existing, required))
+		safeRequired := required.DeepCopy()
+		safeExisting := existing.DeepCopy()
+
+		for s := range safeExisting.Data {
+			safeExisting.Data[s] = []byte("OLD")
```
enj: maybe we want `OLD`, `NEW` and `MODIFIED` as the states to track?

sttts: what do you mean with 3 states? We only have two.

sttts: Ic what you mean. But note that we create a patch below. We cannot have OLD and NEW for all fields, only those which changed.

enj:
/retitle [release-4.1] Bug 1734622: resourceapply: don't log secret data

tnozicka:
> /retitle Bug 1734622: resourceapply: don't log secret data

@enj This one was for master branch, not release-4.1. What am I missing?

PR #474, opened 2019-07-15
wires CustomNoUpgrade to ObserveFeatureFlag

p0lyn0mial:
the new field was added in openshift/api#370


p0lyn0mial:
please review https://github.com/openshift/library-go/pull/474/commits/6f1665e0fabd7e5a2518a498a666bc1979ccdbbf, the second commit contains updated deps.

pkg/operator/configobserver/featuregates/observe_featuregates.go
```diff
@@ -95,3 +82,47 @@ func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers
...
+	formatEnabledFunc := func(fs string) string {
+		return fmt.Sprintf("%s=true", fs)
+	}
+	formatDisabledFunc := func(fs string) string {
+		return fmt.Sprintf("%s=false", fs)
+	}
+
+	if fg.Spec.FeatureSet == configv1.CustomNoUpgrade {
+		if fg.Spec.FeatureGateSelection.CustomNoUpgrade == nil {
+			return nil, fmt.Errorf(".spec.featureGateSelection.CustomNoUpgrade is empty but the flag %s was set", configv1.CustomNoUpgrade)
```
deads2k: We don't want to fail.  This function should on fail on fatal issues.  You can add this to the validating admission plugin.

pkg/operator/configobserver/featuregates/observe_featuregates.go
```diff
@@ -95,3 +82,47 @@ func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers
...
+		return newConfigValue, nil
+	}
+
+	featureSet, ok := configv1.FeatureSets[fg.Spec.FeatureSet]
+	if !ok {
+		return nil, fmt.Errorf(".spec.featureSet %q not found", featureSet)
+	}
+	for _, enable := range featureSet.Enabled {
+		// only add whitelisted feature flags
+		if !f.allowAll && !f.knownFeatures.Has(enable) {
```
deads2k: I think we need this same check on the customnoupgrade or we won't be able to combine featuregates in one spot.

deads2k:
there are some post merge updates needed, but this unsticks us from vendoring downstream

/lgtm

PR #476, opened 2019-07-16
makes FeatureGateSelection.CustonNoUpgrade to tolerate knownFeatures

p0lyn0mial:


p0lyn0mial:
addresses review comments from https://github.com/openshift/library-go/pull/474

PR #478, opened 2019-07-17
rendor: add optional --manifest-operator-image

sttts:
Prereq for https://github.com/openshift/cluster-kube-apiserver-operator/pull/525.


abhinavdahiya:
/lgtm

/hold
so that @deads2k can also take a look.

deads2k:
I'm ok with this, but I prefer the installer to get more opinionated about the input it provides and do things like provide all mappings and directory with certs.

/hold cancel

PR #479, opened 2019-07-19
pkg/asset/create: run UpdateSatus for resources that have status field set

abhinavdahiya:
Resources like proxies.config.openshift.io that have a status subresource require extra call to UpdateStatus to ensure that the status
contents of the manifest are applied to the cluster.

/cc @deads2k @sttts

### DIFF
--- a/pkg/assets/create/creater.go
+++ b/pkg/assets/create/creater.go
@@ -163,13 +163,19 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
 			continue
 		}
 
+		var resource dynamic.ResourceInterface
 		if mappings.Scope.Name() == meta.RESTScopeNameRoot {
-			_, err = client.Resource(mappings.Resource).Create(manifests[path], metav1.CreateOptions{})
+			resource = client.Resource(mappings.Resource)
 		} else {
-			_, err = client.Resource(mappings.Resource).Namespace(manifests[path].GetNamespace()).Create(manifests[path], metav1.CreateOptions{})
+			resource = client.Resource(mappings.Resource).Namespace(manifests[path].GetNamespace())
 		}
-
 		resourceString := mappings.Resource.Resource + "." + mappings.Resource.Version + "." + mappings.Resource.Group + "/" + manifests[path].GetName() + " -n " + manifests[path].GetNamespace()
+
+		incluster, err := resource.Create(manifests[path], metav1.CreateOptions{})
+
+		if err == nil && options.Verbose {
+			fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
+		}
 
 		// Resource already exists means we already succeeded
 		// This should never happen as we remove already created items from the manifest list, unless the resource existed beforehand.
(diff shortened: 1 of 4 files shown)


pkg/assets/create/creater.go
```diff
@@ -163,36 +163,58 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
 			}
-			errs[path] = fmt.Errorf("failed to create: %v", err)
+			errs[path] = fmt.Errorf("failed to create %s: %v", resourceString, err)
 			continue
 		}
 
-		if options.Verbose {
-			fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
+		if _, ok := manifests[path].Object["status"]; ok {
+			unstructured.SetNestedMap(incluster.Object, manifests[path].Object, "status")
```
sttts: what should this line do? Now it sets the status field of incluster to the value of the whole manifest object. You probably only want `manifests[path].Object["status"]` instead.

pkg/assets/create/creater.go
```diff
@@ -163,36 +163,58 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
 				fmt.Fprintf(options.StdErr, "Failed to create %q %s: %v\n", path, resourceString, err)
 			}
-			errs[path] = fmt.Errorf("failed to create: %v", err)
+			errs[path] = fmt.Errorf("failed to create %s: %v", resourceString, err)
 			continue
 		}
 
-		if options.Verbose {
-			fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
+		if _, ok := manifests[path].Object["status"]; ok {
```
sttts: I would expect a check that `include.Object` has no status field. Otherwise, we might race with other actors setting status.

sttts:
Am overall ok with calling `UpdateStatus`. Found some issues in the implementation.

pkg/assets/create/creater.go
abhinavdahiya: Hmm, i'm not sure I follow your suggestion? how does that prevent race... ?

pkg/assets/create/creater.go
abhinavdahiya: Ahh, didn't realize from the API doc.

pkg/assets/create/creater.go
abhinavdahiya: an manifests are file on disk, why would they change mid-execution?

sttts: You creation of object has two step: main create call plus status update. If there is time in between some operator waiting for the object might be update it in between.

Maybe a more theoretical problem. But just checking for /status wouldn't harm I think.

abhinavdahiya: > If there is time in between some operator waiting for the object might be update it in between.
> Maybe a more theoretical problem. But just checking for .status wouldn't harm I think.

So if an operator in the mean time updated the status, cluster-bootstrap should skip UpdateStatus is that what you are suggesting?

abhinavdahiya:
ping @sttts

pkg/assets/create/creater.go
sttts: Yes.

Note this code is potentially not only used in cluster-bootstrap. It is better generically correct.

abhinavdahiya: @sttts updated to skip UpdateStatus when it already exists incluster.. with https://github.com/openshift/library-go/pull/479/commits/30c07c12fcd1bc3ee2093c1558985d154b89c995

pkg/assets/create/creater.go
```diff
@@ -163,36 +163,68 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			_, found, err := unstructured.NestedMap(incluster.Object, "status")
+			if err != nil {
+				if options.Verbose {
+					fmt.Fprintf(options.StdErr, "Failed to find status field %q %s: %v\n", path, resourceString, err)
+				}
+				errs[path] = fmt.Errorf("failed to find status field %s: %v", resourceString, err)
+				continue
+			}
+			if !found {
+				unstructured.SetNestedMap(incluster.Object, manifests[path].Object["status"].(map[string]interface{}), "status")
```
sttts: this will panic if somebody adds a yaml manifest like:
```
metadata:
  ..
spec: ...
status:
```
This renders status to `null` in JSON which is no `map[string]interface{}`.

abhinavdahiya: Fixed with https://github.com/openshift/library-go/pull/479/commits/9e68872b12bb441d339456b80cdbd807360e3db7

also added extra manifest to test this.

abhinavdahiya: @sttts

pkg/assets/create/creater.go
```diff
@@ -163,36 +163,68 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+				incluster, err = resource.UpdateStatus(incluster, metav1.UpdateOptions{})
+				if err != nil && !kerrors.IsNotFound(err) {
+					if options.Verbose {
+						fmt.Fprintf(options.StdErr, "Failed to update the  %q %s: %v\n", path, resourceString, err)
+					}
+					errs[path] = fmt.Errorf("failed to update status for %s: %v", resourceString, err)
+					continue
+				}
+				if options.Verbose {
+					fmt.Fprintf(options.StdErr, "Updated status for %q %s\n", path, resourceString)
```
sttts: nit: this message is slightly wrong in case of notfound error

pkg/assets/create/creater.go
```diff
@@ -163,36 +163,68 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			_, found, err := unstructured.NestedMap(incluster.Object, "status")
+			if err != nil {
+				if options.Verbose {
+					fmt.Fprintf(options.StdErr, "Failed to find status field %q %s: %v\n", path, resourceString, err)
+				}
+				errs[path] = fmt.Errorf("failed to find status field %s: %v", resourceString, err)
+				continue
+			}
+			if !found {
+				unstructured.SetNestedMap(incluster.Object, manifests[path].Object["status"].(map[string]interface{}), "status")
```
sttts: why this complexity with typecasts? Just set it as an untyped value here with`SetNestedField`, and drop the `!= nil` test above.

pkg/assets/create/creater.go
```diff
@@ -163,36 +163,61 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			errs[path] = fmt.Errorf("failed to create %s: %v", resourceString, err)
 			continue
 		}
 
-		if options.Verbose {
-			fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
+		if _, ok := manifests[path].Object["status"]; ok {
+			_, found := incluster.Object["status"]
+			if !found {
+				incluster.Object["status"] = manifests[path].Object["status"]
```
sttts: :+1:

PR #480, opened 2019-07-22
Support GOPATH containing a list of paths

tnozicka:
/cc @sttts 

replaces https://github.com/openshift/oc/pull/10

for some reason ART uses multiple paths in GOPATH which looks weird but is allowed

### DIFF
--- a/alpha-build-machinery/make/lib/golang.mk
+++ b/alpha-build-machinery/make/lib/golang.mk
@@ -1,6 +1,9 @@
 GO ?=go
 GOPATH ?=$(shell $(GO) env GOPATH)
-GO_PACKAGE :=$(subst $(GOPATH)/src/,,$(abspath .))
+gopath_list :=$(subst :, ,$(strip $(GOPATH)))
+# Use every path in GOPATH to try to remove it as a prefix of current dir to determine the package name.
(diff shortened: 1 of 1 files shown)


alpha-build-machinery/make/lib/golang.mk
```diff
@@ -1,6 +1,9 @@
 GO ?=go
 GOPATH ?=$(shell $(GO) env GOPATH)
-GO_PACKAGE :=$(subst $(GOPATH)/src/,,$(abspath .))
+gopath_list :=$(subst :, ,$(strip $(GOPATH)))
+# Use every path in GOPATH to try to remove it as a prefix of current dir to determine the package name.
+# If the prefix is not removed on subtitution, filter-out unchanged paths.
+GO_PACKAGE ?=$(strip $(filter-out $(abspath .),$(foreach p,$(gopath_list),$(patsubst $(p)/src/%,%,$(abspath .)))))
```
sttts: do we really have to support all items in the GOPATH list or just the first?

tnozicka: we are not guaranteed to be in the first one and it's seems negligible since we already have to do the hard part of splitting the gopath and substituting it

PR #483, opened 2019-07-24
Cross build

tnozicka:
/cc @sttts @deads2k

### DIFF
--- a/alpha-build-machinery/Makefile
+++ b/alpha-build-machinery/Makefile
@@ -12,7 +12,7 @@ examples :=$(wildcard ./make/examples/*/Makefile.test)
 # Dynamic values are replaced with "<redacted_for_diff>" so we can do diff against checkout versions.
 define update-makefile-log
 mkdir -p "$(3)"
-$(MAKE) -j 1 -C "$(dir $(1))" -f "$(notdir $(1))" --no-print-directory --warn-undefined-variables $(2) 2>&1 | \
+set -o pipefail; $(MAKE) -j 1 -C "$(dir $(1))" -f "$(notdir $(1))" --no-print-directory --warn-undefined-variables $(2) 2>&1 | \
    sed 's/\.\(buildDate\|versionFromGit\|commitFromGit\|gitTreeState\)="[^"]*" /.\1="<redacted_for_diff>" /g' | \
    tee "$(3)"/"$(notdir $(1))"$(subst ..,.,.$(2).log)
 
--- a/alpha-build-machinery/make/examples/multiple-binaries/Makefile
+++ b/alpha-build-machinery/make/examples/multiple-binaries/Makefile
@@ -1,3 +1,22 @@
 include $(addprefix ../../, \
 	golang.mk \
 )
+
+CROSS_BUILD_BINDIR :=_output/bin
+
+cross-build-darwin-amd64:
+	+@GOOS=darwin GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/darwin_amd64
+.PHONY: cross-build-darwin-amd64
+
+cross-build-windows-amd64:
+	+@GOOS=windows GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/windows_amd64
+.PHONY: cross-build-windows-amd64
+
+cross-build: cross-build-darwin-amd64 cross-build-windows-amd64
+.PHONY: cross-build
+
+clean-cross-build:
+	$(RM) -r '$(CROSS_BUILD_BINDIR)'
+.PHONY: clean-cross-build
+
+clean: clean-cross-build
--- a/alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
+++ b/alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
@@ -1,7 +1,10 @@
 SHELL :=/bin/bash -euo pipefail
 
-all:
-	$(MAKE) -C . build
+test: test-build test-cross-build
+.PHONY: test
+
+test-build:
+	$(MAKE) build
 	[[ -f ./openshift ]]
 	[[ -f ./oc ]]
 
@@ -16,7 +19,25 @@ all:
 		git describe --long --tags --abbrev=7 --match 'v[0-9]*' || echo 'v0.0.0-unknown'; \
 	)
 
-	$(MAKE) -C . clean
+	$(MAKE) clean
 	[[ ! -f ./openshift ]]
 	[[ ! -f ./oc ]]
-.PHONY: all
+.PHONY: test-build
+
+test-cross-build:
+	[[ ! -f ./_output/bin ]]
+	$(MAKE) cross-build
+	[[ ! -f ./openshift ]]
+	[[ ! -f ./oc ]]
+	[[ -f ./_output/bin/darwin_amd64/openshift ]]
+	[[ -f ./_output/bin/darwin_amd64/oc ]]
+	[[ -f ./_output/bin/windows_amd64/openshift.exe ]]
+	[[ -f ./_output/bin/windows_amd64/oc.exe ]]
+
+	$(MAKE) clean
+	[[ ! -f ./_output/bin/darwin_amd64/openshift ]]
+	[[ ! -f ./_output/bin/darwin_amd64/oc ]]
+	[[ ! -f ./_output/bin/windows_amd64/openshift.exe ]]
+	[[ ! -f ./_output/bin/windows_amd64/oc.exe ]]
+	[[ ! -f ./_output/bin ]]
+.PHONY: test-cross-build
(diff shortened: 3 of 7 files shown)


eparis:
why aren't we using `go install` which will put the packages in the well known goroot/bin/arch/os locations?

tnozicka:
I see install and build as two distinct actions, especially given some people have GOPATH/bin in PATH. When building a package you don't necessarily want to install it. (Say I install `oc` from master branch and use that system wide and then work on a bug in older branch - when compiling it I don't want to install it, just use it locally.)

That said I was thinking of building into `_bin` dir locally where the architectures could be separated by subfolders and the default binaries would move there as well to avoid possible name conflicts in root. This was just an easier change that we can build on top of and still get cross-build fast enough because currently those binaries are build from different sources.

alpha-build-machinery/Makefile
```diff
@@ -12,7 +12,7 @@ examples :=$(wildcard ./make/examples/*/Makefile.test)
 # Dynamic values are replaced with "<redacted_for_diff>" so we can do diff against checkout versions.
 define update-makefile-log
 mkdir -p "$(3)"
-$(MAKE) -j 1 -C "$(dir $(1))" -f "$(notdir $(1))" --no-print-directory --warn-undefined-variables $(2) 2>&1 | \
+set -o pipefail; $(MAKE) -j 1 -C "$(dir $(1))" -f "$(notdir $(1))" --no-print-directory --warn-undefined-variables $(2) 2>&1 | \
```
sttts: how much do I love the default error behaviour of bash with pipes.

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -8,12 +8,12 @@ all:
 	# test version is set correctly when linking
 	#   majorFromGit, minorFromGit are deprecated upstream and set to empty value
 	#   we avoid comparing time to avoid flakes
-	diff <( ./oc | sed '$$d' ) <( \
+	#   we avoid comparing git tree state
+	diff <( ./oc | grep -v -e 'clean' -e 'dirty' | sed '$$d' ) <( \
```
sttts: oc?

alpha-build-machinery/Makefile
tnozicka: who doesn't :)

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
tnozicka: ambiguous https://github.com/openshift/library-go/tree/963a391ce8fa3eeed242845a9eee88b9ce24b798/alpha-build-machinery/make/examples/multiple-binaries/cmd/oc

sttts: it's just an example, right? So nvmd.

sttts: but why the change?

alpha-build-machinery/make/targets/golang/build.mk
```diff
@@ -1,7 +1,8 @@
 self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
 
+go_bin_suffix ?=
 define build-package
-	$(strip $(GO) build $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) $(1))
+	$(strip $(GO) build $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) $(if $(go_bin_suffix),-o '$(notdir $(1))$(go_bin_suffix)',) $(1))
```
sttts: what does it do? worth a comment?

tnozicka: fills `-o` only if `$(go_bin_suffix)` is not empty

tnozicka:
also install can conflict on names as you loose namespacing - e.g. kubectl from k8s and from openshift

but I am slightly more inclined to move us to `_bin` folder if we agree on the UX, just didn't want to couple it to this change

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
tnozicka: clean vs. dirty has pretty much always been the issue - if you run `make update` before you commit and after, you get different results. also if part of the update is writing down the updated log files this is pretty much a cycle

eparis:
> That said I was thinking of building into `_bin` dir locally where the architectures could be separated by subfolders and the default binaries would move there as well to avoid possible name conflicts in root. This was just an easier change that we can build on top of and still get cross-build fast enough because currently those binaries are build from different sources.

Ok, how about we copy kinda what golang does and put crossbuilt stuff in _output/bin/$GOOS/$GOARCH/file.$SUFFIX   ?
Leave local arch builds in the root dir.

eparis:
Is this really better than the bash scripts kube/origin use? From a supportability perspective?  How many people will be able to enhance this magic other than @tnozicka ?

tnozicka:
updated with a middle ground (native binaries in root folder and cross in `_bin/$(GOOS)_$(GOARCH)/`, less magic and @eparis had a good point on being able to set different tags for cross builds

alpha-build-machinery/Makefile
sttts: that `set` sets it forever, doesn't it? intentional?

tnozicka: every make line has its own shell

sttts: right

alpha-build-machinery/make/examples/multiple-binaries/Makefile
```diff
@@ -1,3 +1,22 @@
 include $(addprefix ../../, \
 	golang.mk \
 )
+
+CROSS_BUILD_BINDIR :=_bin
```
sttts: elsewhere we call this `output`. `_bin` is super ugly.

alpha-build-machinery/make/examples/multiple-binaries/Makefile
```diff
@@ -1,3 +1,22 @@
...
+
+cross-build-darwin-amd64:
+	+@GOOS=darwin GOARCH=amd64 $(MAKE) --no-print-directory build GO_BUILD_BINDIR:=$(CROSS_BUILD_BINDIR)/darwin_amd64
```
sttts: what is the `+`?

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -1,22 +1,38 @@
...
 	[[ ! -f ./oc ]]
+
+	[[ ! -f ./_bin ]]
+	$(MAKE) cross-build
+	[[ ! -f ./openshift ]]
+	[[ ! -f ./oc ]]
+	[[ -f ./_bin/darwin_amd64/openshift ]]
+	[[ -f ./_bin/darwin_amd64/oc ]]
+	[[ -f ./_bin/windows_amd64/openshift.exe ]]
+	[[ -f ./_bin/windows_amd64/oc.exe ]]
```
sttts: makefiles testing makefiles 😱

alpha-build-machinery/make/targets/golang/build.mk
```diff
@@ -1,7 +1,10 @@
 self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
 
 define build-package
-	$(strip $(GO) build $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) $(1))
+	$(if $(GO_BUILD_BINDIR),mkdir -p '$(GO_BUILD_BINDIR)',)
+	$(strip $(GO) build $(GO_BUILD_FLAGS) $(GO_LD_FLAGS) \
+		$(if $(GO_BUILD_BINDIR),-o '$(GO_BUILD_BINDIR)/$(notdir $(1))$(GOEXE)',) \
```
sttts: `$(TOOMANYDOLLARS)`

alpha-build-machinery/make/examples/multiple-binaries/Makefile
tnozicka: the only intent I had was to save a subdir otherwise you'd need `_output/bin/...`. I don't feel strong if you want to go with that. I have not yet encountered something else then crossbuild artifacts that we need to keep in special dir

alpha-build-machinery/make/examples/multiple-binaries/Makefile
tnozicka: always run. (iow. actually run on dry-run, $(MAKE) should imply it but this is explicit)

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
tnozicka: its shell in the end but we can split it into multiple targets to be able to run a specific test

alpha-build-machinery/make/examples/multiple-binaries/Makefile
tnozicka: I am gonna put RPMs there 2 because rpmbuild isn't able to produce an rpm to current dir and always put the arch subfolder there, so _output it is

alpha-build-machinery/make/examples/multiple-binaries/Makefile
sttts: learned something new :+1:

PR #485, opened 2019-07-25
configobserver: add helper function for observing the Proxy object

stlaz:
OpenShift infrastructure operators should be observing the Proxy object, add a helper so that we can reuse the code.

Useful at least for the kube and openshift apiserver operators.

### DIFF
--- a/pkg/operator/configobserver/proxy/observe_proxy.go
+++ b/pkg/operator/configobserver/proxy/observe_proxy.go
@@ -1,0 +1,88 @@
+package proxy
+
+import (
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+type ProxyLister interface {
+	ProxyLister() configlistersv1.ProxyLister
+}
+
+func NewProxyObserveFunc(configPath []string) configobserver.ObserveConfigFunc {
+	return (&observeProxyFlags{
+		configPath: configPath,
+	}).ObserveProxyConfig
+}
+
+type observeProxyFlags struct {
+	configPath []string
+}
+
+// ObserveProxyConfig observes the proxy.config.openshift.io/cluster object and writes
+// its content to an unstructured object in a string map at the path from the constructor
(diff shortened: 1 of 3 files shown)


stlaz:
Added a unit test, removed logging on Proxy object not-foud to match the featuregates observer behavior.

pkg/operator/configobserver/proxy/observe_proxy.go
```diff
@@ -0,0 +1,86 @@
...
+		configPath: configPath,
+	}).ObserveProxyConfig
+}
+
+type observeProxyFlags struct {
+	configPath []string
+}
+
+// ObserveProxyConfig observes the proxy.config.openshift.io/cluster object and writes
+// its content an unstructured object in a string map at the proxyObserver.proxy path
```
sttts: nit: as an unstructured object

sttts: s/proxyObserver.proxy path/at the path given to the constructor/

pkg/operator/configobserver/proxy/observe_proxy.go
```diff
@@ -0,0 +1,86 @@
...
+func (f *observeProxyFlags) ObserveProxyConfig(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	proxyLister := genericListers.(ProxyLister)
+
+	errs := []error{}
```
sttts: the use of errs is not helpful in the flow here, even wrong (see after the `Get("cluster")`. Would switch to a simple err.

stlaz: This is how the errors are being handled throughout the other observers as well (see `observer_cloudprovider.go` or `observe_featuregates.go`) - they only return the errors that occured during handling of the unstructured object.

I agree that swallowing the error at the `Get("cluster")` part is not optimal, I'll use the even recorder similarly to what's done in `ObserveCloudProviderNames()`

stlaz: Hm, actually, that will only suffice for the `NotFound()` way, but the one with actual error should be added, that could be a problem even for the `clouprovider` observing, I'll fix it there as well.

pkg/operator/configobserver/proxy/observe_proxy.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	observedConfig := map[string]interface{}{}
+	proxyConfig, err := proxyLister.ProxyLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveProxyConfig", "proxy.%s/cluster not found", configv1.GroupName)
+		return observedConfig, errs
+	}
+	if err != nil {
+		errs = append(errs, err)
```
sttts: :+1:

PR #487, opened 2019-07-25
Add generic operator client for ease of use

deads2k:
pending proof of capability.


deads2k:
/assign @damemi 

see green proof pulls.

PR #489, opened 2019-07-26
configobserver/cloudprovider: sync the cloud-config for GCP

abhinavdahiya:
GCP cloud provider needs to be configured to be Regional and MultiZone so that it is looking for resources like instances in all Zones for the region it's deployed in.

/cc @csrwng

### DIFF
--- a/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
+++ b/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
@@ -95,7 +95,7 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 	}
 
 	// we set cloudprovider configmap values only for some cloud providers.
-	validCloudProviders := sets.NewString("azure", "vsphere")
+	validCloudProviders := sets.NewString("azure", "gce", "vsphere")
 	if !validCloudProviders.Has(cloudProvider) {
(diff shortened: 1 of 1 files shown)


pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
```diff
@@ -95,7 +95,7 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 	}
 
 	// we set cloudprovider configmap values only for some cloud providers.
-	validCloudProviders := sets.NewString("azure", "vsphere")
+	validCloudProviders := sets.NewString("azure", "gce", "vsphere")
```
p0lyn0mial: it's not `gcp` ?:)

the title suggests that it should be `gcp` and I have also found https://github.com/openshift/api/commit/21798913ff1a7f03020384e9ab3f36b052146500#diff-cee9360e003319719376a6c9189c2e48

abhinavdahiya: here the values are cloud provider in terms of k8s.. https://github.com/openshift/library-go/blob/b799e5bbbcdc88a34f4bf233bb7001cadc9d5866/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go#L150

PR #490, opened 2019-07-29
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in
golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile.

### DIFF
--- a/alpha-build-machinery/make/targets/golang/verify-update.mk
+++ b/alpha-build-machinery/make/targets/golang/verify-update.mk
@@ -2,6 +2,8 @@ self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
 
 go_files_count :=$(words $(GO_FILES))
 
+GOSEC_GOPATH :=$(shell mktemp -d)
+
 verify-gofmt:
 	$(info Running `$(GOFMT) $(GOFMT_FLAGS)` on $(go_files_count) file(s).)
 	@TMP=$$( mktemp ); \
@@ -25,6 +27,27 @@ verify-govet:
 verify-golint:
 	$(GOLINT) $(GO_PACKAGES)
 .PHONY: verify-govet
+
+verify-gosec: update-gosec
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	$(GOSEC_GOPATH)/bin/$(GOSEC) \
+		-severity $(GOSEC_SEVERITY) -confidence $(GOSEC_CONFIDENCE) \
+		-exclude $(GOSEC_EXCLUDE) \
+		-quiet $(GO_PACKAGES)
+else
+	$(GOSEC) \
+		-severity $(GOSEC_SEVERITY) -confidence $(GOSEC_CONFIDENCE) \
+		-exclude $(GOSEC_EXCLUDE) \
+		-quiet $(GO_PACKAGES)
+endif
+.PHONY: verify-gosec
+
+update-gosec:
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	GOPATH=$(GOSEC_GOPATH) GOBIN=$(GOSEC_GOPATH)/bin GO111MODULE=on go get github.com/securego/gosec/cmd/gosec@4b59c948083cd711b6a8aac8f32721b164899f57
+endif
+.PHONY: update-gosec
+
 
 # We need to be careful to expand all the paths before any include is done
 # or self_dir could be modified for the next include by the included file.
--- a/alpha-build-machinery/make/default.example.mk.help.log
+++ b/alpha-build-machinery/make/default.example.mk.help.log
@@ -15,6 +15,7 @@ update-deps
 update-deps-overrides
 update-generated
 update-gofmt
+update-gosec
 verify
 verify-bindata
 verify-codegen
@@ -22,4 +23,5 @@ verify-deps
 verify-generated
 verify-gofmt
 verify-golint
+verify-gosec
 verify-govet
--- a/alpha-build-machinery/make/golang.example.mk.help.log
+++ b/alpha-build-machinery/make/golang.example.mk.help.log
@@ -8,7 +8,9 @@ test
 test-unit
 update
 update-gofmt
+update-gosec
 verify
 verify-gofmt
 verify-golint
+verify-gosec
 verify-govet
--- a/alpha-build-machinery/make/golang.mk
+++ b/alpha-build-machinery/make/golang.mk
@@ -6,6 +6,7 @@ self_dir := $(dir $(lastword $(MAKEFILE_LIST)))
 
 verify: verify-gofmt
 verify: verify-govet
+verify: verify-gosec
 .PHONY: verify
 
 update: update-gofmt
--- a/alpha-build-machinery/make/lib/golang.mk
+++ b/alpha-build-machinery/make/lib/golang.mk
@@ -11,6 +11,10 @@ GOEXE ?=$(shell $(GO) env GOEXE)
 GOFMT ?=gofmt
 GOFMT_FLAGS ?=-s -l
 GOLINT ?=golint
+GOSEC ?=gosec
+GOSEC_SEVERITY ?=high
+GOSEC_CONFIDENCE ?=high
+GOSEC_EXCLUDE ?=G304
 
 GO_FILES ?=$(shell find . -name '*.go' -not -path '*/vendor/*' -not -path '*/_output/*' -print)
 GO_PACKAGES ?=./...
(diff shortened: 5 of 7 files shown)


alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +26,16 @@ verify-golint:
 	$(GOLINT) $(GO_PACKAGES)
 .PHONY: verify-govet
 
+verify-sec:
+ifeq (, $(shell which gosec 2>/dev/null))
```
tnozicka: `$(GOSEC)`

JAORMX: Done! Thanks for the comment.

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +26,16 @@ verify-golint:
 	$(GOLINT) $(GO_PACKAGES)
 .PHONY: verify-govet
 
+verify-sec:
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	go get -u github.com/securego/gosec/cmd/gosec
```
tnozicka: messing with GOPATH is not desired. I could go for a message and where to get it

sttts: we cannot depend on the master branch. This must be a specific version. Otherwise, CI will break randomly because gosec changed.

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +26,16 @@ verify-golint:
...
+endif
+	$(GOSEC) \
+		-severity medium -confidence medium \
```
tnozicka: those flags need to be configurable and I think the default should really be only high to avoid false positives

/hold

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: Done

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: severity high or confidence high?

I often see value in the medium ones, that's why this was the suggestion.

alpha-build-machinery/make/targets/golang/verify-update.mk
sttts: A message does not work. This is the same problem we had with crd-schema-gen. We have to put versions into our makefiles, not anywhere else (I could live with a variable to override it from an operator, but that won't make a difference to this argument).

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: They're configurable now. thanks for the feedback.

I would still prefer them to be medium. instead of high; any chance to discuss this?

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +26,16 @@ verify-golint:
 	$(GOLINT) $(GO_PACKAGES)
 .PHONY: verify-govet
 
+verify-sec:
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	GO111MODULE=on go get github.com/securego/gosec/cmd/gosec@4b59c948083cd711b6a8aac8f32721b164899f57
+endif
+	$(GOSEC) \
+		-severity $(GOSEC_SEVERITY) -confidence $(GOSEC_CONFIDENCE) \
```
sttts: no default value?

JAORMX: the default value is in another file (similar to the default value of GOFMT_FLAGS)

JAORMX: They are set in aonther file, similar to GOFMT_FLAGS

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +26,16 @@ verify-golint:
 	$(GOLINT) $(GO_PACKAGES)
 .PHONY: verify-govet
 
+verify-sec: GOSEC_GOPATH :=$(shell mktemp -d)
```
sttts: @tnozicka  wondering, we also create `_output` now. Shouldn't we put those things also there? Temp files created from Makefiles have a code smell.

JAORMX: So... should I remove the mktemp call?

JAORMX:
Any more feedback on this PR?

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +28,27 @@ verify-golint:
...
 .PHONY: verify-govet
 
+verify-sec: update-gosec
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	$(GOSEC_GOPATH)/bin/$(GOSEC) \
+		-severity $(GOSEC_SEVERITY) -confidence $(GOSEC_CONFIDENCE) \
+		-exclude $(GOSEC_EXCLUDE) \
+		-quiet $(GO_PACKAGES)
+else
+	$(GOSEC) \
```
tnozicka: Wouldn't it be better to always use the same pinned version so people having gosec installed don't get different results then then the rest and CI?

alpha-build-machinery/make/targets/golang/verify-update.mk
```diff
@@ -26,6 +28,27 @@ verify-golint:
...
+		-severity $(GOSEC_SEVERITY) -confidence $(GOSEC_CONFIDENCE) \
+		-exclude $(GOSEC_EXCLUDE) \
+		-quiet $(GO_PACKAGES)
+endif
+.PHONY: verify-sec
+
+update-gosec:
+ifeq (, $(shell which $(GOSEC) 2>/dev/null))
+	GOPATH=$(GOSEC_GOPATH) GOBIN=$(GOSEC_GOPATH)/bin GO111MODULE=on go get github.com/securego/gosec/cmd/gosec@4b59c948083cd711b6a8aac8f32721b164899f57
```
tnozicka: not sure why this should be selecting GO111MODULE for the user?

alpha-build-machinery/make/targets/golang/verify-update.mk
tnozicka: my reasoning is to limit false positives (given it is configurable you can choose lower confidence if you want). given the false positives means polluting code with dummy comments - I'l like to see what the difference is - have you tried it against one of our big repositories (or origin before the split)? if this ever gets hooked into CI the confidence should be high

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: In order for us to get a pinned version. Which addresses your other comment.

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: That's addressed on the "update-gosec" target

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: So we get a pinner version

alpha-build-machinery/make/targets/golang/verify-update.mk
JAORMX: addressed in the "update-gosec" comment.

alpha-build-machinery/make/targets/golang/verify-update.mk
tnozicka: Not sure, this is pretty tricky and we currently do it in several ways. We checkpoint some tool versions in our build images but that doesn't have effect on local dev so we just assume devs get compatible version. (e.g. glide just errors out and informs the user to get a compatible version). Some tools like bindata we vendor and then create tmpdir to install them into but that's for those that need to be coupled to the project. Vendoring it means common deps issues.

Maybe the least controversial in this case is to output an error message e.g. with the `go get` command for user to install the tool, possibly check version as well. If this get hooked into CI we will pre-install in the builder image.

JAORMX:
In an attempt to make this PR less controversial. I moved the severity and confidence of the security checks to high. This way, it'll only catch critical issues.

PR #496, opened 2019-08-01
Make image building to be more configurable and confirm to 4.x naming conventions

tnozicka:
We need this so we can build images with correct names especially for the case where that is a chained build.

E.g. we need to be able to match the image name in
```
FROM registry.svc.ci.openshift.org/ocp/4.2:cli
```

@openshift/openshift-team-master This need Makefile updated on first bump. (These targets are not exercised in CI so it won't fail.) 

/cc @sttts

### DIFF
--- a/alpha-build-machinery/make/default.example.mk
+++ b/alpha-build-machinery/make/default.example.mk
@@ -20,12 +20,12 @@ CODEGEN_GROUPS_VERSION :=openshiftapiserver:v1alpha1
 #   $ make -n --print-data-base | grep ^CODEGEN
 
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
-# $0 - macro name
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-cluster-openshift-apiserver-operator,./Dockerfile,.)
+$(call build-image,ocp-cli,registry.svc.ci.openshift.org/ocp/4.2:cli,./images/cli/Dockerfile.rhel,.)
 
 # This will call a macro called "add-bindata" which will generate bindata specific targets based on the parameters:
 # $0 - macro name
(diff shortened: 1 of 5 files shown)


alpha-build-machinery/make/operator.example.mk
```diff
@@ -22,12 +22,15 @@ CODEGEN_GROUPS_VERSION :=openshiftapiserver:v1alpha1
...
+# $1 - target name
+# $2 - image registry
+# $3 - image org
+# $4 - image name
+# $5 - image tag
+# $6 - Dockerfile path
+# $7 - context
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-cluster-openshift-apiserver-operator,./Dockerfile,.)
+$(call build-image,ocp-openshift-apiserver-operator,registry.svc.ci.openshift.org,ocp,4.2,openshift-apiserver-operator,./Dockerfile.rhel,.)
```
sttts: this is a bad interface with 7 params. Can't we combine them at the calling site? `$(addsuffix /,$(2))$(addsuffix /,$(3))$(4)$(addprefix :,$(5))` for example

alpha-build-machinery/make/targets/openshift/images.mk
```diff
@@ -1,25 +1,32 @@
...
+# $2 - image registry
+# $3 - image org
+# $4 - image name
+# $5 - image tag
+# $6 - Dockerfile path
+# $7 - context
 define build-image-internal
 image-$(1):
-	$(strip imagebuilder --allow-pull $(IMAGE_BUILD_EXTRA_FLAGS) -f $(2) -t $(addsuffix /,$(IMAGE_REGISTRY))$(addsuffix /,$(IMAGE_ORG))$(1)$(addprefix :,$(IMAGE_TAG)) $(3))
+	$(strip \
```
sttts: what does this add to calling imagebuilder directly?

tnozicka: strips white chars one space

alpha-build-machinery/make/operator.example.mk
tnozicka: I guess, it gave us more flexibility in naming targets automatically, but now that we target as arg we can collapse it

alpha-build-machinery/make/default.example.mk
```diff
@@ -20,12 +20,12 @@ CODEGEN_GROUPS_VERSION :=openshiftapiserver:v1alpha1
...
-# $1 - target suffix
-# $2 - Dockerfile path
-# $3 - context directory for image build
+# $1 - target name
+# $2 - image ref
+# $3 - Dockerfile path
+# $4 - context
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,origin-cluster-openshift-apiserver-operator,./Dockerfile,.)
+$(call build-image,ocp-cli,registry.svc.ci.openshift.org/ocp/4.2:cli,./images/cli/Dockerfile.rhel,.)
```
sttts: much better

PR #499, opened 2019-08-02
observe proxy: don't change target in noconfig -> noconfig case

stlaz:
causes trouble in https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/530/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws/2467/artifacts/e2e-aws/pods/openshift-kube-apiserver-operator_kube-apiserver-operator-564769bd4f-p8s98_kube-apiserver-operator.log

### DIFF
--- a/pkg/operator/configobserver/proxy/observe_proxy.go
+++ b/pkg/operator/configobserver/proxy/observe_proxy.go
@@ -56,7 +56,7 @@ func (f *observeProxyFlags) ObserveProxyConfig(genericListers configobserver.Lis
 	}
 
 	newProxyMap := proxyToMap(proxyConfig)
-	if len(newProxyMap) > 0 {
+	if newProxyMap != nil {
 		if err := unstructured.SetNestedStringMap(observedConfig, newProxyMap, f.configPath...); err != nil {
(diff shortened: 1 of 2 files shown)


stlaz:
The ill behavior can also be observed in `openshift-apiserver-operator` logs, interestingly it does not report Progressing/Degraded

pkg/operator/configobserver/proxy/observe_proxy.go
```diff
@@ -60,6 +60,9 @@ func (f *observeProxyFlags) ObserveProxyConfig(genericListers configobserver.Lis
 		if err := unstructured.SetNestedStringMap(observedConfig, newProxyMap, f.configPath...); err != nil {
 			errs = append(errs, err)
 		}
+	} else if len(currentProxyMap) == 0 {
+		// both are unset
+		return existingConfig, errs
```
sttts: why is this necessary? Because empty map vs. nil is not deep equal?

stlaz: exactly

PR #501, opened 2019-08-06
Update observe proxy to use Status NoProxy

jcpowermac:
When the installer creates the Proxy manifest
Status NoProxy contains the list of cluster urls and
ip address(es) that should not be proxied.  Spec noProxy
only contains the user provided list.


stlaz:
/lgtm
It's a good thing we don't have any e2e tests about this, otherwise this would have broken them (and would be blocked by https://github.com/openshift/cluster-network-operator/pull/245)
Only worth to make QEs life easier until the above PR is merged I guess.

PR #502, opened 2019-08-07
Symlink kube

tnozicka:
/cc @mfojtik @soltysh @deads2k


tnozicka:
This is really just a bug fix to do what hacks in origin did.

ref: https://github.com/openshift/kubernetes-apiserver/pull/1#issuecomment-519489336

mfojtik:
I'm not sure we want to do this anymore, lets wait for @deads2k and @sttts to be back and talk about what we want to do. This can wait.

sttts:
Staging in library-go???? 😱 Let's close this before anybody merges.

tnozicka:
> Staging in library-go???? scream Let's close this before anybody merges.

it is in the very specific section of openshift dependency management. If you vendor the whole kubernetes I think the correct path is to symlink staging

@deads2k ?

sttts:
Do we vendor k/k here? We should not.

sttts:
It is about the build system. But we certainly won't test-vendor k/k here in a library repo. Create a test repo and check it out in some CI test instead if we want to test it continuously.

tnozicka:
> Do we vendor k/k here? We should not.

it is for build machinery, not library-go

sttts:
> it is for build machinery, not library-go

Still. 2 million lines of (dead) code is a nogo. Create a test repo and wire that into the CI tests here.

tnozicka:
Unit tests are conventionally located at the same repo. This is just another reason that shows buil-machinery should live in it's own repo.

tnozicka:
I guess I could wire it so the example vendor isn't committed (and does live restore), I agree we don't wan't to pull it in and commit into deps. Doesn't the `--strip-vendor` prune those though?

@sttts I don't think I've head your opinion on the symlinking itself

sttts:
Not committing it sounds like a solution. Would be ok with that.

`--strip-vendor` strips it, but you still get a huge repository.

PR #505, opened 2019-08-12
Bug 1738432: genericoperatorclient: only preserve fields outside of Operator{Spec,Status}

sttts:
Before this, we just used mergo to merge old spec/status with the new value. It had no concept of specified `OperatorSpec/Status` fields and also merged observed configs with mergo. The consequence was that fields were never removed, although the new observed config did not have those fields anymore.

### DIFF
--- a/pkg/operator/genericoperatorclient/dynamic_operator_client.go
+++ b/pkg/operator/genericoperatorclient/dynamic_operator_client.go
@@ -1,9 +1,9 @@
 package genericoperatorclient
 
 import (
+	"reflect"
+	"strings"
 	"time"
-
-	"github.com/imdario/mergo"
 
 	"k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -17,6 +16,8 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
```
damemi: some weird import grouping here

pkg/operator/genericoperatorclient/dynamic_operator_client_test.go
```diff
@@ -29,10 +30,52 @@ func TestSetOperatorSpecFromUnstructured(t *testing.T) {
...
+					"operatorLogLevel":           "",
+					"unsupportedConfigOverrides": nil,
+					"observedConfig":             nil,
+				},
+			},
+		},
+		{
+			name: "replace-rawextensions",
+			in: map[string]interface{}{
+				"spec": map[string]interface{}{
```
enj: Do we need some tests with both spec and status set?  Along with metadata?

pkg/operator/genericoperatorclient/dynamic_operator_client.go
sttts: fixed

pkg/operator/genericoperatorclient/dynamic_operator_client_test.go
sttts: Kind of out-of-scope, but added anyway.

PR #507, opened 2019-08-23
Bug 1719188: fix IsOperatorAlwaysManaged check so operator conditions are set to u…

gabemontero:
…nknown if unmanaged is NOT supported

Now, are operators which want only to support managed, and not unmanaged etc. already calling SetOperatorAlwaysManaged() at 
https://github.com/gabemontero/library-go/blob/master/pkg/operator/management/management_state.go#L25-L27

Or are they not calling that method and getting desired behavior by accident?

Do we have a precise list of operator who need to be "always managed"

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -124,7 +124,10 @@ func (c StatusSyncer) sync() error {
 	}
 	clusterOperatorObj := originalClusterOperatorObj.DeepCopy()
 
-	if detailedSpec.ManagementState == operatorv1.Unmanaged && !management.IsOperatorAlwaysManaged() {
+	// management.IsOperatorAlwaysManaged() means the operator can't be set to unmanaged state.
+	// so if a cluster admin sets an operator as unmanaged when it can't be unmanaged, then we set the
+	// conditions to unknown, as we are now in an indeterminate state
+	if detailedSpec.ManagementState == operatorv1.Unmanaged && management.IsOperatorAlwaysManaged() {
 		clusterOperatorObj.Status = configv1.ClusterOperatorStatus{}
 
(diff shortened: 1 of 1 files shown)


gabemontero:
/assign @deads2k 

@mfojtik @bparees @openshift/openshift-team-developer-experience ptal / fyi

pkg/operator/status/status_controller.go
```diff
@@ -124,7 +124,10 @@ func (c StatusSyncer) sync() error {
 	}
 	clusterOperatorObj := originalClusterOperatorObj.DeepCopy()
 
-	if detailedSpec.ManagementState == operatorv1.Unmanaged && !management.IsOperatorAlwaysManaged() {
+	// management.IsOperatorAlwaysManaged() means the operator can't be set to unmanaged state.
+	// so if a cluster admin sets an operator as unmanaged when it can't be unmanaged, then we set the
+	// conditions to unknown, as we are now in an indeterminate state
+	if detailedSpec.ManagementState == operatorv1.Unmanaged && management.IsOperatorAlwaysManaged() {
```
deads2k: This change looks off and the existing code looks right. If are always managed, then entering the block below doesn't make any sense because that operator has no way to be be disabled.  That operator is never unmanaged because the operator simply doesn't accept that spec value.

bparees: is the operator is supposed to be always managed, why wouldn't we just stomp their change back to "managed"?

Because we still need to provide a backdoor way for the admin to hack stuff up?  what does "IsOperatorAlwaysManaged" actually reflect/check/mean?

gabemontero: The wording of methods / godoc does not resonate in that way with me, but I'm previously uninitiated to this code repo, so I'll defer and call https://github.com/openshift/library-go/blob/master/pkg/operator/management/management_state.go#L25 in the https://github.com/openshift/cluster-openshift-controller-manager-operator which to date allows for the setting of unmanaged.

Thanks for the clarification

PR #508, opened 2019-08-26
watchdog: add --ready-file support

sttts:
To avoid races between watchee and watcher.


sttts:
oauth flake.

/retest

PR #516, opened 2019-08-29
fileobserver: deprecate ExitOnChangeReactor in favor of TerminateOnChangeReactor

mfojtik:
The `ExitOnChangeReactor` is dangerous because it calls `os.Exit(0)` directly which kills the process immediately without proper shutdown. 
The `TerminateOnChangeReactor` allow consumer to pass the termination function.

The `ExitOnChangeReactor` is not removed as somebody might be using it, but marked as deprecated.

/cc @sttts

### DIFF
--- a/pkg/controller/fileobserver/observer.go
+++ b/pkg/controller/fileobserver/observer.go
@@ -10,7 +10,7 @@ import (
 
 type Observer interface {
 	Run(stopChan <-chan struct{})
-	AddReactor(reaction reactorFn, startingFileContent map[string][]byte, files ...string) Observer
+	AddReactor(reaction ReactorFn, startingFileContent map[string][]byte, files ...string) Observer
 }
 
 // ActionType define a type of action observed on the file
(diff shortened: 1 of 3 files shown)


pkg/controller/fileobserver/observer.go
```diff
@@ -40,20 +40,26 @@ func (t ActionType) String(filename string) string {
...
 // ExitOnChangeReactor provides reactor function that causes the process to exit when the change is detected.
-var ExitOnChangeReactor reactorFn = func(filename string, action ActionType) error {
-	klog.Infof("exiting because %q changed", filename)
-	os.Exit(0)
-	return nil
+// DEPRECATED: Using this function cause process to exit immediately without proper shutdown (context close/etc.)
+//             Use the TerminateOnChangeReactor() instead.
+var ExitOnChangeReactor = TerminateOnChangeReactor(func() { os.Exit(0) })
+
+func TerminateOnChangeReactor(terminateFn func()) ReactorFn {
```
sttts: a builder :D I was joking ;-)

mfojtik: i like it better now ;-) even if it was just a joke

PR #521, opened 2019-09-06
Add crd-schema-gen alpha build targets

damemi:
This attempts to add the common crd-schema-gen to alpha-build-targets, using the new upstream crd-gen tag (v0.2.1)


damemi:
@tnozicka, maybe you could debug my make abilities but when I try to run this (see example update makefile in https://github.com/openshift/cluster-kube-scheduler-operator/pull/171) I get `Makefile:45: *** multiple target patterns.  Stop.`... been trying to debug it but haven't been able to figure it out, any ideas?

Could it be the colons in the included targets (for git clone URL and yq binary)?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,24 @@
+CRD_SCHEMA_GEN_TEMP ?=$(shell mktemp -d)
+
+# $1 - schema gen version
+# $2 - manifests directory
+# $3 - package APIs
+define codegen-crds
```
tnozicka: What value does the macro have if it defines fixed targets? You can run it only once so the macro arguments could be fixed variables and you'd not need a macro to begin with.

Is there a usecase where you'd want to have 2 or more separate runs over a single run with multiple args?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,24 @@
...
+	$(CRD_SCHEMA_GEN_TEMP)/bin/controller-gen \
+		schemapatch:manifests=$(2) \
+		output:dir=$(2) \
+		paths="$(subst $() $(),;,$(3))"
+	for p in manifests/*.crd.yaml-merge-patch; do $(CRD_SCHEMA_GEN_TEMP)/bin/yq m -i "$${p%%.crd.yaml-merge-patch}.crd.yaml" "$$p"; done
+
+verify-codegen-crds: update-codegen-crds
+	git diff -q manifests/ || { echo "Changed manifests: "; echo; git diff; false; }
+
+.PHONY: update-codegen-crds verify-codegen-crds crd-schema-gen
```
tnozicka: by convention, in openshift, we use a separate phony statement marking the end of block

tnozicka:
@damemi you are defining a new target inside a marco, so on the call site it would have to do `$(eval`, that's why you get this error.

tnozicka:
Also this PR calls for a test inside alpha-build-machinery so the code get's exercised.

alpha-build-machinery/make/examples/multiple-binaries/manifests/operator.openshift.io_myoperatorresources.yaml
```diff
@@ -0,0 +1,49 @@
...
+  names:
+    kind: MyOperatorResource
+    plural: myoperatorresources
+  scope: ""
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
+  "validation":
```
damemi: @sttts any idea why the quotes around `"validation"` and `"openAPIV3Schema"`?

sttts: Are they also in my PR?

damemi: @sttts They're not, but testing this it looks like it only puts quotes if these sections weren't already there. Possibly a controller-tools bug?

alpha-build-machinery/make/examples/multiple-binaries/manifests/operator.openshift.io_myoperatorresources.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
tnozicka: I could go for simpler one but why not

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,25 @@
+CRD_SCHEMA_GEN_VERSION ?=v0.2.1
+CRD_SCHEMA_GEN_APIS ?=./vendor/github.com/openshift/api/operator/v1
```
tnozicka: Don't store it into vendor it's easy to get whiped on unrelated changes. `pkg/api`?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,25 @@
...
+# $2 - manifests directory
+# $3 - package APIs
+crd-schema-gen:
+	mkdir -p $(CRD_SCHEMA_GEN_TEMP)/bin
+	git clone -b $(CRD_SCHEMA_GEN_VERSION) --single-branch --depth 1 https://github.com/kubernetes-sigs/controller-tools.git $(CRD_SCHEMA_GEN_TEMP)/gen
+	cd $(CRD_SCHEMA_GEN_TEMP)/gen; GO111MODULE=on go build ./cmd/controller-gen; mv controller-gen ../bin
+	curl -f -L -o $(CRD_SCHEMA_GEN_TEMP)/bin/yq https://github.com/mikefarah/yq/releases/download/2.4.0/yq_$(shell uname -s | tr A-Z a-z)_amd64 && chmod +x $(CRD_SCHEMA_GEN_TEMP)/bin/yq
+.PHONY: crd-schema-gen
+
+update-codegen-crds: CRD_SCHEMA_GEN_TEMP :=$(shell mktemp -d)
```
tnozicka: I'd make it a top level var so you can force it to use the same dir and would detect not to install if present

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,25 @@
...
+update-codegen-crds: crd-schema-gen
+	$(CRD_SCHEMA_GEN_TEMP)/bin/controller-gen \
+		schemapatch:manifests=./manifests \
+		output:dir=./manifests \
+		paths="$(subst $() $(),;,$(CRD_SCHEMA_GEN_APIS))"
```
tnozicka: wonder what this does `$() $()`

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,25 @@
...
+		shopt -s nullglob; for p in manifests/*.crd.yaml-merge-patch; do $(CRD_SCHEMA_GEN_TEMP)/bin/yq m -i "$${p%%.crd.yaml-merge-patch}.crd.yaml" "$$p"; done
+.PHONY: update-codegen-crds
+
+verify-codegen-crds: update-codegen-crds
+	git diff -q manifests/ || { echo "Changed manifests: "; echo; git diff; false; }
```
tnozicka: verify shouldn't require git - generate it into tmp folder and check if it matches what's already generated

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
tnozicka: I guess let's install it by default into `_output/tools/` so we cache it

alpha-build-machinery/make/examples/multiple-binaries/manifests/operator.openshift.io_myoperatorresources.yaml
damemi: this was just the kubebuilder init example

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
damemi: I originally had it as a top level var and the tool would fail for me. We also tried top level before this switch and ended up needing to make it a dep to the target like this

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
damemi: @sttts ^?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
damemi: We're not storing anything in vendor, this is reading `openshift/api` from vendor. When it's run from our repos this is where the API is

sttts: +1 

This vendor path is correct.

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
sttts: leads to complexity. Why no git?

tnozicka: a) it depends on what you have staged and what not
b) if you add a new file, it may not be tracked
c) git can be in various states

verify shouldn't change the files it is verifying - this overwrites them on verify (not on update) which is certainly not expected

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
tnozicka: I am not sure there should be such default if it is repo specific, set it in the repo, not in library

sttts: @tnozicka you are right. This path here is wrong. It belong into the consumer repo. No default here.

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
sttts: controller-gen support to output into another directory. That could be used.

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
+CRD_SCHEMA_GEN_VERSION ?=v0.2.1
+CRD_SCHEMA_GEN_APIS ?=$(error CRD_SCHEMA_GEN_APIS is required) 
+CRD_SCHEMA_GEN_TEMP ?=_output/tools/src/sigs.k8s.io/controller-tools
+CRD_SCHEMA_GEN_OUTPUT ?=./manifests
+
+crd-schema-gen:
+	if [ ! -d $(CRD_SCHEMA_GEN_TEMP) ]; then \
+		mkdir -p $(CRD_SCHEMA_GEN_TEMP); \
+		git clone -b $(CRD_SCHEMA_GEN_VERSION) --single-branch --depth 1 https://github.com/kubernetes-sigs/controller-tools.git $(CRD_SCHEMA_GEN_TEMP); \
```
tnozicka: can you reference the repo with the same url?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
...
+		mkdir -p $(CRD_SCHEMA_GEN_TEMP)/bin; \
+		export GOPATH=$(shell pwd)/_output/tools/ && export GOBIN=$(shell pwd)/_output/tools/bin/ && cd $(CRD_SCHEMA_GEN_TEMP) && go mod vendor && go install -mod=vendor ./cmd/controller-gen/...; \
+	fi
+.PHONY: crd-schema-gen
+
+update-codegen-crds: crd-schema-gen
+	$(shell pwd)/_output/tools/bin/controller-gen \
```
tnozicka: `./_output/tools/bin/controller-gen`

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
...
+		schemapatch:manifests=./manifests \
+		output:dir=$(CRD_SCHEMA_GEN_OUTPUT) \
+		paths="$(subst $() $(),;,$(CRD_SCHEMA_GEN_APIS))"
+.PHONY: update-codegen-crds
+
+# crd-schema-patch should be used for repos with yaml patches. It calls update-codegen-crds then applies any patches after first copying any patch files to the output directory (to ensure that verify works since that target runs in a temp dir). It depends on the `yq` tool to merge yaml
+crd-schema-patch: update-codegen-crds
```
tnozicka: rename it to `crd-schema-update-with-patch`?  so `verify-codegen-crds: crd-schema-patch` makes more sense when you read it?

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
...
+	if [ ! -f $(CRD_SCHEMA_GEN_TEMP)/bin/yq ]; then \
+			curl -f -L -o $(CRD_SCHEMA_GEN_TEMP)/bin/yq https://github.com/mikefarah/yq/releases/download/2.4.0/yq_$(shell uname -s | tr A-Z a-z)_amd64 && chmod +x $(CRD_SCHEMA_GEN_TEMP)/bin/yq; \
+	fi
+	if [ $(CRD_SCHEMA_GEN_OUTPUT) != "./manifests" ]; then \
+		cp -n ./manifests/*.crd.yaml-merge-patch $(CRD_SCHEMA_GEN_OUTPUT)/.; \
+	fi
+	shopt -s nullglob; for p in $(CRD_SCHEMA_GEN_OUTPUT)/*.crd.yaml-merge-patch; do $(CRD_SCHEMA_GEN_TEMP)/bin/yq m -i "$${p%%.crd.yaml-merge-patch}.crd.yaml" "$$p"; done
+.PHONY: crd-schema-patch
+
+verify-codegen-crds: CRD_SCHEMA_GEN_OUTPUT :=$(shell mktemp -d)
```
tnozicka: you don't have to put ` ` before `:=` that just usually done in global variable definitions, but doesn't affect anything

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
damemi: You mean `sigs.k8s.io`? Doesn't work for me:
```
$ git clone sigs.k8s.io/controller-tools
fatal: repository 'sigs.k8s.io/controller-tools' does not exist
```

damemi:
@tnozicka updated with your additional feedback and squashed

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -51,3 +51,7 @@ test-rpm:
 	[[ ! -d ./_output/ ]]
 	$(MAKE) clean
 .PHONY: test-rpm
+
+test-update-codegen:
```
tnozicka: you also need to register the test on line 3 as the others are, this isn't run in CI

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -51,3 +51,7 @@ test-rpm:
...
+	$(MAKE) update-codegen
```
tnozicka: This isn't testing the verify target. I guess they are best tested together so lets make target `test-codegen` - I imagine the flow would be like:
1. update
2. verify && success
3. make a change in generated files
4. verify && failure
5. make update
6. verify && success

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
tnozicka: yeah, looks like `git clone` doesn't handle 403

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
...
+	./_output/tools/bin/controller-gen \
+		schemapatch:manifests=./manifests \
+		output:dir=$(CRD_SCHEMA_GEN_OUTPUT) \
+		paths="$(subst $() $(),;,$(CRD_SCHEMA_GEN_APIS))"
+.PHONY: update-codegen-crds
+
+# crd-schema-update-with-patch should be used for repos with yaml patches. It calls update-codegen-crds then applies any patches after first copying any patch files to the output directory (to ensure that verify works since that target runs in a temp dir). It depends on the `yq` tool to merge yaml
+crd-schema-update-with-patch: update-codegen-crds
+	if [ ! -f $(CRD_SCHEMA_GEN_TEMP)/bin/yq ]; then \
+			curl -f -L -o $(CRD_SCHEMA_GEN_TEMP)/bin/yq https://github.com/mikefarah/yq/releases/download/2.4.0/yq_$(shell uname -s | tr A-Z a-z)_amd64 && chmod +x $(CRD_SCHEMA_GEN_TEMP)/bin/yq; \
```
tnozicka: `s~\t\t\t~\t\t~`

alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -0,0 +1,36 @@
...
+	fi
+	if [ $(CRD_SCHEMA_GEN_OUTPUT) != "./manifests" ]; then \
+		cp -n ./manifests/*.crd.yaml-merge-patch $(CRD_SCHEMA_GEN_OUTPUT)/.; \
+	fi
+	shopt -s nullglob; for p in $(CRD_SCHEMA_GEN_OUTPUT)/*.crd.yaml-merge-patch; do $(CRD_SCHEMA_GEN_TEMP)/bin/yq m -i "$${p%%.crd.yaml-merge-patch}.crd.yaml" "$$p"; done
```
tnozicka: shopt &&

tnozicka: or just do `set -euo pipefail &&` at the start of this shell

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
damemi: I think I added this right, please check @tnozicka. Making the change to the generated CRD works best I think by adding a small yaml merge patch and merging it with `yq` (which should already be downloaded from the previous calls to update/verify). Just writing garbage to the file gave some weird results with subsequent calls to yq, probably due to invalid YAML.

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -51,3 +51,12 @@ test-rpm:
 	[[ ! -d ./_output/ ]]
 	$(MAKE) clean
 .PHONY: test-rpm
+
+test-codegen:
+	$(MAKE) update-codegen
+	$(MAKE) verify-codegen
+	./_output/tools/bin/yq m -i manifests/operator.openshift.io_myoperatorresources.crd.yaml manifest.test.crd.yaml-merge-patch
+	-$(MAKE) verify-codegen 2>/dev/null
```
tnozicka: this is not testing that it fails, it only ignores errors
`! $(MAKE) verify-codegen`

alpha-build-machinery/make/examples/multiple-binaries/Makefile.test
```diff
@@ -51,3 +51,12 @@ test-rpm:
 	[[ ! -d ./_output/ ]]
 	$(MAKE) clean
 .PHONY: test-rpm
+
+test-codegen:
+	$(MAKE) update-codegen
+	$(MAKE) verify-codegen
+	./_output/tools/bin/yq m -i manifests/operator.openshift.io_myoperatorresources.crd.yaml manifest.test.crd.yaml-merge-patch
```
tnozicka: I suppose `echo aaa >> the_file && echo bbb >> a_new_file` would be enough to test it is not matching for the diff

tnozicka: you should not get any error on garbage if you change the generated files, if that's the case the generator has loop (it shouldn't depend on it's output)

alpha-build-machinery/make/examples/multiple-binaries/manifest.test.crd.yaml-merge-patch
```diff
@@ -0,0 +1,6 @@
+spec:
+  validation:
+    openAPIV3Schema:
+      properties:
+        kind:
+          pattern: "^modified$"
```
tnozicka: `\n` (fix your editor ;))

tnozicka:
oh, the generator is really recursive https://github.com/kubernetes-sigs/controller-tools/blob/master/cmd/controller-gen/main.go#L139

I need to figure it out, fyi I am fixing the Makefile locally (just that you and we don't duplicate efforts)

tnozicka:
went in with https://github.com/openshift/library-go/pull/526
/close

PR #522, opened 2019-09-10
Handle get and list options in cached getters

enj:

@sttts @p0lyn0mial

### DIFF
--- a/pkg/operator/v1helpers/core_getters.go
+++ b/pkg/operator/v1helpers/core_getters.go
@@ -1,6 +1,8 @@
 package v1helpers
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -35,6 +37,9 @@ func (g combinedConfigMapGetter) ConfigMaps(namespace string) corev1client.Confi
 }
 
 func (g combinedConfigMapInterface) Get(name string, options metav1.GetOptions) (*corev1.ConfigMap, error) {
+	if err := validateResourceVersion(options.ResourceVersion); err != nil {
+		return nil, err
+	}
 	ret, err := g.lister.Get(name)
 	if err != nil {
 		return nil, err
@@ -42,14 +47,18 @@ func (g combinedConfigMapInterface) Get(name string, options metav1.GetOptions) 
 	return ret.DeepCopy(), nil
 }
 func (g combinedConfigMapInterface) List(opts metav1.ListOptions) (*corev1.ConfigMapList, error) {
-	list, err := g.lister.List(labels.Everything())
+	selector, err := listOptionsToSelector(opts)
+	if err != nil {
+		return nil, err
+	}
+	list, err := g.lister.List(selector)
 	if err != nil {
 		return nil, err
 	}
 
-	ret := &corev1.ConfigMapList{}
+	ret := &corev1.ConfigMapList{Items: make([]corev1.ConfigMap, len(list))}
 	for i := range list {
-		ret.Items = append(ret.Items, *(list[i].DeepCopy()))
+		ret.Items[i] = *(list[i].DeepCopy())
 	}
 	return ret, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/operator/v1helpers/core_getters.go
```diff
@@ -100,3 +116,26 @@ func (g combinedSecretInterface) List(opts metav1.ListOptions) (*corev1.SecretLi
 	}
 	return ret, nil
 }
+
+func validateGetOptions(options metav1.GetOptions) error {
+	return validateResourceVersion(options.ResourceVersion)
+}
+
+func listOptionsToSelector(options metav1.ListOptions) (labels.Selector, error) {
+	if len(options.FieldSelector) > 0 || options.Watch || len(options.Continue) > 0 {
```
p0lyn0mial: what about other fields in `ListOptions` struct? (https://github.com/openshift/library-go/blob/96bee2747f6cf4443917046d7341b01dc5d5e1b4/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L334)

if we don't want to repeat the fields we could create an empty list, rewrite the supported fields and compare the diff.

pkg/operator/v1helpers/core_getters.go
```diff
@@ -100,3 +116,26 @@ func (g combinedSecretInterface) List(opts metav1.ListOptions) (*corev1.SecretLi
...
+	}
+	return labels.Parse(options.LabelSelector)
+}
+
+func validateResourceVersion(resourceVersion string) error {
+	switch resourceVersion {
+	case "", "0":
+		return nil
+	default:
+		return fmt.Errorf("options against cache with resource version are unsupported: %s", resourceVersion)
```
p0lyn0mial: nit: would the following be more readable? 

`fmt.Errorf("options against cache with resource version %q are unsupported", resourceVersion)`

pkg/operator/v1helpers/core_getters.go
```diff
@@ -100,3 +116,26 @@ func (g combinedSecretInterface) List(opts metav1.ListOptions) (*corev1.SecretLi
 	}
 	return ret, nil
 }
+
+func validateGetOptions(options metav1.GetOptions) error {
```
p0lyn0mial: do we really need this method? we could just call `validateResourceVersion` directly.

pkg/operator/v1helpers/core_getters.go
```diff
@@ -89,7 +101,11 @@ func (g combinedSecretInterface) Get(name string, options metav1.GetOptions) (*c
 }
 
 func (g combinedSecretInterface) List(opts metav1.ListOptions) (*corev1.SecretList, error) {
-	list, err := g.lister.List(labels.Everything())
+	selector, err := listOptionsToSelector(opts)
+	if err != nil {
+		return nil, err
+	}
+	list, err := g.lister.List(selector)
```
p0lyn0mial: nit: I know you haven't changed it but I've just noticed that a few lines below we could allocate memory as we know the size of the array instead of appending.

```
ret := &corev1.SecretList{}
for i := range list {
  ret.Items = append(ret.Items, *(list[i].DeepCopy()))
}
```

sttts:
Before improving client cacher which were introduced in desperate times, what would it cost to wire through a proper lister? I don't like to continue going the wrong route.

/hold

pkg/operator/v1helpers/core_getters.go
```diff
@@ -89,14 +101,46 @@ func (g combinedSecretInterface) Get(name string, options metav1.GetOptions) (*c
...
+	}
+
+	// make sure no unsupported fields are set (new fields to ListOptions are not silently ignored)
+	optionsCopy := *(options.DeepCopy())
+	optionsCopy.LabelSelector = ""           // setting label selector is supported and checked below
+	optionsCopy.TypeMeta = metav1.TypeMeta{} // ignore type meta
+	optionsCopy.TimeoutSeconds = nil         // ignore timeout
+	optionsCopy.Limit = 0                    // ignore limit
+	optionsCopy.ResourceVersion = ""         // we have already validated this above
+	if optionsEmpty := (metav1.ListOptions{}); optionsCopy != optionsEmpty {
```
p0lyn0mial: yes, I think it's okay to use `equality operator` in this case.

enj:
@mfojtik @sttts @p0lyn0mial you probably care about this sooner rather than later.

/close

PR #525, opened 2019-09-18
fileobserver: do not hash empty string

mfojtik:
Inside auth operator, we read the trusted ca bundle file but we ignore the error when the file does not exists. In that case the starting content for that file is empty string. Without this fix, the fileobserver will hash the empty string (`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`) and then after the binary/operator run, the `calculateFileHash()` return "" as hash if the file does not exists.

That will trigger immediate restart/termination because the content change.

NOTE: This is blocking 4.1 -> 4.2 upgrade test, this was found in the logs:

```
 1 observer_polling.go:78] Observed change: file:/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt (current: "", lastKnown: "e3b0c44298fc1c149af....
```

### DIFF
--- a/pkg/controller/fileobserver/observer_polling.go
+++ b/pkg/controller/fileobserver/observer_polling.go
@@ -38,6 +38,16 @@ func (o *pollingObserver) AddReactor(reaction ReactorFn, startingFileContent map
 
 		if startingContent, ok := startingFileContent[f]; ok {
 			klog.V(3).Infof("Starting from specified content for file %q", f)
+			// if empty starting content is specified, do not hash the empty string but just return it the same
+			// way as calculateFileHash() does in that case.
+			// in case the file exists and is empty, we don't care about the initial content anyway, because we
+			// are only going to react when the file content change.
+			// in case the file does not exists but empty string is specified as initial content, without this
+			// the content will be hashed and reaction will trigger as if the content changed.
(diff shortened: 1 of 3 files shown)


pkg/controller/fileobserver/observer_polling.go
```diff
@@ -38,6 +38,12 @@ func (o *pollingObserver) AddReactor(reaction ReactorFn, startingFileContent map
 
 		if startingContent, ok := startingFileContent[f]; ok {
 			klog.V(3).Infof("Starting from specified content for file %q", f)
+			// if the starting content is empty (like file does not exists, do not calculate hash of empty string
```
stlaz: the comment is wrong, it may still be desirable to react to `file does not exists` -> `file exists but is empty`, in this case however it's probably just that the file is empty

stlaz:
```
mfojtik 2:45 PM
@stlaz if the file does not exists you don't have starting content

mfojtik 2:46 PM
@stlaz calculateHashFile() handle that
it returns ""
```
looking at the thing as a whole it's not that nice, but good enough for now

mfojtik: comment updated.

PR #531, opened 2019-09-25
update dependencies

ironcladlou:
Update dependencies to pick up the latest openshift/api.


ironcladlou:
@deads2k @sttts @enj @Miciah

PR #532, opened 2019-09-26
Fix small bug in crd-schema-gen target that would fail on nil value

damemi:


alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -9,6 +9,7 @@ crd_patches =$(subst $(CRD_SCHEMA_GEN_MANIFESTS),$(CRD_SCHEMA_GEN_OUTPUT),$(wild
 # $1 - crd file
 # $2 - patch file
 define patch-crd
+	cp -n $(CRD_SCHEMA_GEN_MANIFESTS)/$(notdir $2) '$(CRD_SCHEMA_GEN_OUTPUT)/' || true  # FIXME: centos
```
sttts: what is `$(notdir $2)`?

damemi: just the filename of the patch file, in the case that `crd_patches` passes the patch files' locations in a temp dir (ie, verify)

PR #534, opened 2019-09-27
Bug 1754042: Detect more retriable errors from general registry clients

smarterclayton:
If we see a known 429 or 503 error, or an unknown 503 or 500 error,
we can retry.

Will allow broader retry behavior for cli commands.

Prereq for openshift/oc#112

### DIFF
--- a/pkg/image/registryclient/client.go
+++ b/pkg/image/registryclient/client.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/reference"
+	"github.com/docker/distribution/registry/api/errcode"
 	registryclient "github.com/docker/distribution/registry/client"
 	"github.com/docker/distribution/registry/client/auth"
 	"github.com/docker/distribution/registry/client/auth/challenge"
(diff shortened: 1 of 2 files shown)


pkg/image/registryclient/client_test.go
```diff
@@ -320,13 +321,14 @@ func TestRetryFailure(t *testing.T) {
 		t.Fatalf("unexpected: %v %v %#v", m, err, r)
 	}
 
-	// retry four times
+	// verify docker known errors
```
wking: "verify docker known errors" -> "verify network error handling"?  This is generic HTTP stuff, it's not even specific to registries.

smarterclayton: this is errors that docker client knows about

pkg/image/registryclient/client.go
```diff
@@ -369,8 +370,22 @@ func isTemporaryHTTPError(err error) (time.Duration, bool) {
...
+			return 5 * time.Second, true
+		case t.ErrorCode() == errcode.ErrorCodeTooManyRequests:
+			return 2 * time.Second, true
+		}
 	case *registryclient.UnexpectedHTTPResponseError:
-		if t.StatusCode == http.StatusTooManyRequests {
+		switch {
+		case t.StatusCode == http.StatusInternalServerError:
+			return 5 * time.Second, true
+		case t.StatusCode == http.StatusServiceUnavailable:
```
dmage: Don't we want also retry on StatusBadGateway (502) and StatusGatewayTimeout (504)? These usually indicate temporary server-side problems (when the server is overloaded).

dmage:
One question, otherwise looks good to merge.

pkg/image/registryclient/client.go
smarterclayton: Yeah, I think so.  Let me double check the behavior of the daemon in those cases.

PR #536, opened 2019-10-03
installer: skip old revisions and report stucked installer pods via events

mfojtik:
In case there is newer revision available, we should skip creating the installer pod for revision that is  already "old". Instead, we can wait for revision controller to update node target revision to match the latest level...

This might fix the case when we creating revision 5 pod while revision 7 is already available.

Additionally this add commit that will report Warning event when the installer pod is stucked in Pending for longer than 10 minutes. The event will carry reason and message from the pod.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -50,6 +50,10 @@ const (
 	statusConfigMapName = "revision-status"
 )
 
+// maxInstallerPodPendingTime indicates the maximum time we tolerate installer pods in Pending state before we start
+// reporting the reason and message as warnings.
+// Installer pods in pending state usually signals kubelet/networking/image problems.
+var maxInstallerPodPendingTime = 10 * time.Minute
 // InstallerController is a controller that watches the currentRevision and targetRevision fields for each node and spawn
 // installer pods to update the static pods on the master nodes.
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -300,6 +300,12 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 
 		// if we are in a transition, check to see whether our installer pod completed
 		if currNodeState.TargetRevision > currNodeState.CurrentRevision {
+			// Skip creating the installer pod when there is already a revision higher than target revision
+			// In this case, wait for the revision controller to update the target revision on node.
+			if operatorStatus.LatestAvailableRevision > currNodeState.TargetRevision {
```
sttts: expecting a test to reflect this? I thought this has good test coverage.

mfojtik: this function has zero test coverage... I added tests in last commit.

mfojtik:
/hold

This is pending proof-bump in one of the static pod operators but it is ready for review.

mfojtik:
per @deads2k suggestion, I will rewrite this to separate controller that will observe the status of installer pods and report staled/pending conditions up to the operator condition levels.

PR #537, opened 2019-10-03
Generic crdgen

tnozicka:
`openshift/api` needs to generate CRDs in multiple different directories so we need to macro so it can be instantiated multiple times

### DIFF
--- a/alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
+++ b/alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
@@ -1,27 +1,55 @@
 self_dir :=$(dir $(lastword $(MAKEFILE_LIST)))
 
-CRD_SCHEMA_GEN_APIS ?=$(error CRD_SCHEMA_GEN_APIS is required)
-CRD_SCHEMA_GEN_MANIFESTS ?=./manifests
-CRD_SCHEMA_GEN_OUTPUT ?=./manifests
-
-crd_patches =$(subst $(CRD_SCHEMA_GEN_MANIFESTS),$(CRD_SCHEMA_GEN_OUTPUT),$(wildcard $(CRD_SCHEMA_GEN_MANIFESTS)/*.crd.yaml-merge-patch))
-
 # $1 - crd file
 # $2 - patch file
 define patch-crd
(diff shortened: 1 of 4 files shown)


alpha-build-machinery/make/targets/openshift/crd-schema-gen.mk
```diff
@@ -1,53 +1,74 @@
...
-CRD_SCHEMA_GEN_APIS ?=$(error CRD_SCHEMA_GEN_APIS is required)
-CRD_SCHEMA_GEN_MANIFESTS ?=./manifests
-CRD_SCHEMA_GEN_OUTPUT ?=./manifests
-
-crd_patches =$(subst $(CRD_SCHEMA_GEN_MANIFESTS),$(CRD_SCHEMA_GEN_OUTPUT),$(wildcard $(CRD_SCHEMA_GEN_MANIFESTS)/*.crd.yaml-merge-patch))
-
 # $1 - crd file
 # $2 - patch file
 define patch-crd
-	cp -n $(CRD_SCHEMA_GEN_MANIFESTS)/$(notdir $2) '$(CRD_SCHEMA_GEN_OUTPUT)/' || true  # FIXME: centos
```
damemi: why remove this?

tnozicka: What is the reason to duplicate the patch to the output dir? Seems like it is input only. 

I think the intent of avoiding a diff exists only in the test when manifests and tmpdir is diffed as folders so I moved it to that place. The verify here only compares the crd.yaml files directly so it doesn't care about patch files only in one folder.

damemi: Ah, I didn't notice the change [here](https://github.com/openshift/library-go/pull/537/files#diff-64b9efa1ab03ac55fe2aa1425d910d08R25), so I thought we still ran the patch command *in* the output dir (that's why I was thinking we needed to copy the patch files there). I see we don't need that anymore, lgtm

damemi:
Seems build-machinery verify is failing for some reason

tnozicka:
yes, working on a fix

PR #540, opened 2019-10-04
operator/certrotation: make rotation methods public

squeed:
They're useful outside of the static-pod operators.

### DIFF
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -30,7 +30,7 @@ type SigningRotation struct {
 	EventRecorder events.Recorder
 }
 
-func (c SigningRotation) ensureSigningCertKeyPair() (*crypto.CA, error) {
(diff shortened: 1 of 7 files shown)


squeed:
hey @deads2k could you take a quick look at this?

squeed:
This is probably not needed. Hold please, will close if so.

mhenriks:
+1

pkg/operator/certrotation/signer.go
```diff
@@ -30,7 +30,7 @@ type SigningRotation struct {
 	EventRecorder events.Recorder
 }
 
-func (c SigningRotation) ensureSigningCertKeyPair() (*crypto.CA, error) {
+func (c SigningRotation) EnsureSigningCertKeyPair() (*crypto.CA, error) {
```
sttts: what does it do. Public methods need docs.

PR #546, opened 2019-10-07
Emit event when certificate gets updated

soltysh:
/assign @deads2k

### DIFF
--- a/pkg/operator/staticpod/certsyncpod/certsync_controller.go
+++ b/pkg/operator/staticpod/certsyncpod/certsync_controller.go
@@ -79,6 +79,7 @@ func getSecretDir(targetDir, secretName string) string {
 func (c *CertSyncController) sync() error {
 	errors := []error{}
 
+	klog.Infof("Syncing configmaps: %v", c.configMaps)
 	for _, cm := range c.configMaps {
 		configMap, err := c.configMapLister.ConfigMaps(c.namespace).Get(cm.Name)
 		switch {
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -166,8 +167,10 @@ func (c *CertSyncController) sync() error {
 				continue
 			}
 		}
+		c.eventRecorder.Eventf("CertificateUpdated", "Writing updated configmap: %s/%s", configMap.Namespace, configMap.Name)
```
deads2k: handle the removal case above too.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -79,6 +79,7 @@ func getSecretDir(targetDir, secretName string) string {
 func (c *CertSyncController) sync() error {
 	errors := []error{}
 
+	klog.Infof("Syncing configmaps: %v", c.configMaps)
```
deads2k: do we have a good printer on these? I doubt it.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -256,6 +259,7 @@ func (c *CertSyncController) sync() error {
 				continue
 			}
 		}
+		c.eventRecorder.Eventf("CertificateUpdated", "Writing updated secret: %s/%s", secret.Namespace, secret.Name)
```
deads2k: removal case

pkg/operator/staticpod/certsyncpod/certsync_controller.go
deads2k: emit events on errors from this I think.  We don't expect to see any.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
soltysh: You'll get name + optional pairs from the default writer, which imo is sufficient for now.

soltysh:
@deads2k updated, ptal

deads2k:
sure, let's see what happens.

/lgtm

@soltysh you may want to pick this back.

PR #548, opened 2019-10-08
Add MapToEnvVars helper function

soltysh:
/assign @mfojtik 
since you asked for it in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/285#discussion_r328058781 and on top of that we use it in 3 operators: kcm-o, kas-o and oas-o.

/assign @deads2k

### DIFF
--- a/pkg/operator/v1helpers/helpers.go
+++ b/pkg/operator/v1helpers/helpers.go
@@ -1,9 +1,11 @@
 package v1helpers
 
 import (
+	"sort"
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


pkg/operator/v1helpers/helpers.go
```diff
@@ -259,3 +261,18 @@ func (agg aggregate) Error() string {
 func (agg aggregate) Errors() []error {
 	return []error(agg)
 }
+
+func MapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
+	if proxyConfig == nil {
+		return nil
```
mfojtik: i would return empty list maybe so we can append?

pkg/operator/v1helpers/helpers.go
```diff
@@ -259,3 +261,18 @@ func (agg aggregate) Error() string {
 func (agg aggregate) Errors() []error {
 	return []error(agg)
 }
+
+func MapToEnvVars(proxyConfig map[string]string) []corev1.EnvVar {
```
mfojtik: nit: would make this env var name something generic

mfojtik: godoc

soltysh: Yeah, I forgot about that :wink:

soltysh:
@mfojtik updated, ptal

pkg/operator/v1helpers/helpers.go
```diff
@@ -259,3 +261,19 @@ func (agg aggregate) Error() string {
 func (agg aggregate) Errors() []error {
 	return []error(agg)
 }
+
+// MapToEnvVars converts a string-string map to a slice of corev1.EnvVar-s
+func MapToEnvVars(mapEnvVars map[string]string) []corev1.EnvVar {
+	if mapEnvVars == nil {
+		return []corev1.EnvVar{}
```
sttts: why not nil?

soltysh: @mfojtik convinced me it'll be easier to use this in `append`

pkg/operator/v1helpers/helpers.go
sttts: append works with nil too.

pkg/operator/v1helpers/helpers.go
sttts: it is not. This has smell. Don't do magic.

pkg/operator/v1helpers/helpers.go
```diff
@@ -259,3 +261,19 @@ func (agg aggregate) Error() string {
...
+	}
+
+	envVars := []corev1.EnvVar{}
```
sttts: make with capacity

PR #549, opened 2019-10-08
Bug 1761280: don't log secret data

soltysh:
/assign @tnozicka @sttts

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -234,23 +234,7 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 	existingCopy.Data = required.Data
 
 	if klog.V(4) {
-		safeRequired := required.DeepCopy()
-		safeExisting := existing.DeepCopy()
-
-		for s := range safeExisting.Data {
-			safeExisting.Data[s] = []byte("OLD")
-		}
-		for s := range safeRequired.Data {
-			if _, preexisting := existing.Data[s]; !preexisting {
-				safeRequired.Data[s] = []byte("NEW")
-			} else if !equality.Semantic.DeepEqual(existing.Data[s], safeRequired.Data[s]) {
-				safeRequired.Data[s] = []byte("MODIFIED")
-			} else {
-				safeRequired.Data[s] = []byte("OLD")
-			}
-		}
-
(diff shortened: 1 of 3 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -234,23 +234,7 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
...
-				safeRequired.Data[s] = []byte("NEW")
-			} else if !equality.Semantic.DeepEqual(existing.Data[s], safeRequired.Data[s]) {
-				safeRequired.Data[s] = []byte("MODIFIED")
-			} else {
-				safeRequired.Data[s] = []byte("OLD")
-			}
-		}
-
-		klog.Infof("Secret %q changes: %v", required.Namespace+"/"+required.Name, JSONPatch(safeExisting, safeRequired))
+		klog.Infof("Secret %q changes: %v", required.Namespace+"/"+required.Name, JSONPatchSecret(existing, required))
```
tnozicka: nit: don't concatenate string while doing printf at the same time (`'%s/%s'`)

tnozicka: (inherited)

pkg/operator/staticpod/controller/revision/revision_controller.go
```diff
@@ -193,7 +193,7 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 		}
 		if !equality.Semantic.DeepEqual(existingData, requiredData) {
 			if klog.V(4) {
-				klog.Infof("secret %q changes for revision %d: %s", s.Name, revision, resourceapply.JSONPatch(existing, required))
+				klog.Infof("Secret %q changes for revision %d: %s", s.Name, revision, resourceapply.JSONPatchSecret(existing, required))
```
tnozicka: preexisting but it si missing namespace

pkg/operator/resource/resourceapply/core.go
soltysh: Good point, lemme quickly fix it.

pkg/operator/staticpod/controller/revision/revision_controller.go
soltysh: `RevisionResource` does not hold Namespace, thus we don't print it.

PR #550, opened 2019-10-08
installer: add controller that watch pending installer pods

mfojtik:
This will add controller that watches the installer pods in `Pending` state that are in this state for longer then 5 minutes. If such pods are found, the controller will then make the operator go to Degraded and report the reason and message found for such pod/container state.

This will help improve debugging and triaging bugs caused by slow networking or kubelet that prevents rolling updates to static pod based operators.

/cc @deads2k 
/cc @sttts

### DIFF
--- a/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
+++ b/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
@@ -1,0 +1,254 @@
+package installerstate
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const installerStateControllerWorkQueueKey = "key"
+
+// maxToleratedPodPendingDuration is the maximum time we tolerate installer pod in pending state
+var maxToleratedPodPendingDuration = 5 * time.Minute
+
+type InstallerStateController struct {
+	podsGetter      corev1client.PodsGetter
+	eventsGetter    corev1client.EventsGetter
+	queue           workqueue.RateLimitingInterface
+	cachesToSync    []cache.InformerSynced
+	targetNamespace string
+	operatorClient  v1helpers.StaticPodOperatorClient
+	eventRecorder   events.Recorder
+
+	timeNowFn func() time.Time
+}
+
+func NewInstallerStateController(kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	podsGetter corev1client.PodsGetter,
+	eventsGetter corev1client.EventsGetter,
+	operatorClient v1helpers.StaticPodOperatorClient,
+	targetNamespace string,
+	recorder events.Recorder,
+) *InstallerStateController {
+	c := &InstallerStateController{
+		podsGetter:      podsGetter,
+		eventsGetter:    eventsGetter,
+		targetNamespace: targetNamespace,
+		operatorClient:  operatorClient,
+		eventRecorder:   recorder.WithComponentSuffix("installer-state-controller"),
+		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "InstallerStateController"),
+		timeNowFn:       time.Now,
+	}
+
+	c.cachesToSync = append(c.cachesToSync, kubeInformersForTargetNamespace.Core().V1().Pods().Informer().HasSynced)
+	kubeInformersForTargetNamespace.Core().V1().Pods().Informer().AddEventHandler(c.eventHandler())
+
+	return c
+}
+
+func (c *InstallerStateController) eventHandler() cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+	}
+}
+
+// degradedConditionNames lists all supported condition types.
+var degradedConditionNames = []string{
+	"InstallerPodPendingDegraded",
+	"InstallerPodContainerWaitingDegraded",
+	"InstallerPodNetworkingDegraded",
+}
+
+func (c *InstallerStateController) sync() error {
+	pods, err := c.podsGetter.Pods(c.targetNamespace).List(metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"app": "installer"}).String(),
+	})
+	if err != nil {
+		return err
+	}
+
+	// collect all startingObjects that are in pending state for longer than maxToleratedPodPendingDuration
+	pendingPods := []*v1.Pod{}
+	for _, pod := range pods.Items {
+		if pod.Status.Phase != v1.PodPending || pod.Status.StartTime == nil {
+			continue
+		}
+		if c.timeNowFn().Sub(pod.Status.StartTime.Time) >= maxToleratedPodPendingDuration {
+			pendingPods = append(pendingPods, pod.DeepCopy())
+		}
+	}
+
+	// in theory, there should never be two installer startingObjects pending as we don't roll new installer pod
+	// until the previous/existing pod has finished its job.
+	foundConditions := []operatorv1.OperatorCondition{}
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+// degradedConditionNames lists all supported condition types.
+var degradedConditionNames = []string{
+	"InstallerPodPendingDegraded",
+	"InstallerPodContainerWaitingDegraded",
+}
+
+func (c *InstallerStateController) sync() error {
+	pods, err := c.podsGetter.Pods(c.targetNamespace).List(metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labels.Set{"app": "installer"}).String(),
```
deads2k: huh, I don't remember doing this.  neat.

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// collect all pods that are in pending state for longer than maxToleratedPodPendingDuration
+	pendingPods := []*v1.Pod{}
+	hasPendingPods := false
+	for _, pod := range pods.Items {
+		if pod.Status.Phase != v1.PodPending {
+			continue
+		}
+		// we need to track this so we can requeue faster when there are pending pods
+		hasPendingPods = true
```
deads2k: we're going to want that node name for a message

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+		c.queue.AddAfter(installerStateControllerWorkQueueKey, 5*time.Second)
+	}
+
+	return nil
+}
+
+func (c *InstallerStateController) handlePendingInstallerPods(pods []*v1.Pod) []operatorv1.OperatorCondition {
+	conditions := []operatorv1.OperatorCondition{}
+	for _, pod := range pods {
+		// at this poind we already know the pod is pending for longer than expected
```
deads2k: point

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+				conditions = append(conditions, condition)
+				c.eventRecorder.Warningf(condition.Reason, condition.Message)
+			}
+		}
+	}
+
+	return conditions
+}
+
+func (c *InstallerStateController) checkPendingInstallerPod(pod *v1.Pod) error {
```
deads2k: this looks dead

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, updateConditionFuncs...); err != nil {
+		return err
+	}
+
+	// if we found pending pods, requeue faster and do not wait for informer as the pod can be stuck in Pending
+	// with no update for long time and we want to track that time.
+	if hasPendingPods {
+		c.queue.AddAfter(installerStateControllerWorkQueueKey, 5*time.Second)
```
deads2k: too fast, you'll be thrashing on status changes just updating a time and spewing out events.

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
mfojtik: @deads2k isn't node name included in the pod name?

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
mfojtik: @deads2k 1 minute?

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+	klog.Infof("Starting InstallerStateController")
+	defer klog.Infof("Shutting down InstallerStateController")
+	if !cache.WaitForCacheSync(stopCh, c.cachesToSync...) {
+		return
+	}
+
+	// doesn't matter what workers say, only start one.
+	go wait.Until(c.runWorker, time.Second, stopCh)
+
```
deads2k: Add a time-based trigger here.  Auto-queue every minute or so.

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,218 @@
...
+		AddFunc:    func(obj interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+		UpdateFunc: func(old, new interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+		DeleteFunc: func(obj interface{}) { c.queue.Add(installerStateControllerWorkQueueKey) },
+	}
+}
+
+// degradedConditionNames lists all supported condition types.
+var degradedConditionNames = []string{
+	"InstallerPodPendingDegraded",
+	"InstallerPodContainerWaitingDegraded",
```
deads2k: //TODO add condition to single out the SDN for reporting

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
mfojtik: nvmd. added

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if _, _, err := v1helpers.UpdateStaticPodStatus(c.operatorClient, updateConditionFuncs...); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *InstallerStateController) handlePendingInstallerPodsNetworkEvents(pods []*v1.Pod) ([]operatorv1.OperatorCondition, error) {
+	namespaceEvents, err := c.eventsGetter.Events(c.targetNamespace).List(metav1.ListOptions{})
```
deads2k: only grab these if you have pending pods. This is expensive.

mfojtik: done

mfojtik: good catch

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,249 @@
...
+				continue
+			}
+			condition := operatorv1.OperatorCondition{
+				Type:    "InstallerPodNetworkingDegraded",
+				Status:  operatorv1.ConditionTrue,
+				Reason:  event.Reason,
+				Message: fmt.Sprintf("Pod %q on node %q observed degraded networking: %s", pod.Name, pod.Spec.NodeName, event.Message),
+			}
+			conditions = append(conditions, condition)
+			c.eventRecorder.Warningf(condition.Reason, condition.Message)
```
deads2k: this can come out multiple times because kubelets send multiple events.  You want to sum the events and *then* send these messages.

mfojtik: @deads2k i thought the event recorder is smart enough on backend side to only increase the event counter. if this is broken, I don't mind creating more noise in the cluster?

mfojtik: nvmd. I got what you meant... yeah this will just duplicate existing events and fire them using operator as component... i eliminated that and will only fire one warning event.

mfojtik:
The proof PR is green: https://github.com/openshift/cluster-kube-apiserver-operator/pull/586

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return nil
+}
+
+func (c *InstallerStateController) handlePendingInstallerPodsNetworkEvents(pods []*v1.Pod) ([]operatorv1.OperatorCondition, error) {
+	conditions := []operatorv1.OperatorCondition{}
+	if len(pods) == 0 {
+		return conditions, nil
+	}
+	namespaceEvents, err := c.eventsGetter.Events(c.targetNamespace).List(metav1.ListOptions{})
```
soltysh: Wouldn't it be better to use `Search(scheme, obj)` here? This way you'd get only events for the pods. You would be exchanging one call vs several but each will be about specific object?
I'm not sure if obj can be a list of objects, though.

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, event := range namespaceEvents.Items {
+		if event.InvolvedObject.Kind != "Pod" {
+			continue
+		}
+		if !strings.Contains(event.Message, "failed to create pod network") {
```
stlaz: I wonder whether we could benefit from the more general approach of https://github.com/openshift/cluster-kube-apiserver-operator/pull/571/files#diff-d4d4aa822fed3489ac3aee1560b501b4R94 where the idea was that you can later easily extend the number of regexes to get a specific reason about why the pod is failing and not only stick to network failures in the longer run.

mfojtik: yes, a good follow up I think

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
mfojtik: i'm not sure the .Search will make it faster/more efficient... this is only listing events for single namespace and only when there is actual pending pods :-)

deads2k:
this is ok to start, but I would try to write the code so that it works when kubelet events fail to de-dupe and when we have more than one installer pod

/lgtm

PR #552, opened 2019-10-10
Explicitly disable go.mod for bindata generation

soltysh:
/assign @tnozicka


tnozicka:
/lgtm cancel
I think you need the `GOMODULE=on` too for 1.12

tnozicka:
like here https://github.com/openshift/library-go/blob/master/alpha-build-machinery/make/targets/openshift/controller-gen.mk#L14

PR #553, opened 2019-10-10
add tls1.3 cipher suites

stlaz:
This adds TLS 1.3 cipher names to our ciphers list so that we are able to set them in our components. This aligns with the work that's being done for the TLSSecurityProfiles (https://github.com/openshift/api/pull/458) where each of the profiles already contains these cipher suites.

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -31,7 +31,8 @@ import (
 	"k8s.io/client-go/util/cert"
 )
 
-// TLS versions that are known to golang. Go 1.12 adds TLS 1.3 support with a build flag.
+// TLS versions that are known to golang. Go 1.13 adds support for
+// TLS 1.3 that's opt-out with a build flag.
 var versions = map[string]uint16{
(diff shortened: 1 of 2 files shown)


deads2k:
/lgtm

/hold

holding for proof PRs in openshift/orign and openshift/openshfit-apiserver

stlaz:
/hold

Aside from our validation here, kube has its own for the cipher suites, which does not allow these ciphers to be set. It's a good thing we checked.

stlaz:
I've modified the logic a bit so that it matches more the behavior of golang, we'll still need to see whether setting TLS version to 1.3 does not break {kube,openshift}-apiserver I think

stlaz:
@deads2k @sanchezl please have a look, I don't think this should be breaking either apiservers now, kube seem to allow TLS1.3 as `minVersion` too - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L80

stlaz:
/hold cancel
This PR no longer breaks origin/OAS

PR #554, opened 2019-10-10
List only real go files

tnozicka:
there were several issues of matching files in vendor, .git, _output - this PR should make it less likely

/cc @sttts

### DIFF
--- a/alpha-build-machinery/make/lib/golang.mk
+++ b/alpha-build-machinery/make/lib/golang.mk
@@ -12,7 +12,7 @@ GOFMT ?=gofmt
 GOFMT_FLAGS ?=-s -l
 GOLINT ?=golint
 
-GO_FILES ?=$(shell find . -name '*.go' -not -path '*/vendor/*' -not -path '*/_output/*' -print)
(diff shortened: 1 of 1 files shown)


alpha-build-machinery/make/lib/golang.mk
```diff
@@ -12,7 +12,7 @@ GOFMT ?=gofmt
 GOFMT_FLAGS ?=-s -l
 GOLINT ?=golint
 
-GO_FILES ?=$(shell find . -name '*.go' -not -path '*/vendor/*' -not -path '*/_output/*' -print)
```
sttts: why not `find . -name '*.go' -not -path '*/vendor/*' -not -path '*/_output/*' -not -path '*/.git/*' -print`?

soltysh: Honestly the version provided by Tomas doesn't work for me if I invoked:
```
go list -f '{{range .GoFiles}}{{ printf "%s/%s " $.Dir . }}{{end}}{{range .CgoFiles}}{{ printf "%s/%s " $.Dir . }}{{end}}{{range .TestGoFiles}}{{ printf "%s/%s " $.Dir . }}{{end}}' ./cmd/... 
```
I'm only getting:
```
/home/maszulik/workspace/origin/src/github.com/openshift/oc/cmd/oc/oc.go /home/maszulik/workspace/origin/src/github.com/openshift/oc/cmd/oc/oc_test.go
```
Stefan's version works, although both are ugly to read :sob:

soltysh: `go list` is very powerful, but you need to tweak it a lot to get what you need, let's keep it reasonably simple :wink:

mfojtik:
/close

This moved to separate repo.

PR #556, opened 2019-10-11
add IMAGE_REGISTRY flag to default.example.mk

sallyom:
/assign @tnozicka 

add the option of building with a custom image registry/repo:tag, like so:
`make IMAGE_TAG=quay.io/sallyom/openshift-apiserver:test images`

### DIFF
--- a/alpha-build-machinery/make/default.example.mk
+++ b/alpha-build-machinery/make/default.example.mk
@@ -19,13 +19,17 @@ CODEGEN_GROUPS_VERSION :=openshiftapiserver:v1alpha1
 # You can list all codegen related variables by:
 #   $ make -n --print-data-base | grep ^CODEGEN
 
+IMAGE_REGISTRY?=registry.svc.ci.openshift.org
+IMAGE_REPO?=$(IMAGE_REGISTRY)/ocp/4.3
+IMAGE_TAG?=$(IMAGE_REPO):openshift-apiserver
+
 # This will call a macro called "build-image" which will generate image specific targets based on the parameters:
 # $1 - target name
 # $2 - image ref
 # $3 - Dockerfile path
 # $4 - context
 # It will generate target "image-$(1)" for builing the image an binding it as a prerequisite to target "images".
-$(call build-image,ocp-cli,registry.svc.ci.openshift.org/ocp/4.2:cli,./images/cli/Dockerfile.rhel,.)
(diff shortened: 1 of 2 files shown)


sallyom:
@tnozicka I was going to close this, but decided it is useful.  When building images to test with, you'd want to customize `registry/repo:tag` (I use buildah but it would be nice to be able to use the make target).  Also, as an example I switched to `ocp-openshift-apiserver` rather than `ocp-cli` since there are several oc/images rather than the single image (so you'd want `IMAGE_CLI_TAG`, `IMAGE_CLI_ARTIFACTS_TAG`, `IMAGE_DEPLOYER_TAG`, and `IMAGE_RECYCLER_TAG`) for that repo.

alpha-build-machinery/make/default.example.mk
```diff
@@ -19,13 +19,17 @@ CODEGEN_GROUPS_VERSION :=openshiftapiserver:v1alpha1
 # You can list all codegen related variables by:
 #   $ make -n --print-data-base | grep ^CODEGEN
 
+IMAGE_REGISTRY?=registry.svc.ci.openshift.org
+IMAGE_REPO?=$(IMAGE_REGISTRY)/ocp/4.3
```
tnozicka: IMAGE_REPO doesn't contain the image registry part, same for IMAGE_TAG

I thing the correct nomenclature would be `IMAGE_REF :=$(IMAGE_REGISTRY)/$(IMAGE_REPO)/$(IMAGE_NAME):$(IMAGE_TAG)`

tnozicka: or `IMAGE_REF:=$(IMAGE_REGISTRY)/$(IMAGE_NAMESPACE)/$(IMAGE_NAME):$(IMAGE_TAG)` where repository is `$(IMAGE_REGISTRY)/$(IMAGE_NAMESPACE)/$(IMAGE_NAME)` ? I am seeing conflicting sources. @sallyom wdyt?

tnozicka: e.g. dockers claim the repository as the part before image name in 
```
To tag a local image with ID “0e5574283393” into the “fedora” repository with “version1.0”:

$ docker tag 0e5574283393 fedora/httpd:version1.0
```
https://docs.docker.com/engine/reference/commandline/tag/#tag-an-image-referenced-by-name-and-tag

sallyom:
closing, no longer needed

PR #560, opened 2019-10-18
Add SubjectKeyId and AuthorityKeyId to server certs

marun:
The `Subject Common Name` field was previously assumed to be unique and used in determining whether the CA that generated a serving cert had changed. With service serving cert CA rotation, the `Common Name` will now be retained across CAs and this suggested the use of `SubjectKeyId` and `AuthorityKeyId` as a way to determine if a serving cert required regeneration.

Required by https://github.com/openshift/service-ca-operator/pull/73

cc: @sttts @deads2k


marun:
Rebased

PR #561, opened 2019-10-18
Use pre-built version of controller-gen

soltysh:
/assign @sttts 
fyi @tnozicka


tnozicka:
this took time only the first time and was cached for next runs, right? I don't see a good reason yet to having to deal with matrix of architectures and privileges for uploading new versions of those binaries to github

soltysh:
> this took time only the first time and was cached for next runs, right? 

Nope, the moment I wanted to built a newer version I got bit by:
* higher memory usage
* missing deps

I don't think we want to fight with these problems. Having a downloadable binary for linux and mac, and eventually for windows is simple enough.

PR #562, opened 2019-10-18
Add encryption controllers

sttts:
The encryption controllers will also be used in cluster-openshift-apiserver-operator. From the corresponding enhancement doc https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/encrypting-data-at-datastore-layer.md, the resources to be encrypted (beyond secrets and configmaps):

- routes.route.openshift.io (routes can contain embedded TLS private keys)
- oauthaccesstokens.oauth.openshift.io
- oauthauthorizetokens.oauth.openshift.io


mfojtik:
Assuming this is just plain move.

/lgtm

PR #565, opened 2019-10-21
encryption: expect pods for all nodes in staticpod operator status

sttts:
Without this, if pods disappear the encryption operator make progress although they should wait for all nodes.

### DIFF
--- a/pkg/operator/encryption/deployer/staticpod.go
+++ b/pkg/operator/encryption/deployer/staticpod.go
@@ -12,6 +12,8 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
+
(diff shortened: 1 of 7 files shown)


pkg/operator/encryption/deployer/staticpod.go
```diff
@@ -51,14 +56,23 @@ func NewStaticPodDeployer(
...
 
 // DeployedEncryptionConfigSecret returns the deployed encryption config and whether all
 // instances of the operand have acknowledged it.
 func (d *StaticPodDeployer) DeployedEncryptionConfigSecret() (secret *corev1.Secret, converged bool, err error) {
-	revision, err := getAPIServerRevisionOfAllInstances(d.podClient)
+	_, status, _, err := d.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return nil, false, err
+	}
+	if status == nil || len(status.NodeStatuses) == 0 {
```
p0lyn0mial: what does it mean? no nodes?

sttts: Maybe status was wiped or cluster is installing.

PR #567, opened 2019-10-22
operator: generalize revision controller to be independent of static pods

sttts:


sttts:
/hold

until proven it fits into openshift-apiserver.

PR #571, opened 2019-10-24
encryption: fix var scope bug in categorizePods and add tests

sttts:
The pods references in the loop were all the same. Golang, sigh.


pkg/operator/encryption/deployer/revisionedpod.go
```diff
@@ -213,21 +213,26 @@ func nodeNames(pods []*corev1.Pod) sets.String {
 }
 
 func categorizePods(pods []corev1.Pod) (good []*corev1.Pod, bad []*corev1.Pod, progressing bool, err error) {
+	if len(pods) == 0 {
+		return nil, nil, true, err
+	}
```
mfojtik:
```
return good, bad, len(pods) > 0, nil
```

sttts: that‘s equivalent. But return early is more Go like, isn‘t it?

PR #572, opened 2019-10-24
encryption: fix logf format in e2e

sttts:


p0lyn0mial:
could you fix those as well

https://github.com/openshift/library-go/blob/744cc6828424bde41bd57fcb6221283af3ca258d/test/e2e-encryption/encryption_test.go#L174

https://github.com/openshift/library-go/blob/744cc6828424bde41bd57fcb6221283af3ca258d/test/e2e-encryption/encryption_test.go#L201

https://github.com/openshift/library-go/blob/744cc6828424bde41bd57fcb6221283af3ca258d/test/e2e-encryption/encryption_test.go#L217

and more just grep `Logf`

sttts:
> could you fix those as well

those are correct.

PR #575, opened 2019-10-31
WIP: pkg/operator/v1helpers/deployment: New package for reporting deployment status

wking:
Many operators manage a deployment as one of their operands (e.g. the console operator manages a console deployment).  Make it easier for those operators to report the status of the managed deployment by providing a centralized helper that mostly leans on [deployment conditions][1] [set][2] by [the Kubernetes controller][3].

Test examples are picked from [the console namespace][5] in [this CI job][4].

Things I don't like about the current status output:

* Why is `Progressing=True` for the `downloads` deployment?  Seems like `ReplicaSet ... has successfully progressed.` would be a cause to say `Progressing=False, we're all good`.

* Why is there no `ReplicaFailure` condition on the console deployment?  Seems like `Available=False` and `Progressing=False` would be grounds for some kind of degraded condition on the deployment.

* The messages are pretty terse, I'd expect `1 available replica is less than the minimum 2` or similar instead of a blanket `does not have minimum availability`.  Bonus points for including things like the `CrashLoopBackOff` and OAuth `connection refused` from [the unavailabled pods][6].

But however we sort out those things, I think some sort of centralized helper here will make it easier to get consistent, high-quality reporting out of all the operators that manage deployments.  Thoughts?

[1]: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status
[2]: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progress-deadline-seconds
[3]: https://github.com/kubernetes/kubernetes/tree/b7bc5dc9db16731b6e64053bc5b9e8cdca872e14/pkg/controller/deployment
[4]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.3/507
[5]: https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.3/507/artifacts/e2e-gcp/must-gather/registry-svc-ci-openshift-org-ci-op-xcjcp0gb-stable-sha256-dae1257b516a5c177237cfef5a6a3e241962b0d20cf54bcb2b66dc1671c5035e/namespaces/openshift-console/apps/deployments.yaml
[6]: https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.3/507/artifacts/e2e-gcp/must-gather/registry-svc-ci-openshift-org-ci-op-xcjcp0gb-stable-sha256-dae1257b516a5c177237cfef5a6a3e241962b0d20cf54bcb2b66dc1671c5035e/namespaces/openshift-console/pods/console-7974bb6477-xqrmt/console-7974bb6477-xqrmt.yaml

### DIFF
--- a/pkg/operator/v1helpers/deployment/deployment.go
+++ b/pkg/operator/v1helpers/deployment/deployment.go
@@ -1,0 +1,65 @@
+// Package deployment contains helpers for operators which manage deployments.
+package deployment
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// SetOperatorConditions sets prefixed operator conditions based on the deployment's conditions.
+func SetOperatorConditions(conditions *[]operatorv1.OperatorCondition, prefix string, deployment *appsv1.Deployment) error {
+	available := operatorv1.ConditionUnknown
+	degraded := operatorv1.ConditionUnknown
+	for _, condition := range deployment.Status.Conditions {
+		operatorCondition := operatorv1.OperatorCondition{
+			Status:  v1helpers.ConvertCoreV1StatusToOperatorV1Status(condition.Status),
+			Reason:  condition.Reason,
+			Message: condition.Message,
+		}
+
+		switch condition.Type {
+		case appsv1.DeploymentAvailable:
+			operatorCondition.Type = prefix + "Available"
+			available = operatorCondition.Status
+		case appsv1.DeploymentProgressing:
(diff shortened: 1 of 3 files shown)


wking:
> Things I don't like about the current status output...

We need to decide whether I'm just wrong, and these deployment-controller statuses are appropriate, or if we want to push changes to the upstream deployment controller, or if we want to carry local logic that says "I don't trust the deployment controller, when I see `readyReplicas == updatedReplicas == replicas`, set `Progressing=False` regardless of what the deployment controller has for the deployment's `Progressing`" and similar.  I think pushing fixes upstream scales better, but it might be a long, hard push, and we'd have to be careful about not breaking existing consumers (and we might not know the complete set of existing consumers).

PR #577, opened 2019-11-01
encryption: add events to controllers

mfojtik:
Copy of https://github.com/openshift/cluster-kube-apiserver-operator/pull/610 that was lost during heavy lifting.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -93,7 +93,7 @@ func NewKeyController(
 		apiServerClient: apiServerClient,
 
 		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "EncryptionKeyController"),
-		eventRecorder: eventRecorder.WithComponentSuffix("encryption-key-controller"), // TODO unused
+		eventRecorder: eventRecorder.WithComponentSuffix("encryption-key-controller"),
 
 		encryptedGRs: encryptedGRs,
 		component:    component,
@@ -194,7 +194,14 @@ func (c *keyController) checkAndCreateKeys() error {
 	if errors.IsAlreadyExists(createErr) {
 		return c.validateExistingSecret(keySecret, newKeyID)
 	}
-	return createErr
+	if createErr != nil {
+		c.eventRecorder.Warningf("EncryptionKeyCreateFailed", "Secret %q failed to create: %v", keySecret.Name, err)
+		return createErr
+	}
+
+	c.eventRecorder.Eventf("EncryptionKeyCreated", "Secret %q successfully created: %q", keySecret.Name, reasons)
+
+	return nil
 }
 
 func (c *keyController) validateExistingSecret(keySecret *corev1.Secret, keyID uint64) error {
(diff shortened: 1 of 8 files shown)


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -194,7 +194,14 @@ func (c *keyController) checkAndCreateKeys() error {
 	if errors.IsAlreadyExists(createErr) {
 		return c.validateExistingSecret(keySecret, newKeyID)
 	}
-	return createErr
+	if createErr != nil {
+		c.eventRecorder.Warningf("EncryptionKeyCreateFailed", "Secret %q failed to create: %v", keySecret.Name, err)
```
sttts: should we really create these events? They show up as degraded anyway.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -223,12 +223,16 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
 		}
 
 		// idem-potent migration start
+		c.eventRecorder.Eventf("EncryptionMigrationStarted", fmt.Sprintf("Started migration for %s", gr.String()))
```
sttts: this changed. This does not start migration. Ensure is idempotent. Would move the event into the migrator.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -223,12 +223,16 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
...
+		migrationStartedTime := time.Now()
 		finished, result, err := c.migrator.EnsureMigration(gr, ks.Key.Name)
 		if err != nil {
+			c.eventRecorder.Eventf("EncryptionMigrationFailed", fmt.Sprintf("Failed migration of %s (took %s): %v", gr.String(), time.Since(migrationStartedTime), err))
```
sttts: would skip this. It leads to other events.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -223,12 +223,16 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
...
 			errs = append(errs, err)
 			continue
 		}
 		if finished && result != nil {
+			c.eventRecorder.Eventf("EncryptionMigrationFailed", fmt.Sprintf("Finished migration of %s (took %s) with error: %v", gr.String(), time.Since(migrationStartedTime), result))
```
sttts: would move this into the migrator because you are running into this again and again when result is != nil once.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -238,6 +242,8 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
 			continue
 		}
 
+		c.eventRecorder.Eventf("EncryptionMigrationFinished", fmt.Sprintf("Finished migration for %s in %s", gr.String(), time.Since(migrationStartedTime)))
```
sttts: also this one will fire again and again.

pkg/operator/encryption/controllers/key_controller.go
mfojtik: @sttts why not, at least we will see when this started to fail and which secret

sttts: if we go degraded, we see the error message too, don't we?

PR #579, opened 2019-11-04
Allow rotation "only when certs are expired"

tnozicka:
For automatic recovery we need to run a process that regenerates certificate only when they are expired. (library-go forces rotation in 1/5 of the lifetime.) We need to avoid stealing the rotation from the operator so logs are in one place.

proof this isn't breaking current flow in https://github.com/openshift/cluster-kube-apiserver-operator/pull/635

/cc @soltysh

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -28,6 +28,9 @@ type TargetRotation struct {
 	Name      string
 	Validity  time.Duration
 	Refresh   time.Duration
+	// RefreshOnlyWhenExpired allows rotating only certs that are already expired. (for autorecovery)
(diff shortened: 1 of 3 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -24,10 +24,11 @@ import (
...
 type TargetRotation struct {
-	Namespace string
-	Name      string
-	Validity  time.Duration
-	Refresh   time.Duration
+	Namespace              string
+	Name                   string
+	Validity               time.Duration
+	Refresh                time.Duration
+	RefreshOnlyWhenExpired bool
```
sttts: add a go doc please. `// normally rotation happens after 80% of validity. This allows to wait until expiration`.

PR #584, opened 2019-11-07
fix workqueue and client metrics registration

mfojtik:


mfojtik:
/hold

this might not be the right fix, we should think about custom registry instead

PR #585, opened 2019-11-08
encryption: forgiving error handling for migration for retryable errors

sttts:
Following kube-storage-version-migrator logic (errors.go comes from there).


sttts:
/hold

until we smoke tested.

PR #586, opened 2019-11-11
File observer fixes

mfojtik:
* First commit is adding mechanism for callers to ensure the initial sync passed and the fileobserver see and observing changes on files successfully
* Second commit is critical as when the initial content of file is empty, the reactor for the file is never added
* Third commit is adding much more unit test coverage to ensure all corner cases are handled

There is one TODO in unit test when the initial content is empty and the file content is empty (but exists), the observer will trigger "Created" action, even when the file existed. This should be handled as follow up issue.

### DIFF
--- a/pkg/controller/fileobserver/observer_polling.go
+++ b/pkg/controller/fileobserver/observer_polling.go
@@ -17,7 +17,7 @@ import (
 type pollingObserver struct {
 	interval time.Duration
 	reactors map[string][]ReactorFn
-	files    map[string]string
+	files    map[string]fileHashAndState
 
 	reactorsMutex sync.RWMutex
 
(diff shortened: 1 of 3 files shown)


pkg/controller/fileobserver/observer_polling.go
```diff
@@ -20,6 +20,12 @@ type pollingObserver struct {
 	files    map[string]string
 
 	reactorsMutex sync.RWMutex
+	hasSynced     bool
+}
+
+// HasSynced indicates that the observer synced all observed files at least once.
+func (o *pollingObserver) HasSynced() bool {
+	return o.hasSynced
```
sttts: must this be go routine safe?

mfojtik: i thought so as well, but aren't booleans safe in go?

mfojtik: and I was wrong ;-) the data race in unit test :-)

sttts: No, they are not.

mfojtik: fixed by mutex

mfojtik:
/hold

todo must be fixed

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -74,28 +86,42 @@ func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
...
+			klog.Infof("Observed change: file:%s (current: %q, lastKnown: %q)", filename, currentFileState.hash, lastKnownFileState.hash)
 
 			for i := range reactors {
-				action := FileModified
-				switch {
-				case len(lastKnownHash) == 0:
+				action := NoAction
+
+				// do nothing if file does not exists
+				if !lastKnownFileState.exists && !currentFileState.exists {
```
sttts: all these cases would make a nice switch statement (again).

sttts:
Lgtm. Just some nit.

pkg/controller/fileobserver/observer_polling.go
mfojtik: @sttts i found the switch more confusing, but I can follow up? :-) at least I put comments on what each case mean

PR #587, opened 2019-11-12
image/reference: A ref with tag and ID should prefer ID

smarterclayton:
When we convert a ref like `foo/bar:baz@sha256:...` to Exact(), we
should prefer the digest.

### DIFF
--- a/pkg/image/reference/reference.go
+++ b/pkg/image/reference/reference.go
@@ -174,8 +174,6 @@ func (r DockerImageReference) NameString() string {
 	switch {
 	case len(r.Name) == 0:
 		return ""
-	case len(r.Tag) > 0:
-		return r.Name + ":" + r.Tag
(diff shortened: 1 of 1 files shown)


smarterclayton:
This is related to https://github.com/openshift/oc/pull/157 and must be vendored into openshift-apiserver and oc in order for image stream import to be correct.

smarterclayton:
@soltysh this came up due to the SME thread

smarterclayton:
@dmage @bparees @soltysh I need a review on this so I can propagate it

mfojtik:
@smarterclayton @bparees this broke the unit tests in openshift-apiserver, next time I would want to see a proof bump in openshift-apiserver with green unit tests before merging :-)

Oleg is working on fixing the importer.

smarterclayton:
What’s the break?  Tags don’t have priority over digests.

PR #590, opened 2019-11-12
cert-syncer: Fix events for not-found resources

Miciah:
When emitting an event concerning the absence of a resource, use the namespace and name that were used to try to get the resource, rather than the namespace and name from the result of trying to get the resource (which will both be blank since the resource did not exist).

Remove a superfluous `err` argument in the `Warningf` call made when removing the file for a non-existent secret.  The `err` argument has no corresponding placeholder in the format string; the extra argument, combined with using the empty result's namespace and name, resulted in messages like the following:

    Removed file for secret: /%!(EXTRA *errors.StatusError=secrets "user-serving-cert" not found)

As the error is always a "not found" error, it can be omitted.

Follow-up to #546.

* `pkg/operator/staticpod/certsyncpod/certsync_controller.go` (`sync`): Omit `err` from "Removed file for secret" event.  Use the correct namespace and name for events concerning not-found resources.

### DIFF
--- a/pkg/operator/staticpod/certsyncpod/certsync_controller.go
+++ b/pkg/operator/staticpod/certsyncpod/certsync_controller.go
@@ -102,14 +102,14 @@ func (c *CertSyncController) sync() error {
 
 			// remove missing content
 			if err := os.RemoveAll(getConfigMapDir(c.destinationDir, cm.Name)); err != nil {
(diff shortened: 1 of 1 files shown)


Miciah:
Rebased.

deads2k:
awkward
/approve

PR #591, opened 2019-11-14
verifies input params of JSONPatchSecret function to prevent npe

p0lyn0mial:


p0lyn0mial:
fixes https://github.com/openshift/library-go/pull/591

p0lyn0mial:
It seems that `JSONPatchSecret` is a convenience function used to log diff between the original and modified objects. It uses `JSONPatch` which has similar behavior (store errors in response) and usage. I think that fixing the function instead of the callers seems to be okay. I changed the function names to explicitly warn about handling the errors.

p0lyn0mial:
`JSONPatchSecret` usage https://github.com/search?q=org%3Aopenshift+JSONPatchSecret&type=Code

`JSONPatch` usage https://github.com/search?l=Go&q=org%3Aopenshift+JSONPatch&type=Code

PR #594, opened 2019-11-19
Added api to check if bootstrap user is enabled

vareti:


pkg/authentication/bootstrapauthenticator/bootstrap.go
```diff
@@ -15,7 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
-	"k8s.io/client-go/kubernetes/typed/core/v1"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
```
stlaz: in case we need this rename, it should be `corev1`

pkg/authentication/bootstrapauthenticator/bootstrap.go
```diff
@@ -151,3 +151,8 @@ func (b *bootstrapUserDataGetter) Get() (*BootstrapUserData, bool, error) {
 		UID:          base64.RawURLEncoding.EncodeToString(uidBytes[:]),
 	}, true, nil
 }
+
+func (b *bootstrapUserDataGetter) IsEnabled() (bool, error) {
+	_, state, err := b.Get()
```
stlaz: I think this will need more love, maybe moving the logic that decides whether the bootstrap IdP has been removed outside of `Get()` to a separate unexported function that would be shared between these two methods.

Also the comment above mentions considering the oauth-server running inside kube-apiserver, although you were problably right to disregard that part of it, since I don't see how that would be an option these days.

pkg/authentication/bootstrapauthenticator/bootstrap.go
```diff
@@ -92,11 +92,11 @@ type BootstrapUserData struct {
 type BootstrapUserDataGetter interface {
 	Get() (data *BootstrapUserData, ok bool, err error)
 	// TODO add a method like:
-	// IsPermanentlyDisabled() bool
 	// and use it to gate the wiring of components related to the bootstrap user.
 	// when the oauth server is running embedded in the kube api server, this method would always
 	// return false because the control plane would not be functional at the time of the check.
 	// when running as an external process, we can assume a functional control plane to perform the check.
```
stlaz: The comment does not make sense withouth just this one line, you'll eventually want to get rid of it all

stlaz:
LGTM, squash the commits, also I think this would be worth a unit-test

pkg/authentication/bootstrapauthenticator/bootstrap_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestIsEnabled(t *testing.T) {
+	namespaceSystem := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: metav1.NamespaceSystem,
+		},
+	}
+	testCases := []struct {
+		name        string
+		kubeClient  *fake.Clientset
+		expectedErr string
```
stlaz: Before addressing any other comments - you don't use this field at all, do you need it, then? Or you should remove the `namespaceSystem` in one of the TCs and use it, although such a scenario does not sound very likely.

Also, have go default this field for you (it'll make it be an empty string)

pkg/authentication/bootstrapauthenticator/bootstrap_test.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			b := NewBootstrapUserDataGetter(test.kubeClient.CoreV1(), test.kubeClient.CoreV1())
+			enabled, err := b.IsEnabled()
+			switch {
+			case len(test.expectedErr) == 0 && err == nil:
+			case len(test.expectedErr) == 0 && err != nil,
+				len(test.expectedErr) > 0 && err == nil,
+				len(test.expectedErr) > 0 && err != nil && !strings.Contains(err.Error(), test.expectedErr):
+				t.Errorf("%s: expected %#v, got %#v", test.name, test.expectedErr, err)
```
stlaz: isn't `test.name` automatically included in errors by doing `t.Run`?

pkg/authentication/bootstrapauthenticator/bootstrap_test.go
vareti: i added this field to test out any error cases but was unable to accomplish without lot of mocking. IMO, removing the `namespaceSystem` is the simplest we can do to test. Let me just add that testcase.

pkg/authentication/bootstrapauthenticator/bootstrap_test.go
```diff
@@ -0,0 +1,104 @@
...
+						Namespace:         metav1.NamespaceSystem,
+						Name:              bootstrapUserBasicAuth,
+						CreationTimestamp: metav1.NewTime(time.Now().Add(2 * time.Hour)),
+					},
+					Data: map[string][]byte{bootstrapUserBasicAuth: []byte("foo")},
+				}),
+			expectedVal: false,
+		},
+		{
+			name: "some random error",
```
stlaz: put a better description here, please

stlaz:
/lgtm cancel
squash the commits please

pkg/authentication/bootstrapauthenticator/bootstrap.go
```diff
@@ -151,3 +132,37 @@ func (b *bootstrapUserDataGetter) Get() (*BootstrapUserData, bool, error) {
...
+	}
+	if secret.DeletionTimestamp != nil {
+		klog.V(4).Infof("%s secret is being deleted", bootstrapUserBasicAuth)
+		return nil, nil
+	}
+	namespace, err := b.namespaces.Get(metav1.NamespaceSystem, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	if secret.CreationTimestamp.After(namespace.CreationTimestamp.Add(time.Hour)) {
```
sttts: what is this heuristic? It needs docs why we do this.

stlaz: this is supposed to make it impossible for the bootstrap user to be configured again after it was once removed

PR #595, opened 2019-11-19
adds a list processor for migrating resources in parallel

p0lyn0mial:


pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+					klog.V(4).Infof("Relisting %v after retryable error: %v", gvr, err)
+					continue
+				}
+			}
+
+			migrationStarted := time.Now()
+			klog.V(4).Infof("Migrating %d objects of %v", len(allResource.Items), gvr)
+			err = p.processList(allResource)
+			if err != nil {
+				errs = append(errs, err)
```
p0lyn0mial: I couldn't decide what to do, should we continue or should we stop and restart.

In general, we can't tolerate even a single error, so it could happen that the last item in the list could restart the whole migration.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+	_, listErr := listPager.List(p.ctx, metav1.ListOptions{})
+	klog.V(4).Infof("Migration for %v finished in %v", gvr, time.Now().Sub(migrationStarted))
+	errs = append(errs, listErr)
+	return utilerrors.NewAggregate(errs)
+}
+
+func (p *ListProcessor) processList(l *unstructured.UnstructuredList) error {
+	workCh := make(chan *unstructured.Unstructured)
+	go func() {
+		defer close(workCh)
```
p0lyn0mial: should we be paranoid and add crash handling code?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+		}
+	}()
+
+	var wg sync.WaitGroup
+	wg.Add(p.concurrency)
+	errCh := make(chan error)
+	for i := 0; i < p.concurrency; i++ {
+		go func() {
+			defer wg.Done()
```
p0lyn0mial: should we be paranoid and add crash handling code?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+// Note that this operation block until all resources have been process, we can't get the next page or the context has been cancelled
+func (p *ListProcessor) Run(gvr schema.GroupVersionResource) error {
+	var errs []error
+	listPager := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+		for {
+			allResource, err := p.dynamicClient.Resource(gvr).List(opts)
+			if err != nil {
+				klog.Warningf("List of %v failed: %v", gvr, err)
+				if errors.IsResourceExpired(err) {
+					token, err := inconsistentContinueToken(err)
```
p0lyn0mial: this is how we will recover after a token expiry

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+// Run starts processing all the instance of the given GVR in batches.
+// Note that this operation block until all resources have been process, we can't get the next page or the context has been cancelled
+func (p *ListProcessor) Run(gvr schema.GroupVersionResource) error {
+	var errs []error
+	listPager := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
+		for {
+			allResource, err := p.dynamicClient.Resource(gvr).List(opts)
```
p0lyn0mial: we could also parallel pages retrieval

pkg/operator/encryption/controllers/migrators/inprocess.go
```diff
@@ -92,6 +90,7 @@ func (m *InProcessMigrator) EnsureMigration(gr schema.GroupResource, writeKey st
 }
 
 func (m *InProcessMigrator) runMigration(gvr schema.GroupVersionResource, writeKey string, stopCh <-chan struct{}, doneCh chan<- struct{}) {
+	fmt.Printf("runMigration for %v\n", gvr)
```
p0lyn0mial: remove

p0lyn0mial:
This code is [mirrored in](https://github.com/openshift/cluster-kube-apiserver-operator/pull/654) which additionally runs an e2e encryption test on a larger data set (~25K)

I [deliberately made](https://github.com/openshift/cluster-kube-apiserver-operator/pull/654/files#diff-5ef19c7cde0cb133eb26dc7d6f1b8568R41) the test to fail at the end because I wanted to print the logs (I know).

The following are the most important logs from [the test run](
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/654/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-operator-encryption/364/artifacts/e2e-aws-operator-encryption/unit_report.xml)
```
Created 26993 secrets 
Populating etcd took 34.816079932s 
Checking if all Secrets where encrypted/decrypted for "aescbc" mode 
Verified 27995 Secrets 
Checking if all ConfigMaps where encrypted/decrypted for "aescbc" mode 
Verified 303 ConfigMaps 
AssertSecretsAndConfigMaps took 2.45085828s 
encryption took 15m53.729762549s 
```

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
sttts: Don’t think that helps. We should rather keep processing while the next list is started.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,170 @@
...
+	migrationStarted := time.Now()
+	listPager.FullListIfExpired = false // prevent memory explosion from full list
+	_, listErr := listPager.List(p.ctx, metav1.ListOptions{})
+	klog.V(4).Infof("Migration for %v finished in %v", gvr, time.Now().Sub(migrationStarted))
+	errs = append(errs, listErr)
+	return utilerrors.NewAggregate(errs)
+}
+
+func (p *ListProcessor) processList(l *unstructured.UnstructuredList) error {
+	workCh := make(chan *unstructured.Unstructured)
```
sttts: give it some buffer

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
sttts: we should

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,178 @@
...
+	_, listErr := listPager.List(p.ctx, metav1.ListOptions{})
+	klog.V(4).Infof("Migration for %v finished in %v", gvr, time.Now().Sub(migrationStarted))
+	errs = append(errs, listErr)
+	return utilerrors.NewAggregate(errs)
+}
+
+func (p *ListProcessor) processList(l *unstructured.UnstructuredList) error {
+	workCh := make(chan *unstructured.Unstructured, p.concurrency)
+	go func() {
+		defer utilruntime.HandleCrash()
```
p0lyn0mial: I hope it's okay to use it, do we ever set `ReallyCrash` to `false`?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
p0lyn0mial: @sttts wdyt ?

pkg/operator/encryption/controllers/migrators/inprocess.go
```diff
@@ -3,19 +3,17 @@ package migrators
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
```
mfojtik: nit: move down

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
mfojtik: yes

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,178 @@
...
+	for err := range errCh {
+		errors = append(errors, err)
+	}
+	return utilerrors.NewAggregate(errors)
+}
+
+func (p *ListProcessor) worker(workCh <-chan *unstructured.Unstructured, errCh chan<- error) {
+	for item := range workCh {
+		err := executeWorkerFunc(p.workerFn, item)
+		if err != nil {
```
mfojtik: nit: 
```
if err == nil {
  continue
}
```

(nested code is harder to read)

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,178 @@
...
+			case <-p.ctx.Done():
+				return
+			}
+		}
+	}
+}
+
+func executeWorkerFunc(workerFn WorkerFunc, obj *unstructured.Unstructured) (result error) {
+	defer func() {
+		if r := recover(); r != nil {
```
mfojtik: isn't there something like util.HandlePanic() ?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,178 @@
...
+			if err, ok := r.(error); ok {
+				result = err
+			} else {
+				result = fmt.Errorf("panic: %v", r)
+			}
+		}
+	}()
+	result = workerFn(obj)
```
mfojtik: return workerFn(obj)

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
p0lyn0mial: not if you want to capture the error and return it, notice that this function makes use of `named return value`, panic will be captured, transformed to an error and returned to the caller.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
p0lyn0mial: we could but I would prefer to highlight that this function makes use of `named return value`, wdyt?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,179 @@
...
+	"k8s.io/client-go/tools/pager"
+	"k8s.io/klog"
+)
+
+const (
+	defaultConcurrency = 10
+)
+
+// WorkerFunc function that is executed by workers to process a single item
+type WorkerFunc func(*unstructured.Unstructured) error
```
sttts: do we have to expose any symbol here?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
sttts: I think we have to stop over, i.e. return this error directly.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,179 @@
...
+			defer utilruntime.HandleCrash()
+			defer wg.Done()
+			p.worker(workCh, errCh)
+		}()
+	}
+
+	go func() {
+		defer utilruntime.HandleCrash()
+		wg.Wait()
+		close(errCh)
```
sttts: when the first worker throws an error, we can stop. Now we keep processing the whole page in the non-erroring workes, which can take minutes.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (p *ListProcessor) worker(workCh <-chan *unstructured.Unstructured, errCh chan<- error) {
+	for item := range workCh {
+		err := executeWorkerFunc(p.workerFn, item)
+		if err == nil {
+			continue
+		}
+		select {
+		case errCh <- err:
+			continue
```
sttts: what does this mean: a case with a channel write? It's a non-blocking write? In other words, this whole loop hot-loops?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,185 @@
...
+	listPager.FullListIfExpired = false // prevent memory explosion from full list
+	_, listErr := listPager.List(p.ctx, metav1.ListOptions{})
+	klog.V(4).Infof("Migration for %v finished in %v", gvr, time.Now().Sub(migrationStarted))
+	errs = append(errs, listErr)
+	return utilerrors.NewAggregate(errs)
+}
+
+func (p *ListProcessor) processList(l *unstructured.UnstructuredList) error {
+	workCh := make(chan *unstructured.Unstructured, p.concurrency)
+	onWorkerErrorCtx, onWorkerErrorCancel := context.WithCancel(context.Background())
```
sttts: can we do `ctx := WithCancel(p.ctx)` and only wait for that one?

p0lyn0mial: we could but then I think we will not collect all the potential errors as workers are also listening on that channel.

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,185 @@
...
+func (p *ListProcessor) worker(workCh <-chan *unstructured.Unstructured, errCh chan<- error) {
+	for item := range workCh {
+		err := executeWorkerFunc(p.workerFn, item)
+		if err == nil {
+			continue
+		}
+		select {
+		case errCh <- err:
+			continue
+		case <-p.ctx.Done():
```
sttts: why do we need this case?

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (p *ListProcessor) worker(workCh <-chan *unstructured.Unstructured, errCh chan<- error) {
+	for item := range workCh {
+		err := executeWorkerFunc(p.workerFn, item)
+		if err == nil {
+			continue
+		}
+		select {
+		case errCh <- err:
+			continue
```
sttts: return

pkg/operator/encryption/controllers/migrators/inprocess_processor.go
p0lyn0mial: nothing, it has no effect, `continue` begins the new iteration line `142`, I can remove it and it will work exactly in the same way.

p0lyn0mial:
> ci/prow/e2e-aws-encryption

seems to be infra related
```
level=error msg="When applying changes to module.bootstrap.aws_s3_bucket.ignition, provider"
level=error msg="\"aws\" produced an unexpected new value for was present, but now absent."
```

PR #598, opened 2019-11-21
Bug 1772190: certsyncpod: reduce amount of events produced by removing files

mfojtik:
As reported, the `CertificateRemoved` might cause event spam when the secret is not found permanently. This will reduce the events only to fire when we actually remove the file.

### DIFF
--- a/pkg/operator/staticpod/certsyncpod/certsync_controller.go
+++ b/pkg/operator/staticpod/certsyncpod/certsync_controller.go
@@ -16,7 +16,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	corev1interface "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/listers/core/v1"
+	v1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -198,11 +198,14 @@ func (c *CertSyncController) sync() error {
...
 
 			// remove missing content
-			if err := os.RemoveAll(getSecretDir(c.destinationDir, s.Name)); err != nil {
-				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed removing file for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
-				errors = append(errors, err)
+			secretFile := getSecretDir(c.destinationDir, s.Name)
+			if _, err := os.Stat(secretFile); !os.IsNotExist(err) {
+				if err := os.RemoveAll(secretFile); err != nil {
+					c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed removing file for missing secret: %s/%s: %v", secret.Namespace, secret.Name, err)
+					errors = append(errors, err)
```
p0lyn0mial: I think that we have to `continue` otherwise we will fire `CertificateRemoved.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -198,11 +198,14 @@ func (c *CertSyncController) sync() error {
 			}
 
 			// remove missing content
-			if err := os.RemoveAll(getSecretDir(c.destinationDir, s.Name)); err != nil {
-				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed removing file for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
-				errors = append(errors, err)
+			secretFile := getSecretDir(c.destinationDir, s.Name)
+			if _, err := os.Stat(secretFile); !os.IsNotExist(err) {
```
p0lyn0mial: what if `err` satisfies `!os.IsNotExist(err)` but is != `nil`?

mfojtik: good catch, fixed

pkg/operator/staticpod/certsyncpod/certsync_controller.go
mfojtik: i inverted the logic here and checking if the file exists before deleting it in separate `if`.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -197,12 +197,19 @@ func (c *CertSyncController) sync() error {
...
-			if err := os.RemoveAll(getSecretDir(c.destinationDir, s.Name)); err != nil {
-				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed removing file for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
+			if err := os.RemoveAll(secretFile); err != nil {
+				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed removing file for missing secret: %s/%s: %v", secret.Namespace, secret.Name, err)
 				errors = append(errors, err)
+				continue
 			}
-			c.eventRecorder.Warningf("CertificateRemoved", "Removed file for secret: %s/%s", secret.Namespace, secret.Name, err)
+			c.eventRecorder.Warningf("CertificateRemoved", "Removed file for missing secret: %s/%s", secret.Namespace, secret.Name)
 			continue
```
p0lyn0mial: nit: I think that we don't need `continue` at the end of `case` clause.

mfojtik: @p0lyn0mial i think we do, otherwise it will escape down :-)

p0lyn0mial: ah, ic, we would escape `switch` statement :)

PR #602, opened 2019-11-24
add helper functions for running prometheus query

sallyom:
@marun @sanchezl this PR provides helper functions for prometheus query

I placed a fake bump/testing here:
* https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316
* https://github.com/openshift/cluster-kube-scheduler-operator/pull/191

/cc @sttts

### DIFF
--- a/test/library/metrics/query.go
+++ b/test/library/metrics/query.go
@@ -1,0 +1,95 @@
+package metrics
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/transport"
+
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+)
+
+// NewPrometheusClient returns Prometheus API or error
+// Note: with thanos-querier you must pass an entire Alert as a query. Partial queries return an error, so have to pass the entire alert.
+// Example query for an Alert:
+// `ALERTS{alertname="PodDisruptionBudgetAtLimit",alertstate="pending",namespace="pdbnamespace",poddisruptionbudget="pdbname",prometheus="openshift-monitoring/k8s",service="kube-state-metrics",severity="warning"}==1`
+// Example query:
+// `scheduler_scheduling_duration_seconds_sum`
+func NewPrometheusClient(kclient *kubernetes.Clientset, rc *routeclient.Clientset) (prometheusv1.API, error) {
+	_, err := kclient.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	route, err := rc.RouteV1().Routes("openshift-monitoring").Get("thanos-querier", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	host := route.Status.Ingress[0].Host
+	var bearerToken string
+	secrets, err := kclient.CoreV1().Secrets("openshift-monitoring").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
+	}
+	for _, s := range secrets.Items {
+		if s.Type != corev1.SecretTypeServiceAccountToken ||
+			!strings.HasPrefix(s.Name, "prometheus-k8s") {
+			continue
+		}
+		bearerToken = string(s.Data[corev1.ServiceAccountTokenKey])
+		break
+	}
+	if len(bearerToken) == 0 {
+		return nil, fmt.Errorf("prometheus service account not found")
+	}
+
+	return createClient(kclient, host, bearerToken)
+}
+
+func createClient(kclient *kubernetes.Clientset, host, bearerToken string) (prometheusv1.API, error) {
+	// retrieve router CA
+	routerCAConfigMap, err := kclient.CoreV1().ConfigMaps("openshift-config-managed").Get("router-ca", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	bundlePEM := []byte(routerCAConfigMap.Data["ca-bundle.crt"])
+
+	// make a client connection configured with the provided bundle.
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM(bundlePEM)
+
+	// prometheus API client, configured for route host and bearer token auth
+	client, err := prometheusapi.NewClient(prometheusapi.Config{
+		Address: "https://" + host,
+		RoundTripper: transport.NewBearerAuthRoundTripper(
+			bearerToken,
+			&http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
+				TLSHandshakeTimeout: 10 * time.Second,
+				TLSClientConfig: &tls.Config{
+					RootCAs:    roots,
+					ServerName: host,
+				},
+			},
+		),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return prometheusv1.NewAPI(client), nil
+}
--- a/vendor/github.com/prometheus/client_golang/api/client.go
+++ b/vendor/github.com/prometheus/client_golang/api/client.go
@@ -1,0 +1,156 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package api provides clients for the HTTP APIs.
+package api
+
+import (
+	"context"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+)
+
+type Warnings []string
+
+// DefaultRoundTripper is used if no RoundTripper is set in Config.
+var DefaultRoundTripper http.RoundTripper = &http.Transport{
+	Proxy: http.ProxyFromEnvironment,
+	DialContext: (&net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}).DialContext,
+	TLSHandshakeTimeout: 10 * time.Second,
+}
+
+// Config defines configuration parameters for a new client.
+type Config struct {
+	// The address of the Prometheus to connect to.
+	Address string
+
+	// RoundTripper is used by the Client to drive HTTP requests. If not
+	// provided, DefaultRoundTripper will be used.
+	RoundTripper http.RoundTripper
+}
+
+func (cfg *Config) roundTripper() http.RoundTripper {
+	if cfg.RoundTripper == nil {
+		return DefaultRoundTripper
+	}
+	return cfg.RoundTripper
+}
+
+// Client is the interface for an API client.
+type Client interface {
+	URL(ep string, args map[string]string) *url.URL
+	Do(context.Context, *http.Request) (*http.Response, []byte, Warnings, error)
+}
+
+// DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request.
+func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, Warnings, error) {
+	req, err := http.NewRequest(http.MethodPost, u.String(), strings.NewReader(args.Encode()))
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	resp, body, warnings, err := c.Do(ctx, req)
+	if resp != nil && resp.StatusCode == http.StatusMethodNotAllowed {
+		u.RawQuery = args.Encode()
+		req, err = http.NewRequest(http.MethodGet, u.String(), nil)
+		if err != nil {
+			return nil, nil, warnings, err
+		}
+
+	} else {
+		if err != nil {
+			return resp, body, warnings, err
+		}
+		return resp, body, warnings, nil
+	}
+	return c.Do(ctx, req)
+}
+
+// NewClient returns a new Client.
+//
+// It is safe to use the returned Client from multiple goroutines.
+func NewClient(cfg Config) (Client, error) {
+	u, err := url.Parse(cfg.Address)
+	if err != nil {
+		return nil, err
+	}
+	u.Path = strings.TrimRight(u.Path, "/")
+
+	return &httpClient{
+		endpoint: u,
+		client:   http.Client{Transport: cfg.roundTripper()},
+	}, nil
+}
+
+type httpClient struct {
+	endpoint *url.URL
+	client   http.Client
+}
+
+func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
+	p := path.Join(c.endpoint.Path, ep)
+
+	for arg, val := range args {
+		arg = ":" + arg
+		p = strings.Replace(p, arg, val, -1)
+	}
+
+	u := *c.endpoint
+	u.Path = p
+
+	return &u
+}
+
+func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, Warnings, error) {
+	if ctx != nil {
+		req = req.WithContext(ctx)
+	}
+	resp, err := c.client.Do(req)
+	defer func() {
+		if resp != nil {
+			resp.Body.Close()
+		}
+	}()
+
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	var body []byte
+	done := make(chan struct{})
+	go func() {
+		body, err = ioutil.ReadAll(resp.Body)
+		close(done)
+	}()
+
+	select {
+	case <-ctx.Done():
+		<-done
+		err = resp.Body.Close()
+		if err == nil {
+			err = ctx.Err()
+		}
+	case <-done:
+	}
+
+	return resp, body, nil, err
+}
--- a/vendor/github.com/prometheus/client_golang/api/prometheus/v1/api.go
+++ b/vendor/github.com/prometheus/client_golang/api/prometheus/v1/api.go
@@ -1,0 +1,877 @@
+// Copyright 2017 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package v1 provides bindings to the Prometheus HTTP API v1:
+// http://prometheus.io/docs/querying/api/
+package v1
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"net/http"
+	"strconv"
+	"time"
+	"unsafe"
+
+	json "github.com/json-iterator/go"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/prometheus/client_golang/api"
+)
+
+func init() {
+	json.RegisterTypeEncoderFunc("model.SamplePair", marshalPointJSON, marshalPointJSONIsEmpty)
+	json.RegisterTypeDecoderFunc("model.SamplePair", unMarshalPointJSON)
+}
+
+func unMarshalPointJSON(ptr unsafe.Pointer, iter *json.Iterator) {
+	p := (*model.SamplePair)(ptr)
+	if !iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "SamplePair must be [timestamp, value]")
+		return
+	}
+	t := iter.ReadNumber()
+	if err := p.Timestamp.UnmarshalJSON([]byte(t)); err != nil {
+		iter.ReportError("unmarshal model.SamplePair", err.Error())
+		return
+	}
+	if !iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "SamplePair missing value")
+		return
+	}
+
+	f, err := strconv.ParseFloat(iter.ReadString(), 64)
+	if err != nil {
+		iter.ReportError("unmarshal model.SamplePair", err.Error())
+		return
+	}
+	p.Value = model.SampleValue(f)
+
+	if iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "SamplePair has too many values, must be [timestamp, value]")
+		return
+	}
+}
+
+func marshalPointJSON(ptr unsafe.Pointer, stream *json.Stream) {
+	p := *((*model.SamplePair)(ptr))
+	stream.WriteArrayStart()
+	// Write out the timestamp as a float divided by 1000.
+	// This is ~3x faster than converting to a float.
+	t := int64(p.Timestamp)
+	if t < 0 {
+		stream.WriteRaw(`-`)
+		t = -t
+	}
+	stream.WriteInt64(t / 1000)
+	fraction := t % 1000
+	if fraction != 0 {
+		stream.WriteRaw(`.`)
+		if fraction < 100 {
+			stream.WriteRaw(`0`)
+		}
+		if fraction < 10 {
+			stream.WriteRaw(`0`)
+		}
+		stream.WriteInt64(fraction)
+	}
+	stream.WriteMore()
+	stream.WriteRaw(`"`)
+
+	// Taken from https://github.com/json-iterator/go/blob/master/stream_float.go#L71 as a workaround
+	// to https://github.com/json-iterator/go/issues/365 (jsoniter, to follow json standard, doesn't allow inf/nan)
+	buf := stream.Buffer()
+	abs := math.Abs(float64(p.Value))
+	fmt := byte('f')
+	// Note: Must use float32 comparisons for underlying float32 value to get precise cutoffs right.
+	if abs != 0 {
+		if abs < 1e-6 || abs >= 1e21 {
+			fmt = 'e'
+			fmt = 'e'
+		}
+	}
+	buf = strconv.AppendFloat(buf, float64(p.Value), fmt, -1, 64)
+	stream.SetBuffer(buf)
+
+	stream.WriteRaw(`"`)
+	stream.WriteArrayEnd()
+
+}
+
+func marshalPointJSONIsEmpty(ptr unsafe.Pointer) bool {
+	return false
+}
+
+const (
+	statusAPIError = 422
+
+	apiPrefix = "/api/v1"
+
+	epAlerts          = apiPrefix + "/alerts"
+	epAlertManagers   = apiPrefix + "/alertmanagers"
+	epQuery           = apiPrefix + "/query"
+	epQueryRange      = apiPrefix + "/query_range"
+	epLabels          = apiPrefix + "/labels"
+	epLabelValues     = apiPrefix + "/label/:name/values"
+	epSeries          = apiPrefix + "/series"
+	epTargets         = apiPrefix + "/targets"
+	epTargetsMetadata = apiPrefix + "/targets/metadata"
+	epRules           = apiPrefix + "/rules"
+	epSnapshot        = apiPrefix + "/admin/tsdb/snapshot"
+	epDeleteSeries    = apiPrefix + "/admin/tsdb/delete_series"
+	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
+	epConfig          = apiPrefix + "/status/config"
+	epFlags           = apiPrefix + "/status/flags"
+)
+
+// AlertState models the state of an alert.
+type AlertState string
+
+// ErrorType models the different API error types.
+type ErrorType string
+
+// HealthStatus models the health status of a scrape target.
+type HealthStatus string
+
+// RuleType models the type of a rule.
+type RuleType string
+
+// RuleHealth models the health status of a rule.
+type RuleHealth string
+
+// MetricType models the type of a metric.
+type MetricType string
+
+const (
+	// Possible values for AlertState.
+	AlertStateFiring   AlertState = "firing"
+	AlertStateInactive AlertState = "inactive"
+	AlertStatePending  AlertState = "pending"
+
+	// Possible values for ErrorType.
+	ErrBadData     ErrorType = "bad_data"
+	ErrTimeout     ErrorType = "timeout"
+	ErrCanceled    ErrorType = "canceled"
+	ErrExec        ErrorType = "execution"
+	ErrBadResponse ErrorType = "bad_response"
+	ErrServer      ErrorType = "server_error"
+	ErrClient      ErrorType = "client_error"
+
+	// Possible values for HealthStatus.
+	HealthGood    HealthStatus = "up"
+	HealthUnknown HealthStatus = "unknown"
+	HealthBad     HealthStatus = "down"
+
+	// Possible values for RuleType.
+	RuleTypeRecording RuleType = "recording"
+	RuleTypeAlerting  RuleType = "alerting"
+
+	// Possible values for RuleHealth.
+	RuleHealthGood    = "ok"
+	RuleHealthUnknown = "unknown"
+	RuleHealthBad     = "err"
+
+	// Possible values for MetricType
+	MetricTypeCounter        MetricType = "counter"
+	MetricTypeGauge          MetricType = "gauge"
+	MetricTypeHistogram      MetricType = "histogram"
+	MetricTypeGaugeHistogram MetricType = "gaugehistogram"
+	MetricTypeSummary        MetricType = "summary"
+	MetricTypeInfo           MetricType = "info"
+	MetricTypeStateset       MetricType = "stateset"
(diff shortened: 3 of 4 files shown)


pkg/metrics/query.go
```diff
@@ -0,0 +1,218 @@
...
+		},
+	})
+	return err
+}
+
+// pollForCABundle returns the given configmap data
+func (pq *PromQuery) pollForCABundle() ([]byte, error) {
+	expectedDataSize := 1
+	var ca []byte
+	err := wait.Poll(time.Second*5, time.Second*300, func() (bool, error) {
```
marun: Is there a reason not to use shorter intervals - 1s and 30s for example?

pkg/metrics/query.go
```diff
@@ -0,0 +1,218 @@
...
+type PromQuery struct {
+	Client        *kubernetes.Clientset
+	Query         string
+	ConfigMapName string
+	NsName        string
+	URL           string
+	BearerToken   string
+}
+
+func NewPromQuery(client *kubernetes.Clientset, query, cmName, nsName string) (*PromQuery, error) {
```
marun: I suggest separating ca bundle retrieval from prometheus querying. Ideally something like the following:

```golang
// RetrieveServiceCABundle would internally create an annotated configmap in the provided namespace, retrieve the configmap from it, and delete it. 
caBundle, err := RetrieveServiceCABundle(client, namespace)

// NewPrometheusClient would do the work of initializing a client capable of making queries
prometheusClient, err := NewPrometheusClient(client, namespace, cabundle)

// Actually run the query
result, err := prometheusClient.Query(...)
```

pkg/metrics/query.go
marun: Is there a reason not to use shorter intervals - 1s and 30s for example?

pkg/metrics/query.go
```diff
@@ -0,0 +1,218 @@
...
+			return false, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
+		}
+		for _, secret := range secrets.Items {
+			if secret.Type != corev1.SecretTypeServiceAccountToken {
+				continue
+			}
+			if !strings.HasPrefix(secret.Name, "prometheus-") {
+				continue
+			}
+			bearerToken = string(secret.Data[corev1.ServiceAccountTokenKey])
```
marun: Is this the only way to get access to prometheus?  I'm wondering if there is an option that doesn't require the same privileges as prometheus itself.

pkg/metrics/query.go
sallyom: nope, shortened, thnks

pkg/metrics/query.go
sallyom: updated, _much better thank you_

pkg/metrics/query.go
sallyom: this is how it's done in [origin/test/extended/prometheus/prometheus.go](https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go#L469-#L498)  that's what I based this off of.

sallyom:
@marun thanks for the review, I've updated, ready for another review.  Also, see this functioning in the [cluster-kube-controller-operator PR](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316).

sallyom:
/hold

there's something off w/ this.. will figure it out tomorrow

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	Status string                 `json:"status"`
+	Data   prometheusResponseData `json:"data"`
+}
+
+type prometheusResponseData struct {
+	ResultType string       `json:"resultType"`
+	Result     model.Vector `json:"result"`
+}
+
+func (pq *PrometheusClient) getQueryResponse() ([]byte, error) {
```
marun: Maybe s/getQueryResponse/makeRequest/ ?

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	err := wait.Poll(time.Second*1, time.Second*waitForPrometheusStartSeconds, func() (bool, error) {
+		//TODO when the http/query apis discussed at https://github.com/prometheus/client_golang#client-for-the-prometheus-http-api
+		// and introduced at https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go are vendored into
+		// openshift/origin, look to replace this homegrown http request / query param with that API
+		contents, err := pq.getQueryResponse()
+		if err != nil {
+			return false, fmt.Errorf("waiting for query results from %s", pq.URL)
+		}
+		result := prometheusResponse{}
+		json.Unmarshal(contents, &result)
```
marun: Please check the error returned by `Unmarshal`.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	//cmd := fmt.Sprintf("curl -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
+	req, err := http.NewRequest("GET", pq.URL, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Authorization", pq.BearerToken)
+
+	resp, err := pq.Client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("curl command failed: %v\n%v", err, resp)
```
marun: I'm assuming this error message is a legacy of a previous implementation, and should probably be updated to reflect the switch to using golang http rather than curl.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func LocatePrometheus(client *kubernetes.Clientset) (string, bool) {
+	_, err := client.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if kapierrs.IsNotFound(err) {
+		return "", false
+	}
+
+	var bearerToken string
+	err = wait.Poll(time.Second*1, time.Second*30, func() (bool, error) {
```
marun: The following comments are relevant to all uses in this PR of `wait.Poll`:

- `wait.Poll` waits for the interval before the first invocation. `wait.PollImediate` appears to be preferable in this PR so that the first invocation is executed immediately. 

- Please follow the advice @sttts provided in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316 regarding timeout.

- Suggest logging recoverable errors (i.e. via klog at warning or info level) rather than returning them to ensure that polling can continue. Only fatal errors should be returned to a `wait.Poll*` method since any non-nil error will trigger exit. 

- Note that this strategy often means that a polling method only ever reports a timeout error. Diagnosing the reason for a timeout failure then requires reviewing the logs for an indication of the recoverable error(s). In cases where more detail needs to appear directly (not necessarily in this PR), the last recoverable error can be tracked and returned instead of the timeout error.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+		secrets, err := client.CoreV1().Secrets("openshift-monitoring").List(metav1.ListOptions{})
+		if err != nil {
+			return false, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
```
marun: Please embed the original error in the message to simplify failure analysis.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+			return true, nil
+		}
+	})
+	if err != nil {
+		return "", false
+	}
+	return bearerToken, true
+}
+
+const waitForPrometheusStartSeconds = 240
```
marun: I'm assuming this is necessary in the context of a prometheus-specific test, but can an operator test expect prometheus to be started prior to being invoked? 

If not, would it make sense to add a separate method (e.g. `WaitForPrometheus`) that waited for a health check of prometheus? That would ensure a separation between startup- and query-related failure.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	Data   prometheusResponseData `json:"data"`
+}
+
+type prometheusResponseData struct {
+	ResultType string       `json:"resultType"`
+	Result     model.Vector `json:"result"`
+}
+
+func (pq *PrometheusClient) getQueryResponse() ([]byte, error) {
+	//cmd := fmt.Sprintf("curl -s -k -H 'Authorization: Bearer %s' %q", bearer, url)
```
marun: Please remove.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	return body, nil
+}
+
+// RunPrometheusQuery() checks that query returns non-empty result, errors if query returns empty, and returns metrics
+// example: query := `ALERTS{alertstate="pending",alertname="PodDisruptionBudgetAtLimit",severity="warning"} == 1`
+func (pq *PrometheusClient) RunPrometheusQuery() (model.Vector, error) {
+	// expect all correct metrics within a reasonable time period
+        var metrics model.Vector
```
marun: gofmt?

marun: hmm, or just `make verify`

pkg/metrics/query.go
sallyom: I've added helpers to run exec curl - this seems to be the only way to exercise the cluster CA

sallyom:
> +1 to the separation of concerns you are pursuing.
> 
> What do you think of adding an e2e test that exercises the code?
yea, was thinking this, looking into it.

sallyom:
> @sallyom since these are meant as testing helpers please move them to `pkg/test/library/metrics` I don't want anyone accidentally figuring this one out and using it for other purpose than testing.

oh right! thanks, will do

pkg/metrics/query.go
sallyom: yes, my gofmt was broken somewhere between go.1.12(needed to compile library-go) and go.13.4(needed to compile kubectl) It's working/fixed now.

pkg/metrics/query.go
sallyom: after days of experimenting, i've come to conclusion that best way to exercise CA is to run curl from a pod - any other way I've tried requires eq of `curl -k` .  I've implemented the exec pod.

pkg/metrics/query.go
sallyom: i've removed this, thanks for your patience while I experiment/break/fix everything. I need to be better about `WIP` tagging, but I value every review : )

pkg/metrics/query.go
sallyom: updated

pkg/metrics/query.go
sallyom: updated.  yup ^^ figured that all out through much trial and error

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	return body, nil
+}
+
+// RunPrometheusQuery() checks that query returns non-empty result, errors if query returns empty, and returns metrics
+// example: query := `ALERTS{alertstate="pending",alertname="PodDisruptionBudgetAtLimit",severity="warning"} == 1`
+func (pq *PrometheusClient) RunPrometheusQuery() (model.Vector, error) {
```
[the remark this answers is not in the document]
sallyom: fixed, : )

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type PrometheusClient struct {
+	Client      *http.Client
+	URL         string
+	BearerToken string
+}
+
+func NewPrometheusClient(caBundle []byte, query, bearerToken string) (*PrometheusClient, error) {
+	url := fmt.Sprintf("%s/api/v1/query?%s", metricsBaseURL, (url.Values{"query": []string{query}}).Encode())
```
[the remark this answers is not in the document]
sallyom: i've removed/updated code to the point of this no longer being relevant lol

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return metrics, nil
+}
+
+func RetrieveServiceCABundle(client *kubernetes.Clientset, configMapName, nsName string) ([]byte, error) {
+	// Prompt the injection of the ca bundle into a configmap
+	err := createAnnotatedCABundleInjectionConfigMap(client, configMapName, nsName)
```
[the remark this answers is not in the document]
sallyom: TODO: cleanup - currently removing ns where these functions are called.. will look at if this is sufficient.

pkg/metrics/query.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, err
+	}
+	return ca, nil
+}
+
+func createAnnotatedCABundleInjectionConfigMap(client *kubernetes.Clientset, configMapName, nsName string) error {
+	_, err := client.CoreV1().ConfigMaps(nsName).Create(&corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: configMapName,
```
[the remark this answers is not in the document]
sallyom: updated

sallyom:
Thanks, @marun for your patience while I worked this out.  Note, only way I've found (through days of trial and error) to utilize internal prometheus svc url + cluster CA bundle is through secure curl, within a pod.  This is working nicely.  Will clean up/possibly move to a single e2e here in lib-go rather than in each operator repo.

test/metrics/query.go
```diff
@@ -0,0 +1,260 @@
...
+
+	"github.com/prometheus/common/model"
+)
+
+const (
+	prometheusBaseURL         = "https://prometheus-k8s.openshift-monitoring.svc:9091"
+	serviceCAInjectAnnotation = "service.beta.openshift.io/inject-cabundle"
+)
+
+func LocatePrometheus(client *kubernetes.Clientset) (string, error) {
```
soltysh: Missing godoc, what are the return values.

test/metrics/query.go
```diff
@@ -0,0 +1,260 @@
...
+	Data   prometheusResponseData `json:"data"`
+}
+
+type prometheusResponseData struct {
+	ResultType string       `json:"resultType"`
+	Result     model.Vector `json:"result"`
+}
+
+// CreateAnnotatedCABundleConfigMap results in configmap injected with CA bundle issued by service-ca-operator
+func CreateAnnotatedCABundleInjectionConfigMap(client *kubernetes.Clientset, configMapName, nsName string) error {
```
soltysh: Worth noting down that the method also waits for that CM to be created.

soltysh: `CreateAndWaitFor...` would be a better name

test/metrics/query.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		result := prometheusResponse{}
+		json.Unmarshal([]byte(output), &result)
+		metrics = result.Data.Result
+		if len(metrics) == 0 {
+			return false, nil
+		}
+		return true, nil
+	})
+	if len(metrics) == 0 {
```
soltysh: Analysing whether output is correct or not should be left to the caller.

test/metrics/query.go
```diff
@@ -0,0 +1,260 @@
...
+		return metrics, fmt.Errorf("prometheus returned unexpected results")
+	}
+	return metrics, nil
+}
+
+// PodExecCreate creates a pod from which an exec command can run, to run secure curl command to scrape metrics
+func PodExecCreate(client *kubernetes.Clientset, name, cmName string, labels map[string]string) error {
```
soltysh: `RunPod` maybe?

test/metrics/query.go
```diff
@@ -0,0 +1,260 @@
...
+	defaultArgs := []string{}
+	defaultArgs = append(defaultArgs, "--kubeconfig="+kubeconfig)
+	kubectlArgs := append(defaultArgs, args...)
+	cmd := exec.Command(kubectl, kubectlArgs...)
+	return cmd
+}
+
+// runExecCmd runs the given cmd in the context of the given pod using `kubectl exec`
+// inside of a shell.
+func runExecCmd(kubeconfig, kubectl, name, cmd string) (string, error) {
```
soltysh: There's a lot of these functions, squash to a single one? If we need to we'll extract helpers as we go.

test/metrics/query.go
sallyom: updated

sallyom: updated

test/metrics/query.go
sallyom: updated

test/metrics/query.go
sallyom: updated, thanks

test/metrics/query.go
sallyom: PodCreate? updated.

test/metrics/query.go
sallyom: ok, will do

sallyom:
@marun I also added a route client and accessed the external route, but only way I could get it working is with InsecureSkipVerify in TLSConfig.. looking at cluster-monitor tests now...

sallyom:
> I trust that your work in accessing an internal service via a pod was instructive! :)
> 
> A quick check of how the monitoring operator tests itself reveals the following:
> 
> https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go
> 
https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go#L83  that's what I was working to avoid

pkg/metrics/query.go
sallyom: never mind, we've determined that running the query against the external prometheus URL, no exec, no CAbundle required is sufficient

pkg/metrics/query.go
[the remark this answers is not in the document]
sallyom: updated, no longer necessary

pkg/metrics/query.go
sallyom: lol nm, CA bundle no longer required - so long as external url is accessible, we can verify that metrics are being served/certs are current.

test/metrics/query.go
sallyom: no longer required w/ querying against external url, no exec required.

test/metrics/query.go
sallyom: no longer requred

sallyom:
> In either case, a new constructor for the client is suggested that accepts the ca bundle and validates the connection if provided.

I've backed up to an earlier version of helper functions, now that we've decided CAbundle is not required to verify metrics - it's  very similar to the cluster-monitoring tests you pointed to : ) note that cluster-monitoring tests don't use CAbundle, but configure `InsecureSkipVerify` - If that's sufficient,  all you need to query is a) prometheus route and b) bearer token from service account secret - it's very simple now.

pkg/metrics/query.go
sallyom: no longer required, using InsecureSkipVerify, no CAbundle retrieval

sallyom:
> You might want to move this even further down to `test/library/metrics` directory, right next to encryption directory.
> Also we need @deads2k or @sttts to review and approve this.

moved to test/library/metrics dir, and now much simpler w/out exec/CA retrieval, etc.

marun:
Is there a reason to reimplement the [monitoring test client](https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go) vs moving or copying to library-go?

sallyom:
> Is there a reason to reimplement the [monitoring test client](https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go) vs moving or copying to library-go?

nope, but all you need is a)bearerToken and b)route, it's very simple is all, and I already had it there.  I can move/copy that cluster-monitoring-operator code to library-go, no problem.

marun:
> > Is there a reason to reimplement the [monitoring test client](https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go) vs moving or copying to library-go?
> 
> nope, but all you need is a)bearerToken and b)route, it's very simple is all, and I already had it there. I can move/copy that cluster-monitoring-operator code to library-go, no problem.

I think it's preferable to reuse the code maintained by the monitoring team (i.e. the best code is no (new) code).

sallyom:
> > > Is there a reason to reimplement the [monitoring test client](https://github.com/openshift/cluster-monitoring-operator/blob/master/test/e2e/framework/route_client.go) vs moving or copying to library-go?
> > 
> > 
> > nope, but all you need is a)bearerToken and b)route, it's very simple is all, and I already had it there. I can move/copy that cluster-monitoring-operator code to library-go, no problem.
> 
> I think it's preferable to reuse the code maintained by the monitoring team (i.e. the best code is no (new) code).

@s-urbaniak cluster-monitoring team ok w/ moving that route_client.go to `library-go/test/library/something`? It's your code, I don't want to copy it, better if you move it.  If so, I'll close this PR. : )

s-urbaniak:
@sallyom as discussed out of band. Sure thing, if it helps other teams I can submit a contribution to library-go.

sallyom:
@marun @s-urbaniak on 2nd thought (or more), I think it makes sense to add just what's here in this PR to library-go.  These are helpers for any operator to verify their metrics are accessible.  I've taken a closer look at the cluster-monitoring-operator test/e2e/framework and it would add a lot of code here that we don't need.  This single file is all we need (at least for the 2 operators' e2es I'm writing) and if we find this isn't sufficient, we can bring in cluster-monitoring-operator framework later.  See my test PR here, using this added file's fns: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+	"github.com/prometheus/common/model"
+)
+
+type PrometheusClient struct {
```
mfojtik: i would make this private struct and return `interface` in NewPrometheusClient below... It allows mocking or replacing the client in future. Also a public struct with all fields private is strange to me.

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+	// Host address of Prometheus public route
+	host string
+	// ServiceAccount bearer token to pass though OpenShift oauth proxy
+	token string
+}
+
+func NewPrometheusClient(client *kubernetes.Clientset, rc *routeclient.Clientset) (*PrometheusClient, error) {
```
mfojtik: nit: godoc

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+	_, err := client.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	route, err := rc.RouteV1().Routes("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	host := route.Spec.Host
```
mfojtik: this is bad, use `route.Status` to retrieve the host from `RouteIngress`

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+			continue
+		}
+		if !strings.HasPrefix(secret.Name, "prometheus-") {
+			continue
+		}
+		bearerToken = string(secret.Data[corev1.ServiceAccountTokenKey])
+		break
+	}
+	if len(bearerToken) == 0 {
+		return nil, fmt.Errorf("error getting bearer token from prometheus service account")
```
mfojtik: "prometheus service account not found" would be better :)

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+	var bearerToken string
+	secrets, err := client.CoreV1().Secrets("openshift-monitoring").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
+	}
+	for _, secret := range secrets.Items {
+		if secret.Type != corev1.SecretTypeServiceAccountToken {
+			continue
+		}
+		if !strings.HasPrefix(secret.Name, "prometheus-") {
```
mfojtik: nit: merge this with the `if` above?

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// Query returns metrics from external prometheus url or an error
+// ex: query := `ALERTS{alertstate="firing",alertname="Watchdog",severity="none"} == 1`
+// ex: query := `scheduler_scheduling_duration_seconds_sum`
+func (c *PrometheusClient) Query(query string) (model.Vector, error) {
+	var metrics model.Vector
+
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: true,
```
mfojtik: cheating! what would it take to make this secure?

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+
+	req.Header.Add("Authorization", "Bearer "+c.token)
+
+	resp, err := prometheusClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("prometheus query failed: %v\n%v", err, resp)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Non-200 response from prometheus: %v", resp)
```
mfojtik: nit: lowercase error message

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	result := prometheusResponse{}
+	json.Unmarshal(body, &result)
```
mfojtik: check for unmarshal error?

test/library/metrics/query.go
```diff
@@ -0,0 +1,107 @@
...
+	metrics = result.Data.Result
+	return metrics, nil
```
mfojtik: just return result.Data.Result

test/library/metrics/query.go
sallyom: I had this working in an earlier iteration, we'd need to include the CA (create configmap w/ annotation for service-ca-operator to create cert), then use the internal service url.  I still have that code somewhere locally - was steered in this direction following how cluster-monitoring folks configure their tests.

marun: @mfojtik Given that the monitoring team is also skipping verification in this test-only scenario, what is the value in other teams not doing the same? It's not like the security of the route's endpoints won't be validated by things like the console.

marun: @sallyom If verification is required, I think it would be preferable to verify the externally-exposed route rather than jumping through the hoops required to use the internal service url from tests running outside the cluster.

sanchezl:
@sallyom There is a [Prometheus client](https://github.com/openshift/library-go/blob/9821002633e882489381e51d1bb4442788c0c01f/go.mod#L42) available. I used it for https://github.com/openshift/origin/pull/24291.

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: done, w/ route.Status.Ingress[0].Host

test/library/metrics/query.go
sallyom: added

test/library/metrics/query.go
sallyom: _still working on this_

sallyom: I exported both struct and fields - I'm confused about returning the interface, using private struct.  In Query fn, we need the Host+Token so shouldn't the struct + fields be exported?  Maybe it'll make more sense if i stare at it a bit more.

sallyom:
> @sallyom There is a [Prometheus client](https://github.com/openshift/library-go/blob/9821002633e882489381e51d1bb4442788c0c01f/go.mod#L42) available. I used it for [openshift/origin#24291](https://github.com/openshift/origin/pull/24291).

oh (I know there was some reason I wasn't using that but cannot remember now)! thanks, I'll incorporate that

test/library/metrics/query.go
sallyom: now that I'm using prometheus/client_golang this is resolved.

test/library/metrics/query.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	host := route.Status.Ingress[0].Host
+	var bearerToken string
+	secrets, err := kclient.CoreV1().Secrets("openshift-monitoring").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
+	}
+	for _, secret := range secrets.Items {
+		if secret.Type != corev1.SecretTypeServiceAccountToken ||
+			!strings.HasPrefix(secret.Name, "prometheus-") {
```
marun: Would it make sense to specify the more explicit `prometheus-k8s-` prefix? Tokens sourced from secrets prefixed with both `prometheus-k8s-` and `prometheus-operator-` support querying today, but I'm not sure that will always be true.

marun:
@sallyom Nice work! I've updated https://github.com/openshift/service-ca-operator/pull/90 to use it and tests are passing locally.  Once this PR merges I'll add another helper that checks that metrics are being collected for a given namespace (e.g. `checkMetricsCollected(t, config, namespace)`).

test/library/metrics/query.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// NewPrometheusClient returns Prometheus API or error
+func NewPrometheusClient(kclient *kubernetes.Clientset, rc *routeclient.Clientset) (prometheusv1.API, error) {
+	_, err := kclient.CoreV1().Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	route, err := rc.RouteV1().Routes("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
```
marun: Not something to block merge since the code works as-is., but I seem to recall @s-urbaniak mentioning that it would be preferable to query `thanos-querier`. However, my naive attempts to use the `thanos-querier` route were unsuccessful with tokens sourced from `prometheus-k8s-` and `thanos-querier-` prefixed secrets.

s-urbaniak: Indeed, the `prometheus-k8s` is route is still maintained, but effectively deprecated. Thanos Querier is gated by the same oauth-proxy sidecar as `prometheus-k8s`, hence the same token secrets should apply, else we have a bug on our side.

marun: @s-urbaniak No bug. I was requesting targets via the `thanos-querier` route and it wasn't succeeding, but I realize now that it was a 404 rather than an auth failure presumably because thanos doesn't support querying targets.  Performing an `up` query instead works fine.

@sallyom Please update to use the `thanos-querier` route.

test/library/metrics/query.go
```diff
@@ -0,0 +1,88 @@
...
+	var host string
+	switch queryType {
+	case "query":
+		route, err := rc.RouteV1().Routes("openshift-monitoring").Get("thanos-querier", metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		host = route.Status.Ingress[0].Host
+	case "alert":
+		route, err := rc.RouteV1().Routes("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
```
marun: @s-urbaniak If this route is deprecated, how should tests be querying for alerts?

test/library/metrics/query.go
sallyom: I hit a snag, the thanos-querier route works for a query, but when searching for an alert, that route does not work, have to use the prometheus-k8s route for alerts (like I'm doing in the [kube-controller-manager-operator PR](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/316)).  I added a switch and a param to the NewPrometheusClient to pass either 'query' or 'alert'.  
@s-urbaniak is that expected? thanos-querier route not for Alerts? I get a 404 for /alerts

test/library/metrics/query.go
sallyom: done

test/library/metrics/query.go
sallyom: ^^ lol I found that out also, see above

test/library/metrics/query.go
sallyom: I found that you can pass the _entire alert_ as a query, like so:
`ALERTS{alertname="PodDisruptionBudgetAtLimit",alertstate="pending",namespace="pdbns",poddisruptionbudget="pdbname",prometheus="openshift-monitoring/k8s",service="kube-state-metrics",severity="warning"}==1`
instead of using the deprecated prometheus-k8s route.  From what I can find, you have to pass the full alert, since partial responses return an error w/ how we have thanos setup - that is stretching my knowledge of prometheus/thanos, correct me if you can.

soltysh: @sallyom do you need to pass all of the properties in that query or only those you care about? Seems quite cumbersome at first.

marun:
LGTM

test/library/metrics/query.go
s-urbaniak: @sallyom indeed, thanos querier does not support `/alerts` as it is "only" responsible for the query part. It doesn't scrape targets either as it is "just" a proxy in front of prometheus, hence `/targets` doesn't exist here either. Your `ALERTS` query is fine from my point of view, but passing just `ALERTS{alertname="PodDisruptionBudgetAtLimit"}` should work too.

test/library/metrics/query.go
s-urbaniak: as mentioned above you can use the `ALERTS` query to get the desired result. Indeed, the `/alerts` endpoint is not accessible via Thanos Querier, as it is "just" responsible for querying metrics.

test/library/metrics/query.go
```diff
@@ -0,0 +1,78 @@
...
+		RoundTripper: transport.NewBearerAuthRoundTripper(
+			bearerToken,
+			&http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
+				TLSHandshakeTimeout: 10 * time.Second,
+				TLSClientConfig:     &tls.Config{InsecureSkipVerify: true},
```
deads2k: You cannot send a bearer token to a destination without verifying the server you're speaking to.

marun: What is the potential risk(s) you are concerned with? As per the `test` prefix of this path, this code is intended only to support e2e testing of metrics collection. Also, the bearer token will be only ever be sent to the endpoints backing the `thanos-querier` route owned by the monitoring operator.

sallyom:
@deads2k ptal, thanks!  and thanks for pointing me in the direction of the correct router-ca configmap, that did the trick, I was initially using the service-ca-operator generated CA, not the correct router-ca

PR #609, opened 2019-11-26
Verify outbound manifests now that schema1 support is dropped

smarterclayton:
We can verify that the returned digest matches our expectation cheaply.
May work with schema1, but more focused on schema2 for now.

Testing in openshift/oc#186

### DIFF
--- a/pkg/image/registryclient/client_test.go
+++ b/pkg/image/registryclient/client_test.go
@@ -547,6 +547,64 @@ func Test_verifyManifest_Get(t *testing.T) {
 	}
 }
 
+func Test_verifyManifest_Put(t *testing.T) {
+	tests := []struct {
+		name     string
+		dgst     digest.Digest
+		err      error
+		manifest distribution.Manifest
+		options  []distribution.ManifestServiceOption
+		want     digest.Digest
+		wantErr  string
+	}{
+		{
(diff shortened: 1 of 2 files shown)


pkg/image/registryclient/client_test.go
```diff
@@ -547,6 +547,62 @@ func Test_verifyManifest_Get(t *testing.T) {
...
 
+func Test_verifyManifest_Put(t *testing.T) {
+	tests := []struct {
+		name     string
+		dgst     digest.Digest
+		err      error
+		manifest distribution.Manifest
+		options  []distribution.ManifestServiceOption
+		want     digest.Digest
+		wantErr  bool
```
wking: nit: instead of a "was there an error?" boolean, I'd rather have an `expectedError` string to be used as a regexp or exact match for the returned error.  For example, like [this][1].  Could certainly be punted to follow-up work though.

[1]: https://github.com/openshift/installer/blob/af568556272c7f1d4ce8757a73f3b18c8d7b0c8a/pkg/types/validation/installconfig_test.go#L717-L721

smarterclayton: Updated

smarterclayton:
Tag please (if it looks good)

PR #612, opened 2019-11-27
encryption: enrich current state with secret info to avoid spurious events

sttts:
The current config state was not enriched with the information we store in secret annotations (reasons, migration state). Hence, every comparison of that info between current and desired state was wrong, leading to more events and possibly other unwanted side-effects. This PR also updates the current state with the secrets, like we did with the desired state.

### DIFF
--- a/pkg/operator/encryption/controllers/migration_controller.go
+++ b/pkg/operator/encryption/controllers/migration_controller.go
@@ -56,6 +56,8 @@ const (
 //   encryption.apiserver.operator.openshift.io/migrated-resources annotations on the
 //   current write-key secrets.
 type migrationController struct {
+	component string
+
 	operatorClient operatorv1helpers.OperatorClient
 
 	queue         workqueue.RateLimitingInterface
@@ -74,6 +76,7 @@ type migrationController struct {
 }
 
 func NewMigrationController(
+	component string,
 	deployer statemachine.Deployer,
 	migrator migrators.Migrator,
 	operatorClient operatorv1helpers.OperatorClient,
@@ -84,6 +87,7 @@ func NewMigrationController(
 	encryptedGRs []schema.GroupResource,
 ) *migrationController {
 	c := &migrationController{
+		component:      component,
 		operatorClient: operatorClient,
 
 		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "EncryptionMigrationController"),
@@ -167,7 +171,11 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
 		return nil, nil
 	}
 
-	currentState := encryptionconfig.ToEncryptionState(currentEncryptionConfig)
+	encryptionSecrets, err := secrets.ListKeySecrets(c.secretClient, c.encryptionSecretSelector)
+	if err != nil {
+		return nil, err
+	}
+	currentState, _ := encryptionconfig.ToEncryptionState(currentEncryptionConfig, encryptionSecrets)
 	desiredEncryptedConfig := encryptionconfig.FromEncryptionState(desiredEncryptionState)
 
 	// no storage migration until config is stable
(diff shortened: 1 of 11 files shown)


pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -92,6 +106,14 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 				continue // should never happen
 			}
 
+			// enrich KeyState with values from secrets
+			for _, k := range backedKeys {
+				if state.EqualKeyAndEqualID(&ks, &k) {
+					ks = k
```
p0lyn0mial: okay, that's the spot where we add more info to the keys.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -122,3 +123,16 @@ func (m *MigratedGroupResources) HasResource(resource schema.GroupResource) bool
 	}
 	return false
 }
+
+// ListKeySecrets returns the current key secrets from openshift-config-managed.
+func ListKeySecrets(secretClient corev1client.SecretsGetter, encryptionSecretSelector metav1.ListOptions) ([]*corev1.Secret, error) {
+	encryptionSecretList, err := secretClient.Secrets("openshift-config-managed").List(encryptionSecretSelector)
```
p0lyn0mial: okay, this was moved from `GetEncryptionConfigAndState` method

pkg/operator/encryption/statemachine/transition.go
```diff
@@ -76,21 +72,10 @@ func GetEncryptionConfigAndState(
 // 3. if (2) is the case, the write-key must be the most recent key.
 // 4. if (2) and (3) are the case, all non-write keys should be removed.
 func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.EncryptionConfiguration, encryptionSecrets []*corev1.Secret, toBeEncryptedGRs []schema.GroupResource) map[schema.GroupResource]state.GroupResourceState {
-	backedKeys := make([]state.KeyState, 0, len(encryptionSecrets))
```
p0lyn0mial: okay, we will get `backedKeys` from `ToEncryptionState ` now.

pkg/operator/encryption/statemachine/transition.go
```diff
@@ -114,27 +99,6 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
 		return desiredEncryptionState
 	}
 
-	// enrich KeyState with values from secrets
```
p0lyn0mial: ok, not needed as it will be done in `ToEncryptionState`

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -167,7 +171,11 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
 		return nil, nil
 	}
 
-	currentState := encryptionconfig.ToEncryptionState(currentEncryptionConfig)
+	encryptionSecrets, err := secrets.ListKeySecrets(c.secretClient, c.encryptionSecretSelector)
+	if err != nil {
+		return nil, err
+	}
+	currentState, _ := encryptionconfig.ToEncryptionState(currentEncryptionConfig, encryptionSecrets)
```
p0lyn0mial: okay, now `currentState` will have additional data from the secrets.

pkg/operator/encryption/encryptionconfig/config.go
sttts: yep. Everything else is basically plumbing update. Yes, I should have made two commits.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -205,37 +213,19 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable() (migratingR
 			continue // no write key to migrate to
 		}
 
-		writeSecret, err := findSecretForKeyWithClient(grActualKeys.WriteKey, c.secretClient, c.encryptionSecretSelector)
```
p0lyn0mial: okay, this bit is not needed as it was essentially  calling `ToKeyState`

PR #615, opened 2019-11-28
encryption: fix freezing key controller

sttts:
The key controller was still just expecting one read-key, while with the recent change for backup/restore we always preserve one more key. This made the key controller freeze after the first two read-keys were in the config.

While fixing the upper, we also run into the case that during pruning this one more read-key is preserved, but this read-key must be backed. Hence, we have to preserve also all other unbacked read-keys in between.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller_test.go
+++ b/pkg/operator/encryption/controllers/key_controller_test.go
@@ -1,6 +1,7 @@
 package controllers
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"testing"
@@ -12,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/diff"
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/config/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	clientgotesting "k8s.io/client-go/testing"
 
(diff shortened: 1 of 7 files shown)


pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -210,6 +212,48 @@ func TestKeyController(t *testing.T) {
 			},
 		},
 
+		{
+			name: "create a new write key when the previous key expired and another read key exists",
```
p0lyn0mial: shouldn't we validate if a new key was created? (`validateFunc` function)

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -210,6 +212,48 @@ func TestKeyController(t *testing.T) {
...
+					return ecs
+				}(),
+			},
+			apiServerObjects: apiServerAesCBC,
+			targetNamespace:  "kms",
+			expectedActions: []string{
+				"list:pods:kms",
+				"get:secrets:kms",
+				"list:secrets:openshift-config-managed",
+				"create:secrets:openshift-config-managed",
```
p0lyn0mial: this only says that something was created ;)

pkg/operator/encryption/state/helpers.go
```diff
@@ -34,11 +34,15 @@ func MigratedFor(grs []schema.GroupResource, km KeyState) (ok bool, missing []sc
...
 }
 
-// KeysWithPotentiallyPersistedData returns the minimal, recent secrets which have migrated all given GRs.
-func KeysWithPotentiallyPersistedData(grs []schema.GroupResource, recentFirstSortedKeys []KeyState) []KeyState {
+// KeysWithPotentiallyPersistedDataAndNextReadKey returns the minimal, recent secrets which have migrated all given GRs.
+func KeysWithPotentiallyPersistedDataAndNextReadKey(grs []schema.GroupResource, recentFirstSortedKeys []KeyState) []KeyState {
 	for i, k := range recentFirstSortedKeys {
 		if allMigrated, missing, _ := MigratedFor(grs, k); allMigrated {
-			return recentFirstSortedKeys[:i+1]
+			if i+1 < len(recentFirstSortedKeys) {
```
p0lyn0mial: this looks to be safe, we could have a unit test for it.

pkg/operator/encryption/controllers/key_controller_test.go
sttts: that's a general issue in this test, isn't it? But what else do we create in the key controller than keys? :)

pkg/operator/encryption/state/helpers.go
sttts: what do you want to test?

sttts: we have tests that "one more read-key" is fine. So this is covered. The rest here is just Golang standard behaviour.

p0lyn0mial: sgtm.

pkg/operator/encryption/controllers/key_controller_test.go
p0lyn0mial: yeah, hard to beat that argument :)

PR #616, opened 2019-11-28
staticpod: switch controllers to use context instead of stopCh

mfojtik:


pkg/operator/staticpod/controllers.go
```diff
@@ -157,11 +165,13 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (*staticPodOperator
 			configMapClient,
 			secretClient,
 			eventRecorder,
-		)
+		))
+	} else {
+		errs = append(errs, fmt.Errorf("missing revisionController; cannot proceed"))
```
sttts: why no hard return err?

mfojtik: the original idea was to collect all errors I think, that is pre-existing.

sttts: ic. The controller names are not necessary anymore.

sttts:
/lgtm
/approve

Happy bumping!

PR #618, opened 2019-11-28
inprocess migrator metrics

p0lyn0mial:
Adds two simple metrics to the storage migrator. Since the same metrics will be used by `kas-o` and `oas-o` I think it's fine to have them down in `library-go`

1. `migrated_objects` - represents the total number of objects that have been migrated, labeled with the full resource name.
2. `migrations` - represents the number of completed migration, labeled with the full resource name, and the status of the migration (failed or succeeded)

note that the names and types align with those defined in the upstream migrator.

### DIFF
--- a/pkg/operator/encryption/controllers/migrators/metrics.go
+++ b/pkg/operator/encryption/controllers/migrators/metrics.go
@@ -1,0 +1,93 @@
+package migrators
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+const (
+	namespace = "storage_migrator"
+	subsystem = "core_migrator"
+)
+
+// metrics provides access to all core migrator metrics.
+var metrics *migratorMetrics
+
+func init() {
+	metrics = newMigratorMetrics(legacyregistry.Register)
+}
+
+// migratorMetrics instruments core migrator with prometheus metrics.
+type migratorMetrics struct {
+	objectsMigrated   *k8smetrics.CounterVec
+	migration         *k8smetrics.CounterVec
+	migrationDuration *k8smetrics.HistogramVec
+}
+
+// newMigratorMetrics create a new MigratorMetrics, configured with default metric names.
+func newMigratorMetrics(registerFunc func(k8smetrics.Registerable) error) *migratorMetrics {
+	// objectMigrates is defined in kube-storave-version-migrator
+	objectsMigrated := k8smetrics.NewCounterVec(
+		&k8smetrics.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migrated_objects",
+			Help:      "The total number of objects that have been migrated, labeled with the full resource name",
+		}, []string{"resource"})
+	registerFunc(objectsMigrated)
+
+	// migration is defined in kube-storave-version-migrator
+	migration := k8smetrics.NewCounterVec(
+		&k8smetrics.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migrations",
+			Help:      "The total number of completed migration, labeled with the full resource name, and the status of the migration (failed or succeeded)",
+		}, []string{"resource", "status"})
+	registerFunc(migration)
+
+	// migrationDuration is not defined upstream but uses the same Namespace and Subsystem
+	// as the other metrics that are defined in kube-storave-version-migrator
+	migrationDuration := k8smetrics.NewHistogramVec(
+		&k8smetrics.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migration_duration_seconds",
+			Help:      "How long a successful migration takes in seconds, labeled with the full resource name",
+			Buckets:   prometheus.ExponentialBuckets(120, 2, 7),
+		}, []string{"resource"})
+	registerFunc(migrationDuration)
+
+	return &migratorMetrics{
+		objectsMigrated:   objectsMigrated,
+		migration:         migration,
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/controllers/migrators/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+	migration       *k8smetrics.CounterVec
+}
+
+// newMigratorMetrics create a new MigratorMetrics, configured with default metric names.
+func newMigratorMetrics() *migratorMetrics {
+	objectsMigrated := k8smetrics.NewCounterVec(
+		&k8smetrics.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migrated_objects",
```
p0lyn0mial: I think that this should have `_total` suffix (we could try to upstream that change)

pkg/operator/encryption/controllers/migrators/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+			Help:      "The number of objects that have been migrated, labeled with the full resource name.",
+		}, []string{"resource"})
+	legacyregistry.Register(objectsMigrated)
+
+	migration := k8smetrics.NewCounterVec(
+		&k8smetrics.CounterOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migrations",
```
p0lyn0mial: same here (suffix)

p0lyn0mial:
I tested collecting `migrated_objects` in two modes. The first one (`mode1`) collected samples in batches (`len(allResource.Items)`), the second one (`mode2`) collected samples for each object separately (increase by "one").

Metrics were collected during e2e testing (https://prow.svc.ci.openshift.org/pr-history/?org=openshift&repo=cluster-kube-apiserver-operator&pr=676)

The following are queries I run in the expression browser - https://prometheus.io/docs/prometheus/latest/getting_started/#using-the-expression-browser (note that you need to adjust the time)

```
# query1
# total number of migrations, based on that I can tell that migration #7 started at 16:17:19 and ended at 16:23:04
storage_migrator_core_migrator_migrations

# query2
# gives the total number of migrated objects storage_migrator_core_migrator_migrated_objects

# query3
# gives the total number of object migrated as part of migration #7
storage_migrator_core_migrator_migrated_objects - storage_migrator_core_migrator_migrated_objects offset 6m

# query4
# gives QPS (Object per second) for migration #7
rate(storage_migrator_core_migrator_migrated_objects[6m])
```

p0lyn0mial:
Here are the results for `mode1`:

#query1 (total number of migrations)
query1_batch

#query1 graph (total number of migrations)
query1_graph_batch

#query2 (total number of migrated objects)
query2_batch

#query2 graph (total number of migrated objects)
query2_graph_batch

#query3 (the total number of object migrated as part of migration X)
query3_batch

#query4 ( QPS (Object per second) for migration X)
query4_batch

p0lyn0mial:
Here are the results for `mode2`:

#query1 (total number of migrations)
query1


#query1 graph (total number of migrations)
query1_graph


#query2 (total number of migrated objects)
query2


#query2 graph (total number of migrated objects)
query2_graph


#query3 (the total number of object migrated as part of migration X)
query3


#query4 ( QPS (Object per second) for migration X)
query4

p0lyn0mial:
based on the queries I run I think that those metrics are a good starting point and can answer the basic question we have.

it looks like the two modes lead to the same numbers, `mode1` (collect in batches) is more pleasant for CPU than `mode2`. Not sure what happens when we miss some metrics (network error) `the whole batch` vs `a few samples`.

sttts:
storage_migration_objects is better

pkg/operator/encryption/controllers/migrators/metrics.go
```diff
@@ -0,0 +1,70 @@
+package migrators
+
+import (
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+const (
+	namespace = "storage_migrator"
+	subsystem = "core_migrator"
```
sttts: are these a common pattern?

pkg/operator/encryption/controllers/migrators/metrics.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// migratorMetrics instruments core migrator with prometheus metrics.
+type migratorMetrics struct {
+	objectsMigrated *k8smetrics.CounterVec
+	migration       *k8smetrics.CounterVec
+}
+
+// newMigratorMetrics create a new MigratorMetrics, configured with default metric names.
+func newMigratorMetrics() *migratorMetrics {
```
sttts: nit: make legacyregistry a parameter here

pkg/operator/encryption/controllers/migrators/metrics.go
p0lyn0mial: the upstream tool names it like this.

pkg/operator/encryption/controllers/migrators/metrics.go
```diff
@@ -14,38 +16,50 @@ const (
...
-	legacyregistry.Register(migration)
+	registerFunc(migration)
+
+	migrationDuration := k8smetrics.NewHistogramVec(
+		&k8smetrics.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "migration_duration_seconds",
+			Help:      "How long a successful migration takes in seconds, labeled with the full resource name",
+			Buckets:   prometheus.ExponentialBuckets(300, 2, 7),
```
sttts: actually with this being by resource, should we go down a bit with the minimum? maybe 2m, 4m, 8m, 16m, 32m, 1h, 2h ?

p0lyn0mial: yes, it makes sense, updated.

PR #619, opened 2019-11-28
encryption: retry requests during e2e when kube-apiserver restarts

sttts:


test/library/encryption/helpers.go
```diff
@@ -227,7 +233,7 @@ func ForceKeyRotation(t testing.TB, updateUnsupportedConfig UpdateUnsupportedCon
 		return err
 	}
 
-	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	return onErrorWithTimeout(wait.ForeverTestTimeout, retry.DefaultBackoff, orError(errors.IsConflict, transientAPIError), func() error {
```
p0lyn0mial: do we need `onError` since `IsConflict` is already considered `transientError` ?

p0lyn0mial:
potentially we have many places that would like to use `onErrorWithTimeout` function not only `GetLastKeyMeta`.

test/library/encryption/helpers.go
sttts: IsConflict is not transient

PR #622, opened 2019-11-29
Bug 1776797: encryption: speed-up, Encrypted condition, allow backup/restore

sttts:
This PR syncs 4.4 work with 4.3 mainly doing:

- speeding up migration by a factor: 50k objects in half an hour (before around 1k in 5m)
- added Encrypted condition for the user to see the encryption status
- leave last read-key in encryption config in order to allow backup/restore when encryption-config backup is done after backup, without losing data (compare https://github.com/openshift/enhancements/pull/131).


sttts:
/hold

until smoke tested in ckao.

PR #628, opened 2019-12-04
alha-build-machinery: Force vendoring mode

tnozicka:
When we enable 1.13 it will turn go modules on by default and the
current go implementation ignodes vendor folder when present unless
vendoring mode is specifically requested. While that is being fixed
upstream, we need to force the -mod=vendor flag so we use vendored deps.

Requires:
 - [ ] https://github.com/openshift/release/pull/6226 (so the we validate we don't hit issues like https://github.com/golang/go/issues/35955)

`go vet` will have enforce the vendoring mod by the global env in CI, but because of https://github.com/golang/go/issues/35955 we can't set it explicitly for developers yet.

/cc @sttts 

@openshift/sig-master


tnozicka:
/hold
for https://github.com/openshift/release/pull/6226

tnozicka:
needs 1.13 bump first

tnozicka:
also fixed other go1.13 bugs to get tests green

PR #631, opened 2019-12-06
Switch to go modules and bump k8s to 1.17

vareti:
Giving a try to switch dependency management tool to go modules and bump k8s to 1.17


pkg/operator/configobserver/featuregates/observe_featuregates_test.go
```diff
@@ -47,23 +47,20 @@ func TestObserveFeatureFlags(t *testing.T) {
 			name:        "default",
 			configValue: configv1.Default,
 			expectedResult: []string{
-				"ExperimentalCriticalPodAnnotation=true",
```
p0lyn0mial: has been those features removed?

I checked https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ but it's not clear for me.

p0lyn0mial:
`go.mod` looks good to me, you need to updated generated data to fix the `verify` step. 

`verify-deps` failed to find glide.yaml file in directory tree.

pkg/operator/configobserver/featuregates/observe_featuregates_test.go
p0lyn0mial: @sttts is this okay?

vareti: I found out from this file list of enabled/disabled feature gates

https://github.com/openshift/api/blob/master/config/v1/types_feature.go#L101-L110

alpha-build-machinery/make/targets/openshift/deps.mk
```diff
@@ -8,28 +8,28 @@ update-deps:
 	$(scripts_dir)/$@.sh
 .PHONY: update-deps
 
-# $1 - temporary directory to restore vendor dependencies from glide.lock
```
sttts: @tnozicka ptal

alpha-build-machinery/scripts/update-deps.sh
```diff
@@ -1,27 +1,23 @@
...
 
-verify_glide_version
+verify_go_version
 
-glide update --strip-vendor
-
-# glide doesn't handle mercurial properly and leaves internal files (equivalent of .git/) laying around
-# Given those files differ by mercurial version it was cloned with, verify-deps would break
-find ./vendor -name '.hg_archival.txt' -delete
+go mod vendor
```
tnozicka: first run `go mod tidy`

alpha-build-machinery/scripts/update-deps.sh
```diff
@@ -1,27 +1,23 @@
 #!/bin/bash -e
 
-readonly GLIDE_MINOR_VERSION="13"
-readonly REQUIRED_GLIDE_VERSION="0.$GLIDE_MINOR_VERSION"
+readonly GO_MINOR_VERSION=12
```
tnozicka: 13 as that's the first version with modules on by default, unless these commands start setting the GO111MODULE env var, right?

tnozicka: maybe we can just drop the check entirely since go version is mandated globally and go mod has compatibility guarantees

tnozicka:
If the indent of this PR is just to switch library-go to go.mod it is fine to disable the verify-deps check in the CI and removing https://github.com/openshift/library-go/blob/7d8455a/Makefile#L13 . We have a card to add a new verify-deps for go.mod and other repos already went the path of disabling the check temporarily until we have a replacement

pkg/security/ldapclient/interfaces.go
```diff
@@ -1,6 +1,6 @@
 package ldapclient
 
-import "gopkg.in/ldap.v2"
```
sttts: this is not part of a bump. A bump are just the vendor changes, and potentially the glide/go.mod changes. Move this into a fix-up commit.

PR #632, opened 2019-12-07
pkg/operator/status: Add configurable inertia

wking:
Instead of hard-coding a two-minute inertia, allow the controller-creator to set a custom inertia function.  The previous value was chosen to work fairly well for most cases, but there was contention between important conditions where 2m was pushing too long and less severe conditions where 2m was short enough to catch slow launch-time initialization that triggered distracting Degraded flaps (e.g. the 2m10s missing ConfigMaps in [rhbz#1775922][1], where the bulk of that 2m was taken by other kube-apiserver initialization actions).  With this commit, the kube-apiserver can say:

> 3m of missing ConfigMaps is ok, but only accept 1m of SomeOtherThing before going `Degraded=True`.

or whatever.

/assign @deads2k

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1775922

### DIFF
--- a/pkg/operator/status/inertia.go
+++ b/pkg/operator/status/inertia.go
@@ -1,0 +1,66 @@
+package status
+
+import (
+	"fmt"
+	"regexp"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// Inertia returns the inertial duration for the given condition.
+type Inertia func(condition operatorv1.OperatorCondition) time.Duration
+
+// InertiaCondition configures an inertia duration for a given set of
+// condition types.
+type InertiaCondition struct {
+	// ConditionTypeMatcher is a regular expression selecting condition types
+	// with which this InertiaCondition is associated.
+	ConditionTypeMatcher *regexp.Regexp
+
+	// Duration is the inertial duration for associated conditions.
+	Duration time.Duration
+}
+
+// InertiaConfig holds configuration for an Inertia implementation.
+type InertiaConfig struct {
+	defaultDuration time.Duration
+	conditions      []InertiaCondition
+}
+
+// NewInertia creates a new InertiaConfig object.  Conditions are
+// applied in the given order, so a condition type matching multiple
+// regular expressions will have the duration associated with the first
+// matching entry.
+func NewInertia(defaultDuration time.Duration, conditions ...InertiaCondition) (*InertiaConfig, error) {
+	for i, condition := range conditions {
+		if condition.ConditionTypeMatcher == nil {
+			return nil, fmt.Errorf("condition %d has a nil ConditionTypeMatcher", i)
+		}
+	}
+
+	return &InertiaConfig{
+		defaultDuration: defaultDuration,
+		conditions:      conditions,
+	}, nil
+}
+
+// MustNewInertia is like NewInertia but panics on error.
+func MustNewInertia(defaultDuration time.Duration, conditions ...InertiaCondition) *InertiaConfig {
+	inertia, err := NewInertia(defaultDuration, conditions...)
+	if err != nil {
+		panic(err)
+	}
+
+	return inertia
+}
+
+// Inertia returns the configured inertia for the given condition type.
+func (c *InertiaConfig) Inertia(condition operatorv1.OperatorCondition) time.Duration {
+	for _, matcher := range c.conditions {
+		if matcher.ConditionTypeMatcher.MatchString(condition.Type) {
+			return matcher.Duration
+		}
+	}
+	return c.defaultDuration
+}
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -49,9 +49,10 @@ type StatusSyncer struct {
 	clusterOperatorClient configv1client.ClusterOperatorsGetter
 	clusterOperatorLister configv1listers.ClusterOperatorLister
 
-	cachesToSync  []cache.InformerSynced
-	queue         workqueue.RateLimitingInterface
-	eventRecorder events.Recorder
+	cachesToSync    []cache.InformerSynced
+	queue           workqueue.RateLimitingInterface
+	eventRecorder   events.Recorder
+	degradedInertia Inertia
 }
 
 func NewClusterOperatorStatusController(
@@ -71,6 +72,7 @@ func NewClusterOperatorStatusController(
 		clusterOperatorLister: clusterOperatorInformer.Lister(),
 		operatorClient:        operatorClient,
 		eventRecorder:         recorder.WithComponentSuffix("status-controller"),
+		degradedInertia:       MustNewInertia(2 * time.Minute).Inertia,
 
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "StatusSyncer_"+strings.Replace(name, "-", "_", -1)),
 	}
@@ -82,6 +84,14 @@ func NewClusterOperatorStatusController(
 	c.cachesToSync = append(c.cachesToSync, clusterOperatorInformer.Informer().HasSynced)
 
 	return c
+}
+
+// WithDegradedInertia returns a copy of the StatusSyncer with the
+// requested inertia function for degraded conditions.
+func (c *StatusSyncer) WithDegradedInertia(inertia Inertia) *StatusSyncer {
+	output := *c
+	output.degradedInertia = inertia
+	return &output
 }
 
 // sync reacts to a change in prereqs by finding information that is required to match another value in the cluster. This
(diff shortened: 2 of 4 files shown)


wking:
[verify-alpha-build-machinery][1] failed with:

```
--- ./manifests/operator.openshift.io_myotheroperatorresources.crd.yaml
@@ -9,6 +9,40 @@ spec:
     kind: MyOtherOperatorResource
     plural: myotheroperatorresources
   scope: ""
+  validation:
+    openAPIV3Schema:
+      description: MyOtherOperatorResource is an example operator configuration type
...
+      type: object
   version: v1beta1
   versions:
   - name: v1beta1
make[2]: *** [verify-codegen-crds-manifests] Error 1
```

Seems unrelated and non-blocking.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_library-go/632/pull-ci-openshift-library-go-master-verify-alpha-build-machinery/1674

deads2k:
/unassign
/assign @sanchezl 

The idea seems reasonable.

pkg/operator/status/inertia.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Inertia returns the inertial duration for the given condition type.
+type Inertia func(conditionType string) time.Duration
+
+// InertiaCondition configures an inertia duration for a given set of
+// condition types.
+type InertiaCondition struct {
+	// ConditionType is a regular expression selecting condition types
+	// with which this InertiaCondition is associated.
+	ConditionType *regexp.Regexp
```
mfojtik: nit: `ConditionTypeMatcher` or something similar to highlight this is a regexp ?

wking: > nit: `ConditionTypeMatcher` or something similar to highlight this is a regexp ?

I think `Regexp`-ness is already clear from the property definition here, and Go will throw useful compiler errors for folks who think it's a string or whatever.  But I don't care all that much; rerolled to use your name with 1a1079fd5 -> 72d73fac0.

Also in that reroll, I've changed:

```go
type Inertia func(conditionType string) time.Duration
```

to:

```go
type Inertia func(conditionType string) time.Duration
```

in case folks want to write `Inertia` implementations that switch on the `Reason` or some such.

pkg/operator/status/inertia.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"fmt"
+	"regexp"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// Inertia returns the inertial duration for the given condition.
+type Inertia func(condition operatorv1.OperatorCondition) time.Duration
```
wking: Thoughts on pass-by-value here?  A pointer `condition` would be cheaper, and it seems unlikely that an operator would abuse it to mutate the passed condition.

sanchezl: Without analysis, there is no guarantee it would be cheaper. I would prefer the pass-by-value as it correlates with the expected mutability of the parameter.

wking:
> Can we provide a default inertial func for use by operators that pull in this change and are not ready to think about inertia on a granular level?

That's just `NewInertia(2*time.Minute)`.  Do you need me to wrap that up in something shorter?  Hard to imagine what that would be...

Hmm, actually, now that this is configurable, we probably want this for all types, not [just Degraded][1].  In that case, the default would be:

```go
NewIntertia(0, InertiaCondition{
  ConditionTypeMatcher: regexp.MustCompile("Degraded"),
  Duration:             2 * time.Minute,
})
```

Maybe that is worth a default?  Still seems easy enough to just declare this when you bump your library-go vendor...

[1]: https://github.com/openshift/library-go/blob/acc37612b9f9bd6d637b870661082489eb575aa5/pkg/operator/status/status_controller.go#L147

wking:
> Hmm, actually, now that this is configurable, we probably want this for all types...

Rerolled to do this with 72d73fac0 -> cb0967e89.

pkg/operator/status/inertia.go
```diff
@@ -0,0 +1,56 @@
...
+type InertiaConfig struct {
+	defaultDuration time.Duration
+	conditions      []InertiaCondition
+}
+
+// NewInertia creates a new InertiaConfig object.  Conditions are
+// applied in the given order, so a condition type matching multiple
+// regular expressions will have the duration associated with the first
+// matching entry.
+func NewInertia(defaultDuration time.Duration, conditions ...InertiaCondition) (*InertiaConfig, error) {
```
deads2k: make a `MustNewIntertia` that panics on the error to ease the coding.

pkg/operator/status/status_controller.go
```diff
@@ -144,10 +147,10 @@ func (c StatusSyncer) sync() error {
 	}
 
 	clusterOperatorObj.Status.RelatedObjects = c.relatedObjects
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionInertialCondition("Degraded", operatorv1.ConditionFalse, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Progressing", operatorv1.ConditionFalse, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Available", operatorv1.ConditionTrue, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Upgradeable", operatorv1.ConditionTrue, currentDetailedStatus.Conditions...))
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Degraded", operatorv1.ConditionFalse, c.inertia, currentDetailedStatus.Conditions...))
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Progressing", operatorv1.ConditionFalse, c.inertia, currentDetailedStatus.Conditions...))
```
deads2k: I prefer no intertia on these for now.  just pass nil please.

pkg/operator/status/status_controller.go
```diff
@@ -144,10 +147,10 @@ func (c StatusSyncer) sync() error {
...
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Available", operatorv1.ConditionTrue, c.inertia, currentDetailedStatus.Conditions...))
```
deads2k: I prefer no intertia on these for now.  just pass nil please.

pkg/operator/status/status_controller.go
```diff
@@ -144,10 +147,10 @@ func (c StatusSyncer) sync() error {
...
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, unionCondition("Upgradeable", operatorv1.ConditionTrue, c.inertia, currentDetailedStatus.Conditions...))
```
deads2k: I prefer no intertia on these for now.  just pass nil please.

pkg/operator/status/inertia.go
wking: > make a `MustNewIntertia` that panics on the error to ease the coding

Done with cb0967e89 -> 2d2aa9635, which also rebases onto the current master (no conflicts).

pkg/operator/status/status_controller.go
wking: > I prefer no intertia on these for now. just pass nil please.

Does library-go need to have an opinion here?  Seems like folks could set up their `Inertia` implementation to [restrict themselves to Degraded][1] if they wanted.

[1]: https://github.com/openshift/library-go/pull/632#issuecomment-563359300

wking: > Does library-go need to have an opinion here?

Talked this over out of band.  Issues considered:

* There's currently no use case for inertia in the other unioned cases.  Availability, progressing, and upgradeable should all switch over immediately.
* If someone thinks up a case, they'll need to argue it out in a library-go PR, where library-go maintainers can decide if the case makes sense or if the user is misunderstanding the union/type semantics.
* If the case is convincing, and library-go adds a new `WithAvailableInertia` (or whatever), existing `WithDegradedInertia` can bump to the new library-go without breaking.

So I've pushed 60ab2f004 -> 68a4d3905 leaving inertia out of `NewClusterOperatorStatusController`, adding a new, future-proof `WithDegradedInertia`, and setting `inertia` `nil` for these other types.

pkg/operator/status/status_controller.go
wking: > I prefer no intertia on these for now.

https://github.com/openshift/library-go/pull/632#discussion_r356273639

pkg/operator/status/status_controller.go
wking: > I prefer no intertia on these for now.

https://github.com/openshift/library-go/pull/632#discussion_r356273639

wking:
[unit][1] [logs][2]:

```
=== RUN   TestSyncSecret
--- FAIL: TestSyncSecret (0.11s)
    recorder_testing.go:33: Event: SecretCreated: Created Secret/foo -n operator because it was missing
    resourcesync_controller_test.go:159: expected exactly 1 delete call for this test, got 0
```

Dunno what that's about.

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_library-go/632/pull-ci-openshift-library-go-master-unit/1784
[2]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_library-go/632/pull-ci-openshift-library-go-master-unit/1784/artifacts/unit/junit_report.xml

PR #635, opened 2019-12-10
build: copy shared source-to-image helpers

mfojtik:
This is a clean move from source-to-image for packages that are shared between OpenShift build controllers, CLI (oc new-app) and source-to-image.

Since vendoring source-to-image already causing trouble for `oc` and we don't want to vendor source-to-image docker and other transitive (buildkit, etc.) this establish minimal common ground and set `oc` repo free from source-to-image.

`oc` import list: https://gist.github.com/mfojtik/2df592fe3e3abd61af54e458a3fffd9a

### DIFF
--- a/pkg/build/OWNERS
+++ b/pkg/build/OWNERS
@@ -1,0 +1,14 @@
+reviewers:
+  - bparees
+  - smarterclayton
+  - mfojtik
+  - soltysh
+  - csrwng
+  - gabemontero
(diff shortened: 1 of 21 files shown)


soltysh:
/hold
I don't think we want to make that mistake.

mfojtik:
/close

we found https://github.com/openshift/oc/pull/206 a better solution

PR #638, opened 2019-12-10
bug 1781804: fix revision pending message

deads2k:
I needed a correct message to figure out what state the operator thinks it is in

/assign @sttts

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -307,8 +307,7 @@ func (c *InstallerController) manageInstallationPods(operatorSpec *operatorv1.St
 				return true, err
 			}
 
-			pendingNewRevision := operatorStatus.LatestAvailableRevision > currNodeState.TargetRevision
-			newCurrNodeState, installerPodFailed, reason, err := c.newNodeStateForInstallInProgress(currNodeState, pendingNewRevision)
+			newCurrNodeState, installerPodFailed, reason, err := c.newNodeStateForInstallInProgress(currNodeState, operatorStatus.LatestAvailableRevision)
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

this fixes an immediate problem, but the entire approach looks wrong.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -537,11 +536,11 @@ func (c *InstallerController) newNodeStateForInstallInProgress(currNodeState *op
 
 	switch installerPod.Status.Phase {
 	case corev1.PodSucceeded:
-		if newRevisionPending {
+		if pendingNewRevision := latestRevisionAvailable > currNodeState.TargetRevision; pendingNewRevision {
 			// stop early, don't wait for ready static pod because a new revision is waiting
 			ret.LastFailedRevision = currNodeState.TargetRevision
 			ret.TargetRevision = 0
-			ret.LastFailedRevisionErrors = []string{fmt.Sprintf("static pod of revision has been installed, but is not ready while new revision %d is pending", currNodeState.TargetRevision)}
+			ret.LastFailedRevisionErrors = []string{fmt.Sprintf("static pod of revision %d has been installed, but is not ready while new revision %d is pending", currNodeState.TargetRevision, latestRevisionAvailable)}
```
sttts: oops.

deads2k:
/hold cancel

the approach is weird (we aren't degraded for this reason), but we are running

pkg/operator/staticpod/controller/installer/installer_controller.go
mfojtik: i thought go vet catch these (I fixed some in past)

PR #647, opened 2019-12-16
controllercmd: pass context via func argument

mfojtik:
Passing context.Context via struct field is a Go anti-pattern. We need to pass the context directly via function arguments instead.

After this PR land, the bumps in operator will have to fix the way they pass the context.


mfojtik:
/override ci/prow/verify-alpha-build-machinery

This is unrelated to this change and @tnozicka is looking into the failure

PR #649, opened 2019-12-16
bug 1784052: delete secrets if immutable fields don't match

deads2k:
secret types are immutable, so we need to delete and recreate on certain kinds of errors.

/hold

hold for proof.


deads2k:
/assign @mfojtik @tnozicka 

proof with test passed https://github.com/openshift/cluster-kube-apiserver-operator/pull/697

pkg/operator/resource/resourceapply/core.go
```diff
@@ -252,8 +252,24 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 		klog.Infof("Secret %s/%s changes: %v", required.Namespace, required.Name, JSONPatchSecretNoError(existing, existingCopy))
 	}
 	actual, err := client.Secrets(required.Namespace).Update(existingCopy)
-
 	reportUpdateEvent(recorder, existingCopy, err)
+
+	if err == nil {
+		return actual, true, err
+	}
+	if !strings.Contains(err.Error(), "field is immutable") {
```
mfojtik: nit: can you move this up? `if err == nil || !string....`

PR #654, opened 2020-01-03
fileobserver: prometheus counter for observed actions

mfojtik:
This add `fileobserver_action_count{name=ACTION, filename="file"}=N` prometheus metric that can be used to track changes on particular files over time and indicate problems when files are changed too often or not changed at all.


mfojtik:
(fixed build error, re-applying lgtm)

PR #655, opened 2020-01-03
leaderelection: use proto client

mfojtik:
Use protobuf client for leader election. Should decrease the CPU usage when querying for the configmap (but it really is just cosmetic).


mfojtik:
cc @deads2k or @sttts for approval

deads2k:
small potatoes, but ok

/approve

PR #656, opened 2020-01-03
leaderelection: set release on cancel to true

mfojtik:
This will cause faster operator upgrades as we release the leader when the context that we pass to all control loops in operators is cancelled (typically on graceful shutdown).


deads2k:
The new race to consider is the time between the release and the process exit. This should be small and all users of leader election need to handle cases of control loops taking time to shut down.  Similar to the original 3.x rules, an accidental active/active scenario should result in update conflict, backoff, status detection, skip.  The end result is a lot of stormy, conflicty work, but no wrong actions taken. 

/lgtm

PR #657, opened 2020-01-06
config: default bind network to tcp instead of tcp4

mfojtik:
Based on https://github.com/openshift/cluster-authentication-operator/pull/231 we should default to `tcp` instead of forcing `tcp4`


mfojtik:
/cc @russellb 

We use this in operators to serve metrics and healthz. The https://bugzilla.redhat.com/show_bug.cgi?id=1788140 and 5 other BZ's suggests that this is broken in ipv6 clusters.

mfojtik:
/retitle Bug 1788521: config: default bind network to tcp instead of tcp4

PR #667, opened 2020-01-16
Add controllerset for controlling an apiserver deployment

stlaz:
Add controller for NS finalization of a running API server deployment, controller handling an API server deployment's APIServices, and a controllerset which bundles these controllers + some others that do the API server housekeeping


deads2k:
straigth move?  if so, it can be lgtmd

/approve

pkg/operator/nsfinalizercontroller/finalizer_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	queue workqueue.RateLimitingInterface
+}
+
+// NewFinalizerController is here because
+// When running an aggregated API on the platform, you delete the namespace hosting the aggregated API. Doing that the
+// namespace controller starts by doing complete discovery and then deleting all objects, but pods have a grace period,
+// so it deletes the rest and requeues. The ns controller starts again and does a complete discovery and.... fails. The
+// failure means it refuses to complete the cleanup. Now, we don't actually want to delete the resoruces from our
+// aggregated API, only the server plus config if we remove the apiservices to unstick it, GC will start cleaning
+// everything. For now, we can unbork 4.0, but clearing the finalizer after the pod and daemonset we created are gone.
```
sttts: with so much apiserver specifics, why not put everything into `pkg/operator/apiserver/controller/namespacefinalizer`, and the others in similar places?

pkg/operator/apiservicecontroller/apiservice_controller.go
```diff
@@ -0,0 +1,234 @@
+package apiservicecontroller
```
sttts: pkg/operator/apiserver/controllers/apiserver

pkg/operator/apiservercontrollerset/apiservercontrollerset.go
```diff
@@ -0,0 +1,199 @@
+package apiservercontrollerset
```
sttts: pkg/operator/apiserver/controllerset

stlaz:
@sttts thanks, that makes sense.

@deads2k it is a clean move (with the repackaging now) + a fix to use controllers directly in the slice https://github.com/openshift/library-go/pull/667/files#diff-0c281fce8ed3b258f8b6fb4cec7b5ea7R196-R198 in case the goroutine triggered after the loop variable changed to point to the next controller.

stlaz:
/hold
I can see there were changes to the code in COA-o

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	operatorClient          v1helpers.OperatorClient
+	kubeClient              kubernetes.Interface
+	apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface
+	eventRecorder           events.Recorder
+
+	// queue only ever has one item, but it has nice error handling backoff/retry semantics
+	queue workqueue.RateLimitingInterface
+}
+
+func NewAPIServiceController(
```
p0lyn0mial: this has slightly changed in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/294

worth picking as it made it more generic.

stlaz:
/hold cancel
this should now include latest changes done in COA-o

PR #669, opened 2020-01-16
brings Kubernetes 1.17.1

p0lyn0mial:


mfojtik:
https://github.com/openshift/library-go/pull/670 does it together with openshift/api and client-go

PR #675, opened 2020-01-20
add cached client capabilities to stay out of qps jail

deads2k:
Alright, I think this is the last one in the chain.  We have

1. static resource management for the most common types
2. automatic informers get wired into the controller and the cache waiting
3. automatic status reporting for operators with aggregated status.
4. allow static resource management of CRDs
5. update CRD handling to support v1
6. add the ability to use cached clients based on the namespaced informers.

/assign @mfojtik

### DIFF
--- a/pkg/operator/resource/resourceapply/generic.go
+++ b/pkg/operator/resource/resourceapply/generic.go
@@ -3,6 +3,10 @@ package resourceapply
 import (
 	"fmt"
 
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
 	"github.com/openshift/api"
 	"github.com/openshift/library-go/pkg/operator/events"
 	corev1 "k8s.io/api/core/v1"
@@ -37,14 +41,19 @@ type ApplyResult struct {
 	Error   error
 }
 
-// TODO provide ways to provide cached getters
 type ClientHolder struct {
 	kubeClient          kubernetes.Interface
 	apiExtensionsClient apiextensionsclient.Interface
+	kubeInformers       v1helpers.KubeInformersForNamespaces
 }
 
 func (c *ClientHolder) WithKubernetes(client kubernetes.Interface) *ClientHolder {
 	c.kubeClient = client
+	return c
+}
+
+func (c *ClientHolder) WithKubernetesInformers(kubeInformers v1helpers.KubeInformersForNamespaces) *ClientHolder {
+	c.kubeInformers = kubeInformers
 	return c
 }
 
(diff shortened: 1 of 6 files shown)


pkg/operator/resource/resourceapply/generic.go
```diff
@@ -85,3 +155,23 @@ func ApplyDirectly(kubeClient kubernetes.Interface, recorder events.Recorder, ma
...
+}
+
+func (c *ClientHolder) secretsGetter() corev1client.SecretsGetter {
+	if c.kubeClient == nil {
+		return nil
+	}
+	if c.kubeInformers == nil {
+		return c.kubeClient.CoreV1()
+	}
+	return v1helpers.CachedSecretGetter(c.kubeClient.CoreV1(), c.kubeInformers)
```
sttts: I am still against this `corev1client.SecretsGetter` interface for the purpose of caching. The wrapper throws away the `GetOptions`. Instead make a smaller interface implemented by a more powerful implementation (clientset).

deads2k: non-overlapping sets makes adoption and usage harder,  but I'm happy to let this linger and take the rest.

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -1,19 +1,48 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextclientv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	apiextclientv1beta1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"k8s.io/klog"
```
mfojtik: nit: ...

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -37,17 +41,22 @@ type ApplyResult struct {
 	Error   error
 }
 
-// TODO provide ways to provide cached getters
 type ClientHolder struct {
```
mfojtik: nit: would be nice to have something like `NewClientHolder()` so one don't have to do `(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient),`

mfojtik: or just `NewKubeClientHolder()` ?

mfojtik:
LGTM

(there is missing godoc, but I can live with it or I will add it as follow up).

PR #676, opened 2020-01-21
Add encryption controllers to apiserver controllerset

stlaz:


mfojtik:
https://github.com/openshift/library-go/pull/687

this might be a better way

pkg/operator/apiserver/controllerset/apiservercontrollerset.go
```diff
@@ -166,6 +177,60 @@ func (cs *APIServerControllerSet) WithoutFinalizerController() *APIServerControl
...
+	targetNamespace string,
+	encryptionResources []schema.GroupResource,
+	dynamicClientForMigration dynamic.Interface,
+	apiServerClient configv1client.APIServerInterface,
+	apiServerInformer configv1informers.APIServerInformer,
+	kubeClient kubernetes.Interface,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	resourceSyncController resourcesynccontroller.ResourceSyncer,
+) *APIServerControllerSet {
+	nodeProvider := NewDaemonSetNodeProvider(
```
p0lyn0mial: this is already outdated, we already switched to using `deployments`

p0lyn0mial:
@mfojtik our intend is to build `APIServerControllerSet ` that consists of many more controllers. For example, at the moment it knows `apiServiceController` and `staticResourceController`  just to name a few. In the future, we are planning to add even more, like `workloadController`.

`APIServerControllerSet` will be used by `openshift-apiserver-operator` and `openshift-authentication-operator` and both require etcd encryption. Thus adding the encryption controllers seems to be okay.

p0lyn0mial:
/close

deprecated by https://github.com/openshift/library-go/pull/756

PR #677, opened 2020-01-22
make it possible to nerf feature gates if they are broken

deads2k:
/assign @mfojtik


mfojtik:
/lgtm

We need to make sure that every time this gets used we have blocker bug tracking revert.

PR #678, opened 2020-01-22
Add namespace flag

tnozicka:
Namespace flag is needed for running the controller locally.

Probably by not setting it I found the seqfault fixed in the other commit.

/cc @deads2k @mfojtik

### DIFF
--- a/pkg/controller/controllercmd/flags.go
+++ b/pkg/controller/controllercmd/flags.go
@@ -22,6 +22,8 @@ type ControllerFlags struct {
 	ConfigFile string
 	// KubeConfigFile points to a kubeconfig file if you don't want to use the in cluster config
 	KubeConfigFile string
+	// Namespace points to a base namespace for the controller and related events
+	Namespace string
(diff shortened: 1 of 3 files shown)


pkg/controller/controllercmd/flags.go
```diff
@@ -44,6 +46,7 @@ func (f *ControllerFlags) AddFlags(cmd *cobra.Command) {
 	flags.StringVar(&f.ConfigFile, "config", f.ConfigFile, "Location of the master configuration file to run from.")
 	cmd.MarkFlagFilename("config", "yaml", "yml")
 	flags.StringVar(&f.KubeConfigFile, "kubeconfig", f.KubeConfigFile, "Location of the master configuration file to run from.")
+	flags.StringVar(&f.Namespace, "namespace", f.Namespace, "Namespace where the controller is running. Auto-detected if run in cluster.")
```
mfojtik: make the flag hidden?

tnozicka: done

PR #679, opened 2020-01-23
Controller manager

mfojtik:
This builds on #666 and add graceful termination to all controllers via leader election. This will cause operators to exit zero. We exit non-zero, when graceful period is passed (we need to check operator workloads if we set this) or the controllers return error/panic before we ask them to die.

/cc @tnozicka

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -70,6 +70,9 @@ type ControllerBuilder struct {
 	authenticationConfig *operatorv1alpha1.DelegatedAuthentication
 	authorizationConfig  *operatorv1alpha1.DelegatedAuthorization
 	healthChecks         []healthz.HealthChecker
+
+	// nonZeroExitFn exists for unit testing
+	nonZeroExitFn func(args ...interface{})
 }
 
 // NewController returns a builder struct for constructing the command you want to run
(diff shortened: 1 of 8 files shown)


pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +242,42 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+	// 10s is the graceful termination time we give the controllers to finish their workers.
+	// when this time pass, we exit with non-zero code, killing all controller workers.
+	// NOTE: The pod must set the termination graceful time.
+	leaderElection.Callbacks.OnStartedLeading = b.getOnStartedLeadingFunc(ctx, controllerContext, 10*time.Second)
+
+	leaderelection.RunOrDie(ctx, leaderElection)
+	return nil
+}
+
+func (b ControllerBuilder) getOnStartedLeadingFunc(ctx context.Context, controllerContext *ControllerContext, gracefulTerminationDuration time.Duration) func(ctx context.Context) {
```
mfojtik: i extracted this from Run() because I want to have this strongly unit tested for all cases...

mfojtik:
proof PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/744

mfojtik:
@deads2k this gets us closer to proper graceful shutdown with 0 exit code.

PR #680, opened 2020-01-23
Create reusable package to manage manifests for use by CVO and oc

jottofar:
Create reusable package for use by CVO and 'oc adm release mirror'

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -1,0 +1,120 @@
+package manifest
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// Manifest stores Kubernetes object in Raw from a file.
+// It stores the GroupVersionKind for the manifest.
+// Raw and Obj should always be kept in sync such that
+// each provides the same data but in different formats.
+// To ensure Raw and Obj are always in sync, they should not
+// be set directly but rather only be set by calling
+// either method ManifestsFromFiles or ParseManifests.
+type Manifest struct {
+	// OriginalFilename is set to the filename this manifest was loaded from.
+	// It is not guaranteed to be set or be unique, but we will set it when
+	// loading from disk to provide better debuggability.
+	OriginalFilename string
+
+	Raw []byte
+	GVK schema.GroupVersionKind
+
+	Obj *unstructured.Unstructured
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface for the Manifest
+// type. It unmarshals bytes of a single kubernetes object to Manifest.
+func (m *Manifest) UnmarshalJSON(in []byte) error {
+	if m == nil {
+		return errors.New("Manifest: UnmarshalJSON on nil pointer")
+	}
+
+	// This happens when marshalling
+	// <yaml>
+	// ---	(this between two `---`)
+	// ---
+	// <yaml>
+	if bytes.Equal(in, []byte("null")) {
+		m.Raw = nil
+		return nil
+	}
+
+	m.Raw = append(m.Raw[0:0], in...)
+	udi, _, err := scheme.Codecs.UniversalDecoder().Decode(in, nil, &unstructured.Unstructured{})
+	if err != nil {
+		return errors.Wrapf(err, "unable to decode manifest")
+	}
+	ud, ok := udi.(*unstructured.Unstructured)
+	if !ok {
+		return fmt.Errorf("expected manifest to decode into *unstructured.Unstructured, got %T", ud)
+	}
+
+	m.GVK = ud.GroupVersionKind()
+	m.Obj = ud
+	return nil
+}
+
+// ManifestsFromFiles reads files and returns Manifests in the same order.
+// files should be list of absolute paths for the manifests on disk.
+func ManifestsFromFiles(files []string) ([]Manifest, error) {
+	var manifests []Manifest
+	var errs []error
+	for _, file := range files {
+		file, err := os.Open(file)
+		if err != nil {
+			errs = append(errs, errors.Wrapf(err, "error opening %s", file.Name()))
+			continue
+		}
+		defer file.Close()
+
+		ms, err := ParseManifests(file)
+		if err != nil {
+			errs = append(errs, errors.Wrapf(err, "error parsing %s", file.Name()))
+			continue
+		}
+		for _, m := range ms {
+			m.OriginalFilename = filepath.Base(file.Name())
+		}
+		manifests = append(manifests, ms...)
+	}
+
+	agg := utilerrors.NewAggregate(errs)
+	if agg != nil {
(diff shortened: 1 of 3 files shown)


pkg/manifest/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+	OriginalFilename string
+
+	Raw []byte
+	GVK schema.GroupVersionKind
+
+	Obj *unstructured.Unstructured
+}
+
+// UnmarshalJSON unmarshals bytes of single kubernetes object to Manifest.
+func (m *Manifest) UnmarshalJSON(in []byte) error {
```
sttts: why is this no constructor?

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+// Manifest stores Kubernetes object in Raw from a file.
+// It stores the GroupVersionKind for the manifest.
+type Manifest struct {
```
sttts: what are the invariants? Raw and Obj are always in sync and set?

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+	// OriginalFilename is set to the filename this manifest was loaded from.
+	// It is not guaranteed to be set or be unique, but we will set it when
+	// loading from disk to provide better debuggability.
+	OriginalFilename string
+
+	Raw []byte
+	GVK schema.GroupVersionKind
+
+	Obj *unstructured.Unstructured
```
sttts: Object

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+// Manifest stores Kubernetes object in Raw from a file.
+// It stores the GroupVersionKind for the manifest.
+type Manifest struct {
+	// OriginalFilename is set to the filename this manifest was loaded from.
+	// It is not guaranteed to be set or be unique, but we will set it when
+	// loading from disk to provide better debuggability.
+	OriginalFilename string
+
+	Raw []byte
+	GVK schema.GroupVersionKind
```
sttts: Obj gives the same information. Why a copy?

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+		return fmt.Errorf("expected manifest to decode into *unstructured.Unstructured, got %T", ud)
+	}
+
+	m.GVK = ud.GroupVersionKind()
+	m.Obj = ud.DeepCopy()
+	return nil
+}
+
+// Object returns underlying metav1.Object
+func (m *Manifest) Object() metav1.Object { return m.Obj }
```
sttts: with this, why a public Obj?

pkg/manifest/manifest.go
jottofar: Not sure as this is existing code I simply migrated "as is" from CVO. The guidance I was given is to effect as little change as possible to the CVO code. I can add a constructor but forcing it's use by making the Manifest struct private would have big impact to CVO. Would providing a constructor but leaving Manifest public suffice?

pkg/manifest/manifest.go
jottofar: They are always in sync with it's current use by CVO and my planned use by oc. The struct only gets set here by this call to ParseManifests in CVO:

https://github.com/jottofar/cluster-version-operator/blob/732ae625e991bf1447370b537caf9bcc22c850dc/pkg/payload/payload.go#L147

Also, both manifest.ManifestsFromFiles and manifest.UnmarshalJSON set both

pkg/manifest/manifest.go
jottofar: Assume you want me to change Obj to Object? If so, is this absolutely required? Again, going back to goal of trying to minimize changes to CVO.

pkg/manifest/manifest.go
jottofar: It is currently accessed by CVO through both Obj and Raw.

pkg/manifest/manifest.go
jottofar: CVO accesses it directly in some cases:

https://github.com/jottofar/cluster-version-operator/blob/732ae625e991bf1447370b537caf9bcc22c850dc/pkg/cvo/cvo.go#L304

And in others Object() is used:

https://github.com/jottofar/cluster-version-operator/blob/732ae625e991bf1447370b537caf9bcc22c850dc/pkg/cvo/sync_worker.go#L918

pkg/manifest/manifest.go
sttts: please document that. It's central for the type.

pkg/manifest/manifest.go
sttts: change CVO

pkg/manifest/manifest.go
jottofar: added a ctor

pkg/manifest/manifest.go
jottofar: documented

pkg/manifest/manifest.go
jottofar: documented structure

pkg/manifest/manifest.go
jottofar: I removed the Object func

pkg/manifest/manifest.go
sttts: why not make this func private?

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,132 @@
...
+	udi, _, err := scheme.Codecs.UniversalDecoder().Decode(in, nil, &unstructured.Unstructured{})
+	if err != nil {
+		return errors.Wrapf(err, "unable to decode manifest")
+	}
+	ud, ok := udi.(*unstructured.Unstructured)
+	if !ok {
+		return fmt.Errorf("expected manifest to decode into *unstructured.Unstructured, got %T", ud)
+	}
+
+	m.GVK = ud.GroupVersionKind()
```
sttts: here you see how senseless it is to store GVK independently.

pkg/manifest/manifest.go
```diff
@@ -0,0 +1,132 @@
...
+	m.Obj = ud.DeepCopy()
```
sttts: why deepcopy?

pkg/manifest/manifest.go
sttts: The answer to my questions: `UnmarshalJSON` is called by the yaml or json decoder. In other words, `Manifest` knows how to unmarshal itself. This means we must have this signature. And it also means the constructor is not used at all. Sorry for pointing you into the wrong direction.

jottofar: Removed ctor and documented UnmarshalJSON.

pkg/manifest/manifest.go
jottofar: Must be required. I removed the DeepCopy and existing code gets a SIGSEGV nil pointer when trying to access Obj:

`src := fmt.Sprintf("the config map %s/%s", ms[i].Obj.GetNamespace(), ms[i].Obj.GetName())`

sttts: That's odd. Of course I meant `m.Obj = ud`. Doesn't work?

jottofar: Yes, when I commented out line 65 'm.Obj = ud.DeepCopy()' it failed with SIGSEGV.

jottofar: Changed and does work

PR #682, opened 2020-01-27
Allow adjusting serving info to avoid port collisions

tnozicka:
When we run multiple library-go based binaries in a pod ports will collide.

e.g. KCM collides on 8433 that was hardcoded

/cc @mfojtik @soltysh

### DIFF
--- a/pkg/controller/controllercmd/flags.go
+++ b/pkg/controller/controllercmd/flags.go
@@ -24,6 +24,8 @@ type ControllerFlags struct {
 	KubeConfigFile string
 	// Namespace points to a base namespace for the controller and related events
 	Namespace string
+	// BindAddress is the ip:port to serve on
+	BindAddress string
 	// TerminateOnFiles is a list of files. If any of these changes, the process terminates.
 	TerminateOnFiles []string
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/controllercmd/flags.go
```diff
@@ -48,7 +50,7 @@ func (f *ControllerFlags) AddFlags(cmd *cobra.Command) {
 	flags.StringVar(&f.KubeConfigFile, "kubeconfig", f.KubeConfigFile, "Location of the master configuration file to run from.")
 	cmd.MarkFlagFilename("kubeconfig", "kubeconfig")
 	flags.StringVar(&f.Namespace, "namespace", f.Namespace, "Namespace where the controller is running. Auto-detected if run in cluster.")
-	cmd.Flags().MarkHidden("namespace")
```
tnozicka: @mfojtik it turned out it wasn't just dev flag and KCMO is using it now

pkg/controller/controllercmd/flags.go
```diff
@@ -48,7 +50,7 @@ func (f *ControllerFlags) AddFlags(cmd *cobra.Command) {
...
+	flags.StringVar(&f.BindAddress, "listen", f.Namespace, "The ip:port to serve on.")
```
mfojtik: this seems wrong?

tnozicka: fixed, thanks

sttts:
Wrong approach. We should use the serving info flags.

sttts:
/hold cancel
/approve

I thought there is a better way by just using the serving options from k8s.io/apiserver. But the wiring here reinvents the wheel 🤷‍♂️

PR #685, opened 2020-01-28
controller: deflake controller factory test

mfojtik:
Rework the shutdown test to use channels instead time.Sleep, make it more reliable and flake prone.

### DIFF
--- a/pkg/controller/factory/controller_test.go
+++ b/pkg/controller/factory/controller_test.go
@@ -106,51 +106,52 @@ func TestResyncController(t *testing.T) {
 
 func TestMultiWorkerControllerShutdown(t *testing.T) {
 	controllerCtx, shutdown := context.WithCancel(context.TODO())
-	factory := New().ResyncEvery(5 * time.Minute) // make sure we only call 1 sync manually
+	factory := New().ResyncEvery(10 * time.Minute) // make sure we only call 1 sync manually
 	var workersShutdownMutex sync.Mutex
+	var syncCallCountMutex sync.Mutex
+
 	workersShutdownCount := 0
 	syncCallCount := 0
-	var syncCallCountMutex sync.Mutex
-	recorder := events.NewInMemoryRecorder("shutdown-controller")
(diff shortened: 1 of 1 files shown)


pkg/controller/factory/controller_test.go
```diff
@@ -106,51 +106,52 @@ func TestResyncController(t *testing.T) {
...
+		case 1:
 			syncContext.Queue().Add("TestKey1")
 			syncContext.Queue().Add("TestKey2")
 			syncContext.Queue().Add("TestKey3")
 			syncContext.Queue().Add("TestKey4")
-			syncContext.Queue().Add("TestKey5")
+		case 5:
+			close(allWorkersBusy)
 		}
 		syncCallCountMutex.Unlock()
```
sttts: move up behind `++`?

pkg/controller/factory/controller_test.go
```diff
@@ -106,51 +106,52 @@ func TestResyncController(t *testing.T) {
...
 
-	shutdown() // initiate controller shutdown
-	<-runComplete
+	// this blocks until all workers are shut down.
+	controller.Run(controllerCtx, 5)
 
 	workersShutdownMutex.Lock()
-	if workersShutdownCount != 6 {
-		t.Fatalf("expected all 6 workers to gracefully shutdown, got %d", workersShutdownCount)
+	if workersShutdownCount != 5 {
```
sttts: how do you know that all 5 have finished the sync func?

pkg/controller/factory/controller_test.go
mfojtik: you need syncCallCount in switch

pkg/controller/factory/controller_test.go
mfojtik: you count them when they finish :)

sttts: There is a race.

sttts: Didn't realize `controller.Run(controllerCtx, 5)` blocks.

PR #687, opened 2020-01-28
Migrate encryption controllers to factory

mfojtik:
This migrates encryption controllers to use controller factory.

### DIFF
--- a/pkg/operator/encryption/controllers.go
+++ b/pkg/operator/encryption/controllers.go
@@ -1,6 +1,8 @@
 package encryption
 
 import (
+	"context"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -19,7 +21,7 @@ import (
 )
 
 type runner interface {
-	Run(stopCh <-chan struct{})
+	Run(ctx context.Context, workers int)
 }
 
 func NewControllers(
@@ -101,10 +103,10 @@ type Controllers struct {
 	controllers []runner
 }
 
-func (c *Controllers) Run(stopCh <-chan struct{}) {
+func (c *Controllers) Run(ctx context.Context, workers int) {
 	for _, controller := range c.controllers {
 		con := controller // capture range variable
-		go con.Run(stopCh)
+		go con.Run(ctx, workers)
 	}
-	<-stopCh
+	<-ctx.Done()
 }
--- a/pkg/operator/encryption/controllers/condition_controller.go
+++ b/pkg/operator/encryption/controllers/condition_controller.go
@@ -1,22 +1,18 @@
 package controllers
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/apimachinery/pkg/util/wait"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/workqueue"
-	"k8s.io/klog"
 
-	operatorv1 "github.com/openshift/api/operator/v1"
-
+	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
 	"github.com/openshift/library-go/pkg/operator/encryption/state"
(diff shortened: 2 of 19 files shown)


pkg/operator/encryption/controllers/controller.go
```diff
@@ -28,9 +28,10 @@ func setUpInformers(
 
 	managedSecretsInformer := kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer()
 	managedSecretsInformer.AddEventHandler(eventHandler)
+	deployer.AddEventHandler(eventHandler)
 
-	return append([]cache.InformerSynced{
+	return []cache.InformerSynced{
 		operatorInformer.HasSynced,
 		managedSecretsInformer.HasSynced,
```
sttts: deployer.HasSynched

pkg/operator/encryption/controllers/condition_controller.go
```diff
@@ -54,26 +45,25 @@ func NewConditionController(
...
 	}
 
-	c.preRunCachesSynced = setUpInformers(deployer, operatorClient, kubeInformersForNamespaces, c.eventHandler())
-
-	return c
+	return factory.New().WithInformers(
+		kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
+		operatorClient.Informer(),
+		deployer,
+	).ResyncEvery(time.Second).WithSync(c.sync).ToController("EncryptionConditionController", eventRecorder.WithComponentSuffix("encryption-condition-controller"))
```
sttts: resync every second?

pkg/controller/factory/factory.go
```diff
@@ -16,14 +16,21 @@ type Factory struct {
...
-	informers             []cache.SharedInformer
+	informers             []RegistrableInformer
 	namespaceInformers    []*namespaceInformer
 	cachesToSync          []cache.InformerSynced
 	interestingNamespaces sets.String
 }
 
+// RegistrableInformer represents any structure that allow to register event handlers and informs if caches are synced.
+// Any SharedInformer will comply.
+type RegistrableInformer interface {
```
sttts: this isn't part of the factory library?

mfojtik: it is... it is a subset of SharedInformer interface so the "Deployer" thing can comply... We don't need full shared informer access... just event handler registration and has synced.

pkg/operator/encryption/controllers/controller.go
mfojtik: good catch

pkg/operator/encryption/controllers/condition_controller.go
mfojtik: @sttts that was pre-existing https://github.com/openshift/library-go/pull/687/files/318a3bf613627b8148ed47788882d3f1a1490705#diff-a03dd41db6b9c24ab2e28afc2b14fe5dL199

sttts: wow. Sounds like we would prefer something like 30 sec at least.

mfojtik: @sttts this is all pre-existing... i would do this as a follow up maybe?

sttts: follow-up is fine

pkg/operator/encryption/deployer/unionrevisionedpod.go
```diff
@@ -13,9 +13,10 @@ import (
 // UnionRevisionLabelPodDeployer provides unified state from multiple distinct deployers.
 type UnionRevisionLabelPodDeployer struct {
 	delegates []statemachine.Deployer
+	hasSynced []cache.InformerSynced
 }
 
-var _ statemachine.Deployer = &RevisionLabelPodDeployer{}
+var _ statemachine.Deployer = &UnionRevisionLabelPodDeployer{}
```
mfojtik: @p0lyn0mial @sttts this somehow slipped trough review.

mfojtik:
unit tests are totally flaking

@stevekuznetsov @droslean have we done something in base images that might explain this?

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -58,12 +54,10 @@ const (
...
 type migrationController struct {
 	component string
+	name      string
 
 	operatorClient operatorv1helpers.OperatorClient
 
-	queue         workqueue.RateLimitingInterface
-	eventRecorder events.Recorder
-
 	preRunCachesSynced []cache.InformerSynced
```
p0lyn0mial: is `preRunCachesSynced` used ?

pkg/operator/encryption/controllers.go
```diff
@@ -101,10 +103,10 @@ type Controllers struct {
 	controllers []runner
 }
 
-func (c *Controllers) Run(stopCh <-chan struct{}) {
+func (c *Controllers) Run(ctx context.Context, workers int) {
```
p0lyn0mial: just wondering if we want to expose `workers` IIRC all `encryption` controllers are single-threaded.

droslean:
> unit tests are totally flaking

As far as I know, we didn't make any changes. Can someone debug this and make sure that this is an actual infrastructure problem?

PR #689, opened 2020-01-28
Add batch/CronJob resource helpers

damemi:
These helpers would be beneficial to the descheduler operator, which works by creating a cronjob to run the descheduler image. Currently, it is creating this cronjob manually and switching to bindata (and these helpers) would clean up the code nicely

### DIFF
--- a/pkg/operator/resource/resourceapply/batch.go
+++ b/pkg/operator/resource/resourceapply/batch.go
@@ -1,0 +1,39 @@
+package resourceapply
+
+import (
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourceapply/batch.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	existingCopy := existing.DeepCopy()
+
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	if !*modified {
```
deads2k: don't you need to check if the generations match?

soltysh: The current flow does not use generation, so this can be ignored, at least for now. New one, hopefully will.

damemi:
/close
We don't need cronjobs anymore in descheduler operator

PR #691, opened 2020-01-29
Fortify build machinery

tnozicka:
build-machinery supports only golang >=1.13
(Yes, we have come to a point where we need this enforced. Version check is on my list next.)

There was a case when someone was building it with 1.12 which resulted in silently failing `go list` in a subshell and `GO_BUILD_PACKAGES_EXPANDED` being empty. Since the time we have added building multiple packages and introduced a for loop, with empty var it executes nothing, but ends with exit code `0`. That made the image build continue. Unfortunately this was combined with a bug in the image builder that also didn't error out on `COPY` instruction missing the files.

```
RUN make build --warn-undefined-variables
build flag -mod=vendor only valid when using modules
make: Nothing to be done for `build'.
```

We need to fix our part to be resilient and also follow up on the image builder issue separately.

I understand this doesn't look extra nice, but correctness and checking exit codes should take precedence

Requires:
 - [ ] bumping CI to centos:8 to get make 4.2

/cc @mfojtik @soltysh 

@mfojtik this needs `GNU Make 4.2`, try how it goes on your Mac

@openshift/sig-master

### DIFF
--- a/alpha-build-machinery/Makefile
+++ b/alpha-build-machinery/Makefile
@@ -1,4 +1,5 @@
 SHELL :=/bin/bash
+
 all: verify
 .PHONY: all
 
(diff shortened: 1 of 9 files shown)


tnozicka:
The subshell return value checking needs centos:8 for make 4.2. I have split the build fix into https://github.com/openshift/library-go/pull/692 as we have to have for bumping CI builder images to centos:8

sttts:
Where exactly do we need the new make?

tnozicka:
> Where exactly do we need the new make?
in the bulder `:golang-1.13` image, not the runtime/base

tnozicka:
> Where exactly do we need the new make?

`.SHELLSTATUS` variable need Make 4.2

soltysh:
@tnozicka I'm gonna close this since this will have to move to the new repo.

PR #693, opened 2020-01-30
fix detection of namespace

deads2k:


deads2k:
we've agreed with eparis that library-go doesn't require the label

PR #694, opened 2020-01-30
Add version check and wire it for `go`

tnozicka:
We have been hitting issues with people and tools using an old unsupported version resulting in bugs, especially since go 1.13 switch to modules on by default. The CI golang version enforcement is also super fragile and our Dockerfiles are often outdated.

This PR will make sure we use the correct version for `go`. (The version check can also be wired for other binaries.) Consumers can choose their own version explicitly by adding `GO_REQUIRED_MIN_VERSION :=1.13.5` to their Makefile. If someone want to explicitly build with old version, `make GO_REQUIRED_MIN_VERSION :=` will do the trick.

/cc @mfojtik @soltysh @sttts 

@openshift/sig-master


alpha-build-machinery/make/examples/version-checks/Makefile
```diff
@@ -0,0 +1,15 @@
+include $(addprefix ../../, \
+	lib/version.mk \
+)
+
+PROGRAM_NAME :=go
+
+REQUIRED_VERSION :=1.12.5
+
+CURRENT_VERSION :=1.13.1
```
sttts: who will maintain this?

sttts:
Technically LGTM. Have some concerns that we will properly maintain these values in our 23 repos.

Is there a way to bound the current version to the one used in the Dockerfile?

tnozicka:
> Is there a way to bound the current version to the one used in the Dockerfile?

a) you'd chose to set the version manually in your Makefile and update it together with your dockerfile

b) the value that comes from library-go. We'll bump it with every rebase / release wise. We should have a tool showing where are dependencies (branches) are and see that all the repos are up to date (we need it for other reasons)

a) overrides b)

I suppose for a) you could parse the value from the dockerfile based on naming scheme, but our Dockerfiles are located in different locations, sot that's not a job for the library-go to guess. I am not sure if it's worth it either. You use builder golang-1.14 and still require only golang-1.13. Also e.g. kube requires 1.13.5 so the image refs are missing the minor number to parse.

alpha-build-machinery/make/examples/version-checks/Makefile
sttts: nvmd. It's just the example.

PR #695, opened 2020-02-02
Bug 1775224: Allow controllers to shutdown gracefully

vareti:
It is noticed that when operators are restarted during upgrades or file observer reactions, even though it is supposed to be a graceful shutdown and all resources are expected to be released, leader election lock that is acquired by this operator is not released and remains to be a stale entity until the lease expires. This results in almost a minute delay for operator to acquire a new lease when it restarts. Upgrades and dev workflows can be sped up if the operator is allowed to do graceful shutdown instead of killing the process abruptly with `klog.Fatal(f)`

When these changes were tested with `service-ca-operator`,log messages indicating that operator is waiting to acquire lease are not seen anymore.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -207,7 +207,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		if err := server.PrepareRun().Run(ctx.Done()); err != nil {
 			klog.Error(err)
 		}
-		klog.Fatal("server exited")
+		klog.Error("server exited")
 	}()
 
 	protoConfig := rest.CopyConfig(clientConfig)
@@ -240,7 +240,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 
 	leaderElection.Callbacks.OnStartedLeading = func(ctx context.Context) {
 		if err := b.startFunc(ctx, controllerContext); err != nil {
-			klog.Fatal(err)
+			klog.Error(err)
 		}
 	}
 	leaderelection.RunOrDie(ctx, leaderElection)
--- a/pkg/controller/fileobserver/observer_polling.go
+++ b/pkg/controller/fileobserver/observer_polling.go
@@ -133,7 +133,7 @@ func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
 		return false, nil
 	})
 	if err != nil {
-		klog.Fatalf("file observer failed: %v", err)
+		klog.Errorf("file observer failed: %v", err)
 	}
 }
 


pkg/controller/controllercmd/builder.go
```diff
@@ -207,7 +207,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		if err := server.PrepareRun().Run(ctx.Done()); err != nil {
 			klog.Error(err)
 		}
-		klog.Fatal("server exited")
```
tnozicka: /hold

This won't do graceful shutdown. You can't do graceful termination unless you know that all the business logic is shutdown already. That's why this intentionally calls os.Exit. Graceful termination means having to track all the gorutines running business logic and waiting for them to end, releasing the lock only after they finish. We have been investing some time to that with @mfojtik but it is a long run that needs several fixes in how we start controller and how we need to track gorutines.

Although I don't think you need graceful termination to fix the BZ. ReleaseOnCancel immediately followed by os.Exit / glog.Fatal should release the lock. I think @mfojtik added the ReleaseOnCancel recently - did you check if it is still not releasing it?

pkg/controller/controllercmd/builder.go
```diff
@@ -240,7 +240,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 
 	leaderElection.Callbacks.OnStartedLeading = func(ctx context.Context) {
 		if err := b.startFunc(ctx, controllerContext); err != nil {
-			klog.Fatal(err)
+			klog.Error(err)
```
tnozicka: the call back is asynchronous so if you error instead of an exit is it gonna be still running dead branch?

pkg/controller/controllercmd/builder.go
```diff
@@ -207,7 +207,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		if err := server.PrepareRun().Run(ctx.Done()); err != nil {
 			klog.Error(err)
 		}
-		klog.Fatal("server exited")
+		klog.Error("server exited")
```
tnozicka: async, changes the flow to ignore the error and not to exit the program

pkg/controller/fileobserver/observer_polling.go
```diff
@@ -133,7 +133,7 @@ func (o *pollingObserver) processReactors(stopCh <-chan struct{}) {
 		return false, nil
 	})
 	if err != nil {
-		klog.Fatalf("file observer failed: %v", err)
+		klog.Errorf("file observer failed: %v", err)
```
tnozicka: if file observer fails we can't just continue running

mfojtik:
I think what you have done is not enough... Take a look at https://github.com/openshift/library-go/pull/679 :-)

vareti:
This is not even remotely close to what @mfojtik has done. Closing it in favor of #679

PR #697, opened 2020-02-04
Bug 1793267: Add -mod=vendor fallback for `go list`

tnozicka:
if built without network connectivity, `go list` without `-mod=vendor` flag fails and GO_PACKAGE would be empty

/cc @soltysh


soltysh:
Looks like the bot can't pick up approve label correctly, setting one manually.

PR #698, opened 2020-02-04
build-machinery: use -trimpath by default

mfojtik:
This makes panics looks better by trimming the build root from the paths.
It makes coping the the file paths to editors easier and improve developers life :-)

```
	-trimpath
		remove all file system paths from the resulting executable.
		Instead of absolute file system paths, the recorded file names
		will begin with either "go" (for the standard library),
		or a module path@version (when using modules),
		or a plain import path (when using GOPATH).
```

This was added in go-1.13.


tnozicka:
lgtm but you need to run `make -C alpha-build-machinery update`

PR #699, opened 2020-02-05
Properly terminate controllers on leader election lost/shutdown

mfojtik:
First commit add controller manager that groups multiple controllers and **wait** until all controllers are terminated. This is used later in leader election block for builder where we allow controllers to properly finish before terminating them forcefully. IF controllers exit cleanly and in time, we exit 0 (must be for graceful termination), otherwise we exit 1.

Additionally, this converts resourcesync controller to use factory and allow to build the SyncContext upfront for unit testing or for more tide control on event recorder.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"sync"
 	"time"
 
@@ -70,6 +71,11 @@ type ControllerBuilder struct {
 	authenticationConfig *operatorv1alpha1.DelegatedAuthentication
 	authorizationConfig  *operatorv1alpha1.DelegatedAuthorization
 	healthChecks         []healthz.HealthChecker
+
+	// nonZeroExitFn takes a function that exit the process with non-zero code.
+	// This stub exists for unit test where we can check if the graceful termination work properly.
+	// Default function will klog.Warning(args) and os.Exit(1).
+	nonZeroExitFn func(args ...interface{})
 }
 
 // NewController returns a builder struct for constructing the command you want to run
@@ -78,6 +84,10 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
 		startFunc:        startFunc,
 		componentName:    componentName,
 		observerInterval: defaultObserverInterval,
+		nonZeroExitFn: func(args ...interface{}) {
+			klog.Warning(args...)
+			os.Exit(1)
+		},
 	}
 }
 
@@ -204,9 +214,9 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 
 		go func() {
 			if err := server.PrepareRun().Run(ctx.Done()); err != nil {
-				klog.Error(err)
+				klog.Fatal(err)
 			}
-			klog.Fatal("server exited")
+			klog.Info("server exited")
 		}()
 	}
 
@@ -226,7 +236,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		if err := b.startFunc(ctx, controllerContext); err != nil {
 			return err
 		}
-		return fmt.Errorf("exited")
+		return nil
 	}
 
 	// ensure blocking TCP connections don't block the leader election
@@ -238,13 +248,41 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		return err
 	}
 
-	leaderElection.Callbacks.OnStartedLeading = func(ctx context.Context) {
-		if err := b.startFunc(ctx, controllerContext); err != nil {
-			klog.Fatal(err)
+	// 10s is the graceful termination time we give the controllers to finish their workers.
+	// when this time pass, we exit with non-zero code, killing all controller workers.
+	// NOTE: The pod must set the termination graceful time.
+	leaderElection.Callbacks.OnStartedLeading = b.getOnStartedLeadingFunc(controllerContext, 10*time.Second)
+
+	leaderelection.RunOrDie(ctx, leaderElection)
+	return nil
+}
+
+func (b ControllerBuilder) getOnStartedLeadingFunc(controllerContext *ControllerContext, gracefulTerminationDuration time.Duration) func(ctx context.Context) {
+	return func(ctx context.Context) {
+		stoppedCh := make(chan struct{})
+		go func() {
+			defer close(stoppedCh)
+			if err := b.startFunc(ctx, controllerContext); err != nil {
+				b.nonZeroExitFn(fmt.Sprintf("graceful termination failed, controllers failed with error: %v", err))
+			}
+		}()
+
+		select {
+		case <-ctx.Done(): // context closed means the process likely received signal to terminate
+		case <-stoppedCh:
+			// if context was not cancelled (it is not "done"), but the startFunc terminated, it means it terminated prematurely
+			// when this happen, it means the controllers terminated without error.
+			if ctx.Err() == nil {
+				b.nonZeroExitFn("graceful termination failed, controllers terminated prematurely")
+			}
+		}
+
+		select {
+		case <-time.After(gracefulTerminationDuration): // when context was closed above, give controllers extra time to terminate gracefully
+			b.nonZeroExitFn(fmt.Sprintf("graceful termination failed, some controllers failed to shutdown in %s", gracefulTerminationDuration))
+		case <-stoppedCh: // stoppedCh here means the controllers finished termination and we exit 0
 		}
 	}
-	leaderelection.RunOrDie(ctx, leaderElection)
-	return fmt.Errorf("exited")
 }
 
 func (b *ControllerBuilder) getComponentNamespace() (string, error) {
--- a/pkg/controller/controllercmd/builder_test.go
+++ b/pkg/controller/controllercmd/builder_test.go
@@ -1,0 +1,184 @@
+package controllercmd
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestControllerBuilder_getOnStartedLeadingFunc(t *testing.T) {
+	nonZeroExits := []string{}
+	b := ControllerBuilder{
+		nonZeroExitFn: func(args ...interface{}) {
+			nonZeroExits = append(nonZeroExits, fmt.Sprintf("%#v", args))
+		},
+		startFunc: func(ctx context.Context, controllerContext *ControllerContext) error {
+			time.Sleep(1 * time.Second)
+			return nil
+		},
+	}
+
+	// controllers finished prematurely, without being asked to finish
+	b.getOnStartedLeadingFunc(&ControllerContext{}, 3*time.Second)(context.TODO())
+	if len(nonZeroExits) != 1 || !strings.Contains(nonZeroExits[0], "controllers terminated prematurely") {
+		t.Errorf("expected controllers to exit prematurely, got %#v", nonZeroExits)
+	}
+
+	// controllers finished gracefully after context was cancelled, with zero exit status
+	nonZeroExits = []string{}
+	ctx, cancel := context.WithCancel(context.TODO())
+	go func() {
+		defer cancel()
+		time.Sleep(1 * time.Second)
+	}()
+	b.startFunc = func(ctx context.Context, controllerContext *ControllerContext) error {
+		time.Sleep(2 * time.Second)
+		return nil
+	}
+	b.getOnStartedLeadingFunc(&ControllerContext{}, 5*time.Second)(ctx)
+	if len(nonZeroExits) > 0 {
+		t.Errorf("expected controllers to exit gracefully, but got %#v", nonZeroExits)
+	}
+
+	// controllers passed the graceful termination duration and are force killed
+	nonZeroExits = []string{}
+	ctx, cancel = context.WithCancel(context.TODO())
+	go func() {
+		defer cancel()
+		time.Sleep(1 * time.Second)
+	}()
+	b.startFunc = func(ctx context.Context, controllerContext *ControllerContext) error {
+		time.Sleep(3 * time.Second)
+		return nil
+	}
+	b.getOnStartedLeadingFunc(&ControllerContext{}, 1*time.Second)(ctx)
+	if len(nonZeroExits) != 1 && !strings.Contains(nonZeroExits[0], "some controllers failed to shutdown in 1s") {
+		t.Errorf("expected controllers to failed finish in 1s, got %#v", nonZeroExits)
+	}
+}
+
+func TestControllerBuilder_GracefulShutdown(t *testing.T) {
+	nonZeroExitCh := make(chan struct{})
+	startedCh := make(chan struct{})
+	ctx, shutdown := context.WithCancel(context.Background())
+
+	b := &ControllerBuilder{
+		nonZeroExitFn: func(args ...interface{}) {
+			t.Logf("non-zero exit detected: %+v", args)
+			close(nonZeroExitCh)
+		},
+		startFunc: func(ctx context.Context, controllerContext *ControllerContext) error {
+			close(startedCh)
+			<-ctx.Done()
+			return nil
+		},
+	}
+
+	// wait for controller to run, then give it 1s and shutdown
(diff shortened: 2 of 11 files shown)


mfojtik:
Proof in kube-apiserver-operator: https://github.com/openshift/cluster-kube-apiserver-operator/pull/745

pkg/controller/controllercmd/builder.go
```diff
@@ -70,6 +70,9 @@ type ControllerBuilder struct {
 	authenticationConfig *operatorv1alpha1.DelegatedAuthentication
 	authorizationConfig  *operatorv1alpha1.DelegatedAuthorization
 	healthChecks         []healthz.HealthChecker
+
+	// nonZeroExitFn exists for unit testing
+	nonZeroExitFn func(args ...interface{})
```
sttts: what do they do? comment

sttts: `// ... called before terminating with non-zero exit code` ?

mfojtik: @sttts this is a unit test stub (see comment)... so I'm able to test the Fatalf() behavior.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +242,42 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+	// 10s is the graceful termination time we give the controllers to finish their workers.
+	// when this time pass, we exit with non-zero code, killing all controller workers.
+	// NOTE: The pod must set the termination graceful time.
+	leaderElection.Callbacks.OnStartedLeading = b.getOnStartedLeadingFunc(ctx, controllerContext, 10*time.Second)
+
+	leaderelection.RunOrDie(ctx, leaderElection)
+	return nil
+}
+
+func (b ControllerBuilder) getOnStartedLeadingFunc(ctx context.Context, controllerContext *ControllerContext, gracefulTerminationDuration time.Duration) func(ctx context.Context) {
```
sttts: s/getO/o/

pkg/controller/controllercmd/builder.go
sttts: what are args? If this is input for Sprintf, leave the format string as extra arg.

mfojtik: @sttts this is not public API, right? only used in tests and private... I can surely improve the comments.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +242,42 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+	return func(context.Context) {
```
sttts: I don't understand that we have this context, and the outer context.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +242,42 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+		stopCh := make(chan struct{})
```
sttts: this is not a stop channel, but a `stoppedCh`.

pkg/controller/controllercmd/builder.go
mfojtik: they are basically the same... I can pass the ctx from the outer if it makes it cleaner

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +242,42 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+		go func() {
+			defer close(stopCh)
+			if err := b.startFunc(ctx, controllerContext); err != nil {
+				b.nonZeroExitFn(fmt.Sprintf("controllers failed with error: %v", err))
+			}
+		}()
+
+		// when the stopCh is closed before the context is cancelled, it means the controllers
+		// died before we asked them to die.
```
sttts: This is wrong. You are producing a race here during normal shutdown. The controllers are fast, you will run into the "prematurely" case altough the flow is totally fine.

pkg/controller/factory/base_controller.go
```diff
@@ -84,6 +89,9 @@ func QueueKey(name string) string {
 }
 
 func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.Duration) {
+	if reflect.ValueOf(interval).IsZero() {
```
sttts: reflect? 😱 Why?

pkg/controller/factory/factory.go
```diff
@@ -14,16 +13,24 @@ import (
...
-	informers             []cache.SharedInformer
+	informers             []RegistrableInformer
 	namespaceInformers    []*namespaceInformer
 	cachesToSync          []cache.InformerSynced
 	interestingNamespaces sets.String
 }
 
+// RegistrableInformer represents any structure that allow to register event handlers and informs if caches are synced.
+// Any SharedInformer will comply.
+type RegistrableInformer interface {
```
sttts: is `Registrable` really a word?

pkg/controller/factory/base_controller.go
mfojtik: it seems like in go-1.13 this is the "right" way to check if something is "zero" (iow. not set).

pkg/controller/factory/factory.go
mfojtik: open for better suggestions

> registrable
> Admitting of registration; that may or can be registered.

pkg/controller/factory/factory.go
```diff
@@ -76,13 +83,26 @@ func (f *Factory) WithRuntimeObject() *Factory {
 	return f
 }
 
+// WithSyncContext allows to specify existing sync context to this factory.
+// This is useful during unit testing.
+func (f *Factory) WithSyncContext(ctx SyncContext) *Factory {
```
sttts: can a factory exist without `SyncFunc`? Can a sync func exist without `SyncContext`?

sttts: What is the default if there is any?

pkg/controller/manager/controller_manager.go
```diff
@@ -2,24 +2,56 @@ package manager
...
+
+	"k8s.io/klog"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 )
 
 type ControllerManager interface {
-	Run(ctx context.Context)
+	Start(ctx context.Context)
+	Add(controller factory.Controller, workers int) ControllerManager
```
sttts: WithController

pkg/controller/factory/factory.go
mfojtik: SyncContext is set for each sync() call in case it contain the runtime object for example. Factory can exists without sync context if you don't call sync()...
This stub only exists so you can re-use existing event recorder (in case you have unit test with unit testing event recorder)...

If the sync context is not provided, factory create one for you.

pkg/controller/controllercmd/builder.go
mfojtik: @sttts the controllers can only finish when the context is cancelled... if the function above exit before the context is cancelled, it means something caused the controller to terminate, without error...

I think even if the normal shutdown is fast, the context must be cancelled and therefore the <-ctx.Done() will take over the case will cause the stoppedChan check pass. (i swapped them + I added one more check into stopped branch. Will push shortly.

pkg/controller/controllercmd/builder.go
mfojtik: @sttts improved godoc

pkg/controller/controllercmd/builder.go
sttts: there is no determinism for multiple contexts.

pkg/controller/factory/base_controller.go
sttts: No, it's not. Just compare with `time.Duration(0)`. Duration is just an int64.

pkg/controller/factory/factory.go
sttts: hab about just `Informer` ?

pkg/controller/factory/factory.go
sttts: document that: `// ... If this is not called, a sync context is created by the factory`

pkg/controller/controllercmd/builder_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	// wait for controller to run, then give it 1s and shutdown
+	go func() {
+		defer shutdown()
+		<-startedCh
+		time.Sleep(time.Second)
+	}()
+
+	stopCh := make(chan struct{})
```
sttts: stoppedCh

pkg/controller/controllercmd/builder.go
```diff
@@ -70,6 +70,10 @@ type ControllerBuilder struct {
 	authenticationConfig *operatorv1alpha1.DelegatedAuthentication
 	authorizationConfig  *operatorv1alpha1.DelegatedAuthorization
 	healthChecks         []healthz.HealthChecker
+
+	// nonZeroExitFn exists for unit testing and should be not used outside unit tests
+	// The signature of this function matches klog.Fatalf.
+	nonZeroExitFn func(args ...interface{})
```
sttts: klog.Fatalf has a string as first arg, no?

pkg/controller/controllercmd/builder.go
```diff
@@ -78,6 +82,7 @@ func NewController(componentName string, startFunc StartFunc) *ControllerBuilder
 		startFunc:        startFunc,
 		componentName:    componentName,
 		observerInterval: defaultObserverInterval,
+		nonZeroExitFn:    klog.Fatal,
```
sttts: doesn't `klog.Fatal` panic? We don't want that.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +243,44 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+			defer close(stoppedCh)
+			if err := b.startFunc(ctx, controllerContext); err != nil {
+				b.nonZeroExitFn(fmt.Sprintf("graceful termination failed, controllers failed with error: %v", err))
+			}
+		}()
+
+		// when the stoppedCh is closed before the context is cancelled, it means the controllers
+		// exited before we asked them to do so.
+		select {
+		case <-ctx.Done():
```
sttts: this is still wrong.

To make it right, wait for `stoppedCh` first. Then check that `ctx.Done()` is closed. But in two statements.

pkg/controller/controllercmd/builder_test.go
sttts: stoppedCh

pkg/controller/controllercmd/builder_test.go
```diff
@@ -0,0 +1,132 @@
...
+		b.getonStartedLeadingFunc(&ControllerContext{}, 10*time.Second)(ctx)
+	}()
+
+	select {
+	case <-nonZeroExitCh:
+		t.Fatal("unexpected non-zero shutdown")
+	case <-stopCh:
+	case <-time.After(5 * time.Second):
+		t.Fatal("unexpected timeout while terminating")
+	}
```
sttts: as above, this is non-deterministic.

pkg/controller/controllercmd/builder_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		startFunc: func(ctx context.Context, controllerContext *ControllerContext) error {
+			defer close(startedCh)
+			return fmt.Errorf("controller failed")
+		},
+	}
+
+	go func() {
+		defer close(stopCh)
+		b.getonStartedLeadingFunc(&ControllerContext{}, 10*time.Second)(ctx)
```
sttts: getonStartedLeadingFunc has wrong case

pkg/controller/controllercmd/builder_test.go
```diff
@@ -0,0 +1,132 @@
...
+			t.Logf("non-zero exit detected: %+v", args)
+		},
+		startFunc: func(ctx context.Context, controllerContext *ControllerContext) error {
+			defer close(startedCh)
+			return fmt.Errorf("controller failed")
+		},
+	}
+
+	go func() {
+		defer close(stopCh)
```
sttts: stoppedCh

pkg/controller/controllercmd/builder.go
sttts: It looks wrong. Why do you have the outer one at all?

pkg/controller/controllercmd/builder.go
mfojtik: no :-)

mfojtik: "Fatal" does not.

pkg/controller/controllercmd/builder.go
mfojtik: it exit(1) which is what we want and what we use today

pkg/controller/controllercmd/builder.go
sttts: you comment is wrong

pkg/controller/controllercmd/builder.go
sttts: It writes the stacktrace, checked the code.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +243,39 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+			defer close(stoppedCh)
+			if err := b.startFunc(ctx, controllerContext); err != nil {
+				b.nonZeroExitFn(fmt.Sprintf("graceful termination failed, controllers failed with error: %v", err))
+			}
+		}()
+
+		select {
+		case <-ctx.Done(): // context closed means the process received SIGINT and controllers should terminate
+		case <-stoppedCh: // closed stoppedCh at this point means premature controllers exit, without error if context was not cancelled
+			if ctx.Err() == nil {
```
sttts: does not match comment

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +243,39 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+		stoppedCh := make(chan struct{})
+		go func() {
+			defer close(stoppedCh)
+			if err := b.startFunc(ctx, controllerContext); err != nil {
+				b.nonZeroExitFn(fmt.Sprintf("graceful termination failed, controllers failed with error: %v", err))
+			}
+		}()
+
+		select {
+		case <-ctx.Done(): // context closed means the process received SIGINT and controllers should terminate
```
sttts: nit: sigint is an implementation detail and not guaranteed here. We only know the context closed, potentially because of sigint.

pkg/controller/controllercmd/builder.go
```diff
@@ -238,13 +243,39 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+		case <-stoppedCh: // closed stoppedCh at this point means premature controllers exit, without error if context was not cancelled
+			if ctx.Err() == nil {
+				b.nonZeroExitFn("graceful termination failed, controllers terminated prematurely")
+			}
+		}
+
+		select {
+		case <-time.After(gracefulTerminationDuration): // when context was closed above, give controllers 10s to terminate gracefully
```
sttts: nit: don't repeat hard-coded constants from somewhere else here. In this func it's arbitrary.

pkg/controller/controllercmd/builder_test.go
```diff
@@ -0,0 +1,184 @@
...
+		defer cancel()
+		time.Sleep(1 * time.Second)
+	}()
+	b.startFunc = func(ctx context.Context, controllerContext *ControllerContext) error {
+		time.Sleep(2 * time.Second)
+		return nil
+	}
+	b.getonStartedLeadingFunc(&ControllerContext{}, 5*time.Second)(ctx)
+	if len(fatals) > 0 {
+		t.Errorf("expected controllers to exit prematurely, got %#v", fatals)
```
sttts: not prematurely

pkg/controller/controllercmd/builder.go
mfojtik: fixed

pkg/controller/controllercmd/builder.go
mfojtik: i replaced this by klog.Warning() followed by os.Exit(1).

PR #700, opened 2020-02-05
factory: fix time based resync to use queue instead of calling sync directly

mfojtik:
This prevents factory controllers to run two sync()'s in parallel. Instead the time based resyncs are now handled via queue.

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -3,6 +3,7 @@ package factory
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -77,11 +78,14 @@ func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
 	return c.sync(ctx, syncCtx)
 }
 
+// QueueKey return queue key for given name.
+func QueueKey(name string) string {
+	return strings.ToLower(name) + "Key"
+}
+
 func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.Duration) {
 	go wait.UntilWithContext(ctx, func(ctx context.Context) {
-		if err := c.sync(ctx, c.syncContext); err != nil {
-			utilruntime.HandleError(fmt.Errorf("periodical resync of controller %s failed: %v", c.name, err))
-		}
+		c.syncContext.Queue().Add(QueueKey(c.name))
 	}, interval)
 }
 


pkg/controller/factory/base_controller.go
```diff
@@ -77,11 +78,14 @@ func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
...
+func QueueKey(name string) string {
+	return strings.ToLower(name) + "Key"
+}
+
 func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.Duration) {
 	go wait.UntilWithContext(ctx, func(ctx context.Context) {
-		if err := c.sync(ctx, c.syncContext); err != nil {
-			utilruntime.HandleError(fmt.Errorf("periodical resync of controller %s failed: %v", c.name, err))
-		}
+		c.syncContext.Queue().Add(QueueKey(c.name))
```
sttts: how is this distinct from the resync capabilities of informers.

tnozicka: funny, I asked the same question on slack

mfojtik: @sttts in many controllers we do manual time based resync... I think @deads2k asked for this at some point to make sure that when informers get stucked we still run.

tnozicka: The timer used to start after the snyc finished so it was sync+1s, now its every 1s, but I don't see an issue with that.

Using the queue prevents races.

mfojtik: @tnozicka if we use queue, we do time based sync, we utilize queue (so it show in metrics) and we don't race with informer. i think, this is a good change.

deads2k:
informer resyncs replay the current state of the objects.  This means 
 1. you cannot optimized a zero-diff to no update
 2. you are required to have informers
 3. even if you have informers, you must be sure that they have contents
 4. informers with different resync times result in many notifications
 5. you aren't guaranteed to avoid numerous queuing for every configmap for instance

The time based trigger resolves all of these issues.  It is a different mechanism for "single item queue" controllers as a sub-class of other controllers.

pkg/controller/factory/base_controller.go
```diff
@@ -77,11 +78,14 @@ func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
 	return c.sync(ctx, syncCtx)
 }
 
+// QueueKey return queue key for given name.
+func QueueKey(name string) string {
+	return strings.ToLower(name) + "Key"
```
deads2k: this looks weird.  Why do the other queuing mechanisms not require this?

mfojtik: we usually make `const key` and use it... in factory, we have to construct it... I will need this to make resourcesynccontroller factory based as it need to access the queue from SyncSecret/ConfigMap that we export from controller...

deads2k: > we usually make `const key` and use it... in factory, we have to construct it... I will need this to make resourcesynccontroller factory based as it need to access the queue from SyncSecret/ConfigMap that we export from controller...

the issue isn't the exposure.  The issue is that this should be re-used everywhere, right?

deads2k:
/approve

I'm ok with taking fixes later.

PR #701, opened 2020-02-05
add methods to manipulation observed configuration values

deads2k:
This is straight promotion out of the kube-apiserver operator.

### DIFF
--- a/pkg/operator/resource/resourcemerge/generic_config_merger.go
+++ b/pkg/operator/resource/resourcemerge/generic_config_merger.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"k8s.io/klog"
 	"sigs.k8s.io/yaml"
@@ -133,6 +134,13 @@ func MergePrunedProcessConfig(schema runtime.Object, specialCases map[string]Mer
 
 type MergeFunc func(dst, src interface{}, currentPath string) (interface{}, error)
 
+var _ MergeFunc = RemoveConfig
+
+// RemoveConfig is a merge func that elimintes an entire path from the config
+func RemoveConfig(dst, src interface{}, currentPath string) (interface{}, error) {
+	return dst, nil
+}
+
 // mergeConfig overwrites entries in curr by additional.  It modifies curr.
 func mergeConfig(curr, additional map[string]interface{}, currentPath string, specialCases map[string]MergeFunc) error {
 	for additionalKey, additionalVal := range additional {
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourcemerge/generic_config_merger.go
```diff
@@ -133,6 +134,13 @@ func MergePrunedProcessConfig(schema runtime.Object, specialCases map[string]Mer
 
 type MergeFunc func(dst, src interface{}, currentPath string) (interface{}, error)
 
+var _ MergeFunc = RemoveConfig
+
+// RemoveConfig is a merge func that elimintes an entire path from the config
+func RemoveConfig(dst, src interface{}, currentPath string) (interface{}, error) {
```
sttts: path is misleading. I thought you follow the path and cut it out.

pkg/operator/resource/resourcemerge/generic_config_merger.go
```diff
@@ -228,3 +236,36 @@ func intersectValue(x1, x2 interface{}) interface{} {
...
+		if err != nil {
+			return fmt.Errorf("error reading %v from config, %v", strings.Join(requiredPath, "."), err)
+		}
+		if !found {
+			return fmt.Errorf("%v missing from config", strings.Join(requiredPath, "."))
+		}
+		if configVal == nil {
+			return fmt.Errorf("%v null in config", strings.Join(requiredPath, "."))
+		}
+		if configValSlice, ok := configVal.([]interface{}); ok && len(configValSlice) == 0 {
```
sttts: the last two cases don't match "Present" IMO. Then better call it `IsRequiredConfigNonEmpty`.

pkg/operator/resource/resourcemerge/generic_config_merger.go
deads2k: See type assertion above.  That is the signature of the method.  See usage here; https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/targetconfigcontroller/targetconfigcontroller.go#L223

pkg/operator/resource/resourcemerge/generic_config_merger.go
deads2k: > the last two cases don't match "Present" IMO. Then better call it `IsRequiredConfigNonEmpty`.

That's what the word present means.

pkg/operator/resource/resourcemerge/generic_config_merger.go
```diff
@@ -228,3 +236,36 @@ func intersectValue(x1, x2 interface{}) interface{} {
...
+
+	existingConfig := map[string]interface{}{}
+	if err := json.NewDecoder(bytes.NewBuffer(config)).Decode(&existingConfig); err != nil {
+		return fmt.Errorf("error parsing config, %v", err)
+	}
+
+	for _, requiredPath := range requiredPaths {
+		configVal, found, err := unstructured.NestedFieldNoCopy(existingConfig, requiredPath...)
+		if err != nil {
+			return fmt.Errorf("error reading %v from config, %v", strings.Join(requiredPath, "."), err)
```
mfojtik:
```golang
requiredPath = strings.Join(requiredPath, ".")
```

instead of repeating the strings.Join() 5x ?

mfojtik:
/lgtm

just nit, hope you fix it in next library-go PR :)

PR #703, opened 2020-02-05
Bug 1792501: Increase revision pruner pod memory request

damemi:
Originally added in https://github.com/openshift/library-go/pull/683, we are still seeing revision pruners get OOMKilled, so attempting to increase the memory request again

### DIFF
--- a/pkg/operator/staticpod/controller/prune/bindata/bindata.go
+++ b/pkg/operator/staticpod/controller/prune/bindata/bindata.go
@@ -63,9 +63,9 @@ spec:
     imagePullPolicy: IfNotPresent
     resources:
       requests:
-        memory: 100M
+        memory: 200M
       limits:
-        memory: 100M
+        memory: 200M
     securityContext:
       privileged: true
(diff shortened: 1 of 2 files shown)


damemi:
To follow up with this, I did some investigating on a longer-running cluster. In the operator I looked at (KSO), there was only 1 revision pruner pod (out of 8) that was OOMKilled. That pod was only using 40M at its peak (with request & limit set to 100M). So I'm wondering if increasing memory requests will actually fix this, or if in some cases our revision pruner is just the unlucky pod that got dropped on an otherwise full cluster. 

We can give it another bump, but it isn't essential that no revision pruners ever get OOMKilled (because the next one will clean up any work from the previous one to get killed). And we don't want to request excessive memory to push out other components that are actually critical. This may just be a case where we're chasing the wrong flake

damemi:
Actually, https://github.com/openshift/library-go/pull/707 seems to have more information. Missing CPU limits might have been the case here.

PR #706, opened 2020-02-06
bug 1800321: add substitution for NODE_NAME inside of static pods

deads2k:
by substituting the NODE_NAME, we can select which env vars and arguments we want to use in commands.  This allows a single static pod template to be used on nodes that must have different arguments

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -32,6 +32,7 @@ type InstallOptions struct {
 	KubeClient kubernetes.Interface
 
 	Revision  string
+	NodeName  string
 	Namespace string
 
 	PodConfigMapNamePrefix        string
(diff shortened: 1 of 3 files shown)


deads2k:
/hold 
for proof

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -294,7 +302,9 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 		if !exists {
 			return true, fmt.Errorf("required 'pod.yaml' key does not exist in configmap")
 		}
-		podContent = strings.Replace(podData, "REVISION", o.Revision, -1)
+		podContent = strings.ReplaceAll(podData, "REVISION", o.Revision)
+		podContent = strings.ReplaceAll(podContent, "NODE_NAME", o.NodeName)
+		podContent = strings.ReplaceAll(podContent, "NODE_ENVVAR_NAME", strings.ReplaceAll(strings.ReplaceAll(o.NodeName, "-", "_"), ".", "_"))
```
mfojtik: where is this used?

mfojtik:
/lgtm

unhold when proof go green

PR #709, opened 2020-02-13
Bug 1779438: fix race for controller commands

stlaz:
Scenario:
a) AddDefaultRotationToConfig():
1. no serving cert and key found in expected directory
2. creates self-signed serving cert/key pair
3. serving-cert key-pair appears in the expected directory

b) WithRestartOnChange() -> AddReactor()

4. does not see starting content for the observed files
5. hashes for the expected cert/key pair are counted from the
   actual files but the controller will keep serving with the
   self-signed cert/key pair now

This was most probably caused by the line
```
    certDir = temporaryCertDir
```
which caused that starting content was added for wrong two files.

### DIFF
--- a/pkg/controller/controllercmd/cmd.go
+++ b/pkg/controller/controllercmd/cmd.go
@@ -1,7 +1,6 @@
 package controllercmd
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


stlaz:
Another BZ with the same diagnose - https://bugzilla.redhat.com/show_bug.cgi?id=1820385

stlaz:
Looks like the container itself is failing rather than the tests themselves but the CI does not appear to be very chatty about the failure reason.

PR #710, opened 2020-02-13
pkg/operator/status/condition: Say "All is well" with no interesting conditions

wking:
If none of the interesting conditions have anything to say, and nothing bad is happening, explicitly say "All is well".  This makes a clearer distinction between "nothing is wrong" and "some things are concerning, but inertia or sub-condition status are keeping me from actually going degraded yet".


wking:
#712 walked back the high Go requirement.

/retest

wking:
/remove-lifecycle stale

I still think we want this.  Also, rebased onto master (no conflicts, but it will make the graph history easier to read if/when this lands) with 245ffce78 -> 1d8b2ff84.

PR #711, opened 2020-02-14
pkg/operator/status: Use warning events for concerning status changes

wking:
And provide more context in the event messages, to make it easier to distinguish "we're slightly concerned, but not actually degraded" events from "we're actually sad" events.  For examples where I was distracted by the former, see [rhbz#1775970][1] and [rhbz#1779429][2].

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1775970
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1779429


wking:
Rebased onto master with 5be8b2968 -> 85802c172.

PR #715, opened 2020-02-20
Bug 1805398: Allow disabling serving in ControllerBuilder

tnozicka:
To avoid reaching the serving internal client for delegated auth config map, we need a way to disable it for recovery containers using special SNI endpoint

proof in https://github.com/openshift/cluster-kube-apiserver-operator/pull/768
/cc @deads2k @sttts @soltysh 

/assign @deads2k


tnozicka:
proof is green https://github.com/openshift/cluster-kube-apiserver-operator/pull/768

PR #718, opened 2020-02-25
WIP: Add simple prometheus query observer controller

mfojtik:
This introduces simple prometheus query result observer that can react to query results.

An example usage would be an etcd-operator that watch the fsync metrics (`histogram_quantile(0.99, rate(etcd_disk_wal_fsync_duration_seconds_bucket[5m]))`) and if the etcd criteria are not met with the value returned, the operator can go degraded or unavailable.

Each query handler is identified by name and that name is then translated to degraded condition which fire if the handler return error and it is cleared when the handler return nil.

### DIFF
--- a/pkg/metrics/client/client.go
+++ b/pkg/metrics/client/client.go
@@ -1,0 +1,102 @@
+package client
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/transport"
+
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+
+	routeclient "github.com/openshift/client-go/route/clientset/versioned"
+	routev1client "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+)
+
+// DEPRECATED: This function is left only for test library compatibility.
+func DeprecatedPrometheusClient(kubeClient *kubernetes.Clientset, routeClient *routeclient.Clientset) (prometheusv1.API, error) {
+	return NewPrometheusClient(kubeClient.CoreV1(), kubeClient.CoreV1(), kubeClient.CoreV1(), routeClient.RouteV1())
+}
+
+// NewPrometheusClient returns Prometheus API or error
+// Note: with thanos-querier you must pass an entire Alert as a query. Partial queries return an error, so have to pass the entire alert.
+// Example query for an Alert:
+// `ALERTS{alertname="PodDisruptionBudgetAtLimit",alertstate="pending",namespace="pdbnamespace",poddisruptionbudget="pdbname",prometheus="openshift-monitoring/k8s",service="kube-state-metrics",severity="warning"}==1`
+// Example query:
+// `scheduler_scheduling_duration_seconds_sum`
+func NewPrometheusClient(serviceClient corev1client.ServicesGetter, secretsClient corev1client.SecretsGetter, configsClient corev1client.ConfigMapsGetter, routeClient routev1client.RoutesGetter) (prometheusv1.API, error) {
+	_, err := serviceClient.Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	route, err := routeClient.Routes("openshift-monitoring").Get("thanos-querier", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	host := route.Status.Ingress[0].Host
+	var bearerToken string
+	secrets, err := secretsClient.Secrets("openshift-monitoring").List(metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not list secrets in openshift-monitoring namespace")
+	}
+	for _, s := range secrets.Items {
+		if s.Type != corev1.SecretTypeServiceAccountToken ||
+			!strings.HasPrefix(s.Name, "prometheus-k8s") {
+			continue
+		}
+		bearerToken = string(s.Data[corev1.ServiceAccountTokenKey])
+		break
+	}
+	if len(bearerToken) == 0 {
+		return nil, fmt.Errorf("prometheus service account not found")
+	}
+
+	return createClient(configsClient, host, bearerToken)
+}
+
+func createClient(configClient corev1client.ConfigMapsGetter, host, bearerToken string) (prometheusv1.API, error) {
+	// retrieve router CA
+	routerCAConfigMap, err := configClient.ConfigMaps("openshift-config-managed").Get("router-ca", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	bundlePEM := []byte(routerCAConfigMap.Data["ca-bundle.crt"])
+
+	// make a client connection configured with the provided bundle.
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM(bundlePEM)
+
+	// prometheus API client, configured for route host and bearer token auth
+	client, err := prometheusapi.NewClient(prometheusapi.Config{
+		Address: "https://" + net.JoinHostPort(host, "443"),
+		RoundTripper: transport.NewBearerAuthRoundTripper(
+			bearerToken,
+			&http.Transport{
+				Proxy: http.ProxyFromEnvironment,
+				DialContext: (&net.Dialer{
+					Timeout:   30 * time.Second,
+					KeepAlive: 30 * time.Second,
+				}).DialContext,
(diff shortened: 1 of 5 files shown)


pkg/metrics/observer/query_observer.go
```diff
@@ -0,0 +1,74 @@
...
+// NewPrometheusMetricObserver allows to continuously observe particular Prometheus query result value and react to it.
+// The name should be unique name for this observer.
+// The queryHandlers represents list of Prometheus queries and handlers where the result value and type are passed.
+// The clients are needed to initialize that prometheus API client.
+func NewPrometheusMetricObserver(name string, queryHandlers []QueryHandler, serviceClient corev1client.ServicesGetter, secretsClient corev1client.SecretsGetter, configsClient corev1client.ConfigMapsGetter,
+	routeClient routev1client.RoutesGetter, recorder events.Recorder) (factory.Controller, error) {
+	observer := &prometheusMetricObserver{
+		queryHandlers: queryHandlers,
+	}
+	prometheusClient, err := metricsclient.NewPrometheusClient(serviceClient, secretsClient, configsClient, routeClient)
```
mfojtik: @deads2k I assume the secrets and other material might rotate and we would need to recreate this client... however I guess making this client every 30s for every query handler could be expensive (we can use the cached clients maybe?)

pkg/metrics/observer/query_observer.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (o *prometheusMetricObserver) sync(ctx context.Context, controllerContext factory.SyncContext) error {
+	var errors []error
+	for _, handler := range o.queryHandlers {
+		result, warnings, err := o.prometheusClient.QueryRange(ctx, handler.Query, prometheusv1.Range{
+			Start: time.Now().Add(-defaultPollingDuration),
+			End:   time.Now(),
+		})
+		if err != nil {
```
mfojtik: @deads2k @hexfusion i wonder if we should set degraded conditions here if we are unable to query prometheus for metrics (this might cause some churn during bootstrap?)

mfojtik:
/cc @lilic 
/cc @brancz 

(for visibility as this is in some ways overlapping with alerts, except this is changing the operator degraded status directly)

pkg/metrics/client/client.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	bundlePEM := []byte(routerCAConfigMap.Data["ca-bundle.crt"])
+
+	// make a client connection configured with the provided bundle.
+	roots := x509.NewCertPool()
+	roots.AppendCertsFromPEM(bundlePEM)
+
+	// prometheus API client, configured for route host and bearer token auth
+	client, err := prometheusapi.NewClient(prometheusapi.Config{
+		Address: "https://" + host,
```
mfojtik: @deads2k @russellb will this break in ipv6 ?

russellb: If `host` is ever an IPv6 address, then yes, it will break.

The safe way to handle this would be something like

`"https://" + net.JoinHostPort(host, "443")`

russellb: ^^^ will ensure that an IPv6 address gets wrapped in square brackets.

mfojtik: @russellb this will be a host name (route host name)

mfojtik:
@lilic we want this to communicate to cluster admins that the etcd (just one case) won't perform well on their cluster because their disks are slow, so the operator will report "Degraded=true" if the number returned from promquery is low/high. We could do this via alerts, however in some cases (like etcd) it is reasonable to stop the upgrade when etcd is underwater performance wise.

lilic:
Okay thanks!

> We could do this via alerts, however in some cases (like etcd) it is reasonable to stop the upgrade when etcd is underwater performance wise.

I know the discussion should probably go outside of this PR, but why can't we say if critical alerts are firing we cannot perform upgrades instead. 

As compromise could we have both? :)

mfojtik:
> I know the discussion should probably go outside of this PR, but why can't we say if critical alerts are firing we cannot perform upgrades instead.

I think this is reasonable, but it will probably require CVO to watch the alerts?

lilic:
> I think this is reasonable, but it will probably require CVO to watch the alerts?

Yes, the change would be in CVO, and all teams would just do alerts in critical or some other severity. But thats outside of my paygrade to decide on this. :D

pkg/metrics/client/client.go
russellb: OK, then no problems.

brancz:
I do think this should be alert driven and the particular example that you gave etcd even already has alerts for, I know this, because I wrote them :) 

I do think operators setting their degraded status based on whether critical alerts are firing makes a lot of sense. It’s definitely better than operators being bespoke monitoring systems. We also played with the idea of marking upgradable false for the cluster monitoring operator, if _any_ critical alert is firing, although arguably I do think CVO should somehow gate this.

pkg/metrics/client/client.go
```diff
@@ -0,0 +1,102 @@
...
+	_, err := serviceClient.Services("openshift-monitoring").Get("prometheus-k8s", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	route, err := routeClient.Routes("openshift-monitoring").Get("thanos-querier", metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	host := route.Status.Ingress[0].Host
```
deads2k: this is unfortunate. Can we avoid using network-edge to get this?  For the metrics we want, can we go direct to the etcd metrics endpoint?  it tracks an instantaneous p99, right?

pkg/metrics/observer/query_observer.go
```diff
@@ -0,0 +1,141 @@
...
+				Reason:  "PrometheusUnavailable",
+				Message: fmt.Sprintf("Prometheus client is not available: %v", prometheusClientErr),
+			})
+			continue
+		}
+
+		// pass the polling interval to handler so callers know how often we execute this handler
+		handlerCtx := context.WithValue(ctx, "polling_interval", defaultPollingDuration)
+
+		if err := handler.Handler(handlerCtx, controllerContext.Recorder(), prometheusClient); err != nil {
```
deads2k: one handler per query or one handler per controller?  I guess higher order logic could be created with coordinated handlers and we're likely to use this sparingly enough that we only go degraded for individually named metrics.  `EtcdDiskPerformanceDegraded` as opposed to `EtcdMetricsDegraded`.  or `KubeAPIServerFailoverCapacityDegraded` and not `KubeAPIServerMetricsDegraded`.  

Now that I've typed the examples, this seems fine.

deads2k:
> I do think this should be alert driven and the particular example that you gave etcd even already has alerts for, I know this, because I wrote them :)
> 
> I do think operators setting their degraded status based on whether critical alerts are firing makes a lot of sense. It’s definitely better than operators being bespoke monitoring systems. We also played with the idea of marking upgradable false for the cluster monitoring operator, if _any_ critical alert is firing, although arguably I do think CVO should somehow gate this.

@brancz @lilic We cannot prevent upgrades when alerts are firing for the same reason that we cannot prevent upgrades on available=false or degraded=true.  We have to able to upgrade past a bad openshift release.  If we have a bug in an operator that causes these conditions, we need to be able to upgrade past that situation, so we cannot block upgrades when the current state is bad.

pkg/metrics/client/client.go
```diff
@@ -0,0 +1,102 @@
...
+	return NewPrometheusClient(kubeClient.CoreV1(), kubeClient.CoreV1(), kubeClient.CoreV1(), routeClient.RouteV1())
+}
+
+// NewPrometheusClient returns Prometheus API or error
+// Note: with thanos-querier you must pass an entire Alert as a query. Partial queries return an error, so have to pass the entire alert.
+// Example query for an Alert:
+// `ALERTS{alertname="PodDisruptionBudgetAtLimit",alertstate="pending",namespace="pdbnamespace",poddisruptionbudget="pdbname",prometheus="openshift-monitoring/k8s",service="kube-state-metrics",severity="warning"}==1`
+// Example query:
+// `scheduler_scheduling_duration_seconds_sum`
+func NewPrometheusClient(serviceClient corev1client.ServicesGetter, secretsClient corev1client.SecretsGetter, configsClient corev1client.ConfigMapsGetter, routeClient routev1client.RoutesGetter) (prometheusv1.API, error) {
```
deads2k: @sallyom @soltysh and this is why getting the logic safe to being with mattered :)

deads2k:
> @deads2k I assume the secrets and other material might rotate and we would need to recreate this client... however I guess making this client every 30s for every query handler could be expensive (we can use the cached clients maybe?)

How about only rebuilding the client when the overall controller experiences is an error. This would limit you to only error cases, which are already likely to be slow and expensive.

brancz:
I feel like I’ve read about forcing upgrades somehow even if upgradable is set to false. Either way even if not setting upgradable to false I’d prefer not to make individual operators bespoke monitoring systems, but define alerts and see if they are firing much like done here.

pkg/metrics/client/client.go
brancz: I don’t believe this is necessary. Just request the Thanos query service directly.

hexfusion:
Overall this looks great thanks @mfojtik

brancz:
I still think doing this via alerts makes the most sense. Alerts can have variable/configurable thresholds by making the thresholds recording rules themselves.

mfojtik:
we don't need this here anymore.

PR #719, opened 2020-02-26
DEVEXP-515: Added support for RequestModifiers.

ricardomaraschini:
Added support for user provided `transport.RequestModifiers` on registry client. I found out that the code here is duplicated into two other repositories, each one of the with a slightly different feature set.

With this PR we can get rid of the duplicated code on [image-registry](https://github.com/openshift/image-registry/tree/master/pkg/origin-common/image/registryclient).

### DIFF
--- a/pkg/image/registryclient/client.go
+++ b/pkg/image/registryclient/client.go
@@ -47,14 +47,15 @@ func (e *ErrNotV2Registry) Error() string {
 type AuthHandlersFunc func(transport http.RoundTripper, registry *url.URL, repoName string) []auth.AuthenticationHandler
 
 // NewContext is capable of creating RepositoryRetrievers.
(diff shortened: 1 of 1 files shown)


pkg/image/registryclient/client.go
```diff
@@ -47,14 +47,15 @@ func (e *ErrNotV2Registry) Error() string {
...
+func NewContext(transp, insecureTransport http.RoundTripper) *Context {
 	return &Context{
-		Transport:         transport,
+		Transport:         transp,
 		InsecureTransport: insecureTransport,
 		Challenges:        challenge.NewSimpleManager(),
 		Actions:           []string{"pull"},
 		Retries:           2,
 		Credentials:       NoCredentials,
+		RequestModifiers:  make([]transport.RequestModifier, 0),
```
dmage: You may avoid memory allocation and use nil here

PR #727, opened 2020-03-06
resourceapply: Allow removing run-level labels from namespaces

stlaz:
A couple of components were using run-level labels on their
namespaces even though they did not have to. Make it possible
to remove this label with the library-go functions while not
changing the other labels/annotations in case these are maintained
automatically by a different controller.

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -22,8 +22,10 @@ import (
 func ApplyNamespace(client coreclientv1.NamespacesGetter, recorder events.Recorder, required *corev1.Namespace) (*corev1.Namespace, bool, error) {
 	existing, err := client.Namespaces().Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		actual, err := client.Namespaces().Create(required)
-		reportCreateEvent(recorder, required, err)
+		requiredCopy := required.DeepCopy()
+		actual, err := client.Namespaces().
+			Create(resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*corev1.Namespace))
+		reportCreateEvent(recorder, requiredCopy, err)
 		return actual, true, err
 	}
 	if err != nil {
@@ -53,8 +55,10 @@ func ApplyNamespace(client coreclientv1.NamespacesGetter, recorder events.Record
 func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder, required *corev1.Service) (*corev1.Service, bool, error) {
 	existing, err := client.Services(required.Namespace).Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		actual, err := client.Services(required.Namespace).Create(required)
-		reportCreateEvent(recorder, required, err)
+		requiredCopy := required.DeepCopy()
+		actual, err := client.Services(requiredCopy.Namespace).
+			Create(resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*corev1.Service))
+		reportCreateEvent(recorder, requiredCopy, err)
 		return actual, true, err
 	}
 	if err != nil {
@@ -94,8 +98,10 @@ func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder, 
 func ApplyPod(client coreclientv1.PodsGetter, recorder events.Recorder, required *corev1.Pod) (*corev1.Pod, bool, error) {
 	existing, err := client.Pods(required.Namespace).Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		actual, err := client.Pods(required.Namespace).Create(required)
-		reportCreateEvent(recorder, required, err)
+		requiredCopy := required.DeepCopy()
+		actual, err := client.Pods(requiredCopy.Namespace).
+			Create(resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*corev1.Pod))
+		reportCreateEvent(recorder, requiredCopy, err)
 		return actual, true, err
 	}
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -34,6 +34,14 @@ func ApplyNamespace(client coreclientv1.NamespacesGetter, recorder events.Record
 	existingCopy := existing.DeepCopy()
 
 	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+
+	// allow removing runlevels since the metadata are otherwise merged
+	_, requiredRunLevelPresent := required.ObjectMeta.Labels["run-level"]
+	if _, ok := existingCopy.ObjectMeta.Labels["run-level"]; ok != requiredRunLevelPresent && !requiredRunLevelPresent {
```
sttts: the add case works?

`ok && !requiredRunLevelPresent` is simpler

stlaz: The add case is handled by the `EnsureObjectMeta()` function above.

Good simplification, I modified the condition in latest revision.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -150,6 +154,7 @@ func ApplyServiceAccount(client coreclientv1.ServiceAccountsGetter, recorder eve
 func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, required *corev1.ConfigMap) (*corev1.ConfigMap, bool, error) {
 	existing, err := client.ConfigMaps(required.Namespace).Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
+		resourcemerge.CleanMetaLabelsAnnotations(&required.ObjectMeta)
```
sttts: mutating the input is bad. We need a deep copy (or shallow copy of at least the root and ObjectMeta).

stlaz: I went with the DeepCopy() should we ever want to mutate any other fields of the actual object as well

sttts: performance here does not matter anyway

pkg/operator/resource/resourceapply/core.go
```diff
@@ -22,8 +22,10 @@ import (
 func ApplyNamespace(client coreclientv1.NamespacesGetter, recorder events.Recorder, required *corev1.Namespace) (*corev1.Namespace, bool, error) {
 	existing, err := client.Namespaces().Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		actual, err := client.Namespaces().Create(required)
-		reportCreateEvent(recorder, required, err)
+		requiredCopy := required.DeepCopy()
+		resourcemerge.CleanMetaLabelsAnnotations(&requiredCopy.ObjectMeta)
```
sttts: how about:
```
actual, err := client.Namespaces().Create(WithCleanLabelsAndAnnotations(requiredCopy))`
```

stlaz: done

pkg/operator/resource/resourceapply/core.go
```diff
@@ -22,8 +22,10 @@ import (
 func ApplyNamespace(client coreclientv1.NamespacesGetter, recorder events.Recorder, required *corev1.Namespace) (*corev1.Namespace, bool, error) {
 	existing, err := client.Namespaces().Get(required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		actual, err := client.Namespaces().Create(required)
-		reportCreateEvent(recorder, required, err)
+		requiredCopy := required.DeepCopy()
+		actual, err := client.Namespaces().
+			Create(resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*corev1.Namespace))
+		reportCreateEvent(recorder, requiredCopy, err)
```
sttts: report requiredCopy or actual or cleaned version?

pkg/operator/resource/resourcemerge/object_merger.go
```diff
@@ -124,7 +142,13 @@ func MergeMap(modified *bool, existing *map[string]string, required map[string]s
 	for k, v := range required {
 		if existingV, ok := (*existing)[k]; !ok || v != existingV {
 			*modified = true
-			(*existing)[k] = v
+			// if "required" map contains a key with "-" as suffix, remove that
+			// key from the existing map instead of replacing the value
+			if strings.HasSuffix(k, "-") {
+				delete(*existing, strings.TrimRight(k, "-"))
```
sttts: Intentional to remove arbitrary man `-` here?

stlaz: yes, you need to remove the trailing `-` to get the actual name of the field you're trying to remove:
```
original: run-level: 0
required: run-level-:
```
This is different in case of https://github.com/openshift/library-go/pull/727/files#diff-3567eb2cd10c84b30387be04408ab7b2R27 where the object being cleaned is the object that's actually getting created.

pkg/operator/resource/resourceapply/core.go
stlaz: `actual` could be `nil` here in case of an error, so I guess the cleaned version should be reported. In C I would be sure that `requiredCopy == cleaned`, not sure if Go does additional magic to the pointer being passed to `WithCleanLabelsAndAnnotations` :thinking:

stlaz: Yes, they are equal

pkg/operator/resource/resourcemerge/object_merger.go
sttts: I meant that `TrimRight` also removes `---` from `foo----`.

pkg/operator/resource/resourceapply/core.go
sttts: I don't care much which one you print, just decide conciously :)

PR #731, opened 2020-03-09
Migrate controllers to factory

mfojtik:
This moves more controllers to controller factory.
It also makes one more extension point in factory, which is `PostRunHook`. This allows to register custom triggers or other asynchronous code to run **after** the controller us run.

### DIFF
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -3,6 +3,7 @@ package leaderelection
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"strings"
 	"time"
 
@@ -64,7 +65,8 @@ func ToConfigMapLeaderElection(clientConfig *rest.Config, config configv1.Leader
 		RetryPeriod:     config.RetryPeriod.Duration,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStoppedLeading: func() {
-				klog.Fatalf("leaderelection lost")
+				defer os.Exit(0)
+				klog.Warningf("leader election lost")
 			},
 		},
 	}, nil
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -3,11 +3,9 @@ package factory
 import (
 	"context"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -16,13 +14,12 @@ import (
 
 // baseController represents generic Kubernetes controller boiler-plate
 type baseController struct {
-	name         string
-	cachesToSync []cache.InformerSynced
-	sync         func(ctx context.Context, controllerContext SyncContext) error
-	syncContext  SyncContext
-	resyncEvery  time.Duration
-
-	postRunHooks []PostStartHook
+	name           string
+	cachesToSync   []cache.InformerSynced
+	sync           func(ctx context.Context, controllerContext SyncContext) error
+	syncContext    SyncContext
+	resyncEvery    time.Duration
+	postStartHooks []PostStartHook
 }
 
 var _ Controller = &baseController{}
@@ -68,19 +65,19 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 		}()
 	}
 
-	// run post-run hooks (custom triggers, etc.)
-	if len(c.postRunHooks) > 0 {
+	// run post-start hooks (custom triggers, etc.)
+	if len(c.postStartHooks) > 0 {
 		var hookWg sync.WaitGroup
 		defer func() {
-			defer klog.Infof("All %s post-run hooks have been terminated", c.name)
-			hookWg.Wait() // wait for the post-run hooks
+			hookWg.Wait() // wait for the post-start hooks
+			klog.Infof("All %s post start hooks have been terminated", c.name)
 		}()
-		for i := range c.postRunHooks {
+		for i := range c.postStartHooks {
 			hookWg.Add(1)
 			go func(index int) {
 				defer hookWg.Done()
-				if err := c.postRunHooks[index](ctx, c.syncContext); err != nil {
-					klog.Warningf("%s controller post run hook error: %v", c.name, err)
+				if err := c.postStartHooks[index](ctx, c.syncContext); err != nil {
+					klog.Warningf("%s controller post start hook error: %v", c.name, err)
 				}
 			}(i)
 		}
@@ -102,17 +99,12 @@ func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
 	return c.sync(ctx, syncCtx)
 }
 
-// QueueKey return queue key for given name.
-func QueueKey(name string) string {
-	return strings.ToLower(name) + "Key"
-}
-
 func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.Duration) {
 	if interval == 0 {
 		return
 	}
 	go wait.UntilWithContext(ctx, func(ctx context.Context) {
-		c.syncContext.Queue().Add(QueueKey(c.name))
+		c.syncContext.Queue().Add(DefaultQueueKey)
 	}, interval)
 }
 
@@ -137,18 +129,24 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 }
 
 func (c *baseController) processNextWorkItem(queueCtx context.Context) {
-	syncObject, quit := c.syncContext.Queue().Get()
+	key, quit := c.syncContext.Queue().Get()
 	if quit {
 		return
 	}
-	defer c.syncContext.Queue().Done(syncObject)
+	defer c.syncContext.Queue().Done(key)
 
-	runtimeObject, _ := syncObject.(runtime.Object)
-	if err := c.sync(queueCtx, c.syncContext.(syncContext).withRuntimeObject(runtimeObject)); err != nil {
-		utilruntime.HandleError(fmt.Errorf("%s controller failed to sync %+v with: %w", c.name, syncObject, err))
-		c.syncContext.Queue().AddRateLimited(syncObject)
+	syncCtx := c.syncContext.(syncContext)
+	var ok bool
+	syncCtx.queueKey, ok = key.(string)
+	if !ok {
+		klog.Warningf("Queue key is not string: %+v", key)
+	}
+
+	if err := c.sync(queueCtx, syncCtx); err != nil {
+		utilruntime.HandleError(fmt.Errorf("%s controller failed to sync %+v with: %w", c.name, key, err))
+		c.syncContext.Queue().AddRateLimited(key)
 		return
 	}
 
-	c.syncContext.Queue().Forget(syncObject)
+	c.syncContext.Queue().Forget(key)
 }
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -21,6 +21,8 @@ type baseController struct {
 	sync         func(ctx context.Context, controllerContext SyncContext) error
 	syncContext  SyncContext
 	resyncEvery  time.Duration
+
+	postRunHooks []PostStartHook
 }
 
 var _ Controller = &baseController{}
@@ -36,10 +38,10 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 		panic("timeout waiting for informer cache") // this will be recovered using HandleCrash()
 	}
 
-	var workerWaitGroup sync.WaitGroup
+	var workerWg sync.WaitGroup
 	defer func() {
 		defer klog.Infof("All %s workers have been terminated", c.name)
-		workerWaitGroup.Wait()
+		workerWg.Wait()
 	}()
 
 	// queueContext is used to track and initiate queue shutdown
@@ -47,11 +49,11 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 
 	for i := 1; i <= workers; i++ {
 		klog.Infof("Starting #%d worker of %s controller ...", i, c.name)
-		workerWaitGroup.Add(1)
+		workerWg.Add(1)
 		go func() {
 			defer func() {
 				klog.Infof("Shutting down worker of %s controller ...", c.name)
-				workerWaitGroup.Done()
+				workerWg.Done()
 			}()
 			c.runWorker(queueContext)
 		}()
@@ -59,11 +61,29 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 
 	// runPeriodicalResync is independent from queue
 	if c.resyncEvery > 0 {
-		workerWaitGroup.Add(1)
+		workerWg.Add(1)
 		go func() {
-			defer workerWaitGroup.Done()
+			defer workerWg.Done()
 			c.runPeriodicalResync(ctx, c.resyncEvery)
 		}()
+	}
+
+	// run post-run hooks (custom triggers, etc.)
+	if len(c.postRunHooks) > 0 {
+		var hookWg sync.WaitGroup
+		defer func() {
+			defer klog.Infof("All %s post-run hooks have been terminated", c.name)
+			hookWg.Wait() // wait for the post-run hooks
+		}()
+		for i := range c.postRunHooks {
+			hookWg.Add(1)
+			go func(index int) {
+				defer hookWg.Done()
+				if err := c.postRunHooks[index](ctx, c.syncContext); err != nil {
+					klog.Warningf("%s controller post run hook error: %v", c.name, err)
+				}
+			}(i)
+		}
 	}
 
 	// Handle controller shutdown
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -17,12 +17,9 @@ import (
 // syncContext implements SyncContext and provide user access to queue and object that caused
 // the sync to be triggered.
 type syncContext struct {
-	queue         workqueue.RateLimitingInterface
 	eventRecorder events.Recorder
-
-	// queueRuntimeObject holds the object we got from informer
-	// There is no direct access to this object to prevent cache mutation.
-	queueRuntimeObject runtime.Object
+	queue         workqueue.RateLimitingInterface
+	queueKey      string
 }
 
 var _ SyncContext = syncContext{}
@@ -35,30 +32,16 @@ func NewSyncContext(name string, recorder events.Recorder) SyncContext {
 	}
 }
 
-// GetObject gives the object from the queue.
-// For controllers generated without WithRuntimeObject() this always return nil.
-func (c syncContext) GetObject() runtime.Object {
-	if c.queueRuntimeObject == nil {
-		return nil
-	}
-	return c.queueRuntimeObject.DeepCopyObject()
-}
-
 func (c syncContext) Queue() workqueue.RateLimitingInterface {
 	return c.queue
 }
 
-func (c syncContext) Recorder() events.Recorder {
-	return c.eventRecorder
+func (c syncContext) QueueKey() string {
+	return c.queueKey
 }
 
-// withRuntimeObject make a copy of existing sync context and set the queueRuntimeObject.
-func (c syncContext) withRuntimeObject(obj runtime.Object) SyncContext {
-	return syncContext{
-		eventRecorder:      c.Recorder(),
-		queue:              c.Queue(),
-		queueRuntimeObject: obj,
-	}
+func (c syncContext) Recorder() events.Recorder {
+	return c.eventRecorder
 }
 
 func (c syncContext) isInterestingNamespace(obj interface{}, interestingNamespaces sets.String) (bool, bool) {
@@ -76,78 +59,46 @@ func (c syncContext) isInterestingNamespace(obj interface{}, interestingNamespac
 }
 
 // eventHandler provides default event handler that is added to an informers passed to controller factory.
-func (c syncContext) eventHandler(keyName string, objectQueue bool, interestingNamespaces sets.String) cache.ResourceEventHandler {
+func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeyFunc, interestingNamespaces sets.String) cache.ResourceEventHandler {
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			isNamespace, isInteresting := c.isInterestingNamespace(obj, interestingNamespaces)
-			if !objectQueue {
-				if !isNamespace {
-					c.Queue().Add(keyName)
-				} else if isInteresting {
-					c.Queue().Add(keyName)
-				}
-				return
-			}
 			runtimeObj, ok := obj.(runtime.Object)
 			if !ok {
 				utilruntime.HandleError(fmt.Errorf("added object %+v is not runtime Object", obj))
 				return
 			}
-			if !isNamespace {
-				c.Queue().Add(runtimeObj)
-			} else if isInteresting {
-				c.Queue().Add(runtimeObj)
+			if !isNamespace || (isNamespace && isInteresting) {
+				c.Queue().Add(queueKeyFunc(runtimeObj))
 			}
 		},
 		UpdateFunc: func(old, new interface{}) {
 			isNamespace, isInteresting := c.isInterestingNamespace(new, interestingNamespaces)
-			if !objectQueue {
-				if !isNamespace {
-					c.Queue().Add(keyName)
-				} else if isInteresting {
-					c.Queue().Add(keyName)
-				}
-				return
-			}
 			runtimeObj, ok := new.(runtime.Object)
 			if !ok {
 				utilruntime.HandleError(fmt.Errorf("updated object %+v is not runtime Object", runtimeObj))
 				return
 			}
-			if !isNamespace {
-				c.Queue().Add(runtimeObj)
-			} else if isInteresting {
-				c.Queue().Add(runtimeObj)
+			if !isNamespace || (isNamespace && isInteresting) {
+				c.Queue().Add(queueKeyFunc(runtimeObj))
 			}
 		},
 		DeleteFunc: func(obj interface{}) {
 			isNamespace, isInteresting := c.isInterestingNamespace(obj, interestingNamespaces)
-			if !objectQueue {
-				if !isNamespace {
-					c.Queue().Add(keyName)
-				} else if isInteresting {
-					c.Queue().Add(keyName)
-				}
-				return
-			}
 			runtimeObj, ok := obj.(runtime.Object)
 			if !ok {
 				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
 				if ok {
-					if !isNamespace {
-						c.Queue().Add(tombstone.Obj.(runtime.Object))
-					} else if isInteresting {
-						c.Queue().Add(tombstone.Obj.(runtime.Object))
+					if !isNamespace || (isNamespace && isInteresting) {
+						c.Queue().Add(queueKeyFunc(tombstone.Obj.(runtime.Object)))
 					}
 					return
 				}
 				utilruntime.HandleError(fmt.Errorf("updated object %+v is not runtime Object", runtimeObj))
 				return
 			}
-			if !isNamespace {
-				c.Queue().Add(runtimeObj)
-			} else if isInteresting {
-				c.Queue().Add(runtimeObj)
+			if !isNamespace || (isNamespace && isInteresting) {
+				c.Queue().Add(queueKeyFunc(runtimeObj))
 			}
 		},
 	}
--- a/pkg/controller/factory/factory.go
+++ b/pkg/controller/factory/factory.go
@@ -1,6 +1,7 @@
 package factory
 
 import (
+	"context"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -17,6 +18,7 @@ type Factory struct {
 	resyncInterval        time.Duration
 	objectQueue           bool
 	informers             []Informer
+	postStartHooks        []PostStartHook
 	namespaceInformers    []*namespaceInformer
 	cachesToSync          []cache.InformerSynced
 	interestingNamespaces sets.String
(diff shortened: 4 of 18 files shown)


pkg/controller/factory/factory.go
```diff
@@ -54,6 +61,12 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
 	return f
 }
 
+// WithPostRunHooks allows to register functions that will run asynchronously after the controller is started via Run command.
+func (f *Factory) WithPostRunHooks(hooks ...PostRunHook) *Factory {
```
mfojtik: see: https://github.com/openshift/library-go/pull/731/commits/80202dc074c17a2981110c7ce0b7fc706a8fde8d#diff-2261944c4e72b8534466fe56d0a53f41R119 for example.

pkg/controller/factory/base_controller.go
```diff
@@ -66,6 +68,24 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 		}()
 	}
 
+	// run post-run hooks (custom triggers, etc.)
+	if len(c.postRunHooks) > 0 {
+		var hookWaitGroup sync.WaitGroup
```
sttts: why not `wg` like everywhere else?

mfojtik: we already have one above

pkg/controller/factory/base_controller.go
```diff
@@ -66,6 +68,24 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
+		defer func() {
+			defer klog.Infof("All %s post-run hooks have been terminated", c.name)
+			hookWaitGroup.Wait() // wait for the post-run hooks
+		}()
+		for i := range c.postRunHooks {
+			hookWaitGroup.Add(1)
+			go func() {
+				defer hookWaitGroup.Done()
+				if err := c.postRunHooks[i](ctx, c.syncContext); err != nil {
```
sttts: `i` race. Make it a param of the go func.

pkg/controller/factory/factory.go
```diff
@@ -34,6 +36,11 @@ type namespaceInformer struct {
 	namespaces sets.String
 }
 
+// PostRunHook specify a function that will run after controller is ran.
+// The context is cancelled when the controller is asked to shutdown and the post start hook should terminate as well.
+// The syncContext allow access to controller queue and event recorder.
+type PostRunHook func(ctx context.Context, syncContext SyncContext) error
```
sttts: ERRTOOMANYCONTEXTS

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -116,25 +100,29 @@ func (c *APIServiceController) sync() error {
...
 	ready, err := c.precondition(apiServices)
 	if err != nil {
-		v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+		if _, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:    "APIServicesAvailable",
 			Status:  operatorv1.ConditionFalse,
 			Reason:  "ErrorCheckingPrecondition",
 			Message: err.Error(),
-		}))
+		})); updateErr != nil {
```
sttts: for the reasons given in Slack, would keep returning `err`.

sttts: Or at least combine both into one string.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
```diff
@@ -1,14 +1,16 @@
...
 
-	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/tools/cache"
 	"sort"
 	"strings"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+
```
sttts: no empty line

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -27,10 +23,10 @@ const (
 	CertificateIssuer = "auth.openshift.io/certificate-issuer"
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
+	// RunOnceContextKey is a context value key that can be used to call the controller Sync() and make it only run the syncWorker once and report error.
+	RunOnceContextKey = "cert-rotation-controller.openshift.io/run-once"
```
sttts: this is odd. Have never seen this.

pkg/controller/factory/base_controller.go
sttts: ok, missed that. hookWG? :)

mfojtik: sure, characters are cheap, right? ;-)

pkg/controller/factory/base_controller.go
mfojtik: good catch, I will add some unit-test as follow up.

pkg/controller/factory/factory.go
mfojtik: upstream calls SyncContext `ReconcileRequest`... I don't think it is worth renaming it now...

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: it is legit and I don't see other way to achieve prior behavior and keeping the `factory.Controller` interface there... I think we want to suppress snowflakes in controllers. This will give reasonable replacement for controller.RunOnce() method :-)

pkg/controller/factory/base_controller.go
mfojtik: fixed.

pkg/controller/factory/base_controller.go
mfojtik: fixed.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
mfojtik: fixed.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
mfojtik: fixed.

pkg/controller/factory/base_controller.go
```diff
@@ -36,36 +38,54 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
+		var hookWg sync.WaitGroup
+		defer func() {
+			defer klog.Infof("All %s post-run hooks have been terminated", c.name)
+			hookWg.Wait() // wait for the post-run hooks
+		}()
+		for i := range c.postRunHooks {
+			hookWg.Add(1)
+			go func(index int) {
+				defer hookWg.Done()
+				if err := c.postRunHooks[index](ctx, c.syncContext); err != nil {
```
sttts: are you sure you want to call them here and not after the context closed?!

mfojtik: @sttts we have two controllers that are adding "custom" triggers (if A change add to workqueue...). Maybe "PostRun" is not the perfect name for this... This hook should run after the controller is started... "PostStart" ?

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: am still lost when reading "context key". What is a context key?

mfojtik: https://golang.org/pkg/context/#example_WithValue

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -93,6 +91,7 @@ func (c CertRotationController) sync() error {
 		newCondition.Message = syncErr.Error()
 	}
 	if _, _, updateErr := v1helpers.UpdateStaticPodStatus(c.OperatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition)); updateErr != nil {
+		syncCtx.Recorder().Warningf("RotationError", newCondition.Message)
```
sttts: does this really need an event?

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: @sttts in other words... if you have run-once context passed to Sync(), return right after the syncWorker() is called with error it returns and don't try to update status or something.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -117,97 +116,24 @@ func (c CertRotationController) syncWorker() error {
...
 	go wait.Until(func() {
-		ticker := time.NewTicker(time.Minute)
-		defer ticker.Stop()
-
 		for {
-			c.queue.Add(workQueueKey)
 			select {
-			case <-ticker.C:
+			case <-targetRefresh:
+				syncCtx.Queue().Add(factory.QueueKey(c.name))
```
sttts: why do we queue names?

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: wrong code... I should get `updated` bool and only fire this event once when we update the condition.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -57,32 +50,37 @@ func NewCertRotationController(
...
+			targetRotation.Informer.Informer(),
+		).
+		WithPostRunHooks(
+			c.targetCertRecheckerPostRunHook,
+		).
+		ToController("CertRotationController", recorder.WithComponentSuffix("cert-rotation-controller"))
 }
 
-func (c CertRotationController) sync() error {
+func (c CertRotationController) Sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
sttts: why is this a non-pointer now?

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: prometheus metrics? (each queue have one, makes it easier to see which queue is busy)

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: it does not mutate the CertRotationController and still satisfy the interface.

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: fixed (locally):

```golang
	_, updated, updateErr := v1helpers.UpdateStaticPodStatus(c.OperatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition))
	if updateErr != nil {
		return updateErr
	}
	if updated && syncErr != nil {
		syncCtx.Recorder().Warningf("RotationError", newCondition.Message)
	}
```

pkg/controller/factory/base_controller.go
sttts: I think I got it. The name postRun is super bad. This is not post run. It is post start. Please rename it to avoid confusion.

pkg/controller/factory/factory.go
sttts: as commented about. This is not post-run. It is post-start. Pleaes rename.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -117,97 +116,24 @@ func (c CertRotationController) syncWorker() error {
...
 }
 
-func (c *CertRotationController) WaitForReady(stopCh <-chan struct{}) {
-	klog.Infof("Waiting for CertRotationController - %q", c.name)
-	defer klog.Infof("Finished waiting for CertRotationController - %q", c.name)
-
-	if !cache.WaitForCacheSync(stopCh, c.cachesToSync...) {
-		utilruntime.HandleError(fmt.Errorf("caches did not sync"))
-		return
+func (c CertRotationController) targetCertRecheckerPostRunHook(ctx context.Context, syncCtx factory.SyncContext) error {
```
sttts: pointer receiver

mfojtik: @sttts do we need one? (i'm fine adding it)

pkg/operator/staleconditions/remove_stale_conditions.go
```diff
@@ -63,55 +42,3 @@ func (c RemoveStaleConditions) sync() error {
...
-	klog.Infof("Starting RemoveStaleConditions")
-	defer klog.Infof("Shutting down RemoveStaleConditions")
-
-	if !cache.WaitForCacheSync(ctx.Done(), c.cachesToSync...) {
-		utilruntime.HandleError(fmt.Errorf("caches did not sync"))
-		return
-	}
-
-	// doesn't matter what workers say, only start one.
-	go wait.Until(c.runWorker, time.Second, ctx.Done())
```
sttts: this is lost. The factory will run as many as the consumer asks for.

mfojtik: `return factory.New().ResyncEvery(time.Second)` does this.

mfojtik: we doing this for almost every controller....

pkg/operator/staleconditions/remove_stale_conditions.go
```diff
@@ -63,55 +42,3 @@ func (c RemoveStaleConditions) sync() error {
...
-
-	return true
-}
-
-// eventHandler queues the operator to check spec and status
-func (c *RemoveStaleConditions) eventHandler() cache.ResourceEventHandler {
-	return cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { c.queue.Add(workQueueKey) },
-		UpdateFunc: func(old, new interface{}) { c.queue.Add(workQueueKey) },
-		DeleteFunc: func(obj interface{}) { c.queue.Add(workQueueKey) },
```
sttts: as far as I see the key names are out of control of the controller authors. That's bad. This is a core decission what to queue when. In the base controller we queue `name+"Key"` while other controllers want actual object names to be queued. That makes a huge difference in the context of multiple works, but also in the context of `c.queue.Forget/Done`.

pkg/operator/status/status_controller.go
```diff
@@ -64,26 +62,24 @@ func NewClusterOperatorStatusController(
...
 
-	operatorClient.Informer().AddEventHandler(c.eventHandler())
-	clusterOperatorInformer.Informer().AddEventHandler(c.eventHandler())
-
-	c.cachesToSync = append(c.cachesToSync, operatorClient.Informer().HasSynced)
-	c.cachesToSync = append(c.cachesToSync, clusterOperatorInformer.Informer().HasSynced)
-
-	return c
+func (c *StatusSyncer) Run(ctx context.Context, workers int) {
+	c.controllerFactory.WithPostRunHooks(c.watchVersionGetterPostRunHook).WithSync(c.Sync).ToController("StatusSyncer_"+c.Name(), c.recorder).Run(ctx, workers)
```
sttts: why not call all those With* funcs from the constructor. Continueing construction here in `Run` looks wrong.

pkg/operator/status/status_controller.go
```diff
@@ -125,11 +121,11 @@ func (c StatusSyncer) sync() error {
 		if apierrors.IsNotFound(createErr) {
 			// this means that the API isn't present.  We did not fail.  Try again later
 			klog.Infof("ClusterOperator API not created")
-			c.queue.AddRateLimited(workQueueKey)
+			syncCtx.Queue().AddRateLimited(factory.QueueKey("StatusSyncer_" + c.clusterOperatorName))
```
sttts: this is bad that we manually construct the key here. This quickly diverges from the constant key used elsewhere.

sttts: In other words: have one place where to get the constant key from.

pkg/operator/status/status_controller.go
```diff
@@ -192,71 +188,20 @@ func OperatorConditionToClusterOperatorCondition(condition operatorv1.OperatorCo
...
-	// doesn't matter what workers say, only start one.
-	go wait.UntilWithContext(ctx, c.runWorker, time.Second)
-
-	<-ctx.Done()
-}
-
-func (c *StatusSyncer) watchVersionGetter(stopCh <-chan struct{}) {
+func (c *StatusSyncer) watchVersionGetterPostRunHook(ctx context.Context, syncCtx factory.SyncContext) error {
 	defer utilruntime.HandleCrash()
+	workQueueKey := factory.QueueKey("StatusSyncer_" + c.clusterOperatorName)
```
sttts: again, don't have more than one place where the constant key is defined

pkg/operator/status/status_controller.go
mfojtik: @sttts the reason is that we have mutating function that mutates the controller internal variables, which effectively changes the controller internal state... I saw that when you run Sync() then it might produce wrong results.

pkg/operator/staleconditions/remove_stale_conditions.go
mfojtik: @sttts I made `WithQueueKeyFunc` that allows you to specify custom queue key function. By default I use the one we already had (based on controller name). I'm open to add more extensibility if we need to (like add map[informer]keyFunc or something).

pkg/operator/status/status_controller.go
mfojtik: fixed.

pkg/operator/status/status_controller.go
mfojtik: fixed.

pkg/controller/factory/factory.go
mfojtik: renamed.

pkg/controller/factory/base_controller.go
mfojtik: renamed.

pkg/controller/factory/base_controller.go
```diff
@@ -82,17 +103,19 @@ func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
 	return c.sync(ctx, syncCtx)
 }
 
-// QueueKey return queue key for given name.
-func QueueKey(name string) string {
-	return strings.ToLower(name) + "Key"
+// DefaultQueueKeyFunc is the default queue key function that transforms the name into workqueue key.
+// Eg. If name is 'FooController' the key will be "foocontrollerKey"
+// It also automatically convert '-' to '_' as '-' is not allowed in queue keys.
+func DefaultQueueKeyFunc(name interface{}) string {
```
sttts: name is odd. What is passed here? the object?

sttts: the cast below will panic possibly

pkg/controller/factory/factory.go
```diff
@@ -34,6 +37,14 @@ type namespaceInformer struct {
 	namespaces sets.String
 }
 
+// PostStartHook specify a function that will run after controller is ran.
```
sttts: is started

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: It's for tests?

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: am not convinced we want to create events on every error.

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: but this is not the queue, it's the key in a queue. I hope this is not exposed as metrics label as that would be CVE worths (memory explosion attack)

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: I hate this in Go that call-by-value receivers semantics is misused for read-only enforcement. It is slow to copy around huge objects every time.

pkg/operator/staleconditions/remove_stale_conditions.go
sttts: we have controllers which cannot run with multiple workers, yet the interfaces allow it, leading to races and worse things.

pkg/operator/status/status_controller.go
sttts: I cannot follow. Please explain. This sounds very fishy.

pkg/controller/factory/base_controller.go
```diff
@@ -36,36 +35,54 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
-		workerWaitGroup.Add(1)
+		workerWg.Add(1)
 		go func() {
-			defer workerWaitGroup.Done()
+			defer workerWg.Done()
 			c.runPeriodicalResync(ctx, c.resyncEvery)
 		}()
 	}
 
+	// run post-run hooks (custom triggers, etc.)
```
sttts: post-start

pkg/controller/factory/base_controller.go
```diff
@@ -36,36 +35,54 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
+	if len(c.postStartHooks) > 0 {
+		var hookWg sync.WaitGroup
+		defer func() {
+			defer klog.Infof("All %s post-run hooks have been terminated", c.name)
```
sttts: post-start

sttts: you like defers ;)

pkg/controller/factory/factory.go
```diff
@@ -91,6 +126,16 @@ func (f *Factory) WithSyncContext(ctx SyncContext) *Factory {
 	return f
 }
 
+// DefaultConstQueueKey returns a "key" const that is used to trigger a string based controllers.
+func DefaultConstQueueKey() string {
+	return defaultConstQueueKey(nil)
+}
+
+// defaultConstQueueKey is a wrapper for string const queue key function.
+func defaultConstQueueKey(runtime.Object) string {
```
sttts: why do we need this and the one above?

sttts: Would make `defaultConstQueueKey` available as `DefaultConstKeyFunc` and `defaultStringQueueKey` available as `DefaultConstQueueKey`.

pkg/controller/factory/base_controller.go
```diff
@@ -36,36 +35,54 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
+			hookWg.Wait() // wait for the post-run hooks
+		}()
+		for i := range c.postStartHooks {
+			hookWg.Add(1)
+			go func(index int) {
+				defer hookWg.Done()
+				if err := c.postStartHooks[index](ctx, c.syncContext); err != nil {
+					klog.Warningf("%s controller post run hook error: %v", c.name, err)
```
sttts: post-start

pkg/controller/factory/factory.go
```diff
@@ -54,6 +77,26 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
...
+
+// WithQueueKeyFunc allows to specify custom queue key function.
+// By default, the queue key func returns "key" constant.
+func (f *Factory) WithStringQueueKeyFunc(queueKey StringQueueKeyFunc) *Factory {
+	f.queueKeyFn = queueKey
+	return f
+}
+
+// WithPostRunHooks allows to register functions that will run asynchronously after the controller is started via Run command.
+func (f *Factory) WithPostRunHooks(hooks ...PostStartHook) *Factory {
```
sttts: PostStart

pkg/controller/factory/factory.go
```diff
@@ -54,6 +77,26 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
 	return f
 }
 
+// WithInformerQueueKeys allows to pass multiple InformerQueueKeyFunc that return informer and the corresponding queue key function
+// that will be used for the particular informer.
+func (f *Factory) WithInformerQueueKeyFuncs(mappings ...InformerQueueKeyFunc) *Factory {
```
sttts: this looks like Scala

sttts: Why no simple `WithInformers(keyFunc, informers....)`? If you want different keyFuncs, call it multiple times. The functional informer type works, but feels misplaced and oversized.

pkg/controller/factory/factory.go
```diff
@@ -54,6 +77,26 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
...
+	f.informerQueueKeys = append(f.informerQueueKeys, mappings...)
+	return f
+}
+
+// WithQueueKeyFunc allows to specify custom queue key function.
+// By default, the queue key func returns "key" constant.
+func (f *Factory) WithStringQueueKeyFunc(queueKey StringQueueKeyFunc) *Factory {
```
sttts: Why do we need a string transformer, i.e. no constant string key func?

mfojtik: we don't need this and we don't use it anywhere... i think we don't need to allow controllers to customize the queue "key" value at all. i removed this

pkg/controller/factory/factory.go
mfojtik: neat, right?

pkg/controller/factory/factory.go
mfojtik: no, keep it simple, just make one public const and controllers can reference

pkg/controller/factory/base_controller.go
mfojtik: removed :)

pkg/config/leaderelection/leaderelection.go
```diff
@@ -64,7 +65,8 @@ func ToConfigMapLeaderElection(clientConfig *rest.Config, config configv1.Leader
 		RetryPeriod:     config.RetryPeriod.Duration,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStoppedLeading: func() {
-				klog.Fatalf("leaderelection lost")
+				defer os.Exit(0)
```
sttts: you like defer, I repeat myself.

pkg/controller/factory/base_controller.go
```diff
@@ -36,36 +35,54 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
 			c.runPeriodicalResync(ctx, c.resyncEvery)
 		}()
 	}
 
+	// run post-start hooks (custom triggers, etc.)
+	if len(c.postStartHooks) > 0 {
+		var hookWg sync.WaitGroup
+		defer func() {
+			hookWg.Wait() // wait for the post-run hooks
+			klog.Infof("All %s post-run hooks have been terminated", c.name)
```
sttts: post start

pkg/controller/factory/base_controller.go
```diff
@@ -117,18 +129,17 @@ func (c *baseController) runWorker(queueCtx context.Context) {
...
 		return
 	}
-	defer c.syncContext.Queue().Done(syncObject)
+	defer c.syncContext.Queue().Done(key)
 
-	runtimeObject, _ := syncObject.(runtime.Object)
-	if err := c.sync(queueCtx, c.syncContext.(syncContext).withRuntimeObject(runtimeObject)); err != nil {
-		utilruntime.HandleError(fmt.Errorf("%s controller failed to sync %+v with: %w", c.name, syncObject, err))
-		c.syncContext.Queue().AddRateLimited(syncObject)
+	if err := c.sync(queueCtx, c.syncContext); err != nil {
```
sttts: not passing the key is odd

sttts: can we agree on strings in the queue?

pkg/controller/factory/controller_context.go
```diff
@@ -76,78 +74,54 @@ func (c syncContext) isInterestingNamespace(obj interface{}, interestingNamespac
...
-				return
-			}
 			runtimeObj, ok := obj.(runtime.Object)
 			if !ok {
 				utilruntime.HandleError(fmt.Errorf("added object %+v is not runtime Object", obj))
 				return
 			}
 			if !isNamespace {
-				c.Queue().Add(runtimeObj)
+				c.Queue().Add(queueKeyFunc(runtimeObj))
```
sttts: merge both if cases. they have the same body.

pkg/controller/factory/factory.go
```diff
@@ -34,6 +41,23 @@ type namespaceInformer struct {
...
+// The syncContext allow access to controller queue and event recorder.
+type PostStartHook func(ctx context.Context, syncContext SyncContext) error
+
+// ObjectQueueKeyFunc is used to make a string work queue key out of the runtime object that is passed to it.
+// This can extract the "namespace/name" if you need to or just return "key" if you building controller that only use string
+// triggers.
+type ObjectQueueKeyFunc func(runtime.Object) string
+
+// StringQueueKeyFunc is used to make a string work queue key based on string input.
+type StringQueueKeyFunc func(string) string
```
sttts: is this still used? I hope not. This signature confuses me.

pkg/controller/factory/factory.go
sttts: Scala does not mean "neat". It means "what the hell are you doing here?" ;-)

pkg/controller/factory/factory.go
mfojtik: not used

pkg/config/leaderelection/leaderelection.go
mfojtik: this is right way :) you want os exit in defer so other defers will run... it is good to get a habit ;)

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: @sttts no, it is for one single snowflake controller that we need to expose this mode for (RunOnce)... I would dwell on it too much :-)

pkg/operator/certrotation/client_cert_rotation_controller.go
mfojtik: events make timelines

sttts: logs are timelines.

pkg/operator/certrotation/client_cert_rotation_controller.go
sttts: ok, got it

pkg/controller/factory/factory.go
```diff
@@ -54,6 +75,24 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
 	return f
 }
 
+// WithInformersQueueKeyFunc is used to register event handlers and get the caches synchronized functions.
+// Pass informers you want to use to react to changes on resources. If informer event is observed, then the Sync() function
+// is called.
+// Pass the queueKeyFn you want to use to transform the informer runtime.Object into string key used by work queue.
+func (f *Factory) WithInformersQueueKeyFunc(queueKeyFn ObjectQueueKeyFunc, informers ...Informer) *Factory {
```
sttts: 😍

PR #732, opened 2020-03-10
implements a generic workload controller

p0lyn0mial:
it will be used by `openshift-apiserver` and `oauth-apiserver`

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -35,8 +35,14 @@ const (
 	workQueueKey = "key"
 )
 
-// SyncFunc a function that will be used for delegation. It should bring the desired workload into operation.
-type SyncFunc func() (*appsv1.Deployment, bool, []error)
+// Delegate captures a set of methods that hold a custom logic
+type Delegate interface {
+	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
+	Sync() (*appsv1.Deployment, bool, []error)
+
+	// PreconditionFulfilled a method that indicates whether all prerequisites are met and we can Sync.
+	PreconditionFulfilled() (bool, error)
+}
 
 // Controller is a generic workload controller that deals with Deployment resource.
 // Callers must provide a sync function for delegation. It should bring the desired workload into operation.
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
I wired it manually in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/333

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	workload, operatorConfigAtHighestGeneration, errs := c.syncFn()
+
+	return c.updateOperatorStatus(workload, operatorConfigAtHighestGeneration, errs)
+}
+
+// Run starts workload controller and blocks until stopCh is closed.
+// Note that setting workers doesn't have any effect, the controller is single-threaded.
+func (c *Controller) Run(ctx context.Context, workers int) {
```
p0lyn0mial: Probably we should drop `workers` param

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -0,0 +1,447 @@
...
+	operatorSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+
+	if run, err := c.shouldSync(operatorSpec); !run {
+		return err
+	}
+
+	if fulfilled, err := c.preconditionFulfilled(operatorSpec); !fulfilled {
```
p0lyn0mial: I think I am going to add a custom function for `precoditions`, for example when a config hasn't been populated there is no need to run the `sync` function. That function would be provided by the callers.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -0,0 +1,457 @@
...
+
+// AddInformer queues the given informer to check spec, status and managed resources
+func (c *Controller) AddInformer(informer cache.SharedIndexInformer) *Controller {
+	informer.AddEventHandler(c.eventHandler())
+	c.preRunCachesSynced = append(c.preRunCachesSynced, informer.HasSynced)
+	return c
+}
+
+// AddNamespaceInformer queues the given ns informer for the targetNamespace
+func (c *Controller) AddNamespaceInformer(informer cache.SharedIndexInformer) *Controller {
```
sttts: how odd that we have a special namespace informer code path. @mfojtik do you have that on the radar?

PR #733, opened 2020-03-10
apiserver observers: add switch to observe to apiServerArguments

stlaz:
/cc @p0lyn0mial @sttts

### DIFF
--- a/pkg/operator/configobserver/apiserver/observe_cors.go
+++ b/pkg/operator/configobserver/apiserver/observe_cors.go
@@ -16,9 +16,18 @@ var clusterDefaultCORSAllowedOrigins = []string{
 }
 
 // ObserveAdditionalCORSAllowedOrigins observes the additionalCORSAllowedOrigins field
-// of the APIServer resource
-func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
-	corsAllowedOriginsPath := []string{"corsAllowedOrigins"}
+// of the APIServer resource and sets the corsAllowedOrigins field of observedConfig
+func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	return innerObserveAdditionalCORSAllowedOrigins(genericListers, recorder, existingConfig, []string{"corsAllowedOrigins"})
+}
+
+// ObserveAdditionalCORSAllowedOriginsToArguments observes the additionalCORSAllowedOrigins field
+// of the APIServer resource and sets the cors-allowed-origins field in observedConfig.apiServerArguments
+func ObserveAdditionalCORSAllowedOriginsToArguments(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	return innerObserveAdditionalCORSAllowedOrigins(genericListers, recorder, existingConfig, []string{"apiServerArguments", "cors-allowed-origins"})
+}
+
+func innerObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, corsAllowedOriginsPath []string) (ret map[string]interface{}, _ []error) {
 	defer func() {
 		ret = configobserver.Pruned(ret, corsAllowedOriginsPath)
 	}()
--- a/pkg/operator/configobserver/apiserver/observe_cors_test.go
+++ b/pkg/operator/configobserver/apiserver/observe_cors_test.go
@@ -8,6 +8,7 @@ import (
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/tools/cache"
 )
 
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
-			if tt.corsConfig != nil {
-				if err := indexer.Add(&configv1.APIServer{
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "cluster",
-					},
-					Spec: configv1.APIServerSpec{
-						AdditionalCORSAllowedOrigins: tt.corsConfig,
-					},
-				}); err != nil {
-					t.Fatal(err)
+		for _, useAPIServerArguments := range []bool{false, true} {
+			var corsPath []string
+			if useAPIServerArguments {
+				corsPath = []string{"apiServerArguments", "cors-allowed-origins"}
+			} else {
+				corsPath = []string{"corsAllowedOrigins"}
+			}
+			t.Run(tt.name, func(t *testing.T) {
+				indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+				if tt.corsConfig != nil {
+					if err := indexer.Add(&configv1.APIServer{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "cluster",
+						},
+						Spec: configv1.APIServerSpec{
+							AdditionalCORSAllowedOrigins: tt.corsConfig,
+						},
+					}); err != nil {
+						t.Fatal(err)
+					}
 				}
-			}
-			listers := testLister{
-				lister: configlistersv1.NewAPIServerLister(indexer),
-			}
-
-			gotConfig, errs := ObserveAdditionalCORSAllowedOrigins(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
-			if len(errs) > 0 {
-				t.Errorf("ObserveAdditionalCORSAllowedOrigins() expected no errors, got %v", errs)
-			}
-
-			gotCors := gotConfig["corsAllowedOrigins"].([]interface{})
-			for i := range tt.expectedCORS {
-				if gotCors[i] != tt.expectedCORS[i] {
-					t.Fatalf("ObserveAdditionalCORSAllowedOrigins() got = %v, want %v", gotCors, tt.expectedCORS)
+				listers := testLister{
+					lister: configlistersv1.NewAPIServerLister(indexer),
 				}
-			}
-		})
+
+				var gotConfig map[string]interface{}
+				var errs []error
+				if useAPIServerArguments {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOriginsToArguments(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				} else {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOrigins(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				}
+				if len(errs) > 0 {
+					t.Errorf("expected no errors, got %v", errs)
+				}
+
+				gotCors, _, _ := unstructured.NestedStringSlice(gotConfig, corsPath...)
+				for i := range tt.expectedCORS {
+					if gotCors[i] != tt.expectedCORS[i] {
+						t.Fatalf("got = %v, want %v", gotCors, tt.expectedCORS)
+					}
+				}
+			})
+		}
 	}
 }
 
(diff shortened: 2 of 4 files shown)


pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -24,23 +24,47 @@ type APIServerLister interface {
...
+// configuration (or their apiServerArguments mirrors)
+func NewTLSSSecurityProfileObserver(useAPIServerArguments bool) configobserver.ObserveConfigFunc {
+	observer := &tlsSecurityObserver{
+		minTlSVersionPath: minTLSVersionPath[useAPIServerArguments],
+		cipherSuitesPath:  cipherSuitesPath[useAPIServerArguments],
+	}
+	return observer.observeTLSSecurityProfile
+}
 
+func (o *tlsSecurityObserver) observeTLSSecurityProfile(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
p0lyn0mial: what are we going to do on upgrades, imagine that `existingConfig` was created with `servingInfo`, we rolled out a new version and now we are going to put values into `apiServerArguments`, should we clean up?

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -24,23 +24,47 @@ type APIServerLister interface {
...
+	cipherSuitesPath = map[bool][]string{
+		false: {"servingInfo", "cipherSuites"},
+		true:  {"apiServerArguments", "tls-cipher-suites"},
+	}
+)
+
+// NewTLSSSecurityProfileObserver creates and observer for the APIServer.Spec.TLSSecurityProfile field
+// which sets the ServingInfo.MinTLSVersion and ServingInfo.CipherSuites fields values in the observed
+// configuration (or their apiServerArguments mirrors)
+func NewTLSSSecurityProfileObserver(useAPIServerArguments bool) configobserver.ObserveConfigFunc {
```
p0lyn0mial: Just wondering if we could do something like the following for minimal disruption to components that use library-go  . So we could preserve the old interface, extract the common logic and provide a new function.


```
func ObserveTLSSecurityProfile(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
	var (
		minTlSVersionPath = []string{"servingInfo", "minTLSVersion"}
		cipherSuitesPath  = []string{"servingInfo", "cipherSuites"}
	)
        // here we could call the common function passing the old `minTlSVersionPath` and `cipherSuitesPath`

}
```

```
func NewObserveTLSSecurityProfile(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
	var (
		minTlSVersionPath = []string{"apiServerArguments", "tls-min-version"}
		cipherSuitesPath  = []string{"apiServerArguments", "tls-cipher-suites"}
	)
        // here we would call the common function passing different `minTlSVersionPath` and `cipherSuitesPath` 
       //  TODO: do we have to clean up the existingConfig - remove the old entries (`servingInfo`) ?

}
```

In the future releases we could just simply remove  `ObserveTLSSecurityProfile` and rename `NewObserveTLSSecurityProfile`

pkg/operator/configobserver/apiserver/helpers.go
```diff
@@ -0,0 +1,18 @@
+package apiserver
+
+// prepareEmptyInterfaceForPath creates empty map[string]interface{} all the way to
+// before the last node of a path
+func prepareEmptyInterfaceForPath(path ...string) map[string]interface{} {
```
sttts: same as `SetNestedField(map[string]interface{}{}, map[string]interface{}{}, path[:len(path)-1]...)`

pkg/operator/configobserver/apiserver/observe_cors.go
```diff
@@ -10,33 +10,50 @@ import (
...
-	currentCORSAllowedOrigins, _, err := unstructured.NestedStringSlice(existingConfig, corsAllowedOriginsPath)
+	currentCORSAllowedOrigins, _, err := unstructured.NestedStringSlice(existingConfig, o.corsAllowedOriginsPath...)
 	if err != nil {
 		return defaultConfig, append(errs, err)
 	}
 	currentCORSSet := sets.NewString(currentCORSAllowedOrigins...)
 	currentCORSSet.Insert(clusterDefaultCORSAllowedOrigins...)
 
-	observedConfig := map[string]interface{}{}
+	observedConfig := prepareEmptyInterfaceForPath(o.corsAllowedOriginsPath...)
```
sttts: even start with `map[string]interface{}{}`. No need to prepare anything. `SetNestedStringSlice` will create the path on the way.

stlaz: I could have sworn that that was failing for me with non-nil asserts before, perhaps I was doing something wrong because now they are not. Thanks

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
stlaz: Definitely makes more sense from the perspective of the consumer of these functions, I'll make it so.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
stlaz: I think we should, I'll add that

stlaz: actually, since we are not setting that, that will be cleaned up, won't it?

pkg/operator/configobserver/apiserver/observe_cors_test.go
```diff
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
...
+				}
+				if len(errs) > 0 {
+					t.Errorf("ObserveAdditionalCORSAllowedOrigins() expected no errors, got %v", errs)
+				}
 
-			gotCors := gotConfig["corsAllowedOrigins"].([]interface{})
-			for i := range tt.expectedCORS {
-				if gotCors[i] != tt.expectedCORS[i] {
-					t.Fatalf("ObserveAdditionalCORSAllowedOrigins() got = %v, want %v", gotCors, tt.expectedCORS)
+				gotCors, _, _ := unstructured.NestedStringSlice(gotConfig, corsPath...)
```
p0lyn0mial: okay, this actually checks the `paths` so we now that the config is in good shape.

pkg/operator/configobserver/apiserver/observe_cors_test.go
```diff
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
...
+				for i := range tt.expectedCORS {
+					if gotCors[i] != tt.expectedCORS[i] {
+						t.Fatalf("ObserveAdditionalCORSAllowedOrigins() got = %v, want %v", gotCors, tt.expectedCORS)
```
p0lyn0mial: nit: rm the function name from the err msg ?

pkg/operator/configobserver/apiserver/observe_cors_test.go
```diff
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
...
-			}
+				var gotConfig map[string]interface{}
+				var errs []error
+				if useAPIServerArguments {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOriginsToArguments(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				} else {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOrigins(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				}
+				if len(errs) > 0 {
+					t.Errorf("ObserveAdditionalCORSAllowedOrigins() expected no errors, got %v", errs)
```
p0lyn0mial: here too, the name of the function in the err msg is not relevant anymore.

pkg/operator/configobserver/apiserver/observe_cors.go
```diff
@@ -17,8 +17,15 @@ var clusterDefaultCORSAllowedOrigins = []string{
 
 // ObserveAdditionalCORSAllowedOrigins observes the additionalCORSAllowedOrigins field
 // of the APIServer resource
-func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
-	corsAllowedOriginsPath := []string{"corsAllowedOrigins"}
+func ObserveAdditionalCORSAllowedOrigins(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	return innerObserveAdditionalCORSAllowedOrigins(genericListers, recorder, existingConfig, []string{"corsAllowedOrigins"})
+}
+
+func ObserveAdditionalCORSAllowedOriginsToArguments(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
```
p0lyn0mial: consider describing what it does

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile_test.go
```diff
@@ -67,51 +62,74 @@ func TestObserveTLSSecurityProfile(t *testing.T) {
...
+				gotSuites, _, err := unstructured.NestedStringSlice(result, cipherSuitesPath...)
+				if err != nil {
+					t.Errorf("couldn't get cipherSuites from the returned object: %v", err)
+				}
+
+				if !reflect.DeepEqual(gotSuites, tt.expectedSuites) {
+					t.Errorf("ObserveTLSSecurityProfile() got cipherSuites = %v, expected %v", gotSuites, tt.expectedSuites)
+				}
+				if gotMinTLSVersion != tt.expectedMinTLSVersion {
+					t.Errorf("ObserveTLSSecurityProfile() got minTlSVersion = %v, expected %v", gotMinTLSVersion, tt.expectedMinTLSVersion)
```
p0lyn0mial: nit: rm the function name from the err msg also in line `127`.

pkg/operator/configobserver/apiserver/observe_cors_test.go
```diff
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
...
-			}
+				var gotConfig map[string]interface{}
+				var errs []error
+				if useAPIServerArguments {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOriginsToArguments(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				} else {
+					gotConfig, errs = ObserveAdditionalCORSAllowedOrigins(listers, events.NewInMemoryRecorder(t.Name()), tt.existingConfig)
+				}
+				if len(errs) > 0 {
+					t.Errorf("expected no errors, got %v", errs)
```
p0lyn0mial: nit, could be changed to `t.Fatal` in case of an error we fail anyway . . .

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile_test.go
```diff
@@ -67,51 +62,74 @@ func TestObserveTLSSecurityProfile(t *testing.T) {
...
-			}
+				var result map[string]interface{}
+				var errs []error
+				if useAPIServerArgs {
+					result, errs = ObserveTLSSecurityProfileToArguments(listers, events.NewInMemoryRecorder(t.Name()), existingConfig)
+				} else {
+					result, errs = ObserveTLSSecurityProfile(listers, events.NewInMemoryRecorder(t.Name()), existingConfig)
+				}
+				if len(errs) > 0 {
+					t.Errorf("expected 0 errors, got %v", errs)
```
p0lyn0mial: nit, here to `t.Fatal`

pkg/operator/configobserver/apiserver/observe_cors_test.go
```diff
@@ -42,36 +43,50 @@ func TestObserveAdditionalCORSAllowedOrigins(t *testing.T) {
...
+				}
 
-			gotCors := gotConfig["corsAllowedOrigins"].([]interface{})
-			for i := range tt.expectedCORS {
-				if gotCors[i] != tt.expectedCORS[i] {
-					t.Fatalf("ObserveAdditionalCORSAllowedOrigins() got = %v, want %v", gotCors, tt.expectedCORS)
+				gotCors, _, _ := unstructured.NestedStringSlice(gotConfig, corsPath...)
+				for i := range tt.expectedCORS {
+					if gotCors[i] != tt.expectedCORS[i] {
+						t.Fatalf("got = %v, want %v", gotCors, tt.expectedCORS)
```
p0lyn0mial: nit, to print all incorrect values we could use `t.Errorf`

p0lyn0mial:
added a few additional comments (`nits`), but this pull LGTM

assigning to @sttts for approval.

PR #737, opened 2020-03-12
logging: properly set log level

mfojtik:
Properly determine the current log level and set the `klog.V()` level so we don't run with `V(0)`.
Also add unit tests to make sure this never break.

### DIFF
--- a/pkg/operator/loglevel/logging_controller.go
+++ b/pkg/operator/loglevel/logging_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -12,11 +11,19 @@ import (
 
 type LogLevelController struct {
 	operatorClient operatorv1helpers.OperatorClient
+
+	// for unit tests only
+	setLogLevelFn func(operatorv1.LogLevel) error
+	getLogLevelFn func() (operatorv1.LogLevel, bool)
 }
 
 // sets the klog level based on desired state
 func NewClusterOperatorLoggingController(operatorClient operatorv1helpers.OperatorClient, recorder events.Recorder) factory.Controller {
-	c := &LogLevelController{operatorClient: operatorClient}
+	c := &LogLevelController{
+		operatorClient: operatorClient,
+		setLogLevelFn:  SetLogLEvel,
+		getLogLevelFn:  GetLogLevel,
+	}
 	return factory.New().WithInformers(operatorClient.Informer()).WithSync(c.sync).ToController("LoggingSyncer", recorder)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/operator/loglevel/logging_controller.go
```diff
@@ -4,19 +4,26 @@ import (
...
+	currentLogLevelFn func() (operatorv1.LogLevel, bool)
 }
 
 // sets the klog level based on desired state
 func NewClusterOperatorLoggingController(operatorClient operatorv1helpers.OperatorClient, recorder events.Recorder) factory.Controller {
-	c := &LogLevelController{operatorClient: operatorClient}
+	c := &LogLevelController{
+		operatorClient:    operatorClient,
+		setVerbosityFn:    SetVerbosityValue,
+		currentLogLevelFn: CurrentLogLevel,
```
sttts: why do we have three terms for the same thing? verbosity, verbosityValue, logLevel

pkg/operator/loglevel/logging_controller.go
```diff
@@ -28,24 +35,37 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
 		return err
 	}
 
-	currentLogLevel := CurrentLogLevel()
+	currentLogLevel, isUnknown := c.currentLogLevelFn()
```
sttts: haha, I had started like that with a second return value and reverted

pkg/operator/loglevel/logging_controller.go
```diff
@@ -28,24 +35,37 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
...
 	desiredLogLevel := detailedSpec.OperatorLogLevel
 
+	// if operator operatorSpec OperatorLogLevel is empty, default to "Normal"
+	// TODO: This should be probably done by defaulting the CR field?
```
sttts: doesn't this apply to startup as well when loglevel is not set?

mfojtik: yes.

pkg/operator/loglevel/logging_controller.go
mfojtik: we can pick one and stick to it, I can unify them

pkg/operator/loglevel/logging_controller.go
```diff
@@ -28,24 +35,37 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+
+	// Set the new loglevel if the operator operatorSpec changed
+	if err := c.setVerbosityFn(desiredLogLevel); err != nil {
 		syncCtx.Recorder().Warningf("OperatorLoglevelChangeFailed", "Unable to change operator log level from %q to %q: %v", currentLogLevel, desiredLogLevel, err)
 		return err
 	}
 
+	// give better message when we setting the log level for the first time
+	if !isUnknown {
+		syncCtx.Recorder().Eventf("OperatorLoglevelChange", "Operator log level set to %q", currentLogLevel)
```
sttts: How do we make sure this does not fire on every restart?

mfojtik: it will. we can skip this event.

sttts: yes, please. If you can contain it, fine. But it must not show up multiple times, only on CRD user action.

pkg/operator/loglevel/logging_controller.go
sttts: yes please

pkg/operator/loglevel/logging_controller.go
```diff
@@ -4,19 +4,26 @@ import (
...
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 type LogLevelController struct {
 	operatorClient operatorv1helpers.OperatorClient
+
+	// for unit tests only
+	setLogLevelFn func(operatorv1.LogLevel) error
+	getLogLevelFn func() (operatorv1.LogLevel, bool)
```
sttts: uniformity 😍

pkg/operator/loglevel/logging_controller.go
```diff
@@ -28,24 +35,36 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+	// if operator operatorSpec OperatorLogLevel is empty, default to "Normal"
+	// TODO: This should be probably done by defaulting the CR field?
 	if len(desiredLogLevel) == 0 {
 		desiredLogLevel = operatorv1.Normal
 	}
 
-	// When the current loglevel is the desired one, do nothing
-	if currentLogLevel == desiredLogLevel {
+	// correct log level is set and it matches the expected log level from operator operatorSpec, do nothing.
+	if isUnknown && currentLogLevel == desiredLogLevel {
```
sttts: `!isUnknown`

mfojtik: ooops.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -660,7 +660,7 @@ func (c *InstallerController) ensureInstallerPod(nodeName string, operatorSpec *
 	}
 
 	args := []string{
-		fmt.Sprintf("-v=%d", loglevel.LogLevelToKlog(operatorSpec.LogLevel)),
+		fmt.Sprintf("-v=%d", loglevel.LogLevelToVerbosity(operatorSpec.LogLevel)),
```
sttts: isn't there the same for the pruner?

sttts: Is this function not also called for the operand? Do we have two of them?

mfojtik:
reapplying LGTM (just fixed the logic in `isUnknown`

PR #739, opened 2020-03-12
Bug 1810528: Allow setting if a CA is required or optional when creating the bundle

tnozicka:
When operators are creating CA bundles from particular CAs, some are required and some are optional. So far all were optional and we managed to deploy kube-scheduler with CA bundle not having kube-apiserver's CAs which is essential for the pod to work.

The question is if we keep the default to be required or optional but required seems safer. The optional (but required) case manifests in races, while if we make required something that doesn't have to be and isn't running at the time we need it we fail in CI and fix it before merging.

If we agree on changing the default from optional to required, I'll let the other operator maintainers know so they are not surprised on bumps.

/cc @deads2k @soltysh 
@openshift/sig-master

### DIFF
--- a/pkg/operator/resourcesynccontroller/core.go
+++ b/pkg/operator/resourcesynccontroller/core.go
@@ -10,6 +10,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/cert"
+	"k8s.io/klog"
 
 	"github.com/openshift/library-go/pkg/crypto"
 )
@@ -18,16 +19,23 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
-			continue
-		}
 		if err != nil {
-			return nil, err
+			if apierrors.IsNotFound(err) && !input.Required {
+				klog.V(2).Infof("Optional ConfigMap %s/%s doesn't exist yet. Skipping.", input.Namespace, input.Name)
+				continue
+			} else {
+				return nil, err
+			}
 		}
 
 		// configmaps must conform to this
 		inputContent := inputConfigMap.Data["ca-bundle.crt"]
 		if len(inputContent) == 0 {
+			if input.Required {
+				return nil, fmt.Errorf("configmap %s/%s doesn't contain 'ca-bundle.crt'", input.Namespace, input.Name)
+			}
+
+			klog.V(2).Infof("Optional ConfigMap %s/%s doesn't contain 'ca-bundle.crt' yet. Skipping.", input.Namespace, input.Name)
 			continue
 		}
 		inputCerts, err := cert.ParseCertsPEM([]byte(inputContent))
--- a/pkg/operator/resourcesynccontroller/interfaces.go
+++ b/pkg/operator/resourcesynccontroller/interfaces.go
@@ -4,6 +4,7 @@ package resourcesynccontroller
 type ResourceLocation struct {
 	Namespace string `json:"namespace"`
 	Name      string `json:"name"`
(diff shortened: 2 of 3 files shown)


tnozicka:
/hold
for proof when we agree on the reviews here

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -4,6 +4,7 @@ package resourcesynccontroller
 type ResourceLocation struct {
 	Namespace string `json:"namespace"`
 	Name      string `json:"name"`
+	Optional  bool   `json:"optional"`
```
deads2k: `Required`.  requiring this is exceptional, not normal.

tnozicka: ok

tnozicka:
proof in https://github.com/openshift/cluster-kube-scheduler-operator/pull/224

tnozicka:
/hold cancel
green e2e

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -18,7 +19,8 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) && !input.Required {
+			klog.V(2).Infof("Optional ConfigMap %s/%s doesn't exist yet. Skipping.", input.Namespace, input.Name)
```
sttts, line 23: even higher than 2?

tnozicka: What level would you like to see it on? I don't think this deserves that level of attention as the CM is optional so it's valid to be missing.

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -4,6 +4,7 @@ package resourcesynccontroller
 type ResourceLocation struct {
```
deads2k, line 4: This struct is also used in the ResourceSyncController. You need to honor the value in that controller to.

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -18,7 +19,8 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) && !input.Required {
```
deads2k, line 22: comment here so it's obvious that the fall through to if err != nil is obvious

tnozicka: I've restructured the code so it's easier to read and doesn't need the comment

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -18,16 +19,23 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
-			continue
-		}
 		if err != nil {
-			return nil, err
+			if apierrors.IsNotFound(err) && !input.Required {
```
deads2k, line 23: kube style avoids this extra nesting.

pkg/operator/resourcesynccontroller/interfaces.go
deads2k: here https://github.com/openshift/library-go/blob/4a44f9500348c607b3ab376637c3126a6081715f/pkg/operator/resourcesynccontroller/resourcesync_controller.go#L155-L156

tnozicka: When source is empty, this reads the cache to see if destination is present and and if it is present, it issues a delete call. In what way would you expect `Required` concept to change that flow?

PR #735, opened 2020-03-11
Rebase kubernetes to v1.18.0-beta.2

damemi:

PR #740, opened 2020-03-12
Prebase 1.18 

damemi:
This has all the changes from https://github.com/openshift/library-go/pull/735, squashed to a reasonable size and pushed to the prebase-1.18 branch


deads2k:
I wonder
/lgtm

PR #741, opened 2020-03-13
resourcesynccontroller: fix unit test flakes

mfojtik:


mfojtik:
/hold

test still flaky :/

PR #744, opened 2020-03-13
config-observers: prune returned config

sttts:
Config observers are supposed to return the existing config in case of error. This is not uniformely implemented. Some observers even return the entire observed config, including fields from other observers.

This PR switch existing config reconstruction trivial by just filter the whole observed config by the keys the observer is interested in. This is a three liner in each observer, super simple and uniform.

### DIFF
--- a/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
+++ b/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
@@ -60,8 +60,7 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 		return observedConfig, errs
 	}
 	if err != nil {
-		errs = append(errs, err)
-		return existingConfig, errs
+		return existingConfig, append(errs, err)
 	}
 
 	cloudProvider := getPlatformName(infrastructure.Status.Platform, recorder)
@@ -88,16 +87,13 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 		sourceLocation = resourcesynccontroller.ResourceLocation{}
 	}
 
-	err = listers.ResourceSyncer().SyncConfigMap(
+	if err := listers.ResourceSyncer().SyncConfigMap(
 		resourcesynccontroller.ResourceLocation{
 			Namespace: c.targetNamespaceName,
 			Name:      "cloud-config",
 		},
-		sourceLocation)
-
-	if err != nil {
-		errs = append(errs, err)
-		return observedConfig, errs
+		sourceLocation); err != nil {
+		return existingConfig, append(errs, err)
 	}
 
 	if len(sourceCloudConfigMap) == 0 {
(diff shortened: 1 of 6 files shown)


pkg/operator/configobserver/featuregates/observe_featuregates.go
```diff
@@ -41,20 +41,13 @@ type featureFlags struct {
 }
 
 // ObserveFeatureFlags fills in --feature-flags for the kube-apiserver
-func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, _ []error) {
+	defer func() {
+		ret = configobserver.Pruned(ret, f.configPath)
```
p0lyn0mial: I think this is what will be returned from this function.

pkg/operator/configobserver/featuregates/observe_featuregates.go
```diff
@@ -70,13 +63,17 @@ func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers
...
+		return existingConfig, errs
 	}
 
 	newConfigValue, err := f.getWhitelistedFeatureNames(configResource)
 	if err != nil {
 		errs = append(errs, err)
-		return prevObservedConfig, errs
+		return existingConfig, errs
+	}
+	currentConfigValue, _, err := unstructured.NestedStringSlice(existingConfig, f.configPath...)
```
p0lyn0mial: shouldn't we then assign to `ret` variable?

p0lyn0mial: all good - https://play.golang.org/p/MMSUo8I1U4m

pkg/operator/configobserver/featuregates/observe_featuregates.go
```diff
@@ -70,13 +63,17 @@ func (f *featureFlags) ObserveFeatureFlags(genericListers configobserver.Listers
...
+	if err != nil {
+		errs = append(errs, err)
```
p0lyn0mial: shouldn't we `return` on `err`?

pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
```diff
@@ -78,12 +61,12 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 	}
 	if err != nil {
 		errs = append(errs, err)
-		return previouslyObservedConfig, errs
+		return existingConfig, errs
```
p0lyn0mial: nit could be changed to `return existingConfig, append(errs, err)`

pkg/operator/configobserver/featuregates/observe_featuregates.go
p0lyn0mial: shouldn't we return `existingConfig` in line `85` in this file? (when we failed to `SetNestedStringSlice`)

pkg/operator/configobserver/featuregates/observe_featuregates.go
sttts: yes

pkg/operator/configobserver/featuregates/observe_featuregates.go
sttts: this case means that the previously observed value was wrongly typed. We have to keep going here to fix that, i.e. not return.

pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
sttts: done

pkg/operator/configobserver/featuregates/observe_featuregates.go
p0lyn0mial: ic, it make sense.

PR #749, opened 2020-03-17
factory: fix appending to caches

mfojtik:


pkg/controller/factory/factory.go
```diff
@@ -136,33 +136,34 @@ func (f *Factory) ToController(name string, eventRecorder events.Recorder) Contr
...
 
 	c := &baseController{
-		name:        name,
-		sync:        f.sync,
-		resyncEvery: f.resyncInterval,
-		syncContext: ctx,
+		name:         name,
+		sync:         f.sync,
+		resyncEvery:  f.resyncInterval,
+		cachesToSync: f.cachesToSync,
```
sttts: `append(nil, f.cachesToSync...)`

mfojtik: fixed

PR #750, opened 2020-03-17
Add CSIDriver to resourceread and resourceapply

bertinatto:
The CSIDriver object contains information about a CSI driver.

Up until recenty, this object was created by the cluster-driver-registrar sidecar container, however, this sidecar has been deprecated. As a result, operators need to create this object when installing a CSI driver.

CC @openshift/storage

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -6,10 +6,12 @@ import (
 	"k8s.io/klog"
 
 	storagev1 "k8s.io/api/storage/v1"
+	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


bertinatto:
@deads2k would you be able to take a look at this? We'd like to use this in the CSI operators.

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -48,3 +50,32 @@ func ApplyStorageClass(client storageclientv1.StorageClassesGetter, recorder eve
...
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	if !*modified {
+		return existingCopy, false, nil
+	}
+
+	if klog.V(4) {
+		klog.Infof("CSIDriver %q changes: %v", required.Name, JSONPatchNoError(existing, existingCopy))
+	}
+
+	actual, err := client.CSIDrivers().Update(existingCopy)
```
deads2k: Do CSI drivers have any immutable fields?  Those can sometimes cause issues.  You don't have to fix them all now, but you can look at what we did for secrets to handle the situation

bertinatto:
Rebased and added contexts to client methods (Create/Update)

PR #751, opened 2020-03-17
factory: improve unit test coverage

mfojtik:
This covers ~89% factory with unit tests.

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -139,11 +139,12 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 	var ok bool
 	syncCtx.queueKey, ok = key.(string)
 	if !ok {
-		klog.Warningf("Queue key is not string: %+v", key)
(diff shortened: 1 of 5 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -74,12 +74,12 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 		}()
 		for i := range c.postStartHooks {
 			hookWg.Add(1)
-			go func(index int) {
+			go func(postStartHookCtx context.Context, index int) {
```
sttts: ctx is no loop variable. No need for aliasing it.

pkg/controller/factory/base_controller.go
```diff
@@ -139,11 +139,12 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 	var ok bool
 	syncCtx.queueKey, ok = key.(string)
 	if !ok {
-		klog.Warningf("Queue key is not string: %+v", key)
+		utilruntime.HandleError(fmt.Errorf("%q controller failed to process key %q (not a string)", c.name, key))
+		return
```
sttts: this `return` is helpful.

PR #752, opened 2020-03-17
Fix up more encryption helpers in prebase branch

damemi:


soltysh:
This is not needed any more.

PR #753, opened 2020-03-19
UnionRevisionLabelPodDeployer

p0lyn0mial:
This PR implements `UnionRevisionLabelPodDeployer` that returns a unified state from multiple distinct deployers. That is: none has reported an error, all have converged, all have observed exactly the same encryption configuration.

Visit https://github.com/openshift/enhancements/blob/master/enhancements/etcd/etcd-encryption-for-separate-oauth-apis.md for more details.

### DIFF
--- a/pkg/operator/encryption/deployer/unionrevisionedpod.go
+++ b/pkg/operator/encryption/deployer/unionrevisionedpod.go
@@ -1,0 +1,79 @@
+package deployer
+
+import (
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
+	"github.com/openshift/library-go/pkg/operator/encryption/statemachine"
+)
+
+// UnionRevisionLabelPodDeployer provides unified state from multiple distinct deployers.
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/deployer/unionrevisionedpod.go
```diff
@@ -0,0 +1,79 @@
...
+// That means:
+//  - none has reported an error
+//  - all have converged
+//  - all have observed exactly the same encryption configuration
+func NewUnionRevisionLabelPodDeployer(delegates ...statemachine.Deployer) *UnionRevisionLabelPodDeployer {
+	return &UnionRevisionLabelPodDeployer{delegates: delegates}
+}
+
+// DeployedEncryptionConfigSecret returns the actual encryption configuration across multiple deployers
+func (d *UnionRevisionLabelPodDeployer) DeployedEncryptionConfigSecret() (secret *corev1.Secret, converged bool, err error) {
```
mfojtik: any reason why to use named return values?

mfojtik:
/lgtm

one nit.

pkg/operator/encryption/deployer/unionrevisionedpod.go
p0lyn0mial: no reason indeed will remove, thx

PR #754, opened 2020-03-23
adds revision controller to APIServerControllerSet

p0lyn0mial:


p0lyn0mial:
an example usage https://github.com/openshift/cluster-authentication-operator/blob/2091248b4b2e56a08544d63f273e0cc86ff98795/pkg/operator2/starter.go#L333

PR #756, opened 2020-03-25
moves encryption controllers to apiservercontrollerset

p0lyn0mial:


p0lyn0mial:
/hold

haven't tested

p0lyn0mial:
this pull deprecates https://github.com/openshift/library-go/pull/676

p0lyn0mial:
to test it in `openshift-apiserver` I need https://github.com/openshift/cluster-openshift-apiserver-operator/pull/345 first

pkg/operator/apiserver/controllerset/apiservercontrollerset.go
```diff
@@ -276,6 +288,62 @@ func (cs *APIServerControllerSet) WithoutRevisionController() *APIServerControll
...
+	component string,
+	encryptedGRs []schema.GroupResource,
+	migrationClient kubemigratorclient.Interface,
+	kubeClient kubernetes.Interface,
+	apiServerClient configv1client.APIServerInterface,
+	secretsClient corev1client.SecretsGetter,
+	apiServerInformer configv1informers.APIServerInformer,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	resourceSyncController resourcesynccontroller.ResourceSyncer,
+	deployer statemachine.Deployer,
```
p0lyn0mial: `deployer` is optional, a default one will be created if `nil`

for now, it will be different for `OAS-O` (`UnionRevisionLabelPodDeployer`) and `CAO`

pkg/operator/apiserver/controllerset/apiservercontrollerset.go
```diff
@@ -276,6 +288,62 @@ func (cs *APIServerControllerSet) WithoutRevisionController() *APIServerControll
...
+
+		deployer, err = encryptiondeployer.NewRevisionLabelPodDeployer("revision", component, kubeInformersForNamespaces, resourceSyncController, kubeClient.CoreV1(), kubeClient.CoreV1(), nodeProvider)
+		if err != nil {
+			err = fmt.Errorf("unable to create encryption controllers because failed to create a deployer, due to %v", err)
+			cs.unhandledErrors = append(cs.unhandledErrors, err)
+			return cs
+		}
+	}
+
+	migrationInformer := migrationv1alpha1informer.NewSharedInformerFactory(migrationClient, time.Minute*30)
```
stlaz: why create the informer here instead of it being passed as an argument (and thus not having to special-case it for informer start)?

p0lyn0mial: if we do that, then clients would have to deal with "migrator" - wanted to avoid that as they don't need to know about it. Does it make sense?

stlaz: They need to pass the migration client anyway, I'd say pass it as an argument in the name of uniform informer handling

p0lyn0mial: okay, but I will move `deployer` as well, so that dependencies are created in the main program.

p0lyn0mial:
/hold cancel

it works https://github.com/openshift/cluster-openshift-apiserver-operator/pull/346

stlaz:
/lgtm
great to see it work \o/

PR #757, opened 2020-03-26
factory: add WithSyncDegradedOnError that allows handling degraded for simple sync()

mfojtik:
Introduce `WithSyncDegradedOnError(operatorClient)` to factory which in case of `sync()` function return error set the `ControllerNameDegraded` condition automatically (and remove it when the sync does not return any error).

This is useful for simple operator controller loops and minimize the boiler plate in sync() calls. Also David called this idiot-proof :-)

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -10,16 +10,22 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
 // baseController represents generic Kubernetes controller boiler-plate
 type baseController struct {
-	name           string
-	cachesToSync   []cache.InformerSynced
-	sync           func(ctx context.Context, controllerContext SyncContext) error
-	syncContext    SyncContext
-	resyncEvery    time.Duration
-	postStartHooks []PostStartHook
+	name               string
+	cachesToSync       []cache.InformerSynced
+	sync               func(ctx context.Context, controllerContext SyncContext) error
+	syncContext        SyncContext
+	syncDegradedClient operatorv1helpers.OperatorClient
+	resyncEvery        time.Duration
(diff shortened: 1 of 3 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -128,6 +133,33 @@ func (c *baseController) runWorker(queueCtx context.Context) {
...
+	}
+	if err != nil {
+		_, _, updateErr := v1helpers.UpdateStatus(c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    c.name + "Degraded",
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "SyncError",
+			Message: err.Error(),
+		}))
+		if updateErr != nil {
+			syncCtx.Recorder().Warning(c.name+"UpdatingStatus", updateErr.Error())
```
sttts: won't this spam the apiserver with events?

pkg/controller/factory/factory.go
```diff
@@ -122,6 +124,13 @@ func (f *Factory) WithSyncContext(ctx SyncContext) *Factory {
 	return f
 }
 
+// WithSyncDegradedOnError encapsulate the controller sync() function, so when this function return an error, the operator client
+// is used to set the degraded condition to (eg. "ControllerFooDegraded"). The degraded condition name is set based on the controller name.
+func (f *Factory) WithSyncDegradedOnError(operatorClient operatorv1helpers.OperatorClient) *Factory {
+	f.operatorClient = operatorClient
```
sttts: rename operatorClient to something SyncDegradedOnError specific (this var drives this feature, so it should be named accordingly).

pkg/controller/factory/base_controller.go
mfojtik: only on conflicts, but I don't this it will do much spam... on the other hand, i would want to know if the controller can't update its status.

pkg/controller/factory/factory.go
mfojtik: renamed.

pkg/controller/factory/base_controller.go
sttts: Am not a fan of this (as you might have noticed in previous reviews :-) ). It would be fine if this is an ongoing issue.

sttts: Or in other words: events are for edges, not for levels.

PR #759, opened 2020-03-27
Fix StorageClasses comparison

bertinatto:
StorageClasses are currently always updated on every operator resync.

The reason is `ApplyStorageClass()` does a `DeepEqual` comparison between the SCs, however, 

1) creation time are always different and 
2) the object got from the API doesn't contain `TypeMeta` set.

As a result, `DeepEqual()` returns `false` and the SC is always updated.

CC @openshift/storage

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -30,25 +30,29 @@ func ApplyStorageClass(client storageclientv1.StorageClassesGetter, recorder eve
 		return nil, false, err
 	}
 
+	// First, let's compare ObjectMeta from both objects
 	modified := resourcemerge.BoolPtr(false)
 	existingCopy := existing.DeepCopy()
-
 	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
-	contentSame := equality.Semantic.DeepEqual(existingCopy, required)
+
+	// Second, let's compare the other fields. StorageClass doesn't have a spec and we don't
+	// want to miss fields, so we have to copy required to get all fields
+	// and then overwrite ObjectMeta and TypeMeta from the original.
+	requiredCopy := required.DeepCopy()
+	requiredCopy.ObjectMeta = *existingCopy.ObjectMeta.DeepCopy()
+	requiredCopy.TypeMeta = existingCopy.TypeMeta
+
+	contentSame := equality.Semantic.DeepEqual(existingCopy, requiredCopy)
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourceapply/storage.go
```diff
@@ -34,6 +34,15 @@ func ApplyStorageClass(client storageclientv1.StorageClassesGetter, recorder eve
 	existingCopy := existing.DeepCopy()
 
 	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+
+	// Now that we copied everything that matters from required.ObjectMeta, we
+	// should reset required.ObjectMeta so a DeepEqual() comparison is possible
+	required.ObjectMeta = *existingCopy.ObjectMeta.DeepCopy()
```
deads2k: You don't want to mutate your inputs.  I think I see the bug a few lines down: `existingCopy = required.DeepCopy()`.  I see why you did it, but how about doing the required.DeepCopy() up here and then something like this

```go
//storage class doesn't have a spec and we don't want to miss fields.  We have to copy the required to get all fields and then overwrite objectmeta and typemeta from the original
existingCopy := required.DeepCopy()
existingCopy.ObjectMeta = existing.ObjectMeta.DeepCopy()
existingCopy.TypeMeta = existing.TypeMeta.DeepCopy()
```

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -34,6 +34,15 @@ func ApplyStorageClass(client storageclientv1.StorageClassesGetter, recorder eve
...
+
+	// We also need to reset required.TypeMeta because existingCopy.TypeMeta,
+	// that's comming from the apiserver, isn't set (see https://issues.k8s.io/3030)
+	required.TypeMeta = existingCopy.TypeMeta
+
 	contentSame := equality.Semantic.DeepEqual(existingCopy, required)
```
deads2k: Then this line has to be updated to `contentSame := equality.Semantic.DeepEqual(existing, required)`

pkg/operator/resource/resourceapply/storage.go
bertinatto: `existingCopy.ObjectMeta = existing.ObjectMeta.DeepCopy()`: here we'd miss the "merged" ObjectMeta from line 36 above.

I tried to incorporate your suggestions the best I could. I think the main reason this code is so confusing is that we need to mutate the objects at each step.

PR #760, opened 2020-03-28
add WithCustomInstaller method to Builder interface

alaypatel07:
There is a need in cluster etcd operator to check if it is safe
to install a revision on a node during install time.

Will allow to solve openshift/cluster-etcd-operator#278 with 
something like openshift/cluster-etcd-operator#281

### DIFF
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -7,11 +7,13 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
 	"github.com/openshift/library-go/pkg/operator/revisioncontroller"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controllers.go
```diff
@@ -124,6 +129,15 @@ func (b *staticPodOperatorControllerBuilder) WithCerts(certDir string, certConfi
 
 func (b *staticPodOperatorControllerBuilder) WithInstaller(command []string) Builder {
 	b.installCommand = command
+	b.installerPodMutationFunc = func(pod *corev1.Pod, nodeName string, operatorSpec *operatorv1.StaticPodOperatorSpec, revision int32) error {
+		return nil
+	}
+	return b
+}
+
+func (b *staticPodOperatorControllerBuilder) WithMutatingInstallerPod(command []string, installerPodMutationFunc installer.InstallerPodMutationFunc) Builder {
```
sttts: what does this do? It needs docs.

sttts: MutatingInstallerPod sounds like the pod is mutating something. In reality we pass a func which mutates, right? `WithMutatedInstaller` or `WithCustomInstaller`?

alaypatel07: @sttts yes, WithCustomInstaller with do to explain makes more sense, will update

pkg/operator/staticpod/controllers.go
```diff
@@ -25,6 +25,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	corev1 "k8s.io/api/core/v1"
```
sttts: order

hexfusion:
/title Bug 1817028: add WithCustomInstaller method to Builder interface

hexfusion:
/retitle Bug 1817028: add WithCustomInstaller method to Builder interface

PR #763, opened 2020-03-30
changes UnionRevisionLabelPodDeployer to deal with `nil` secrets

p0lyn0mial:


pkg/operator/encryption/deployer/unionrevisionedpod.go
```diff
@@ -36,6 +36,9 @@ func (d *UnionRevisionLabelPodDeployer) DeployedEncryptionConfigSecret() (secret
 		if !converged || err != nil {
 			return nil, converged, err
 		}
+		if secret == nil {
```
p0lyn0mial: I will use it in `OAS-O` - the custom `deployer` will not return any resources when the annotation was `set`.

sttts: What do we see if the oauth-apiserver secret does not exist?

p0lyn0mial: the same - https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/deployer/revisionedpod.go#L116

p0lyn0mial:
```
// if encryption is not enabled at this revision or the secret was deleted, we should not error
if errors.IsNotFound(err) {
	return nil, true, nil
}
```

sttts: My thought was that we might want to block COAO from progressing if its secret exists, but CAO's secret does not.

p0lyn0mial: if https://github.com/openshift/cluster-openshift-apiserver-operator/pull/348 makes sense then we need to merge this pull.

PR #766, opened 2020-04-03
Bug 1826230: Switch from jteeuwen/go-bindata to go-bindata/go-bindata

marun:
This change is for consistency with upstream kubernetes so there is one less difference when rebasing.

It's necessary to bump build-machinery-go and api since they also depend on go-bindata.


marun:
Marking WIP until https://github.com/openshift/build-machinery-go/pull/13 merges.

PR #767, opened 2020-04-06
reaper: Add a periodic reaper funcion

mrunalp:
This commit adds a periodic reaper function that is intended
to be used when the go process uses the os/exec package to
launch child processes. The period argument could be adjusted
according to how long the os/exec calling code waits to
gather exit status from child processes.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>


mrunalp:
@smarterclayton ptal. If we can confirm that this works well with router, I can optimize the code further for parsing /proc. Also, we can consolidate both types of reaping with ReaperOptions and a single function or keep it separate like the current approach.

smarterclayton:
/hold

Can you test vendor this into a router PR and look at the log output?  If this fixes it, we should get none of the "can't wait for process anymore"

mrunalp:
@smarterclayton sure, I can do that.

mrunalp:
Opened https://github.com/openshift/router/pull/111

mrunalp:
Updated.

mrunalp:
I have a couple of logs remaining that I will take out after one more round of testing on the router PR.

mrunalp:
Updated the PR. This is ready.

pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,75 @@
...
+// from a pid 1 process.
+func StartReaper(period time.Duration) {
+	if os.Getpid() == 1 {
+		go func() {
+			var zs []int
+			var err error
+			for {
+				zs, err = parseProcForZombies()
+				if err != nil {
+					klog.V(4).Infof(err.Error())
```
smarterclayton, line 62: Give this an error prefix so we know it was from reaping

pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,75 @@
...
+			}
+		}
+	}
+
+	return zombies, nil
+}
+
+// StartReaper starts a goroutine to reap processes periodically if called
+// from a pid 1 process.
+func StartReaper(period time.Duration) {
```
smarterclayton, line 54: If period is less than or equal to zero set it to five seconds and add a comment above.  Also describe why you might change the default period (if you create tens of thousands of defunct processes every five seconds).

mrunalp: Added.

mrunalp:
@smarterclayton could you take another look?

pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+		}
+	}
+
+	return zombies, nil
+}
+
+// StartReaper starts a goroutine to reap processes periodically if called
+// from a pid 1 process.
+// If period is less than 5 seconds, then it is set to 5 seconds.
```
smarterclayton, line 54: I would have it be only when 0, not when less than five.  If you have a very fast subprocess creating app, you might need to set it faster.

pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,85 @@
...
+		if period.Seconds() < minReaperPeriodSeconds {
+			period = minReaperPeriodSeconds * time.Second
+		}
+		go func() {
+			var zs []int
+			var err error
+			for {
+				zs, err = parseProcForZombies()
+				if err != nil {
+					klog.Errorf("Reaper: failed to parse zombies: %v", err)
```
smarterclayton, line 72: `failed to parse proc filesystem to find processes to terminate`

klog will already set the filename so `reaper: ` is unnecessary

pkg/proc/proc_linux.go
```diff
@@ -0,0 +1,85 @@
...
+// If period is less than 5 seconds, then it is set to 5 seconds.
+// A caller can adjust the period depending on how many and how frequently zombie
+// processes are created and need to reaped. A process generating lots of zombie
+// processes should set the period to 5 seconds. A process that has fewer zombie
+// processes and can tolerate zombies being around longer could set it to a higher
+// value.
+func StartReaper(period time.Duration) {
+	if os.Getpid() == 1 {
+		const minReaperPeriodSeconds = 5
+		if period.Seconds() < minReaperPeriodSeconds {
```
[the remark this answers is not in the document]
mrunalp: sounds good! Default to 5.

PR #768, opened 2020-04-07
Fix ApplySecret for Type changes

tnozicka:
Kubernetes validation silently ignores secret changes and doesn't return an error.
https://github.com/kubernetes/kubernetes/blob/98e65951dccfd40d3b4f31949c2ab8df5912d93e/pkg/apis/core/validation/validation.go#L5048
This leads to the resource sync controller constantly trying to update to the desired state. (mutating admission vs. DeepEqual)

/cc @deads2k

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -273,6 +273,11 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 
 	existingCopy.Type = required.Type
 
+	// Server defaults some values and we need to do it as well or it will never equal.
+	if existingCopy.Type == "" {
+		existingCopy.Type = corev1.SecretTypeOpaque
+	}
+
 	if equality.Semantic.DeepEqual(existingCopy, existing) {
 		return existing, false, nil
 	}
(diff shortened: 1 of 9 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -280,14 +280,23 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 	if klog.V(4) {
 		klog.Infof("Secret %s/%s changes: %v", required.Namespace, required.Name, JSONPatchSecretNoError(existing, existingCopy))
 	}
-	actual, err := client.Secrets(required.Namespace).Update(context.TODO(), existingCopy, metav1.UpdateOptions{})
-	reportUpdateEvent(recorder, existingCopy, err)
```
deads2k, line 283: You need to properly default empty with what the the server does.

tnozicka: added the defaulting

tnozicka:
proof in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/392

tnozicka:
the proof is green

PR #771, opened 2020-04-14
Add WithVersion() to command builder that reports binary version to prometheus

mfojtik:
This helper will avoid making mistakes when copying the pkg/version content for control plane operators and also templatize the description.

It will also be easier to replace the legacyregistry from this point in future.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -5,12 +5,16 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/version"
 	genericapiserver "k8s.io/apiserver/pkg/server"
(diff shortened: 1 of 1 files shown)


deads2k:
should the command builder bit have .WithVersion(version.Get)?

/lgtm
/hold

hold for question.  I think it would smooth usage.

pkg/controller/controllercmd/builder.go
```diff
@@ -134,6 +138,23 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	return b
 }
 
+// WithVersion accepts a getting that provide binary version information that is used to report build_info information to prometheus
+func (b *ControllerBuilder) WithVersion(versionGetter func() version.Info) *ControllerBuilder {
```
deads2k, line 142: I think you just need the version.Info

pkg/controller/controllercmd/builder.go
```diff
@@ -134,6 +138,23 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
...
+	buildInfo := metrics.NewGaugeVec(
+		&metrics.GaugeOpts{
+			Name: strings.Replace(b.componentNamespace, "-", "_", -1) + "_build_info",
```
deads2k, line 145: you want to do this part during the actual .Run so you're guaranteed this componentNamespace is set.

mfojtik:
/retest

both infra flakes

PR #773, opened 2020-04-16
Update Apply{DaemonSet,Deployment} to rely on a hash of the spec

marun:
This removes the need for the caller to supply an expected generation and be able to force a rollout. Any state that is not present in the spec should instead be added as an annotation so that a rollout will be triggered when the annotation changes.

This is intentionally a breaking change to encourage switching to the revised function. Callers who want to continue using the previous (deprecated) implementation will need to change the name of the function they are calling to `ApplyDeploymentWithForce`.

Canaries:
 - https://github.com/openshift/cluster-authentication-operator/pull/272
 - https://github.com/openshift/cluster-openshift-apiserver-operator/pull/354

### DIFF
--- a/pkg/operator/resource/resourceapply/apps.go
+++ b/pkg/operator/resource/resourceapply/apps.go
@@ -2,6 +2,9 @@ package resourceapply
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
 
 	"k8s.io/klog"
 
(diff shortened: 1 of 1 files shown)


marun:
> @marun also, we probably want the behaviour of `ApplyDaemonSet()` to match `ApplyDeployment()`.

For sure. Want to make sure the approach is sound first.
> Instead of doing a breaking change, what about introducing a new function adds the annotation and calls `Apply[Deployment|DaemonSet]()`?

As per the PR description, the breakage is intentional to force callers to choose between keeping the old behavior (for now) or adapting to the new. I don't think this is too much to ask - a quick survey of the code of existing operators shows that majority have only a single call to `ApplyDeployment` while some have at most a few.

jsafrane:
While I am in favor of this change, it breaks unit tests  - a new annotations appears and needs to be be added on operator input (when the tests checks that operator does nothing when nothing is expected) and ignored when comparing expected / actual objects on output.

Exposing the hash computation as a public function would help to fix the unit tests.

bertinatto:
> > @marun also, we probably want the behaviour of `ApplyDaemonSet()` to match `ApplyDeployment()`.
> 
> For sure. Want to make sure the approach is sound first.
> 
> > Instead of doing a breaking change, what about introducing a new function adds the annotation and calls `Apply[Deployment|DaemonSet]()`?
> 
> As per the PR description, the breakage is intentional to force callers to choose between keeping the old behavior (for now) or adapting to the new. I don't think this is too much to ask - a quick survey of the code of existing operators shows that majority have only a single call to `ApplyDeployment` while some have at most a few.

Right, I guess I'm not a big fan of a breaking change here, but we'll use whatever approach you decide.

pkg/operator/resource/resourceapply/apps.go
```diff
@@ -15,13 +18,42 @@ import (
...
+// should set an annotation representing that external state.
+func ApplyDeployment(client appsclientv1.DeploymentsGetter, recorder events.Recorder,
+	required *appsv1.Deployment) (*appsv1.Deployment, bool, error) {
+
+	// Record the hash of the spec in an annotation to prompt rollout if it changes.
+	jsonBytes, err := json.Marshal(required.Spec)
+	if err != nil {
+		return nil, false, err
+	}
+	specHash := fmt.Sprintf("%x", sha256.Sum256(jsonBytes))
```
bertinatto, line 35: The deployment controller creates the replicaSet with a hash in its name. This hash is calculated from the deployment's pod template [0]. Same for the daemonSet controller.

It's not the same as we're doing here, as we're hashing the whole deployment's spec, but it'd be nice to use a similar approach. For that, we could either:

1) Hash the `deployment.Spec.Template` as well, and use the public available utility function [1] that does that. The downside is that `ApplyDeployment()` wouldn't update the deployment if we change the number of replicas, for instance.

2) Create our own hashing function based on [1].

Does this make sense?

[0] https://github.com/kubernetes/kubernetes/blob/98e65951dccfd40d3b4f31949c2ab8df5912d93e/pkg/controller/deployment/sync.go#L189
[1] https://github.com/kubernetes/kubernetes/blob/98e65951dccfd40d3b4f31949c2ab8df5912d93e/pkg/controller/controller_utils.go#L1129

marun: deployment controller: hash to come up with a unique replicaset name
operator: hash to ensure that a change in intent results in a rollout

Given that the use cases are different, why would it be desirable to match the implementation of the deployment controller?

marun:
> While I am in favor of this change, it breaks unit tests - a new annotations appears and needs to be be added on operator input (when the tests checks that operator does nothing when nothing is expected) and ignored when comparing expected / actual objects on output.

I think a unit test that cares about an internal detail of ApplyDeployment is a brittle test. Why not just strip the annotation for the purposes of comparison?

jsafrane:
The unit test needs to *add* the annotation to fake API server object, so when ApplyDeployment gets it (as `existing`), the annotation is there.

Stripping the annotation *after* the test, when comparing result is a good idea, but it's orthogonal to ^.

pkg/operator/resource/resourceapply/apps.go
```diff
@@ -15,13 +18,42 @@ import (
...
 )
 
-// ApplyDeployment merges objectmeta and requires matching generation. It returns the final Object, whether any change as made, and an error
-func ApplyDeployment(client appsclientv1.DeploymentsGetter, recorder events.Recorder, required *appsv1.Deployment, expectedGeneration int64,
+const specHashAnnotation = "operator.openshift.io/spec-hash"
+
+// ApplyDeployment merges ObjectMeta of the provided deployment with an existing one
+// if it exists and updates the API if the deployment spec and metadata differ. To
+// ensure an update in response to state external to the deployment spec, the caller
+// should set an annotation representing that external state.
```
deads2k, line 26: give an example of such an annotation

pkg/operator/resource/resourceapply/apps.go
```diff
@@ -15,13 +18,42 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
-// ApplyDeployment merges objectmeta and requires matching generation. It returns the final Object, whether any change as made, and an error
-func ApplyDeployment(client appsclientv1.DeploymentsGetter, recorder events.Recorder, required *appsv1.Deployment, expectedGeneration int64,
+const specHashAnnotation = "operator.openshift.io/spec-hash"
+
+// ApplyDeployment merges ObjectMeta of the provided deployment with an existing one
+// if it exists and updates the API if the deployment spec and metadata differ. To
```
deads2k, line 24: if the deployment spec or metadata differ from the previously required spec and metadata.  To be reliable, the input of the required spec from the operator should be stable.  It does not need to set all fields, since some fields are defaulted server-side.  

Detection of spec drift from intent by other actors is determined by generation, not by spec comparison.

pkg/operator/resource/resourceapply/apps.go
```diff
@@ -15,13 +18,42 @@ import (
...
+	if err != nil {
+		return nil, false, err
+	}
+	specHash := fmt.Sprintf("%x", sha256.Sum256(jsonBytes))
+	if required.Annotations == nil {
+		required.Annotations = map[string]string{}
+	}
+	required.Annotations[specHashAnnotation] = specHash
+
+	return ApplyDeploymentWithForce(client, recorder, required, 0, false)
```
deads2k, line 41: I think you need an expected generation in order to know if another actor mutated the spec.  The hash of the input spec lets you know if the operator decided a new value was needed (--loglevel for instance) and the generation lets you know if an external actor changed the deployment spec. 

Because defaults can change on the server (new field appearing in a new level of kube for instance), you cannot reliably read and compare a previous hash.

deads2k:
> > > @marun also, we probably want the behaviour of `ApplyDaemonSet()` to match `ApplyDeployment()`.
> > 
> > 
> > For sure. Want to make sure the approach is sound first.
> > > Instead of doing a breaking change, what about introducing a new function adds the annotation and calls `Apply[Deployment|DaemonSet]()`?
> > 
> > 
> > As per the PR description, the breakage is intentional to force callers to choose between keeping the old behavior (for now) or adapting to the new. I don't think this is too much to ask - a quick survey of the code of existing operators shows that majority have only a single call to `ApplyDeployment` while some have at most a few.
> 
> Right, I guess I'm not a big fan of a breaking change here, but we'll use whatever approach you decide.

@bertinatto we're sensitive to the fact that this causes some pain, but working with @jsafrane I think we've found a better pattern for handling updates in a reliable way that works how our callers really expect them to run.

To make the change "easy" to adopt, we'll first make the change obvious and @marun should point to the renamed legacy method for a release before we remove or hide it.  Given that the old method defied expectations, I think it's better in this case to break compilation and make callers look at the doc briefly to decide about updating right then or switching to the older method.

pkg/operator/resource/resourceapply/apps.go
marun: Done

pkg/operator/resource/resourceapply/apps.go
marun: Done

pkg/operator/resource/resourceapply/apps.go
marun: Done

marun:
> The unit test needs to _add_ the annotation to fake API server object, so when ApplyDeployment gets it (as `existing`), the annotation is there.
> 
> Stripping the annotation _after_ the test, when comparing result is a good idea, but it's orthogonal to ^.

Understood. Updated to expose `SetSpecHashAnnotation`.

pkg/operator/resource/resourceapply/apps.go
```diff
@@ -15,13 +18,100 @@ import (
...
+// revised to set that external state as an annotation e.g.
+//
+//   myoperator.openshift.io/my-resource: <resourceVersion>
+//
+// - Update the call to use ApplyDeploymentWithForce. This is available as a temporary measure
+// but the method is deprecated and will be removed in the future.
+func ApplyDeployment(client appsclientv1.DeploymentsGetter, recorder events.Recorder,
+	required *appsv1.Deployment, expectedGeneration int64) (*appsv1.Deployment, bool, error) {
+
+	err := SetSpecHashAnnotation(&required.ObjectMeta, required.Spec)
```
deads2k, line 94: need to make a copy of required so you don't mutate your input.

marun: Done. Also added a commit to ensure ApplyDeploymentWithForce and ApplyDaemonset copy before mutating.

marun:
Added a new commit to perform the same change to ApplyDaemonSet.

PR #775, opened 2020-04-21
Bug 1826259: fix ingress CA cert location

soltysh:
This is coming from https://github.com/openshift/cluster-ingress-operator/pull/377

/assign @sttts @mfojtik


sttts:
This only react to a change openshift-config-managed resource. We should start a discussion whether we broke an API with it. This PR is fine though as reaction.

/approve

PR #776, opened 2020-04-21
Remove bitbucket.org/ww/goautoneg dependency

alvaroaleman:
It forces ppl to install mercurial. Use instead munnerz github copy of
it.


sttts:
Is the copy a fork? Is it maintained?

alvaroaleman:
> Is the copy a fork? Is it maintained?

It is a fork in that it has one additional patch: https://github.com/munnerz/goautoneg/pull/3 . WRT maintenance, the original hasn't seen any change since 2012: https://bitbucket.org/ww/goautoneg/src/default/

`github.com/munnerz/goautoneg` is already a transient dependency in this repo because its used in  k8s.io/apiserver: https://github.com/kubernetes/kubernetes/blob/fc9d1741022a8b367e8f91236d0c41111d47684d/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/negotiation/negotiate.go#L25

PR #777, opened 2020-04-21
events: make kube broadcaster shutdown gracefully and tune correlator so we don't loose events

mfojtik:
This change will wire the broadcaster `Shutdown()` function into library-go event recorder.
This is later facilitated in controllercmd builder, where this is called when the binary leader election change.

In addition, this change will provide more fine-tuned correlator options that operators which send a lot of events and don't want to loose events should use. These are default now for controllercmd builder based operators.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	"k8s.io/client-go/tools/record"
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog"
@@ -64,6 +65,7 @@ type ControllerBuilder struct {
 	leaderElection          *configv1.LeaderElection
 	fileObserver            fileobserver.Observer
 	fileObserverReactorFn   func(file string, action fileobserver.ActionType) error
+	eventRecorderOptions    record.CorrelatorOptions
 
 	startFunc          StartFunc
 	componentName      string
@@ -175,6 +177,14 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
 	return b
 }
 
+// WithEventRecorderOptions allows to override the default Kubernetes event recorder correlator options.
+// This is needed if the binary is sending a lot of events.
+// Using events.DefaultOperatorEventRecorderOptions here makes a good default for normal operator binary.
+func (b *ControllerBuilder) WithEventRecorderOptions(options record.CorrelatorOptions) *ControllerBuilder {
+	b.eventRecorderOptions = options
+	return b
+}
+
 // Run starts your controller for you.  It uses leader election if you asked, otherwise it directly calls you
 func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstructured) error {
 	clientConfig, err := b.getClientConfig()
@@ -195,7 +205,7 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 	if err != nil {
 		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
 	}
-	eventRecorder := events.NewKubeRecorder(kubeClient.CoreV1().Events(namespace), b.componentName, controllerRef)
+	eventRecorder := events.NewKubeRecorderWithOptions(kubeClient.CoreV1().Events(namespace), b.eventRecorderOptions, b.componentName, controllerRef)
 
 	// if there is file observer defined for this command, add event into default reaction function.
 	if b.fileObserverReactorFn != nil {
@@ -298,6 +308,7 @@ func (b ControllerBuilder) getOnStartedLeadingFunc(controllerContext *Controller
 
 		select {
 		case <-ctx.Done(): // context closed means the process likely received signal to terminate
+			controllerContext.EventRecorder.Shutdown()
 		case <-stoppedCh:
 			// if context was not cancelled (it is not "done"), but the startFunc terminated, it means it terminated prematurely
 			// when this happen, it means the controllers terminated without error.
--- a/pkg/operator/events/recorder_upstream.go
+++ b/pkg/operator/events/recorder_upstream.go
@@ -2,22 +2,32 @@ package events
 
 import (
 	"fmt"
-
-	"k8s.io/klog"
+	"strings"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes/scheme"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog"
 )
 
-// NewKubeRecorder returns new event recorder.
-func NewKubeRecorder(client corev1client.EventInterface, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
+// NewKubeRecorder returns new event recorder with tweaked correlator options.
+func NewKubeRecorderWithOptions(client corev1client.EventInterface, options record.CorrelatorOptions, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
 	return (&upstreamRecorder{
 		client:            client,
 		component:         sourceComponentName,
 		involvedObjectRef: involvedObjectRef,
+		options:           options,
+		fallbackRecorder:  NewRecorder(client, sourceComponentName, involvedObjectRef),
 	}).ForComponent(sourceComponentName)
+}
+
+// NewKubeRecorder returns new event recorder with default correlator options.
+func NewKubeRecorder(client corev1client.EventInterface, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
+	return NewKubeRecorderWithOptions(client, record.CorrelatorOptions{}, sourceComponentName, involvedObjectRef)
 }
 
 // upstreamRecorder is an implementation of Recorder interface.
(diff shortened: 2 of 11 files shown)


deads2k:
/hold

this will cause panics if we try to write events after shutdown.  You probably don't want that.

mfojtik:
@deads2k this will flush what is in sink and waits until all those events are recorded... it is true that "new" events that arrived after we called shutdown won't make it... but I dunno how can we stop them coming. we can time the shutdown as the last thing we call AFTER all controllers are down?

deads2k:
> @deads2k this will flush what is in sink and waits until all those events are recorded... it is true that "new" events that arrived after we called shutdown won't make it... but I dunno how can we stop them coming. we can time the shutdown as the last thing we call AFTER all controllers are down?

The issue isn't that we don't get them. The issue is that the event write will panic because the incoming channel is closed.

mfojtik:
@deads2k proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/837

I made some tweaks:

* During shutdown we switch to normal client, so we keep creating events even if the broadcaster is flushing
* I added prometheus counter for events, so in future we can check how many events we send and how many events we see
* I tweaked the upstream correlator to increase the throughput from 10 events with same reason/component per 10 minutes  to 50 events per same reason/component per 5 minutes.

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +30,60 @@ type upstreamRecorder struct {
...
+	// in that case we create the events directly.
+	fallbackRecorder Recorder
+}
+
+var eventsCounterMetric = metrics.NewCounterVec(&metrics.CounterOpts{
+	Subsystem:      "event_recorder",
+	Name:           "total_events_count",
+	Help:           "Total count of events processed by this event recorder per involved object",
+	StabilityLevel: metrics.ALPHA,
+}, []string{"namespace", "name"})
```
mfojtik, line 48: @deads2k i think this will have reasonable cardinality and can show us how many events we receive per operator... i wonder if we want to break this into warnings vs. normal events :-)

i imagine an alert that can fire if we see "abnormal" amount of "warning" events in a period of time...

mfojtik: cc @smarterclayton

mfojtik:
/retest

found typo in resourcesynccontroller unit test: https://github.com/openshift/library-go/pull/780

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +30,58 @@ type upstreamRecorder struct {
...
-	broadcaster := record.NewBroadcaster()
+	newRecorderForComponent := upstreamRecorder{
+		client:            r.client,
+		fallbackRecorder:  r.fallbackRecorder.WithComponentSuffix(componentName),
+		involvedObjectRef: r.involvedObjectRef,
+		shuttingDown:      r.shuttingDown,
+	}
+	// tweak the event correlator, so we don't loose important events.
+	broadcaster := record.NewBroadcasterWithCorrelatorOptions(record.CorrelatorOptions{
+		BurstSize: 50,        // default: 25 (change allows a single source to send 50 events about object per minute)
```
deads2k, line 65: this needs to be configurable so we can choose something good for our operators.

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +30,58 @@ type upstreamRecorder struct {
...
+		QPS:       1. / 120., // default: 1/300 (change allows refill rate to 1 new event every 2 minutes)
```
deads2k, line 66: this isn't enough.  To see how many we're dumping, I think we want 1 qps or so with a 60 burst

pkg/operator/events/recorder_upstream.go
mfojtik: @deads2k who not pick something good here for all our operators? ;-)

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +30,58 @@ type upstreamRecorder struct {
...
-	newRecorderForComponent := *r
-	broadcaster := record.NewBroadcaster()
+	newRecorderForComponent := upstreamRecorder{
+		client:            r.client,
+		fallbackRecorder:  r.fallbackRecorder.WithComponentSuffix(componentName),
+		involvedObjectRef: r.involvedObjectRef,
+		shuttingDown:      r.shuttingDown,
+	}
+	// tweak the event correlator, so we don't loose important events.
+	broadcaster := record.NewBroadcasterWithCorrelatorOptions(record.CorrelatorOptions{
```
deads2k, line 64: you forgot to modify the key func.  without that you'll still be combining I think

mfojtik: i think we want to combine? so the default key function seems fine? or we never want to combine events with same reason?

mfojtik:
@deads2k updated, made it configurable.

mfojtik:
updated the proof as well: https://github.com/openshift/cluster-kube-apiserver-operator/pull/837

pkg/controller/controllercmd/builder.go
```diff
@@ -175,6 +177,14 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
 	return b
 }
 
+// WithEventRecorderOptions allows to override the default Kubernetes event recorder correlator options.
+// This is needed if the binary is sending a lot of events.
+// Using events.DefaultOperatorEventRecorderOptions here makes a good default for normal operator binary.
+func (b *ControllerBuilder) WithEventRecorderOptions(options record.CorrelatorOptions) *ControllerBuilder {
```
deads2k, line 183: I recommend providing a function that produces these options like `RecommendedClusterSingletonCorrelatorOptions()` as  starting point for mutation.

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +31,63 @@ type upstreamRecorder struct {
...
 	involvedObjectRef *corev1.ObjectReference
+	options           record.CorrelatorOptions
+
+	// shuttingDown indicates that the broadcaster for this recorder is being shut down
+	shuttingDown  bool
+	shutdownMutex sync.RWMutex
+
+	// fallbackRecorder is used when the kube recorder is shutting down
+	// in that case we create the events directly.
+	fallbackRecorder Recorder
```
deads2k, line 42: oh what a wicked web we weave

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +31,63 @@ type upstreamRecorder struct {
...
+}
+
+var DefaultOperatorEventRecorderOptions = record.CorrelatorOptions{
```
deads2k, line 45: still forgot the keyFunc that needs to include the message to avoid de-duping different messages like we have today.

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +31,63 @@ type upstreamRecorder struct {
...
+	BurstSize: 60,       // default: 25 (change allows a single source to send 50 events about object per minute)
+	QPS:       1. / 60., // default: 1/300 (change allows refill rate to 1 new event every 2 minutes)
```
deads2k, line 47: I think is just `1.0` for one per second.  We want them all!

pkg/operator/events/recorder_upstream.go
deads2k: > i think we want to combine? so the default key function seems fine? or we never want to combine events with same reason?

we never want to combine events with different messages.

pkg/operator/events/recorder_upstream.go
mfojtik: i can just log the events that come after shutdown... i think the chance we leak events after shutdown is triggered is really small (the window is basically ~1-2s, but this makes sure we don't miss any event at all...

pkg/operator/events/recorder_upstream.go
mfojtik: done

pkg/operator/events/recorder_upstream.go
mfojtik: keyFunc added

pkg/controller/controllercmd/builder.go
mfojtik: done

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
```diff
@@ -82,7 +83,7 @@ func (o *CertSyncControllerOptions) Run() error {
 
 	kubeInformers := informers.NewSharedInformerFactoryWithOptions(o.kubeClient, 10*time.Minute, informers.WithNamespace(o.Namespace))
 
-	eventRecorder := events.NewKubeRecorder(o.kubeClient.CoreV1().Events(o.Namespace), "cert-syncer",
+	eventRecorder := events.NewKubeRecorder(o.kubeClient.CoreV1().Events(o.Namespace), record.CorrelatorOptions{}, "cert-syncer",
```
deads2k, line 86: we want the operator level one for this

mfojtik: fixed

deads2k:
before change, openshift-kube-apiserver-operator had 186 events, now it has 383 events.  There's a lot, but I think we can manage it.

mfojtik:
@deads2k those messages are truncated, see https://github.com/openshift/library-go/pull/784

i think increase of events is reasonable... if should consider which events are useful and which are not and how many we fire from controllers perhaps?

pkg/operator/events/recorder_upstream.go
```diff
@@ -27,20 +32,82 @@ type upstreamRecorder struct {
...
+				event.InvolvedObject.Kind,
+				event.InvolvedObject.Namespace,
+				event.InvolvedObject.Name,
+				string(event.InvolvedObject.UID),
+				event.InvolvedObject.APIVersion,
+				event.Type,
+				event.Reason,
+				event.Message,
+			},
+				""), event.Message
```
sttts, line 65: odd idention

pkg/operator/events/recorder_upstream.go
```diff
@@ -59,12 +126,35 @@ func (r *upstreamRecorder) Warningf(reason, messageFmt string, args ...interface
...
+func (r *upstreamRecorder) isShuttingDown() bool {
+	r.shutdownMutex.RLock()
+	defer r.shutdownMutex.RUnlock()
+	return r.shuttingDown
+}
+
 // Event emits the normal type event.
 func (r *upstreamRecorder) Event(reason, message string) {
+	defer r.incrementEventsCounter(corev1.EventTypeNormal)
+	if r.isShuttingDown() {
```
sttts, line 145: this races

pkg/operator/events/recorder_upstream.go
```diff
@@ -59,12 +126,35 @@ func (r *upstreamRecorder) Warningf(reason, messageFmt string, args ...interface
...
+		r.fallbackRecorder.Event(reason, message)
+		return
+	}
 	r.eventRecorder.Event(r.involvedObjectRef, corev1.EventTypeNormal, reason, message)
 }
 
 // Warning emits the warning type event.
 func (r *upstreamRecorder) Warning(reason, message string) {
+	defer r.incrementEventsCounter(corev1.EventTypeWarning)
+	if r.isShuttingDown() {
```
sttts, line 155: this races

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
```diff
@@ -82,7 +82,7 @@ func (o *CertSyncControllerOptions) Run() error {
 
 	kubeInformers := informers.NewSharedInformerFactoryWithOptions(o.kubeClient, 10*time.Minute, informers.WithNamespace(o.Namespace))
 
-	eventRecorder := events.NewKubeRecorder(o.kubeClient.CoreV1().Events(o.Namespace), "cert-syncer",
+	eventRecorder := events.NewKubeRecorder(o.kubeClient.CoreV1().Events(o.Namespace), events.RecommendedClusterSingletonCorrelatorOptions(), "cert-syncer",
```
sttts, line 85: award for the longest identifier! 🏆

mfojtik: goes to @deads2k for suggesting it :)

pkg/operator/events/recorder_upstream.go
mfojtik: races how?

pkg/operator/events/recorder_upstream.go
sttts: shutdown can be called after isShuttingDown left the lock. We would lose events.

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
sttts: it has some Java smell, am not surprised :D

pkg/operator/events/recorder_upstream.go
```diff
@@ -2,21 +2,26 @@ package events
...
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog"
 )
 
 // NewKubeRecorder returns new event recorder.
-func NewKubeRecorder(client corev1client.EventInterface, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
+func NewKubeRecorder(client corev1client.EventInterface, options record.CorrelatorOptions, sourceComponentName string, involvedObjectRef *corev1.ObjectReference) Recorder {
```
sttts, line 18: do we want to break this interface?

pkg/operator/events/recorder_upstream.go
```diff
@@ -59,12 +125,33 @@ func (r *upstreamRecorder) Warningf(reason, messageFmt string, args ...interface
...
+	if r.involvedObjectRef == nil {
+		return
+	}
+	eventsCounterMetric.WithLabelValues(severity).Inc()
+}
+
 // Event emits the normal type event.
 func (r *upstreamRecorder) Event(reason, message string) {
+	r.shutdownMutex.Lock()
+	defer r.shutdownMutex.Unlock()
```
sttts, line 138: this is a long lock. Use a RW lock and lock here read-only, and write-lock in the shutdown call func.

pkg/operator/events/recorder_upstream.go
mfojtik: I can make `NewKubeRecorderWithOptions` ?

pkg/controller/controllercmd/builder.go
```diff
@@ -175,6 +177,14 @@ func (b *ControllerBuilder) WithInstanceIdentity(identity string) *ControllerBui
...
+	b.eventRecorderOptions = options
```
sttts, line 184: I expected to set it to the recommended options by default. Is that the case?

sttts: I guess https://github.com/openshift/library-go/pull/777/files#diff-ff937793ae2933db156923ac8eaebbd8R271 is doing this, right?

PR #785, opened 2020-04-23
return an image digest from DockerImageReference string

sallyom:
I've found I need this for oc here: https://github.com/openshift/oc/pull/395
This adds function to return an image digest from a string rather than going through a registry (to get the digest for tools from release payload mirrored from a registry you may not have access to in disconnected environments) 
ie, this requires registry access: https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/info.go#L675

### DIFF
--- a/pkg/image/reference/reference.go
+++ b/pkg/image/reference/reference.go
@@ -1,6 +1,7 @@
 package reference
 
 import (
+	"fmt"
 	"net"
 	"net/url"
 	"strings"
@@ -60,6 +61,19 @@ func Parse(spec string) (DockerImageReference, error) {
 	}
 
 	return ref, nil
+}
+
+// GetSpecDigest returns the image digest from a DockerImageReference string.
+// example 'sha:22204fa2...' in an image 'registry/ocp/release@sha:22204fa2...'
+func GetSpecDigest(spec string) (string, error) {
+	namedRef, err := reference.ParseNamed(spec)
+	if err != nil {
+		return "", err
+	}
+	if named, ok := namedRef.(reference.Digested); ok {
+		return named.Digest().String(), nil
+	}
+	return "", fmt.Errorf("could not parse digest for spec: %v", spec)
 }
 
 // Equal returns true if the other DockerImageReference is equivalent to the


pkg/image/reference/reference.go
```diff
@@ -62,6 +63,19 @@ func Parse(spec string) (DockerImageReference, error) {
...
 
+// GetSpecDigestTag returns the image digest from a DockerImageReference string.
+// example '@sha:22204fa2...' in an image 'registry/ocp/release@sha:22204fa2...'
+func GetSpecDigestTag(spec string) (string, error) {
+	namedRef, err := reference.ParseNamed(spec)
+	if err != nil {
+		return "", err
+	}
+	if named, ok := namedRef.(reference.Canonical); ok {
+		return fmt.Sprintf("@" + named.Digest().String()), nil
```
dmage, line 74: Why do you need this '@' prefix?

pkg/image/reference/reference.go
```diff
@@ -62,6 +63,19 @@ func Parse(spec string) (DockerImageReference, error) {
...
 }
 
+// GetSpecDigestTag returns the image digest from a DockerImageReference string.
+// example '@sha:22204fa2...' in an image 'registry/ocp/release@sha:22204fa2...'
+func GetSpecDigestTag(spec string) (string, error) {
+	namedRef, err := reference.ParseNamed(spec)
+	if err != nil {
+		return "", err
+	}
+	if named, ok := namedRef.(reference.Canonical); ok {
```
dmage, line 73: `reference.Digested`

pkg/image/reference/reference.go
sallyom: right, i'm not sure about that.. we can add that in a fmt.Sprintf here instead: https://github.com/openshift/oc/pull/395/files#r414557373

pkg/image/reference/reference.go
sallyom: eh.. i reused that from here, btw : ) https://github.com/openshift/library-go/blob/master/pkg/image/reference/reference.go#L52  from Parse

sallyom: ok, I've updated that, thanks

pkg/image/reference/reference.go
sallyom: I updated, func is now 'GetSpecDigest' and will add the "@" when needed in oc.

pkg/image/reference/reference.go
```diff
@@ -62,6 +63,19 @@ func Parse(spec string) (DockerImageReference, error) {
 	return ref, nil
 }
 
+// GetSpecDigest returns the image digest from a DockerImageReference string.
+// example 'sha:22204fa2...' in an image 'registry/ocp/release@sha:22204fa2...'
+func GetSpecDigest(spec string) (string, error) {
```
soltysh, line 68: Why this is needed when we can use `Parse` and then read resulting `ref.ID`?

sallyom: correct, closing this lol

sallyom:
this is not needed see above comment, closing

PR #788, opened 2020-04-30
ApplyService should compare existing and required Ports

jhadvig:
Since the `Ports` field are editable we should be checking for changes.

/assign @mfojtik

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -79,12 +79,15 @@ func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder, 
 		typeSame = true
 	}
 
-	if selectorSame && typeSame && !*modified {
+	portsSame := equality.Semantic.DeepEqual(existingCopy.Spec.Ports, required.Spec.Ports)
+
+	if selectorSame && typeSame && portsSame && !*modified {
 		return existingCopy, false, nil
 	}
 
 	existingCopy.Spec.Selector = required.Spec.Selector
 	existingCopy.Spec.Type = required.Spec.Type // if this is different, the update will fail.  Status will indicate it.
+	existingCopy.Spec.Ports = required.Spec.Ports
 
 	if klog.V(4) {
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -79,12 +79,15 @@ func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder,
 		typeSame = true
 	}
 
-	if selectorSame && typeSame && !*modified {
+	portsSame := equality.Semantic.DeepEqual(existingCopy.Spec.Ports, required.Spec.Ports)
```
deads2k, line 82: This can fail when `ServiceAppProtocol` feature gate is off.  In 1.18, it is off by default.

deads2k: > This can fail when `ServiceAppProtocol` feature gate is off. In 1.18, it is off by default.

If someone is setting the `AppProtocol`, I think you can just fail to apply in general for now.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -79,12 +79,15 @@ func ApplyService(client coreclientv1.ServicesGetter, recorder events.Recorder,
...
+
+	if selectorSame && typeSame && portsSame && !*modified {
 		return existingCopy, false, nil
 	}
 
 	existingCopy.Spec.Selector = required.Spec.Selector
 	existingCopy.Spec.Type = required.Spec.Type // if this is different, the update will fail.  Status will indicate it.
+	existingCopy.Spec.Ports = required.Spec.Ports
```
deads2k, line 90: this should be a deep copy to avoid accidental mutations.

pkg/operator/resource/resourceapply/core.go
jhadvig: So I was testing to update the port protocol field in a service and that was done without any issue (on latest cluster - `4.5.0-0.ci-2020-05-04-102311`)

jhadvig: Also tested the change manually changes in our repo and it was working, without erroring our

jhadvig:
Closing the PR since the Protocol field is feature gated.

PR #789, opened 2020-04-30
Fix missing event in node controller

tkashem:
* `MasterNodesReadyChanged` event is never fired since it is at the end of the UpdateStaticPodStatusFunc chain. This issue has been fixed.
* Add defensive guard, if Ready condition is missing then `NodeControllerDegraded` should be set to True.

### DIFF
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -88,12 +88,19 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	// detect and report master nodes that are not ready
 	notReadyNodes := []string{}
 	for _, node := range nodes {
-		for _, con := range node.Status.Conditions {
-			if con.Type == coreapiv1.NodeReady && con.Status != coreapiv1.ConditionTrue {
-				notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready since %s because %s (%s)", node.Name, con.LastTransitionTime, con.Reason, con.Message))
-			}
+		nodeReadyCondition := nodeConditionFinder(&node.Status, coreapiv1.NodeReady)
+
+		// If a "Ready" condition is not found, that node should be deemed as not Ready by default.
+		if nodeReadyCondition == nil {
+			notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready, no Ready condition found in status block", node.Name))
+			continue
+		}
+
+		if nodeReadyCondition.Status != coreapiv1.ConditionTrue {
+			notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready since %s because %s (%s)", node.Name, nodeReadyCondition.LastTransitionTime, nodeReadyCondition.Reason, nodeReadyCondition.Message))
 		}
 	}
+
 	newCondition := operatorv1.OperatorCondition{
 		Type: condition.NodeControllerDegradedConditionType,
 	}
@@ -108,12 +115,12 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	}
 
 	oldStatus := &operatorv1.StaticPodOperatorStatus{}
-	_, updated, updateError := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition), func(status *operatorv1.StaticPodOperatorStatus) error {
-		status.NodeStatuses = newTargetNodeStates
-		return nil
-	}, func(status *operatorv1.StaticPodOperatorStatus) error {
-		//a hack for storing the old status (before the update)
+	_, updated, updateError := v1helpers.UpdateStaticPodStatus(c.operatorClient, func(status *operatorv1.StaticPodOperatorStatus) error {
+		//a hack for storing the old status (before we mutate it)
 		oldStatus = status
+		return nil
+	}, v1helpers.UpdateStaticPodConditionFn(newCondition), func(status *operatorv1.StaticPodOperatorStatus) error {
+		status.NodeStatuses = newTargetNodeStates
 		return nil
 	})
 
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -86,14 +86,26 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+			}
+		}
+
+		return nil
+	}
+
 	notReadyNodes := []string{}
 	for _, node := range nodes {
-		for _, con := range node.Status.Conditions {
-			if con.Type == coreapiv1.NodeReady && con.Status != coreapiv1.ConditionTrue {
```
tkashem, line 92: it's a bit more defensive, if the "Ready" condition is not found then we should deem the node as `not ready`, correct?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -108,13 +120,13 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	}
 
 	oldStatus := &operatorv1.StaticPodOperatorStatus{}
-	_, updated, updateError := v1helpers.UpdateStaticPodStatus(c.operatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition), func(status *operatorv1.StaticPodOperatorStatus) error {
-		status.NodeStatuses = newTargetNodeStates
-		return nil
-	}, func(status *operatorv1.StaticPodOperatorStatus) error {
```
tkashem, line 114: this was at the end of the chain, so it received the status object that has been mutated already by the previous `UpdateStaticPodStatusFunc` funcs. `oldStatus` is basically the new status we are trying to save and it will have `newCondition`, right? 

So we may not see the event "MasterNodesReadyChanged" I think.

pkg/operator/v1helpers/helpers.go
```diff
@@ -57,11 +57,13 @@ func SetOperatorCondition(conditions *[]operatorv1.OperatorCondition, newConditi
 		return
 	}
 
-	if existingCondition.Status != newCondition.Status {
-		existingCondition.Status = newCondition.Status
+	if existingCondition.Status != newCondition.Status ||
```
tkashem, line 60: change in `status` or `reason` should be considered a transition, correct?

pkg/operator/staticpod/controller/node/node_controller.go
sanchezl: I concur.

pkg/operator/v1helpers/helpers.go
sanchezl: No, it is the 
```
// Last time the condition transitioned from one status to another.
```
I have seen some conditions that have a `lastUpdateTime` also (like `DeploymentCondition`).

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: yes, it seems it should have been at the beginning of the chain.

p0lyn0mial: or could we simply relay on `originalOperatorStatus` to detect changes?

pkg/operator/v1helpers/helpers.go
tkashem: thanks @sanchezl, so it applies to the condition as a whole, the last time the condition changed state, irrespective of the reason. I think it makes sense.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -86,14 +86,26 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
 
 	// detect and report master nodes that are not ready
+	nodeConditionFinder := func(node *coreapiv1.Node) *coreapiv1.NodeCondition {
+		for i := range node.Status.Conditions {
+			if node.Status.Conditions[i].Type == coreapiv1.NodeReady {
+				return &node.Status.Conditions[i]
+			}
+		}
+
+		return nil
```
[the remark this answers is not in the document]
tkashem: If a `NodeReady` condition is missing then I would want to log this as a separate message. I added a new error message when `NodeReady` is missiong from the status block.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -86,14 +86,26 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	}
 
 	// detect and report master nodes that are not ready
+	nodeConditionFinder := func(node *coreapiv1.Node) *coreapiv1.NodeCondition {
```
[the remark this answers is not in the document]
tkashem: I moved it to a new function. I wish the code-gen generated these utility functions along with the type.

pkg/operator/staticpod/controller/node/node_controller.go
tkashem: @p0lyn0mial are you suggesting we change the signature of the `UpdateStatus` function? I am in favor of changing the signature of the function (we have to add a v2 of this package I assume to maintain backward compatibility?)

pkg/operator/staticpod/controller/node/node_controller.go
tkashem: in implementation logic we have to take into account the possibility of the condition not being in the slice. In my opinion, we should be defensive and guard for nil or not found although by contract the `Nodeready` condition should be there.

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: I was thinking about reading the old status from [originalOperatorStatus](https://github.com/openshift/library-go/blob/fe935f679430bcf7dfee61ddf7bec7c37db79db4/pkg/operator/staticpod/controller/node/node_controller.go#L42) but it seems to be not enough since we need to update both `newCondition` and `NodeStatus`

No need to create a new function (`v2`) as we don't have any guarantees of backward compatibility for `library-go`

PR #794, opened 2020-05-12
Bug 1801437: Add unknownRevisionStatus limit to static pod operator prune controllers

damemi:
This depends on https://github.com/openshift/api/pull/643

Considers the new `unknownRevisionLimit` field in the static pod operator spec provided by the above PR when calculating revision resources to exclude from pruning. Previously, this just excluded all unknown revisions meaning they were never pruned.

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/prune"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -118,28 +122,28 @@ func (c *PruneController) excludedRevisionHistory(ctx context.Context, recorder
...
-				// protect things you don't understand
 				unknownStatusRevisions = append(unknownStatusRevisions, revisionNumber)
 				recorder.Event("UnknownRevisionStatus", fmt.Sprintf("unknown status for revision %d: %v", revisionNumber, configMap.Data["status"]))
 			}
 		}
 	}
 
 	// Return early if nothing to prune
-	if len(succeededRevisions)+len(failedRevisions) == 0 {
+	if len(succeededRevisions)+len(failedRevisions)+len(unknownStatusRevisions) == 0 {
```
deads2k, line 132: If I understand the code correctly, the revision controller will never set  the configmap status out of InProgress today even if it is no longer being worked.

You probably want to set it to something like "NeverCompleted", right?

You should probably update that controller to set proper status on old revisions I'd think.

Sample data here:

```
- apiVersion: v1
  data:
    revision: "1"
    status: InProgress
  kind: ConfigMap
  metadata:
    creationTimestamp: "2019-12-05T18:24:01Z"
    name: revision-status-1
    namespace: openshift-kube-apiserver
    resourceVersion: "2760"
    selfLink: /api/v1/namespaces/openshift-kube-apiserver/configmaps/revision-status-1
    uid: 68df822f-178c-11ea-a4b6-02ba0195095b
- apiVersion: v1
  data:
    revision: "115"
    status: InProgress
  kind: ConfigMap
  metadata:
    creationTimestamp: "2020-04-22T18:41:10Z"
    name: revision-status-115
    namespace: openshift-kube-apiserver
    resourceVersion: "109420856"
    selfLink: /api/v1/namespaces/openshift-kube-apiserver/configmaps/revision-status-115
    uid: d3d6e579-130c-4e34-ac37-deb922fb3340
- apiVersion: v1
  data:
    revision: "121"
    status: InProgress
  kind: ConfigMap
  metadata:
    creationTimestamp: "2020-04-23T00:52:07Z"
    name: revision-status-121
    namespace: openshift-kube-apiserver
    resourceVersion: "109631969"
    selfLink: /api/v1/namespaces/openshift-kube-apiserver/configmaps/revision-status-121
    uid: 90bc20c5-1a8f-4a79-a9aa-605c440ab387
```

Side note, storing a revisionTriggered reason would be pretty nice.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -191,20 +192,40 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 }
 
 func (c RevisionController) createNewRevision(recorder events.Recorder, revision int32) error {
+	// Check if the previous status never progressed past InProgress that means this new revision was kicked off before
+	// the previous finished -- mark it Abandoned
+	if revision > 1 {
```
deads2k, line 197: check all previous versions

damemi: updated

damemi:
This will need a rebase once https://github.com/openshift/api/pull/643 merges

damemi:
/close
this isn't adding a limit for unknownRevisions anymore and is more focused on fixing the bug at hand around Abandoned revisions. So I'm moving the changes to a new PR to preserve these unknownRevisions commits (in case we end up wanting them) and focus on that: https://github.com/openshift/library-go/pull/806

PR #795, opened 2020-05-13
Process Secret.StringData when reading manifests

tnozicka:
When manifests contain secret with `stringData` we need to merge it into `Data` so any logic can process it in unified way. `ApplySecret` explicitly forbids `stringData` (for obvious apply/DeepEqual issues) 
https://github.com/openshift/library-go/blob/609e1511d2de69e6d84afb1f3e8a5a6b3c805fc2/pkg/operator/resource/resourceapply/core.go#L238-L240

/cc @deads2k

### DIFF
--- a/pkg/operator/resource/resourceread/core.go
+++ b/pkg/operator/resource/resourceread/core.go
@@ -30,7 +30,20 @@ func ReadSecretV1OrDie(objBytes []byte) *corev1.Secret {
 	if err != nil {
 		panic(err)
 	}
-	return requiredObj.(*corev1.Secret)
+
+	secret := requiredObj.(*corev1.Secret)
+	if len(secret.StringData) > 0 {
+		if secret.Data == nil {
+			secret.Data = map[string][]byte{}
+		}
+		for k, v := range secret.StringData {
+			secret.Data[k] = []byte(v)
+		}
+
+		secret.StringData = nil
+	}
+
+	return secret
 }
 
 func ReadNamespaceV1OrDie(objBytes []byte) *corev1.Namespace {


pkg/operator/resource/resourceread/core.go
```diff
@@ -30,7 +30,20 @@ func ReadSecretV1OrDie(objBytes []byte) *corev1.Secret {
 	if err != nil {
 		panic(err)
 	}
-	return requiredObj.(*corev1.Secret)
+
+	secret := requiredObj.(*corev1.Secret)
+	if len(secret.StringData) > 0 {
```
deads2k, line 35: This fixes one possible inbound source.  Instead of this, we should update the appySecret function to put the StringData into the Data field and then update based on that.

tnozicka: The reason I did it here and not there is that you need to process/modify the `.Data` before sending it to `ApplySecret`. E.g. I need to inject certificates. If we don't do it at read time, any processing loop will need to be duplicated to handle both `.Data` and `.StringData`. Generally the API assumptions is you get only `.Data` when reading. Unifying that assumption for reading manifests seems like the closest we can get to have sane (non-duplicated) processing. As the base64 difference is meant just for writing binary data in text files I don't see a use case for setting `StringData` on any code path, hence `ApplySecret` should't need to handle it and it explicitly forbids it in case someone made a mistake.

tnozicka: (the base64 conversion happens when decoding the object)

deads2k: > `ApplySecret` should't need to handle it and it explicitly forbids it in case someone made a mistake.

We should match.  Either this isn't allowed and the code is correct and people will be smacked.  Or we change the code to allow on both sides. But a halfway state isn't good for anyone.

tnozicka: Half way prevents bugs like if somebody reads secrets with `StringData` without merging it into `.Data`, processes it with some function (those generally handle only `.Data`), any data stored in `.StringData` are missed in the processing but ApplySecret stops you on StringData and you discover the bug.

deads2k:
I think https://github.com/openshift/library-go/pull/797 just fixes this.  If it's easy to fix, we should just make these things work

tnozicka:
> I think #797 just fixes this. If it's easy to fix, we should just make these things work

I can live with ApplySecret handling it as well, but on its own it's not fixing the issue presented in this PR. We need the read fix as well, that's what makes processing manifests possible, apply is only the final step at which you already have only `Data`.

For real use case, see https://github.com/openshift/cluster-kube-apiserver-operator/pull/858/files#diff-6501874255487b3161b3dfc0f3f34070R487

PR #797, opened 2020-05-15
update ApplySecret to handle StringData

deads2k:
just make it work everywhere instead of just a single flow like https://github.com/openshift/library-go/pull/795

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -234,10 +234,18 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
 }
 
 // ApplySecret merges objectmeta, requires data
-func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
-	if len(required.StringData) > 0 {
-		return nil, false, fmt.Errorf("Secret.stringData is not supported")
+func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret) (*corev1.Secret, bool, error) {
+	// copy the stringData to data.  Error on a data content conflict inside required.  This is usually a bug.
+	required := requiredInput.DeepCopy()
+	for k, v := range required.StringData {
+		if dataV, ok := required.Data[k]; ok {
+			if string(dataV) != v {
+				return nil, false, fmt.Errorf("Secret.stringData[%q] conflicts with Secret.data[%q]", k, k)
+			}
+		}
+		required.Data[k] = []byte(v)
 	}
+	required.StringData = nil
 
 	existing, err := client.Secrets(required.Namespace).Get(context.TODO(), required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {


pkg/operator/resource/resourceapply/core.go
```diff
@@ -234,10 +234,18 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
...
-func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
-	if len(required.StringData) > 0 {
-		return nil, false, fmt.Errorf("Secret.stringData is not supported")
+func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret) (*corev1.Secret, bool, error) {
+	// copy the stringData to data.  Error on a data content conflict inside required.  This is usually a bug.
+	required := requiredInput.DeepCopy()
+	for k, v := range required.StringData {
+		if dataV, ok := required.Data[k]; ok {
+			if string(dataV) != v {
+				return nil, false, fmt.Errorf("Secret.stringData[%q] conflicts with Secret.data[%q]", k, k)
```
tnozicka, line 243: the api docs say this always overrides data. if you were to change a field in the code using stringData in the code (I don't see why), wouldn't this break the update?

deads2k: > the api docs say this always overrides data. if you were to change a field in the code using stringData in the code (I don't see why), wouldn't this break the update?

I think that would mean inconsistent information was provided as requiredInput, so a failure here is reasonable.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -234,10 +234,18 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
...
+			}
+		}
+		required.Data[k] = []byte(v)
```
vareti, line 246: why is this not base64 encoded before writing into `Data`?

pkg/operator/resource/resourceapply/core.go
```diff
@@ -234,10 +234,18 @@ func ApplyConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Record
 }
 
 // ApplySecret merges objectmeta, requires data
-func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
-	if len(required.StringData) > 0 {
-		return nil, false, fmt.Errorf("Secret.stringData is not supported")
+func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret) (*corev1.Secret, bool, error) {
+	// copy the stringData to data.  Error on a data content conflict inside required.  This is usually a bug.
+	required := requiredInput.DeepCopy()
```
vareti, line 239: just curios why requiredInput was not checked for `nil` value here? It seems to be the case previously too.

pkg/operator/resource/resourceapply/core.go
deads2k: > why is this not base64 encoded before writing into `Data`?

encoding is done automatically during marshalling

pkg/operator/resource/resourceapply/core.go
deads2k: > just curios why `requiredInput` was not checked for `nil` value here? It is the same case before you made any changes too.

uh, hopefully it panics at some point.  don't do that.

pkg/operator/resource/resourceapply/core.go
vareti: Good to know. Thanks!

PR #802, opened 2020-05-20
pkg/operator/status/condition: Default to Upgradeable=True AsExpected

wking:
Having `Unknown` and `NoData` for other conditions is reasonable, because all operators should have strong opinions on `Available`, `Degraded`, and `Progressing`.  But it's possible to have operators which will never set `Upgradeable=False`, and it seems tedious to force them to declare an opinion to avoid a potentially alarming message ("why don't they have data?  Is something broken?").  This commit will get us `Upgradeable=True` `AsExpected` for operators like:

```console
$ JQ='.items[] | .unknown = ([.status.conditions[] | select(.status != "True" and .status != "False") | {type, status, reason}]) | select((.unknown | length) > 0) | .metadata.name + " " + (.unknown | tostring)'
$ curl -s https://storage.googleapis.com/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/11222/artifacts/launch/clusteroperators.json | jq -r "${JQ}"
kube-storage-version-migrator [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
openshift-controller-manager [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
service-ca [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
$ curl -s https://storage.googleapis.com/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-4.5/2086/artifacts/e2e-aws/clusteroperators.json | jq -r "${JQ}"
kube-storage-version-migrator [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
openshift-controller-manager [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
service-ca [{"type":"Upgradeable","status":"Unknown","reason":"NoData"}]
```

where the examples are from [a 4.4.3 -> 4.4.4 promotion test][1] and [a 4.5 nightly test][2] respectively.

[1]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/11222
[2]: https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-4.5/2086

### DIFF
--- a/pkg/operator/status/condition.go
+++ b/pkg/operator/status/condition.go
@@ -44,8 +44,13 @@ func UnionCondition(conditionType string, defaultConditionStatus operatorv1.Cond
 
 	unionedCondition := operatorv1.OperatorCondition{Type: conditionType, Status: operatorv1.ConditionUnknown}
 	if len(interestingConditions) == 0 {
-		unionedCondition.Status = operatorv1.ConditionUnknown
-		unionedCondition.Reason = "NoData"
+		if conditionType == "Upgradeable" {
+			unionedCondition.Status = defaultConditionStatus
+			unionedCondition.Reason = "AsExpected"
+		} else {
+			unionedCondition.Status = operatorv1.ConditionUnknown
+			unionedCondition.Reason = "NoData"
+		}
 		return unionedCondition
 	}
 


wking:
/assign @deads2k

Unless you want to delegate ;)

wking:
Per [renewed interest][1]:

/reopen
/remove-lifecycle rotten

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1928141#c4

wking:
Rebased around a5a9a5e626d (#842) with 7dcbab36 -> 57898ddc.

pkg/operator/status/condition.go
```diff
@@ -44,8 +44,13 @@ func UnionCondition(conditionType string, defaultConditionStatus operatorv1.Cond
 
 	unionedCondition := operatorv1.OperatorCondition{Type: conditionType, Status: operatorv1.ConditionUnknown}
 	if len(interestingConditions) == 0 {
-		unionedCondition.Status = operatorv1.ConditionUnknown
-		unionedCondition.Reason = "NoData"
+		if conditionType == "Upgradeable" {
```
tkashem, line 47: I am thinking of an operator that has a strong opinion on `Upgradeable`. Should we set the default value to `AsExpected` for those operators? (The operator finds an Upgradeable=False condition but fails to update the status for some reason)

Maybe I am missing something here, but to me it feels like that the default `Unknown` is safe to start with here.

wking: > The operator finds an `Upgradeable=False` condition but fails to update the status for some reason

I think the flow you're concerned about is:

1. API server is unavailable, operator fails to write some `UpgradeableWhatever=False` operator condition.
2. API server becomes available, operator forgets that it just wanted to write `UpgradeableWhatever=False`, doesn't see anything concerning on the in-cluster status, and writes `Upgradeable=True` `AsExpected`.
3. API server remains available, operator successfully writes `UpgradeableWhatever=False`.
4. API server remains available, operator successfully notices `UpgradeableWhatever=False` and writes `Upgradeable=False`.

I'm a bit fiddly on how client-side caching works.  Ideally "forgets that it just wanted to write" would not be possible.  If that's possible today, library-go should be able to add additional client-side guards to work around it.  But even if it is possible today, it seems unlikely to me that the gap between 1 and 4 would be all that large or likely to receive and update request.  And if that gap exists, you're still exposed to the gap between 1 and 3 in:

1. API server is available, operator writes `UpgradeableWhatever=False`.
2. API server becomes unavailable, operator cannot write `Upgradeable=False`.
3. API server becomes available, operator successfully notices `UpgradeableWhatever=False` and writes `Upgradeable=False`.

So I'm not sure how much effort we want to address in driving down this latency.  It's hard to imagine cases that would be really terrible for an update that did not exist minutes before the update was requested.  And if those sorts of cases do exist, they could also occur minutes after the update was requested when we were already committed.  Thoughts?

tkashem: My apology, I misunderstood the changes. So if I am not mistaken, `UnionCondition` aggregates the `*Upgradeable` conditions reported by an operator and it synthesizes a new `Upgradeable` condition to be written into `ClusterOperator` status. 
And by default, if `len(interestingConditions)` is zero then we set `Upgradeable` to `Unknown`. 

So if we have an operator that does not have *any* opinion on `Upgradeable`:
`interestingConditions` will be empty -> `Upgradeable` is set to `Unknown` -> blocks upgrade?

If I have understood this correctly, let me know if the following scenario makes sense:
- We have an operator that has a strong opinion on `Upgradeable` and it maintains only one condition type for that -  `Foo_Upgradeable` condition. Due to a bug in the code it stops writing this condition. When we aggregate inside `UnionCondition` function, `interestingConditions` will be empty and we will write `AsExpected`.  (our defaulting was optimistic and led to this issue)  

A couple of points:
- In general, I think we should always set the default value to be pessimistic or `Unknown` in this case so we don't let a bug go unnoticed. we don't automatically conclude that it's working as expected, when it may not be the case.
- Not have special handling of a certain condition type inside `UnionCondition`. This change is not desirable for an operator that has a strong opinion of `Upgradeable`, right? Maybe we can parameterize this change so each author can choose the default value appropriately or add a new function that writes `AsExpected`. So it becomes more of an opt in process vs automatic.

Thoughts?

wking: > So if we have an operator that does not have any opinion on Upgradeable: interestingConditions will be empty -> Upgradeable is set to Unknown -> blocks upgrade?

That's all right except "blocks updates".  The CVO only blocks minor-version updates on `Upgradeable=False`.  The issue with `Upgradeable=Unknown` is just that [folks find it confusing][1].

> We have an operator that has a strong opinion on `Upgradeable` and it maintains only one condition type for that - `Foo_Upgradeable` condition. Due to a bug in the code it stops writing this condition.

There's no freshness cutoff in any of these conditions.  If you stop writing the condition, this union code will continue to pass along whatever value you'd set before when it updates `Upgradeable`.

> When we aggregate inside `UnionCondition` function, `interestingConditions` will be empty and we will write `AsExpected`.

It will only be empty if the operator has not written any `*Upgradeable` operator conditions for the union code to consume.  That's possible, but I don't think guarding against it is worth setting `Upgradeable=False`, and without doing that, you get no protection from the CVO.

> This change is not desirable for an operator that has a strong opinion of `Upgradeable`, right?

This change is irrelevant for operators with opinions, because they will always have content in `interestingConditions`.  The way you express "I have an opinion" is just by setting `*Upgradeable` operator conditions.  I don't think we need a parallel knob to emphasize that.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1928141#c0

wking:
I still think this is a sane path forward, but haven't hit upon the magic phrasing that makes my arguments convincing ;).  Closing, and library-go maintainers and consumers can sort out the `Unknown` conditions however they like.  Ping me if y'all come around and want this re-opened again later :p.

/close

PR #806, opened 2020-05-27
Update interrupted InProgress revisions to Abandoned

damemi:
This carries some of the changes in https://github.com/openshift/library-go/pull/794, specifically when creating a new revision it checks for interrupted `InProgress` revisions and marks them as `Abandoned`

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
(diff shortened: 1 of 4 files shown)


pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -191,20 +192,41 @@ func (c RevisionController) isLatestRevisionCurrent(revision int32) (bool, strin
 }
 
 func (c RevisionController) createNewRevision(recorder events.Recorder, revision int32) error {
+	// Before we create a new revision, check if any of the existing revisions got interrupted while InProgress
+	// If so, mark them Abandoned.
+	configMaps, err := c.configMapGetter.ConfigMaps(c.targetNamespace).List(context.TODO(), metav1.ListOptions{})
```
deads2k, line 197: do this *after* the new one is created and exclude the new one.  That way if you crash, you don't accidentally mark the last on as Abandoned before creating the new current one.

PR #816, opened 2020-06-15
Add Config Observers for OAuth Token Config

vareti:
This PR adds config observers for OAuth [TokenConfig's](https://github.com/openshift/api/blob/de0cf99e4f4f1c38153a61d48757bd86c9309c09/osin/v1/types.go#L393) `AccessTokenMaxAgeSeconds` and `AccessTokenInactivityTimeout`. It adds separate functions for observing these two fields. 

These config observers would be reused by both cluster-authentication-operator and cluster-kube-apiserver-operator to write appropriate config for oauth-server and kube-apiserver respectively.


pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	return mergedConfig, errs
+}
+
+// ObserveAccessTokenMaxAgeSeconds returns the observed config for access token max age.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config,
+// it uses the default value for token max age.
+// If the function only fails to read OAuth cluster config, it returns existing config.
+// If token max age value is zero in OAuth cluster config, it uses default value.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
marun, line 58: (No action required) I think it might be simpler to separate mixing value retrieval and return value construction. Maybe separate out retrieval of the value from oauth config into a separate function?

pkg/operator/configobserver/oauth/observe_tokenconfig_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (l testLister) ResourceSyncer() resourcesynccontroller.ResourceSyncer {
+	return nil
+}
+
+func (l testLister) PreRunHasSynced() []cache.InformerSynced {
+	return nil
+}
+
+func TestObserveTokenConfig(t *testing.T) {
```
marun, line 34: (No action required) I don't see all paths being tested. Maybe test corner cases of the helper functions (`ObserveAccessTokenInactivityTimeoutSeconds` and `ObserveAccessTokenMaxAgeSeconds`) independently to make this easier?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	if existingAccessTokenInactivityTimeoutSeconds != nil {
+		if f, ok := existingAccessTokenInactivityTimeoutSeconds.(float64); ok {
+			existingAccessTokenInactivityTimeoutSecondsFloat = &f
+		}
+	}
+
+	oauthConfig, err := listers.OAuthLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warning("oauth.config.openshift.io/cluster: not found")
+		return buildUnstructuredTokenConfig("", nil), nil
```
marun, line 135: Why not return the error slice?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	errs = []error{}
+	mergedConfig := map[string]interface{}{}
+
+	for _, observeConfig := range []configobserver.ObserveConfigFunc{ObserveAccessTokenMaxAgeSeconds, ObserveAccessTokenInactivityTimeoutSeconds} {
+		observedConfig, err := observeConfig(genericlisters, recorder, existingConfig)
+		if err != nil {
+			errs = append(errs, err...)
+		}
+
+		if err := mergo.Merge(&mergedConfig, observedConfig); err != nil {
```
sttts, line 44: why not wire two observers? the merging is done by the caller.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+// ObserveAccessTokenMaxAgeSeconds returns the observed config for access token max age.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config,
+// it uses the default value for token max age.
+// If the function only fails to read OAuth cluster config, it returns existing config.
+// If token max age value is zero in OAuth cluster config, it uses default value.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	listers := genericlisters.(OAuthLister)
+	accessTokenMaxAgePath := append(tokenConfigPath, fieldAccessTokenMaxAgeSeconds)
```
sttts, line 61: this func does not own tokenConfigPath. You **must not** append. You risk read/write races.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	defer func() {
+		// Prune the observed config so that it only contains access token max age field.
+		ret = configobserver.Pruned(ret, accessTokenMaxAgePath)
```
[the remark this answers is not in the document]
sttts: this is the pattern we have in all observers. So defer is fine. Please don't diverge.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+// If token inactivity timeout is not set in OAuth cluster config, it return empty config.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenInactivityTimeoutSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	listers := genericlisters.(OAuthLister)
+	observedConfig := map[string]interface{}{}
+	accessTokenInactivityTimeoutPath := append(tokenConfigPath, fieldAccessTokenInactivityTimeoutSeconds)
+	defer func() {
+		// Prune the observed config so that it only contains access token inactivity timeout field.
+		ret = configobserver.Pruned(ret, accessTokenInactivityTimeoutPath)
```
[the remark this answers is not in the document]
sttts: yes. We have that everywhere because `existingConfig` is a full config objects, but we must not return it here as it overrides values this observer does not own.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return buildUnstructuredTokenConfig(fieldAccessTokenMaxAgeSeconds, observedAccessTokenMaxAgeSeconds), errs
+}
+
+// ObserveAccessTokenInactivityTimeoutSeconds returns the observed config for access token inactivity timeout.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config, it return empty config.
+// If the function only fails to read OAuth cluster config, it returns existing config.
+// If token inactivity timeout is not set in OAuth cluster config, it return empty config.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenInactivityTimeoutSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
sttts, line 108: why do we have two observers in the first place? Aren't both values from the same config?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+		if err := mergo.Merge(&mergedConfig, observedConfig); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return mergedConfig, errs
+}
+
+// ObserveAccessTokenMaxAgeSeconds returns the observed config for access token max age.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config,
```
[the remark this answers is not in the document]
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+// it uses the default value for token max age.
+// If the function only fails to read OAuth cluster config, it returns existing config.
+// If token max age value is zero in OAuth cluster config, it uses default value.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	listers := genericlisters.(OAuthLister)
```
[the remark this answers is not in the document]
vareti: My understanding is that this function can be called from outside this package and argument type must be generic. So, I am trying to follow the expectations set [here](https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/config_observer_controller.go#L29). What am i missing?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	accessTokenMaxAgePath := append(tokenConfigPath, fieldAccessTokenMaxAgeSeconds)
+	defer func() {
+		// Prune the observed config so that it only contains access token max age field.
+		ret = configobserver.Pruned(ret, accessTokenMaxAgePath)
+	}()
+
+	existingAccessTokenMaxAgeSeconds, _, err := unstructured.NestedFloat64(existingConfig, accessTokenMaxAgePath...)
+	if err != nil {
+		// If failed to parse the config for access token max age field, continue execution with the
```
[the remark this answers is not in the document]
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+		// hope to find the value for it in OAuth config.
+		errs = append(errs, err)
+	}
+
+	oauthConfig, err := listers.OAuthLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		klog.Warning("oauth.config.openshift.io/cluster: not found")
+		return buildUnstructuredTokenConfig(fieldAccessTokenMaxAgeSeconds, defaultAccessTokenMaxAgeSeconds), nil
```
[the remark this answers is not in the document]
vareti: Agreed, I am also not convinced about this. Updated the code now to not overwrite existing config.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	} else if err != nil {
+		// Failed to read OAuth cluster config.
+		if existingAccessTokenMaxAgeSeconds == 0 {
+			// If the attempt to read from existing config failed previously, fallback to default value.
```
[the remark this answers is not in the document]
vareti: Point taken. Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+			return buildUnstructuredTokenConfig(fieldAccessTokenMaxAgeSeconds, defaultAccessTokenMaxAgeSeconds), append(errs, err)
+		}
+		// Return whatever is present in existing config.
+		return existingConfig, append(errs, err)
+	}
+
+	observedAccessTokenMaxAgeSeconds := float64(oauthConfig.Spec.TokenConfig.AccessTokenMaxAgeSeconds)
+	if observedAccessTokenMaxAgeSeconds == 0 {
```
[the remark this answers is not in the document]
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+		recorder.Eventf("ObserveTokenConfig", "%s changed from %d to %d", fieldAccessTokenMaxAgeSeconds,
+			existingAccessTokenMaxAgeSeconds,
+			observedAccessTokenMaxAgeSeconds)
+	}
+
+	return buildUnstructuredTokenConfig(fieldAccessTokenMaxAgeSeconds, observedAccessTokenMaxAgeSeconds), errs
+}
+
+// ObserveAccessTokenInactivityTimeoutSeconds returns the observed config for access token inactivity timeout.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config, it return empty config.
```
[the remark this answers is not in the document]
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+// If the function only fails to read OAuth cluster config, it returns existing config.
+// If token inactivity timeout is not set in OAuth cluster config, it return empty config.
+// In any other case, it uses the value present in OAuth cluster config.
+func ObserveAccessTokenInactivityTimeoutSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	listers := genericlisters.(OAuthLister)
```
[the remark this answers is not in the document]
vareti: My understanding is that this function can be called from outside this package and argument has to be of generic type. So, we must follow the expectations set [here](https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/config_observer_controller.go#L29). What am i missing?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+	observedConfig := map[string]interface{}{}
+	accessTokenInactivityTimeoutPath := append(tokenConfigPath, fieldAccessTokenInactivityTimeoutSeconds)
+	defer func() {
+		// Prune the observed config so that it only contains access token inactivity timeout field.
+		ret = configobserver.Pruned(ret, accessTokenInactivityTimeoutPath)
+	}()
+
+	existingAccessTokenInactivityTimeoutSeconds, _, err := unstructured.NestedFieldCopy(existingConfig, accessTokenInactivityTimeoutPath...)
+	if err != nil {
+		// If failed to parse existing config for access token inactivity timeout, continue execution with the
```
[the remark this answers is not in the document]
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+
+func buildUnstructuredTokenConfig(key string, val interface{}) map[string]interface{} {
+	defaultConfig := map[string]interface{}{}
+
+	observedMaxAgeFieldMap := map[string]interface{}{
+		key: val,
+	}
+
+	if err := unstructured.SetNestedMap(defaultConfig, observedMaxAgeFieldMap, tokenConfigPath...); err != nil {
+		// This must never happen because defaultConfig is empty.
```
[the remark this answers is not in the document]
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: I will look into this.

pkg/operator/configobserver/oauth/observe_tokenconfig_test.go
vareti: I have the same thought. I will try to add separate tests for paths that are not tested.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: This thought crossed my mind. It is a trade-off between wiring two observers versus the caller knowing about only one observer. I chose the later because I felt it is simple. But, I don't have a strong opinion about this and can change it as you would like.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Agreed. No easy way to declare global constant array in Go. Updated the code now to make this variable local to the function.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Yes, they are from same config. But both the values are not needed simultaneously in all use cases. For example, kube-apiserver would only need to know about `accessTokenInactivityTimeoutSeconds`

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+
+	return mergedConfig, errs
+}
+
+// ObserveAccessTokenMaxAgeSeconds returns the observed config for access token max age.
+// If the function fails to read both AccessTokenMaxAgeSeconds from existing config and OAuth cluster config,
+// it uses the default value for token max age.
+// If the function only fails to read OAuth cluster config, it returns existing config.
```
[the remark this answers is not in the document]
vareti: I updated the comments. Please take a look.

vareti:
@stlaz I stole the code from your [commit](https://github.com/openshift/cluster-authentication-operator/pull/222/commits/0ad89802cb942b058f7ec5e55f6057c53f6dc739) and created this PR. I would really appreciate if you could review it.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: Would prefer simpler code that sets two values instead of all the boilerplate twice. No strong opinion though. Just do what makes more sense i.e. is more elegant.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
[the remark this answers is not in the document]
sttts: it's the pattern we use in the observers. It's not pretty, but Go does not give us tools to make it better.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,169 @@
...
+			observedAccessTokenInactivityTimeout)
+	}
+
+	return observedConfig, errs
+}
+
+func buildUnstructuredTokenConfig(key string, val interface{}, fields []string) map[string]interface{} {
+	defaultConfig := map[string]interface{}{}
+
+	observedMaxAgeFieldMap := map[string]interface{}{
```
sttts, line 158: nit: this is not about maxagefield anymore, is it? It's generic.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,169 @@
...
+		key: val,
+	}
+
+	if err := unstructured.SetNestedMap(defaultConfig, observedMaxAgeFieldMap, fields...); err != nil {
+		// SetNestedMap can return only return an error if the fields are not present in map.
+		// As defaultConfig is empty, this error must never happen.
+		klog.Warningf("Failed to write default config for access token max age: %v", err)
```
sttts, line 165: same here, warning must be generic

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig_test.go
vareti: I refactored the code and added few tests. Now the coverage is close to 100%.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done. Removed the boilerplate code.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		// Failed to read OAuth cluster config.
+		if errors.IsNotFound(err) {
+			klog.Warning("oauth.config.openshift.io/cluster: not found")
+		}
+
+		if existingAccessTokenMaxAgeSeconds == 0 {
+			// As the value 0 here indicates that a read from existing config failed previously, use the default value.
+			return buildUnstructuredTokenConfig(defaultAccessTokenMaxAgeSeconds, tokenMaxAgePath), append(errs, err)
+		}
```
sttts, line 59: just return existingConfig here and the err appended to errs. This extra logic is wrong IMO. The nested float might be missing, and the lister returns an error for whatever reason. This code here tries to be clever, but makes the situation worse.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	if !ok {
+		return existingConfig, append(errs, fmt.Errorf("failed to assert: given lister does not implement an OAuth lister"))
+	}
+
+	existingAccessTokenMaxAgeSeconds, _, err := unstructured.NestedFloat64(existingConfig, tokenMaxAgePath...)
+	if err != nil {
+		// If failed to read the config for access token max age field, continue execution with the
+		// hope to find the value for it in OAuth cluster config.
+		errs = append(errs, err)
+	}
```
sttts, line 47: with https://github.com/openshift/library-go/pull/816/files#r444861773, move this block down where it is (will be) used

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+
+	observedAccessTokenMaxAgeSeconds := float64(oauthConfig.Spec.TokenConfig.AccessTokenMaxAgeSeconds)
+	if observedAccessTokenMaxAgeSeconds == 0 {
+		// As the value 0 indicates that this field is not set in OAuth cluster config, use the default value.
+		observedAccessTokenMaxAgeSeconds = defaultAccessTokenMaxAgeSeconds
+	}
+
+	if existingAccessTokenMaxAgeSeconds != observedAccessTokenMaxAgeSeconds {
+		recorder.Eventf("ObserveTokenConfig", "%s changed from %d to %d", fieldAccessTokenMaxAgeSeconds,
```
sttts, line 71: ObserveOauthTokenConfig

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	existingAccessTokenInactivityTimeout, _, err := unstructured.NestedFieldCopy(existingConfig, tokenInactivityTimeoutPath...)
+	if err != nil {
+		// If failed to read existing config for access token inactivity timeout, continue execution with the
+		// hope to find the value for it in OAuth cluster config.
+		errs = append(errs, err)
+	}
+
+	var existingAccessTokenInactivityTimeoutFloat float64
+	if existingAccessTokenInactivityTimeout != nil {
+		if f, ok := existingAccessTokenInactivityTimeout.(float64); ok {
```
sttts, line 105: this is a string, isn't it?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		// Failed to read OAuth cluster config.
+		if errors.IsNotFound(err) {
+			klog.Warning("oauth.config.openshift.io/cluster: not found")
+		}
+
+		if existingAccessTokenInactivityTimeoutFloat == 0 {
+			// As the nil value indicates that an attempt to read from existing config failed previously, return empty config.
+			return observedConfig, append(errs, err)
+		}
```
sttts, line 120: same comment as for max timeout. Don't be clever here.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: It's an int32 in the Go type. Unmarshaling could give us a float64 or int64, depending on the library in use.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+		// Return whatever is present in existing config
+		return existingConfig, append(errs, err)
+	}
+
+	observedAccessTokenInactivityTimeout := oauthConfig.Spec.TokenConfig.AccessTokenInactivityTimeout.Seconds()
```
sttts, line 125: if the user sets `"3600.5s"` for the metav1.Duration, it will marshal into the same number in the kube apiserver config as float and the unmarshalling will fail in kube-apiserver because comma values do not work with an int32 in https://github.com/openshift/api/blob/11eee842dafd4698259c5de6b0746183ea6998f5/osin/v1/types.go#L393.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	if observedAccessTokenInactivityTimeout != 0 {
+		// Build observed config only if the inactivity timeout value is set in OAuth cluster config.
+		observedConfig = buildUnstructuredTokenConfig(observedAccessTokenInactivityTimeout, tokenInactivityTimeoutPath)
+	}
+
+	if existingAccessTokenInactivityTimeoutFloat != observedAccessTokenInactivityTimeout {
+		recorder.Eventf("ObserveTokenConfig", "%s changed from %v to %v", fieldAccessTokenInactivityTimeoutSeconds,
```
sttts, line 132: ObserveOauthTokenConfig

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return observedConfig, errs
+}
+
+func buildUnstructuredTokenConfig(val interface{}, fields []string) map[string]interface{} {
+	unstructuredConfig := map[string]interface{}{}
+
+	if err := unstructured.SetNestedField(unstructuredConfig, val, fields...); err != nil {
+		// SetNestedField can return only return an error if the fields are not present in map.
```
sttts, line 144: double "return"

I don't think the sentence is correct. Instead `SetNestedField` returns an error if they *are* present, but are of wrong kind (i.e. of float type if an object is expected).

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: if i understand correctly, `float64` is the default type for numbers when decoding json formatted data. It seems like Go standard library also follows the same. You meant to say if we use a library other than standard library, this value can be decoded to int64?

Anyways I think this issue will be resolved with inactivity timeout field in osin config is changed to time.Duration.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Good catch! I am guessing this will also be resolved with new inactivity timeout field in osin config.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: my understanding was that this observer must either return the default value or whatever it observed. didn't realize that an empty config can be returned. hence the clever tricks ;)

removed the code now.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: I checked other config observers in the library-go. All those observers are using default values when cluster config is not found. Shouldn't this observer follow the same?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: > Anyways I think this issue will be resolved with inactivity timeout field in osin config is changed to time.Duration.

Any side-effects when doing that?

cc @stlaz

sttts: > You meant to say if we use a library other than standard library, this value can be decoded to int64?

the yaml decode (sigs.k8s.io/yaml) produces int64 too I think.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: yes

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,137 @@
...
+	existingAccessTokenMaxAgeSeconds, _, err := unstructured.NestedFloat64(existingConfig, tokenMaxAgePath...)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if existingAccessTokenMaxAgeSeconds != observedAccessTokenMaxAgeSeconds {
+		recorder.Eventf("ObserveOAuthTokenConfig", "%s changed from %d to %d", fieldAccessTokenMaxAgeSeconds,
+			existingAccessTokenMaxAgeSeconds,
+			observedAccessTokenMaxAgeSeconds)
+	}
```
sttts, line 68: the observer logic looks fine now, in both observers :+1:

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,137 @@
...
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	tokenConfigPath := []string{"oauthConfig", "tokenConfig"}
+	tokenMaxAgePath := append(tokenConfigPath, fieldAccessTokenMaxAgeSeconds)
+	defer func() {
+		// Prune the observed config so that it only contains access token max age field.
+		ret = configobserver.Pruned(ret, tokenMaxAgePath)
+	}()
+
+	listers, ok := genericlisters.(OAuthLister)
```
stlaz, line 38: (no action needed) I'd leave that panicking on use instead, it being null is a programmer's fault

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	observedAccessTokenInactivityTimeout := oauthConfig.Spec.TokenConfig.AccessTokenInactivityTimeout.Seconds()
+	if observedAccessTokenInactivityTimeout != 0 {
+		// Build observed config only if the inactivity timeout value is set in OAuth cluster config.
+		observedConfig = buildUnstructuredTokenConfig(observedAccessTokenInactivityTimeout, tokenInactivityTimeoutPath)
+	}
+
+	if existingAccessTokenInactivityTimeoutFloat != observedAccessTokenInactivityTimeout {
+		recorder.Eventf("ObserveOAuthTokenConfig", "%s changed from %v to %v", fieldAccessTokenInactivityTimeout,
```
stlaz, line 119: `ObserveAccessTokenInactivityTimeout`

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return buildUnstructuredTokenConfig(observedAccessTokenMaxAgeSeconds, tokenMaxAgePath), errs
+}
+
+// ObserveAccessTokenInactivityTimeout returns an unstructured fragment of KubeAPIServerConfig that has access token inactivity timeout,
+// if there is a valid value for it in OAuth cluster config.
+func ObserveAccessTokenInactivityTimeout(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	observedConfig := map[string]interface{}{}
```
stlaz, line 77: move to point of use

pkg/operator/configobserver/oauth/observe_tokenconfig.go
stlaz: if you set it observed as int, the original Go decoded config will appear different in the final comparison https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/config_observer_controller.go#L110 because the decoder at https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/config_observer_controller.go#L79 decodes numbers as floats by default.

I tried to work around this when I was writing these time-related observers for the authn operator, but keeping these values as float ended up being the most sane option.

I don't think there should be side-effects to having this field as time.Duration (I got to this PR after the openshift/api one).

Note that `osin` and `kubecontrolplane` APIs need to match type if you want to be able to reuse this observer, and `kubecontrolplane` will require the field to be pointer due to https://github.com/openshift/origin/blob/295a35c33057749cc7bdee1f2d33ddaed5b77d60/vendor/k8s.io/kubernetes/pkg/kubeapiserver/authenticator/patch_authenticator.go#L60

sttts: rounded to the next integer, but represented as float is fine with me. Make sure we don't loop. 

I wouldn't mind to see a klog.Warningf here if the observed value is not a float64.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: as discussed in Slack, just round to the next integer. That should be enough. Sub-second durations are possible, but don't make much sense.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Yes. along with checking for null, this check makes sure that `genericlisters` implements `OAuthLister` interface

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,137 @@
...
+		observedAccessTokenMaxAgeSeconds = defaultAccessTokenMaxAgeSeconds
+	}
+
+	existingAccessTokenMaxAgeSeconds, _, err := unstructured.NestedFloat64(existingConfig, tokenMaxAgePath...)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if existingAccessTokenMaxAgeSeconds != observedAccessTokenMaxAgeSeconds {
+		recorder.Eventf("ObserveOAuthTokenConfig", "%s changed from %d to %d", fieldAccessTokenMaxAgeSeconds,
```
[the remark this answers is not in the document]
vareti: Done

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: https://github.com/openshift/api/pull/674 converts the destination of this value to type `time.Duration`. With that, I am guessing we would not need to worry about losing the precision here?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
vareti: Done.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// ObserveAccessTokenMaxAgeSeconds returns an unstructured fragment of KubeAPIServerConfig that changes the default value for access token max age,
+// if there is a valid value for it in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
+	errs = []error{}
+	tokenConfigPath := []string{"oauthConfig", "tokenConfig"}
+	tokenMaxAgePath := append(tokenConfigPath, fieldAccessTokenMaxAgeSeconds)
+	defer func() {
+		// Prune the observed config so that it only contains access token max age field.
```
deads2k, line 33: why would you ever put something in here you don't want?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+		ret = configobserver.Pruned(ret, tokenMaxAgePath)
+	}()
+
+	listers, ok := genericlisters.(OAuthLister)
+	if !ok {
+		return existingConfig, append(errs, fmt.Errorf("failed to assert: given lister does not implement an OAuth lister"))
```
deads2k, line 39: this is wrong.  This is asserting that this function owns every field in the existing config.  This isn't what you want.  Follow the existing pattern to pull your config out so you don't claim everyone's fields.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	oauthConfig, err := listers.OAuthLister().Get("cluster")
+	if err != nil {
+		// Failed to read OAuth cluster config.
+		if errors.IsNotFound(err) {
```
deads2k, line 45:
```go
if errors.isnotfound(){
}
if err != nil{
// do other
}
```

to avoid nesting.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+			klog.Warning("oauth.config.openshift.io/cluster: not found")
```
deads2k, line 46: if its worth a warning, it's worth an event.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+	return observedConfig, errs
+}
+
+func buildUnstructuredTokenConfig(val interface{}, fields []string) map[string]interface{} {
+	unstructuredConfig := map[string]interface{}{}
+
+	if err := unstructured.SetNestedField(unstructuredConfig, val, fields...); err != nil {
+		// SetNestedField can return an error if one of the nesting level is not map[string]interface{}.
+		// As unstructuredConfig is empty, this error must never happen.
+		klog.Warningf("failed to write unstructured config for fields %v: %v", fields, err)
```
deads2k, line 125: this doesn't look right.  Looks like you're hiding an error.  Flatten so you don't mess up.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+
+const (
+	defaultAccessTokenMaxAgeSeconds   = float64(86400) // a day
+	fieldAccessTokenMaxAgeSeconds     = "accessTokenMaxAgeSeconds"
+	fieldAccessTokenInactivityTimeout = "accessTokenInactivityTimeout"
+)
+
+// ObserveAccessTokenMaxAgeSeconds returns an unstructured fragment of KubeAPIServerConfig that changes the default value for access token max age,
+// if there is a valid value for it in OAuth cluster config.
+func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
deads2k, line 28: no named parameters.  It makes the function harder to reason about because I have to track current state.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: because we did it in 27 different observers and 7 code paths each.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: that's why we prune.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: +1. I don't think it is, especially not repeatedly.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: See below about pruning. We fixed all library-go observers months ago and this style was the only one we didn't scream about. Every other error logic reaching the same goal was error prone and ugly.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -0,0 +1,129 @@
...
+		if errors.IsNotFound(err) {
+			klog.Warning("oauth.config.openshift.io/cluster: not found")
+		}
+		// Return whatever is present in existing config
+		return existingConfig, append(errs, err)
+	}
+
+	existingAccessTokenInactivityTimeout, _, err := unstructured.NestedString(existingConfig, tokenInactivityTimeoutPath...)
+	if err != nil {
+		errs = append(errs, err)
```
sttts, line 100: this lacks the comment why we keep going. I think the other observers have it. Without a comment this looks like a coding error.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: The comment is right. Have no strong opinion about extra error handling or not.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
sttts: ok, looks like event code is gone. klog does not create events.

PR #817, opened 2020-06-16
operatorclient: add metadata methods to interface

bertinatto:
This PR is intended to get more clarification. Will fix the tests once I know the direction.

Use case: we are creating a library to use in CSI operators. This library uses the `genericoperatorclient`. At some point, we need to update the CR status and set the right generation, like it's done here:

https://github.com/openshift/machine-api-operator/blob/5688547505e7963783f04ad0737740cfac4b6457/pkg/controller/machineset/status.go#L85-L89

As a result, I need access to the CR metadata as well (not only spec and status).

@deads2k @sttts could you take a look, please?

My main question is: is it OK to change the `OperatorClient`  interface to get the CR metadata? Or that should be done differently?

### DIFF
--- a/pkg/operator/genericoperatorclient/dynamic_operator_client.go
+++ b/pkg/operator/genericoperatorclient/dynamic_operator_client.go
@@ -47,6 +47,15 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
 	return c.informer.Informer()
 }
 
+func (c dynamicOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
+	uncastInstance, err := c.informer.Lister().Get(globalConfigName)
+	if err != nil {
+		return nil, err
+	}
+	instance := uncastInstance.(*unstructured.Unstructured)
+	return getObjectMetaFromUnstructured(instance.UnstructuredContent())
+}
+
 func (c dynamicOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
 	uncastInstance, err := c.informer.Lister().Get(globalConfigName)
 	if err != nil {
@@ -120,6 +129,22 @@ func (c dynamicOperatorClient) UpdateOperatorStatus(resourceVersion string, stat
 	}
 
 	return retStatus, nil
+}
+
+func getObjectMetaFromUnstructured(obj map[string]interface{}) (*metav1.ObjectMeta, error) {
+	uncastMeta, exists, err := unstructured.NestedMap(obj, "metadata")
+	if !exists {
+		return &metav1.ObjectMeta{}, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	ret := &metav1.ObjectMeta{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(uncastMeta, ret); err != nil {
+		return nil, err
+	}
+	return ret, nil
 }
 
 func getOperatorSpecFromUnstructured(obj map[string]interface{}) (*operatorv1.OperatorSpec, error) {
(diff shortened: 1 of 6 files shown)


pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -47,6 +47,36 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
...
+
+	ret, err := c.client.Update(context.TODO(), copy, metav1.UpdateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return getObjectMetaFromUnstructured(ret.UnstructuredContent())
+}
+
 func (c dynamicOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
```
bertinatto, line 80: As an alternative approach, I could return the metadata from here (instead of only the `resourceVersion`).

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -47,6 +47,36 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
...
+func (c dynamicOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
+	uncastInstance, err := c.informer.Lister().Get(globalConfigName)
+	if err != nil {
+		return nil, err
+	}
+	instance := uncastInstance.(*unstructured.Unstructured)
+	return getObjectMetaFromUnstructured(instance.UnstructuredContent())
+}
+
+func (c dynamicOperatorClient) UpdateObjectMeta(resourceVersion string, meta *metav1.ObjectMeta) (*metav1.ObjectMeta, error) {
```
sttts, line 59: updating objectmeta unconditionally with an enforced resource vesion looks fishy.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -47,6 +47,36 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
 	return c.informer.Informer()
 }
 
+func (c dynamicOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
```
sttts, line 50: reading is fine.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -47,6 +47,36 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
...
+
+func (c dynamicOperatorClient) UpdateObjectMeta(resourceVersion string, meta *metav1.ObjectMeta) (*metav1.ObjectMeta, error) {
+	uncastOriginal, err := c.informer.Lister().Get(globalConfigName)
+	if err != nil {
+		return nil, err
+	}
+	original := uncastOriginal.(*unstructured.Unstructured)
+
+	copy := original.DeepCopy()
+	copy.SetResourceVersion(resourceVersion)
```
sttts, line 67: this way we risk to lose data from the spec. For the other two updaters this is slightly different as the status and the spec are the input and the operator claims ownership of these values.

bertinatto: That's fair, I hadn't put too much thought on this one.

It turns out I don't need this method, so I dropped it. I just need to read the metadata.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
bertinatto: Is this a good name?

sttts: fine with me

PR #818, opened 2020-06-17
eventwatch: add controller

mfojtik:
This controller can be used to filter interesting event reasons and take action on them.
For example, this allow to register handler, that will update Prometheus counter when an interesting event is observed (or update operator conditions, etc..).

Example:

```golang
eventwatch.New().WithEventHandler("kube-apiserver", "Reason", func(*corev1.Event) error {
  // prometheusCounter++
  return nil
}).ToController(informers, eventClient, eventRecorder)
```

### DIFF
--- a/pkg/operator/eventwatch/controller.go
+++ b/pkg/operator/eventwatch/controller.go
@@ -1,0 +1,152 @@
+package eventwatch
+
+import (
+	"context"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/eventwatch/controller.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const (
+	eventAckAnnotationName = "eventwatch.openshift.io/acknowledged"
+)
+
+type Controller struct {
```
sttts, line 23: godocs

mfojtik: updated.

mfojtik:
@deads2k this will provide generic bridge between events and metrics

PR #826, opened 2020-06-29
Add helpers for service node port range configuration

abhat:
This PR adds a helper API to library-go for observing changes to the service node port range configuration.


pkg/operator/configobserver/network/observe_network.go
```diff
@@ -127,3 +127,27 @@ func validateCIDRs(in []string) error {
...
+	if err != nil {
+		recorder.Warningf("GetServiceNodePortRangeFailed", "error getting networks.%s/cluster: %v", configv1.GroupName, err)
+		return "", err
+	}
+
+	if network.Spec.ServiceNodePortRange == "" {
+		return "", nil
+	}
+
+	if _, err = utilnet.ParsePortRange(
```
squeed, line 147: nit: this linebreak seems odd?

squeed:
one very minor thing, otherwise looks good.

pkg/operator/configobserver/network/observe_network.go
abhat: fixed.

pkg/operator/configobserver/network/observe_network.go
```diff
@@ -127,3 +127,26 @@ func validateCIDRs(in []string) error {
 	}
 	return nil
 }
+
+// GetServiceNodePortRange retrieves the ServiceNodePortRange for the cluster.
+func GetServiceNodePortRange(lister configlistersv1.NetworkLister, recorder events.Recorder) (string, error) {
```
sttts, line 132: why public?

abhat: this will get called from the cluster-kube-apiserver-operator's `ConfigObserver`.

PR #829, opened 2020-07-07
CSI: add controllers for CSI operators 

bertinatto:
The following operators are being developed using this library. Eventually they will live in the `openshift` org:

AWS EBS: https://github.com/bertinatto/aws-ebs-csi-driver-operator-1/blob/lib/pkg/operator/starter.go
OpenStack Cinder : https://github.com/bertinatto/openstack-cinder-csi-driver-operator
GCE PD: https://github.com/bertinatto/gcp-pd-csi-driver-operator
Azure Disk: https://github.com/bertinatto/azure-disk-csi-driver-operator

Manila and oVirt CSI driver operators are in the works as well.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -3,6 +3,7 @@ package csicontrollerset
 import (
 	"context"
 
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -68,7 +69,10 @@ func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
 		c.csiDriverController,
 	} {
 		if controller != nil {
-			go controller.Run(ctx, 1)
+			go func() {
+				defer utilruntime.HandleCrash()
+				controller.Run(ctx, 1)
+			}()
 		}
 	}
 }
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -1,0 +1,174 @@
+package csicontrollerset
+
+import (
+	"context"
+
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/loglevel"
+	"github.com/openshift/library-go/pkg/operator/management"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/library-go/pkg/operator/csi/credentialsrequestcontroller"
+	"github.com/openshift/library-go/pkg/operator/csi/csidrivercontroller"
+)
+
+// CSIDriverControllerOptions contains file names for manifests for both CSI Driver Deployment and DaemonSet.
+type CSIDriverControllerOptions struct {
+	controllerManifest string
+	nodeManifest       string
+}
+
+// CSIDriverControllerOption is a modifier function for CSIDriverControllerOptions.
+type CSIDriverControllerOption func(*CSIDriverControllerOptions)
+
+// WithControllerService returns a CSIDriverControllerOption
+// with a Deployment (CSI Controller Service) manifest file.
+func WithControllerService(file string) CSIDriverControllerOption {
+	return func(o *CSIDriverControllerOptions) {
+		o.controllerManifest = file
+	}
+}
+
+// WithNodeService returns a CSIDriverControllerOption
+// with a DaemonSet (CSI Node Service) manifest file.
+func WithNodeService(file string) CSIDriverControllerOption {
+	return func(o *CSIDriverControllerOptions) {
+		o.nodeManifest = file
+	}
+}
+
+// CSIControllerSet contains a set of controllers that are usually used to deploy CSI Drivers.
+type CSIControllerSet struct {
+	logLevelController           factory.Controller
+	managementStateController    factory.Controller
+	staticResourcesController    factory.Controller
+	credentialsRequestController factory.Controller
+	csiDriverController          *csidrivercontroller.CSIDriverController
+
+	operatorClient v1helpers.OperatorClient
+	eventRecorder  events.Recorder
+}
+
+// Run starts all controllers initialized in the set.
+func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
+	for _, controller := range []interface {
+		Run(context.Context, int)
+	}{
+		c.logLevelController,
+		c.managementStateController,
+		c.staticResourcesController,
+		c.credentialsRequestController,
+		c.csiDriverController,
+	} {
+		if controller != nil {
+			go controller.Run(ctx, 1)
+		}
+	}
+}
+
+// WithLogLevelController returns a *ControllerSet with a log level controller initialized.
+func (c *CSIControllerSet) WithLogLevelController() *CSIControllerSet {
+	c.logLevelController = loglevel.NewClusterOperatorLoggingController(c.operatorClient, c.eventRecorder)
+	return c
+}
+
+// WithManagementStateController returns a *ControllerSet with a management state controller initialized.
+func (c *CSIControllerSet) WithManagementStateController(operandName string, supportsOperandRemoval bool) *CSIControllerSet {
+	c.managementStateController = management.NewOperatorManagementStateController(operandName, c.operatorClient, c.eventRecorder)
+	if supportsOperandRemoval {
+		management.SetOperatorNotRemovable()
+	}
+	return c
+}
+
+// WithStaticResourcesController returns a *ControllerSet with a static resources controller initialized.
+func (c *CSIControllerSet) WithStaticResourcesController(
+	name string,
+	kubeClient kubernetes.Interface,
+	kubeInformersForNamespace v1helpers.KubeInformersForNamespaces,
+	manifests resourceapply.AssetFunc,
+	files []string,
+) *CSIControllerSet {
+	c.staticResourcesController = staticresourcecontroller.NewStaticResourceController(
+		name,
+		manifests,
+		files,
+		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient),
+		c.operatorClient,
+		c.eventRecorder,
+	).AddKubeInformers(kubeInformersForNamespace)
+	return c
+}
+
+// WithCredentialsRequestController returns a *ControllerSet with a CredentialsRequestController initialized.
+func (c *CSIControllerSet) WithCredentialsRequestController(
+	name string,
+	operandNamespace string,
+	assetFunc func(string) []byte,
+	file string,
+	dynamicClient dynamic.Interface,
+) *CSIControllerSet {
+	manifestFile := assetFunc(file)
+	c.credentialsRequestController = credentialsrequestcontroller.NewCredentialsRequestController(
+		name,
+		operandNamespace,
+		manifestFile,
+		dynamicClient,
+		c.operatorClient,
+		c.eventRecorder,
+	)
+	return c
+}
+
+// WithCSIDriverController returns a *ControllerSet with a CSI Driver controller initialized.
+func (c *CSIControllerSet) WithCSIDriverController(
+	name string,
+	csiDriverName string,
+	csiDriverNamespace string,
+	assetFunc func(string) []byte,
+	kubeClient kubernetes.Interface,
+	namespacedInformerFactory informers.SharedInformerFactory,
+	setters ...CSIDriverControllerOption,
+) *CSIControllerSet {
+	cdc := csidrivercontroller.NewCSIDriverController(
+		name,
+		csiDriverName,
+		csiDriverNamespace,
+		c.operatorClient,
+		assetFunc,
+		kubeClient,
+		c.eventRecorder,
+	)
+
+	opts := &CSIDriverControllerOptions{}
+	for _, setter := range setters {
+		setter(opts)
+	}
+
+	if opts.controllerManifest != "" {
+		cdc = cdc.WithControllerService(namespacedInformerFactory.Apps().V1().Deployments(), opts.controllerManifest)
+	}
+
+	if opts.nodeManifest != "" {
+		cdc = cdc.WithNodeService(namespacedInformerFactory.Apps().V1().DaemonSets(), opts.nodeManifest)
+	}
+
+	c.csiDriverController = cdc
+
+	return c
+}
+
+// New returns a basic *ControllerSet without any controller.
+func NewCSIControllerSet(operatorClient v1helpers.OperatorClient, eventRecorder events.Recorder) *CSIControllerSet {
+	return &CSIControllerSet{
+		operatorClient: operatorClient,
+		eventRecorder:  eventRecorder,
+	}
+}
--- a/pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
+++ b/pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
@@ -1,0 +1,374 @@
+package csidrivercontroller
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+
+	opv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	driverImageEnvName              = "DRIVER_IMAGE"
+	provisionerImageEnvName         = "PROVISIONER_IMAGE"
+	attacherImageEnvName            = "ATTACHER_IMAGE"
+	resizerImageEnvName             = "RESIZER_IMAGE"
+	snapshotterImageEnvName         = "SNAPSHOTTER_IMAGE"
+	nodeDriverRegistrarImageEnvName = "NODE_DRIVER_REGISTRAR_IMAGE"
+	livenessProbeImageEnvName       = "LIVENESS_PROBE_IMAGE"
+
+	globalConfigName = "cluster"
+
+	maxRetries = 15
+)
+
+// CSIDriverController is a controller that deploys a CSI driver operand to a given namespace.
+//
+// A CSI driver operand is typically made of of a DaemonSet and a Deployment. The DaemonSet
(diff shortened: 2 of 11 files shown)


bertinatto:
@deads2k could you take a look at this and let me know if it's OK to have these controllers here?

pkg/operator/csi/controller/driver/sync.go
```diff
@@ -0,0 +1,321 @@
...
+	if c.images.livenessProbe != "" {
+		if idx := getIndex(containers, livenessProbeContainerName); idx > -1 {
+			containers[idx].Image = c.images.livenessProbe
+		}
+	}
+
+	logLevel := getLogLevel(spec.LogLevel)
+	for i, container := range containers {
+		for j, arg := range container.Args {
+			if strings.HasPrefix(arg, "--v=") {
```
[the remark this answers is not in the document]
bertinatto: Yes, this flag is initialized by `klog` package and all operators use it (I checked EBS, PD, Manila, Cinder, Azure Disk, oVirt).

I'll make a not to document this.

pkg/operator/csi/controllerset/controllerset.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, controller := range []interface {
+		Run(context.Context, int)
+	}{
+		c.logLevelController,
+		c.managementStateController,
+		c.staticResourcesController,
+		c.credentialsController,
+		c.csiDriverController,
+	} {
+		go controller.Run(ctx, 1)
```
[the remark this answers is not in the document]
bertinatto: Fixed

pkg/operator/csi/controller/credentials/controller.go
```diff
@@ -0,0 +1,106 @@
...
+		Reason: "AsExpected",
+	}
+
+	_, syncError := c.syncCredentialsRequest(status, syncContext)
+	if syncError != nil {
+		cond.Status = operatorv1.ConditionTrue
+		cond.Reason = "SyncError"
+		cond.Message = syncError.Error()
+	}
+
```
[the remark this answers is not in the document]
bertinatto: The CSI driver controller just expects the secret to be there, just like the static resources deployed by the StaticResourcesController. If any of these static resources don't exist yet, like the namespace, kubelets will fail to start the pods, but they'll try again and succeed when the resource is ready.

If the resource is is never created, the CSI driver (deployment/daemonset) will fail to start and the CSI operator will be (correctly) marked as `Degraded`.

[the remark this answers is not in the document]
bertinatto: With regards to the `c.name + "Progressing" / c.name +"Available"` conditions, typically "auxiliary" library-go controller only have a `c.name+ "Degraded"` condition.

This level of detail could be useful in the credentials controller, but IMO it's not essential, so I'd propose adding it in a next iteration (if we do want it).

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,328 @@
...
+
+	return syncErr
+}
+
+func (c *Controller) updateSyncError(status *operatorv1.OperatorStatus, err error) {
+	if err != nil {
+		// Operator is Degraded: could not finish what it was doing
+		v1helpers.SetOperatorCondition(&status.Conditions,
+			operatorv1.OperatorCondition{
+				Type:    operatorv1.OperatorStatusTypeDegraded,
```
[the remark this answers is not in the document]
bertinatto: > The overall operator status is union of CredentialsRequest and CSIDriver controllers, not just CSIDriver controller itself.

IMO we shouldn't give the credentials operator the same status as the CSI driver controller. There are operators where it's not even necessary (Azure Disk) and if it fails to provision the secret the CSI driver controller will be degraded anyway.

> Or, taking one step back, what conditions will AWSEBSDriver CR expose?

Sounds like both options are OK, but the first one seems to be a common pattern [1] so I'd go with it.

[1] https://github.com/bertinatto/library-go/blob/5e0e2b5516dd9078818d7dd572dc37389a0edc60/pkg/operator/apiserver/controllerset/apiservercontrollerset.go#L119

pkg/operator/csi/controller/credentials/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	operatorClient  v1helpers.OperatorClient
+	targetNamespace string
+	manifest        []byte
+	dynamicClient   dynamic.Interface
+}
+
+func New(
+	name,
+	targetNamespace string,
+	manifest []byte,
```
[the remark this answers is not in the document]
jsafrane: ack

pkg/operator/csi/controller/credentials/controller.go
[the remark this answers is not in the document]
jsafrane: Ok, let's add it in a next iteration.

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	return true
+}
+
+func (c *Controller) handleErr(err error, key interface{}) {
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	if c.queue.NumRequeues(key) < maxRetries {
```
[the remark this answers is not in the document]
jsafrane: `AddRateLimited` is expected and useful and it does its own exponential backoff. I was curious why *not* to call `AddRateLimited`. And the answer is below, I missed `AddAfter(key, 1*time.Minute)`. So the faulty item is re-queued, resetting the exponential backoff.

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	operatorClient.Informer().AddEventHandler(controller.eventHandler(csiDriverName))
+	controller.informersSynced = append(controller.informersSynced, operatorClient.Informer().HasSynced)
+	return controller
+}
+
+func (c *Controller) WithControllerService(informer appsinformersv1.DeploymentInformer, file string) *Controller {
+	informer.Informer().AddEventHandler(c.eventHandler("deployment"))
+	c.informersSynced = append(c.informersSynced, informer.Informer().HasSynced)
+	c.controllerManifest = c.assetFunc(file)
```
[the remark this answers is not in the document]
jsafrane: I think we can add it later, with potential `WithControllerServiceHook(<callback>)` that allows AWS EBS operator to inject cluster ID.

pkg/operator/csi/controller/driver/controller.go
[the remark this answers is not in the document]
jsafrane: `ClusterOperatorStatusController` syncs CR (say AWSEBSDriver) with ClusterOperator CR. It syncs RelatedObjects that are not available in AWSEBSDriver.Status

jsafrane:
LGTM from my side, it's a good start. We can improve the controllers in subsequent PRs.
@deads2k, can you please take a look?

pkg/operator/csi/controller/driver/sync.go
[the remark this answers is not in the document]
bertinatto: Added a note in the docs.

pkg/operator/csi/controller/driver/controller.go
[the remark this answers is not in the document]
bertinatto: Sounds good, I'm working on this but will submit a separate PR once this is merged.

pkg/operator/csi/controller/credentials/controller.go
```diff
@@ -0,0 +1,91 @@
...
+	opv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// Controller is a simple controller that maintains a CredentialsRequest static manifest.
+// It uses unstructured.Unstructured as currently there's no API type for this resource.
+type Controller struct {
```
deads2k, line 21: kube project standard realized that naming everything controller.go did bad things to glog output.  Please name your file something more specific even if you wish to keep this name (and I recommend making the struct name more specific, but that isn't externally visible)

deads2k, line 21: CredentialsRequestController?

pkg/operator/csi/controller/credentials/controller.go
```diff
@@ -0,0 +1,91 @@
...
+		operatorClient.Informer(),
+	).WithSync(
+		c.sync,
+	).ResyncEvery(
+		time.Minute,
+	).WithSyncDegradedOnError(
+		operatorClient,
+	).ToController(
+		name,
+		recorder.WithComponentSuffix(name),
```
deads2k, line 55: I recommend suffixing or prefixing the name so that you're guaranteed of a good name in events and metrics even by notice users

pkg/operator/csi/controller/credentials/resource.go
```diff
@@ -0,0 +1,118 @@
...
+
+func readCredentialRequestsOrDie(objBytes []byte) *unstructured.Unstructured {
+	udi, _, err := scheme.Codecs.UniversalDecoder().Decode(objBytes, nil, &unstructured.Unstructured{})
+	if err != nil {
+		panic(err)
+	}
+	return udi.(*unstructured.Unstructured)
+}
+
+func applyCredentialsRequest(
```
deads2k, line 44: I'm +1 to having these in resourceapply and resourceread.  Any reason not to place them there?

pkg/operator/csi/controller/credentials/controller.go
```diff
@@ -0,0 +1,91 @@
...
+type Controller struct {
+	name            string
+	operatorClient  v1helpers.OperatorClient
+	targetNamespace string
+	manifest        []byte
+	dynamicClient   dynamic.Interface
+}
+
+// New returns a Controller.
+func New(
```
deads2k, line 30: This is a good controller, but I think our overall experience would be better if your read and apply functions were in resourceread and resourceapply, and then we simply make https://github.com/openshift/library-go/blob/master/pkg/operator/staticresourcecontroller/static_resource_controller.go#L61  compatible.

Is there a reason this is impractical?  If so, simply moving the resourceread and resourceapply functions so we can later add it is ok with me, but I'd like a TODO describing the limitations here.

pkg/operator/csi/controller/credentials/resource.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
+)
+
+const (
+	credentialsRequestGroup    = "cloudcredential.openshift.io"
+	credentialsRequestVersion  = "v1"
+	credentialsRequestResource = "credentialsrequests"
+
+	specHashAnnotation = "operator.openshift.io/spec-hash"
```
deads2k, line 25: please expose the const in resourceapply and use that value.

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,374 @@
...
+// LIVENESS_PROBE_IMAGE
+//
+// 2. Log level
+//
+// The controller can also override the log level passed in to the CSI driver container.
+//
+// In order to do that, the flag `--v=` passed in to the csi-driver container is replaced with value specified
+// in the OperatorClient resource (Spec.LogLevel). That being said, it's expected that the csi-driver container sets
+// its log level based on the `--v=` flag, which is initialized by default if the driver uses the package klog.
+type Controller struct {
```
deads2k, line 80: same comment about filenames and struct names. Filename is a required change, struct name is a recommended, but optional change.

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,374 @@
...
+		images:             imagesFromEnv(),
+		queue:              workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), csiDriverName),
+		eventRecorder:      eventRecorder,
+	}
+	operatorClient.Informer().AddEventHandler(controller.eventHandler(csiDriverName))
+	controller.informersSynced = append(controller.informersSynced, operatorClient.Informer().HasSynced)
+	return controller
+}
+
+// WithControllerService returns a Controller with the CSI Controller Service.
```
deads2k, line 138: godoc needs greater specificity about this being required to be a deployment.

pkg/operator/csi/controller/driver/sync.go
```diff
@@ -0,0 +1,321 @@
...
+					Message: msg,
+					Reason:  "AsExpected",
+				})
+		}
+	}
+
+	return nil
+}
+
+func (c *Controller) getExpectedDeployment(spec *opv1.OperatorSpec) *appsv1.Deployment {
```
deads2k, line 171: One of the things we regretted about our early controllers was requiring particular shapes to the deployments.  We found that if we instead manipulated strings of yaml, we could do simple `sed s/orig/new/g` like replacements that ended up more flexible.

Notable wins about the `sed` approach
1. no special names on containers
2. no pre-determined substitution locations
3. original yaml does not need to parse into a deployment or daemonset (this is big for integer replacements for instance)
4. it's easy to wrap commands inside of bash or change the order of things.

And in your case it would allow a mutation function to come in as part of the `WithDeployment` function that accepts the bytes and spec if someone needs additional customization.  You would probably run your `sed` after the customization.

pkg/operator/csi/controller/driver/controller.go
```diff
@@ -0,0 +1,374 @@
...
+	c.queue.AddAfter(key, 1*time.Minute)
+}
+
+func logInformerEvent(kind, obj interface{}, message string) {
+	if klog.V(6) {
+		objMeta, err := meta.Accessor(obj)
+		if err != nil {
+			return
+		}
+		klog.V(6).Infof("Received event: %s %s %s", kind, objMeta.GetName(), message)
```
deads2k, line 360: super noisy, you sure about this?

deads2k:
I like your daemonset and deployment control loops a lot.  Question though, why not separate them into two different control loops that manage deployments and daemonsets separately and then compose the separate loops if you want vanity names to guide users (like them needing the node sidecar and the other thing).

pkg/operator/csi/controller/credentials/controller.go
bertinatto: Only the event recorder or the controller name as well?

We use the controller name to prefix the conditions, so I'm afraid if I prefix it with `CSIDriverController` and someone sets the controller name as `AWSEBSCSIDriverController` we are going to end up with ackward conditions in the status,

pkg/operator/csi/controller/credentials/resource.go
bertinatto: Moved this code to resourceapply, so I don't need this const here anymore

pkg/operator/csi/controller/driver/controller.go
bertinatto: Removed

pkg/operator/csi/controller/driver/controller.go
bertinatto: Added

pkg/operator/csi/controller/credentials/controller.go
bertinatto: Move the read/apply functions to resourceread and resourceapply.

I started adding code to handle the CredentialsRequest in the generic apply function but that's going to take more time that I anticipated. I'll work on that (and remove this controller) in a subsequent PR.

pkg/operator/csi/controller/driver/sync.go
bertinatto: Will address this next

bertinatto: Done. The mutation function part is a work-in-progress, I'll add that after this is merged.

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -0,0 +1,174 @@
...
+		Run(context.Context, int)
+	}{
+		c.logLevelController,
+		c.managementStateController,
+		c.staticResourcesController,
+		c.credentialsRequestController,
+		c.csiDriverController,
+	} {
+		if controller != nil {
+			go controller.Run(ctx, 1)
```
deads2k, line 71:
```go
go func(){
  utilruntime.HandleCrash()
  controller.Run(ctx, 1)
}()
```

otherwise a panic can crash the process.  That line will wire to a global panic handler configured in openshift processes.

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -0,0 +1,174 @@
...
+	staticResourcesController    factory.Controller
+	credentialsRequestController factory.Controller
+	csiDriverController          *csidrivercontroller.CSIDriverController
+
+	operatorClient v1helpers.OperatorClient
+	eventRecorder  events.Recorder
+}
+
+// Run starts all controllers initialized in the set.
+func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
```
deads2k, line 61: at some point, not in the PR, I think we want a `.Complete` step like we use in apiservers that can let users know if they're missing a new controller.  We added a `WithoutX` to one of them.

pkg/operator/csi/csicontrollerset/csi_controller_set.go
bertinatto: Fixed

pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
```diff
@@ -0,0 +1,369 @@
...
+		return err
+	}
+
+	// We only support Managed for now
+	if opSpec.ManagementState != opv1.Managed {
+		return nil
+	}
+
+	startTime := time.Now()
+	klog.Info("Starting syncing operator at ", startTime)
```
deads2k, line 200: I recommend V(4).  You'll see it when you crank up logs, but you won't get noise during normal running.

pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+func (c *CSIDriverController) handleSync(
+	resourceVersion string,
+	meta *metav1.ObjectMeta,
+	spec *opv1.OperatorSpec,
+	status *opv1.OperatorStatus,
+) error {
+	var controllerService *appsv1.Deployment
+	if c.controllerManifest != nil {
+		c, err := c.syncDeployment(spec, status)
```
deads2k, line 282: my recommendation remains to create separate controllers for these, but with the substitution as it is, we can merge this.

If you had separate controllers, each individual piece becomes an additional controller you run, right?

pkg/operator/csi/csidrivercontroller/sync.go
```diff
@@ -0,0 +1,266 @@
...
+			}
+		}
+	}
+
+	// Finally, set the conditions
+
+	// The operator does not have any prerequisites (at least now)
+	v1helpers.SetOperatorCondition(&status.Conditions,
+		opv1.OperatorCondition{
+			Type:   fmt.Sprintf("%s%s", c.name, opv1.OperatorStatusTypePrereqsSatisfied),
```
deads2k, line 83: This may have been an evolutionary dead end. I don't remember using it anytime recently.

pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
bertinatto: bumped to 4

pkg/operator/csi/csicontrollerset/csi_controller_set.go
bertinatto: Ack. Added to my TODO

pkg/operator/csi/csidrivercontroller/csi_driver_controller.go
bertinatto: Yes, some CSI operators will need another controller (not from library-go) to fetch some information from the cluster and customize the StorageClass.

I like the idea of having separate controllers, I hadn't considered it before. I haven't addressed that in this PR because it'll require lots of changes (including tests), but I'll work in this next.

pkg/operator/csi/csidrivercontroller/sync.go
bertinatto: If that's OK, I'll remove it. I thought it was mandatory

deads2k:
/lgtm
/approve

/assign @jsafrane
I'll be out tomorrow, assign to jsafrane in case there are minor tweaks.

PR #832, opened 2020-07-09
factory: handle panic in sync

mfojtik:
In case of panic() inside the sync() the controller can crash the operator and cause it to crashloop. This will prevent that using the same way as API server is handing this and it allows the controllers to recover themself.

NOTE: This might lead to wedged controller loops that never clear their conditions, but it is better than crashlooping pods?

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -156,6 +156,7 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 	var workerWaitGroup sync.WaitGroup
 	workerWaitGroup.Add(1)
 	go func() {
+		defer utilruntime.HandleCrash()
 		defer workerWaitGroup.Done()
 		for {
 			select {
@@ -214,7 +215,7 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 	if err := c.reconcile(queueCtx, syncCtx); err != nil {
 		if err == SyntheticRequeueError {
 			// logging this helps detecting wedged controllers with missing pre-requirements
-			klog.Infof("%q controller requested synthetic requeue with key %q", c.name, key)
+			klog.V(5).Infof("%q controller requested synthetic requeue with key %q", c.name, key)
 		} else {
 			utilruntime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", c.name, key, err))
 		}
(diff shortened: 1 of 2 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -137,7 +137,10 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 }
 
 func (c *baseController) Sync(ctx context.Context, syncCtx SyncContext) error {
-	return c.sync(ctx, syncCtx)
+	var err error
```
deads2k, line 140: This needs to go where you launch the go func

```
func (c *baseController) runWorker(queueCtx context.Context) {
	var workerWaitGroup sync.WaitGroup
	workerWaitGroup.Add(1)
	go func() {
		defer workerWaitGroup.Done()
		for {
			select {
			case <-queueCtx.Done():
				return
			default:
				c.processNextWorkItem(queueCtx)
			}
		}
	}()
	workerWaitGroup.Wait()
}
```

You want a defer utilruntime.HandleCrash() like we do `wait.BackoffUntil`

pkg/controller/factory/base_controller.go
```diff
@@ -171,7 +171,14 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 
 // reconcile wraps the sync() call and if operator client is set, it handle the degraded condition if sync() returns an error.
 func (c *baseController) reconcile(ctx context.Context, syncCtx SyncContext) error {
-	err := c.sync(ctx, syncCtx)
+	var err error
```
deads2k, line 174: push this to where the go func is started, like kube. That's where the risk starts.  Also, use utilrutnime.HandleCrash to be consistent with literally every other binary we have that honors `OPENSHIFT_ON_PANIC`.  We even wire this in `library-go/pkg/controller/contrrollercmd`

pkg/controller/factory/base_controller.go
```diff
@@ -156,6 +156,14 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 	var workerWaitGroup sync.WaitGroup
 	workerWaitGroup.Add(1)
 	go func() {
+		defer utilruntime.HandleCrash(func(r interface{}) {
```
deads2k, line 159: correct spot, but I suggest separately adding to `runtime.PanicHandlers` as soon as you have the recorder in the command.

PR #833, opened 2020-07-09
configobserver/cloudprovider: support cluster-config-operator

patrickdillon:
Adds support for cloud provider configs generated by the cluster-config-operator.
Configobserver should prefer configs generated by the config operator, but fall back
to config listed in infrastructure resource for backwards compatibility.

https://issues.redhat.com/browse/CORS-1483

This has been tested successfully vendoring my personal branch to the kube-controller-manager-operator. 

Before patching KCM-O:
```
$ oc describe cm cloud-config-7 -n openshift-kube-controller-manager
Name:         cloud-config-7
Namespace:    openshift-kube-controller-manager
Labels:       <none>
Annotations:  <none>

Data
====
config:
```

After patching: 
```
$ oc describe cm cloud-config-9 -n openshift-kube-controller-manager
Name:         cloud-config-9
Namespace:    openshift-kube-controller-manager
Labels:       <none>
Annotations:  <none>

Data
====
cloud.conf:
----
```

We can see that the key `config` has changed to `cloud.conf` which indicates the KCM cloud config has been updated from the `openshift-config/cloud-provider-config` to the `openshift-config-managed/kube-cloud-config`
cc @abhinavdahiya @deads2k

### DIFF
--- a/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
+++ b/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
@@ -7,24 +7,27 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
 
 	configv1 "github.com/openshift/api/config/v1"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
-
-	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
 )
 
 const (
-	cloudProviderConfFilePath = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/%s"
-	configNamespace           = "openshift-config"
+	cloudProviderConfFilePath       = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/%s"
+	configNamespace                 = "openshift-config"
+	machineSpecifiedConfigNamespace = "openshift-config-managed"
+	machineSpecifiedConfig          = "kube-cloud-config"
 )
 
 // InfrastructureLister lists infrastrucre information and allows resources to be synced
 type InfrastructureLister interface {
 	InfrastructureLister() configlistersv1.InfrastructureLister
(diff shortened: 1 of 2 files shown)


pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
```diff
@@ -72,13 +75,27 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
 
 	sourceCloudConfigMap := infrastructure.Spec.CloudConfig.Name
 	sourceCloudConfigNamespace := configNamespace
+	sourceCloudConfigKey := infrastructure.Spec.CloudConfig.Key
+
+	// If a managed cloud-provider config is available, it should be used instead of the default.
+	_, err = listers.ConfigMapLister().ConfigMaps(machineSpecifiedConfigNamespace).Get(machineSpecifiedConfig)
+	if err != nil {
```
soltysh, line 82:
```go
if !errors.IsNotFound(err) {
```
is sufficient.

patrickdillon: @soltysh thank you for the review. I think using your suggestion with my original code would omit the case where `err == nil && err.IsNotFound(err)` in which case we want to keep the default values. So I think my code was unclear and I rewrote this conditional. PTAL and let me know if there are any further changes.

soltysh: Yeah, this is much better now. Thx!

pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
```diff
@@ -72,13 +75,25 @@ func (c *cloudProviderObserver) ObserveCloudProviderNames(genericListers configo
...
+	}
+
 	sourceLocation := resourcesynccontroller.ResourceLocation{
 		Namespace: sourceCloudConfigNamespace,
 		Name:      sourceCloudConfigMap,
 	}
 
 	// we set cloudprovider configmap values only for some cloud providers.
-	validCloudProviders := sets.NewString("azure", "gce", "openstack", "vsphere")
+	validCloudProviders := sets.NewString("aws", "azure", "gce", "openstack", "vsphere")
```
patrickdillon, line 96: Hm I need to make sure this is ok. I added "aws" because in the case of custom region endpoints, aws will have a cloud config, but otherwise won't.

Perhaps that missing cloud config will pose a problem now that aws is in the list. I am running another e2e-aws test in openshift/cluster-kube-controller-manager-operator#430 but perhaps someone may know off the top of their head

patrickdillon: I tested locally on aws and everything looked fine. Also e2e-aws on dependent pr openshift/cluster-kube-controller-manager-operator#430 passed so everything looks good to me.

PR #834, opened 2020-07-14
loglevel: allow to override the default log level

mfojtik:
This allows to have debug CI jobs against PR's that have debug commit which increases the operator loglevel (which is normally set by operator config).

### DIFF
--- a/pkg/operator/loglevel/logging_controller.go
+++ b/pkg/operator/loglevel/logging_controller.go
@@ -15,14 +15,23 @@ type LogLevelController struct {
 	// for unit tests only
 	setLogLevelFn func(operatorv1.LogLevel) error
 	getLogLevelFn func() (operatorv1.LogLevel, bool)
+
+	defaultLogLevel operatorv1.LogLevel
 }
 
-// sets the klog level based on desired state
+// NewClusterOperatorLoggingController sets a klog level for the operator based on the operator config.
+// If the loglevel is not set the default "Normal" level will be used.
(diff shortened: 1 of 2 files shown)


pkg/operator/loglevel/logging_controller.go
```diff
@@ -15,14 +15,23 @@ type LogLevelController struct {
...
 }
 
-// sets the klog level based on desired state
+// NewClusterOperatorLoggingController sets a klog level for the operator based on the operator config.
+// If the loglevel is not set the default "Normal" level will be used.
 func NewClusterOperatorLoggingController(operatorClient operatorv1helpers.OperatorClient, recorder events.Recorder) factory.Controller {
+	return NewClusterOperatorLoggingControllerWithLogLevel(operatorClient, operatorv1.Normal, recorder)
+}
+
+// NewClusterOperatorLoggingControllerWithLogLevel sets a klog level for the operator based on the operator config.
```
sttts, line 28: the comment does not match

mfojtik: what does not match?

mfojtik: it will still set what you specify in operator CR :-) but if you don't it will default

sttts: NewClusterOperatorLoggingControllerWithLogLevel sets a klog level for the operator based on the operator config, using the given default log level if the operator config does not specify anything.

PR #835, opened 2020-07-14
eventwatch: allow to match reason substring

mfojtik:
This allows to match the event reasons just by substring. An example is `FooBarPanic` event reason (we send this events when a controller panic occur) can be matched with `.WithEventHandler("namespace", "Panic", handlerFn)`.

### DIFF
--- a/pkg/operator/eventwatch/controller.go
+++ b/pkg/operator/eventwatch/controller.go
@@ -33,9 +33,9 @@ type Controller struct {
 }
 
 type eventHandler struct {
-	reason    string
-	namespace string
-	process   func(event *corev1.Event) error
+	reasonPattern string
+	namespace     string
+	process       func(event *corev1.Event) error
 }
 
 type Builder struct {
@@ -47,13 +47,14 @@ func New() *Builder {
 	return &Builder{}
 }
 
-// WithEventHandler add handler for event matching the namespace and the reason.
+// WithEventHandler add handler for event matching the namespace and the reason pattern.
 // This can be called multiple times.
-func (b *Builder) WithEventHandler(namespace, reason string, processEvent func(event *corev1.Event) error) *Builder {
+// The reason pattern can contain glob ("*") as prefix or suffix. If no glob is used a strict match is required
+func (b *Builder) WithEventHandler(namespace, reasonPattern string, processEvent func(event *corev1.Event) error) *Builder {
 	b.eventConfig = append(b.eventConfig, eventHandler{
-		reason:    reason,
-		namespace: namespace,
-		process:   processEvent,
+		reasonPattern: reasonPattern,
+		namespace:     namespace,
+		process:       processEvent,
 	})
 	return b
 }
@@ -99,7 +100,7 @@ func (c *Controller) getEventHandler(eventKey string) *eventHandler {
 		return nil
 	}
 	for i := range c.events {
-		if c.events[i].namespace == namespace && c.events[i].reason == reason {
+		if c.events[i].namespace == namespace && reasonMatch(c.events[i].reasonPattern, reason) {
 			return &c.events[i]
 		}
 	}
--- a/pkg/operator/eventwatch/glob.go
+++ b/pkg/operator/eventwatch/glob.go
@@ -1,0 +1,61 @@
+package eventwatch
+
+import (
+	"strings"
+)
+
+// reasonMatch will test a string pattern, potentially containing globs, against a
+// subject string. The result is a simple true/false, determining whether or
+// not the glob pattern matched the subject text.
+//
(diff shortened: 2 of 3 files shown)


pkg/operator/eventwatch/controller.go
```diff
@@ -99,7 +99,7 @@ func (c *Controller) getEventHandler(eventKey string) *eventHandler {
 		return nil
 	}
 	for i := range c.events {
-		if c.events[i].namespace == namespace && c.events[i].reason == reason {
+		if c.events[i].namespace == namespace && strings.Contains(reason, c.events[i].reason) {
```
sttts, line 102: this means `reason` is actually `reasonSubstring` everywhere? Would prefer to make that explicit rather than introducing magic behaviour here.

pkg/operator/eventwatch/controller.go
```diff
@@ -24,7 +24,7 @@ const (
 )
 
 // Controller observes the events in given informer namespaces and match them with configured event handlers.
-// If the event reason and namespace match the configured, then the process() function is called in handler on that event.
+// If the event reasonSubstring and namespace match the configured, then the process() function is called in handler on that event.
```
sttts, line 27: needs a little update

pkg/operator/eventwatch/controller.go
```diff
@@ -47,13 +47,13 @@ func New() *Builder {
 	return &Builder{}
 }
 
-// WithEventHandler add handler for event matching the namespace and the reason.
+// WithEventHandler add handler for event matching the namespace and the reason substring.
 // This can be called multiple times.
-func (b *Builder) WithEventHandler(namespace, reason string, processEvent func(event *corev1.Event) error) *Builder {
+func (b *Builder) WithEventHandler(namespace, reasonSubstring string, processEvent func(event *corev1.Event) error) *Builder {
```
sttts, line 52: I wonder whether this is really a good approach. I don't like this fuzzy matching. Alternatives: `*Panic` for postfix match.

mfojtik: i was thinking to add some matcher (we use some in other places), but I thought it might be an overkill

pkg/operator/eventwatch/controller.go
mfojtik: goland...

pkg/operator/eventwatch/glob.go
```diff
@@ -0,0 +1,61 @@
...
+		// No globs in pattern, so test for equality
+		return reason == pattern
+	}
+
+	leadingGlob := strings.HasPrefix(pattern, "*")
+	trailingGlob := strings.HasSuffix(pattern, "*")
+	end := len(parts) - 1
+
+	// Go over the leading parts and ensure they match.
+	for i := 0; i < end; i++ {
```
sttts, line 39: `for i := range parts`?

pkg/operator/eventwatch/glob.go
```diff
@@ -0,0 +1,61 @@
...
+		idx := strings.Index(reason, parts[i])
```
sttts, line 40: the case of `*foo` seems to go through fine, but is full of edge cases. Maybe make that explicit. Easier to read.

pkg/operator/eventwatch/controller.go
```diff
@@ -24,7 +24,7 @@ const (
 )
 
 // Controller observes the events in given informer namespaces and match them with configured event handlers.
-// If the event reason and namespace match the configured, then the process() function is called in handler on that event.
+// If the event reason substring and namespace match the configured, then the process() function is called in handler on that event.
```
sttts, line 27: sorry for being dense: it's not a substring anymore.

pkg/operator/eventwatch/glob.go
mfojtik: this is copied from the original repo and it is tested with unit test... i don't want to hack this up too much

pkg/operator/eventwatch/controller.go
```diff
@@ -24,7 +24,7 @@ const (
 )
 
 // Controller observes the events in given informer namespaces and match them with configured event handlers.
-// If the event reason and namespace match the configured, then the process() function is called in handler on that event.
+// If the event reason  and namespace match the configured, then the process() function is called in handler on that event.
```
sttts, line 27: nit: double space

pkg/operator/eventwatch/glob.go
sttts: I was impressed by the clever algorithm. Now you steal me my amazement :D

sttts:
/lgtm
/approve
/hold

Unhold at will if you hate double spaces less than I do.

PR #839, opened 2020-07-20
move the etcd config observer to library-go

p0lyn0mial:
this PR moves existing observer to the shared library as it will be used by CAO and OAS-O.

I slightly changed the code so that it supports passing a custom path under which the configuration will be saved.

### DIFF
--- a/pkg/operator/configobserver/apiserver/listers.go
+++ b/pkg/operator/configobserver/apiserver/listers.go
@@ -1,0 +1,9 @@
+package apiserver
+
+import (
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+)
+
+type APIServerLister interface {
+	APIServerLister() configlistersv1.APIServerLister
+}
--- a/pkg/operator/configobserver/etcd/listers.go
+++ b/pkg/operator/configobserver/etcd/listers.go
@@ -4,8 +4,8 @@ import (
 	corelistersv1 "k8s.io/client-go/listers/core/v1"
 )
 
-type ConfigmapLister interface {
-	ConfigmapLister() corelistersv1.ConfigMapLister
+type ConfigMapLister interface {
+	ConfigMapLister() corelistersv1.ConfigMapLister
 }
 
 type EndpointsLister interface {
--- a/pkg/operator/configobserver/etcd/listers.go
+++ b/pkg/operator/configobserver/etcd/listers.go
@@ -1,0 +1,13 @@
+package etcd
+
+import (
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+)
+
+type ConfigmapLister interface {
+	ConfigmapLister() corelistersv1.ConfigMapLister
+}
+
+type EndpointsLister interface {
+	EndpointsLister() corelistersv1.EndpointsLister
+}
--- a/pkg/operator/configobserver/apiserver/listers_test.go
+++ b/pkg/operator/configobserver/apiserver/listers_test.go
@@ -1,0 +1,23 @@
+package apiserver
+
+import (
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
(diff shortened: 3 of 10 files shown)


p0lyn0mial:
/hold 

for https://github.com/openshift/cluster-openshift-apiserver-operator/pull/371 and squashing

pkg/operator/configobserver/apiserver/listers.go
```diff
@@ -1 +1,27 @@
...
+// Listers provides a unified interface for config observers used in apiserver package
+type Listers interface {
+	APIServerLister
+	ConfigmapLister
+	EndpointsLister
+}
+
+type APIServerLister interface {
+	APIServerLister() configlistersv1.APIServerLister
+	PreRunHasSynced() []cache.InformerSynced
```
sttts, line 10: what is this? Why only in one?

pkg/operator/configobserver/apiserver/observe_etcd_endpoints.go
```diff
@@ -104,3 +104,77 @@ func ObserveStorageURLs(genericListers configobserver.Listers, recorder events.R
 
 	return observedConfig, errs
 }
+
+// observeStorageURLsFromOldEndPoint observes the storage URL config.
+func observeStorageURLsFromOldEndPoint(genericListers configobserver.Listers, recorder events.Recorder, currentConfig map[string]interface{}) (map[string]interface{}, []error) {
```
sttts: this long func surprises me in an "adjust" commit.

pkg/operator/configobserver/apiserver/listers.go
```diff
@@ -2,27 +2,15 @@ package apiserver
 
 import (
 	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
-	corelistersv1 "k8s.io/client-go/listers/core/v1"
```
sttts: could this commit be factored into the previous ones?

pkg/operator/configobserver/etcd/listers.go
```diff
@@ -1,4 +1,4 @@
-package etcdobserver
```
sttts: factor into previous commit

hexfusion:
etcd end of this looks good, more or less the same logic as used by OAS. The only comment is it would be nice to have a single version of this defined I think the only difference between KAS and OAS here is KAS appends localhost to etcdURLs.

p0lyn0mial:
> etcd end of this looks good, more or less the same logic as used by OAS

@hexfusion it must look exactly the same as the logic used by OAS :) The only change I did was to add support for passing a custom path under which the configuration will be saved.

p0lyn0mial:
> The only comment is it would be nice to have a single version of this defined I think the only difference between KAS and OAS here is KAS appends localhost to etcdURLs

let me see what it takes to change the code to support KAS.

p0lyn0mial:
it looks like KAS:

1. doesn't support a fall back function when the config map wasn't found https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go#L54

2. always appends 'localhost' https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go#L91

3. supports having two config paths (old and new) - https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/etcdendpoints/observe_etcd_endpoints.go#L26

p0lyn0mial:
the proof PRs look good https://github.com/openshift/cluster-kube-apiserver-operator/pull/912 and https://github.com/openshift/cluster-openshift-apiserver-operator/pull/371

hexfusion:
thanks @p0lyn0mial 

/lgtm

PR #841, opened 2020-07-21
NewNestedConfigObserver controller

p0lyn0mial:
This PR adds `NewNestedConfigObserver` controller that creates a config observer that watches changes to a nested field (nestedConfigPath) in the config.

The new function is useful when the config is shared across multiple controllers in the same process.

For example, passing `operandOne` in the `nestedConfigPath` argument will create a controller that watches only values stored under that path.

```
"operandOne": {
 "foo": "one",
 "bar": "two",
 "baz": "three"
 },
 "operandTwo": {
  "foo1": "one",
  "bar1": "two",
  "baz1": "three"
 }
}`
``

### DIFF
--- a/pkg/operator/configobserver/config_observer_controller.go
+++ b/pkg/operator/configobserver/config_observer_controller.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/imdario/mergo"
@@ -13,6 +14,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/tools/cache"
@@ -45,8 +47,11 @@ type ConfigObserver struct {
 	observers []ObserveConfigFunc
 
 	operatorClient v1helpers.OperatorClient
+
 	// listers are used by config observers to retrieve necessary resources
 	listers Listers
+
+	nestedConfigPath []string
 }
 
 func NewConfigObserver(
@@ -56,10 +61,48 @@ func NewConfigObserver(
 	informers []factory.Informer,
 	observers ...ObserveConfigFunc,
 ) factory.Controller {
+	return NewNestedConfigObserver(
+		operatorClient,
+		eventRecorder,
+		listers,
+		informers,
+		nil,
+		observers...,
+	)
+}
+
+// NewNestedConfigObserver creates a config observer that watches changes to a nested field (nestedConfigPath) in the config.
+// Useful when the config is shared across multiple controllers in the same process.
+//
+// Example:
+//
+// Given the following configuration, you could run two separate controllers and point each to its own section.
+// The first controller would be responsible for "oauthAPIServer" and the second for "oauthServer" section.
+//
+// "observedConfig": {
+//   "oauthAPIServer": {
+//     "apiServerArguments": {"tls-min-version": "VersionTLS12"}
+//   },
+//   "oauthServer": {
+//     "corsAllowedOrigins": [ "//127\\.0\\.0\\.1(:|$)","//localhost(:|$)"]
+//   }
+// }
+//
+// oauthAPIController    := NewNestedConfigObserver(..., []string{"oauthAPIServer"}
+// oauthServerController := NewNestedConfigObserver(..., []string{"oauthServer"}
+func NewNestedConfigObserver(
+	operatorClient v1helpers.OperatorClient,
+	eventRecorder events.Recorder,
+	listers Listers,
+	informers []factory.Informer,
+	nestedConfigPath []string,
+	observers ...ObserveConfigFunc,
+) factory.Controller {
 	c := &ConfigObserver{
-		operatorClient: operatorClient,
-		observers:      observers,
-		listers:        listers,
+		operatorClient:   operatorClient,
+		observers:        observers,
+		listers:          listers,
+		nestedConfigPath: nestedConfigPath,
 	}
 
 	return factory.New().ResyncEvery(time.Second).WithSync(c.sync).WithInformers(append(informers, listersToInformer(listers)...)...).ToController("ConfigObserver", eventRecorder.WithComponentSuffix("config-observer"))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/operator/configobserver/config_observer_controller.go [hunk 1]
```diff
@@ -55,11 +58,32 @@ func NewConfigObserver(
...
+		eventRecorder,
+		listers,
+		informers,
+		nil,
+		observers...,
+	)
+}
+
+// NewNestedConfigObserver creates a config observer that watches changes to a nested field (nestedConfigPath) in the config.
+// Useful when the config is shared across multiple controllers in the same process.
```
deads2k, line 73: A concrete example of two controllers and their divisions would help make the reason for this clear.

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -135,6 +153,50 @@ func (c ConfigObserver) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	return configError
 }
 
+func (c ConfigObserver) updateObservedConfig(syncCtx factory.SyncContext, existingConfig map[string]interface{}, mergedObservedConfig map[string]interface{}) error {
+	existingConfigMaybeNested := map[string]interface{}{}
+	mergedObservedConfigMaybeNested := map[string]interface{}{}
+	if len(c.nestedConfigPath) > 0 {
```
deads2k, line 159: extract the filtering to a single method for me?  I think I want to re-use the filtering to avoid the defer handling we have in other spots.

deads2k:
```go
filterConfig(nestedConfigPath, config, existingErr) (mergedObservedConfig map[string]interface{}, error){
}
```

called twice.

deads2k:
```go

func configobserver() (map[string]string, error) {
	return map[string]string{"foo": "bar"}, nil
}

type filter struct {
	path []string
}

func (f filter) Filter(in map[string]string, err error) (map[string]string, error) {
	// do the filter
	return in, err
}

func main() {
	ret, err := filter{path: []string{"foo"}}.Filter(configobserver())

	fmt.Printf("%v and %v\n", ret, err)
}

```

this looks pretty good.  clean filter

deads2k: I see I wrote a non-sequitor.  I'll do that some other time.

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -135,6 +153,50 @@ func (c ConfigObserver) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	return configError
 }
 
+func (c ConfigObserver) updateObservedConfig(syncCtx factory.SyncContext, existingConfig map[string]interface{}, mergedObservedConfig map[string]interface{}) error {
```
deads2k, line 156: I think something like this would help me have parsed this.

```go
func() {
  if c.nestedConfigPath == 0 {
    if !deepEqual(existingConfig, mergedObservedConfig){
      return c.updateConfig(mergedObservedConfig)
    }
    return nil
  }

  // do the other stuff
}

func (c ConfigObserver) updateConfig(config map) error{
  if _, _, err := v1helpers.UpdateSpec(c.operatorClient, v1helpers.UpdateObservedConfigFn(updatedConfig)); err != nil {
    // At this point we failed to write the updated config. If we are permanently broken, do not pile the errors from observers
    // but instead reset the errors and only report single error condition.
    syncCtx.Recorder().Warningf("ObservedConfigWriteError", "Failed to write observed config: %v", err)
    return fmt.Errorf("error writing updated observed config: %v", err)
  }
  return nil
}

```

deads2k:
fine with this in principle.

/approve

pkg/operator/configobserver/config_observer_controller.go
p0lyn0mial: done.

pkg/operator/configobserver/config_observer_controller.go
p0lyn0mial: done

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -135,6 +171,46 @@ func (c ConfigObserver) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+	}
+	mergedObservedConfigNested, _, err := unstructured.NestedMap(mergedObservedConfig, c.nestedConfigPath...)
+	if err != nil {
+		return fmt.Errorf("unable to extract the merged config under %v, err %v", c.nestedConfigPath, err)
+	}
+	if !equality.Semantic.DeepEqual(existingConfigNested, mergedObservedConfigNested) {
+		syncCtx.Recorder().Eventf("ObservedConfigChanged", "Writing updated section (%q) of observed config: %q", strings.Join(c.nestedConfigPath, "/"), diff.ObjectDiff(existingConfigNested, mergedObservedConfigNested))
+
+		updatedConfig := existingConfig
+		if err := unstructured.SetNestedField(updatedConfig, mergedObservedConfigNested, c.nestedConfigPath...); err != nil {
```
stlaz, line 195: You should only modify your values to prevent races.

At this point, `existingConfig` may already contain obsolete values for all the other observers running in parallel with you, wheareas these other observers might already be in a new cycle where their `existingConfig` may contain obsolete values to your new config.

I think you should create a new function based on `type UpdateOperatorSpecFunc func(spec *operatorv1.OperatorSpec) error` that just takes the observed config and only replaces its managed prefixed config.

p0lyn0mial: yeah, I can't write only my changes because that would remove everything else. I thought that the potential conflicts would be automatically removed - PATH - will check.

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -135,6 +172,57 @@ func (c ConfigObserver) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+	if _, _, err := v1helpers.UpdateSpec(c.operatorClient, updateConfigHelper(updatedMaybeNestedConfig)); err != nil {
+		// At this point we failed to write the updated config. If we are permanently broken, do not pile the errors from observers
+		// but instead reset the errors and only report single error condition.
+		syncCtx.Recorder().Warningf("ObservedConfigWriteError", "Failed to write observed config: %v", err)
+		return fmt.Errorf("error writing updated observed config: %v", err)
+	}
+	return nil
+}
+
+// updateConfigHelper returns a helper function for updating the nested config.
```
stlaz, line 211: nit `updateNestedConfigHelper`

PR #845, opened 2020-07-22
Audit policy config observer

tkashem:


tkashem:
holding it, as soon as the openshift/api PR https://github.com/openshift/api/pull/681 merges we can remove the replace from go.mod

/hold

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// AuditPolicyPathGetter allows the observer to be agnostic of the source of audit profile(s).
+type AuditPolicyPathGetter interface {
+	// AuditPolicyPathGetter returns the path to the audit policy file (associated with the
+	// given profile) in the static manifest folder.
+	GetAuditPolicyPath(profile string) (string, error)
+
+	// GetDefaultAuditPolicyPath returns the path to the default audit policy.
+	GetDefaultAuditPolicyPath() string
```
sttts, line 21: why isn't that `GetAuditPolicyPath("Default")`?

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+		if len(current) > 0 {
+			auditPolicyPath = current[0]
+		}
+
+		defaultObservedConfig := map[string]interface{}{}
+		if err := unstructured.SetNestedStringSlice(defaultObservedConfig, []string{auditPolicyPath}, apiServerArgumentsAuditPath...); err != nil {
+			// If we are here, then we are not setting any default for observed audit profile.
+			// looks like returning the existing config is the best alternative here. This puts a constraint on the
+			// apiserver to have a default configuration in its bindata.
+			return existingConfig, append(errs, err)
```
sttts, line 51: this is wrong. You have to prune the existing config: the input has values from all observers. But you are only allowed to return your fields.

Compare pruning in the defer func in the cloud provider observer.

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+		apiServer, err := listers.APIServerLister().Get("cluster")
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				klog.Warningf("apiserver.config.openshift.io/cluster: not found")
+
+				// return the audit policy path from the current config if set, otherwise we use the default configuration.
+				return defaultObservedConfig, errs
+			}
+
+			return defaultObservedConfig, append(errs, err)
```
sttts, line 64: on non-not-found errors we return the existing config. This here would change the config for the apiserver. We don't want that. It creates flapping on error situations.

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		desiredProfile := string(apiServer.Spec.Audit.Profile)
+		if len(desiredProfile) == 0 {
+			return defaultObservedConfig, errs
```
sttts, line 69: `// this should never happen because we have a default value on the field`

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+		// return it as the observed config in case of error, otherwise we resort to the default policy path.
+		auditPolicyPath := pathGetter.GetDefaultAuditPolicyPath()
+		current, _, err := unstructured.NestedStringSlice(existingConfig, apiServerArgumentsAuditPath...)
+		if err != nil {
+			// best-effort approach, let's keep going.
+			errs = append(errs, err)
+		}
+		if len(current) > 0 {
+			auditPolicyPath = current[0]
+		}
```
sttts, line 44: This should only be need further down where you compare with the new value and create an event on change. Move it there.

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return defaultObservedConfig, append(errs, fmt.Errorf("audit profile is not valid name=%s", desiredProfile))
+		}
+		if desiredAuditPolicyPath == auditPolicyPath {
+			return defaultObservedConfig, errs
+		}
+
+		observedConfig := map[string]interface{}{}
+		if err := unstructured.SetNestedStringSlice(observedConfig, []string{desiredAuditPolicyPath}, apiServerArgumentsAuditPath...); err != nil {
+			return defaultObservedConfig, append(errs, fmt.Errorf("failed to set desired audit profile in observed config name=%s", desiredProfile))
```
sttts, line 82: return old config, not default. Avoids flapping.

pkg/operator/configobserver/apiserver/observe_audit.go
tkashem: to reduce coupling, library-go doesn't need to know the defaulting mechanism, it just needs to know:
- the path to the default audit profle
- given an audit profile, it needs to know the desired path.
on the other hand, the apiserver/component that defines the audit profiles should know well which audit profile is default.

I know we are defaulting at the CRD level, but i think the observer implementation should still guard for empty profile name and other error conditions.

pkg/operator/configobserver/apiserver/observe_audit.go
tkashem: yeah, I have added the pruner.

pkg/operator/configobserver/apiserver/observe_audit.go
sttts: This func is only called to default the value in the old config. This whole block of code is only needed for the event. Just compare to empty string. Same result.

tkashem: It resorts to the default value if the existing/current config is not set. when the existing config is not set and the observer function encounters a lister error or any other before it can set the new/desired configuration, what should the observer return?

pkg/operator/configobserver/apiserver/observe_audit.go
tkashem: on any lister error, the observer function is returning the current/existing config.

pkg/operator/configobserver/apiserver/observe_audit.go
tkashem: the CRD is not co-located; the observer being a core func, it's a guard against if the CRD schema is accidentally changed.

pkg/operator/configobserver/apiserver/observe_audit.go
tkashem: it's returning the current config, I changed the variable name. I think it boils down to the question I posted here - https://github.com/openshift/library-go/pull/845#discussion_r459668601.

tkashem:
good for review

/hold cancel

pkg/operator/configobserver/apiserver/observe_audit.go
```diff
@@ -0,0 +1,87 @@
...
+
+				return existingConfig, errs
+			}
+
+			return existingConfig, append(errs, err)
+		}
+
+		desiredProfile := string(apiServer.Spec.Audit.Profile)
+		if len(desiredProfile) == 0 {
+			return existingConfig, errs
```
sttts, line 50: empty config

PR #847, opened 2020-07-23
CSI: Accept custom instance name

bertinatto:
PR https://github.com/oVirt/csi-driver-operator/pull/12/files landed, so the CSI Driver controller needs to react to a different resource: `clustercsidrivers`. In addition to that, the controller should only react if the CR has a certain name.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivercontroller/csi_driver_controller_test.go
+++ b/pkg/operator/csi/csidrivercontroller/csi_driver_controller_test.go
@@ -124,6 +124,7 @@ func newOperator(test testCase, t *testing.T) *testContext {
 		controllerName,
 		operandName,
 		operandNamespace,
+		"cluster",
 		fakeOperatorClient,
 		makeFakeManifest,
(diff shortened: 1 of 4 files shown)


pkg/operator/csi/csidrivercontroller/csi_driver_controller_test.go
```diff
@@ -124,6 +124,7 @@ func newOperator(test testCase, t *testing.T) *testContext {
 		controllerName,
 		operandName,
 		operandNamespace,
+		"cluster",
```
huffmanca, line 127: Do we want to change the name away from `cluster` in this e2e test? Will this cause issues if other tests are running in parallel? For instance, if we ever get the EFS CSI driver and both are used in an AWS cluster?

jsafrane: This is unit test, the CR name IMO does not really matter there.

PR #848, opened 2020-07-23
config/clusteroperator: add etcd revision resource

hexfusion:
usecase: for disaster recovery and in general it is important to have an expected list of resources that should be found in each revision for each static pod operator. By moving these definitions to library-go we can import without needing to add operators as dependencies directly.


hexfusion:
proof https://github.com/openshift/cluster-etcd-operator/pull/403

hexfusion:
After thinking about this for a bit, going to move the payload for revision into revision-status in a followup PR.

PR #849, opened 2020-07-24
adds missing tests for ApplyDeployment and ApplyDeploymentWithForce

p0lyn0mial:


p0lyn0mial:
now, when I have the tests I can start actually changing the methods.

PR #851, opened 2020-07-24
add configobserver for the master url arg for kcm and ks

deads2k:
promotion of https://github.com/openshift/cluster-kube-controller-manager-operator/tree/master/pkg/operator/configobservation/masterurl for inclusion in kube-scheduler.

will open proof


deads2k:
proof PR https://github.com/openshift/cluster-kube-scheduler-operator/pull/266

PR #860, opened 2020-07-31
tweaks audit policies

p0lyn0mial:
makes it more generic so that it can be used by the API servers.

### DIFF
--- a/pkg/operator/apiserver/audit/audit_policies.go
+++ b/pkg/operator/apiserver/audit/audit_policies.go
@@ -24,19 +24,22 @@ const (
 
 // WithAuditPolicies is meant to wrap a standard Asset function usually provided by an operator.
 // It delegates to GetAuditPolicies when the filename matches the predicate for retrieving a audit policy config map for target namespace.
-func WithAuditPolicies(targetNamespace string, assetDelegateFunc resourceapply.AssetFunc) resourceapply.AssetFunc {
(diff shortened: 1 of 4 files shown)


pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -33,7 +33,17 @@ func WithAuditPolicies(targetName string, targetNamespace string, assetDelegateF
...
 
-// getRawAuditPolicies returns a raw config map that holds the audit policies for the target namespaces
+// GetAuditPolicies returns a config map that holds the audit policies for the target namespaces and name
+func GetAuditPolicies(targetName, targetNamespace string) (*corev1.ConfigMap, error) {
+	rawAuditPolicies, err := getRawAuditPolicies(targetName, targetNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	return resourceread.ReadConfigMapV1OrDie(rawAuditPolicies), nil
```
tkashem, line 43: do we have a `ReadConfigMapV1` version that returns an `error`?

p0lyn0mial: I’m not aware of any, but this is a common practice. It is considered a programmer error when it fails

p0lyn0mial:
/assign @sttts 

for approval

PR #861, opened 2020-08-03
prefix nested observer condition to avoid racing

stlaz:
without this, if there are multiple observers, these would be stepping on each other's toes, trying to set the same condition

/assign @p0lyn0mial

### DIFF
--- a/pkg/operator/configobserver/config_observer_controller.go
+++ b/pkg/operator/configobserver/config_observer_controller.go
@@ -51,7 +51,8 @@ type ConfigObserver struct {
 	// listers are used by config observers to retrieve necessary resources
 	listers Listers
 
-	nestedConfigPath []string
+	nestedConfigPath      []string
+	degradedConditionType string
 }
 
 func NewConfigObserver(
@@ -67,6 +68,7 @@ func NewConfigObserver(
 		listers,
 		informers,
 		nil,
+		"",
 		observers...,
 	)
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -96,13 +98,18 @@ func NewNestedConfigObserver(
 	listers Listers,
 	informers []factory.Informer,
 	nestedConfigPath []string,
+	conditionPrefix string,
```
p0lyn0mial, line 101: Just thinking if it would be possible to take it from `nestedConfigPath` somehow?

p0lyn0mial: for example, CAO uses `NewNestedConfigObserver(..., []string{"oauthAPIServer"}`.

So we could take the first element and make it pretty so that it ends up being  `OauthAPIServerConfigObservationDegraded`, not practical?

stlaz: I was wondering about the same, and thought whether anyone would go as deep as layer 2 different nesting. Probably not, so we can just as well use the first element, either seems fine.

p0lyn0mial: yeah, taking the first element seems to be enough.

How we will make it pretty? Is there an easy way to convert it to a camel case string?

stlaz: I don't think there's an easy way, the simplest thing we can do is just to take the first rune of the string and turn it into upper case :shrug:

p0lyn0mial: alright, so let's make it explicit.

stlaz:
/hold cancel
we agreed off-github that this is fine, too :)

pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -99,10 +102,20 @@ func NewNestedConfigObserver(
...
-		listers:          listers,
-		nestedConfigPath: nestedConfigPath,
+		operatorClient:        operatorClient,
+		observers:             observers,
+		listers:               listers,
+		nestedConfigPath:      nestedConfigPath,
+		degradedConditionType: condition.ConfigObservationDegradedConditionType,
+	}
+	if len(nestedConfigPath) > 0 {
+		nestedRoot := nestedConfigPath[0]
```
sttts, line 112: put a line here with an example. I have a hard time to follow your printf code and rune logic.

PR #864, opened 2020-08-12
Validate operatorSpec.logLevel

jsafrane:
Only report an event (and log it) on failures and fall back to the default log level (`Normal` in most cases).

We can't do hard validation to prevent from breaking working clusters with potentially bad `logLevels`.

### DIFF
--- a/pkg/operator/loglevel/logging_controller.go
+++ b/pkg/operator/loglevel/logging_controller.go
@@ -51,6 +51,11 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
 		desiredLogLevel = c.defaultLogLevel
 	}
 
+	if !ValidLogLevel(desiredLogLevel) {
+		syncCtx.Recorder().Warningf("OperatorLogLevelInvalid", "Invalid logLevel %q, falling back to %q", desiredLogLevel, c.defaultLogLevel)
+		desiredLogLevel = c.defaultLogLevel
+	}
+
 	// correct log level is set and it matches the expected log level from operator operatorSpec, do nothing.
 	if !isUnknown && currentLogLevel == desiredLogLevel {
 		return nil
--- a/pkg/operator/loglevel/logging_controller_test.go
+++ b/pkg/operator/loglevel/logging_controller_test.go
@@ -88,6 +88,16 @@ func TestClusterOperatorLoggingController(t *testing.T) {
 			},
 			startingVerbosity: 4,
 			expectedVerbosity: 2,
+			defaultLoglevel:   &operatorv1.Normal,
+			evalEvents: func(events []*corev1.Event, t *testing.T) {
+				if len(events) != 1 {
+					t.Errorf("expected exactly one event, got %d", len(events))
+					return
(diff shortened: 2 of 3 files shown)


jsafrane:
The unit tests are disabled, but I checked the newly added one works.

sttts:
> The unit tests are disabled

What do you mean?

jsafrane:
The tests are waiting for some fixes after rebase: https://github.com/openshift/library-go/blob/master/pkg/operator/loglevel/logging_controller_test.go#L33

sanchezl:
> The tests are waiting for some fixes after rebase:
> 
> https://github.com/openshift/library-go/blob/f5cb4d19a4fea9298bcb1b34f45dd52caba4130d/pkg/operator/loglevel/logging_controller_test.go#L33

Maybe @marun could tell us more. Seems only one of the existing test cases (`when_OperatorLogLevel_is_set_to_Debug_operator_must_set_V(4)_when_it_is_currently_V(2)`) is broken. Maybe we can enable all but that on in a follow up.

marun:
> > The tests are waiting for some fixes after rebase:
> > https://github.com/openshift/library-go/blob/f5cb4d19a4fea9298bcb1b34f45dd52caba4130d/pkg/operator/loglevel/logging_controller_test.go#L33
> 
> Maybe @marun could tell us more. Seems only one of the existing test cases (`when_OperatorLogLevel_is_set_to_Debug_operator_must_set_V(4)_when_it_is_currently_V(2)`) is broken. Maybe we can enable all but that on in a follow up.

The test in question relied on being able to fake out the logging implementation's Verbose parameter. klog v2 changed the type of that value from boolean to a struct whose members are private, which precludes that strategy from working. The test will need to be rewritten to account for this structural change before it will be able to pass.

pkg/operator/loglevel/logging_controller.go
```diff
@@ -51,6 +51,11 @@ func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContex
 		desiredLogLevel = c.defaultLogLevel
 	}
 
+	if !ValidLogLevel(desiredLogLevel) {
+		syncCtx.Recorder().Warningf("OperatorLogLevelInvalid", "Invalid logLevel %q, falling back to %q", desiredLogLevel, c.defaultLogLevel)
+		desiredLogLevel = c.defaultLogLevel
```
sttts, line 56: @deads2k Jordan was working on this mechanism to get warning strings to the user (kubectl), used for deprecation warnings in his case. Is that live already? Could it be used via admission to tell the user that the loglevel is invalid without failing?

deads2k: > @deads2k Jordan was working on this mechanism to get warning strings to the user (kubectl), used for deprecation warnings in his case. Is that live already? Could it be used via admission to tell the user that the loglevel is invalid without failing?

Maybe.  I guess I would want to just force clear bad data one release.

deads2k:
very gentle.  If you'd like to force clear bad data, I'm +1 to that.

/approve

PR #865, opened 2020-08-13
static pod operator should not update versions until the image is new

deads2k:
This fixes a bug, but I don't know which one.

Proof PRs are needed for the big static pod operator.  Will link shortly.
- [ ] https://github.com/openshift/cluster-kube-apiserver-operator/pull/932
- [ ] https://github.com/openshift/cluster-etcd-operator/pull/422
- [ ] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/434
- [ ] https://github.com/openshift/cluster-kube-scheduler-operator/pull/272

### DIFF
--- a/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
+++ b/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
@@ -117,19 +117,28 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
 		}
 	}
 
-	if len(images) == 0 {
+	switch {
+	case len(images) == 0:
 		syncCtx.Recorder().Warningf("MissingVersion", "no image found for operand pod")
-	} else if len(images) > 1 {
+
+	case len(images) > 1:
 		syncCtx.Recorder().Eventf("MultipleVersions", "multiple versions found, probably in transition: %v", strings.Join(images.List(), ","))
-	} else {
-		c.versionRecorder.SetVersion(
-			c.operandName,
-			status.VersionForOperandFromEnv(),
-		)
-		c.versionRecorder.SetVersion(
-			"operator",
-			status.VersionForOperatorFromEnv(),
-		)
+
+	default: // we have one image
+		// if have a consistent image and if that image the same as the current operand image, then we can update the version to reflect our new version
+		if images.List()[0] == status.ImageForOperandFromEnv() {
+			c.versionRecorder.SetVersion(
+				c.operandName,
+				status.VersionForOperandFromEnv(),
+			)
+			c.versionRecorder.SetVersion(
+				"operator",
+				status.VersionForOperatorFromEnv(),
+			)
+
+		} else {
+			// otherwise, we have one image, but it is NOT the current operand image so we don't update the version
+		}
 	}
 
 	// update failing condition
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -117,19 +117,28 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
...
+	default: // we have one image
+		// if have a consistent image and if that image the same as the current operand image, then we can update the version to reflect our new version
+		if images.List()[0] == status.ImageForOperandFromEnv() {
+			c.versionRecorder.SetVersion(
+				c.operandName,
+				status.VersionForOperandFromEnv(),
+			)
+			c.versionRecorder.SetVersion(
+				"operator",
+				status.VersionForOperatorFromEnv(),
```
wking, line 136: Is this setting the `operator` version for the operator as a whole, or just for this one static pod controller (with the operator's `operator` version leveling after all its sub-controllers have leveled)?

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -117,19 +117,28 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
...
-			status.VersionForOperandFromEnv(),
-		)
-		c.versionRecorder.SetVersion(
-			"operator",
-			status.VersionForOperatorFromEnv(),
-		)
+
+	default: // we have one image
+		// if have a consistent image and if that image the same as the current operand image, then we can update the version to reflect our new version
+		if images.List()[0] == status.ImageForOperandFromEnv() {
```
wking, line 129: nit:

```go
case images.List()[0] == status.ImageForOperandFromEnv():
  ...
default:
  // we have one image, but it is NOT the current operand image so we don't update the version
}
```

would unroll one level of nesting, right?

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
deads2k: > Is this setting the `operator` version for the operator as a whole, or just for this one static pod controller (with the operator's `operator` version leveling after all its sub-controllers have leveled)?

it's voting. anything needing a different version sets it separately.  But static pod operators don't get choices, it changes when their static pods are up to date.

wking: > ...it's voting. anything needing a different version sets it separately...

Is there aggregation logic that tallies votes somewhere, similar to what `UnionClusterCondition` does for conditions?  Grepping for `"operator"` is not turning up an aggregator for me...

wking: For what it's worth, I'd expect each controller to set an `operandName`, as you have a bit further up here.  And then operator-level logic that looks at all the controllers and bumps `operator` when they all match their target operand version.

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
sanchezl: I think the intent it clearer as is.

PR #869, opened 2020-08-24
Create a resuable verify package for image release verification

jottofar:
Both the oc client application and the CVO do image release verification. Image release verification was recently added to the oc client and at that time any code required was simply copied from the CVO were it was originally developed. This change moves the code required for image release verification, the CVO package verify, from the CVO into library-go for reuse by CVO and the oc client application and applies any required changes to make it reusable by both.


wking:
/close

Issue with me approving #837 was the `vendor` bump.

PR #871, opened 2020-08-26
Bug 1858763: static-pod-installer: remove deadlock by recreating installers that disappear

sttts:
The line

```
```

**Note:** just before this code there is the `ensureInstallerPod` invocation. We run into the bug when the pod apply call fails (e.g. of some network issue).

**Note:** when the installer pod disappears again and again (for whatever reason), this could hotloop (of course we do the usual backoff in the controller though). But this is the same
for the existing OOM case.


sttts:
/hold

until proof bump https://github.com/openshift/cluster-kube-apiserver-operator/pull/939 shows some green.

PR #874, opened 2020-08-31
factory: do not panic on cache sync when we shutdown

mfojtik:
This change fix a case when the operator is shutting down **before** the caches are synced. In that case, the main context is cancelled and the WaitForCacheSync() receives stop which result in panic. While these panics are harmless, they look ugly and they might trigger some post-processing filters in CI that look for panics.

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -24,6 +24,8 @@ import (
 // This can be also done by re-adding the key to queue, but this is cheaper and more convenient.
 var SyntheticRequeueError = errors.New("synthetic requeue request")
 
+var defaultCacheSyncTimeout = 10 * time.Minute
+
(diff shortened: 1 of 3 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -60,11 +60,18 @@ func (s *scheduledJob) Run() {
...
+	if !cache.WaitForNamedCacheSync(c.name, cacheSyncCtx.Done(), c.cachesToSync...) {
+		// the parent context is closed, it means we are shutting down, do not call panic()
+		if ctx.Err() != nil {
+			return
+		}
+		panic("failed to wait for cache to sync")
+	}
+
 	// HandleCrash recovers panics
 	defer utilruntime.HandleCrash()
```
sttts, line 74: move that to the top again ?

PR #875, opened 2020-09-02
Fix crypto configuration for golang 1.14

marun:
The 1.19.0 bump tried to ensure library-go compatibility with the addition of a _SHA256 suffix to the following cipher suites:

 - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
 - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256

The naive approach taken by the bump's test fixes was to include both the new and old names in the cipher map, which satisfied the check in TestConstantMaps that all golang cipher suites were mapped to a name.

However, CipherSuitesToNamesOrDie was not tested, and when called by kube-apiserver would panic due to the suite ids of the ciphers in question mapping to both the old and new cipher names.

This commit ensures CipherSuitesToNamesOrDie is tested to not panic when called with the default set of ciphers, and ensures that attempting to convert the suite ids in question always returns the new name.


soltysh (approved):
/lgtm

pkg/crypto/crypto.go
```diff
@@ -155,17 +155,17 @@ var openSSLToIANACiphersMap = map[string]string{
...
-	"ECDHE-RSA-AES128-SHA256":       "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",   // 0xC0,0x27
-	"AES128-GCM-SHA256":             "TLS_RSA_WITH_AES_128_GCM_SHA256",         // 0x00,0x9C
-	"AES256-GCM-SHA384":             "TLS_RSA_WITH_AES_256_GCM_SHA384",         // 0x00,0x9D
-	"AES128-SHA256":                 "TLS_RSA_WITH_AES_128_CBC_SHA256",         // 0x00,0x3C
+	"ECDHE-ECDSA-AES128-GCM-SHA256": "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",       // 0xC0,0x2B
+	"ECDHE-RSA-AES128-GCM-SHA256":   "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",         // 0xC0,0x2F
+	"ECDHE-ECDSA-AES256-GCM-SHA384": "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",       // 0xC0,0x2C
+	"ECDHE-RSA-AES256-GCM-SHA384":   "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",         // 0xC0,0x30
+	"ECDHE-ECDSA-CHACHA20-POLY1305": "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256", // 0xCC,0xA9
+	"ECDHE-RSA-CHACHA20-POLY1305":   "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",   // 0xCC,0xA8
```
sttts, line 163: are these the same?

sttts: ic, they are

PR #876, opened 2020-09-02
controller factory: allow conditional queue key add

stlaz:
The conditional work queue pushing simplifies the logic for
controllers that take a certain resource type but only want to
react to the resource changes when a condition is met,
e.g. an annotation is set to a certain value.

Would be useful at https://github.com/openshift/service-ca-operator/pull/121/files#diff-eec55cf875708a3ee2197838c800c3faR80

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -44,62 +44,66 @@ func (c syncContext) Recorder() events.Recorder {
 	return c.eventRecorder
 }
 
-func (c syncContext) isInterestingNamespace(obj interface{}, interestingNamespaces sets.String) (bool, bool) {
-	ns, ok := obj.(*corev1.Namespace)
-	if !ok {
-		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
-		if ok {
-			if ns, ok := tombstone.Obj.(*corev1.Namespace); ok {
-				return true, interestingNamespaces.Has(ns.Name)
-			}
-		}
-		return false, false
-	}
-	return true, interestingNamespaces.Has(ns.Name)
-}
-
 // eventHandler provides default event handler that is added to an informers passed to controller factory.
-func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeyFunc, interestingNamespaces sets.String) cache.ResourceEventHandler {
-	return cache.ResourceEventHandlerFuncs{
+func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeyFunc, filter EventFilterFunc) cache.ResourceEventHandler {
+	resourceEventHandler := cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			isNamespace, isInteresting := c.isInterestingNamespace(obj, interestingNamespaces)
 			runtimeObj, ok := obj.(runtime.Object)
 			if !ok {
 				utilruntime.HandleError(fmt.Errorf("added object %+v is not runtime Object", obj))
 				return
 			}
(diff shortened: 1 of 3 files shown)


pkg/controller/factory/factory.go
```diff
@@ -97,6 +101,13 @@ func (f *Factory) WithBareInformers(informers ...Informer) *Factory {
 // is called.
 // Pass the queueKeyFn you want to use to transform the informer runtime.Object into string key used by work queue.
 func (f *Factory) WithInformersQueueKeyFunc(queueKeyFn ObjectQueueKeyFunc, informers ...Informer) *Factory {
+	return f.WithInformersCondQueueKeyFunc(
+		func(obj runtime.Object) (string, bool) { return queueKeyFn(obj), true },
+		informers...,
+	)
+}
+
+func (f *Factory) WithInformersCondQueueKeyFunc(queueKeyFn ConditionalObjectQueueKeyFunc, informers ...Informer) *Factory {
```
mfojtik, line 110: make it ugly! WithInformersConditionalQueueKeyFunc``

mfojtik, line 110: please add godoc and include example

pkg/controller/factory/factory.go
```diff
@@ -82,6 +92,18 @@ func (f *Factory) WithInformers(informers ...Informer) *Factory {
 	return f
 }
 
+// WithFilteredEventsInformers is used to register event handlers and get the caches synchronized functions.
+// Pass the informers you want to use to react to changes on resources. If informer event is observed, then the Sync() function
+// is called.
+// Pass filter to filter out events that should not trigger Sync() call.
+func (f *Factory) WithFilteredEventsInformers(filter EventFilterFunc, informers ...Informer) *Factory {
+	f.filteredInformers = append(f.filteredInformers, filteredInformers{
```
sttts, line 100: having two informer lists is looking pretty strange. Can't we unify that?

stlaz: sure can, I'll merge those

stlaz:
fixed an issue where the filters were not in fact passed to the `eventHandler()`

pkg/controller/factory/controller_context.go
```diff
@@ -44,62 +44,74 @@ func (c syncContext) Recorder() events.Recorder {
...
+// as a filter for objects that are namespaces and they concern us in any way
+type namespaceChecker struct {
+	interestingNamespaces sets.String
+}
+
+func namespaceCheckingFunc(interestingNamespaces []string) func(obj interface{}) bool {
+	return namespaceChecker{
+		interestingNamespaces: sets.NewString(interestingNamespaces...),
+	}.isInterestingNamespace
+}
```
sttts, line 98: use a closure directly.

```
func namespaceCheckingFunc(interestingNamespaces []string) func(obj interface{}) bool {
  return func(obj interface{}) bool {
    ...
  }
}
```

pkg/controller/factory/factory.go
```diff
@@ -78,7 +87,19 @@ func (f *Factory) WithSync(syncFn SyncFunc) *Factory {
 // Pass informers you want to use to react to changes on resources. If informer event is observed, then the Sync() function
 // is called.
 func (f *Factory) WithInformers(informers ...Informer) *Factory {
-	f.informers = append(f.informers, informers...)
+	f.informers = append(f.informers, filteredInformers{informers: informers})
```
sttts, line 90: call WithFilteredEventsInformers here with nil

pkg/controller/factory/controller_context.go
stlaz: it's now returning closure :+1:

PR #881, opened 2020-09-07
[release-4.4] Bug 1876484: static-pod-installer: recreate installers that disappeared

soltysh:
The line

```
```

/assign @sttts


sttts:
Overriding the valid-bug, because these are just the library PRs. The bumps will go the normal way.

PR #882, opened 2020-09-07
[release-4.3] Bug 1876486: static-pod-installer: recreate installers that disappeared

soltysh:
The line

```
```

/assign @sttts


glide.lock
```diff
@@ -273,7 +273,7 @@ imports:
   - libcontainer/system
   - libcontainer/user
 - name: github.com/openshift/api
-  version: 371e06d3a8a73b95b67bf6d0b7fb6259c67a5e62
+  version: a843dc3262a0cdc5fe6c20620f3e47f5f905efd1
```
sttts, line 276: why was this bumped? Is it the right branch?

soltysh: yup, it's 4.3 https://github.com/openshift/library-go/blob/release-4.3/glide.yaml#L16

sttts:
Overriding the valid-bug, because these are just the library PRs. The bumps will go the normal way.

PR #883, opened 2020-09-07
resourcesync: add provided by to sync error message

mfojtik:
If the `Provider` field is set for the sync source and the sync will fail with error, the error message will include the `... (provided by FOO)` string that helps identify the resource provider.
Practically, when an ingress operator is not running and we fail to sync a config map provided by this operator, the auth operator will error out with "config map foo not found", but it does not say what provides that config map (or it does not mention where to start looking).

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -133,6 +133,14 @@ func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation
 	return nil
 }
 
+// errorWithProvider provides a finger of blame in case a source resource cannot be retrieved.
+func errorWithProvider(provider string, err error) error {
+	if len(provider) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -133,6 +133,14 @@ func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation
 	return nil
 }
 
+// errorWithProvider provides a finger of blame in case a source resource cannot be retrieved.
+func errorWithProvider(provider string, err error) error {
+	if len(provider) > 0 {
+		return fmt.Errorf("%w (for more details check the %q that is supposed to provide this resource", err, provider)
```
sttts, line 139: )

sttts: s/for more details//

PR #885, opened 2020-09-08
Bug 1876913: Add global git configs

adambkaplan:
Allow git global configurations to be added.


adambkaplan:
Required for fix in https://github.com/openshift/builder/pull/174

PR #887, opened 2020-09-09
installer: log connection errors on retry

mfojtik:
Log retries on connection errors with the connection errors so we know the installer actually retry them.

### DIFF
--- a/pkg/operator/resource/retry/retry.go
+++ b/pkg/operator/resource/retry/retry.go
@@ -6,6 +6,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/retry/retry.go
```diff
@@ -3,6 +3,8 @@ package retry
 import (
 	"context"
 
+	"k8s.io/klog/v2"
+
```
tnozicka, line 7: nit: space

pkg/operator/resource/retry/retry_test.go
```diff
@@ -3,9 +3,12 @@ package retry
 import (
 	"context"
 	"fmt"
+	"syscall"
 	"testing"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/net"
+
```
tnozicka, line 11: nit: space

tnozicka (approved):

PR #888, opened 2020-09-09
installer: add reason with blame

mfojtik:
Provide blame given the reason message. When the installer pod fails because of networking is not working, suggest admins to look into networking-operator for possible reason.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/blame.go
+++ b/pkg/operator/staticpod/controller/installer/blame.go
@@ -1,0 +1,17 @@
+package installer
+
+import (
+	"fmt"
+	"strings"
+)
+
+// reasonWithBlame inspects the reason given by the installer pod, which usually include the last log line from the installer pod
+// and try to point a finger on component that likely caused its failure to provide better operator message.
+func reasonWithBlame(originalReason string) string {
+	switch {
+	case strings.Contains(originalReason, "no route to host"):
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/blame.go
```diff
@@ -0,0 +1,17 @@
+package installer
+
+import (
+	"fmt"
+	"strings"
+)
+
+// inspectInstallerReason inspects the reason given by the installer pod, which usually include the last log line from the installer pod
```
sttts, line 8: reasonWithBlame

pkg/operator/staticpod/controller/installer/blame.go
```diff
@@ -0,0 +1,17 @@
...
+// and try to point a finger on component that likely caused its failure to provide better operator message.
+func reasonWithBlame(reason string) string {
+	switch {
+	case strings.Contains(reason, "no route to host"):
+		return fmt.Sprintf("%s (check the network-operator for possible cause)")
```
sttts, line 13: maybe even clearer like: most probably node networking is misbehaving

mfojtik: updated

mfojtik:
fixed broken Sprintf, adding lgtm back

PR #890, opened 2020-09-09
Bug 1878007: Validate operatorSpec.logLevel

vareti:
Report event (and log it) on failures and fall back to the default log level
(Normal in most cases).

We can't do hard validation to prevent from breaking working clusters with
potentially bad logLevels.

(cherry picked from commit 29b87426f36212e3a2873a707fa03869507a9082)


sttts:
Would not log an event for this. A warning is fine in the logs.

PR #891, opened 2020-09-10
Fix panic in cache sync

tnozicka:
Avoid panic on failure to sync caches and just exit.

/cc @mfojtik @deads2k

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -62,6 +62,18 @@ func (s *scheduledJob) Run() {
 	s.queue.Add(DefaultQueueKey)
 }
 
+func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheSyncs ...cache.InformerSynced) error {
+	klog.Infof("Waiting for caches to sync for %s", controllerName)
+
+	if !cache.WaitForCacheSync(stopCh, cacheSyncs...) {
+		return fmt.Errorf("unable to sync caches for %s", controllerName)
+	}
+
+	klog.Infof("Caches are synced for %s ", controllerName)
+
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -62,19 +63,33 @@ func (s *scheduledJob) Run() {
...
+	err := waitForNamedCacheSync(c.name, cacheSyncCtx.Done(), c.cachesToSync...)
+	if err != nil {
+		select {
+		case <-ctx.Done():
+			// Exit gracefully
 			return
-		}
-		panic("failed to wait for cache to sync")
+		default:
+			klog.Exit(err)
```
deads2k, line 92: tell me which ones didn't sync

mfojtik, line 92: one nit: pass a controller name here pls.

deads2k, line 92: add a comment explaining
```
// if caches did not sync after 10 minutes, it has taken oddly long and we should provide feedback. Since the control loops will never start, it is safer to exit with a good message than to continue with a dead loop. 
// TODO consider making this behavior configurable.
```

tnozicka: it's already there from https://github.com/openshift/library-go/pull/891/files#diff-dfb55c7c09634be481990d4f70ce291cR70

tnozicka: done

mfojtik:
@deads2k to approve

pkg/controller/factory/base_controller.go
tnozicka: we agreed on slack that this would be complicated so printing the controller name and having informer errors in the logs is enough for now

PR #893, opened 2020-09-14
apps: add human readable deployment pod containers status helper

mfojtik:


pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,45 @@
...
+	"strings"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+)
+
+// PodContainersStatus return detailed information about deployment pods and the containers status in human readable format.
+// This can be used for cluster operator condition messages or logging.
+func PodContainersStatus(deployment *appsv1.Deployment, podClient corelistersv1.PodLister) ([]string, error) {
```
sttts, line 14: and wire this into deployment controller.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -4,8 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/openshift/library-go/pkg/apps/deployment"
+
```
sttts, line 11: order

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,52 @@
...
+		podConditions := []string{
+			fmt.Sprintf("phase=%s", deploymentPods[i].Status.Phase),
+		}
+		for _, c := range deploymentPods[i].Status.Conditions {
+			podConditions = append(podConditions, fmt.Sprintf("%s=%s (%s#%s)", c.Type, c.Status, c.Reason, c.Message))
+		}
+		deploymentPodsStates = append(deploymentPodsStates, fmt.Sprintf("pod %q is %s (%s)", deploymentPods[i].Name, strings.Join(podConditions, ","), strings.Join(containerStates, ",")))
+	}
+	if len(deploymentPodsStates) == 0 {
+		deploymentPodsStates = append(deploymentPodsStates, fmt.Sprintf("no pods matching labels %q", labels.SelectorFromSet(deployment.Spec.Template.Labels).String()))
```
sttts, line 49: no pods found with labels

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,52 @@
...
+	for i := range deploymentPods {
+		containerStates := []string{}
+		for _, c := range append(deploymentPods[i].Status.ContainerStatuses, deploymentPods[i].Status.InitContainerStatuses...) {
+			ready := "ready"
+			if !c.Ready {
+				ready = "NOT ready"
+			}
+			switch {
+			case c.State.Running != nil:
+				containerStates = append(containerStates, fmt.Sprintf("container %q is %s running since %s", c.Name, ready, c.State.Running.StartedAt))
```
sttts, line 29: skip this

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,52 @@
...
+			case c.State.Waiting != nil:
+				containerStates = append(containerStates, fmt.Sprintf("container %q is %s waiting: %s - %s", c.Name, ready, c.State.Waiting.Reason, c.State.Waiting.Message))
```
sttts, line 31: don't use colon `:` in these messages. It's hard to read the final string for a pod

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,95 @@
...
+		succeededCount := 0
+		restartCount := 0
+		notReadyCount := 0
+
+		for _, c := range append(deploymentPods[i].Status.ContainerStatuses, deploymentPods[i].Status.InitContainerStatuses...) {
+			if !c.Ready {
+				notReadyCount++
+			}
+			if c.RestartCount > 1 {
+				restartCount++
```
sttts, line 55: >= 1 ?

mfojtik: no, i lost my comment there, but I want to tolerate 1 restart because many pods reload their certs

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,95 @@
...
+		if succeededCount > 0 {
+			containerStates = append(containerStates, fmt.Sprintf("%s succeeded in pod %s", containerPlural(succeededCount), deploymentPods[i].Name))
+		}
+		if restartCount > 0 {
+			containerStates = append(containerStates, fmt.Sprintf("%s crash-looping in %s", containerPlural(restartCount), podPhase(deploymentPods[i])))
+		}
+		if notReadyCount > 0 {
+			containerStates = append(containerStates, fmt.Sprintf("%s not ready in %s", containerPlural(notReadyCount), podPhase(deploymentPods[i])))
+		}
+	}
```
sttts, line 89: these are still too many. One message per pod, not more.

sttts: And only print non-normal states. If a container is running there is nothing to say. If the whole pod is ready, there is nothing to say.

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			// we tolerate 1 restart for cert reloading
+			if c.RestartCount > 1 {
+				restartCount++
+			}
+			if c.State.Terminated != nil && c.State.Terminated.ExitCode != 0 {
+				failedCount++
+			}
+			if c.State.Waiting != nil {
+				waitingCount++
```
sttts, line 60: Where would we see this common message from kubelet like "CNI not ready, bla bla" ?

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,82 @@
...
+
+		for _, c := range append(deploymentPods[i].Status.ContainerStatuses, deploymentPods[i].Status.InitContainerStatuses...) {
+			if !c.Ready {
+				notReadyCount++
+			}
+			// we tolerate 1 restart for cert reloading
+			if c.RestartCount > 1 {
+				restartCount++
+			}
+			if c.State.Terminated != nil && c.State.Terminated.ExitCode != 0 {
```
sttts, line 56: would this pass the c.Ready above?

I would have expected `if c.Ready { continue }` above and a default case below.

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,82 @@
...
+			containerStates = append(containerStates, fmt.Sprintf("%s crash-looping in %s", containerPlural(restartCount), podPhase(deploymentPods[i])))
+		}
+		if notReadyCount > 0 {
+			containerStates = append(containerStates, fmt.Sprintf("%s not ready in %s", containerPlural(notReadyCount), podPhase(deploymentPods[i])))
+		}
+	}
+
+	if len(deploymentPods) == 0 {
+		containerStates = append(containerStates, fmt.Sprintf("no pods found with labels %q", labels.SelectorFromSet(deployment.Spec.Template.Labels).String()))
+	}
```
sttts, line 80: what happens if one pod is missing?

pkg/apps/deployment/status.go
mfojtik: in container status eventually (message)... we have controller for installer pods

pkg/apps/deployment/status.go
mfojtik: that is reported in the higher level where we count the available replicas vs. replicas. This helper is just a pretty printer.

pkg/apps/deployment/status.go
sttts: ack

pkg/apps/deployment/status.go
```diff
@@ -0,0 +1,96 @@
...
+	if crashloop {
+		crash = "crashlooping "
+	}
+	if c == 1 {
+		return fmt.Sprintf("%d %scontainer is", c, crash)
+	}
+	return fmt.Sprintf("%d %scontainers are", c, crash)
+}
+
+func podPhase(pod *corev1.Pod) string {
```
sttts, line 24: nit: helpers at the bottom

pkg/apps/deployment/status.go
sttts: worth to expose some special cases here? Like pointing directly to networking instead of nodes.

mfojtik: we don't point anywhere here, we just saying the container is pending/waiting... you wanted people to go inspect the containers and now you want special handling :-)

pkg/apps/deployment/status_test.go
```diff
@@ -0,0 +1,223 @@
...
+					State: corev1.ContainerState{
+						Waiting: &corev1.ContainerStateWaiting{
+							Reason:  "IsSlow",
+							Message: "Slow container",
+						},
+					},
+				})
+			})),
+			deployment:       makeDeployment("foo", defaultLabels),
+			expectedMessages: []string{"1 container is waiting in pending pod test"},
```
sttts, line 131: nit: drop "1"

pkg/apps/deployment/status_test.go
```diff
@@ -0,0 +1,223 @@
...
+					State: corev1.ContainerState{
+						Waiting: &corev1.ContainerStateWaiting{
+							Reason:  "IsSlow",
+							Message: "Slow container",
+						},
+					},
+				})
+			})),
+			deployment:       makeDeployment("foo", defaultLabels),
+			expectedMessages: []string{"1 crashlooping container is waiting in pending pod test"},
```
sttts, line 150: `pod "test"`

pkg/apps/deployment/status_test.go
mfojtik: i like to know how many :-)

pkg/apps/deployment/status_test.go
mfojtik: i hate quotes :)

pkg/apps/deployment/status.go
sttts: can be follow-up. but sure, I think about the next step :)

pkg/apps/deployment/status_test.go
sttts: `1 container is waiting` does not contradict that there is another one. Remember I am mathematician.

pkg/apps/deployment/status_test.go
sttts: quotes where quotes make sense.

sttts: this is a name. Either `pod %q` or `%s pod`. The current variant causes parse errors.

pkg/apps/deployment/status_test.go
mfojtik: can't argue with that. changed.

pkg/apps/deployment/status_test.go
mfojtik: ok :-) second variant then. fixed.

PR #896, opened 2020-09-18
Fix reading of CRD v1

jsafrane:
Use the right SchemeGroupVersion. Reading v1 CRD manifest panics with:
```
panic: converting (v1.CustomResourceDefinition) to (v1beta1.CustomResourceDefinition): unknown conversion [recovered]
```

/assign @sttts @deads2k


jsafrane:
`fileobserver TestObserverPolling` failures do not look related
/retest

PR #897, opened 2020-09-18
Bug 1879991: Fix bug in reflector not recovering from "Too large resource version" 

vareti:
This bumps imports the recent fix in reflector/informer that was made in upstream client-go

https://github.com/kubernetes/kubernetes/pull/92688


vareti:
/close

it was decided to bump only the operators to keep the changes to minimal

PR #898, opened 2020-09-18
Replace bitbucket.org/ww/goautoneg with its mirror

dmage:
The package bitbucket.org/ww/goautoneg is not available anymore, so it is replaced with its mirror github.com/munnerz/goautoneg.


dmage:
/assign @deads2k 

Do we need BZ for this?

deads2k:
/approve

leaving lgtm with @sttts and/or @tnozicka

pkg/apiserver/httprequest/httprequest.go
```diff
@@ -5,7 +5,7 @@ import (
 	"net/http"
 	"strings"
 
-	"bitbucket.org/ww/goautoneg"
```
sttts, line 8: I would expect a replace in go.mod.

What is upstream kube doing?

dmage: replace doesn't work, upstream uses github.com/munnerz/goautoneg

```
go: github.com/munnerz/goautoneg@v0.0.0-20191010083416-a7dc8b61c822 used for two different module paths (bitbucket.org/ww/goautoneg and github.com/munnerz/goautoneg)
```

dmage: https://github.com/kubernetes/kubernetes/blob/2d8fbd61afe9e2bf207be88320fd9c3001885a96/go.mod#L75

PR #901, opened 2020-09-22
Bug 1879475: Update status reasons to be more explicit

huffmanca:
This PR updates the status reasons in progressing scenarios to be a bit more explicit. This should reduce the number of "AsExpected" reasons that we see.


huffmanca:
/assign sttts

This one should be ready for your approval.

PR #903, opened 2020-09-29
Return key slice from QueueKeysFunc

qiujian16:


pkg/controller/factory/factory.go
```diff
@@ -119,22 +124,54 @@ func (f *Factory) WithBareInformers(informers ...Informer) *Factory {
 // Pass the queueKeyFn you want to use to transform the informer runtime.Object into string key used by work queue.
 func (f *Factory) WithInformersQueueKeyFunc(queueKeyFn ObjectQueueKeyFunc, informers ...Informer) *Factory {
 	f.informerQueueKeys = append(f.informerQueueKeys, informersWithQueueKey{
-		informers:  informers,
-		queueKeyFn: queueKeyFn,
+		informers: informers,
+		queueKeysFn: func(obj runtime.Object) []string {
```
pmorie, line 128: Do we want to change the existing method, or add a new one?

qiujian16: I think this one is not exposed. We will still maintain the existing method in factory.

pkg/controller/factory/controller_context.go
```diff
@@ -45,35 +45,47 @@ func (c syncContext) Recorder() events.Recorder {
...
+func (c syncContext) eventHandler(queueKeysFunc ObjectQueueKeysFunc, filter EventFilterFunc) cache.ResourceEventHandler {
 	resourceEventHandler := cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
 			runtimeObj, ok := obj.(runtime.Object)
 			if !ok {
 				utilruntime.HandleError(fmt.Errorf("added object %+v is not runtime Object", obj))
 				return
 			}
-			c.Queue().Add(queueKeyFunc(runtimeObj))
+			keys := queueKeysFunc(runtimeObj)
```
mfojtik, line 56:
```
for _, key := range queueKeyFunc(runtimeObj) {
  c.Queue().Add(key)
}
```

(you don't need keys variable)

pkg/controller/factory/factory.go
```diff
@@ -68,6 +68,11 @@ type PostStartHook func(ctx context.Context, syncContext SyncContext) error
 // triggers.
 type ObjectQueueKeyFunc func(runtime.Object) string
```
mfojtik, line 69: do we need this?

pkg/controller/factory/factory.go
mfojtik: I would refactor this to:

```
func WithInformersSingleQueueKeyFunc()
func WithInformersMultipleQueueKeysFunc()
```

pkg/controller/factory/factory.go
```diff
@@ -119,22 +124,54 @@ func (f *Factory) WithBareInformers(informers ...Informer) *Factory {
 // Pass the queueKeyFn you want to use to transform the informer runtime.Object into string key used by work queue.
 func (f *Factory) WithInformersQueueKeyFunc(queueKeyFn ObjectQueueKeyFunc, informers ...Informer) *Factory {
 	f.informerQueueKeys = append(f.informerQueueKeys, informersWithQueueKey{
-		informers:  informers,
-		queueKeyFn: queueKeyFn,
+		informers: informers,
+		queueKeysFn: func(obj runtime.Object) []string {
+			key := queueKeyFn(obj)
+			return []string{key}
```
mfojtik, line 130: just return `[]string{queueKeyFn(obj)}`

pkg/controller/factory/factory.go
```diff
@@ -119,22 +124,54 @@ func (f *Factory) WithBareInformers(informers ...Informer) *Factory {
...
 }
 
 // WithFilteredEventsInformersQueueKeyFunc is used to register event handlers and get the caches synchronized functions.
 // Pass informers you want to use to react to changes on resources. If informer event is observed, then the Sync() function
 // is called.
-// Pass the queueKeyFn you want to use to transform the informer runtime.Object into string key used by work queue.
+// Pass the queueKeysFn you want to use to transform the informer runtime.Object into string key used by work queue.
 // Pass filter to filter out events that should not trigger Sync() call.
 func (f *Factory) WithFilteredEventsInformersQueueKeyFunc(queueKeyFn ObjectQueueKeyFunc, filter EventFilterFunc, informers ...Informer) *Factory {
 	f.informerQueueKeys = append(f.informerQueueKeys, informersWithQueueKey{
```
mfojtik, line 154: s/informersWithQueueKey/informersWithQueueKeys/

pkg/controller/factory/factory.go
```diff
@@ -119,22 +124,54 @@ func (f *Factory) WithBareInformers(informers ...Informer) *Factory {
...
+	})
+	return f
+}
+
+// WithFilteredEventsInformersQueueKeysFunc is used to register event handlers and get the caches synchronized functions.
+// Pass informers you want to use to react to changes on resources. If informer event is observed, then the Sync() function
+// is called.
+// Pass the queueKeysFn you want to use to transform the informer runtime.Object into a string slice of keys used by work queue.
+// Pass filter to filter out events that should not trigger Sync() call.
+func (f *Factory) WithFilteredEventsInformersQueueKeysFunc(queueKeysFn ObjectQueueKeysFunc, filter EventFilterFunc, informers ...Informer) *Factory {
```
mfojtik, line 170: is there use-case for this?

mfojtik:
/hold

Please hold merging this until we after master == 4.7.0

pkg/controller/factory/factory.go
qiujian16: change to WithInformersMultipleQueueKeysFunc(), should we keep the name of WithInformersQueueKeyFunc for compatible?

pkg/controller/factory/controller_context.go
qiujian16: done

pkg/controller/factory/factory.go
qiujian16: it is still needed as the parameter of WithInformersQueueKeyFunc

pkg/controller/factory/factory.go
qiujian16: this is still to use a func that returns single key

pkg/controller/factory/factory.go
qiujian16: done

pkg/controller/factory/factory.go
qiujian16: currently no,  i will remove this func

PR #907, opened 2020-09-30
controller factory: add sync wrapper for unmanaged-state operators

stlaz:
cc @mfojtik @sttts


pkg/controller/factory/factory.go
```diff
@@ -268,3 +269,20 @@ func (f *Factory) ToController(name string, eventRecorder events.Recorder) Contr
 
 	return c
 }
+
+// WithHandleUnmanagedStateSync wraps a factory SyncFunc with a handler that no-ops
+// when the operator's ManagementState field is set to "Unmanaged"
+func WithHandleUnmanagedStateSync(operatorClient v1helpers.OperatorClient, delegate SyncFunc) SyncFunc {
+	return func(ctx context.Context, syncContext SyncContext) error {
+		spec, _, _, err := operatorClient.GetOperatorState()
```
sttts, line 277: is this cached?

stlaz: This is not a cached call

sttts: we already have that state in the wrapped sync func. Any way to access that?

stlaz: I don't think we'll always have the state in the wrapped `sync()` (unless I'm mistaken), but I think I could make a few changes that would allow using the `Factory.syncDegradedClient` (=> rename to `Factory.operatorClient`) and register handlers for different operator management states.

stlaz: ... but that still does not access the management state of the operator anywhere...

stlaz: I just noticed that the CAO actually does cache the state:
https://github.com/openshift/cluster-authentication-operator/blob/87b3524d807223f102cb45a07cc3de7c3a4c8947/pkg/operator/operatorclient.go#L34-L42 and perhaps the operators too:
https://github.com/openshift/cluster-openshift-apiserver-operator/blob/f3b2cade6e7d7fa7f4b9109b9fa075815ea3026d/vendor/github.com/openshift/library-go/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L59-L76
https://github.com/openshift/cluster-openshift-apiserver-operator/blob/f3b2cade6e7d7fa7f4b9109b9fa075815ea3026d/pkg/operator/operatorclient/operatorclient.go#L23-L30

My apologies, I really thought we haven't cached it. This adds a requirement to make sure the client's Informer is synced, though

stlaz:
This grew into something that looks a bit more useful than I originally intended, see if we might want what the PR offers now. I'll need to add unit tests if we'd like to adopt this, ofc.

PR #909, opened 2020-09-30
csi: add ${CLUSTER_ID} to placeholders

bertinatto:
Some CSI Drivers need to be given a cluster ID so that they can tag volumes and snapshots they create.

This is done so that the installer can filter those resources out and delete them on cluster deletion.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
@@ -114,6 +115,7 @@ func (c *CSIControllerSet) WithCSIDriverControllerService(
 	file string,
 	kubeClient kubernetes.Interface,
 	namespacedInformerFactory informers.SharedInformerFactory,
+	optionalConfigInformer configinformers.SharedInformerFactory,
 ) *CSIControllerSet {
 	manifestFile := assetFunc(file)
 	c.csiDriverControllerServiceController = csidrivercontrollerservicecontroller.NewCSIDriverControllerServiceController(
@@ -122,6 +124,7 @@ func (c *CSIControllerSet) WithCSIDriverControllerService(
 		c.operatorClient,
 		kubeClient,
 		namespacedInformerFactory.Apps().V1().Deployments(),
+		optionalConfigInformer,
 		c.eventRecorder,
 	)
 	return c
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
@@ -10,6 +10,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	opv1 "github.com/openshift/api/operator/v1"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
@@ -29,6 +30,8 @@ const (
 	resizerImageEnvName       = "RESIZER_IMAGE"
 	snapshotterImageEnvName   = "SNAPSHOTTER_IMAGE"
 	livenessProbeImageEnvName = "LIVENESS_PROBE_IMAGE"
+
+	infraConfigName = "cluster"
 )
 
 // CSIDriverControllerServiceController is a controller that deploys a CSI Controller Service to a given namespace.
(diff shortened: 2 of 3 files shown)


pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -29,6 +30,8 @@ const (
 	resizerImageEnvName       = "RESIZER_IMAGE"
 	snapshotterImageEnvName   = "SNAPSHOTTER_IMAGE"
 	livenessProbeImageEnvName = "LIVENESS_PROBE_IMAGE"
+
+	infraConfigName = "cluster"
```
sttts, line 34: over-abstraction

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -77,19 +87,28 @@ func NewCSIDriverControllerServiceController(
...
+		deployInformer:  deployInformer,
+		configInformers: configInformers,
 	}
 
-	return factory.New().WithInformers(
+	informers := []factory.Informer{
 		operatorClient.Informer(),
 		deployInformer.Informer(),
+	}
+	if c.configInformers != nil {
```
sttts, line 106: why should this ever be nil? Feels like unneeded complexity.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -225,6 +255,11 @@ func replacePlaceholders(manifest []byte, spec *opv1.OperatorSpec) []byte {
 		pairs = append(pairs, []string{"${LIVENESS_PROBE_IMAGE}", livenessProbe}...)
 	}
 
+	// Cluster ID
+	if clusterID != "" {
+		pairs = append(pairs, []string{"${CLUSTER_ID}", clusterID}...)
```
sttts, line 260: leaving `${CLUSTER_ID}` in there is odd.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Some CSI drivers (operands) don't have the option to pass in a cluster ID, so their operators will let this to be nil

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -119,7 +138,18 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
...
 	}
 
-	manifest := replacePlaceholders(c.manifest, opSpec)
+	var clusterID string
+	if c.configInformers != nil {
+		infra, err := c.configInformers.Config().V1().Infrastructures().Lister().Get(infraConfigName)
+		if err != nil {
+			return err
+		}
+		if infra.Status.PlatformStatus != nil {
```
sttts, line 147: this is correct? InfraName depends on another field? This at least needs a comment here. Looks like a bug now.

bertinatto: Yup, we've seen this on baremetal. Added a comment

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Removed the condition

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Sorry, do you mind elaborating?

Not sure if you're talking about this, but I've seen this pattern (const with singleton CR name) used  across most controllers in library-go.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Also, the docstring above tries to explain that

sttts: call it optionalConfigInformer. The current pattern is a recipe to null pointer exceptions the next time somebody modifies this code and is not aware of this invariant.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
sttts: this does not answer my question when reading this code. Why does `PlatformStatus != nil` influence why we care about cluster id?

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Renamed to `optionalConfigInformer`

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -114,6 +115,7 @@ func (c *CSIControllerSet) WithCSIDriverControllerService(
 	file string,
 	kubeClient kubernetes.Interface,
 	namespacedInformerFactory informers.SharedInformerFactory,
+	configInformers configinformers.SharedInformerFactory,
```
sttts, line 118: this is optional too?

sttts:
/approve
lgtm.

@jsafrane please take a final look.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: Sorry, I fat-fingered this.

Now that I look at it, this check doesn't make sense. I can just use `.Status.InfrastructureName` directly.

PR #910, opened 2020-10-05
verifier: expose Interface and hide implementation details

soltysh:
The code currently wrongly uses implementation types rather than interfaces, this code fixes that by exposing an interface and ensuring all the types implement it properly. 

/assign @wking @jottofar

### DIFF
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/verify/store"
+	"github.com/openshift/library-go/pkg/verify/store/serial"
 	"github.com/openshift/library-go/pkg/verify/util"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/verify/verify.go
```diff
@@ -48,10 +63,10 @@ const maxSignatureSearch = 10
 // validReleaseDigest is a verification rule to filter clearly invalid digests.
 var validReleaseDigest = regexp.MustCompile(`^[a-zA-Z0-9:]+$`)
 
-// ReleaseVerifier implements a signature intersection operation on a provided release
+// releaseVerifier implements a signature intersection operation on a provided release
 // digest - all verifiers must have at least one valid signature attesting the release
 // digest. If any failure occurs the caller should assume the content is unverified.
-type ReleaseVerifier struct {
+type releaseVerifier struct {
```
jottofar, line 69: CVO, which will be reusing this package as well, needs access to field Store. Prior to any of my changes or refactoring to library-go ReleaseVerifier was also exported.

soltysh: If we want to re-use the interface, I'd suggest adding `Store` method to interface, directly accessing struct fields won't help here. And the current approach is causing nil-pointer exception when used in oc.

pkg/verify/verify.go
```diff
@@ -200,8 +222,13 @@ func (v *ReleaseVerifier) Signatures() map[string][][]byte {
 	return copied
 }
 
+// AddStore adds additional stores for signature verification.
+func (v *releaseVerifier) AddStore(additionalStore store.Store) {
+	v.store = &serial.Store{Stores: []store.Store{additionalStore, v.store}}
```
wking, line 227: do we always want to add with a serial prepend?  I think we might want a getter/setter method pair so folks can pick the combination on the caller-side.

jottofar: This was just lifted exactly from CVO's current setting of 'verifier.Store` but also fine to go with the more flexible approach as well.

soltysh: That's the only usage I've noticed. I'd suggest you guys have a closer look at that package implementation and reconsider the overall shape, imo. For now I'd go with as is, and I'd definitely discourage getters & setters.

jottofar: As discussed with @wking, this will satisfy CVO's needs now and we can change it later if necessary.

PR #911, opened 2020-10-05
pkg/route: Add a package with route helpers

wking:
Constructing a usable URI for a route is non-trivial code that is useful in many route consumers.  Save consumers a few lines by creating a helper in a shared location for centralized maintenance.

Owners from [`router/OWNERS`][1].

CC @Miciah

[1]: https://github.com/openshift/router/blob/189bd7fd5594e09e85b9af178738ae3550a39995/OWNERS

### DIFF
--- a/pkg/route/OWNERS
+++ b/pkg/route/OWNERS
@@ -1,0 +1,14 @@
+approvers:
+  - frobware
+  - knobunc
+  - Miciah
+  - sgreene570
+  - smarterclayton
+reviewers:
+  - danehans
+  - frobware
+  - knobunc
+  - Miciah
+  - sgreene570
+  - smarterclayton
+component: Routing
--- a/pkg/route/routeapihelpers/routeapihelpers.go
+++ b/pkg/route/routeapihelpers/routeapihelpers.go
@@ -1,0 +1,44 @@
+// Package routeapihelpers contains utilities for handling OpenShift route objects.
+package routeapihelpers
+
+import (
+	"fmt"
+	"net/url"
+
+	routev1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// IngressURI calculates an admitted ingress URI.
+// If 'host' is nonempty, only the ingress for that host is considered.
+// If 'host' is empty, the first admitted ingress is used.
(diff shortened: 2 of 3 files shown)


pkg/route/route.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"net/url"
+
+	routev1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// IngressURI calculates an admitted ingress URI.
+// If 'host' is set, only the ingress for that host is considered.
+// If 'host' is unset, the first admitted ingress is used.
```
Miciah, line 14: Instead of "set" and "unset", maybe better to say "nonempty" and "empty."

wking:
Some potential consumers:

* [installer][1], which currently assumes HTTPS and the spec host.
* [auth operator's `RouteHasCanonicalHost`][2], which becomes `_, _, err := IngressURI(route, host); err != nil`
* openshift/cincinnati-operator#66

[1]: https://github.com/openshift/installer/blob/b2199d8ee9c6611476f59ce6deb08f96a8bd5bae/cmd/openshift-install/create.go#L447
[2]: https://github.com/openshift/cluster-authentication-operator/blob/adbb442cc7bbcb6a71f6118729d994a4f4dc8985/pkg/controllers/common/route.go#L151

Miciah (commented):
It might make sense to add a test case with `Admitted=False`.

wking:
> It might make sense to add a test case with `Admitted=False`.

Done with `explicitly non-admitted ingress, host-agnostic` in 08e24b164 -> c125aa6f7.

pkg/route/route.go
wking: > Instead of "set" and "unset", maybe better to say "nonempty" and "empty."

Done in 08e24b164 -> c125aa6f7.

deads2k:
I'm generally wary of "bucket of helpers", but I'm willing to see where it goes.  The last three packages I saw like this took nearly six months to untangle enough to separate.

I would like to see `pkg/route/routeapihelpers` in case a different category of route related code arrives.

/approve
/hold

holding for package update.  Feel free to merge after.

PR #913, opened 2020-10-07
staticpod: improve operator messages

mfojtik:
* Do proper pluralization in messages
* Show total node count (so we can determine if all expected control plane nodes are there)
* Unify on terminology ("achieve" vs. "is on")
* Add messages to unit test to make sure we have them correct

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
 	}
 }
 
+type revisionDescriptionPrinter struct {
+	nodes          []operatorv1.NodeStatus
+	latestRevision int32
+	progressing    bool
+}
+
+func newRevisionDescriptionPrinter(nodes []operatorv1.NodeStatus, latestRevision int32, progressing bool) *revisionDescriptionPrinter {
+	return &revisionDescriptionPrinter{nodes: nodes, latestRevision: latestRevision, progressing: progressing}
+}
+
+func pluralizedNode(i int) string {
+	if i < 2 {
+		return "node"
+	}
+	return "nodes"
+}
+
+func (r revisionDescriptionPrinter) String() string {
+	currentRevisions := sets.NewInt32()
+	latestCount := 0
+	availableCount := 0
+	for i := range r.nodes {
+		c := r.nodes[i].CurrentRevision
+		currentRevisions.Insert(c)
+		if c == r.latestRevision {
+			latestCount++
+		}
+		if c != 0 {
+			availableCount++
+		}
+	}
+
+	notAvailable := fmt.Sprintf("%d %s are not available and ", availableCount, pluralizedNode(availableCount))
+	if availableCount == len(r.nodes) {
+		notAvailable = ""
+	}
+	if availableCount == 0 {
+		return "none of the nodes are available"
+	}
+
+	if !currentRevisions.Has(r.latestRevision) {
+		if r.progressing {
+			return fmt.Sprintf("%snodes progressing towards revision %d", notAvailable, r.latestRevision)
+		}
+		return fmt.Sprintf("%snone of the nodes reached latest available revision %d", notAvailable, r.latestRevision)
+	}
+	if currentRevisions.Equal(sets.NewInt32(r.latestRevision)) {
+		return fmt.Sprintf("%sall nodes reached revision %d", notAvailable, r.latestRevision)
+	}
+	if currentRevisions.Has(r.latestRevision) {
+		oldRevisionStringList := []string{}
+		for _, i := range currentRevisions.Difference(sets.NewInt32(r.latestRevision)).List() {
+			oldRevisionStringList = append(oldRevisionStringList, fmt.Sprintf("%d", i))
+		}
+		return fmt.Sprintf("%ssome nodes are on old revisions (%s), %d %s reached latest available revision %d", notAvailable, strings.Join(oldRevisionStringList, ","), latestCount, pluralizedNode(latestCount), r.latestRevision)
+	}
+	return ""
+}
+
 // setAvailableProgressingConditions sets the Available and Progressing conditions
 func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1.StaticPodOperatorStatus) error {
 	// Available means that we have at least one pod at the latest level
 	numAvailable := 0
 	numAtLatestRevision := 0
 	numProgressing := 0
+	numNodes := len(newStatus.NodeStatuses)
 	counts := map[int32]int{}
 	failingCount := map[int32]int{}
 	failing := map[int32][]string{}
+	var latestRevision int32
+
 	for _, currNodeStatus := range newStatus.NodeStatuses {
 		counts[currNodeStatus.CurrentRevision] = counts[currNodeStatus.CurrentRevision] + 1
 		if currNodeStatus.CurrentRevision != 0 {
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -452,27 +453,41 @@ func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1
 
 	revisionStrings := []string{}
 	for _, currentRevision := range Int32KeySet(counts).List() {
-		count := counts[currentRevision]
-		revisionStrings = append(revisionStrings, fmt.Sprintf("%d nodes are at revision %d", count, currentRevision))
+		if count := counts[currentRevision]; count == 1 {
+			revisionStrings = append(revisionStrings, fmt.Sprintf("1 node achieved revision %d", currentRevision))
```
sttts, line 457: is "achieved" the right word?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -452,27 +453,41 @@ func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1
...
+				revisionStrings = append(revisionStrings, fmt.Sprintf("all nodes achieved revision %d", currentRevision))
+			} else {
+				revisionStrings = append(revisionStrings, fmt.Sprintf("%d nodes achieved revision %d", count, currentRevision))
+			}
+		}
 	}
 	// if we are progressing and no nodes have achieved that level, we should indicate
 	if numProgressing > 0 && counts[newStatus.LatestAvailableRevision] == 0 {
-		revisionStrings = append(revisionStrings, fmt.Sprintf("%d nodes have achieved new revision %d", 0, newStatus.LatestAvailableRevision))
+		revisionStrings = append(revisionStrings, fmt.Sprintf("none of the nodes have achieved new revision %d", newStatus.LatestAvailableRevision))
```
sttts, line 468: all these strings are a lot of text. Shouldn't we rather try to be more brief?

> some nodes are on old revisions (34, 35), 1 is on latest.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -452,27 +453,41 @@ func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1
...
 	}
 	revisionDescription := strings.Join(revisionStrings, "; ")
 
 	if numAvailable > 0 {
+		message := fmt.Sprintf("%d nodes are active; %s", numAvailable, revisionDescription)
```
sttts, line 473: what does active mean?

mfojtik: good question, I think "active" means "available" which means "they achieved the desired levels and are running" ?

pkg/operator/staticpod/controller/installer/installer_controller.go
mfojtik: we should pick one and be consistent "are at" vs. "achieved" I don't care much which one, I liked "achieved" better.

pkg/operator/staticpod/controller/installer/installer_controller.go
mfojtik: yeah, I think it will read better than 3 semicolons

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
...
+	latestCount := 0
+	availableCount := 0
+	for i := range r.nodes {
+		c := r.nodes[i].CurrentRevision
+		currentRevisions.Insert(c)
+		if c == r.latestRevision {
+			latestCount++
+		}
+		if c != 0 {
+			availableCount++
```
sttts, line 453: why does this mean anything about availability?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
...
+	if availableCount == len(r.nodes) {
+		notAvailable = ""
+	}
+	if availableCount == 0 {
+		return "none of the nodes are available"
+	}
+
+	if !currentRevisions.Has(r.latestRevision) {
+		if r.progressing {
+			return fmt.Sprintf("%snodes progressing towards revision %d", notAvailable, r.latestRevision)
```
sttts, line 467: am confused about `notAvailable` (e.g. `none of the nodes are available`) to be plugged into the first `%s`.

mfojtik: i tried to incorporate that into message, so we don't lose it but also get rid of the `;` .... I think it will read as "none of the nodes are available and nodes progressing towards revision N"

pkg/operator/staticpod/controller/installer/installer_controller.go
mfojtik: in the existing code "numAvailable" is determined by CurrentRevision!=0 ... So i guess CurrentRevision==0 means the node is not available or not actively making progress? i'm a bit confused here as well.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
 	}
 }
 
+type revisionDescriptionPrinter struct {
```
sttts, line 425: doc

sttts: Why is this an object in the first place and not just a simple func?

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: this has nothing to do with availability. It only says that this revision was the last that got ready eventually.

mfojtik: `lastReadyRevision` ?

sttts: call it `nonZeroCount`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
...
+		currentRevisions.Insert(c)
+		if c == r.latestRevision {
+			latestCount++
+		}
+		if c != 0 {
+			availableCount++
+		}
+	}
+
+	notAvailable := fmt.Sprintf("%d %s are not available and ", availableCount, pluralizedNode(availableCount))
```
sttts, line 457: cannot follow. This must be `len(nodes)-availableCount`, no?

And better turn the wording into: `%d %s have never been available and`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
...
+	if availableCount == len(r.nodes) {
+		notAvailable = ""
+	}
+	if availableCount == 0 {
+		return "none of the nodes are available"
```
sttts, line 462: none of the nodes was ever available.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: `!currentRevisions.Has(r.latestRevision)` does not match message

sttts: oh. you mean "**all** nodes progressing ..."

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -422,15 +422,77 @@ func setNodeStatusFn(status *operatorv1.NodeStatus) v1helpers.UpdateStaticPodSta
...
+			return fmt.Sprintf("%snodes progressing towards revision %d", notAvailable, r.latestRevision)
+		}
+		return fmt.Sprintf("%snone of the nodes reached latest available revision %d", notAvailable, r.latestRevision)
+	}
+	if currentRevisions.Equal(sets.NewInt32(r.latestRevision)) {
+		return fmt.Sprintf("%sall nodes reached revision %d", notAvailable, r.latestRevision)
+	}
+	if currentRevisions.Has(r.latestRevision) {
+		oldRevisionStringList := []string{}
+		for _, i := range currentRevisions.Difference(sets.NewInt32(r.latestRevision)).List() {
```
sttts, line 476: loop over currentRevisions and exclude `latestRevision`. No need for set arithmetic.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: > lastReadyRevision

It's a count.

PR #915, opened 2020-10-09
Add support for kubevirt platform

ravidbro:
Signed-off-by: Ravid Brown <ravid@redhat.com>


pkg/operator/genericoperatorclient/dynamic_operator_client_test.go
```diff
@@ -33,7 +33,6 @@ func TestSetOperatorSpecFromUnstructured(t *testing.T) {
 					"non-standard-field":         "value",
 					"logLevel":                   "Trace",
 					"managementState":            "",
-					"operatorLogLevel":           "",
```
sttts, line 36: these changes look unrelated. If necessary, put into extra commit and describe why youneed them.

sttts:
Looks nearly fine, only https://github.com/openshift/library-go/pull/915#discussion_r503169668.

pkg/operator/genericoperatorclient/dynamic_operator_client_test.go
ravidbro: It is related and a must, the test fails without this change.
The new version of 'api' changes the behavior of the response and that what is expected from now.

sttts: Then move it into its own commit.

Also please don't mix vendor bump with other changes.

ravidbro: done

ravidbro: I reorganized the commits in this PR

PR #917, opened 2020-10-12
Check errors if we can't annotate a secret

Fedosin:
Now we ignore errors from setResourceMigrated function, which can lead to unpredictable behavior in runtime. To prevent this we add the standard error checking in the function.


sttts:
/lgtm
/approve

There is only a marshalling error. This will never happen. So I think this merely cosmetics.

PR #920, opened 2020-10-12
add UID to all static pods to trick kubelet into honoring grace period

deads2k:


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -320,9 +321,16 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 	// Write secrets, config maps and pod to disk
 	// This does not need timeout, instead we should fail hard when we are not able to write.
 
+	pod, err := resourceread.ReadPodV1([]byte(podContent))
+	if err != nil {
+		return err
+	}
+	pod.UID = uuid.NewUUID()
```
sttts, line 328: this needs a comments with the summary of the slack thread.

deads2k:
comment added, retagging

/cherrypick release-4.6

PR #922, opened 2020-10-13
don't wait on kube-apiserver available

deads2k:
if the APIs are present, the kube-apiserver is up enough.  We don't need the additional check.


deads2k:
compare to https://github.com/openshift/cluster-openshift-apiserver-operator/pull/401/files

PR #925, opened 2020-10-14
Change GetServiceCIDR() to GetServiceCIDRs() for dual-stack support

danwinship:
This has been the subject of FIXMEs for a while...

AFAIK this is used by cluster-kube-apiserver-operator and cluster-kube-controller-manager-operator, and both of their operands have been updated to be able to deal with multiple values.


squeed:
Is the PR already filed on kube-apiserver the only downstream consumer of this function?

PR #926, opened 2020-10-19
Set TCP_USER_TIMEOUT socket option

p0lyn0mial:
This PR sets the `TCP_USER_TIMEOUT` (https://man7.org/linux/man-pages/man7/tcp.7.html) socket option which controls for how long transmitted data may be unacknowledged before the connection is forcefully closed.

Without that option, we rely on the TCP stack to detect broken network connection. This can take up to 15 minutes. During that time our platform might be unavailable. 

There are already reported cases in which aggregated APIs (i.e. `openshift-apiserver`) were unable to establish a new connection to the Kube API for 15 minutes after "ungraceful termination" (https://bugzilla.redhat.com/show_bug.cgi?id=1881878 and https://bugzilla.redhat.com/show_bug.cgi?id=1879232#c39)

It looks like detecting broken connections on the application level is getting more traction and is preferable. Unfortunately, it is on the slow track and will require backporting to golang's std library. Until that time we would like to take advantage of `TCP_USER_TIMEOUT`

[1] https://go-review.googlesource.com/c/net/+/198040
[2] https://go-review.googlesource.com/c/net/+/236498#message-7bd657ac6960f0dc7acbbe28cbe3d80ac4f3a34b

### DIFF
--- a/pkg/config/client/client_config.go
+++ b/pkg/config/client/client_config.go
@@ -2,14 +2,12 @@ package client
 
 import (
 	"io/ioutil"
-	"net"
-	"net/http"
-	"time"
-
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"net/http"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/network"
 )
 
 // GetKubeConfigOrInClusterConfig loads in-cluster config if kubeConfigFile is empty or the file if not,
@@ -101,10 +99,8 @@ func (c ClientTransportOverrides) DefaultClientTransport(rt http.RoundTripper) h
 		return rt
 	}
 
-	transport.DialContext = (&net.Dialer{
-		Timeout:   30 * time.Second,
-		KeepAlive: 30 * time.Second,
-	}).DialContext
+	dialer := network.DefaultClientDialer()
+	transport.DialContext = dialer.DialContext
 
 	// Hold open more internal idle connections
 	transport.MaxIdleConnsPerHost = 100
--- a/pkg/config/client/client_config.go
+++ b/pkg/config/client/client_config.go
@@ -99,8 +99,7 @@ func (c ClientTransportOverrides) DefaultClientTransport(rt http.RoundTripper) h
 		return rt
 	}
 
-	dialer := network.DefaultClientDialer()
-	transport.DialContext = dialer.DialContext
+	transport.DialContext = network.DefaultClientDialContext()
 
 	// Hold open more internal idle connections
 	transport.MaxIdleConnsPerHost = 100
--- a/pkg/network/dialer_others.go
+++ b/pkg/network/dialer_others.go
@@ -1,0 +1,18 @@
+// +build !linux
+
+package network
+
+import (
+	"net"
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+func dialerWithDefaultOptions() *net.Dialer {
+	klog.V(2).Info("Creating the default network Dialer (unsupported platform). It may take up to 15 minutes to detect broken connections and establish a new one")
+	return &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+}
--- a/pkg/network/dialer_others.go
+++ b/pkg/network/dialer_others.go
@@ -9,10 +9,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func dialerWithDefaultOptions() *net.Dialer {
+func dialerWithDefaultOptions() DialContext {
 	klog.V(2).Info("Creating the default network Dialer (unsupported platform). It may take up to 15 minutes to detect broken connections and establish a new one")
-	return &net.Dialer{
+	nd := &net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
 	}
+	return nd.DialContext
 }
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,7 @@ require (
 	go.etcd.io/etcd v0.5.0-alpha.5.0.20200819165624-17cef6e3e9d5
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	golang.org/x/net v0.0.0-20200707034311-ab3426394381
+	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	gopkg.in/asn1-ber.v1 v1.0.0-20181015200546-f715ec2f112d // indirect
 	gopkg.in/ldap.v2 v2.5.1
--- a/pkg/network/dialer.go
+++ b/pkg/network/dialer.go
@@ -1,10 +1,13 @@
 package network
 
 import (
+	"context"
 	"net"
 )
 
-// DefaultClientDialer returns a network dialer with default options sets.
-func DefaultClientDialer() *net.Dialer {
+type DialContext func(ctx context.Context, network, address string) (net.Conn, error)
+
+// DefaultDialContext returns a DialContext function from a network dialer with default options sets.
+func DefaultClientDialContext() DialContext {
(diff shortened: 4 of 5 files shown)


pkg/config/client/client_config.go
```diff
@@ -101,10 +103,23 @@ func (c ClientTransportOverrides) DefaultClientTransport(rt http.RoundTripper) h
...
-	transport.DialContext = (&net.Dialer{
+	dialer := &net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
-	}).DialContext
+		Control: func(network, address string, c syscall.RawConn) error {
+			// Supported only on Linux
+			// TODO: provide code that will be platform depended
+			err := c.Control(func(fd uintptr) {
+				err := syscall.SetsockoptInt(int(fd), syscall.IPPROTO_TCP, /*TCP_USER_TIMEOUT taken from https://godoc.org/golang.org/x/sys/unix*/ 0x12, int(10 * time.Second/time.Millisecond))
```
p0lyn0mial, line 113: not sure what's the best value here

pkg/config/client/client_config.go
```diff
@@ -101,10 +103,23 @@ func (c ClientTransportOverrides) DefaultClientTransport(rt http.RoundTripper) h
 		return rt
 	}
 
-	transport.DialContext = (&net.Dialer{
+	dialer := &net.Dialer{
 		Timeout:   30 * time.Second,
 		KeepAlive: 30 * time.Second,
-	}).DialContext
+		Control: func(network, address string, c syscall.RawConn) error {
+			// Supported only on Linux
```
p0lyn0mial, line 110: do we have to detect connections to the local host?

pkg/config/client/client_config.go
squeed: According to https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/, this option will short-circuit keepalive TCP_KEEPCNT * TCP_KEEPINTVL. Given that client-go sets the KEEPINTVL to 30 seconds, this means that we will send exactly one keepalive probe before this kicks in and we kill the connection.

That's too tight - a single dropped packet while waiting for a large result will drop the connection.

If we would like faster detection of downed peers, we should
* set TCP_KEEPINTVL - the interval at which probes are sent - to something smaller, like 5 seconds
* set TCP_KEEPIDLE - the threshold for sending the first probe - to something very small, like 2 seconds
* set TCP_USER_TIMEOUT to allow for a few dropped probes - 25 seconds or so.

squeed: the 30 seconds comes from here: https://github.com/kubernetes/client-go/blob/master/transport/cache.go#L94

squeed:
Given that I don't think many people will be accessing the apiserver over sattelite internet, these timing parameters seem just fine. However, we need to verify that they do what we want, since these knobs are non-orthogonal and never obvious.

The easiest way I can think of to test this is
1. spin up a cluster with this PR
2. Start a pod, start a tcpdump on a master with all traffic from / to that pod ip
3. issue a watch for something that has no updates
4. verify that tcp keepalives do what we expect
5. using `iptables`, block access to that pod from the apiserver
6. verify that the connection is closed.

pkg/network/dialer_unix.go
```diff
@@ -0,0 +1,74 @@
...
+	"syscall"
+	"time"
+
+	"golang.org/x/sys/unix"
+)
+
+func dialerWithDefaultOptions() *net.Dialer {
+	return &net.Dialer{
+		// TCP_USER_TIMEOUT does affect the behaviour of connect() which is controlled by this field so we set it to the same value
+		Timeout: 25 * time.Second,
```
p0lyn0mial, line 16: does it make sense?

pkg/config/client/client_config.go
p0lyn0mial: sgtm, thx, updated.

pkg/network/dialer_unix.go
squeed: Timeout, in this case, is timeout to establish the TCP session. 25 seconds might be gigantic, if this is exclusively intra-cluster traffic.

p0lyn0mial: yes, this function is primarily used by the operators (https://github.com/search?q=org%3Aopenshift+GetKubeConfigOrInClusterConfig&type=code).

Setting it to `5s` would be okay?

squeed: actually, now that I think about it, 20 is probably about right. In a cluster that is thrashing but making progress, setting too low a timeout just leads to additional resource exhaustion. As always, it's a balancing act between keeping lightly-loaded clusters performant vs. tolerating heavliy-loaded ones.

I wish we had better numbers to make this decision. Let me see if we have some metrics.

pkg/config/client/client_config.go
```diff
@@ -1,13 +1,11 @@
 package client
 
 import (
+	"github.com/openshift/library-go/pkg/network"
```
sttts, line 4: order

pkg/network/dialer_others.go
```diff
@@ -0,0 +1,18 @@
...
+	"time"
+
+	"k8s.io/klog/v2"
+)
+
+func dialerWithDefaultOptions() *net.Dialer {
+	klog.V(2).Info("Creating the default network Dialer (unsupported platform). It may take up to 15 minutes to detect broken connections and establish a new one")
+	return &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
```
sttts, line 16: keepalive in linux is gone?

p0lyn0mial: no, `KeepAlive` setts both `TCP_KEEPINTVL` and `TCP_KEEPIDLE` to the same value. Since we want distinct values we are now setting them in `setDefaultSocketOptions` function

p0lyn0mial:
/hold

for testing

p0lyn0mial:
re e8be1278bbea83f89b05eaa47c4c874b5d18b77d commit

it turned out that `TCP_KEEPINTVL` and `TCP_KEEPIDLE` weren't set because they got overwritten after calling `DialContext ` method, here https://github.com/golang/go/blob/master/src/net/dial.go#L431

p0lyn0mial:
I did test it on an idle connection and it looks like it actually works.

I created [a simple app](https://github.com/p0lyn0mial/simple-watch/blob/1883212e485ab37dfedbd4df2051d94cf82c8b87/main.go) that establishes a watch on never changing `test01/secrets` resources to the Kube API server and simply dropped network traffic originating from the app.

Withouth the patch, the connection is terminated after ~`270s=30s * 9` , here is[ a tcpdump](https://github.com/p0lyn0mial/simple-watch/blob/patch/watch-ka-before-patch.pcap)

Screenshot 2020-10-28 at 12 56 34


With the patch, the connection is terminated after ~`25s`, after dropping the traffic approximately `4` keep alive probes were sent every `5s`, here is [a tcpdump](https://github.com/p0lyn0mial/simple-watch/blob/patch/watch-ka-after-patch.pcap) and the [modified app](https://github.com/p0lyn0mial/simple-watch/blob/7e93966642e67b62c84d4eab1da9ab3855840a22/main.go)

Screenshot 2020-10-28 at 12 38 09

p0lyn0mial:
I did test it on an active connection and it looks like it actually works.

I created [a simple app](https://github.com/p0lyn0mial/simple-get/blob/main/main.go) that sends `3 req` per second to list `test01/secrets` resources from the Kube API server and simply dropped network traffic originating from the app.

Withouth the patch, the connection is terminated after `~16m`, here is[ a tcpdump](https://github.com/p0lyn0mial/simple-get/blob/patch/get-ka-before-patch.pcap)

Screenshot 2020-10-29 at 12 11 41

With the patch, the connection is terminated after `~25s`, here is [a tcpdump ](https://github.com/p0lyn0mial/simple-get/blob/patch/get-ka-after-patch.pcap) and [the modified app](https://github.com/p0lyn0mial/simple-get/blob/patch/main.go)

Screenshot 2020-10-29 at 11 42 48

squeed:
post-merge /lgtm - this is awesome!

PR #927, opened 2020-10-20
add a check for an unhandled error in ApplyServiceMonitor

dhellmann:
If ApplyServiceMonitor fails to Get() an existing resource because it
is not there, it correctly handles the error and creates the
resource. However, if Get() fails for some other reason the error was
left unhandled, causing a panic later when a pointer is
uninitialized. This change handles the error to avoid the panic.

See the machine-api-operator logs in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_machine-api-operator/671/pull-ci-openshift-machine-api-operator-master-e2e-metal-ipi/1318544074000568320 for an example of the panic. I suspect it's caused by not having RBAC set up for ServiceMonitor.


dhellmann:
@deads2k this is a pretty small bug fix to handle a nil pointer error that I encountered. Could you put it on your review queue, please?

PR #928, opened 2020-10-21
fix image reference parsing when tag includes a `.`

ecordell:
this was incorrectly detecting images without a registry, but with a
tag that included a '.' as a registry-only image reference, instead
of a name+tag image reference.

The corrects an issue introduced in 670e819


ecordell:
@bparees fixed a gofmt issue

PR #933, opened 2020-10-26
collapse static pod bytes

deads2k:
We write the static pod yaml separately from the the backup static pod.yaml. This creates one set of bytes for both.  Still written twice though.

/cherrypick release-4.6


mfojtik:
@shodan-bot bump openshift/cluster-kube-apiserver-operator

PR #939, opened 2020-11-06
registry client: Add alternative sources methods to gather sources from ImageContentSourcePolicies

sallyom:
Allow for multiple image source options when retrieving repository (from icsp, from mirrored)

There may be multiple image source options (from icsp, mirrored) rather than a single image source.
This PR adds registry client Context interface and methods to gather ICSP sources for a given Context.

/cc @smarterclayton 
/cc @soltysh 

fake bump here: https://github.com/openshift/oc/pull/439


soltysh:
/assign @dmage 
who is also looking at the other PR

dmage (changes requested):
No, RepositoryList cannot exist.

To create an instance of `distribution.Repository`, the client sends an HTTP request to the /v2/ endpoint. If the first mirror is available and has all necessary images, the client shouldn't try to access other mirrors nor the original repository. It means `distribution.Repository` for the second mirror shouldn't be created until the client has checked the first mirror.

On disconnected clusters firewall can drop requests to external registries, i.e. attempt to connect to an external registry can take few minutes. So this extra action may be very time consuming.

sallyom:
@dmage, I've updated this PR, ptal. Is there a reason to return a repository that won't be accessible? This simply takes the list (only a list when dealing with a Context that has >1 Context.AlternativeSources, so only when the caller has populated that - like with oc, when dealing with ImageContentSourcePolicies) and returns the one repository that is appropriate for a given *registryclient.Context.

pkg/image/registryclient/client.go
```diff
@@ -102,6 +104,7 @@ func (c *Context) Copy() *Context {
 		Limiter:           c.Limiter,
 
 		DisableDigestVerification: c.DisableDigestVerification,
+		AlternativeImageSources:   c.AlternativeImageSources,
```
smarterclayton, line 107: This needs to be an interface, not a string array.  By doing a string array, you are forcing the caller of the context to up front know what the alternatives for.  Instead, you should accept an interface like:

```
type AlternateImageSource interface {
  AlternateImageSource(registry *url.URL, repoName string, tag, digest string) ([]imagereference.Reference, error)
  ...
}
```

that provides you the list of alternates.  Then you have to calculate that at the time you hit the error, not in the `Repository()` call.

sallyom: updated - moved the alternative sources gathering here, to library-go, as a method of Context.  The Alternative Sources are now only a field in Context here, not in oc SecurityOptions. 
To get alternative sources (populate Context.AlternativeSources), caller has to provide a list of ImageContentSourcePolicies and an imageRef.

sallyom: unit tests exercising ^^ are currently in the oc PR

sallyom: I added the interface, ptal

pkg/image/registryclient/client.go
```diff
@@ -199,7 +209,40 @@ func (c *Context) Ping(ctx context.Context, registry *url.URL, insecure bool) (h
...
+		}
+		repoList = append(repoList, repo)
+	}
+	for _, r := range repoList {
+		manifests, err = repo.Manifests(context.TODO())
+		if err != nil {
+			err = fmt.Errorf("unable to get local manifest service: %v", err)
+			continue
+		}
+		return r, manifests, nil
```
dmage, line 238: @sallyom sorry, I don't understand this PR at all. Let's focus on this code.

Creation a new network connection is expensive. Especially for disconnected clusters that drop packages to the external world. Let's imagine that you've mirrored `192.0.2.1/busybox` to `quay.io/prometheus/busybox` and you want to pull `192.0.2.1/busybox@sha256:64db0ecb0d989773dea64212d366e618c29a5552e3e386ac1dd603bf770f484e`. `192.0.2.1` is a dead address and normally is not available (https://tools.ietf.org/html/rfc5735, TEST-NET-1). `addPossibleRepository` will take few minutes for this address. Disconnected clusters see `quay.io` in a similar way: all traffic might be silently dropped and in this case `addPossibleRepository` will take few minutes before it returns an error.

Note that it's perfectly valid and expected that `quay.io` is not available on disconnected clusters.

And once you've created all these repositories, you take the first one, call `Manifests` (it will always succeed in our case), and return it. Everything else is thrown away.

dmage: *upd:* `192.0.2.1` is a dead address, but `quay.io` is available. Therefore, if you have ICSP like

```
- mirrors:
  - quay.io/prometheus/busybox
  source: 192.0.2.1/busybox
```

pulling from `192.0.2.1/busybox` should work without any delays.

sallyom: I don't understand this scenario.  My use case for the change, and addressing https://bugzilla.redhat.com/show_bug.cgi?id=1823143 , is this:
* A release is mirrored from quay to a private disconnected registry
* A user wants to extract from that mirrored release

Currently this is not possible, because ICSPs are not considered, so _only_ the quay image is tried.  
This change adds mirrors from ICSP and returns the repository that is accessible given the Context.

smarterclayton: Part of this is a tradeoff - the client has different resiliency needs than a cluster.  If we *always* check ICSP in the client, then *every* image CLI command is slower.  I think unlike a cluster where we assume that the ICSP load is free (it's distributed up front to every node for resiliency), in the client we're biasing towards latency, so we want the client to avoid having to do the lazy fetch of ICSP unless it has to.  I think it's the right tradeoff.

soltysh: If we're making a client-specific tradeoff, I think it's best to have that functionality only in oc, since library-go is so widely used.

dmage: @smarterclayton I see several options:

1. clients try to use ICSP only when they are explicitly asked to do it (via `--icsp-file` or `--use-cluster-icsp`), and they follow normal ICSP flow (mirrors first, then the original repo)
2. even if `--icsp-file` is provided, ICSP is loaded only after the initial attempt to access the original repo has failed (either from the cluster or from the file)
3. similar to the 2nd approach, but order of requests depends on the used flag (it's cheap to load ICSP from a file, it's more expensive to load it from a cluster)

You prefer the second option, right?

And no matter what option we choose, we shouldn't ping all registries in this method. We should try to use them one by one.

sallyom: we settled on this awhile ago:
* if --icsp-file is set, then decode/use that - don't look further
* image reference (this is the original image from original registry/repo) 
* if ICSP found in cluster, add that to list of image source options, those will be:
    * icsp from cluster
    * user-passed release image (substitute registry/repo/name from user-given w/ same digest)

My preference is that oc just works, it will look for image sources in that order without user specifying anything (except the --icsp-file)

sallyom: wrt @soltysh question, I can implement the RepositoryRetriever and keep these changes within oc, just have to add the RepositoryRetriever interface to Context and export a few methods.

sallyom: I've added the RepositoryRetriever interface to the registry client Context, see [the oc PR](https://github.com/openshift/oc/pull/439) for implementation.

dmage: I'll try to explain what this code is doing again (it's in a different repo now, but changes in this repo are needed for it).

```go
	for _, ics := range c.ImageSources {
		repo, err = c.addPossibleRepository(ctx, ics.RegistryURL(), ics.RepositoryName(), insecure)
		if err != nil {
			return nil, nil, err
		}
		repoList = append(repoList, repo)
	}
```

For every ImageSource it'll send a GET request to something like `https://registry.example.com/v2/`. It may also try to send a request to `http://...` if the first request has failed and `insecure` is set. If the registry is not available, it will return an error.

I.e. for each ImageSource will have an instance of `Repository`. If at least one image source is not available, it'll be a fatal error.

```go
	for _, r := range repoList {
		manifests, err = repo.Manifests(context.TODO())
		if err != nil {
			err = fmt.Errorf("unable to get local manifest service: %v", err)
			continue
		}
		return r, manifests, nil
	}
```

`repoList` contains at least one element at this point, repo.Manifests [cannot fail](https://github.com/docker/distribution/blob/f4506b517a7e9e64cfdcc73c991d9e85b784dc3e/registry/client/repository.go#L177-L185), so the first element will be always returned. Other repoList elements will be thrown away.

ICSPs are created to allow clients to pull images when the source repository is not accessible. ICSPs can contain multiple mirrors so that if a single mirror is not available, the image is still pullable. This code assumes that all sources including all mirrors are available and this is against the idea of mirroring.

sallyom: Right, this should break when it returns a first repo. I've updated that in the oc PR, thank you. repoList is not necessary there, only ever dealing w/ a single repo 
```
	for _, ics := range c.ImageSources {
		repo, err = c.addPossibleRepository(ctx, ics.RegistryURL(), ics.RepositoryName(), insecure)
		if err != nil {
			return nil, nil, err
		}
		break
	}
	manifests, err = repo.Manifests(context.TODO())
	if err != nil {
		return nil, nil, fmt.Errorf("unable to get local manifest service: %v", err)
	}
	return repo, manifests, nil
```

@smarterclayton @dmage other than that, ok w/ this  PR?

dmage: No, as I can see, `addPossibleRepository` is a copy of `(*registryclient.Context).Repository`. Instead of creating a copy and making helpers public, you can call the original function from`(c *registryContext).Repository` using `c.Context.Repository`. This function shouldn't return a ManifestService as the caller can do it itself. So these changes aren't necessary.

dmage: The snippet isn't ok either, there is no guarantee that the first repository has necessary images. OpenShift uses ICSPs to generate registries.conf on nodes, from [containers-registries.conf(5)](https://github.com/containers/image/blob/master/docs/containers-registries.conf.5.md):

> The mirrors are attempted in the specified order; the first one that can be contacted and contains the image will be used (and if none of the mirrors contains the image, the primary location specified by the `registry.location` field, or using the unmodified user-specified reference, is tried last).

From [ICSP docs](https://github.com/openshift/api/blob/31aac52f4998f11617144a0d917c8ea1d2a1a9fc/operator/v1alpha1/types_image_content_source_policy.go#L59-L66):

> mirrors is one or more repositories that _may_ also contain the same images.

We've agreed that `oc` can use a different order, but it doesn't mean that `oc` can assume that all mirrors contain the same set of images.

sallyom: > No, as I can see, `addPossibleRepository` is a copy of `(*registryclient.Context).Repository`. Instead of creating a copy and making helpers public, you can call the original function from`(c *registryContext).Repository` using `c.Context.Repository`. This function shouldn't return a ManifestService as the caller can do it itself. So these changes aren't necessary.

I see that addPossibleRepository/making helpers public is not necessary, I'll update that now. Why not return ManifestService, since 99% of the time that's what we're after when looking for an appropriate Repository? The caller can easily call that but the duplicate code adds up so i don't see why we shouldn't include it.

sallyom: the manifestService returned guarantees the repo contains the image - I feel we're going in circles here.
If you don't agree we should return the ManifestService from the Repository fn, then I'll add that check in oc, in oc's RepositoryRetriever implementation.  However, I'll have to check for an error and throw out the manifestService returned from Repository(bc it won't return the ManifestService) but then I'll have to grab it elsewhere, as everywhere (almost) oc calls Repository, oc wants the ManifestService.  Is there a different way to ensure the Repository returned contains the image we're passing to find the Repository? If so  please let me know.

smarterclayton:
Some high level comments:

> If we're making a client-specific tradeoff, I think it's best to have that functionality only in oc, since library-go is so widely used.

This is not desirable - it needs to be part of library-go and the general client implementation because the behavior of ICSP is general to all OpenShift clusters, and ICSP is useful to both OpenShift specific commands (oc adm release) and generic registry commands (oc image), and in general no oc command really cares about the details or should have to care.

Oleg raised the point about explicit request to use cluster ICSP vs implicit - and that's a good distinction to make.  `oc image` does not talk to a cluster by default, and should not, because it's generic to registries.  A user *may* want to tell the command to use a cluster ICSP, but that has to be opt in, so it's reasonable to support `--use-cluster-icsp` on the `oc image *` commands.  Conversely, `oc adm release *` should be able to honor the cluster ICSP implicitly (because a user can run `oc adm release extract --tools` and expects the that to succeed regardless).  However, for a lot of frequent interactions with the server the contract of `oc adm release *` in general needs to be fast and looking up the cluster ICSP is not appropriate by default.  In a node context ICSP is opinionated and tries to access the exact list order to allow admins to optimize - but when implicitly retrieving a cluster ICSP we can't say for certain that's the best outcome for the user.  In specific commands we may very well do so - such as `oc adm release mirror` or `oc adm release new` where the ICSP likely has the most performant entry first - because that benefits the admin on an expensive call.  However, another goal of ICSP is to improve resiliency, so we have to keep consulting the list of repositories even on subsequent calls to the same distribution.Repository object (i.e. the shim we provide has to have some dynamism).

pkg/image/registryclient/client.go
sallyom: met w/ @smarterclayton i have a different path forward to allow specific ordering of retrieving Manifests from a list of possible if/when that is appropriate (as opposed to lazy lookup/return otherwise) - i'll push for review once i sort out details

sallyom:
I've added the `--lookup-cluster-icsp` flag in the oc pr, and have added the alternative image source awareness to all 
`oc image` commands.
There isn't much to change here in library-go, I added a method to return the manifests while keeping the other less useful method for returning a Repository only. Also added the interface with the `PreferredImageSource` method, implemented in the oc PR.

pkg/image/registryclient/client.go
dmage: @sallyom 

> the manifestService returned guarantees the repo contains the image

I gave you this link before, please take a look at the implementation of repo.Manifests, it cannot fail: https://github.com/docker/distribution/blob/f4506b517a7e9e64cfdcc73c991d9e85b784dc3e/registry/client/repository.go#L177-L185

ManifestService will always be created, its creation doesn't guarantee anything at all.

sallyom: I see, got it, the caller will still verify, then. It still makes sense to return the ManifestService with the Repository, IMO, since it saves a lot of duplicate code.

pkg/image/registryclient/alternative-sources.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// AddICSPsFromCluster will lookup ICSPs from cluster. Unless explicit lookup is set to false, logs errors rather than returning errors.
+func (a *Context) AddICSPsFromCluster(explicit bool) error {
+	icspClient, err := a.ImageSources.ICSPClientFn()
+	if err != nil {
+		return err
+	}
+	if explicit && icspClient == nil {
+		return fmt.Errorf("flag --lookup-cluster-icsp was set to true, but no method was set to find ImageContentSourcePolicy objects in cluster")
```
soltysh, line 37: You don't have flags set here, complain that the client is nil here.

pkg/image/registryclient/alternative-sources.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if icspClient != nil {
```
soltysh, line 39: The above condition shouldn't matter on `explicit` and then this check is not needed.

soltysh (changes requested):
Also seem my comments in https://github.com/openshift/oc/pull/439

pkg/image/registryclient/alternative-sources.go
sallyom: right, updated, fixed, and removed flag reference

pkg/image/registryclient/alternative-sources.go
sallyom: updated

pkg/image/registryclient/alternative-sources.go
sallyom: I need the check for icspClient != nil when !explicit, bc it will be nil whenever user is not connected to a cluster and in that case will get a `nil pointer dereference` when trying to list icsps

pkg/image/registryclient/client.go
sallyom: removed the interface in favor of adding all functions to library-go, as suggested

soltysh (commented):
One more nit and you're good to go

pkg/image/registryclient/alternative-sources.go
```diff
@@ -0,0 +1,221 @@
+package registryclient
```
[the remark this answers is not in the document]
sallyom: good to know, fixing, thanks

pkg/image/registryclient/client.go
```diff
@@ -223,7 +224,12 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
 	if limiter == nil {
 		limiter = rate.NewLimiter(rate.Limit(5), 5)
 	}
-	return NewLimitedRetryRepository(repo, c.Retries, limiter), nil
+	repo = NewLimitedRetryRepository(repo, c.Retries, limiter)
+	manifests, err := repo.Manifests(ctx)
```
smarterclayton, line 228: I'm confused why this was added?  Manifests isn't required to be returned up front, it is a lazy initialization.

sallyom: right, I can remove it, i left it because it saves a lot of code elsewhere

sallyom: updated

pkg/image/registryclient/client.go
sallyom: removed the returning of the ManifestService, updated.

sallyom:
@soltysh pushed the nit-fix
@smarterclayton removed the ManifestService return from Repository
also, updated https://github.com/openshift/oc/pull/439
thanks for your review

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,229 @@
...
+				manifests, err = repo.Manifests(ctx)
+				return replacedImage, repo, manifests, err
+			}
+		}
+	}
+
+	// error if explicitly asked to look in cluster (from command line flag) and didn't connect to image repository
+	if c.ImageSources.LookupClusterICSP {
+		// now try to look for ICSPs from cluster
+		if c.ImageSources.ICSPClientFn == nil {
```
soltysh, line 183: I think you could get away with just passing `ICSPClientFn` which would be a sign that a user requested that, wdyt?
This will mean that in oc in `SecurityOptions#Complete` you'd initialize `o.ICSPClientFn` only if `--lookup-cluster-icsp` was requested. This will simplify this piece of code which is the only relying on the presence of `LookupClisterICSP`.

sallyom: yes, I had to rework use of flags in oc a bit, but now we don't need the LookupClusterICSP.

pkg/image/registryclient/alternativesources.go
```diff
@@ -160,62 +146,40 @@ func (c *Context) PreferredImageSource(image imagereference.DockerImageReference
...
-		}
-		for _, icsRef := range altSources {
-			replacedImage = replaceImage(icsRef, image)
-			if repo, err = c.Repository(ctx, replacedImage.DockerClientDefaults().RegistryURL(), replacedImage.RepositoryName(), insecure); err == nil {
-				manifests, err = repo.Manifests(ctx)
-				return replacedImage, repo, manifests, err
-			}
-		}
+	// now try to look for ICSPs from cluster
+	icspClient, err := c.ImageSources.ICSPClientFn()
```
soltysh, line 166: You still want a nil check here, in case `ICSPClientFn` wasn't passed

pkg/image/registryclient/alternativesources.go
```diff
@@ -18,56 +18,42 @@ import (
...
+	if icspClient == nil {
+		klog.V(4).Infof("no client exists to access ImageContentSourcePolicies in cluster")
+		return
+	}
+	icsps, err := icspClient.List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
-		return err
+		// may or may not have access to ICSPs in cluster
+		// don't error if can't access ICSPs
+		klog.V(4).Infof("did not access any ImageContentSourcePolicies in cluster: %v", err)
```
soltysh, line 46: Return early is the rule, so I'd just add `return` here, log is fine.

pkg/image/registryclient/alternativesources.go
```diff
@@ -160,62 +146,40 @@ func (c *Context) PreferredImageSource(image imagereference.DockerImageReference
...
-			return imagereference.DockerImageReference{}, nil, nil, err
-		}
-		for _, icsRef := range altSources {
-			replacedImage = replaceImage(icsRef, image)
-			if repo, err = c.Repository(ctx, replacedImage.DockerClientDefaults().RegistryURL(), replacedImage.RepositoryName(), insecure); err == nil {
-				manifests, err = repo.Manifests(ctx)
-				return replacedImage, repo, manifests, err
-			}
+	c.addICSPsFromCluster(icspClient)
+	altSources, err = c.addImageSources(image)
```
soltysh, line 171: Now that I'm re-reading this `addImageSources` is more like getter for alternative image sources for the provided image. How about renaming that to `alternateImageSources`?

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	if len(icsps.Items) == 0 {
+		// might be looking up implicitly, log only
+		klog.V(4).Info("no ImageContentSourcePolicies found in cluster")
+	}
+	a.ImageSources.ICSPList = append(a.ImageSources.ICSPList, icsps.Items...)
+}
+
+// addImageSourcePoliciesFromFile appends to list of alternative image sources from ICSP file
+func (a *Context) addImageSourcePoliciesFromFile() error {
```
soltysh, line 56: If you have separate function for reading, I'd move here also the check for `ICSPFile` nil.

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,193 @@
...
+		ICSPFile:     icspFile,
+		ICSPClientFn: icspClientFn,
+	}
+	return c
+}
+
+// addICSPsFromCluster will lookup ICSPs from cluster. Logs errors rather than returning errors. By design,
+// this function has no way of knowing whether it expects to connect to a cluster or not. That is up the the caller.
+func (a *Context) addICSPsFromCluster(icspClient operatorv1alpha1client.ImageContentSourcePolicyInterface) {
+	if icspClient == nil {
```
soltysh, line 38: Why do you need to pass client here directly, if you can easily access `a.ICSPClientFn` check if it's nil or not and act on it.

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,193 @@
...
+			return imagereference.DockerImageReference{}, nil, nil, err
+		}
+		manifests, err = repo.Manifests(ctx)
+		return image, repo, manifests, err
+	}
+
+	var altSources []imagereference.DockerImageReference
+	// error if given an icsp file and don't successfully connect to image repository
+	if len(c.ImageSources.ICSPFile) > 0 {
+		if err = c.addImageSourcePoliciesFromFile(); err != nil {
```
soltysh, line 149: With the above comments about nil checks for `ICSPFile` and `ICSPClientFn` you'd just call here `addImageSourcePoliciesFromFile` and `addICSPsFromCluster` and have a single for loop over the alternative sources. WDYT?

soltysh (commented):
A few more nits...

pkg/image/registryclient/alternativesources.go
sallyom: looking - after pushing the last commit i felt something was still not quite right, I'll update w/ your suggestion :)

sallyom: yes, much better, thanks for seeing that :blush:

pkg/image/registryclient/alternativesources.go
sallyom: right, updated

pkg/image/registryclient/alternativesources.go
sallyom: updated

pkg/image/registryclient/alternativesources.go
sallyom: fixed, updated.

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,170 @@
...
+	"context"
+	"fmt"
+	"io/ioutil"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+
+	"github.com/docker/distribution"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
```
smarterclayton, line 13: It's not appropriate for image/registryclient to take dependencies directly on openshift/api or openshift/client-go, so at a minimum this needs to be moved to another package and the interface (that exposes only incoming repository and alternates) is what you pass in to the context.

pkg/image/registryclient/alternativesources.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	klog.V(2).Infof("Found sources: %v for image: %v", uniqueMirrors, imageRef)
+	return uniqueMirrors, nil
+}
+
+// PreferredImageSource chooses appropriate DockerImageReference for a Context from possible image sources gathered from
+// ImageContentSourcePolicy objects and user-passed image. PreferredImageSource will lookup from cluster
+// if alwaysTryAlternativeSources is true, otherwise will lookup from file if file given, or from
+// original image-reference of user-given image (that may be different from user-given in case of mirrored images).
+func (c *Context) PreferredImageSource(image imagereference.DockerImageReference, alwaysTryAlternativeSources, insecure bool) (imagereference.DockerImageReference, distribution.Repository, distribution.ManifestService, error) {
```
smarterclayton, line 132: As commented in the other PR, this isn't the interface we want (defeats the purpose of putting it here).  We put it here to hide the details of alternates behind the interface.

/hold

I think we should probably set aside time and walk through a basic impl that would accomplish that goal - I may not be clearly expressing the limitations I think should be in place for this, but we want to encapsulate and hide ICSP, not expose it.

soltysh: OK, lemme try to work this out with @sallyom and we'll ping you again. I might have misunderstood or forgotten between when we last spoke and now.

pkg/image/registryclient/alternativesources.go
sallyom: moved all ICSP fns to own package and instead of passing to context, pass context to it :shrug:  so no changes in registryclient.

pkg/image/registryclient/alternativesources.go
sallyom: i pushed an update ptal thanks

pkg/image/registryclient/client.go
```diff
@@ -226,6 +240,32 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
 	return NewLimitedRetryRepository(repo, c.Retries, limiter), nil
 }
 
+// PreferredImageSource takes a list of alternative image sources and returns an appropriate image reference, repository, and manifest service  for a given Context, or an error
+func (c *Context) PreferredImageSource(ctx context.Context, image imagereference.DockerImageReference, alternativeSources []imagereference.DockerImageReference, insecure bool) (imagereference.DockerImageReference, distribution.Repository, distribution.ManifestService, error) {
```
sallyom, line 244: Note: No reference to operatorv1..ICSP

sallyom: If no alternative sources, try the given image ref and return (no change)

pkg/image/registryclient/client.go
```diff
@@ -35,6 +37,11 @@ type RepositoryRetriever interface {
 	Repository(ctx context.Context, registry *url.URL, repoName string, insecure bool) (distribution.Repository, error)
 }
 
+// ImageSource
+type AdditionalImageSources interface {
```
soltysh, line 41: I'd call this interface `ImageSources`

pkg/image/registryclient/client.go
```diff
@@ -35,6 +37,11 @@ type RepositoryRetriever interface {
...
+	AlternativeSources(image imagereference.DockerImageReference) ([]imagereference.DockerImageReference, error)
```
soltysh, line 42: Just `Alternative`, with the field being same as interface, you'll be getting a nice `ImageSources.Alternative` call

pkg/image/registryclient/client.go
soltysh: Yeah, but you're still exposing completely new method, instead of embedding `c.ImageSource != nil` and reading alternatives in `Repository` method.

pkg/image/registryclient/client.go
sallyom: updated, done

pkg/image/registryclient/client.go
sallyom: updated, done

pkg/image/registryclient/client.go
sallyom: _hm...._

pkg/image/registryclient/client.go
sallyom: added interface and moved all icsp fns to new package, as suggested

pkg/image/imagecontentsources/alternativesources.go
```diff
@@ -0,0 +1,134 @@
...
+	klog.V(2).Infof("Found sources: %v for image: %v", uniqueMirrors, imageRef)
+	return uniqueMirrors, nil
+}
+
+// Alternative returns possible DockerImageReferences
+// from image sources gathered from ImageContentSourcePolicy objects and user-passed image.
+// Alternative will lookup from cluster or from ImageContentSourcePolicy file passed from user.
+// Image reference of user-given image may be different from original in case of mirrored images.
+// Satisfies registrycontext.ImageSources interface
+func (a *ImageSources) Alternative(image imagereference.DockerImageReference) ([]imagereference.DockerImageReference, error) {
```
sallyom, line 123: all that's required by Repository call is the registry URL and the repoName (string), so I can rework alternative fns to only return those, rather than the []DockerImageReference

soltysh:
Notes from yesterday's meeting so we don't forget:
1. library-go should only expose interface, the implementation initially will live in oc, since that's the primary consumer. We'll revisit the initial implementation and parts might land in library-go.
2. @dmage will provide his use-cases for this alternative interface, how to best consume it and how registry might need them.

dmage:
I'd suggest to have these interfaces:

```go
// ImageSources is a collection of sources for the image. It is designed as an
// iterator so that it can discover additional sources as needed.
type ImageSources inteface {
    // Next returns an image reference for the image. If there are no more
    // known sources, it returns an empty string. Whenever it's possible, full
    // references (i.e. with registry names and tags/digests) should be
    // preferred.
    Next() (string, error)
}

// ImageResolver retrieves image sources for image references.
type ImageResolver inteface {
    Resolve(imageReference string) (ImageSources, error)
}
```

ImageSources for `oc` can return the original image references on the first call, and then retrieve ICSPs and generate additional sources on subsequent calls.
For the registry it should watch and cache ICSPs and fully obey ICSPs (i.e. return mirrors first and the original reference at the end).

ImageSources should return image references as the caller needs to get credentials for the registry, understand if an insecure connection can be used or if the repository is blocked (we have global policies: https://docs.openshift.com/container-platform/4.6/openshift_images/image-configuration.html).

ImageResolver can be smart enough and resolve short image references into full references.

registryclient.Context should accept a custom ImageResolver. Unless an ImageResolver is provided, an identity resolver should be used (i.e. a resolver that returns only imageReference without any changes).

As Repository accepts `url` and `repoName`, we'll need a higher level method that will accept `imageReference` and be able to use this resolver.

sallyom:
@dmage thanks for the suggestion, working on it

sallyom:
@dmage @smarterclayton @soltysh I implemented the interfaces oleg suggested above, ptal - also updated[ the oc PR](https://github.com/openshift/oc/pull/439)
I appreciate your reviews.

pkg/image/registryclient/client.go
```diff
@@ -35,6 +35,34 @@ type RepositoryRetriever interface {
...
+	// preferred.
+	Next() (string, error)
+}
+
+// ImageResolver retrieves image sources for image references.
+type ImageResolver interface {
+	Resolve(image string) (ImageSources, error)
+}
+
+type defaultImageResolver struct{}
```
soltysh, line 53: There's no value in having this `defaultImageResolver` which is no-op. `ImageResolver` is an optional element of client. You use it if it was provided.

pkg/image/imagecontentsources/alternativesources.go
```diff
@@ -0,0 +1,173 @@
...
+	current := a.Sources
+	newList := append(current[:0], current[0+1:]...)
+	a.Sources = newList
+	return current[0], nil
+}
+
+// Resolve returns a possible image sources for a given image
+// Satisfies registrycontext.ImageResolver interface
+func (a *ICSPSources) Resolve(image string) (registryclient.ImageSources, error) {
+	sources, err := a.alternative(image)
```
soltysh, line 139: Nit: no need for separate method, I'd just embed the contents of `alternative` method in here.

pkg/image/imagecontentsources/alternativesources.go
```diff
@@ -0,0 +1,173 @@
...
+// Image reference of user-given image may be different from original in case of mirrored images.
+func (a *ICSPSources) alternative(image string) ([]string, error) {
+	imageRef, err := imagereference.Parse(image)
+	if err != nil {
+		return nil, err
+	}
+	if err := a.addImageSourcePoliciesFromFile(); err != nil {
+		return nil, err
+	}
+	if len(a.ICSPFile) == 0 {
```
soltysh, line 158: To make it nicer, you could return from `addImageSourcePoliciesFromFile` information how many sources your read or boolean flag indicating if you read anything, and error. Just a suggestion :wink:

pkg/image/imagecontentsources/alternativesources.go
```diff
@@ -0,0 +1,173 @@
...
+
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	operatorv1alpha1scheme "github.com/openshift/client-go/operator/clientset/versioned/scheme"
+	operatorv1alpha1client "github.com/openshift/client-go/operator/clientset/versioned/typed/operator/v1alpha1"
+	imagereference "github.com/openshift/library-go/pkg/image/reference"
+	"github.com/openshift/library-go/pkg/image/registryclient"
+)
+
+// ICSPSources holds ImageContentSourcePolicy variables to look up image sources
+type ICSPSources struct {
```
soltysh, line 20: I'd consider having a contructor for this struct: `NewICSPSources(file string, client peratorv1alpha1client.ImageContentSourcePolicyInterface) ImageResolver` all the remaining implementation details I'd hide behind private members of this package.

soltysh (commented):
A few nits about the code itself, I like this interface, but I'll defer to @smarterclayton and @dmage for the shape of it.

soltysh:
A suggestion, I'd split this PR into 3 commits, for readability:
1. that 2 new interfaces that Oleg proposed
2. reference implementation with ICSPFile and ICSPClient
3. bump
This will make the reviews nicer.

pkg/image/registryclient/client.go
sallyom: I also don't see a value of the extra ImageSources interface. I have a single ImageResolver interface with ResolveNext(string) (string, error)

pkg/image/imagecontentsources/alternativesources.go
sallyom: done, thanks

pkg/image/imagecontentsources/alternativesources.go
sallyom: thanks for the suggestion, done!

pkg/image/imagecontentsources/alternativesources.go
sallyom: > A suggestion, I'd split this PR into 3 commits, for readability:
> 
> 1. that 2 new interfaces that Oleg proposed
> 2. reference implementation with ICSPFile and ICSPClient
> 3. bump
>    This will make the reviews nicer.

haven't split yet, but I've found I only need the single `ImageResolver` interface with the method `ResolveNext` I've pushed this

sallyom:
> A suggestion, I'd split this PR into 3 commits, for readability:
> 
> 1. that 2 new interfaces that Oleg proposed
> 2. reference implementation with ICSPFile and ICSPClient
> 3. bump
>    This will make the reviews nicer.

haven't split yet, but I've found I only need the single `ImageResolver` interface with the method `ResolveNext` I've pushed this

pkg/image/imagecontentsources/alternativesources.go
sallyom: I think it's fine as/is, since whenever an icsp file is passed we want to error if no icsp obj was decoded. We don't want to  return nil if file was successfully accessed but no icsp was found, that should be always an error.

soltysh:
I'm happy with the current shape, we need final ack from @smarterclayton and @dmage

dmage:
I'm OK with the current state, we can continue to work on it in follow up PRs and land the current version to unblock `oc` (unless @smarterclayton has another opinion).
/lgtm

pkg/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return nil, ref, fmt.Errorf("did not find alternative image reference for %s", ref.String())
+}
+
+// addICSPsFromCluster will lookup ImageContentSourcePolicy resources in cluster.
+func (s *simpleLookupICSP) addICSPsFromCluster() error {
+	if s.icspClient == nil {
+		return fmt.Errorf("no client to access ImageContentSourcePolicies in cluster")
+	}
+	icsps, err := s.icspClient.List(context.TODO(), metav1.ListOptions{})
```
ricardomaraschini, line 102: nit: wouldn't it be better to pass down the context from `WithAlternateSources` here?

sallyom: yes, thanks, fixed

sallyom:
> I'm OK with the current state, we can continue to work on it in follow up PRs and land the current version to unblock `oc` (unless @smarterclayton has another opinion).
> /lgtm

@dmage based on discussion w/ @smarterclayton I've updated this, it will require another review, thank you

pkg/image/registryclient/client.go
```diff
@@ -226,6 +244,37 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
 	return NewLimitedRetryRepository(repo, c.Retries, limiter), nil
 }
 
+// RepositoryWithAlternateReference will return the next cached repository and ref, or will gather alternate sources then return first repo, ref, or error
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, registry *url.URL, insecure bool) (distribution.Repository, imagereference.DockerImageReference, error) {
```
ricardomaraschini, line 248: Do we need this `registry` argument? Seems like it is not used.

ricardomaraschini (changes requested):

pkg/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,221 @@
...
+		repoDigestMirrors := icsp.Spec.RepositoryDigestMirrors
+		var sourceMatches bool
+		for _, rdm := range repoDigestMirrors {
+			rdmRef, err := reference.Parse(rdm.Source)
+			if err != nil {
+				return nil, err
+			}
+			if imageRef.AsRepository() == rdmRef.AsRepository() {
+				klog.V(2).Infof("%v RepositoryDigestMirrors source matches given image", imageRef.AsRepository())
+				sourceMatches = true
```
ricardomaraschini, line 148: nit: this could be easier to read and a little bit faster if we inverted the logic to something:

```golang
if imageRef.AsRepository() != rdmRef.AsRepository() {
    continue
}

klog.V(2).Infof("%v RepositoryDigestMirrors source matches given image", imageRef.AsRepository())
```

We would avoid a unnecessary run over the `Mirrors` slice and the `sourceMatches` would not be needed anymore.

pkg/image/registryclient/client.go
sallyom: probably left over from previous implementation I'll take a look, thanks

pkg/image/strategy/simplelookup.go
ricardomaraschini: In fact seems like I might have hit an issue with the logic in here, taking into account the following ImageContentSourcePolicy :

```yaml
apiVersion: operator.openshift.io/v1alpha1
kind: ImageContentSourcePolicy
metadata:
  name: goiaba
spec:
  repositoryDigestMirrors:
  - source: quay.io/rmarasch/does-not-exist
    mirrors:
    - goiaba.news:5000/rmarasch/simple-web-server
  - source: docker.io/rmarasch/does-not-exist
    mirrors:
    - i.do.not.exist/rmarasch/simple-web-server
```

When pulling `quay.io/rmarasch/does-not-exist` I would expect to have as an alternative `goiaba.news:5000/rmarasch/simple-web-server` but I am seeing `i.do.not.exist/rmarasch/simple-web-server` instead.

sallyom: I have a unit test for that: [test icspfile](https://github.com/openshift/library-go/blob/076e9be4cebf1a7f9b8c8434967dd058bac8622e/pkg/image/strategy/simplelookup_test.go#L14-#L39)  and [test case](https://github.com/openshift/library-go/blob/076e9be4cebf1a7f9b8c8434967dd058bac8622e/pkg/image/strategy/simplelookup_test.go#L96-#L97)  but hm, if you are seeing different I'll take a closer look.

pkg/image/registryclient/client.go
sallyom: removed, fixed

pkg/image/registryclient/client.go
```diff
@@ -226,6 +244,37 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
...
+		if err != nil {
+			return nil, locator, err
+		}
+	}
+	// iterate to get preferred ref, repo
+	altRepo, altRef, err := c.AlternateRepositoryStrategy.RepositoryForAlternateReference(ctx, locator, c.cachedRepos)
+	if err != nil {
+		return nil, locator, err
+	}
+	delete(c.cachedRepos, altRef)
```
smarterclayton, line 274: You can't do this without holding the context lock

pkg/image/registryclient/client.go
```diff
@@ -374,6 +426,24 @@ func NewLimitedRetryRepository(repo distribution.Repository, retries int, limite
...
+	}
+	if len(cachedRepos) == 0 {
+		return fmt.Errorf("did not find any repositories for %s", locator)
+	}
+	c.cachedRepos = cachedRepos
+	return nil
+}
+
 // isTemporaryHTTPError returns true if the error indicates a temporary or partial HTTP failure
 func isTemporaryHTTPError(err error) (time.Duration, bool) {
```
smarterclayton, line 448: Was expecting to see one example of this in use with the blob service / manifest service wrapper (the Get method) showing how you will use these methods

sallyom: You mean a method that does more than return the first repo and ref? `oc` uses the `retryRepository` - I found it isn't necessary to create a new repository wrapper.  and the manifests.Get is called in the strategy method, `RepositoryForAlternateReference`.  I've only implemented the`simpleLookupICSP` strategy that `oc` will use.   
I added the Context method `NextRepositoryWithAlternateReference` to return the next (only&preferred for oc) repo, ref found from RepositoryForAlternateReference. Depending on the strategy,  RepoForAlternateRef may have more complex ordering, another strategy can be used but I only added the simpleLookupICSP to get the next 'successful' repo, ref. 
I may be misunderstanding your comment.

pkg/image/registryclient/client.go
sallyom: added

pkg/image/registryclient/client.go
smarterclayton: No one in this PR calls NextRepositoryWithAlternateReference, which means someone outside this library has to interpret it.  I was expecting `retryRepository` to invoke alternate lookups instead - so that someone outside this library doesn't (and I could see the structure change for retryRepository.

pkg/image/registryclient/client.go
```diff
@@ -374,6 +429,24 @@ func NewLimitedRetryRepository(repo distribution.Repository, retries int, limite
...
+		altRepo, err := c.Repository(ctx, altRef.DockerClientDefaults().RegistryURL(), altRef.RepositoryName(), true)
+		if err != nil {
+			return err
+		}
+		if altRepo != nil {
+			cachedRepos[altRef] = altRepo
+		}
+	}
+	if len(cachedRepos) == 0 {
+		return fmt.Errorf("did not find any repositories for %s", locator)
```
ricardomaraschini, line 444: I think it would be better here to return an error indicating that we have no more "alternates" to try. Something similar to what happens when we reach the end of a file (`io.EOF`).  That way one can easily distinguish between an expected error (there is no more alternates) than an unexpected one (like the one returned on line 435 above).

ricardomaraschini:
I have coded an example of usage of this change in [here](https://github.com/ricardomaraschini/dump-place/blob/master/pull-through-mirror/icspaware.go#L80-L134). What I noticed is that it could be a little bit easier if `NextRepositoryWithAlternateReference` returned the original repository as the last call. If it did that [this](https://github.com/ricardomaraschini/dump-place/blob/master/pull-through-mirror/icspaware.go#L119-L133) would not be needed.

pkg/image/registryclient/client.go
sallyom: the commit is outdated, as i updated/pushed thinking there were no more outstanding reviews - however, I updated the error message there to indicate "alternates" instead of an error fetching the repository.

pkg/image/registryclient/client.go
sallyom: I've updated, `nextRepositoryWithReference` is retryRepository method, and 

> I have coded an example of usage of this change in [here](https://github.com/ricardomaraschini/dump-place/blob/master/pull-through-mirror/icspaware.go#L80-L134). What I noticed is that it could be a little bit easier if `NextRepositoryWithAlternateReference` returned the original repository as the last call. If it did that [this](https://github.com/ricardomaraschini/dump-place/blob/master/pull-through-mirror/icspaware.go#L119-L133) would not be needed.

I reworked this based on @smarterclayton  comment - the exported Context method `RepositoryWithAlternateReference` returns the first alternate repo from the retryRepository method `nextRepositoryForAlternateReference`  and the Context.cachedRepos is updated accordingly. The exported `RepositoryWithAlternateReference` only returns repos for which the ManifestService.Get(ctx, locator-digest) succeeds. The caller can collect those and do more ordering if necessary.

pkg/image/registryclient/client.go
ricardomaraschini: I see. The use case I am trying to tackle is the following:

1) try all alternate registry for a given manifest
2) if could not find the manifest in any of the alternate registries give the original registry a try

On such scenario I would iterate over `NextRepositoryWithAlternateReference()` until it returns an error and then move to the original registry but the only way I can distinguish between the end of the alternate registries list and any other error is by comparing the string within the error. That is what I would like to avoid. What I was proposing is an specific error that then I can compare, something like `if err == EndOfAlternates`.

If the proposed scenario is not a valid one then please let me know.

/cc @dmage

ricardomaraschini: Nevermind, this has been addressed by your last commits. I don't need to iterate anymore. Thanks.

pkg/image/strategy/simplelookup.go
ricardomaraschini: I can't reproduce this issue anymore, please disregard this comment.

pkg/image/registryclient/client.go
sallyom: oh great! and yea, if it makes sense anywhere we'll add a custom error.

pkg/image/registryclient/client.go
```diff
@@ -223,7 +248,48 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
...
+		retryRepo.alternateStrategy = c.AlternateRepositoryStrategy
+	}
+	return retryRepo, nil
+}
+
+// RepositoryWithAlternateReference returns the next cached repository and ref, or will gather alternate sources then return first repo, ref, or error
+// Once the repository is returned it is removed from the cache.
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, cachedRepos map[imagereference.DockerImageReference]distribution.Repository) (distribution.Repository, imagereference.DockerImageReference, error) {
+	repo, err := c.repository(ctx, locator.DockerClientDefaults().RegistryURL(), locator.RepositoryName(), true)
+	if repo.alternateStrategy == nil {
```
ricardomaraschini, line 262: I am porting `image-registry` to use this patch and I am seeing the following panic here:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
```

That is because `repo` is `nil`.

ricardomaraschini (changes requested):

pkg/image/registryclient/client.go
```diff
@@ -223,7 +248,48 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
...
+	if c.AlternateRepositoryStrategy != nil {
+		retryRepo.alternateStrategy = c.AlternateRepositoryStrategy
+	}
+	return retryRepo, nil
+}
+
+// RepositoryWithAlternateReference returns the next cached repository and ref, or will gather alternate sources then return first repo, ref, or error
+// Once the repository is returned it is removed from the cache.
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, cachedRepos map[imagereference.DockerImageReference]distribution.Repository) (distribution.Repository, imagereference.DockerImageReference, error) {
+	repo, err := c.repository(ctx, locator.DockerClientDefaults().RegistryURL(), locator.RepositoryName(), true)
```
ricardomaraschini, line 261: As a side question: are we always consider "insecure = true" for alternate registries?

sallyom: if so, I'll carry the `insecure` variable through from the caller - I'm not sure, I recall a discussion where we decided we can default to true but I might be mis-remembering. 
@smarterclayton can `insecure` be true always here?

pkg/image/registryclient/client.go
sallyom: I added a check for if retryRepository.Repository == nil, return an error (if no alternative image strategy). I think that is where the nil panic was coming from? you were returned a nil Repository, then tried to use it? If not, I'll take a closer look, get more information from you. ie, was the retryRepository nil or the returned retryRepository.Repository?

sallyom: got it, thanks

ricardomaraschini: Thanks @sallyom. That addressed a problem with our tests in `image-registry`

pkg/image/registryclient/client.go
ricardomaraschini: Image registry only uses "insecure = true" when:

1. ImageStream has an Annotation (`openshift.io/image.InsecureRepository`) set to "true" 
2. The tag being pulled has its `importPolicy.Insecure` set to `true`

As far as I understand this `insecure` argument makes the registry client use the insecure transport and, in some cases, fallback to http instead of https. If it is always true then a dozen tests in image-registry fail (the ones expecting it not to be able to pull from an insecure registry).

Can we have the `insecure` parameter also in `RepositoryWithAlternateReference()`.

pkg/image/registryclient/client.go
```diff
@@ -352,19 +421,43 @@ func (c *Context) repositoryTransport(t http.RoundTripper, registry *url.URL, re
 	return c.cachedTransport(t, c.scopes(repoName))
 }
 
+func (c *Context) reposForRef(ctx context.Context, locator imagereference.DockerImageReference, sources []imagereference.DockerImageReference) error {
+	cachedRepos := make(map[imagereference.DockerImageReference]distribution.Repository)
+	for _, altRef := range sources {
+		altRepo, err := c.Repository(ctx, altRef.DockerClientDefaults().RegistryURL(), altRef.RepositoryName(), true)
```
dmage, line 427: No, a repository for the next source shouldn't be created until it is proved that the current source cannot provide the image.

dmage (changes requested):
ICSP is often used when the original source is not available, the client should access registries one by one. It shouldn't try to access all sources at the same time.

Creating a repository instance causes an HTTP request. So it's wrong to create repositories for all known sources in advance.

pkg/image/registryclient/client.go
```diff
@@ -223,7 +248,51 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
...
+		sources, err := repo.alternateStrategy.FirstRequest(ctx, locator)
+		if err != nil {
+			return nil, locator, err
+		}
+		if sources == nil {
+			sources, err = repo.alternateStrategy.WithAlternateSources(ctx, locator, err)
+			if err != nil {
+				return nil, locator, err
+			}
+		}
```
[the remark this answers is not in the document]
sallyom: @smarterclayton  suggested the 2 methods, with only the WithAlternativeSources used in the simple lookup method. Other strategies (that I have not added, as I'm only adding the simple lookup strategy, will utilize the FirstRequest) - He suggested `FirstRequest` and `OnFailure`, but I made OnFailure WithAlternateSources.  - He will have to give more context for strategies other than the one I've provided, bc I don't have that context.

pkg/image/registryclient/client.go
sallyom: yes, I will add that, got it

pkg/image/registryclient/client.go
sallyom: ok, I can update that, that was my original assumption but somehow in conversations I was led to believe there was a use-case for collecting possible Repositories, choosing from them - again, I don't know of that use-case just that it possibly exists - @ricardomaraschini also let me know that this is not something image-registry would need.

pkg/image/registryclient/client.go
```diff
@@ -223,7 +248,51 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
...
+	retryRepo := NewLimitedRetryRepository(repo, c.Retries, limiter)
+	if c.AlternateRepositoryStrategy != nil {
+		retryRepo.alternateStrategy = c.AlternateRepositoryStrategy
+	}
+	return retryRepo, nil
+}
+
+// RepositoryWithAlternateReference returns the next cached repository and ref, or will gather alternate sources then return first repo, ref, or error
+// Once the repository is returned it is removed from the cache.
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, cachedRepos map[imagereference.DockerImageReference]distribution.Repository) (distribution.Repository, imagereference.DockerImageReference, error) {
```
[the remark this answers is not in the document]
sallyom: no, that would be used in a different strategy than the one I've provided.

pkg/image/registryclient/client.go
sallyom: great, thank you for the update

sallyom:
> ICSP is often used when the original source is not available, the client should access registries one by one. It shouldn't try to access all sources at the same time.
> 
> Creating a repository instance causes an HTTP request. So it's wrong to create repositories for all known sources in advance.

Agreed, and will update

pkg/image/registryclient/client.go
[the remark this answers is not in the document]
sallyom: in `oc` it will be used 1 time, to get the first valid Repository. To get multiple valid Repositories, caller would repeat. Again, I was under the assumption there was a need for this. It's not required by `oc` and not required by image-registry.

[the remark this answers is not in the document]
sallyom: I've updated. ptal

pkg/image/registryclient/client.go
sallyom: added, pushed

pkg/image/registryclient/client.go
sallyom: and I updated that just now to only return if repo==nil, since that's what is needed (otherwise will skip looking at alt sources if err != nil && repo != nil

pkg/image/registryclient/client.go
sallyom: ^^ outdated ^^ removed the cachedRepos

sallyom:
@smarterclayton this has been updated, ptal

pkg/image/registryclient/client.go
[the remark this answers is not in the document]
soltysh: I'll leave it up to @smarterclayton to decide on naming, I don't feel too strongly in either direction.

pkg/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+	if len(s.icspFile) == 0 {
+		// log errors from accessing cluster since this function has no way of knowing whether it should or should not succeed
+		if err := s.addICSPsFromCluster(ctx); err != nil {
+			klog.V(4).Infof("No alternative image sources gathered from cluster: %v", err)
+		}
+	}
+	icsRefList, err := s.alternativeImageSources(imageRef)
+	if err != nil {
```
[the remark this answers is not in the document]
sallyom: thanks, i made that change - I removed icspList from simpleLookupICSP, and I'm returning the icspList fromFile & fromCluster, then passing it to alternativeImageSources as you suggested. you're right, it's easier to follow that way.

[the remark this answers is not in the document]
soltysh: I'm not a fan of switches :stuck_out_tongue_closed_eyes: in such cases, but I know @smarterclayton likes them :wink: but this reads so much clearer, thank you!

pkg/image/registryclient/client.go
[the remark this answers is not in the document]
dmage: I suspect @smarterclayton meant something different and there should be some actions between `FirstRequest` and `OnFailure`.

I suspect the client should check presence of the image in sources from `FirstRequest` and if it cannot find the the requested image, then it needs to get another list of sources from `OnFailure`.

pkg/image/registryclient/client.go
```diff
@@ -352,19 +417,66 @@ func (c *Context) repositoryTransport(t http.RoundTripper, registry *url.URL, re
...
+	default:
+		dgst, err = digest.Parse(locator.ID)
+		if err != nil {
+			return nil, altRef, err
+		}
+	}
+	_, err = manifests.Get(ctx, dgst)
+	if err != nil {
+		return nil, altRef, err
+	}
```
dmage, line 458: You don't need TagService:
```go
    if locator.ID == "" {
        _, err := ms.Get(ctx, "", distribution.WithTag(locator.Tag))
	if err != nil {
		return nil, altRef, err
	}
    } else {
        _, err := ms.Get(ctx, digest.Digest(locator.ID))
	if err != nil {
		return nil, altRef, err
	}
    }
```

pkg/image/registryclient/client.go
```diff
@@ -200,30 +215,80 @@ func (c *Context) Ping(ctx context.Context, registry *url.URL, insecure bool) (h
...
-	limiter := c.Limiter
-	if limiter == nil {
-		limiter = rate.NewLimiter(rate.Limit(5), 5)
+	retryRepo.Repository = repo
+	return retryRepo, nil
+}
+
+// RepositoryWithAlternateReference returns a a valid repository and ref, or will gather alternate sources then return first repo, ref, or error
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, insecure bool) (distribution.Repository, imagereference.DockerImageReference, error) {
+	repo, err := c.repository(ctx, locator.DockerClientDefaults().RegistryURL(), locator.RepositoryName(), insecure)
```
dmage, line 260: This call is not desirable for the image registry, the registry should access mirrors first.

pkg/image/registryclient/client.go
```diff
@@ -352,19 +417,66 @@ func (c *Context) repositoryTransport(t http.RoundTripper, registry *url.URL, re
 	return c.cachedTransport(t, c.scopes(repoName))
 }
 
+func (c *Context) repoForRef(ctx context.Context, locator imagereference.DockerImageReference, altRef imagereference.DockerImageReference, insecure bool) (distribution.Repository, imagereference.DockerImageReference, error) {
```
dmage, line 420: Why does this method need both `locator` and `altRef`?

pkg/image/registryclient/client.go
sallyom: ok, will remove

pkg/image/registryclient/client.go
sallyom: I'll update but oc may prefer to try original image sources first, since the `oc adm release` commands are opt out for alternate image source lookup, they implicitly lookup icsp. If there is a strong feeling against oc trying mirrored sources first, then i'll (add a boolean? separate fn?)

pkg/image/registryclient/client.go
sallyom: updated

pkg/image/registryclient/client.go
sallyom: they are not always the same. The locator is the passed imagereference and altRef is the possible mirror alternative found from ICSP object. If you have an ICSP spec like this:
```
spec:
  repositoryDigestMirrors:
  - mirrors:
    - registry.ci.openshift.org/ocp/4.7
    source: does/not/exist
```
you can run something like
`oc image extract does/not/exist:tests --path=/:/tmp/ext  --icsp-file ~icsp.yaml` 
The locator: does/not/exist:tests
The altRef: registry.ci.openshift.org/ocp/4.7

pkg/image/strategy/simplelookup.go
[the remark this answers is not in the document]
sallyom: no thank you! I made it an if/else :smile:

pkg/image/registryclient/client.go
dmage: I'd expect altRef to be `registry.ci.openshift.org/ocp/4.7:tests`

pkg/image/registryclient/client.go
```diff
@@ -226,6 +249,46 @@ func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName st
 	return NewLimitedRetryRepository(repo, c.Retries, limiter), nil
 }
 
+// RepositoryWithAlternateReference returns a a valid repository and ref, it gathers alternate sources then returns first repo, ref, or error
+func (c *Context) RepositoryWithAlternateReference(ctx context.Context, locator imagereference.DockerImageReference, insecure bool) (distribution.Repository, imagereference.DockerImageReference, error) {
```
smarterclayton, line 253: Since the point of FirstRequest is to be called the first time get is invoked, I don't see how this code knows that this is the "first request", nor do I see the logic I asked for before that shows this in the context of a specific method.

Who calls this?

smarterclayton: To help frame what I'm asking let me take a stab at implementing the logic inline for the wrappers that I expect, then we can circle back

sallyom: This is called by `oc` (pkg/cli/image/imagesource/options.go, replacing the `regcontext.Repository`) [link to oc PR](https://github.com/openshift/oc/pull/439/files#diff-22e127520b7182edf749daf679f1963186ce0e55e13edf001780b06f4b59d4f7)

smarterclayton: I asked for oc not to need to do that though - that means that multiple callers have to know about this.

I tried to capture this into a semi working form in https://github.com/openshift/library-go/pull/1021 - see how methods like Get and Open are retried (by calling the alternates, which defers and shares the source with multiple invocations), while calls like Put and Create are only invoked once.  Once you get a chance to look through that code, we should schedule some time and see if we can work through where gaps may be between this PR and my example code.

sallyom: Image-registry will try the mirrored references first, b4 the source ref, which is why I am gathering sources in that order . They don't afaict have a need for FirstRequest, and neither does `oc` at least with the issues that i've been working to resolve. I don't have enough context to implement this - you suggested earlier to add it and for the simple strategy return nil sources & nil error, which is what I have currently done - I'll need to know more about a use case for this in order to implement it further.

pkg/image/registryclient/client.go
sallyom: altRef is an ref.AsRepository() and the altRef:locator ID is tried

pkg/image/registryclient/client.go
sallyom: ^^ i did not see your reply when i wrote that, did not refresh the page.. i'm looking at your example now

pkg/image/registryclient/client.go
dmage: I understand, but altRef could (and I think should) include all necessary information.

sallyom: but you can't know the digest, the alternates only tell you that the registry is accessible, for release images that reference multiple digests that can't be stored. (w/ oc adm release extract --command openshift-install, you _first_ need to know you can access the release image by its digest, then you can check for the digest of the installer - so the installer ID has to be attached to the accessible AsRepository() ref ) -anyways, clayton has suggested an extensive rewrite, so i'm working on that now

pkg/image/registryclient/client.go
sallyom: It's taking me a bit of time to refactor this to include clayton's suggestion, but I'm making progress,  and will reach out if/when stuck.

sallyom:
Incorporated clayton's https://github.com/openshift/library-go/pull/1021 with some minor diff - i'd cherry-pick then edit as another commit if the commit msg is updated (it's "DO NOT MERGE" atm) - I've also updated oc to pull this in, however oc is not utilizing most of the function in the alternative lookup strategy, only using the replace of original registry/repo/name with mirrored registry/repo/name without ICSP lookup - image-registry will have to verify the ICSP lookup (simple lookup strategy)  is working b4 this merges.

pkg/image/registryclient/client.go
sallyom: ok i've pushed w/ clayton's #1021, ptal

bparees:
@sallyom this is blocking an important 4.8 bug (https://bugzilla.redhat.com/show_bug.cgi?id=1918376), who's signoff are we waiting on?

pkg/image/strategy/simplelookup.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// simpleLookupICSP holds ImageContentSourcePolicy variables to look up image sources
+// satisfies *Context AlternativeRepositoryStrategy interface
+type simpleLookupICSP struct {
+	lock sync.Mutex
+
+	alternates []reference.DockerImageReference
+	icspFile   string
+	icspClient operatorv1alpha1client.ImageContentSourcePolicyInterface
```
smarterclayton, line 27: This should be in a separate package as discussed before

pkg/image/registryclient/client.go
```diff
@@ -26,7 +26,6 @@ import (
 	"github.com/docker/distribution/registry/client/auth/challenge"
 	"github.com/docker/distribution/registry/client/transport"
 	"github.com/opencontainers/go-digest"
-	imagereference "github.com/openshift/library-go/pkg/image/reference"
```
smarterclayton: This commit merged already - surprised it still shows up here

pkg/image/strategy/simplelookup.go
sallyom: cleaned up commits & cherry-picks

pkg/image/registryclient/client.go
sallyom: i cleaned up the commits & cherry-picks, it's no longer there

sallyom:
@smarterclayton the small modifications from your commits are in  0e622f2 , otherwise I picked yours and updated the commit msgs

sallyom:
@ricardomaraschini  can you please run with this PR to detect issues? Thank you, reach out on slack if needed.

soltysh (approved):

sallyom:
I'm fine with this approach because I can make it work to solve the `oc` issue, but I prefer the simpler solution https://github.com/openshift/library-go/compare/master...sallyom:icsp_old?expand=1 [integrated with oc](https://github.com/openshift/oc/compare/master...sallyom:icsp_old?expand=1). @ricardomaraschini afaict verified that solution works for image-registry, however, I don't know if this mirrored_client has been tested with image-registry. `oc` is not using the majority of the changes introduced here, so I cannot verify them.

pkg/image/registryclient/client.go
```diff
@@ -200,18 +208,67 @@ func (c *Context) Ping(ctx context.Context, registry *url.URL, insecure bool) (h
...
+// Repository instance is threadsafe but the ManifestService, TagService, or BlobService are not.
+func (c *Context) RepositoryForRef(ctx context.Context, ref imagereference.DockerImageReference, insecure bool) (distribution.Repository, error) {
+	return c.connectToRegistry(ctx, repositoryLocator{ref: ref}, insecure)
+}
+
+// Repository returns a distribution.Repository against the provided registry and repository name. If insecure
+// is true, HTTP connections are allowed and HTTPS certificate verification errors will be ignored. The returned
+// Repository instance is threadsafe but the ManifestService, TagService, or BlobService are not. Note - the caller
+// is responsible for providing a valid registry url for docker.io - use RepositoryForRef() to avoid that.
+func (c *Context) Repository(ctx context.Context, registry *url.URL, repoName string, insecure bool) (RepositoryWithLocation, error) {
```
ricardomaraschini, line 222: This change made `*Context` to stop implementing `RepositoryRetriever` interface. Was that expected?

sallyom: Are you currently using that interface? I'm wondering if anybody uses it.  I can update the interface though.

sallyom:
@ricardomaraschini I was able to add back to `oc` the ICSP logic, so I am now using more of the mirrored_client. Have you tried to run w/ it yet? I don't have it working quite yet, so I think perhaps something in the mirrored_client might need to be tweaked. Or, maybe I have to update my implementation of it in `oc` - if you have it working  it will help! If not, I'll keep working on it.

sallyom:
@smarterclayton @ricardomaraschini I have the ICSP stuff working w/ the mirrored_client now- using the *WithLocation fns, I'll push to the oc PR so you can check it out

pkg/image/registryclient/client.go
ricardomaraschini: Yes, image registry repository leverages this interface in a few places.

ricardomaraschini:
> @ricardomaraschini I was able to add back to `oc` the ICSP logic, so I am now using more of the mirrored_client. Have you tried to run w/ it yet? I don't have it working quite yet, so I think perhaps something in the mirrored_client might need to be tweaked. Or, maybe I have to update my implementation of it in `oc` - if you have it working it will help! If not, I'll keep working on it.

I have managed to make it work this morning. The problem now (for the image registry) is the one related to the interface I mentioned [here](https://github.com/openshift/library-go/pull/939#discussion_r611457264). 

Other problem I can solve later on is the fact that  `simpleLookupICSP` prefers to try the original repository first before moving to the mirrors and we would like to go for the mirrors first. I can solve that by providing a new implementation of `AlternateBlobSourceStrategy`.

sallyom:
> > @ricardomaraschini I was able to add back to `oc` the ICSP logic, so I am now using more of the mirrored_client. Have you tried to run w/ it yet? I don't have it working quite yet, so I think perhaps something in the mirrored_client might need to be tweaked. Or, maybe I have to update my implementation of it in `oc` - if you have it working it will help! If not, I'll keep working on it.
> 
> I have managed to make it work this morning. The problem now (for the image registry) is the one related to the interface I mentioned [here](https://github.com/openshift/library-go/pull/939#discussion_r611457264).
> 
> Other problem I can solve later on is the fact that `simpleLookupICSP` prefers to try the original repository first before moving to the mirrors and we would like to go for the mirrors first. I can solve that by providing a new implementation of `AlternateBlobSourceStrategy`.

Right, preferring the original allows `oc` to implicitly lookup alternate sources, but only if the original cannot be fetched. This is what is required of oc, but i see this is the opposite of image-registry. We can add another strategy in a follow-up PR, or, can keep the new strategy local to image-registry/not in lib-go, or, can add to this PR. 

I updated the Repository method in the RepositoryRetriever interface, ptal.

ricardomaraschini:
> > > @ricardomaraschini I was able to add back to `oc` the ICSP logic, so I am now using more of the mirrored_client. Have you tried to run w/ it yet? I don't have it working quite yet, so I think perhaps something in the mirrored_client might need to be tweaked. Or, maybe I have to update my implementation of it in `oc` - if you have it working it will help! If not, I'll keep working on it.
> > 
> > 
> > I have managed to make it work this morning. The problem now (for the image registry) is the one related to the interface I mentioned [here](https://github.com/openshift/library-go/pull/939#discussion_r611457264).
> > Other problem I can solve later on is the fact that `simpleLookupICSP` prefers to try the original repository first before moving to the mirrors and we would like to go for the mirrors first. I can solve that by providing a new implementation of `AlternateBlobSourceStrategy`.
> 
> Right, preferring the original allows `oc` to implicitly lookup alternate sources, but only if the original cannot be fetched. This is what is required of oc, but i see this is the opposite of image-registry. We can add another strategy in a follow-up PR, or, can keep the new strategy local to image-registry/not in lib-go, or, can add to this PR.
> 
> I updated the Repository method in the RepositoryRetriever interface, ptal.

We can do that in another PR for sure. As long as the interface has been updated I am OK with it :-)

ricardomaraschini:
Could we get this in? I have dependent work waiting to be done in https://github.com/openshift/image-registry/pull/267

soltysh:
Replaced with #1084
/close

PR #940, opened 2020-11-10
Bug 1851810: pkg/crypto: add function to whitelist valid etcd ciphers

hexfusion:
This PR adds a function that will take a list of cipherSuites and return a list of whitelisted cipherSuites for use by etcd.

proof: https://github.com/openshift/cluster-etcd-operator/pull/480

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -27,6 +27,7 @@ import (
 
 	"k8s.io/klog/v2"
 
+	etcdtlsutil "go.etcd.io/etcd/pkg/tlsutil"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/client-go/util/cert"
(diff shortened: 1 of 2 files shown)


pkg/crypto/crypto.go
```diff
@@ -284,6 +285,18 @@ func DefaultCiphers() []uint16 {
...
+// WhitelistEtcdCipherSuites whitelists valid ciphers for use with etcd.
+func WhitelistEtcdCipherSuites(cipherSuites []string) []string {
+	whitelist := []string{}
+	for _, cipher := range cipherSuites {
+		_, ok := etcdtlsutil.GetCipherSuite(cipher)
+		if ok {
+			whitelist = append(whitelist, cipher)
+		}
+	}
+	return whitelist
```
hexfusion, line 297: probably should sort this list.

hexfusion:
/hold 

proof https://github.com/openshift/cluster-etcd-operator/pull/480

hexfusion:
/hold cancel

proof passed.

pkg/crypto/crypto.go
```diff
@@ -284,6 +285,19 @@ func DefaultCiphers() []uint16 {
 	}
 }
 
+// WhitelistEtcdCipherSuites whitelists valid ciphers for use with etcd.
+func WhitelistEtcdCipherSuites(cipherSuites []string) []string {
```
sttts, line 289: why is this function in library-go?

hexfusion: Thought was that apiserver should also define whitelisted ciphers in clientv3 tlsConfig so having this here would allow reasonable reuse but I can put it upstream etcd.

hexfusion:
/close
moving to cluster-etcd-operator for interim until we backport upstream.

PR #941, opened 2020-11-18
status controller: allow setting RelatedObjects dynamically

squeed:
Some operators (e.g. cno) don't know the set of related objects at startup. So, allow that to be dynamically generated.

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -33,9 +33,12 @@ type VersionGetter interface {
 	VersionChangedChannel() <-chan struct{}
 }
 
+type RelatedObjectsFunc func() (isset bool, objs []configv1.ObjectReference)
+
 type StatusSyncer struct {
 	clusterOperatorName string
 	relatedObjects      []configv1.ObjectReference
+	relatedObjectsFunc  RelatedObjectsFunc
 
 	versionGetter         VersionGetter
 	operatorClient        operatorv1helpers.OperatorClient
@@ -76,6 +79,21 @@ func NewClusterOperatorStatusController(
 		),
 		recorder: recorder.WithComponentSuffix("status-controller"),
 	}
+}
+
+// WithRelatedObjectsFunc allows the set of related objects to be dynamically
+// determined.
+//
+// The function returns (isset, objects)
+//
+// If isset is false, then the set of related objects is copied over from the
+// existing ClusterOperator object. This is useful in cases where an operator
+// has just restarted, and hasn't yet reconciled.
+//
+// Any statically-defined related objects (in NewClusterOperatorStatusController)
+// will always be included in the result.
+func (c *StatusSyncer) WithRelatedObjectsFunc(f RelatedObjectsFunc) {
+	c.relatedObjectsFunc = f
 }
 
 func (c *StatusSyncer) Run(ctx context.Context, workers int) {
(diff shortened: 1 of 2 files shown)


squeed:
@deads2k can I get a review please? this would help us migrate CNO to library-go tooling.

pkg/operator/status/status_controller.go
```diff
@@ -149,7 +158,16 @@ func (c StatusSyncer) Sync(ctx context.Context, syncCtx factory.SyncContext) err
 		return nil
 	}
 
-	clusterOperatorObj.Status.RelatedObjects = c.relatedObjects
+	if c.relatedObjectsFunc != nil {
+		ro := c.relatedObjectsFunc()
+		if ro != nil {
+			clusterOperatorObj.Status.RelatedObjects = ro
+		} else { // related objects unknown, carry over existing.
```
mfojtik, line 165: i would still append the relatedObjects from constructor? it is not clear from the NewClusterOperatorStatusController() that when WithRelatedObjectsFunc() is used the relatedObjects passed to it are dropped.

squeed: good point; updated.

squeed:
@mfojtik updated, thanks!

mfojtik:
/lgtm

@sttts second look/approve?

pkg/operator/status/status_controller.go
```diff
@@ -149,7 +164,18 @@ func (c StatusSyncer) Sync(ctx context.Context, syncCtx factory.SyncContext) err
...
 
-	clusterOperatorObj.Status.RelatedObjects = c.relatedObjects
+	if c.relatedObjectsFunc != nil {
+		ro := c.relatedObjectsFunc()
+		if ro != nil {
+			// append any statically-defined related objects
+			ro = append(ro, c.relatedObjects...)
+			clusterOperatorObj.Status.RelatedObjects = ro
+		} else { // related objects (temporarily) unknown, carry over existing.
+			clusterOperatorObj.Status.RelatedObjects = originalClusterOperatorObj.Status.RelatedObjects
```
sttts, line 174: this has smell. Why not `c.relatedObjects`? You seem to misuse a nil return value as an error case.

pkg/operator/status/status_controller.go
```diff
@@ -78,6 +79,20 @@ func NewClusterOperatorStatusController(
 	}
 }
 
+// WithRelatedObjectsFunc allows the set of related objects to be dynamically
+// determined.
+//
+// If the function returns nil, the set of related objects is carried over
+// from the existing ClusterOperator. This is useful if the set of related
+// objects is not yet known, perhaps because a control loop hasn't yet
+// executed. To set no related objects, return empty list.
```
sttts, line 88: strange signature for this behaviour. I would expect at least c.relatedObject to be applied, always.

squeed: Right, but then we could flap RelatedObjects.

Here's the motivating example: on restart, the CNO may wish to set a Degraded status condition immediately, but not wish to touch RelatedObjects.

Returning nil signifies "have some inertia - don't touch anything". One could, I suppose, ensure that the static `c.relatedObjects` is in the ClusterOperator object.

pkg/operator/status/status_controller.go
squeed: It's not an error case, it's "I don't know yet". But you're right, that's kind of overloading.

I can change the signature to be more explicit.

squeed:
@sttts updated, thanks for the review. PTAL.

pkg/operator/status/status_controller.go
```diff
@@ -33,9 +33,12 @@ type VersionGetter interface {
 	VersionChangedChannel() <-chan struct{}
 }
 
+type RelatedObjectsFunc func() (isset bool, objs []configv1.ObjectReference)
```
sttts, line 36: go doc

PR #943, opened 2020-11-20
revert of #926: provides DefaultClientDialer that returns a network dialer with default options sets

p0lyn0mial:
https://github.com/openshift/library-go/pull/943 was meant as a temporary solution for detecting broken persistent connections.

This mechanism was replaced by HTTP/2 connection health check which is now part of `http.Transport` and will be applied by default to all clients created via `client-go` library.


sttts:
better call it "revert of ..."

p0lyn0mial:
> better call it "revert of ..."

done.

sttts:
/hold

until 1.20 prebase branches go merged into master

PR #944, opened 2020-11-23
improve DefaultClientDialContext

p0lyn0mial:
wrapping the DialContext method and setting the socket options on a copy of the file descriptor is not deterministic.
as per documentation attempting to change properties of the original using this duplicate
may or may not have the desired effect.

instead we should set the options on the original file descriptor.

### DIFF
--- a/pkg/network/dialer_linux.go
+++ b/pkg/network/dialer_linux.go
@@ -3,41 +3,38 @@
 package network
 
 import (
-	"context"
 	"net"
 	"os"
 	"syscall"
 	"time"
 
 	"golang.org/x/sys/unix"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 func dialerWithDefaultOptions() DialContext {
 	nd := &net.Dialer{
 		// TCP_USER_TIMEOUT does affect the behaviour of connect() which is controlled by this field so we set it to the same value
 		Timeout: 25 * time.Second,
-	}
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
I performed the manual tests described https://github.com/openshift/library-go/pull/926#issuecomment-717888348 and https://github.com/openshift/library-go/pull/926#issuecomment-718694746 and haven't found any issues. 

I got the same results.

pkg/network/dialer_linux.go
```diff
@@ -16,28 +15,18 @@ func dialerWithDefaultOptions() DialContext {
 	nd := &net.Dialer{
 		// TCP_USER_TIMEOUT does affect the behaviour of connect() which is controlled by this field so we set it to the same value
 		Timeout: 25 * time.Second,
+		// KeepAlive must to be set to a negative value to stop std library from applying the default values
+		// by doing so we ensure that the options we are interested in won't be overwritten
+		KeepAlive: time.Duration(-1),
+		Control: func(network, address string, con syscall.RawConn) error {
+			var err error
+			err = con.Control(func(fd uintptr) {
+				err = setDefaultSocketOptions(int(fd))
```
sttts, line 24: which err will survive?

p0lyn0mial: /hold

it should have been `return setDefaultSocketOptions(int(fd))`

p0lyn0mial: /hold cancel

PR #946, opened 2020-11-23
Bug 1896446: Add git configuration to the environment default

adambkaplan:
- Adds capability to add git configuration values to the environment
default. This is one of the following:
  - The file or option specified by the GIT_CONFIG environment variable.
  - The repository-specific git configuration ($location/.git/config)
- Document that AddLocalConfig and AddGlobalConfig can fail if the
GIT_CONFIG environment variable is set.

### DIFF
--- a/pkg/git/repository.go
+++ b/pkg/git/repository.go
@@ -37,6 +37,7 @@ type Repository interface {
 	Add(dir string, spec string) error
 	Commit(dir string, message string) error
 	AddRemote(dir string, name, url string) error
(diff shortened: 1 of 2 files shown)


adambkaplan:
/assign @bparees 

This is the root of why git clone is failing behind a proxy. Git doesn't know which file to write to because we have `GIT_CONFIG` set if user provides a source secret.

adambkaplan:
PR where this is used:

https://github.com/openshift/builder/pull/189

PR #949, opened 2020-11-27
tests: provides WaitForAPIServerToStabilizeOnTheSameRevision waits until all API Servers are running at the same revision

p0lyn0mial:
Provides a convenience function for ensuring that APIs have converged to the same revision. The function is meant to be used in tests that require rolling out a new version (i.e. config change).

The API Servers must stay on the same revision for at least `waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval`.
Mainly because of the difference between the propagation time of triggering a new release and the actual roll-out.

Observations:
- rolling out a new version is not instant you need to account for a propagation time (~1/2 minutes)
-  for some API servers (KAS) rolling out a new version can take ~10 minutes

### DIFF
--- a/test/library/apiserver/apiserver.go
+++ b/test/library/apiserver/apiserver.go
@@ -1,0 +1,133 @@
+package apiserver
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+var (
+	// the following parameters specify for how long apis must
+	// stay on the same revision to be considered stable
+	waitForAPIRevisionSuccessThreshold = 3
+	waitForAPIRevisionSuccessInterval  = 1 * time.Minute
+
+	// the following parameters specify max timeout after which
+	// apis are considered to not converged
+	waitForAPIRevisionPollInterval = 30 * time.Second
+	waitForAPIRevisionTimeout      = 15 * time.Minute
+)
+
+// WaitForAPIServerToStabilizeOnTheSameRevision waits until all API Servers are running at the same revision.
+// The API Servers must stay on the same revision for at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval.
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
this function was introduced in https://github.com/openshift/cluster-kube-apiserver-operator/pull/870

moving here so it can be reused

test/library/apiserver/apiserver.go
```diff
@@ -0,0 +1,123 @@
...
+// The API Servers must stay on the same revision for at least waitForAPIRevisionSuccessThreshold * waitForAPIRevisionSuccessInterval.
+// Mainly because of the difference between the propagation time of triggering a new release and the actual roll-out.
+//
+// Observations:
+//  rolling out a new version is not instant you need to account for a propagation time (~1/2 minutes)
+//  for some API servers (KAS) rolling out a new version can take ~10 minutes
+func WaitForAPIServerToStabilizeOnTheSameRevision(
+	t *testing.T,
+	podClient corev1client.PodInterface,
+	waitForAPIRevisionSuccessThreshold int,
```
sttts, line 26: do we need all these parameters in every test? Isn't there a good default?

p0lyn0mial: sure we could try with what we have atm. Will work for KAS and OAS

```
waitForAPIRevisionSuccessThreshold = 3
waitForAPIRevisionSuccessInterval  = 1 * time.Minute

waitForAPIRevisionPollInterval = 30 * time.Second
waitForAPIRevisionTimeout      = 15 * time.Minute
```

test/library/apiserver/apiserver.go
```diff
@@ -0,0 +1,123 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+// WaitForAPIServerToStabilizeOnTheSameRevision waits until all API Servers are running at the same revision.
```
sttts, line 16: what does "all API servers" mean?

p0lyn0mial: all pods with `apiserver=true` label specified.

p0lyn0mial: updated docs.

sttts: .... given by the namespaced pod client

p0lyn0mial: done

PR #950, opened 2020-11-30
encryption controllers: reduce GRs

p0lyn0mial:
it turns out that the list of resources to encrypt provided to the encryption controllers is not static and can change over time.
this PR makes the necessary changes to support removing unused GR from the encryption config.

### DIFF
--- a/pkg/operator/encryption/statemachine/transition.go
+++ b/pkg/operator/encryption/statemachine/transition.go
@@ -91,6 +91,24 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
 		}
 	}
 
+	// remove unused GRs from the desired encryption configuration
+	// toBeEncryptedGRs is not static and can change over time
+	// here we are removing resources that his operator doesn't manage anymore
+	for actualGR := range desiredEncryptionState {
+		found := false
+		for _, desiredGR := range toBeEncryptedGRs {
+			if actualGR == desiredGR {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			delete(desiredEncryptionState, actualGR)
+			klog.V(4).Infof("removed %s from the encryption config as this operator doesn't manage this GR anymore", actualGR.String())
+		}
+	}
+
 	//
 	// STEP 1: without secrets, wait for the key controller to create one
 	//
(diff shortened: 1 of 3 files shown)


pkg/operator/encryption/statemachine/transition.go
```diff
@@ -171,7 +171,26 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
 	}
 
 	//
-	// STEP 4: with consistent read-keys and write-keys, remove every read-key other than the write-key and one last read key.
+	// STEP 4: with consistent keys remove a GR that this operator no longer manages
+	//
+	// TODO: what if len(toBeEncryptedGRs)  == 0
```
p0lyn0mial, line 176: do we need an additional check?

pkg/operator/encryption/statemachine/transition.go
```diff
@@ -171,7 +171,26 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
 	}
 
 	//
-	// STEP 4: with consistent read-keys and write-keys, remove every read-key other than the write-key and one last read key.
+	// STEP 4: with consistent keys remove a GR that this operator no longer manages
```
p0lyn0mial, line 174: if we add it after the `STEP 5` then we can only reduce a list once the keys are stable and the migration was successful

sttts: I wonder why this does not happen much further up behind the loop https://github.com/openshift/library-go/pull/950/files#diff-c4e81966536385a51d736f8985effa981b911fdbe4c32e59b1bffa64852c80ddR86.

pkg/operator/encryption/statemachine/transition.go
sttts: I don't think we need that. In that case we remove the whole encryption mechanism.

pkg/operator/encryption/statemachine/transition.go
```diff
@@ -91,6 +91,24 @@ func getDesiredEncryptionState(oldEncryptionConfig *apiserverconfigv1.Encryption
...
+		for _, desiredGR := range toBeEncryptedGRs {
+			if actualGR == desiredGR {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			delete(desiredEncryptionState, actualGR)
+			klog.V(4).Infof("removed  %s from the encryption config as this operator doesn't manage this GR anymore", actualGR.String())
```
sttts, line 108: nit: one space too much

sttts:
/approve
Lgtm. Only one nit left.

PR #951, opened 2020-11-30
template: remove usage of legacy api

mfojtik:
The `/oapi` endpoint is removed since 4.x so it is time to remove support for non-grouped API resource even in templates.

### DIFF
--- a/pkg/template/templateprocessing/template.go
+++ b/pkg/template/templateprocessing/template.go
@@ -13,7 +13,6 @@ import (
 
 	appsv1 "github.com/openshift/api/apps/v1"
 	templatev1 "github.com/openshift/api/template/v1"
-	"github.com/openshift/library-go/pkg/legacyapi/legacygroupification"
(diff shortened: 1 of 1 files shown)


pkg/template/templateprocessing/template.go
```diff
@@ -94,10 +93,12 @@ func (p *Processor) Process(template *templatev1.Template) field.ErrorList {
...
 
-		// this changes oapi GVKs to groupified GVKs so they can be submitted to modern, aggregated servers
-		// It is done after substitution in case someone substitutes a kind.
+		// This can only happen when currObj is nil, but better be safe than sorry.
+		if newItem == nil {
+			continue
+		}
+
 		gvk := currObj.GetObjectKind().GroupVersionKind()
-		legacygroupification.OAPIToGroupifiedGVK(&gvk)
```
soltysh, line 100: We want to keep this as is still.

soltysh (changes requested):
/hold

Issue #953, opened 2020-11-30
ApplyValidatingWebhookConfiguration does not work


jsafrane:
There are couple of issues in ApplyValidatingWebhookConfiguration introduced in https://github.com/openshift/library-go/pull/902 

1. When used in the `StaticController`, `ApplyDirectly` fails with:
    ```E1130 17:41:41.847063   99786 base_controller.go:250] "StaticResourceController" controller failed to sync "key", err: 
    "webhook_configuration.yaml" (string): cannot decode "webhook_configuration.yaml": no kind "ValidatingWebhookConfiguration" 
    is registered for version "admissionregistration.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
    ```
    This can be fixed by registering admission/v1 API around these parts: https://github.com/openshift/library-go/blob/bd449d1e2e25157fc0e304e0d5fcd239a2a7ff36/pkg/operator/resource/resourceapply/generic.go#L32-L33

---

2. `ApplyValidatingWebhookConfiguration` itself fails on updating a webhook config with:
    ```
    validatingwebhookconfigurations.admissionregistration.k8s.io "snapshot.storage.k8s.io" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
    ```
    It fails because the function tries to call `Update()` with `required`:
    https://github.com/openshift/library-go/blob/bd449d1e2e25157fc0e304e0d5fcd239a2a7ff36/pkg/operator/resource/resourceapply/admissions.go#L43
    All the other `Apply*` functions merge `required` and `existingCopy` to make sure all the metadata is there...

---

3. Change detection in `ApplyValidatingWebhookConfiguration` does not work at all. In OpenShift, it's common that an admission plugin injects the actual `caBundle` into `webhooks.clientConfig`. Therefore, the `required.webhooks` and `existingCopy.webhooks` will never be the same and the function will *always* update the `ValidatingWebhookConfiguration` object.
    https://github.com/openshift/library-go/blob/bd449d1e2e25157fc0e304e0d5fcd239a2a7ff36/pkg/operator/resource/resourceapply/admissions.go#L34-L37

@qiujian16, is the `ApplyValidatingWebhookConfiguration` used anywhere? How did you test it?

qiujian16:
Sorry, the code is moved from here https://github.com/open-cluster-management/registration-operator/blob/master/pkg/helpers/helpers.go#L202-L227 where I miss some part.... 

[...]

PR #954, opened 2020-12-01
Fix webhook apply bug

qiujian16:
fix this https://github.com/openshift/library-go/issues/953

### DIFF
--- a/pkg/operator/resource/resourceapply/admissions.go
+++ b/pkg/operator/resource/resourceapply/admissions.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	admissionv1 "k8s.io/api/admissionregistration/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	admissionclientv1 "k8s.io/client-go/kubernetes/typed/admissionregistration/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/resource/resourceapply/admissions.go
```diff
@@ -71,7 +72,8 @@ func ApplyMutatingWebhookConfiguration(client admissionclientv1.MutatingWebhookC
 		klog.Infof("ValidatingWebhookConfiguration %q changes: %v", required.Name, JSONPatchNoError(existing, required))
 	}
 
-	actual, err := client.MutatingWebhookConfigurations().Update(context.TODO(), required, metav1.UpdateOptions{})
+	existingCopy.Webhooks = required.Webhooks
```
jsafrane, line 75: This will clear injected CA bundles. Please check with https://github.com/openshift/library-go/pull/836, which does it IMO a bit better, still, it does not solve Generation tracking properly.

qiujian16: I do not think we can make it in applyDirectly func if we use generation to track. And it won't help if we want to change the spec. In this case, should we have a special annotation to hash the applied spec as here https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L113-L117?

qiujian16: @jsafrane I changed to use generation similar as #836, do you think it makes more sense? I think the generation tracking has to be done on higher layer in this case.

qiujian16:
close this since it has been fixed

PR #959, opened 2020-12-03
Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of
https://github.com/openshift/enhancements#482 and does not change behavior.
Initially, all library-go manifests are included in the
single-node-developer cluster profile. Follow-on PRs may exclude any of
these that are not needed in the profile.


cfergeau:
This was fixed by commit 0b5a0f5.

PR #960, opened 2020-12-03
factory: use the cron schedule to run periodic resyncs

mfojtik:
This removes the "special" case "ResyncEvery()" and use the cron scheduler to execute periodic resyncs.
It also fixes a case when initial periodic jobs were not run and instead we waited the scheduled period to trigger initial sync (imagine you set "schedule ever 1h", then operator start and the first resync is done after 1h...)

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/robfig/cron"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
@@ -33,8 +32,7 @@ type baseController struct {
 	sync               func(ctx context.Context, controllerContext SyncContext) error
 	syncContext        SyncContext
 	syncDegradedClient operatorv1helpers.OperatorClient
-	resyncEvery        time.Duration
-	resyncSchedules    []cron.Schedule
+	resyncSchedules    []string
 	postStartHooks     []PostStartHook
 	cacheSyncTimeout   time.Duration
 }
@@ -63,6 +61,9 @@ func (s *scheduledJob) Run() {
 }
 
 func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheSyncs ...cache.InformerSynced) error {
+	if len(cacheSyncs) == 0 {
+		return nil
+	}
 	klog.Infof("Waiting for caches to sync for %s", controllerName)
 
 	if !cache.WaitForCacheSync(stopCh, cacheSyncs...) {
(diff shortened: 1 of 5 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -63,6 +61,9 @@ func (s *scheduledJob) Run() {
 }
 
 func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheSyncs ...cache.InformerSynced) error {
+	if len(cacheSyncs) == 0 {
```
mfojtik, line 64: removes log spam if there are no informers and we only schedule periodically

pkg/controller/factory/base_controller.go
```diff
@@ -120,35 +121,37 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
-		workerWg.Add(1)
-		go func() {
-			defer workerWg.Done()
-			c.runPeriodicalResync(ctx, c.resyncEvery)
-		}()
+		// always schedule initial run, regardless of the schedule
+		go scheduledJob.Run()
+	} else {
+		// Make this warning, usually we want every controller to be able to resync on its own without depending on informers.
+		klog.Warningf("Controller %s will never resync without informer change, consider adding fixed resync schedule", c.Name())
```
mfojtik, line 139: every controller should have manual resync, unless you know what you are doing

pkg/controller/factory/base_controller.go
```diff
@@ -120,35 +121,37 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 	// if scheduled run is requested, run the cron scheduler
 	if c.resyncSchedules != nil {
 		scheduler := cron.New()
+		scheduledJob := newScheduledJob(c.Name(), c.syncContext.Queue())
 		for _, s := range c.resyncSchedules {
-			scheduler.Schedule(s, newScheduledJob(c.name, c.syncContext.Queue()))
+			schedule, err := cron.ParseStandard(s)
+			if err != nil {
+				panic(fmt.Errorf("invalid controller schedule %q: %v", s, err))
```
sttts, line 128: why panic?

pkg/controller/factory/base_controller.go
```diff
@@ -120,35 +121,37 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
-
-	// runPeriodicalResync is independent from queue
-	if c.resyncEvery > 0 {
-		workerWg.Add(1)
-		go func() {
-			defer workerWg.Done()
-			c.runPeriodicalResync(ctx, c.resyncEvery)
-		}()
+		// always schedule initial run, regardless of the schedule
+		go scheduledJob.Run()
```
sttts, line 136: why not queue directly?

pkg/controller/factory/base_controller.go
mfojtik: it is programmer error (you specified wrong schedule format), this is not retriable and controller should panic

pkg/controller/factory/base_controller.go
mfojtik: scheduledJob.Run() does queue, I found this more cleaner for the intent

PR #965, opened 2020-12-07
Bug 1905119: csi: allow for custom string replaces in controller asset

staebler:
To support AWS C2S, the AWS EBS CSI driver needs to be able to modify the controller deployment that it creates based upon whether there is a custom CA bundle to use when accessing the AWS API.


bertinatto:
There's [another bug](https://bugzilla.redhat.com/show_bug.cgi?id=1898045) where we also need to modify the controller Deployment (i.e., by adding some annotations to it), but string replacements aren't a good fit for that.

I'm working on a PR to add a hook to modify the Deployment object instead: https://github.com/openshift/aws-ebs-csi-driver-operator/pull/104. I think we could use that to add the custom CA bundle. Any thoughts?

CC @jsafrane

staebler:
This is obsolete.
/close

PR #966, opened 2020-12-08
resourceapply: fix reporting modified on EnsureMeta

stlaz:
EnsureMeta would report modified=true even if the annotation marked
for removal was already not present in the existing object

should be proven to work by https://github.com/openshift/service-ca-operator/pull/109

### DIFF
--- a/pkg/operator/resource/resourcemerge/object_merger.go
+++ b/pkg/operator/resource/resourcemerge/object_merger.go
@@ -140,15 +140,27 @@ func MergeMap(modified *bool, existing *map[string]string, required map[string]s
 		*existing = map[string]string{}
 	}
 	for k, v := range required {
-		if existingV, ok := (*existing)[k]; !ok || v != existingV {
-			*modified = true
-			// if "required" map contains a key with "-" as suffix, remove that
-			// key from the existing map instead of replacing the value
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourcemerge/object_merger.go
```diff
@@ -140,15 +140,27 @@ func MergeMap(modified *bool, existing *map[string]string, required map[string]s
...
+
+		if existingV, ok := (*existing)[actualKey]; removeKey {
+			if !ok {
+				continue
 			}
+			// value found -> it should be removed
+			delete(*existing, actualKey)
+			*modified = true
+
+		} else if !ok || v != existingV {
```
p0lyn0mial, line 161: if we don't need to remove the key `removeKey=false` then we need to compare only values. What if the map contains complex types? Is it always used to compare `labels` and `annotations`?

I would change it to:
```
else {
*modified =. true
(*existing)[actualKey] = v
}
```

p0lyn0mial: > What if the map contains complex types?

the map can't contain complex types, only `string`s:

`MergeMap(modified *bool, existing *map[string]string, required map[string]string)`

PR #968, opened 2020-12-09
csi: add hooks in both Node and Controller Services

bertinatto:
CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
@@ -2,6 +2,7 @@ package csidrivercontrollerservicecontroller
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -33,6 +34,9 @@ const (
 
 	infraConfigName = "cluster"
 )
+
+// DeploymentHookFunc is a hook function to modify the Deployment.
+type DeploymentHookFunc func(*appsv1.Deployment) error
 
 // CSIDriverControllerServiceController is a controller that deploys a CSI Controller Service to a given namespace.
 //
@@ -77,8 +81,13 @@ type CSIDriverControllerServiceController struct {
 	operatorClient v1helpers.OperatorClient
 	kubeClient     kubernetes.Interface
 	deployInformer appsinformersv1.DeploymentInformer
-	// Optional, used by CSI drivers to tag volumes and snapshots
+	// Optional config informer used to get cluster information.
 	optionalConfigInformer configinformers.SharedInformerFactory
+	// Optional hook functions to modify the Deployment.
+	// If one of these functions returns an error, the sync
+	// fails indicating the ordinal position of the failed function.
+	// Also, in that scenario the Degraded status is set to True.
+	optionalDeploymentHooks []DeploymentHookFunc
 }
 
 func NewCSIDriverControllerServiceController(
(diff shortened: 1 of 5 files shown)


bertinatto:
/hold

This is a WIP design.

jsafrane:
Should there be a similar hook for DaemonSet?

bertinatto:
> Should there be a similar hook for DaemonSet?

Added, but no operator is using it ATM

bertinatto:
/hold cancel

I've tested this with openshift/aws-ebs-csi-driver-operator#104 and it works.

bertinatto:
@deads2k could you take a look, please?

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -89,14 +92,16 @@ func NewCSIDriverControllerServiceController(
 	deployInformer appsinformersv1.DeploymentInformer,
 	optionalConfigInformer configinformers.SharedInformerFactory,
 	recorder events.Recorder,
+	optionalDeploymentHooks ...func(*appsv1.Deployment) error,
```
jsafrane, line 95: Would it make sense to either give owner of the controller a way how to poke the controller for resync (i.e. the owner has informer on secrets / config maps / whatnot and is able to enqueue "fake" event into CSIDriverControllerServiceController' queue) or pass informers to watch into the `NewCSIDriverControllerServiceController`?

I can see there is `ResyncEvery(time.Minute)` below, so this PR will work even without it.

bertinatto: Would it make sense to address this in another PR?

Also, could you clarify a use case for poking the controller for resync?

jsafrane: Ack

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller_test.go
```diff
@@ -422,33 +469,6 @@ func TestSync(t *testing.T) {
 					withFalseConditions(conditionProgressing)),
 			},
 		},
-		{
-			// Deployment is fully deployed and its status is synced to CR
```
deads2k, line 426: this rubs me the wrong way.  Why did it need to be removed?

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -63,6 +64,8 @@ type CSIDriverNodeServiceController struct {
 	operatorClient v1helpers.OperatorClient
 	kubeClient     kubernetes.Interface
 	dsInformer     appsinformersv1.DaemonSetInformer
+	// Optional hook functions to modify the DaemonSet.
+	optionalDaemonSetHooks []func(*appsv1.DaemonSet) error
```
deads2k, line 68: typed functions please. Also, make clear what happens on errors and where responsibility for identifying *which* function is failing lies in the error.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -116,6 +121,13 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
 	manifest := replacePlaceholders(c.manifest, opSpec)
 	required := resourceread.ReadDaemonSetV1OrDie(manifest)
 
+	for _, fn := range c.optionalDaemonSetHooks {
+		err := fn(required)
+		if err != nil {
+			return fmt.Errorf("error running hook function: %w", err)
```
deads2k, line 127: you can at least give me an ordinal (`i`)

deads2k:
I'm curious what you're trying to replace/mutate.  Some sorts of mutations are easier while you still have bytes (this is why we have the `replacePlaceholders`).

bertinatto (commented):
> I'm curious what you're trying to replace/mutate. Some sorts of mutations are easier while you still have bytes (this is why we have the replacePlaceholders).

As more CSI operators use these control loops, we want to give them freedom to mutate the deployment/daemonset without adding placeholders to library-go, specially if the mutation is specific to that operator, like this one: https://github.com/openshift/aws-ebs-csi-driver-operator/pull/102

Most operators won't need this, and the default mutation approach should be the string replacements, but I figured for some situations mutating the object is better.

Does that make sense? Do you think another controller would be a better approach?

deads2k:
> Does that make sense?

I understand the desire.  I'd be somewhat interested in what you're combining because I've regretted it in the past.  But I'm ok adding the option

PR #970, opened 2020-12-14
Bug 1887392: allow configuring authn/z caches TTLs

stlaz:
This should contribute to reducing traffic inside a cluster since not every metrics request to our operators (once every 30s per operator) will need to pass an authn/z check (tokenreview/SAR).

/assign @sttts 

/hold
contains fake bump

### DIFF
--- a/pkg/config/serving/server.go
+++ b/pkg/config/serving/server.go
@@ -39,6 +39,8 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
 	if !authenticationConfig.Disabled {
 		authenticationOptions := genericapiserveroptions.NewDelegatingAuthenticationOptions()
 		authenticationOptions.RemoteKubeConfigFile = kubeConfigFile
+		// the platform generally uses 30s for /metrics scraping, avoid API request for every other /metrics request to the component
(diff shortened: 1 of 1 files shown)


stlaz:
/retest
google was out

pkg/config/serving/server.go
```diff
@@ -58,6 +59,7 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
 	if !authorizationConfig.Disabled {
 		authorizationOptions := genericapiserveroptions.NewDelegatingAuthorizationOptions()
 		authorizationOptions.RemoteKubeConfigFile = kubeConfigFile
+		authorizationOptions.AllowCacheTTL = authorizationConfig.AllowCacheTTL.Duration
```
sttts, line 62: what is the default? None?

stlaz: it's 10s as set by the `NewDelegatingAuthenticationOptions`/`NewDelegatingAuthorizationOptions` functions

stlaz:
depends on https://github.com/openshift/api/pull/832

stlaz:
/hold cancel
no longer needs API change, the timeout is hardcoded

PR #971, opened 2020-12-15
Add unit test to expose undeterministic merge with empty string key

csrwng:
Exposes undeterministic merge introduced in mergo v0.3.9

To test locally...
`go get github.com/imdario/mergo@v0.3.9` and run this test


pkg/operator/configobserver/config_observer_controller_test.go
```diff
@@ -112,7 +115,7 @@ func TestSyncStatus(t *testing.T) {
 
 			expectError: false,
 			expectedObservedConfig: &unstructured.Unstructured{Object: map[string]interface{}{
-				"foo": "one",
+				"foo": map[string]interface{}{"one": "1", "two": ""},
```
deads2k, line 118: so what happens on the latest mergo?

csrwng: on 0.3.11 it passes

csrwng: so presumably it's ok :) ?

PR #973, opened 2020-12-16
DockerImageReference String() fix

kevinrizza:
Currently, the internal DockerImageReference type does not return the
correct response when only the image name and not Registry field is set.
It should return the Registry and Namespace even when name is not set

This commit introduces a fix for the Exact() method called by String()
to properly return the Registry + Namespace

### DIFF
--- a/pkg/image/reference/reference.go
+++ b/pkg/image/reference/reference.go
@@ -205,10 +205,6 @@ func (r DockerImageReference) NameString() string {
 
 // Exact returns a string representation of the set fields on the DockerImageReference
 func (r DockerImageReference) Exact() string {
-	name := r.NameString()
-	if len(name) == 0 {
-		return name
-	}
 	s := r.Registry
 	if len(s) > 0 {
 		s += "/"
@@ -217,6 +213,15 @@ func (r DockerImageReference) Exact() string {
 	if len(r.Namespace) != 0 {
 		s += r.Namespace + "/"
 	}
+
+	name := r.NameString()
+	if len(name) == 0 {
+		if len(r.Registry) == 0 {
+			return ""
+		}
+		return s
+	}
+
 	return s + name
 }
 
--- a/pkg/image/reference/reference_test.go
+++ b/pkg/image/reference/reference_test.go
@@ -8,84 +8,98 @@ import (
 
 func TestParse(t *testing.T) {
 	testCases := []struct {
-		From                               string
-		Registry, Namespace, Name, Tag, ID string
-		Err                                bool
+		From                                       string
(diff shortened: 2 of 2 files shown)


ecordell:
/lgtm

(symbolic)

kevinrizza:
@sttts @smarterclayton Can I get a quick review on this? It's needed for this oc bz:

https://github.com/openshift/oc/pull/673

deads2k:
Let's get fake bumps in openshift/openshift-apiserver and openshift/openshift-controller-manager and openshift/image-registry with this change and then get one of the image guys to weigh in.  My memory is that the image code here is used in those three places.

ecordell:
https://github.com/openshift/image-registry/pull/260
https://github.com/openshift/openshift-apiserver/pull/172
https://github.com/openshift/openshift-controller-manager/pull/155

Waiting on CI

dmage:
/hold

Name should always be set. `quay.io` is a shorthand for `docker.io/library/quay.io`, i.e. `quay.io` in this case should be placed into `Name`.

ecordell:
@dmage This PR didn't change that behavior, though an earlier one did. This was originally done for "bare" parsing of a registry name for re-homing a collection of images under a new registry (common in disconnected mirroring).

To me that seems like the right assumption to make - if we always assume `quay.io` is really `docker.io/library/quay.io` then we have no option to force the parser to read `quay.io` as `quay.io`. 

But if `quay.io` is incorrectly assumed to be `quay.io` when it should be `docker.io/library/quay.io`, you then have the option to fully specify the name as an escape hatch.

(not the mention all the caveats around configurable default registry names).

dmage:
@ecordell can you elaborate on re-homing?

After this change, output of String() is not always understandable by Parse(). This is not desirable behavior. And it is incompatible with Docker/skopeo, please add one more test:

```
From: "rocket.chat",
String: "rocket.chat"
```

This image reference should be parseable and OpenShift should understand it as `docker.io/library/rocket.chat` (assuming `docker.io` is used as a default registry).

pkg/image/reference/reference.go
```diff
@@ -217,6 +213,15 @@ func (r DockerImageReference) Exact() string {
 	if len(r.Namespace) != 0 {
 		s += r.Namespace + "/"
```
smarterclayton, line 214: If name is not set, appending this slash is also not correct.

however, to Oleg’s point there should never be a scenario where name is empty but namespace is, because in that case name should be set to namespace.  And likewise we can’t trust only registry because some registry values can be valid names (localhost:8080 is a valid name + tag)

PR #975, opened 2020-12-22
Bug 1896977: add library function for host name validation

candita:
Router host name validation does not match between the apiserver validation and the route admission validation. Add the function to the library so it can be shared between the two processes.

Use a k8s apimachinery host name validation that reports the issue properly, as well as enforcing host name must have at least two labels (new), with each label less than 63 chars, and total length less than 253 chars. Trailing dots are allowed (new). The old behavior may be retained for customers who need it, by using a route annotation set to "true".

### DIFF
--- a/pkg/route/routeapihelpers/routeapihelpers.go
+++ b/pkg/route/routeapihelpers/routeapihelpers.go
@@ -7,6 +7,8 @@ import (
 
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
+	kvalidation "k8s.io/apimachinery/pkg/util/validation"
+	field "k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 // IngressURI calculates an admitted ingress URI.
@@ -41,4 +43,31 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
 		return nil, nil, fmt.Errorf("no admitted ingress for route %s in namespace %s", route.ObjectMeta.Name, route.ObjectMeta.Namespace)
 	}
 	return nil, nil, fmt.Errorf("no ingress for host %s in route %s in namespace %s", host, route.ObjectMeta.Name, route.ObjectMeta.Namespace)
+}
+
+// ValidateHost checks that a route's host name satisfies DNS requirements, with
+// the assumption that the caller has already checked for an empty host name.
+// Unless the allowNonCompliant annotation is set to true, host name must have
+// at least two labels, with each label no more than 63 characters from the set of
+// alphanumeric characters, '-' or '.', and must start and end with an alphanumeric
+// character. A trailing dot is allowed. The total host name length must be no more
+// than 253 characters.
+// If allowNonCompliant is set to true, it uses a smaller set of conditions from
+// IsDNS1123Subdomain, e.g. character set as described above, and total host name
+// length must be no more than 253 characters.
+func ValidateHost(host string, allowNonCompliant string, hostPath *field.Path) field.ErrorList {
+	result := field.ErrorList{}
+
+	if allowNonCompliant == "true" {
+		errs := kvalidation.IsDNS1123Subdomain(host)
+		if len(errs) != 0 {
+			result = append(result, field.Invalid(hostPath, host, fmt.Sprintf("host must conform to DNS naming conventions: %v", errs)))
+		}
+	} else {
+		errs := kvalidation.IsFullyQualifiedDomainName(hostPath, host)
+		if len(errs) != 0 {
+			result = append(result, field.Invalid(hostPath, host, fmt.Sprintf("host must conform to DNS 1123 naming conventions: %v", errs)))
+		}
+	}
+	return result
 }
--- a/pkg/route/routeapihelpers/routeapihelpers_test.go
+++ b/pkg/route/routeapihelpers/routeapihelpers_test.go
@@ -9,6 +9,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	field "k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 func TestIngressURI(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/route/routeapihelpers/routeapihelpers_test.go
```diff
@@ -299,3 +299,67 @@ func TestIngressURI(t *testing.T) {
...
+			name:              "Non-DNS-compliant host with blank override annotation",
+			host:              "host",
+			allowNonCompliant: "",
+			expectedErrors:    1,
+		},
+		{
+			name:           "Non-DNS-compliant host without override annotation",
+			host:           "host",
+			expectedErrors: 1,
+		},
```
Miciah, line 328: These two test cases have identical inputs.

pkg/route/routeapihelpers/routeapihelpers_test.go
```diff
@@ -299,3 +299,67 @@ func TestIngressURI(t *testing.T) {
...
+			host:              "**",
+			allowNonCompliant: "",
+			expectedErrors:    1,
+		},
+	}
+
+	for _, tc := range tests {
+		errs := ValidateHost(tc.host, tc.allowNonCompliant)
+		if len(errs) != tc.expectedErrors {
+			t.Errorf("Test case %s expected %d error(s), got %d. %v", tc.name, tc.expectedErrors, len(errs), errs)
```
Miciah, line 362:
```suggestion
			t.Errorf("Test case %q expected %d error(s), got %d: %v", tc.name, tc.expectedErrors, len(errs), errs)
```

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,29 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
 	}
 	return nil, nil, fmt.Errorf("no ingress for host %s in route %s in namespace %s", host, route.ObjectMeta.Name, route.ObjectMeta.Namespace)
 }
+
+// ValidateHost checks that a route's host name satisfies DNS requirements.
+// Unless the allowNonCompliant annotation is set to true, host name must have
+// at least two labels, with each label less than 63 chars from a given character
+// set, and total length less than 253 chars.  Trailing dots are allowed.
+// If allowNonCompliant is set to true, only check that the host satisfies
```
Miciah, line 52:
```suggestion
// set, and total length less than 253 chars.  A trailing dot is allowed.
// If allowNonCompliant is set to true, only check that the host satisfies
```

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,29 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
...
+// basic DNS subdomain conditions, e.g. length and character set.
+func ValidateHost(host string, allowNonCompliant string) field.ErrorList {
+	result := field.ErrorList{}
+	specPath := field.NewPath("spec")
+
+	// host is not required but if it is set ensure it meets certain DNS requirements
+	if len(host) > 0 {
```
Miciah, line 59: I think it would make more sense for the caller to check if `spec.host` is empty because only the caller knows if empty `host` is valid in the caller's particular context.

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,29 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
...
+		if allowNonCompliant == "true" {
+			if len(kvalidation.IsDNS1123Subdomain(host)) != 0 {
+				result = append(result, field.Invalid(specPath.Child("host"), host, "host must conform to DNS naming conventions"))
```
Miciah, line 62: Is it intentional not to include `kvalidation.IsDNS1123Subdomain(host)` in the error message?  You do include `kvalidation.IsFullyQualifiedDomainName(specPath.Child("host"), host)` in the error message below, and it seems like `kvalidation.IsDNS1123Subdomain(host)` could have useful information.

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,29 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
...
+
+// ValidateHost checks that a route's host name satisfies DNS requirements.
+// Unless the allowNonCompliant annotation is set to true, host name must have
+// at least two labels, with each label less than 63 chars from a given character
+// set, and total length less than 253 chars.  Trailing dots are allowed.
+// If allowNonCompliant is set to true, only check that the host satisfies
+// basic DNS subdomain conditions, e.g. length and character set.
+func ValidateHost(host string, allowNonCompliant string) field.ErrorList {
+	result := field.ErrorList{}
+	specPath := field.NewPath("spec")
```
Miciah, line 56: It seems strange to construct the `field.Path` value and return a `field.ErrorList` value.  My inclination would be to make `ValidateHost` return a more general `error` value and leave it up to the caller to build the more specific `field.ErrorList` value.  However, `kvalidation.IsFullyQualifiedDomainName` needs a `field.Path` value, so maybe it would make more sense for `ValidateHost` both to have a `field.Path` parameter and a `field.ErrorList` return value like other validation functions.

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,29 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
 	}
 	return nil, nil, fmt.Errorf("no ingress for host %s in route %s in namespace %s", host, route.ObjectMeta.Name, route.ObjectMeta.Namespace)
 }
+
+// ValidateHost checks that a route's host name satisfies DNS requirements.
+// Unless the allowNonCompliant annotation is set to true, host name must have
+// at least two labels, with each label less than 63 chars from a given character
+// set, and total length less than 253 chars.  Trailing dots are allowed.
```
Miciah, line 51: Both limits are "less than or equal to", aren't they?  
https://github.com/kubernetes/kubernetes/blob/6715318ee738668db0ddb84415f6926208761b9b/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L231
https://github.com/kubernetes/kubernetes/blob/6715318ee738668db0ddb84415f6926208761b9b/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L210
```suggestion
// at least two labels, with each label no more than 63 chars from a given character
// set, and total length no more than 253 chars.  Trailing dots are allowed.
```

Miciah, line 51: What does "from a given character set" mean?  It needs to be alphanumeric (plus the hyphen) ASCII, right?

pkg/route/routeapihelpers/routeapihelpers_test.go
```diff
@@ -299,3 +299,67 @@ func TestIngressURI(t *testing.T) {
...
+		host              string
+		allowNonCompliant string
+		expectedErrors    int
+	}{
+		{
+			name:              "Non-DNS-compliant host with override annotation",
+			host:              "host",
+			allowNonCompliant: "true",
+			expectedErrors:    0,
+		},
```
Miciah, line 317: Could you add another test case or two with `allowNonCompliant: "true"` to show what it does reject?

pkg/route/routeapihelpers/routeapihelpers.go
candita: I'll remove it and add a comment.

pkg/route/routeapihelpers/routeapihelpers.go
candita: Sounds reasonable.

pkg/route/routeapihelpers/routeapihelpers.go
candita: As implemented by `IsDNS1123Subdomain`, the character set is `lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character`.  I'll add it to the comment.

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,30 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
...
+// If allowNonCompliant is set to true, it uses a smaller set of conditions from
+// IsDNS1123Subdomain, e.g. character set as described above, and total host name
+// length must be no more than 253 characters.
+func ValidateHost(host string, allowNonCompliant string, specPath *field.Path) field.ErrorList {
+	result := field.ErrorList{}
+
+	if allowNonCompliant == "true" {
+		errs := kvalidation.IsDNS1123Subdomain(host)
+		if len(errs) != 0 {
+			result = append(result, field.Invalid(specPath.Child("host"), host, fmt.Sprintf("host must conform to DNS naming conventions: %v", errs)))
```
Miciah, line 64: It would make sense for the caller to pass in the full path to the field that is being validated (i.e., `spec.host`).  Then `ValidateHost` would use that `*field.Path` value directly instead of using `specPath.Child("host")` here and below.

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -42,3 +44,30 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
 	}
 	return nil, nil, fmt.Errorf("no ingress for host %s in route %s in namespace %s", host, route.ObjectMeta.Name, route.ObjectMeta.Namespace)
 }
+
+// ValidateHost checks that a route's host name satisfies DNS requirements, with
+// the assumption that the caller has already checked for an empty host name.
+// Unless the allowNonCompliant annotation is set to true, host name must have
+// at least two labels, with each label no more than 63 characters from the set of
+// alphanumeric characters, '-' or '.', and must start and end with an alphanumeric
+// character. A trailing dot is allowed. The total host name length must be no more
```
Miciah, line 53: Sorry, I missed this before; the actual rule is that it must start with an alphabetic character and end with an alphanumeric character.

candita: I don't think you missed it, in https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L177 we allow it to start with alphanumeric, not alphabetic.

Miciah: Ah, you're right.  I was looking at 1035, which is more restrictive than 1123.

PR #976, opened 2021-01-07
csi: add configobserver controller and hook helpers

bertinatto:
This PR adds two new functionalities to the CSI pieces of library-go:

**csi: add configobserver to watch proxy info**

Cloud CSI drivers perform all sorts of operations using the cloud API.  However, that's not possible in network-restricted clusters, where requests need to go through a given proxy, unless they target the API server.

This patch adds a CSI Config Observer controller to observe the proxy resource and record the information in the operator CR. It also patches the CSI Controller Service Controller to read this information and pass it to the CSI containers that are part of the Deployment.

**csi: add hook that annotates deployment with secret's hash**

This is useful for when the secret changes. In this case we want the deployment to be re-deployed.

/assign @deads2k

### DIFF
--- a/pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
+++ b/pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
@@ -1,0 +1,81 @@
+package csiconfigobservercontroller
+
+import (
+	"strings"
+
+	"k8s.io/client-go/tools/cache"
+
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	configlistersv1 "github.com/openshift/client-go/config/listers/config/v1"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/configobserver/proxy"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// ProxyConfigPath returns the path for the observed proxy config. This is a
+// function to avoid exposing a slice that could potentially be appended.
+func ProxyConfigPath() []string {
+	return []string{"targetcsiconfig", "proxy"}
+}
+
+// Listers implement the configobserver.Listers interface.
+type Listers struct {
+	ProxyLister_ configlistersv1.ProxyLister
+
+	ResourceSync       resourcesynccontroller.ResourceSyncer
+	PreRunCachesSynced []cache.InformerSynced
+}
+
+func (l Listers) ProxyLister() configlistersv1.ProxyLister {
+	return l.ProxyLister_
+}
+
+func (l Listers) ResourceSyncer() resourcesynccontroller.ResourceSyncer {
+	return l.ResourceSync
+}
+
+func (l Listers) PreRunHasSynced() []cache.InformerSynced {
+	return l.PreRunCachesSynced
+}
+
+// CISConfigObserverController watches information that's relevant to CSI driver operators.
+// For now it only observes proxy information, (through the proxy.config.openshift.io/cluster
+// object), but more will be added.
+type CSIConfigObserverController struct {
+	factory.Controller
+}
+
+// NewCSIConfigObserverController returns a new CSIConfigObserverController.
+func NewCSIConfigObserverController(
+	name string,
+	operatorClient v1helpers.OperatorClient,
+	configinformers configinformers.SharedInformerFactory,
+	eventRecorder events.Recorder,
+) *CSIConfigObserverController {
+	informers := []factory.Informer{
+		operatorClient.Informer(),
+		configinformers.Config().V1().Proxies().Informer(),
+	}
+
+	c := &CSIConfigObserverController{
+		Controller: configobserver.NewConfigObserver(
+			operatorClient,
+			eventRecorder.WithComponentSuffix("csi-config-observer-controller-"+strings.ToLower(name)),
+			Listers{
+				ProxyLister_: configinformers.Config().V1().Proxies().Lister(),
+				PreRunCachesSynced: append([]cache.InformerSynced{},
+					operatorClient.Informer().HasSynced,
+					configinformers.Config().V1().Proxies().Informer().HasSynced,
+				),
+			},
+			informers,
+			proxy.NewProxyObserveFunc(ProxyConfigPath()),
+		),
+	}
+
+	return c
+}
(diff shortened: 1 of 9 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -109,6 +128,81 @@ func (c *CSIControllerSet) WithCredentialsRequestController(
 	return c
 }
 
+func (c *CSIControllerSet) WithSecretHash(
```
bertinatto, line 131: TODO: find a better name

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -0,0 +1,66 @@
...
+	corev1 "k8s.io/client-go/informers/core/v1"
+
+	opv1 "github.com/openshift/api/operator/v1"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehash"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// WithObservedProxyHook injects into the deployment's containers the observed proxy config.
+func WithObservedProxyHook() DeploymentHookFunc {
```
gnufied, line 17: This basically assumes only deployment needs to be updated when proxy configuration changes, but what if pods managed by daemonset needs to be restarted?

pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
```diff
@@ -0,0 +1,75 @@
...
+			eventRecorder.WithComponentSuffix("csi-config-observer-controller-"+strings.ToLower(name)),
+			Listers{
+				ProxyLister_: configinformers.Config().V1().Proxies().Lister(),
+				PreRunCachesSynced: append([]cache.InformerSynced{},
+					operatorClient.Informer().HasSynced,
+					configinformers.Config().V1().Proxies().Informer().HasSynced,
+				),
+			},
+			informers,
+			proxy.NewProxyObserveFunc([]string{"csidrivercontrollerservicecontroller", "proxy"}),
```
gnufied, line 70: given that this could in future affect node side as well - can we call the path something else? may be a simple - `target-csi-config` or something else.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -0,0 +1,66 @@
...
+	return func(opSpec *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		deploymentCopy, err := v1helpers.InjectObservedProxyInDeploymentContainers(
+			deployment,
+			opSpec.ObservedConfig.Raw,
+			"csidrivercontrollerservicecontroller",
```
gnufied, line 22: Since this string comes from a path that was injected elsewhere in `opSpec` - how about we define a constant for it?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -0,0 +1,66 @@
...
+			"proxy",
```
gnufied, line 23: Same for `proxy` string.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -0,0 +1,66 @@
...
+			return fmt.Errorf("invalid dependency reference: %w", err)
+		}
+		if deployment.Annotations == nil {
+			deployment.Annotations = map[string]string{}
+		}
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = map[string]string{}
+		}
+		for k, v := range inputHashes {
+			annotationKey := fmt.Sprintf("operator.openshift.io/dep-%s", k)
```
gnufied, line 55: lets move this into a var/constant

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: Added a hook for the DaemonSet

pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
bertinatto: Done

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: done

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -8,20 +8,21 @@ import (
 	"strings"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
+	"k8s.io/client-go/kubernetes"
```
deads2k, line 15: nit: I don't like separate blocks after the standard imports/.

pkg/operator/v1helpers/helpers.go
```diff
@@ -310,3 +313,30 @@ func MapToEnvVars(mapEnvVars map[string]string) []corev1.EnvVar {
...
+		return nil, fmt.Errorf("couldn't get the proxy config from observedConfig: %w", err)
+	}
+
+	if !found {
+		return containers, nil
+	}
+
+	proxyEnvVars := MapToEnvVars(proxyConfig)
+	patchedContainers := []corev1.Container{}
+	for i := range containers {
```
deads2k, line 335: please use the annotation from https://github.com/openshift/cluster-version-operator/pull/224 to identify the containers to inject.

deads2k:
I think https://github.com/openshift/library-go/pull/976/files#r556138811 is my only blocker before approval.  Good PR.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
bertinatto: done

pkg/operator/v1helpers/helpers.go
bertinatto: done

jsafrane:
/refresh
(CI & github have some issues and github UI is not refreshed. See #announce-testplatform
.

deads2k:
Thanks,

/approve

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: I'd like to leave this on as is. I prefer to leave format strings inline, plus I couldn't re-use the const for the annotation below (`%x` != `%s`)

PR #983, opened 2021-01-22
optimize library-go to eliminate unnecessary SARs

deads2k:
healthz and readyz have to accessible to unauthenticated so the kubelet can health check. Because of this, there is no reason to run an authz check on them.

system:masters has full power, so the authorization check is unnecessary and just uses an extra call.

### DIFF
--- a/pkg/config/serving/server.go
+++ b/pkg/config/serving/server.go
@@ -58,7 +58,10 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
 	}
 
 	if !authorizationConfig.Disabled {
-		authorizationOptions := genericapiserveroptions.NewDelegatingAuthorizationOptions()
+		authorizationOptions := genericapiserveroptions.NewDelegatingAuthorizationOptions().
+			WithAlwaysAllowPaths("/healthz", "/readyz", "/livez"). // this allows the kubelet to always get health and readiness without causing an access check
+			WithAlwaysAllowGroups("system:masters")                // in a kube cluster, system:masters can take any action, so there is no need to ask for an authz check
+
 		authorizationOptions.RemoteKubeConfigFile = kubeConfigFile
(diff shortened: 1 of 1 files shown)


pkg/config/serving/server.go
```diff
@@ -58,7 +58,10 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
 	}
 
 	if !authorizationConfig.Disabled {
-		authorizationOptions := genericapiserveroptions.NewDelegatingAuthorizationOptions()
+		authorizationOptions := genericapiserveroptions.NewDelegatingAuthorizationOptions().
+			WithAlwaysAllowPaths("/healthz", "/healthz/", "/readyz"). // this allows the kubelet to always get health and readiness without causing an access check
```
p0lyn0mial, line 62: > healthz and readyz have to accessible to unauthenticated so the kubelet can health check. Because of this, there is no reason to run an authz check on them.

how does it work today, then?

pkg/config/serving/server.go
```diff
@@ -58,7 +58,10 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
...
+			WithAlwaysAllowGroups("system:masters") // in a kube cluster, system:masters can take any action, so there is no need to ask for an authz check
```
p0lyn0mial, line 63: is this for the `metrics` endpoint ?

pkg/config/serving/server.go
deads2k: > how does it work today, then?

it does a subject access review  (remote call)

deads2k:
manually checked efficacy against kas-o

pkg/config/serving/server.go
p0lyn0mial: I was asking because from what I have seen this method is used in the operators to expose the `metrics` endpoint.

PR #984, opened 2021-01-25
release-3.11: Set TCP_USER_TIMEOUT socket option

p0lyn0mial:
clean pick of https://github.com/openshift/library-go/pull/926 and https://github.com/openshift/library-go/pull/944. I only removed `klog/v2`

### DIFF
--- a/pkg/network/dialer.go
+++ b/pkg/network/dialer.go
@@ -1,0 +1,13 @@
+package network
+
+import (
+	"context"
+	"net"
+)
+
+type DialContext func(ctx context.Context, network, address string) (net.Conn, error)
+
+// DefaultDialContext returns a DialContext function from a network dialer with default options sets.
+func DefaultClientDialContext() DialContext {
+	return dialerWithDefaultOptions()
+}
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
> Travis CI - Pull Request Failing after 1m — Build Failed

pkg/network/dialer_linux.go:23:10: unknown field `Control` in struct literal of type net.Dialer

p0lyn0mial:
The `net.Dialer.Control` (https://golang.org/pkg/net/#Dialer) method was added in golang `1.11` as a way of setting various sockets options. 

For `1.9` (used in 3.11) we could try to extract the underlying [TCPConn.File](https://golang.org/pkg/net/#TCPConn.File
) but it might not work. As the documentation states 
`The returned os.File's file descriptor is different from the connection's. Attempting to change properties of the original using this duplicate may or may not have the desired effect.`


I would recommend bumping golang version for 3.11 at least to `1.11` or backport [the PR](https://github.com/golang/go/commit/3c4d3bdd3b454ef45ce00559d705fe5dc6f57cad) that introduced the `Control` method to `1.9`

/close

PR #985, opened 2021-01-26
introduces a few convenient functions for querying netlink subsystem

p0lyn0mial:
`netlink` can be used for obtaining information about network sockets of various address families from the Linux kernel.

more information can be found at https://man7.org/linux/man-pages/man7/sock_diag.7.html

### DIFF
--- a/pkg/network/networkutils/netlink.go
+++ b/pkg/network/networkutils/netlink.go
@@ -1,0 +1,167 @@
+// +build linux
+
+package networkutils
+
+import (
+	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	nl "github.com/elastic/gosigar/sys/linux"
+)
+
+// SocketDiagMsg struct that holds data returned from netlink and filtered for our purposes
+//
+// full description can be found at https://man7.org/linux/man-pages/man7/sock_diag.7.html
+// the original struct name is inet_diag_msg
+type SocketDiagMsg struct {
+	// state holds a human-readable value of the socket state, like ESTABLISHED or LISTEN
+	state string
+
+	// recvQ - from https://man7.org/linux/man-pages/man7/sock_diag.7.html:
+	//  for listening sockets: the number of pending connections.
+	//  for other sockets: the amount of data in the incoming queue.
+	//
+	// in practice it tells you the number of bytes of data that have been received by the kernel but haven't been copied by the process
+	recvQ int
+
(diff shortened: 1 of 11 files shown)


pkg/network/networkutils/netlink.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// remoteAddress holds an IP address and the port of the remote process
+	remoteAddress string
+
+	// timer - see timerFieldToHumanReadable for possible values
+	timer string
+
+	// retransmissions - for timer values 1, 2, and 4, this field contains
+	// the number of retransmits. For other timer values, this field is set to 0.
+	retransmissions int
```
p0lyn0mial, line 45: should be renamed to `timerRetransmissions`

pkg/network/networkutils/netlink.go
```diff
@@ -0,0 +1,167 @@
...
+// This method uses netlink subsystem for obtaining information
+// about sockets of various address families from the linux kernel.
+//
+// in the future we could:
+//  use InetDiagReqV2 to specify an address family (IPv4 or IPv6) and a protocol (TCP, UDP)
+//  ask the kernel to only return sockets for specific port or address (filtering)
+//
+// Note:
+//  depending on the platform this function is run on, it might require setting custom permissions.
+func GetNetworkSocketsDiagnosticMsg() ([]*SocketDiagMsg, error) {
```
sttts, line 91: "Get" sounds like a no-cost operation. This is more like "Collect" or "Gather".

pkg/network/networkutils/netlink.go
```diff
@@ -0,0 +1,167 @@
+// +build linux
```
sttts, line 1: we need a no-op replacement for other platforms

p0lyn0mial: for each method?

Would it be okay to add a replacement when composing the final app ?

For example, the final app could have two files `app.go and app_others.go`.
Then `app_others.go` could simply print that running it is not supported on that platform.

sttts: yes, could work too

deads2k: > Would it be okay to add a replacement when composing the final app ?

Seems like the library shouldn't just flat out fail to compile on other platforms.

p0lyn0mial:
are we actually still interested in it?

PR #989, opened 2021-02-05
serviceability: Remove Sentry

smarterclayton:
We haven't used it in quite a while and have other solutions now
for finding panics in our binaries.


soltysh (approved):
/lgtm

PR #990, opened 2021-02-05
pkg/image: Remove dockerv1client client

smarterclayton:
We no longer need the v1 client and can cut our dependency on the
fsouza/go-dockerclient library. Preserve the types for compatibility.


smarterclayton:
/assign @dmage 

v1 docker is dead

PR #991, opened 2021-02-09
Enable static pod installer substitution of NODE_UID

marun:
Identifying etcd member certs by node uid rather than node name ensures that node deletion followed by addition with the same name can be detected as a member change.

/cc @hexfusion @deads2k

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -34,6 +34,7 @@ type InstallOptions struct {
 
 	Revision  string
 	NodeName  string
+	NodeUID   string
 	Namespace string
 
 	PodConfigMapNamePrefix        string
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -362,6 +363,7 @@ func (o *InstallOptions) substituteConfigMap(obj *corev1.ConfigMap) *corev1.Conf
 		newContent := strings.ReplaceAll(oldContent, "REVISION", o.Revision)
 		newContent = strings.ReplaceAll(newContent, "NODE_NAME", o.NodeName)
 		newContent = strings.ReplaceAll(newContent, "NODE_ENVVAR_NAME", strings.ReplaceAll(strings.ReplaceAll(o.NodeName, "-", "_"), ".", "_"))
+		newContent = strings.ReplaceAll(newContent, "NODE_ENVVAR_UID", strings.ReplaceAll(strings.ReplaceAll(o.NodeUID, "-", "_"), ".", "_"))
```
retroflexer, line 366: nit: UID proably needs replacing dashes (-), but not dots (.). 

Example node UID is: ec2b7f7d-12f5-533a-89c6-2f99086de704

marun: Good catch. I've updated to just use the uid verbatim.

deads2k:
I'm ok with this philosophically.  I'm less sure about this in usage.  If a node object got deleted and the kubelet restarts, the node object would be recreated but the information previously on disk would still be present.  I'm not entirely clear what you would expect to happen since this information on disk would be different, but the configmaps in the API would still be the same.


Is there an overall flow described somewhere?  @hexfusion do you know the full flow and I'm just slow seeing it?

PR #995, opened 2021-02-17
add harcoded authorizer type for well-known rules

deads2k:
Notably, we want to allow the metrics scraper since that rule is always present in openshift clusters

### DIFF
--- a/pkg/authorization/hardcodedauthorizer/hardcoded_authorizer.go
+++ b/pkg/authorization/hardcodedauthorizer/hardcoded_authorizer.go
@@ -1,0 +1,82 @@
+package hardcodedauthorizer
+
+import (
+	"context"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+type hardcodedAuthorizer struct {
+	usernames sets.String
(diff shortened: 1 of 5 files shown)


pkg/authorization/hardcodedauthorizer/hardcoded_authorizer.go
```diff
@@ -0,0 +1,82 @@
...
+	ret, err := NewHardcodedAuthorizer(rules, usernames...)
+	if err != nil {
+		panic(err)
+	}
+	return ret
+}
+
+// NewHardcodedMetricsScaperAuthorizer provides an authorizer that allows metrics scraping without contacting the
+// kube-apiserver for an authorization check on a rule we always have in place.
+func NewHardcodedMetricsScaperAuthorizer() authorizer.Authorizer {
```
deads2k, line 76: @s-urbaniak please confirm this rule and serviceaccount

sttts:
What is this? Fork of RBAC?

/hold

pkg/controller/controllercmd/builder.go
```diff
@@ -9,6 +9,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/library-go/pkg/authorization/hardcodedauthorizer"
```
mfojtik, line 12: order

deads2k:
replaced with approach in https://github.com/openshift/library-go/pull/998

PR #996, opened 2021-02-18
Allow SCC to be created during bootstrapping

deads2k:
Because discovery for security.openshift.io is served by openshift-apiserver, which is unavailable from the bootstrap node and not present early in an installation flow.  This is a minimal change to pin the RESTMapping so the existing dynamic client flows know where to direct securitycontextconstraints without actual discovery data.

This mechanism is directly lifted from our kubernetes fork where we did this to resolve an ownerref problem.


deads2k:
/hold 

still debugging

PR #998, opened 2021-02-22
add hardcoded authorizers for metrics and token reviews

deads2k:
/assign @sttts 

proof in https://github.com/openshift/kubernetes/pull/588

### DIFF
--- a/pkg/authorization/hardcodedauthorizer/metrics.go
+++ b/pkg/authorization/hardcodedauthorizer/metrics.go
@@ -1,0 +1,38 @@
+package hardcodedauthorizer
+
+import (
+	"context"
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/authorization/hardcodedauthorizer/tokenreview.go
```diff
@@ -0,0 +1,41 @@
+package hardcodedauthorizer
+
+import (
+	"context"
+
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+type tokenReviewAuthorizer struct{}
```
sttts, line 9: why is this in library-go?

deads2k: > why is this in library-go?

only because it was easy to write the tests here.  I'm not particularly attached.

PR #999, opened 2021-02-24
Support generating client certs for a raw subject

marun:
`CA.MakeClientCertificateForDuration` accepts `user.Info` and converts it to a pkix.Name to generate the client cert. This is fine for generating client certs for kube users, but generating client certs for etcd requires setting the raw subject.

/cc @deads2k @mfojtik @hexfusion

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -827,8 +827,12 @@ func (ca *CA) MakeClientCertificate(certFile, keyFile string, u user.Info, expir
 }
 
 func (ca *CA) MakeClientCertificateForDuration(u user.Info, lifetime time.Duration) (*TLSCertificateConfig, error) {
+	return ca.MakeClientCertificateForSubjectAndDuration(userToSubject(u), lifetime)
+}
+
+func (ca *CA) MakeClientCertificateForSubjectAndDuration(subject pkix.Name, lifetime time.Duration) (*TLSCertificateConfig, error) {
 	clientPublicKey, clientPrivateKey, _ := NewKeyPair()
-	clientTemplate := newClientCertificateTemplateForDuration(userToSubject(u), lifetime, time.Now)
+	clientTemplate := newClientCertificateTemplateForDuration(subject, lifetime, time.Now)
 	clientCrt, err := ca.signCertificate(clientTemplate, clientPublicKey)
 	if err != nil {
 		return nil, err


pkg/crypto/crypto.go
```diff
@@ -644,15 +644,15 @@ func MakeSelfSignedCAConfigForSubject(subject pkix.Name, expireDays int) (*TLSCe
...
 }
 
 func MakeSelfSignedCAConfigForDuration(name string, caLifetime time.Duration) (*TLSCertificateConfig, error) {
 	subject := pkix.Name{CommonName: name}
-	return makeSelfSignedCAConfigForSubjectAndDuration(subject, caLifetime)
+	return MakeSelfSignedCAConfigForSubjectAndDuration(subject, caLifetime)
 }
 
-func makeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, caLifetime time.Duration) (*TLSCertificateConfig, error) {
+func MakeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, caLifetime time.Duration) (*TLSCertificateConfig, error) {
```
deads2k, line 655: it's not clear why this should be exposed instead of the already exposed `MakeSelfSignedCAConfigForSubject` which has some checking and then delegates here.

pkg/crypto/crypto.go
```diff
@@ -827,8 +827,12 @@ func (ca *CA) MakeClientCertificate(certFile, keyFile string, u user.Info, expir
 }
 
 func (ca *CA) MakeClientCertificateForDuration(u user.Info, lifetime time.Duration) (*TLSCertificateConfig, error) {
+	return ca.MakeClientCertificateForSubjectAndDuration(userToSubject(u), lifetime)
+}
+
+func (ca *CA) MakeClientCertificateForSubjectAndDuration(subject pkix.Name, lifetime time.Duration) (*TLSCertificateConfig, error) {
```
deads2k, line 833: what are you setting that cannot be a user.Info?  `userToSubject` has some warnings about golang issues with certain orderings.

marun: The installer is currently setting `OrganizationalUnit` (and no groups): https://github.com/openshift/installer/blob/master/pkg/asset/tls/etcd.go#L25

pkg/crypto/crypto.go
marun: I think it makes sense to enable callers to provide the lifetime as a duration constant rather than a magic number of days, but that's just me. If you feel strongly I can remove this change.

deads2k: > I think it makes sense to enable callers to provide the lifetime as a duration constant rather than a magic number of days, but that's just me. If you feel strongly I can remove this change.

I'd like to remove this one.

pkg/crypto/crypto.go
deads2k: Your public method should check to see that this subject conforms to rules here: https://github.com/openshift/library-go/blob/9723292f0698a984407307fa241880ea8b5ca9a0/pkg/crypto/crypto.go#L867-L879

pkg/crypto/crypto.go
marun: Done

PR #1000, opened 2021-02-24
introduces health check monitor

p0lyn0mial:
This PR adds a health check monitor that periodically sends requests to the provided targets to check their health status.
Users can configure:

 - `unhealthyProbesThreshold`: specifies consecutive failed health checks after which a target is considered unhealthy
 - `healthyProbesThreshold`: specifies consecutive successful health checks after which a target is considered healthy
 - `probeResponseTimeout`: specifies a time limit for requests made by the HTTP client for the health check
 - `probeInterval`: specifies a time interval at which health checks are send

The HM also implements `Listener` and `Notifier` interfaces.

It automatically registers for notification if the target provided also implements the `Notifier` interface.
It is implicit so that the provider can provide a static or a dynamic list of targets.

Interested parties can register a listener for notifications about healthy/unhealthy targets changes via AddListener.

This PR also defines the following metrics:

1. `health_monitor_healthy_target_total` - for tracking the number of healthy instances registered partitioned by targets.
2. `health_monitor_unhealthy_target_total` - for tracking the number of unhealthy instances registered partitioned by targets.
3. `health_monitor_current_healthy_targets` - for tracking the number of current healthy targets observed by the health monitor.

### DIFF
--- a/pkg/healthmonitor/health_monitor.go
+++ b/pkg/healthmonitor/health_monitor.go
@@ -1,0 +1,338 @@
+package healthmonitor
+
+import (
+	"fmt"
+	"net/http"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+	"k8s.io/klog/v2"
+)
+
+type HealthMonitor struct {
+	// targetProvider provides a list of targets to monitor
+	// it also can schedule refreshing the list by simply calling Enqueue method
+	targetProvider TargetProvider
+
+	// client is an HTTP client that is used to probe health checks for targets
+	client *http.Client
+
+	// probeResponseTimeout specifies a time limit for requests made by the HTTP client for the health check
+	probeResponseTimeout time.Duration
+
+	// probeInterval specifies a time interval at which health checks are send
+	probeInterval time.Duration
+
+	// unhealthyProbesThreshold specifies consecutive failed health checks after which a target is considered unhealthy
+	unhealthyProbesThreshold int
+
+	// healthyProbesThreshold  specifies consecutive successful health checks after which a target is considered healthy
+	healthyProbesThreshold int
+
+	healthyTargets   []string
+	unhealthyTargets []string
+	targetsToMonitor []string
+
+	consecutiveSuccessfulProbes map[string]int
+	consecutiveFailedProbes     map[string][]error
+
+	refreshTargetsLock sync.Mutex
+	refreshTargets     bool
+
+	// exportedHealthyTargets holds a copy of healthyTargets
+	exportedHealthyTargets atomic.Value
+
+	// exportedUnhealthyTargets holds a copy of unhealthyTargets
+	exportedUnhealthyTargets atomic.Value
+
+	// listeners holds a list of interested parties to be notified when the list of healthy targets changes
+	listeners []Listener
+}
+
+var _ Listener = &HealthMonitor{}
+var _ Notifier = &HealthMonitor{}
+
+// New creates a health monitor that periodically sends requests to the provided targets to check their health.
+// This method also allows you to configure behaviour of the monitor upfront.
+//
+//   unhealthyProbesThreshold - that specifies consecutive failed health checks after which a target is considered unhealthy
+//   healthyProbesThreshold   - that specifies consecutive successful health checks after which a target is considered healthy
+//   probeResponseTimeout     - that specifies a time limit for requests made by the HTTP client for the health check
+//   probeInterval            - that specifies a time interval at which health checks are send
+//
+//
+// Additionally the monitor implements Listener and Notifier interfaces.
+//
+// The health monitor automatically registers for notification if the target provided also implements the Notifier interface.
+// It is implicit so that the provider can provide a static or a dynamic list of targets.
+//
+// Interested parties can register a listener for notifications about healthy/unhealthy targets changes via AddListener.
+// TODO: instead of restConfig we could accept transport so that it is reused instead of creating a new connection to targets
+//       reusing the transport has the advantage of using the same connection as other clients
+func New(targetProvider TargetProvider, restConfig *rest.Config, unhealthyProbesThreshold int, healthyProbesThreshold int, probeResponseTimeout, probeInterval time.Duration) (*HealthMonitor, error) {
+	client, err := createHealthCheckHTTPClient(probeResponseTimeout, restConfig)
+	if err != nil {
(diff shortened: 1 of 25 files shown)


pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+		// notify listeners about the new healthy/unhealthy targets
+		for _, listener := range sm.listeners {
+			listener.Enqueue()
+		}
+	}
+}
+
+func (sm *HealthMonitor) healthCheckSingleTarget(target string) error {
+	// TODO: make the protocol, port and the path configurable
+	url := fmt.Sprintf("https://%s/%s", target, "readyz")
```
mfojtik, line 289: i would use the `url` package here maybe... also we might need to deal with ipv6

pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+	newReq, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return err
+	}
+
+	resp, err := sm.client.Do(newReq)
```
mfojtik, line 295: we probably need to manage some context here or deal with timeouts no? what happen if we do a request into a blackhole?

pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+	tlsConfig, err := transport.TLSConfigFor(transportConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: utilnet.SetTransportDefaults(&http.Transport{
+			TLSClientConfig: tlsConfig,
+		}),
+		Timeout: responseTimeout,
```
mfojtik, line 322: probably ignore the other comment, but make sure this timeout really is the maximum timeout the request will take... including DNS resolving

pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+
+func logUnhealthyTargets(unhealthyTargets []string, currentHealthCheckProbes []targetErrTuple) {
+	for _, unhealthyTarget := range unhealthyTargets {
+		errorsForUnhealthyTarget := []error{}
+		for _, svrErrTuple := range currentHealthCheckProbes {
+			if svrErrTuple.target == unhealthyTarget {
+				errorsForUnhealthyTarget = append(errorsForUnhealthyTarget, svrErrTuple.err)
+			}
+		}
+		klog.V(2).Infof("the following target %v became unhealthy due to %v", unhealthyTarget, utilerrors.NewAggregate(errorsForUnhealthyTarget).Error())
```
mfojtik, line 336: we probably want to turn this into a Prometheus metrics at some point

pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+func (sm *HealthMonitor) AddListener(listener Listener) {
+	sm.listeners = append(sm.listeners, listener)
+}
+
+type targetErrTuple struct {
+	target string
+	err    error
+}
+
+// refreshTargetsLocked updates the internal targets list to monitor if it was requested (via the Enqueue method)
```
mfojtik, line 145: will this allow to dynamically add/remove endpoints? i wonder whether we want to support it or not... I would rather lock adding new endpoints

pkg/healthmonitor/health_monitor.go
```diff
@@ -0,0 +1,338 @@
...
+	if notifier, ok := targetProvider.(Notifier); ok {
+		notifier.AddListener(hm)
+	}
+
+	return hm, nil
+}
+
+// Run starts monitoring the provided targets until stop channel is closed
+// This method is blocking and it is meant to be launched in a separate goroutine
+func (sm *HealthMonitor) Run(stopCh <-chan struct{}) {
```
mfojtik, line 110: please replace stopCh with context :-)

pkg/healthmonitor/health_monitor.go
p0lyn0mial: there is `probeResponseTimeout` parameter for specifying the maximum time limit for requests made by the HTTP client

pkg/healthmonitor/health_monitor.go
p0lyn0mial: yes, already on the TODO list https://github.com/openshift/library-go/pull/1000/commits/ed4d6682c69d3a81e4a6c2bd02144a5ba4e89d60#diff-723588c49708ae3a6e8ea0f3974fb4f2c0ba5ed298630a546a4cdb7f8e746aeeR207

pkg/healthmonitor/health_monitor.go
p0lyn0mial: yes, we already know that the list can be dynamic or static. For example, for the aggregator the list will be dynamic

pkg/healthmonitor/health_monitor.go
p0lyn0mial: +1

pkg/healthmonitor/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	compbasemetrics "k8s.io/component-base/metrics"
+)
+
+type registerables []compbasemetrics.Registerable
+
+var (
+	healthyTargetsTotal = compbasemetrics.NewCounterVec(
+		&compbasemetrics.CounterOpts{
+			Name:           "health_monitor_healthy_target_total",
+			Help:           "Number of healthy instances registered with the health monitor. Partitioned by targets",
```
sttts, line 13: this should be a gauge, no?

pkg/healthmonitor/metrics.go
```diff
@@ -0,0 +1,64 @@
...
+	)
+
+	unHealthyTargetsTotal = compbasemetrics.NewCounterVec(
+		&compbasemetrics.CounterOpts{
+			Name:           "health_monitor_unhealthy_target_total",
+			Help:           "Number of unhealthy instances registered with the health monitor. Partitioned by targets",
+			StabilityLevel: compbasemetrics.ALPHA,
+		},
+		[]string{"target"},
+	)
```
sttts, line 26: we need metrics for:
- counter for toggling
- counter for network issues (connection refused) that are not following a period of N seconds of red `/readyz`

p0lyn0mial: > counter for toggling

I think that having the total for healthy/unhealthy is enough to alert on `toggling`

pkg/healthmonitor/metrics.go
p0lyn0mial: we could have a gauge to track the current number of healthy targets, alert when is `0` for some time.

pkg/healthmonitor/metrics.go
p0lyn0mial: > counter for network issues (connection refused) that are not following a period of N seconds of red /readyz

do you mean detect connection refused errors when an ep turned red on `/readyz` since X seconds ago?

sttts: > do you mean detect connection refused errors when an ep turned red on /readyz since X seconds ago?

Want to count when endpoint become unhealthy without following the `/readyz` protocol.

PR #1001, opened 2021-02-25
csi: give informers from controllerset 10 minutes to sync

bertinatto:
When informers added to the controllerset with the method `.WithExtraInformers()` are **not started**, the cache-sync operation will wait forever and this message is shown in the log:

>  base_controller.go:66] Waiting for caches to sync for OpenStackCinderDriverControllerServiceController

This PR adds a custom context that will make the sync operation time out after 10 minutes. Once this happens, the following error message will be logged:

> caches did not sync

CC @openshift/storage @Fedosin

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -3,6 +3,7 @@ package csicontrollerset
 import (
 	"context"
 	"fmt"
+	"time"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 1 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -24,6 +25,8 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+var defaultCacheSyncTimeout = 10 * time.Minute
+
```
bertinatto, line 29: Same value as `factory.Factory`:

https://github.com/bertinatto/library-go/blob/158312305434c3b3250ca6cf9144f4b237d7135a/pkg/controller/factory/base_controller.go#L27

Fedosin (commented):
/lgtm

PR #1002, opened 2021-03-01
Bug 1933263: pkg/assets: get resources before create to cope with pre-etcd-create errors

sttts:
cluster-bootstrap creates object that either fail on admission on recreate or in different ways than `AlreadyExists`. In resourceapply funcs we already follow that pattern, but in cluster-bootstrap using the assets package we did not.

### DIFF
--- a/pkg/assets/create/creater.go
+++ b/pkg/assets/create/creater.go
@@ -178,34 +178,36 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
 		}
 		resourceString := mappings.Resource.Resource + "." + mappings.Resource.Version + "." + mappings.Resource.Group + "/" + manifests[path].GetName() + " -n " + manifests[path].GetNamespace()
 
-		incluster, err := resource.Create(ctx, manifests[path], metav1.CreateOptions{})
-
-		if err == nil && options.Verbose {
-			fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
-		}
-
-		// Resource already exists means we already succeeded
-		// This should never happen as we remove already created items from the manifest list, unless the resource existed beforehand.
-		if kerrors.IsAlreadyExists(err) {
+		incluster, err := resource.Get(ctx, manifests[path].GetName(), metav1.GetOptions{})
+		switch {
+		case err == nil:
 			if options.Verbose {
 				fmt.Fprintf(options.StdErr, "Skipped %q %s as it already exists\n", path, resourceString)
 			}
-			incluster, err = resource.Get(ctx, manifests[path].GetName(), metav1.GetOptions{})
-			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/assets/create/creater.go
```diff
@@ -178,36 +178,37 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+		if err == nil {
 			if options.Verbose {
 				fmt.Fprintf(options.StdErr, "Skipped %q %s as it already exists\n", path, resourceString)
 			}
-			incluster, err = resource.Get(ctx, manifests[path].GetName(), metav1.GetOptions{})
-			if err != nil {
+			// fall through as if it was just created
+		} else if !kerrors.IsNotFound(err) {
+			if options.Verbose {
+				fmt.Fprintf(options.StdErr, "Failed to get already existing %q %s: %v\n", path, resourceString, err)
```
mfojtik, line 189: this is now wrong, isn't it? the error is not NotFound, so we don't know whether the resource is existing or not.

pkg/assets/create/creater.go
```diff
@@ -178,36 +178,37 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			}
+			errs[path] = fmt.Errorf("failed to get %s: %v", resourceString, err)
+			continue
+		} else {
```
mfojtik, line 193: maybe a switch here would be cleaner... this else is only reached for IsNotFound ?

pkg/assets/create/creater.go
```diff
@@ -178,36 +178,37 @@ func create(ctx context.Context, manifests map[string]*unstructured.Unstructured
...
+			incluster, err = resource.Create(ctx, manifests[path], metav1.CreateOptions{})
+			if err == nil && options.Verbose {
+				fmt.Fprintf(options.StdErr, "Created %q %s\n", path, resourceString)
+			}
+			if kerrors.IsAlreadyExists(err) {
+				if options.Verbose {
+					fmt.Fprintf(options.StdErr, "Skipped %q %s as it already exists\n", path, resourceString)
```
mfojtik, line 200: can you change the message to "Skipped creating %q ..." so we can identify this happening (should be rare?)

pkg/assets/create/creater.go
sttts: yes. Copied it from the create case without reading the message :D

sttts: fixed

pkg/assets/create/creater.go
sttts: yes. Made it clearer.

pkg/assets/create/creater.go
sttts: done

PR #1003, opened 2021-03-02
workloadscontroller: don't prefix workloads name on empty prefix

stlaz:
Required for https://github.com/openshift/cluster-authentication-operator/pull/414

/assign @p0lyn0mial

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -291,7 +291,11 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
 	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
 	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
-		c.versionRecorder.SetVersion(fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name), c.targetOperandVersion)
+		operandName := workload.Name
+		if len(c.operandNamePrefix) > 0 {
+			operandName = fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name)
+		}
+		c.versionRecorder.SetVersion(operandName, c.targetOperandVersion)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -291,7 +291,11 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
 	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
 	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
-		c.versionRecorder.SetVersion(fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name), c.targetOperandVersion)
+		operandName := workload.Name
+		if len(c.operandNamePrefix) > 0 {
+			operandName = fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name)
```
p0lyn0mial, line 296: ok, ic, in case of `oauth-server` the workload.Name is `oauth-openshift`

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -291,7 +291,11 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
 	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
 	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
-		c.versionRecorder.SetVersion(fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name), c.targetOperandVersion)
+		operandName := workload.Name
+		if len(c.operandNamePrefix) > 0 {
```
p0lyn0mial, line 295: since this is the only place we use it, could we be more explicit and change it to something else? To indicate it is only used when setting the version.

stlaz:
(added context wiring to preconditions to this PR)

pkg/operator/apiserver/controller/workload/workload.go
p0lyn0mial: `operandVersionPrefix` ?

PR #1004, opened 2021-03-04
convenience functions for client-go transport

p0lyn0mial:
this PR adds a few convenience functions that I need in https://github.com/openshift/kubernetes/pull/603

### DIFF
--- a/pkg/config/client/transport.go
+++ b/pkg/config/client/transport.go
@@ -1,0 +1,58 @@
+package client
+
+import (
+	"net/http"
+	"net/url"
+
+	"k8s.io/client-go/rest"
+)
+
+// AnonymousClientConfigWithWrapTransport returns a copy of the given config with all user credentials (cert/key, bearer token, and username/password) and custom transports (Transport) removed.
+// This function preserves WrapTransport for clients that care about custom HTTP behavior.
+func AnonymousClientConfigWithWrapTransport(config *rest.Config) *rest.Config {
+	newConfig := rest.AnonymousClientConfig(config)
+	newConfig.WrapTransport = config.WrapTransport
+	return newConfig
+}
+
+// DefaultServerName extract the hostname from the config.Host and sets it in config.ServerName
+// the ServerName is passed to the server for SNI and is used in the client to check server certificates.
+//
+// note:
+// if the ServerName has been already specified calling this method has no effect
+func DefaultServerName(config *rest.Config) error {
+	if len(config.ServerName) > 0 {
+		return nil
+	}
+	u, err := url.Parse(config.Host)
+	if err != nil {
+		return err
+	}
+	config.ServerName = u.Host
+	return nil
+}
+
+// NewPreferredHostRoundTripper a simple middleware for changing the destination host for each request to the provided one.
+// If the preferred host doesn't exists (an empty string) then this RT has no effect.
+func NewPreferredHostRoundTripper(preferredHostFn func() string) func(http.RoundTripper) http.RoundTripper {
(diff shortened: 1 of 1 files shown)


pkg/config/client/transport.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (t *preferredHostRT) RoundTrip(r *http.Request) (*http.Response, error) {
+	preferredHost := t.preferredHostFn()
+
+	if len(preferredHost) == 0 {
+		return t.baseRT.RoundTrip(r)
+	}
+
+	r.Host = preferredHost
```
p0lyn0mial, line 35: copying here means that we won't see it in the logs, is that what we want?

pkg/config/client/transport.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+
+	"k8s.io/client-go/rest"
+)
+
+// AnonymousClientConfigWithWrapTransport returns a copy of the given config with all user credentials (cert/key, bearer token, and username/password) and custom transports (Transport) removed.
+// This function preserves WrapTransport for clients that care about custom HTTP behavior.
+func AnonymousClientConfigWithWrapTransport(config *rest.Config) *rest.Config {
+	newConfig := rest.AnonymousClientConfig(config)
+	newConfig.WrapTransport = config.WrapTransport
```
soltysh, line 13: I literally asked for that upstream :wink:

soltysh (approved):
/ltm

pkg/config/client/transport.go
p0lyn0mial: heh :)

p0lyn0mial:
had to fix the `verify` job

pkg/config/client/transport.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (t *preferredHostRT) RoundTrip(r *http.Request) (*http.Response, error) {
+	preferredHost := t.preferredHostFn()
+
+	if len(preferredHost) == 0 {
+		return t.baseRT.RoundTrip(r)
+	}
+
+	r.Host = preferredHost
+	r.URL.Host = preferredHost
```
sttts, line 38: don't we have to keep the tls server name in order to verify the serving cert successfully? Or do we have localhost certs?

p0lyn0mial: yes, we do have CA that knows how to verify the serving cert for localhost at least for KCM.

sttts: That way you could also use a node IP if necessary.

p0lyn0mial: we decided to set the `ServerName` at https://github.com/openshift/kubernetes/pull/603/commits/e5a8075fb932d3f91f8134d941d280a21149ded1#diff-d073b3d85bfc4877d354264d569dd3e19d544d1775a67f1650d661299869aa7aR448

pkg/config/client/transport.go
```diff
@@ -0,0 +1,52 @@
...
+func AnonymousClientConfigWithWrapTransport(config *rest.Config) *rest.Config {
+	newConfig := rest.AnonymousClientConfig(config)
+	newConfig.WrapTransport = config.WrapTransport
+	return newConfig
+}
+
+// OverwriteHostInTLSServerName extract the hostname from the config.Host and sets it in config.ServerName
+// the ServerName is passed to the server for SNI and is used in the client to check server certificates.
+func OverwriteHostInTLSServerName(config *rest.Config) error {
+	u, err := url.Parse(config.Host)
```
p0lyn0mial, line 21: `config.Host` is always scheme:host:port, where port is optional

PR #1005, opened 2021-03-05
factory: produce nicer message when reconcile fail

mfojtik:
Instead of 

`"OAuthRouteCheckEndpointAccessibleController" controller failed to sync "key", err: Get "https://oauth-openshift.apps.prod..com/healthz": EOF` 

produce error like:

`OAuthRouteCheckEndpointAccessibleController reconciliation failed: Get "https://oauth-openshift.apps.prod..com/healthz": EOF`

95% of all controller using factory just use string `"key"` as queue key and the controller word is always included in the controller name. This makes the message shorter and nicer to users (this message usually show up in logs and in some cases in degraded reason).

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -247,7 +247,11 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 			// logging this helps detecting wedged controllers with missing pre-requirements
 			klog.V(5).Infof("%q controller requested synthetic requeue with key %q", c.name, key)
 		} else {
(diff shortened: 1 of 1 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -247,7 +247,11 @@ func (c *baseController) processNextWorkItem(queueCtx context.Context) {
 			// logging this helps detecting wedged controllers with missing pre-requirements
 			klog.V(5).Infof("%q controller requested synthetic requeue with key %q", c.name, key)
 		} else {
-			utilruntime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", c.name, key, err))
+			if klog.V(4).Enabled() {
```
sttts, line 250: `|| key != "key"`?

PR #1008, opened 2021-03-09
Bug 1896558: Bump to use k8s/apiserver openshift fork

damemi:
This bumps and reacts to the changes in https://github.com/openshift/kubernetes/pull/558/files (pinning to openshift/kubernetes forks).


damemi (commented):
/cc @sttts @mfojtik 
With the changes from https://github.com/openshift/kubernetes/pull/558 that added this parameter, do we want to wire `true` to these controllers in https://github.com/openshift/library-go/pull/1008/commits/7ea38ea69a8b134a0e8873ba14b7f9ca41c15d82?

sttts:
/hold

We cannot use o/k here. This library-go is used in many other places where we don't want to use a kube fork.

damemi:
/close
we won't need this with https://github.com/openshift/kubernetes/pull/616

PR #1010, opened 2021-03-11
DefaultServerName extracts only the host portion from the network address

p0lyn0mial:


p0lyn0mial:
ref: https://github.com/openshift/kubernetes/pull/603

PR #1013, opened 2021-03-16
WIP: Support removable operators

jsafrane:
**Goal:** Add support for removable operators managed by OLM, namely CSI drivers. Operands (a CSI driver) should be installed when user creates a CR and removed when user deletes it.

Includes https://github.com/openshift/library-go/pull/1012 and https://github.com/openshift/library-go/pull/1017 to test with AWS EFS CSI driver operator (which is in progress)

* Most library-go controllers log error / warning messages when operator CR is missing. I updated `dynamicOperatorClient` to optionally (!) return a fake `OperatorSpec`/`OperatorStatus` when the CR is deleted / not yet created. The fake CR has `ManagementState: Unmanaged` to tell controllers to do nothing when CR is missing. [ Faking `ManagementState: Removed` would work too. ]

* Reusing `ManagementState: Removed` and  `management.SetOperatorNotRemovable()` to tell controllers to start removing the operands.

  * Most controllers ignore `ManagementState: Removed`  and most operators don't call `SetOperatorNotRemovable`, even though they don't support removal. I touched only `ManagementStateController` and CSI controllers in this PR. More controllers may come in future with removal support.
 
*  Extended `ManagementStateController` to force set `ManagementState: Removed` when the operator CR meets **all these conditions**:

      * Has a finalizer.
      * Has deletion timestamp.
      * The operator itself is removable (`SetOperatorNotRemovable` was not called)
  
    This way, other controllers need to react only to `ManagementState: Removed` and they don't need to check DeletionTimestamp or finalizers. Existing operators don't use finalizers on their CRs, therefore they won't be affected by this logic, even though they may not set themselves as `NotRemovable`.

    Similarly, users may set `ManagementState: Removed` to un-install the operands and `Managed` to install it back. This is just a side-effect, users should create / delete CRs to install/remove the operands.

  * Added helper `GetSyncAction` to further simplify controllers that may support Removed state.

* Removable operators may need to add/remove finalizers in the operator CR. I added a new interface `OperatorClientWithFinalizers` that can add/remove finalizers as the controller manage the operands / finished their removal. I avoided `OperatorClient` changes, because it's implemented on too many places.

* See the last commit how it all fits together, where CSI Node controller creates / deletes DaemonSet + adds/removes finalizers accordingly when its CR is created / gets DeletionTimestamp.

WIP:
* Fix/add unit tests
* Add all CSI Driver controllers (namely, the control plane Deployment one)
* Maybe update StaticResourceController to remove the objects it creates, however, this must be explicit opt-in, as many operators use it and I doubt they set `NotRemovable` correctly.

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -38,7 +38,7 @@ type CSIControllerSet struct {
 	csiDriverNodeServiceController       factory.Controller
 
 	preRunCachesSynced []cache.InformerSynced
-	operatorClient     v1helpers.OperatorClient
+	operatorClient     v1helpers.OperatorClientWithFinalizers
 	eventRecorder      events.Recorder
 }
 
@@ -84,7 +84,7 @@ func (c *CSIControllerSet) WithLogLevelController() *CSIControllerSet {
 // WithManagementStateController returns a *ControllerSet with a management state controller initialized.
 func (c *CSIControllerSet) WithManagementStateController(operandName string, supportsOperandRemoval bool) *CSIControllerSet {
 	c.managementStateController = management.NewOperatorManagementStateController(operandName, c.operatorClient, c.eventRecorder)
-	if supportsOperandRemoval {
+	if !supportsOperandRemoval {
 		management.SetOperatorNotRemovable()
 	}
 	return c
@@ -196,7 +196,7 @@ func (c *CSIControllerSet) WithExtraInformers(informers ...cache.SharedIndexInfo
 }
 
 // New returns a basic *ControllerSet without any controller.
-func NewCSIControllerSet(operatorClient v1helpers.OperatorClient, eventRecorder events.Recorder) *CSIControllerSet {
+func NewCSIControllerSet(operatorClient v1helpers.OperatorClientWithFinalizers, eventRecorder events.Recorder) *CSIControllerSet {
 	return &CSIControllerSet{
 		operatorClient: operatorClient,
 		eventRecorder:  eventRecorder,
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
@@ -137,7 +137,7 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
 }
 
 func (c *CSIDriverNodeServiceController) syncManaged(opSpec *opv1.OperatorSpec, opStatus *opv1.OperatorStatus, ctx context.Context, syncContext factory.SyncContext) error {
-	if !management.IsOperatorNotRemovable() {
+	if management.IsOperatorRemovable() {
 		err := c.operatorClient.EnsureFinalizer(c.name)
 		if err != nil {
 			return err
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
@@ -8,10 +8,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/management"
 	appsv1 "k8s.io/api/apps/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 
 	opv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -56,6 +59,8 @@ type DaemonSetHookFunc func(*opv1.OperatorSpec, *appsv1.DaemonSet) error
 // In order to do that, the placeholder ${LOG_LEVEL} from the manifest file is replaced with the value specified
 // in the OperatorClient resource (Spec.LogLevel).
 //
+// This controller supports removable operands, as configured in pkg/operator/management.
+//
 // This controller produces the following conditions:
 //
 // <name>Available: indicates that the CSI Node Service was successfully deployed.
@@ -64,7 +69,7 @@ type DaemonSetHookFunc func(*opv1.OperatorSpec, *appsv1.DaemonSet) error
 type CSIDriverNodeServiceController struct {
 	name           string
 	manifest       []byte
-	operatorClient v1helpers.OperatorClient
+	operatorClient v1helpers.OperatorClientWithFinalizers
 	kubeClient     kubernetes.Interface
 	dsInformer     appsinformersv1.DaemonSetInformer
 	// Optional hook functions to modify the DaemonSet.
@@ -77,7 +82,7 @@ type CSIDriverNodeServiceController struct {
 func NewCSIDriverNodeServiceController(
 	name string,
 	manifest []byte,
-	operatorClient v1helpers.OperatorClient,
+	operatorClient v1helpers.OperatorClientWithFinalizers,
 	kubeClient kubernetes.Interface,
 	dsInformer appsinformersv1.DaemonSetInformer,
 	recorder events.Recorder,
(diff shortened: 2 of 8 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -84,7 +84,7 @@ func (c *CSIControllerSet) WithLogLevelController() *CSIControllerSet {
 // WithManagementStateController returns a *ControllerSet with a management state controller initialized.
 func (c *CSIControllerSet) WithManagementStateController(operandName string, supportsOperandRemoval bool) *CSIControllerSet {
 	c.managementStateController = management.NewOperatorManagementStateController(operandName, c.operatorClient, c.eventRecorder)
-	if supportsOperandRemoval {
+	if !supportsOperandRemoval {
```
deads2k, line 87: fun

deads2k: > fun. do you know ahead of time if this breaks someone and/or how it does so?

I can see this is only in your package, but you could consider adding another param to break someone doing this upgrade in a way that they notice.  `reallySupportsOperandRemoval`

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -114,26 +119,36 @@ func (c *CSIDriverNodeServiceController) Name() string {
...
 	}
+}
 
-	manifest := replacePlaceholders(c.manifest, opSpec)
-	required := resourceread.ReadDaemonSetV1OrDie(manifest)
-
-	for i := range c.optionalDaemonSetHooks {
-		err := c.optionalDaemonSetHooks[i](opSpec, required)
+func (c *CSIDriverNodeServiceController) syncManaged(opSpec *opv1.OperatorSpec, opStatus *opv1.OperatorStatus, ctx context.Context, syncContext factory.SyncContext) error {
+	if !management.IsOperatorNotRemovable() {
```
deads2k, line 140: not-not is unfortunate.  Maybe provide a positive method?

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -114,26 +119,36 @@ func (c *CSIDriverNodeServiceController) Name() string {
...
+		err := c.operatorClient.AddFinalizer(c.name)
```
deads2k, line 141: give me a comment here with an example finalizer please.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -75,7 +92,11 @@ func (c dynamicOperatorClient) Informer() cache.SharedIndexInformer {
 func (c dynamicOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
 	uncastInstance, err := c.informer.Lister().Get(c.configName)
 	if err != nil {
+		if errors.IsNotFound(err) && c.fakeMissingInstance {
```
deads2k, line 95: nit
```go
if errors.IsNotFound(err) && c.fakeMissingInstance {
  // do thing
}
if err != nil{
  return err
}

```

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -84,6 +105,13 @@ func (c dynamicOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
 func (c dynamicOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
 	uncastInstance, err := c.informer.Lister().Get(c.configName)
 	if err != nil {
+		if errors.IsNotFound(err) && c.fakeMissingInstance {
```
deads2k, line 108: unnest

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -106,6 +134,9 @@ func (c dynamicOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *op
 func (c dynamicOperatorClient) UpdateOperatorSpec(resourceVersion string, spec *operatorv1.OperatorSpec) (*operatorv1.OperatorSpec, string, error) {
 	uncastOriginal, err := c.informer.Lister().Get(c.configName)
 	if err != nil {
+		if errors.IsNotFound(err) && c.fakeMissingInstance {
```
deads2k, line 137: unnest throughout

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -156,6 +190,76 @@ func (c dynamicOperatorClient) UpdateOperatorStatus(resourceVersion string, stat
 	return retStatus, nil
 }
 
+func (c dynamicOperatorClient) AddFinalizer(finalizer string) error {
```
deads2k, line 193: This looks like `EnsureFinalizer`

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -156,6 +190,76 @@ func (c dynamicOperatorClient) UpdateOperatorStatus(resourceVersion string, stat
...
+	klog.V(42).Infof("Removing finalizer %q: %v", finalizer, newFinalizers)
+	err = c.saveFinalizers(instance, newFinalizers)
+	if err != nil {
+		return err
+	}
+	klog.V(2).Infof("Removed finalizer %q", finalizer)
+	return nil
+}
+
+func (c dynamicOperatorClient) saveFinalizers(instance *unstructured.Unstructured, finalizers []string) error {
```
deads2k, line 256: this looks wrong. You'll be rewriting finalizers if someone adds in between.  We made an update helper to that can take a function to handle the mutation.

pkg/operator/management/management_state.go
```diff
@@ -29,6 +29,8 @@ func SetOperatorAlwaysManaged() {
 // SetOperatorNotRemovable is one time choice the operator author can make to indicate the operator does not support
 // removing of his operand. This makes sense for operators like kube-apiserver where removing operand will lead to a
 // bricked, non-automatically recoverable state.
+// Unless explicitly set, an operator supports Removed state.
+// Note that most library-go controllers do not support Removed state, please check each controller separately.
 func SetOperatorNotRemovable() {
 	allowOperatorRemovedState = false
```
deads2k, line 35: wow, we made a global?  really?  Please fix while you're in here.

pkg/operator/management/management_state.go
```diff
@@ -67,3 +70,34 @@ func IsOperatorManaged(state v1.ManagementState) bool {
...
 }
+
+// SyncAction tells caller of GetSyncAction what to do on operator Sync() based on the management state
+type SyncAction string
+
+const (
+	// SyncActionManage means the operator should manage its operands (create them if missing, update when needed).
+	SyncActionManage = "Manage"
+	// SyncActionDelete means the operator should delete its operands.
+	SyncActionDelete = "Delete"
```
deads2k, line 81: nit: prefer names to match the values set in the API.

deads2k:
Why return a fake instance instead of simply having the consumers handle a 404?

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
jsafrane: added `IsOperatorRemovable` as a separate commit.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
jsafrane: Each library-go controller adds its own finalizer, and I used controller name for that. In EFS it looks like:

```
  finalizers:
  - CSIStaticResourceController
  - AWSEFSDriverNodeServiceController
```

Other removable CSI operators will have one more `*DriverControllerServiceController`. BTW, non-removabe operators / CSI drivers won't have any finalizers.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
jsafrane: unnested

pkg/operator/genericoperatorclient/dynamic_operator_client.go
jsafrane: unnested

pkg/operator/genericoperatorclient/dynamic_operator_client.go
jsafrane: Should it get Conflict error then and the operator re-tries shortly?

I did not notice any special logic when removing finalziers elsewhere in library-go
https://github.com/openshift/library-go/blob/17708d019f9cb170ff1f31265d5acfa82b9a82e9/pkg/operator/encryption/controllers/prune_controller.go#L154-L158

Or do you mean `RetryOnConflict` loop?
https://github.com/openshift/library-go/blob/17708d019f9cb170ff1f31265d5acfa82b9a82e9/pkg/operator/secretspruner/prune_controller.go#L86-L87

pkg/operator/management/management_state.go
jsafrane: Done

pkg/operator/management/management_state.go
jsafrane: That's a bigger refactoring. Will try something (`OperatorType` struct with `IsOperatorAlwaysManaged()` and similar functions?)

jsafrane: I am afraid this may be too big for this PR. `IsOperatorManaged` is called deep in other controllers, for example in this "base" controller that's used in 5 other encryption controllers

https://github.com/openshift/library-go/blob/17708d019f9cb170ff1f31265d5acfa82b9a82e9/pkg/operator/encryption/controllers/controller.go#L30-L30

Passing `OperatorType` down there seems possible, but it will be quite bug PR on its own.

jsafrane:
> Why return a fake instance instead of simply having the consumers handle a 404?

To make other controllers (e.g. logging, management state) to stop logging weird errors and constantly re-queuing new sync().

jsafrane:
/close
Trying less intrusive changes in #1029, but it's WIP.

PR #1018, opened 2021-03-17
 add minreadyseconds concept to static pod operator

deads2k:
// minReadySeconds is the time to wait between the completion of an operand becoming ready (all containers ready)
	// and starting the rollout onto the next node.  This avoids a problem with an external load balancer that looks like
	//  1. for some reason we have two instances, maybe a liveness check blipped on one node. it doesn't matter why
	//  2. we bring down an instance on m0 to start a new revision
	//  3. at this point we have one instance running on m1
	//  4. m0 starts up and goes ready, but the LB ready check just timed out and is waiting for X seconds
	//  5. we bring down an instance on m1 to start the new revision.
	//  6. the LB thinks all backends are down and routes randomly
	//  7. no profit.
	// setting this field to 30s can prevent the kube-apiserver from triggering the above flow on AWS.

/hold for proof

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -8,21 +8,10 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/davecgh/go-spew/spew"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/informers"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
-	"k8s.io/klog/v2"
-
 	operatorv1 "github.com/openshift/api/operator/v1"
-
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/condition"
 	"github.com/openshift/library-go/pkg/operator/events"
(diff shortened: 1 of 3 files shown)


sttts:
Great find!

p0lyn0mial:
"Anything that can go wrong will go wrong" :)

p0lyn0mial:
>  6. the LB thinks all backends are down and routes randomly

is this documented behavior?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -262,6 +281,41 @@ func nodeToStartRevisionWith(ctx context.Context, getStaticPodState func(ctx con
 	return 0, reason, nil
 }
 
+// determine the amount of time to delay before creating the next installer pod.  We delay to avoid
```
p0lyn0mial, line 284: nit: fix the gdoc

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -284,6 +338,15 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
...
+	// determine the amount of time to delay before creating the next installer pod.  We delay to avoid an LB outage (see godoc on minReadySeconds)
+	sleepTime := c.timeToWaitBeforeInstallingNextPod(ctx, operatorStatus.NodeStatuses)
+	if sleepTime > 0 {
+		select {
+		case <-ctx.Done():
+		case <-time.After(sleepTime):
+		}
+	}
+
 	for l := 0; l < len(operatorStatus.NodeStatuses); l++ {
```
p0lyn0mial, line 350: the code below ensures that only one node can transition to a new revision at a time, right?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -262,6 +281,41 @@ func nodeToStartRevisionWith(ctx context.Context, getStaticPodState func(ctx con
...
+			continue
+		}
+		for _, podCondition := range pod.Status.Conditions {
+			if podCondition.Type != corev1.PodReady {
+				continue
+			}
+			if podCondition.Status != corev1.ConditionTrue {
+				continue
+			}
+			durationPodHasBeenReady := c.clock.Now().Sub(podCondition.LastTransitionTime.Time)
```
p0lyn0mial, line 304: okay, it takes `min` from all ready nodes.

deads2k:
> > 1. the LB thinks all backends are down and routes randomly
> 
> is this documented behavior?

Uh, probably documented?  It either routes to all or routes to none. Both outcomes are bad.

pkg/operator/staticpod/controller/installer/installer_controller.go
deads2k: > the code below ensures that only one node can transition to a new revision at a time, right?

I can find a ref if you want it, but yes, we check the previous installer pod status

p0lyn0mial: just wanted to make sure that `durationPodHasBeenReady` counting makes sense. thanks.

deads2k:
proof worked, godoc updated.

/hold cancel

PR #1019, opened 2021-03-19
Migrate rsa keypair validation and generation from ckao and ckcmo

marun:
This functionality is common between ckao and kcmo and moving it to lib-go ensures that the maintenance can be shared.

/cc @stlaz @sttts @soltysh 

Canaries: 
 - https://github.com/openshift/cluster-kube-apiserver-operator/pull/1072
 - https://github.com/openshift/cluster-kube-controller-manager-operator/pull/512

### DIFF
--- a/pkg/operator/encryption/crypto/rsa.go
+++ b/pkg/operator/encryption/crypto/rsa.go
@@ -1,0 +1,76 @@
+package crypto
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"reflect"
+
+	"k8s.io/client-go/util/keyutil"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/crypto/rsa.go
```diff
@@ -0,0 +1,76 @@
...
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"reflect"
+
+	"k8s.io/client-go/util/keyutil"
+)
+
+// Size for generated RSA key pairs
+const keySize = 4096
```
soltysh, line 15: Any particular reason we're bumping this to 4096 vs 2048 which is currently used in both operators? I'm mainly asking about backwards compatibility?

soltysh (approved):
One question.
/lgtm 
otherwise

PR #1020, opened 2021-03-19
shell args

p0lyn0mial:
This PR provides convenient functions for dealing with command-line arguments.

### DIFF
--- a/pkg/operator/v1helpers/args.go
+++ b/pkg/operator/v1helpers/args.go
@@ -1,0 +1,61 @@
+package v1helpers
+
+import (
+	"fmt"
+	"sort"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// FlagsFromUnstructured process the unstructured arguments usually retrieved from an operator's configuration file under a specific key.
+// There are only two supported/valid types for arguments, that is []sting and/or string.
+// Passing a different type yield an error.
+//
+// Use ToFlagSlice function to get a slice of string flags.
+func FlagsFromUnstructured(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
+	return flagsFromUnstructured(unstructuredArgs)
+}
+
+// ToFlagSlice transforms the provided arguments to a slice of string flags.
+// A flag name is taken directly from the key and the value is simply attached.
+// A flag is repeated iff it has more than one value.
+func ToFlagSlice(args map[string][]string) []string {
+	var keys []string
+	for key := range args {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	var flags []string
+	for _, key := range keys {
+		for _, token := range args[key] {
+			flags = append(flags, fmt.Sprintf("--%s=%s", key, token))
+		}
+	}
+	return flags
+}
+
+// flagsFromUnstructured process the unstructured arguments (interface{}) to a map of strings.
+// There are only two supported/valid types for arguments, that is []sting and/or string.
+// Passing a different type yield an error.
+func flagsFromUnstructured(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
+	ret := map[string][]string{}
+	for argName, argRawValue := range unstructuredArgs {
+		var argsSlice []string
+		var found bool
+		var err error
+
+		argsSlice, found, err = unstructured.NestedStringSlice(unstructuredArgs, argName)
+		if !found || err != nil {
+			str, found, err := unstructured.NestedString(unstructuredArgs, argName)
+			if !found || err != nil {
+				return nil, fmt.Errorf("unable to process an argument, incorrect value %v under %v key, expected []string or string", argRawValue, argName)
+			}
+			argsSlice = append(argsSlice, str)
+		}
+
+		ret[argName] = argsSlice
+	}
+
+	return ret, nil
+}
--- a/pkg/operator/v1helpers/args_test.go
+++ b/pkg/operator/v1helpers/args_test.go
@@ -1,0 +1,104 @@
+package v1helpers
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/util/diff"
+)
+
+func TestFlagsFromUnstructured(t *testing.T) {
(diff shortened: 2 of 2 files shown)


p0lyn0mial:
/hold

need unit test for `ToShellEscapedArguments`

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// shellEscapePattern determines if a string should be enclosed in single quotes
+// so that it can safely be passed to shell command line.
+var shellEscapePattern *regexp.Regexp
+
+func init() {
+	// some characters have special meaning to the shell and may need to be escaped
+	// the following regexp defines safe characters that don't require escaping
```
sttts, line 18: where is this from?

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+func ToShellEscapedArguments(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
+	return toStructuredArguments(unstructuredArgs, true)
+}
+
+// ToStructuredArguments process the unstructured arguments usually retrieved from an operator's configuration file under a specific key.
+// There are only two supported/valid types for arguments, that is []sting or string.
+// Passing a different type yeld an error.
+//
+// Use ToFlagSlice function to get a slice of string flags.
+func ToStructuredArguments(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
```
sttts, line 35: `FlagsFromUnstructured` would say more, no?

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+				return nil, fmt.Errorf("unable to process an argument, incorrect value %v under %v key, expected []string or string", argRawValue, argName)
+			}
+			argsSlice = append(argsSlice, str)
+		}
+
+		escapedArgsSlice := make([]string, len(argsSlice))
+		for index, str := range argsSlice {
+			arg := str
+			if escape {
+				arg = maybeQuote(str)
```
sttts, line 93: escaping here is at the wrong spot, isn't it? You should have a `ToShellFlagString` func that does the escaping.

sttts: ... and that only returns one string.

p0lyn0mial: good catch, thx!

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+	// some characters have special meaning to the shell and may need to be escaped
+	// the following regexp defines safe characters that don't require escaping
+	shellEscapePattern = regexp.MustCompile(`[^\w@%+=:,./-]`)
+}
+
+// FlagsFromUnstructuredEscaped process the given arguments and determines if they should be escaped so that they can be safely passed as shell command line arguments.
+// unstructuredArgs holds unprocessed arguments usually retrieved from an operator's configuration file under a specific key.
+//
+// Use ToFlagSlice function to get a slice of string flags.
+func FlagsFromUnstructuredEscaped(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
```
sttts, line 26: I still think this is the wrong spot for escaping. Escape when turning a `[]string` into *one* string.

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// shellEscapePattern determines if a string should be enclosed in single quotes
+// so that it can safely be passed to shell command line.
+var shellEscapePattern *regexp.Regexp
+
+func init() {
+	// some characters have special meaning to the shell and may need to be escaped
+	// the following regexp defines safe characters that don't require escaping
+	shellEscapePattern = regexp.MustCompile(`[^\w@%+=:,./-]`)
```
mfojtik, line 19: this looks strange, why is this in init and not in var declaration?

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+func FlagsFromUnstructured(unstructuredArgs map[string]interface{}) (map[string][]string, error) {
+	return flagsFromUnstructured(unstructuredArgs, false)
+}
+
+// ToFlagSlice transforms the provided arguments to a slice of string flags.
+// A flag name is taken directly from the key and the value is simply attached.
+// A flag is repeated iff it has more than one value.
+func ToFlagSlice(args map[string][]string) []string {
+	var keys []string
+	for key := range args {
```
mfojtik, line 44: won't `keys := append([]string{}, args...)` do the same?

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
+
+	var flags []string
+	for _, key := range keys {
+		for _, token := range args[key] {
+			flags = append(flags, fmt.Sprintf("--%s=%v", key, token))
```
mfojtik, line 52: s/%v/%s/ ?

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+		return "''"
+	}
+	if shellEscapePattern.MatchString(s) {
+		return "'" + strings.Replace(s, "'", "'\"'\"'", -1) + "'"
+	}
+
+	return s
+}
+
+func flagsFromUnstructured(unstructuredArgs map[string]interface{}, escape bool) (map[string][]string, error) {
```
mfojtik, line 73: nit: this is long and complicated enough to deserve one-line comment :-)

pkg/operator/v1helpers/args.go
```diff
@@ -0,0 +1,103 @@
...
+	ret := map[string][]string{}
+	for argName, argRawValue := range unstructuredArgs {
+		var argsSlice []string
```
mfojtik, line 76: nit: group this?

pkg/operator/v1helpers/args.go
p0lyn0mial: `args` is a map, we need to iterate to get the keys.

PR #1021, opened 2021-03-22
DO NOT MERGE: Support dynamic fallback down registry chains

smarterclayton:
Implements the core of the discussion around supporting multiple
sources of input. Each wrapped method in the mirroredRepository
structure that needs to check alternates (methods that invoke digest
lookups) will use that structure. Methods that cannot be retried
will use the first alternative that has a working client (such as
ServeBlob). Methods that must always use the source (mutable calls
or those that don't deal with a digest) will only look at the
original source, although the logic is structured so we can use
data about these requests in the future.

The strategy is clarified to call either FirstRequest and OnFailure,
or just FirstRequest. Each repository will make a single call to the
strategy, and it's up to the strategy to include the original source
when invoking FirstRequest.

Did a very, very, very minimal test of this against docker.io but
did not check anything except get.


pkg/image/registryclient/client_mirrored.go
```diff
@@ -0,0 +1,477 @@
...
+}
+
+var (
+	errNoValidAlternates = fmt.Errorf("no valid alterative sources for this content located")
+	errNoValidSource     = fmt.Errorf("no source repository defined for accessing the repository")
+)
+
+// initialRepos returns a list of locations to attempt to access, a boolean indicating that alternates
+// were suggested, or an error.
+func (r *mirroredRepository) initialRepos(ctx context.Context) ([]reference.DockerImageReference, bool, error) {
```
dmage, line 92: You can safely ditch this bool without sacrificing performance/readability.

pkg/image/registryclient/client_mirrored.go
```diff
@@ -0,0 +1,477 @@
...
+	if err := r.attemptRepos(ctx, repos, fn); err != nil {
+		if loaded {
+			return err
+		}
+		alternates, err := r.errorRepos(ctx)
+		if err != nil {
+			return err
+		}
+		if alternateErr := r.attemptRepos(ctx, alternates, fn); alternateErr != nil {
+			return err
```
dmage, line 189: err is nil here

pkg/image/registryclient/client_test.go
```diff
@@ -841,3 +848,132 @@ func (s *fakeBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter, r
...
+}
+
+func TestMirroredRegistry_ManifestGet(t *testing.T) {
+	opt := distribution.WithManifestMediaTypes([]string{
+		manifestlist.MediaTypeManifestList,
+		schema2.MediaTypeManifest,
+	})
+
+	klog.InitFlags(flag.CommandLine)
+	//cliflag.InitFlags()
```
dmage, line 923: WIP?

dmage (commented):
Honestly I don't like that it lookups alternatives from scratch from every method. If I want to pull (mirror) an image, I want to get all its blobs from the same repository. Most likely it'll be the first suggested repository and this approach will work, but what if the first mirror is down?
Otherwise, looks good, nice catch that write methods shouldn't use mirrors.

smarterclayton:
> Honestly I don't like that it lookups alternatives from scratch from every method. If I want to pull (mirror) an image, I want to get all its blobs from the same repository. Most likely it'll be the first suggested repository and this approach will work, but what if the first mirror is down?

It should only lookup alternatives one time per Repository() object returned from the main call (maybe a bug if I'm looking wierdly)

pkg/image/registryclient/client_mirrored.go
```diff
@@ -0,0 +1,477 @@
...
+// mirroredTags lazily accesses the source repository
+type mirroredTags struct {
+	repo *mirroredRepository
+}
+
+var _ distribution.TagService = mirroredTags{}
+
+func (f mirroredTags) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
+	var desc distribution.Descriptor
+	err := f.repo.source(ctx, func(r distribution.Repository) error {
```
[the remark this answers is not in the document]
smarterclayton: no because you don't know whether they are in sync.  I think that would be a different type of behavior - we could call this "blobMirrored" to clarify because everything has to be content addressable.

pkg/image/registryclient/client_mirrored.go
smarterclayton: It's subtle and necessary - the presence of the bool tells us whether the FirstRequest() gave us alternates or not.  If it did, we never call OnFailure().  If it doesn't, we have to call OnFailure later.  The difference is because the returned list of alternates always contains the source in the default path, so an empty alternates list (or nil) doesn't tell us what expected outcome from FirstRequest was().

pkg/image/registryclient/client_test.go
smarterclayton: Was just hacking it in while i debugged code, would not go in to final.

pkg/image/registryclient/client_mirrored.go
smarterclayton: yikes good catch.

pkg/image/registryclient/client_mirrored.go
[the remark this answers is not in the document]
smarterclayton: Renamed to make it a bit clearer and updated the godoc on the top most item.

pkg/image/registryclient/client_mirrored.go
smarterclayton: renamed it `attemptErr`

pkg/image/registryclient/client_mirrored.go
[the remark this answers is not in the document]
sallyom: i've been stuck w/ the repository going from mirrored(blobMirroredRepository) to retryRepository w/ client.go `connectToRegistry` and when I try to mirror any image I see a manifestUnknown w/ the manifests Put  - If I don't fix this I'll push my oc branch and have you try it/help tomorrow. (first I'll incorporate the name changes above)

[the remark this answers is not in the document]
sallyom: ooooh got it to work - `oc` was considering a nil manifest and an empty Descriptor as 'already existing' once I fixed that the push goes through.. now, why is the regclient returning a nil manifest and empty Descriptor?

PR #1023, opened 2021-03-23
resourcemerge: set LastGeneration to 1 when no previous instances exist

acardace:
The ObjecMeta.generation field is set to 1 when a resource is first
created, if no previous generation of the same resource exists
it's better to init LastGeneration to 1 otherwise the resource might get patched needlessly
right after creation.

Signed-off-by: Antonio Cardace <acardace@redhat.com>


acardace:
This happens for example in Kubevirt https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-operator/resource/apply/admissionregistration.go#L76, where the new generation is created with a newly created resource that has just been posted to the API Server, obviously, the in-memory resource has ObjectMeta.generation initialized to 0.

This causes the first next iteration of the sync loop to patch the resource needlessly because after GETting the resource it sees `generation = 1` and `expectedGeneration = 0`.

acardace:
I actually noticed there is a bug in kubevirt rather than in this library.

PR #1024, opened 2021-03-23
make the union conditions sort stable

deads2k:
We union conditions, but we forgot to sort the constituent conditions to have a stable message.  This causes lots of extra updates.


sanchezl (approved):
/lgtm

PR #1025, opened 2021-03-23
add Default Audit Policy getter function

EmilyM1:
Currently our cluster-kube-apiserver-operator does not use library go audit pkg and uses internal static files.
This is the addition of missing functions to library-go so that the package of library-go audit pkg will work.
Essentially, this fixes build errors.

### DIFF
--- a/pkg/operator/apiserver/audit/audit_policies.go
+++ b/pkg/operator/apiserver/audit/audit_policies.go
@@ -22,6 +22,21 @@ const (
 	auditPolicyAsset = "pkg/operator/apiserver/audit/manifests/audit-policies-cm.yaml"
 )
 
+// DefaultPolicy brings back the default.yaml audit policy to init the api
+func DefaultPolicy() ([]byte, error) {
+	// none is used on target name and namespace as it's not a key in the default.yaml
+	cm, err := GetAuditPolicies("none", "none")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retreive audit policies - %w", err)
+	}
+
+	cmData := []byte(cm.Data["default.yaml"])
+	if len(cmData) == 0 {
+		return nil, errors.New("failed to locate the default policy from configmap")
+	}
+	return cmData, nil
+}
+
 // WithAuditPolicies is meant to wrap a standard Asset function usually provided by an operator.
 // It delegates to GetAuditPolicies when the filename matches the predicate for retrieving a audit policy config map for target namespace and name.
 func WithAuditPolicies(targetName string, targetNamespace string, assetDelegateFunc resourceapply.AssetFunc) resourceapply.AssetFunc {
--- a/pkg/operator/apiserver/audit/audit_policies_test.go
+++ b/pkg/operator/apiserver/audit/audit_policies_test.go
@@ -155,6 +155,29 @@ func TestNewAuditPolicyPathGetter(t *testing.T) {
 	}
 }
 
+func TestDefaultPolicy(t *testing.T) {
+	scenarios := []struct {
+		name string
+	}{
+		{
+			name: "Get default audit policy for the kube-apiserver",
+		},
+	}
+	for _, test := range scenarios {
+		t.Run(test.name, func(t *testing.T) {
+			// act
+			data, err := DefaultPolicy()
+			// assert
+			if err != nil {
+				t.Errorf("expected no error, but got: %v", err)
+			}
+			if len(data) == 0 {
+				t.Error("expected a non empty default policy")
+			}
+		})
+	}
+}
+
 func readBytesFromFile(t *testing.T, filename string) []byte {
 	file, err := os.Open(filename)
 	if err != nil {


pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,39 @@ const (
 	auditPolicyAsset = "pkg/operator/apiserver/audit/manifests/audit-policies-cm.yaml"
 )
 
+func DefaultPolicy() ([]byte, error) {
+	return getPolicyFromResource("default.yaml")
```
tkashem, line 26: `getPolicyFromResource` does not resolve the template varaibles. So we need to use something similar to the beow snippet

```
func DefaultPolicy() ([]byte, error) {
  cm, err := GetAuditPolicies("kube-apiserver-audit-policies", "openshift-kube-apiserver")
  cm.Data["default.yaml"]
}
```

we need to handle error appropriately and a unit test.

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,39 @@ const (
...
+}
+
+func getPolicyFromResource(key string) ([]byte, error) {
```
tkashem, line 29: we can remove it

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,39 @@ const (
 	auditPolicyAsset = "pkg/operator/apiserver/audit/manifests/audit-policies-cm.yaml"
 )
 
+func DefaultPolicy() ([]byte, error) {
```
tkashem, line 25: if we want to keep it library-go, maybe make it usable by other operators:
```
func GetDefaultPolicy(name, namespace string) ([]byte, error)
```

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,21 @@ const (
 	auditPolicyAsset = "pkg/operator/apiserver/audit/manifests/audit-policies-cm.yaml"
 )
 
+// DefaultPolicy brings back the default.yaml audit policy to init the api
+func DefaultPolicy() ([]byte, error) {
+	cm, err := GetAuditPolicies("kube-apiserver-audit-policies", "openshift-kube-apiserver")
```
p0lyn0mial, line 27: `cluster-kube-apiserver-operator` could use `GetAuditPolicies` directly :)

tkashem, line 27:
```
cm, err := GetAuditPolicies("kube-apiserver-audit-policies", "openshift-kube-apiserver")
```

this makes it specific to kas operator, in library-go we usually put logic that can be shared by all operators. My original suggestion was to pass the name and namespace as parameters:
```
func GetDefaultPolicy(name, namespace string) ([]byte, error) {
   cm, err := GetAuditPolicies(name, namespace)
}
```

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,21 @@ const (
...
+
+		if err != nil {
+		return nil, fmt.Errorf("failed to retreive audit policies", err)
+	}
+	cmData, err := []byte(cm.Data["default.yaml"]), err
```
tkashem, line 32: is what you have equivalent to the snippets below?
```
defaultData, ok := cm.Data["default.yaml"]
if !ok || len(defaultData) == 0 {
  return nil, fmt.Errorf("failed to retrieve the default policy", err)
}
```

pkg/operator/apiserver/audit/audit_policies_test.go
```diff
@@ -155,6 +155,29 @@ func TestNewAuditPolicyPathGetter(t *testing.T) {
...
+			name: "Get default audit policy",
+			targetNamespace: "ScenarioOne",
+			targetName:      "audit",
+			goldenFile:      "./testdata/audit-policies-cm-scenario-1.yaml",
+			expectedPath: "/var/run/configmaps/audit/default.yaml"
+		},
+	}
+	{
+		// act
+		cm, err := GetAuditPolicies(scenarios.targetName, scenarios.targetNamespace)
```
tkashem, line 176: we are adding a new function `DefaultPolicy` which should be tested I believe, not `GetAuditPolicies`.

pkg/operator/apiserver/audit/audit_policies.go
tkashem: yeah, I agree, if we want to keep it in library-go we will need to make it reusable:
```
func DefaultPolicy(name, namespace string) ([]byte, error)
```

pkg/operator/apiserver/audit/audit_policies.go
EmilyM1: Thanks for the rec here!

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,23 @@ const (
 	auditPolicyAsset = "pkg/operator/apiserver/audit/manifests/audit-policies-cm.yaml"
 )
 
+// DefaultPolicy brings back the default.yaml audit policy to init the api
+func DefaultPolicy() ([]byte, error) {
+	// none is used on target name and namespace as it's not a key in the default.yaml
+	cm, err := GetAuditPolicies("none", "none")
+
```
tkashem, line 29: nit: do we need this extra line ^?

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,23 @@ const (
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to retreive audit policies - %w", err)
+	}
+
+	cmData := []byte(cm.Data["default.yaml"])
+
```
tkashem, line 35: nit: do we need this extra line ^?

pkg/operator/apiserver/audit/audit_policies_test.go
```diff
@@ -155,6 +155,32 @@ func TestNewAuditPolicyPathGetter(t *testing.T) {
 	}
 }
 
+func TestDefaultPolicy(t *testing.T) {
+	scenarios := []struct {
+		name string
+	}{
+		{
+			name: "Get default audit policy for the kube-apiserver",
+		},
```
tkashem, line 164: we don't need the following test case, it has the same setup as ^

pkg/operator/apiserver/audit/audit_policies.go
EmilyM1: sure will remove rn

pkg/operator/apiserver/audit/audit_policies.go
EmilyM1: will remove

pkg/operator/apiserver/audit/audit_policies_test.go
EmilyM1: removed

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -22,6 +22,21 @@ const (
...
+func DefaultPolicy() ([]byte, error) {
+	// none is used on target name and namespace as it's not a key in the default.yaml
+	cm, err := GetAuditPolicies("none", "none")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retreive audit policies - %w", err)
+	}
+
+	cmData := []byte(cm.Data["default.yaml"])
+	if len(cmData) == 0 {
+		return nil, errors.New("failed to locate the default policy from configmap")
```
sttts, line 35: can this happen? Can't we make sure with a test that `default.yaml` always available?

EmilyM1: Test is there previously:
https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/audit/audit_policies_test.go#L102

sttts: I think I can live with this. I would have preferred no error in the signature, but that's just taste.

sttts:
Please be clear in ckao that the default.yaml policy is fine because it is the safe one (= is secure even with non-sha256 tokens in the system), and we are fine with it in the bootstrapping phase. We will switch to the other default policy for sha256-only tokens when @stlaz's PRs have merged to remove old tokens.

PR #1029, opened 2021-03-25
Add support for removable csi drivers

jsafrane:
This is minimal alternative of https://github.com/openshift/library-go/pull/1013. Only the controllers that are used by CSI driver operators are changed

Goal: Add support for removable operators managed by OLM, namely CSI drivers. Operands (a CSI driver) should be installed when user creates a CR and removed when user deletes it.

* Most library-go controllers log error / warning messages when the operator CR is missing. I updated only the controllers that are used in CSI driver operators to handle missing CRs and return silently.
* Reusing `management.SetOperatorNotRemovable()` to tell CSI controllers to support removing the operands.
* Removable operators may need to add/remove finalizers in the operator CR. I added a new interface `OperatorClientWithFinalizers` that can add/remove finalizers as the controller manage the operands / finish their removal. I avoided `OperatorClient` changes, because it's implemented on too many places.
  *  Again, updated only CSI controllers to add/remove finalizers for removable CSI drivers.

TODO:
  * Fix [`baseController.reportDegraded`](https://github.com/openshift/library-go/blob/4a361189f3da6f13a2a4b5c8fb8bdae68842f791/pkg/controller/factory/base_controller.go#L234) not to log errors / requeue sync with exp. backoff when the operator CR is missing.
      * CSI controllers often use `Factory.WithSyncDegradedOnError()` to simplify error handling.
      * `baseController.reportDegraded` can't use `management.IsOperatorRemovable()` to ignore NotFound for removable operators because of import loop. Shall I move `managementstate.go` to a different package?

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/robfig/cron"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -15,7 +16,7 @@ import (
 	"k8s.io/klog/v2"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-
+	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
@@ -202,7 +203,12 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 // reconcile wraps the sync() call and if operator client is set, it handle the degraded condition if sync() returns an error.
 func (c *baseController) reconcile(ctx context.Context, syncCtx SyncContext) error {
 	err := c.sync(ctx, syncCtx)
-	return c.reportDegraded(ctx, err)
+	degradedErr := c.reportDegraded(ctx, err)
+	if apierrors.IsNotFound(degradedErr) && management.IsOperatorRemovable() {
+		// The operator tolerates missing CR, therefore don't report it up.
+		return err
+	}
+	return degradedErr
 }
 
 // degradedPanicHandler will go degraded on failures, then we should catch potential panics and covert them into bad status.
--- a/pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
+++ b/pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
@@ -28,7 +28,7 @@ import (
 // <name>Degraded: produced when the sync() method returns an error.
 type CredentialsRequestController struct {
 	name            string
-	operatorClient  v1helpers.OperatorClient
+	operatorClient  v1helpers.OperatorClientWithFinalizers
 	targetNamespace string
 	manifest        []byte
 	dynamicClient   dynamic.Interface
@@ -40,7 +40,7 @@ func NewCredentialsRequestController(
 	targetNamespace string,
 	manifest []byte,
 	dynamicClient dynamic.Interface,
-	operatorClient v1helpers.OperatorClient,
+	operatorClient v1helpers.OperatorClientWithFinalizers,
 	recorder events.Recorder,
 ) factory.Controller {
 	c := &CredentialsRequestController{
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
@@ -5,11 +5,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
-
 	"github.com/openshift/library-go/pkg/controller/factory"
 	dc "github.com/openshift/library-go/pkg/operator/deploymentcontroller"
 	"github.com/openshift/library-go/pkg/operator/events"
-
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
@@ -46,6 +44,8 @@ import (
 // The placeholder ${CLUSTER_ID} specified in the static file is replaced with the cluster ID (sometimes referred as infra ID).
 // This is mostly used by CSI drivers to tag volumes and snapshots so that those resources can be cleaned up on cluster deletion.
 //
+// This controller supports removable operands, as configured in pkg/operator/management.
+//
 // This controller produces the following conditions:
 //
 // <name>Available: indicates that the CSI Controller Service was successfully deployed and at least one Deployment replica is available.
@@ -56,7 +56,7 @@ func NewCSIDriverControllerServiceController(
 	name string,
 	manifest []byte,
 	recorder events.Recorder,
-	operatorClient v1helpers.OperatorClient,
+	operatorClient v1helpers.OperatorClientWithFinalizers,
 	kubeClient kubernetes.Interface,
 	deployInformer appsinformersv1.DeploymentInformer,
 	configInformer configinformers.SharedInformerFactory,
@@ -67,5 +67,4 @@ func NewCSIDriverControllerServiceController(
 	var optionalManifestHooks []dc.ManifestHookFunc
 	optionalManifestHooks = append(optionalManifestHooks, WithPlaceholdersHook(configInformer))
 	return dc.NewDeploymentController(name, manifest, recorder, operatorClient, kubeClient, deployInformer, optionalInformers, optionalManifestHooks, optionalDeploymentHooks...)
-
 }
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
@@ -142,8 +142,7 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
 func (c *CSIDriverNodeServiceController) syncManaged(ctx context.Context, opSpec *opv1.OperatorSpec, opStatus *opv1.OperatorStatus, syncContext factory.SyncContext) error {
 	klog.V(4).Infof("syncManaged")
 	if management.IsOperatorRemovable() {
-		err := c.operatorClient.EnsureFinalizer(c.name)
-		if err != nil {
+		if err := v1helpers.EnsureFinalizer(c.operatorClient, c.name); err != nil {
 			return err
 		}
 	}
@@ -274,9 +273,5 @@ func (c *CSIDriverNodeServiceController) syncDeleting(ctx context.Context, opSpe
 	}
 
 	// All removed, remove the finalizer as the last step
-	err = c.operatorClient.RemoveFinalizer(c.name)
-	if err != nil {
-		return err
-	}
-	return nil
+	return v1helpers.RemoveFinalizer(c.operatorClient, c.name)
 }
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
@@ -8,19 +8,21 @@ import (
 	"strings"
 	"time"
 
-	appsv1 "k8s.io/api/apps/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
-	"k8s.io/client-go/kubernetes"
-
 	opv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
+	"github.com/openshift/library-go/pkg/operator/management"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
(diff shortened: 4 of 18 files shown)


pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -217,6 +231,50 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
...
+	var clusterID string
+	if c.configInformer != nil {
+		infra, err := c.configInformer.Config().V1().Infrastructures().Lister().Get(infraConfigName)
+		if err != nil {
+			return nil, err
+		}
+		clusterID = infra.Status.InfrastructureName
+	}
+
+	manifest := replacePlaceholders(c.manifest, opSpec, clusterID)
```
deads2k, line 266: signature changed to include a clusterid?

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -217,6 +231,50 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
...
+
+	// All removed, remove the finalizer as the last step
+	err = c.operatorClient.RemoveFinalizer(c.name)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *CSIDriverControllerServiceController) getDeployment(opSpec *opv1.OperatorSpec) (*appsv1.Deployment, error) {
```
deads2k, line 256: @ravisantoshgudimetla you had a PR trying to collapse this functionality for deployments.  Is it ready to merge?  You're racing this PR now.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -227,3 +258,25 @@ func replacePlaceholders(manifest []byte, spec *opv1.OperatorSpec) []byte {
...
+	}
+
+	err = c.kubeClient.AppsV1().DaemonSets(required.Namespace).Delete(ctx, required.Name, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	} else {
+		klog.V(2).Infof("Deleted DaemonSet %s/%s", required.Namespace, required.Name)
+	}
+
+	// All removed, remove the finalizer as the last step
```
deads2k, line 276: is the removal pattern actually deleting the daemonset or do we want to remove the entire namespace?

deads2k:
The idea here is reasonable.  I'd like to see the deployment and daemonset controllers extracted to be generic first.  @ravisantoshgudimetla had a PR, but it needs to merge without growing in size or it will never land due to risk.

It does seem to me that you have three steps
1. make existing controllers stop their actions.
2. make some controller (potentially a new one, possibly some of the existing ones), delete the entire namespace.
3. make sure all the content is removed before removing the finalizer.

jsafrane:
Sure, I can wait for @ravisantoshgudimetla. 

> 2. make some controller (potentially a new one, possibly some of the existing ones), delete the entire namespace.

All our CSI drivers are in `openshift-cluster-csi-drivers` namespace and we don't want to delete it. Other operators can implement their own logic and e.g. remove a namespace, but we don't need it for our operators.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
jsafrane: not in this PR...

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
jsafrane: Deleting the daemonset only.

jsafrane:
I've been thinking about removing namespaces again and another argument *not* to do it is OLM - OLM creates a namespace for the operator and then either user chooses namespace of the operand (by crating operator's CR in the namespace)  or the operator chooses the namespace and in CSI case it's going to be `openshift-cluster-csi-drivers`. In both cases the operator expects that the namespace already exists and IMO it should not delete it.

jsafrane:
Rebased to current library-go with DeploymentController.

pkg/controller/factory/base_controller.go
```diff
@@ -7,7 +7,9 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/management"
```
mfojtik, line 10: nit: sort import

pkg/controller/factory/base_controller.go
```diff
@@ -202,7 +204,12 @@ func (c *baseController) runWorker(queueCtx context.Context) {
 // reconcile wraps the sync() call and if operator client is set, it handle the degraded condition if sync() returns an error.
 func (c *baseController) reconcile(ctx context.Context, syncCtx SyncContext) error {
 	err := c.sync(ctx, syncCtx)
-	return c.reportDegraded(ctx, err)
+	degradedErr := c.reportDegraded(ctx, err)
```
mfojtik, line 207: can you split this out to separate PR? I can approve it right now

pkg/controller/factory/base_controller.go
jsafrane: fixed

pkg/controller/factory/base_controller.go
jsafrane: Not sure what to split. Handling of `management.IsOperatorRemovable()`? IMO it belongs here, where the other controllers are updated too.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
jsafrane: It has been merged

ravisantoshgudimetla (approved):
> In both cases the operator expects that the namespace already exists and IMO it should not delete it.

FWIW, if you're creating namespace via staticresource controllers, we can have [conditionals going forward ](https://github.com/openshift/library-go/pull/1095/commits/59ec4d268990a31f4c0604ce2dc2c509ab5cac2c ) which can help in removing the namespace based on certain conditions. 

But the PR as is LGTM. Once this merges, I can open DS controller pod which you can reuse @jsafrane

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -114,27 +118,41 @@ func (c *CSIDriverNodeServiceController) Name() string {
...
+	}
+	return c.syncManaged(ctx, opSpec, opStatus, syncContext)
+}
 
-	for i := range c.optionalDaemonSetHooks {
-		err := c.optionalDaemonSetHooks[i](opSpec, required)
+func (c *CSIDriverNodeServiceController) syncManaged(ctx context.Context, opSpec *opv1.OperatorSpec, opStatus *opv1.OperatorStatus, syncContext factory.SyncContext) error {
+	klog.V(4).Infof("syncManaged")
+	if management.IsOperatorRemovable() {
+		err := c.operatorClient.EnsureFinalizer(c.name)
```
deads2k, line 145: nice.  We're allowed to have something.openshift.io/something, right? Will this be easier as `removal.operator.openshift.io/name`?  that fans out cleanly and doesn't burn a "good" name.

deads2k: do we need to encode an overall operator name in there too?  We haven't made that concept so far, but we could to distinguish which operator needs to act.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
```diff
@@ -156,6 +157,68 @@ func (c dynamicOperatorClient) UpdateOperatorStatus(resourceVersion string, stat
...
+		return err
+	}
+	klog.V(2).Infof("Removed finalizer %q", finalizer)
+	return nil
+}
+
+func (c dynamicOperatorClient) saveFinalizers(instance *unstructured.Unstructured, finalizers []string) error {
+	clone := instance.DeepCopy()
+	clone.SetFinalizers(finalizers)
+	_, err := c.client.Update(context.TODO(), clone, metav1.UpdateOptions{})
```
deads2k, line 218: this doesn't look right. We have an UpdateStatus that takes a func to apply a mutation against conflicts.  I think you should do the same here.

jsafrane: All other functions of `dynamicOperatorClient` do not retry on conflict, adding retry here would be odd. I added retry funcs to v1helpers, following `UpdateStatus` pattern.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
jsafrane: added `<operator name>.operator.openshift.io/<controller name>`. Where `operator_name` is either `os.Args[0]` or `$OPERATOR_NAME`.

pkg/operator/v1helpers/helpers.go
```diff
@@ -234,6 +235,40 @@ func UpdateStaticPodConditionFn(cond operatorv1.OperatorCondition) UpdateStaticP
...
+	return err
+}
+
+// getFinalizerName computes a nice finalizer name from controllerName and the operator name ($OPERATOR_NAME or os.Args[0]).
+func getFinalizerName(controllerName string) string {
+	return fmt.Sprintf("%s.operator.openshift.io/%s", getOperatorName(), controllerName)
+}
+
+func getOperatorName() string {
+	if name := os.Getenv("OPERATOR_NAME"); name != "" {
```
deads2k, line 266: I think a followup that actually plumbs the name into the controller constructors makes more sense.  We should have something in our events that reflects this content already.

I'm willing accept this so you can prove the rest, but before we ship 4.9, the env var needs to be removed.  Binaries know this information, so it's not a configuration option.

pkg/operator/genericoperatorclient/dynamic_operator_client.go
deads2k: > All other functions of `dynamicOperatorClient` do not retry on conflict, adding retry here would be odd. I added retry funcs to v1helpers, following `UpdateStatus` pattern.

sounds good.

deads2k:
I haven't reviewed the tests.  The concept looks good and the naming looks plausible.  The env var is inappropriate in the library since this isn't a configuration option: the binary must know this name and it cannot change.  I'd like to see proper plumbing in 4.9, but your credit is good enough for a followup if you like.

/approve

pkg/controller/factory/base_controller.go
```diff
@@ -202,7 +203,12 @@ func (c *baseController) runWorker(queueCtx context.Context) {
...
+	if apierrors.IsNotFound(degradedErr) && management.IsOperatorRemovable() {
+		// The operator tolerates missing CR, therefore don't report it up.
+		return err
+	}
```
bertinatto, line 210: I might be a little confused by this change, but shouldn't this check to done before `reportDegrade()`? That function sets `Degraded=true` if `err != nil`, even if `err` is `NotFound`.

Or am I missing something?

pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
```diff
@@ -40,7 +40,7 @@ func NewCredentialsRequestController(
 	targetNamespace string,
 	manifest []byte,
 	dynamicClient dynamic.Interface,
-	operatorClient v1helpers.OperatorClient,
+	operatorClient v1helpers.OperatorClientWithFinalizers,
```
bertinatto, line 43: This controller should be removed, actually. In practice, these objects shouldn't be created by any operator expect CVO. This is to support customers running CCO in [Manual Mode](https://github.com/openshift/cloud-credential-operator/blob/master/docs/mode-manual-creds.md).

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -227,3 +257,24 @@ func replacePlaceholders(manifest []byte, spec *opv1.OperatorSpec) []byte {
...
+
+	err = c.kubeClient.AppsV1().DaemonSets(required.Namespace).Delete(ctx, required.Name, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	} else {
+		klog.V(2).Infof("Deleted DaemonSet %s/%s", required.Namespace, required.Name)
+	}
+
+	// All removed, remove the finalizer as the last step
+	if err := v1helpers.RemoveFinalizer(c.operatorClient, c.name); err != nil {
```
bertinatto, line 276: nit: `return v1helpers.RemoveFinalizer()`

bertinatto (commented):
What about resources created by the static controller? Are you going to leave them behind?

pkg/operator/loglevel/logging_controller.go
```diff
@@ -40,6 +44,9 @@ func NewClusterOperatorLoggingControllerWithLogLevel(operatorClient operatorv1he
 // must be information that is logically "owned" by another component.
 func (c LogLevelController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	detailedSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if management.IsOperatorRemovable() && errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
jsafrane: fixed

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
jsafrane: fixed

jsafrane:
> What about resources created by the static controller? Are you going to leave them behind?

Yes, at least for now. For EFS, I am adding a custom controller to create the static assets: https://github.com/openshift/aws-efs-csi-driver-operator/pull/2
There is an attempt to remove the assets in https://github.com/openshift/library-go/pull/1110.

pkg/controller/factory/base_controller.go
jsafrane: `c.sync` may return `NotFound` from various reasons. `reportDegraded` returns `NotFound` only when the CR does not exist, therefore I check only that.

bertinatto (commented):
/lgtm

PR #1030, opened 2021-03-26
provides CancelRequest method for the PreferredHost RoundTripper to facilitate cancellation

p0lyn0mial:
`CancelRequest` exists to facilitate the cancellation. In general, there are at least three ways of canceling a request by an HTTP client:
 1. Transport.CancelRequest (depreciated)
 2. Request.Cancel
 3. Request.Context (preferred)

While using client-go callers can specify a timeout value that gets passed directly to an http.Client.
The HTTP client cancels requests to the underlying Transport as if the Request's Context ended.
For compatibility, the Client will also use the deprecated CancelRequest method on Transport if found.
New RoundTripper implementations should use the Request's Context for cancellation instead of implementing CancelRequest.

 Because this wrapper might be the first or might be actually wrapped with already existing wrappers that already implement CancelRequest we need to simply conform.

See for more details:
https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/staging/src/k8s.io/client-go/transport/transport.go#L257

https://github.com/kubernetes/kubernetes/blob/e29c568c4a9cd45d15665345aa015e21bcff52dd/staging/src/k8s.io/client-go/rest/config.go#L328  

https://github.com/kubernetes/kubernetes/blob/3b2746c9ea9e0fa247b01dca27634e509b385eda/staging/src/k8s.io/client-go/transport/round_trippers.go#L302

### DIFF
--- a/pkg/config/client/transport.go
+++ b/pkg/config/client/transport.go
@@ -52,14 +52,40 @@ type preferredHostRT struct {
 	preferredHostFn func() string
 }
 
-func (t *preferredHostRT) RoundTrip(r *http.Request) (*http.Response, error) {
-	preferredHost := t.preferredHostFn()
+func (rt *preferredHostRT) RoundTrip(r *http.Request) (*http.Response, error) {
+	preferredHost := rt.preferredHostFn()
 
 	if len(preferredHost) == 0 {
-		return t.baseRT.RoundTrip(r)
+		return rt.baseRT.RoundTrip(r)
 	}
 
 	r.Host = preferredHost
 	r.URL.Host = preferredHost
-	return t.baseRT.RoundTrip(r)
+	return rt.baseRT.RoundTrip(r)
+}
+
+// CancelRequest exists to facilitate cancellation.
+//
+// In general there are at least three ways of cancelling a request by an HTTP client:
+// 1. Transport.CancelRequest (depreciated)
+// 2. Request.Cancel
+// 3. Request.Context (preferred)
+//
+// While using client-go callers can specify a timeout value that gets passed directly to an http.Client.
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
/hold

for manual testing

pkg/config/client/transport.go
```diff
@@ -52,14 +52,40 @@ type preferredHostRT struct {
...
+// For compatibility, the Client will also use the deprecated CancelRequest method on Transport if found.
+// New RoundTripper implementations should use the Request's Context for cancellation instead of implementing CancelRequest.
+//
+// Because this wrapper might be the first or might be actually wrapped with already existing wrappers that already implement CancelRequest we need to simply conform.
+//
+// See for more details:
+//   https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/staging/src/k8s.io/client-go/transport/transport.go#L257
+//   https://github.com/kubernetes/kubernetes/blob/e29c568c4a9cd45d15665345aa015e21bcff52dd/staging/src/k8s.io/client-go/rest/config.go#L328
+//   https://github.com/kubernetes/kubernetes/blob/3b2746c9ea9e0fa247b01dca27634e509b385eda/staging/src/k8s.io/client-go/transport/round_trippers.go#L302
+func (rt *preferredHostRT) CancelRequest(req *http.Request) {
```
sttts, line 85: could we instead just embed the base roundtripper?
```
type preferredHostRT struct {
	http.RoundTripper
	preferredHostFn func() string
}
```

sttts: Doesn't work: https://play.golang.org/p/wCfvp5aX4qR

PR #1031, opened 2021-03-26
workloads controller: stay progressing as long as updatedReplicas != desired replicas

stlaz:
TL;DR: report progressing until all deployment replicas are updated

----

When a workload gets updated, the workload controller would report `Progressing == False` once the observed generation matches the generation of the observed deployment. However, that happens almost immediately and the deployment then still keeps updating its replicas.

While updating the replicas, the workload controller would think that the deployment is `Degraded` as it sees an unavailable replica, and would report it as such. However, an unavailable, updating replica is nothing too terrible, therefore reporting the deployment as "Progressing" sounds more natural for this process.

This PR fixes the early "Progressing=False" and unnecessary "Degraded=True" during deployment update by:
- reporting Progressing=True and Degraded=False when there are replicas for a given deployment that still need updating
- reporting Progressing=False and Degraded=True when updating the replicas is taking too much time (>5min without Progressing state change)

/assign @p0lyn0mial 
/assign @sttts 
cc @deads2k (it's the issue with tests anticipating Progressing until all apiservers rolled out we talked about yesterday)


stlaz:
proof PR: https://github.com/openshift/cluster-authentication-operator/pull/423

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -298,26 +311,28 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
...
-	}
-
 	if len(errs) > 0 {
 		return kerrors.NewAggregate(errs)
 	}
 	return nil
 }
 
+func isUpdatingTooLong(operatorClient v1helpers.OperatorClient, progressingConditionType string) (bool, error) {
+	_, status, _, err := operatorClient.GetOperatorState()
```
p0lyn0mial, line 327: could we pass and use the spec from the `sync` method to avoid one extra call to API?

stlaz: it's usually cached, but yeah, I did not notice the call up in the sync, that would be better :+1:

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -298,26 +311,28 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
...
+	if err != nil {
+		return false, err
+	}
+
+	progressing := v1helpers.FindOperatorCondition(status.Conditions, progressingConditionType)
+	return progressing != nil && time.Now().After(progressing.LastTransitionTime.Add(5*time.Minute)), nil
```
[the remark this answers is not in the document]
stlaz: good catch

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -298,26 +311,23 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
...
-
 	if len(errs) > 0 {
 		return kerrors.NewAggregate(errs)
 	}
 	return nil
 }
 
+func isUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingConditionType string) (bool, error) {
+	progressing := v1helpers.FindOperatorCondition(operatorStatus.Conditions, progressingConditionType)
+	return progressing != nil && progressing.Status == operatorv1.ConditionTrue && time.Now().After(progressing.LastTransitionTime.Add(5*time.Minute)), nil
```
sttts, line 328: these 5 min are over all instances, aren't they? If you deploy in a (soon) 5 node cluster with graceful termination we can can reach 5 min in normal operation can't we?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -251,27 +255,36 @@ func (c *Controller) updateOperatorStatus(workload *appsv1.Deployment, operatorC
...
 	}
 
+	desiredReplicas := int32(1)
+	if workload.Spec.Replicas != nil {
+		desiredReplicas = *(workload.Spec.Replicas)
+	}
+
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
+	workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas
```
sttts, line 265: this really only means that we deploy, not that we update.

pkg/operator/apiserver/controller/workload/workload.go
stlaz: it's per instance due to:
```
deploymentProgressingCondition.Message = fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation", workload.Name, c.targetNamespace,
```
which will update the progressing condition every time a new pod was updated to the latest version

pkg/operator/apiserver/controller/workload/workload.go
sttts: does `status.UpdatedReplicas` stay down until pods are ready?

pkg/operator/apiserver/controller/workload/workload.go
sttts: add a short comment please. This is not obvious from the code.

pkg/operator/apiserver/controller/workload/workload.go
sttts: Nvmd, `UpdatedReplicas`, not status. replicas.

stlaz: It's non-terminated pods, so once a new pod appears, this number gets increased.

I was thinking of extending this, but that might make this PR more complicated. For now, I'm ok with what it does, but there's certainly room for improvement.

PR #1043, opened 2021-03-31
factory: fix warning event when no resync interval is specified

mfojtik:
For controllers that do not use the time-based resync trigger a warning event is produced about they resyncing with `0s`, eg:

`FastControllerResync Controller "ConnectivityCheckController" resync interval is set to 0s which might lead to client request throttling`

This fix handle that case also add a warning when NO resync trigger is defined for a controller (which is a programmer bug as the controller should trigger based on informer or time-based trigger).

### DIFF
--- a/pkg/controller/factory/factory.go
+++ b/pkg/controller/factory/factory.go
@@ -238,8 +238,14 @@ func (f *Factory) ToController(name string, eventRecorder events.Recorder) Contr
 
 	// Warn about too fast resyncs as they might drain the operators QPS.
 	// This event is cheap as it is only emitted on operator startup.
-	if c.resyncEvery.Seconds() < 60 {
+	if c.resyncEvery.Seconds() > 0 && c.resyncEvery.Seconds() < 60 {
 		ctx.Recorder().Warningf("FastControllerResync", "Controller %q resync interval is set to %s which might lead to client request throttling", name, c.resyncEvery)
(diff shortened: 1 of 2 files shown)


pkg/controller/factory/factory.go
```diff
@@ -238,10 +238,16 @@ func (f *Factory) ToController(name string, eventRecorder events.Recorder) Contr
...
 	// This event is cheap as it is only emitted on operator startup.
-	if c.resyncEvery.Seconds() < 60 {
+	if c.resyncEvery.Seconds() > 0 && c.resyncEvery.Seconds() < 60 {
 		ctx.Recorder().Warningf("FastControllerResync", "Controller %q resync interval is set to %s which might lead to client request throttling", name, c.resyncEvery)
 	}
 
+	// Warn when a controller is created without any resync mechanism defined
+	// This include scheduled resync or informer based resync.
+	if c.resyncEvery.Seconds() == 0 && len(f.resyncSchedules) == 0 && len(f.informers) == 0 && len(f.bareInformers) == 0 && len(f.namespaceInformers) == 0 && len(f.informerQueueKeys) == 0 {
+		ctx.Recorder().Warningf("ControllerWithNoSync", "Controller %q has no informers or resync interval specified which mean it is not running", name)
```
sttts, line 248: which means

PR #1046, opened 2021-04-06
Controller for cloud config synchronization

Danil-Grigorev:
In order to move KCM and KAPI to use out-of-tree cloud provider code they need to have an separate controller which will manage cloud config synchronization between source and destination, independent from current cluster configuration observation for cloud flags.

This configobserver is not responsible for actual change of cloud config or any of the KAPI/KCM arguments, yet is required for other flags to work properly. So it has to be connected to config observation logic, and triggered by same collection of events in order to be kept in sync.

Implementation follows proposed idea in https://github.com/openshift/library-go/pull/895#issuecomment-806021903


pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,122 @@
...
+		return existingConfig, errs
+	}
+	if err != nil {
+		return existingConfig, append(errs, err)
+	}
+
+	cloudProvider := getPlatformName(infrastructure.Status.Platform, recorder)
+
+	// we copy cloudprovider configmap values only for some cloud providers.
+	validCloudProviders := sets.NewString("aws", "azure", "gce", "openstack", "vsphere")
```
mfojtik, line 60: this should probably somewhere on top as var ? (in case we need to add to this list)

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+type cloudConfigSyncer struct {
+	targetNamespaceName string
+}
+
+// ObserveCloudProviderNames observes the cloud provider from the global cluster infrastructure resource.
+func (c *cloudConfigSyncer) SyncConfigMap(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(InfrastructureLister)
+	var errs []error
```
mfojtik, line 46: get rid of this, you are not using it :-)

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,122 @@
...
+
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if errors.IsNotFound(err) {
+		recorder.Warningf("ObserveCloudProviderNames", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		return existingConfig, errs
```
mfojtik, line 51: you never return the original error here

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if err != nil {
+		return existingConfig, append(errs, err)
```
mfojtik, line 54: errs is always empty

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	cloudProvider := getPlatformName(infrastructure.Status.Platform, recorder)
+
+	// we copy cloudprovider configmap values only for some cloud providers.
+	validCloudProviders := sets.NewString("aws", "azure", "gce", "openstack", "vsphere")
+	if !validCloudProviders.Has(cloudProvider) {
+		return existingConfig, errs
```
mfojtik, line 62: this return no error

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
Danil-Grigorev: Right, I would love to, but this method should be one of the config observers and implement [this](https://github.com/openshift/library-go/blob/2df205bdd55006a63d3fd22099db73432a0f9bd8/pkg/operator/configobserver/config_observer_controller.go#L42). I refactored the code so this will be more obvious when the error is returned.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+
+const (
+	configNamespace                 = "openshift-config"
+	machineSpecifiedConfigNamespace = "openshift-config-managed"
+	machineSpecifiedConfig          = "kube-cloud-config"
+)
+
+var validCloudProviders = sets.NewString("aws", "azure", "gce", "openstack", "vsphere")
+
+// InfrastructureLister lists infrastrucre information and allows resources to be synced
```
sttts, line 25: typo

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+type InfrastructureLister interface {
+	InfrastructureLister() configlistersv1.InfrastructureLister
+	ResourceSyncer() resourcesynccontroller.ResourceSyncer
+	ConfigMapLister() corelisterv1.ConfigMapLister
+}
+
+// NewCloudConfigSyncer returns a new cloudprovider dependent observer for syncing cloud config per provider
+// for controller-manager and api-server.
+func NewCloudConfigSyncer(targetNamespaceName string) configobserver.ObserveConfigFunc {
```
sttts, line 34: who will consume this?

This package here is about config observers. A controller is misplaced.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+		targetNamespaceName: targetNamespaceName,
+	}
+	return cloudObserver.SyncCloudConfig
+}
+
+type cloudConfigSyncer struct {
+	targetNamespaceName string
+}
+
+// ObserveCloudProviderNames observes the cloud provider from the global cluster infrastructure resource.
```
sttts, line 45: misplaced comment?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+		return existingConfig, []error{err}
+	}
+
+	sourceLocation := resourcesynccontroller.ResourceLocation{
+		Namespace: sourceCloudConfigNamespace,
+		Name:      sourceCloudConfigMap,
+		Provider:  cloudProvider,
+	}
+
+	if err := listers.ResourceSyncer().SyncConfigMap(
```
sttts, line 82: an observer never does any actions like this. It must be free of side-effect of any kind.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
+)
+
+const (
+	configNamespace                 = "openshift-config"
+	machineSpecifiedConfigNamespace = "openshift-config-managed"
+	machineSpecifiedConfig          = "kube-cloud-config"
+)
+
+var validCloudProviders = sets.NewString("aws", "azure", "gce", "openstack", "vsphere")
```
sttts, line 23: what does "valid" mean?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	return existingConfig, nil
+}
+
+func getPlatformName(platformType configv1.PlatformType, recorder events.Recorder) string {
+	cloudProvider := ""
+	switch platformType {
+	case "":
+		recorder.Warningf("ObserveCloudProvidersFailed", "Required status.platform field is not set in infrastructures.%s/cluster", configv1.GroupName)
```
sttts, line 99: side-effects like this in a getter have smell, also at the bottom.

This also creates events, doesn't it? Not good. Only create events on edges, never due to state.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ObserveCloudProviderNames observes the cloud provider from the global cluster infrastructure resource.
+func (c *cloudConfigSyncer) SyncCloudConfig(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	listers := genericListers.(InfrastructureLister)
+
+	infrastructure, err := listers.InfrastructureLister().Get("cluster")
+	if err != nil {
+		if errors.IsNotFound(err) {
+			recorder.Warningf("ObserveCloudProviderNames", "Required infrastructures.%s/cluster not found", configv1.GroupName)
```
sttts, line 52: don't we print that already just by returning the error?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		return existingConfig, []error{err}
```
sttts, line 54: Returning existingConfig is wrong. Only return values your observer owns.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+	sourceCloudConfigMap := infrastructure.Spec.CloudConfig.Name
+	sourceCloudConfigNamespace := configNamespace
+
+	// If a managed cloud-provider config is available, it should be used instead of the default. If the configmap is not
+	// found, the default values should be used.
+	if _, err = listers.ConfigMapLister().ConfigMaps(machineSpecifiedConfigNamespace).Get(machineSpecifiedConfig); err == nil {
+		sourceCloudConfigMap = machineSpecifiedConfig
+		sourceCloudConfigNamespace = machineSpecifiedConfigNamespace
+	} else if !errors.IsNotFound(err) {
+		return existingConfig, []error{err}
```
sttts, line 73: Returning existingConfig is wrong. Only return values your observer owns.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	if err := listers.ResourceSyncer().SyncConfigMap(
+		resourcesynccontroller.ResourceLocation{
+			Namespace: c.targetNamespaceName,
+			Name:      "cloud-config",
+			Provider:  cloudProvider,
+		},
+		sourceLocation); err != nil {
+		return existingConfig, []error{err}
```
sttts, line 89: Returning existingConfig is wrong. Only return values your observer owns.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+	if !validCloudProviders.Has(cloudProvider) {
+		return existingConfig, nil
+	}
+
+	sourceCloudConfigMap := infrastructure.Spec.CloudConfig.Name
+	sourceCloudConfigNamespace := configNamespace
+
+	// If a managed cloud-provider config is available, it should be used instead of the default. If the configmap is not
+	// found, the default values should be used.
+	if _, err = listers.ConfigMapLister().ConfigMaps(machineSpecifiedConfigNamespace).Get(machineSpecifiedConfig); err == nil {
```
sttts, line 69: is this configmap watched? The observer won't run if it changes otherwise.

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
elmiko: seems like we could just call this `cloudProviders`, unless there might be some distinction in the future for "valid" and "experimental"(?) providers. wdyt @Danil-Grigorev ?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+			recorder.Warningf("ObserveCloudProviderNames", "Required infrastructures.%s/cluster not found", configv1.GroupName)
+		}
+		return existingConfig, []error{err}
+	}
+
+	cloudProvider := getPlatformName(infrastructure.Status.Platform, recorder)
+
+	// we copy cloudprovider configmap values only for some cloud providers.
+	if !validCloudProviders.Has(cloudProvider) {
+		return existingConfig, nil
```
elmiko, line 61: @sttts, seems like this should also not return existingConfig?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
```diff
@@ -0,0 +1,121 @@
...
+		resourcesynccontroller.ResourceLocation{
+			Namespace: c.targetNamespaceName,
+			Name:      "cloud-config",
+			Provider:  cloudProvider,
+		},
+		sourceLocation); err != nil {
+		return existingConfig, []error{err}
+	}
+
+	return existingConfig, nil
```
elmiko, line 92: so, it seems like we shouldn't return the existingConfig, but what is the preferred action here?

it seems like this function _wants_ to return existingConfig back to the caller, could we maybe turn this into a true/false return or something so that the caller knows to just use the value it provided?

but, since it seems like this function is mainly updating the configmap, is there a better place for this to live?

elmiko (commented):
given the comments here about side-effects in the observer, i wonder if we need to re-think how this should be structured?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
elmiko: iiuc, it seems like this behavior is existent in the current implementation, https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go#L105

are you saying we should refactor to remove this behavior?

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
elmiko: this seems similar to the current pattern, https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go#L66

is this a case where we should be refactoring to improve behavior in this new implementation? (just trying to fully understand the issue)

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
sttts: you need a controller or add the configmap to the existing resource syncers

Danil-Grigorev: @stts It is already there, as the observer will be called [here](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L84-L98) and [here](https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L119-L152) the informers list contain a [list](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L37-L38) of namespaces where the configMap is [watched](https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/pkg/operator/configobservation/configobservercontroller/observe_config_controller.go#L55)

pkg/operator/configobserver/cloudconfig/sync_cloud_config.go
sttts: This is still a blocker for this PR.

sttts:
/hold

https://github.com/openshift/library-go/pull/1046/files#r608731020 is still a blocker

PR #1048, opened 2021-04-07
static-pods: always set installer reason string

sttts:
In pending and running phase the reason string was empty, leading to strange incomplete condition messages.


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -707,6 +707,13 @@ func (c *InstallerController) newNodeStateForInstallInProgress(ctx context.Conte
 		}
 		ret.LastFailedRevisionErrors = errors
 		return ret, true, fmt.Sprintf("installer pod failed: %v", strings.Join(errors, "\n")), nil
+
+	default:
+		if len(installerPod.Status.Message) > 0 {
+			reason = fmt.Sprintf("installer is not finished: %s", installerPod.Status.Message)
+		} else {
+			reason = fmt.Sprintf("installer is not finished, but in phase %s", installerPod.Status.Phase)
```
mfojtik, line 715: minor nit:

`installer is not finished, but in Pending phase`

reads better to me?

PR #1051, opened 2021-04-13
allow static resource controller to set degraded on panic

deads2k:
proof in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1101

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -76,7 +76,7 @@ func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheS
 
 func (c *baseController) Run(ctx context.Context, workers int) {
 	// HandleCrash recovers panics
-	defer utilruntime.HandleCrash()
+	defer utilruntime.HandleCrash(c.degradedPanicHandler)
 
 	// give caches 10 minutes to sync
 	cacheSyncCtx, cacheSyncCancel := context.WithTimeout(ctx, c.cacheSyncTimeout)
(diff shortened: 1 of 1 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -183,40 +183,53 @@ func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.
...
-	go func() {
-		defer utilruntime.HandleCrash()
-		defer workerWaitGroup.Done()
-		for {
-			select {
-			case <-queueCtx.Done():
-				return
-			default:
-				c.processNextWorkItem(queueCtx)
+	wait.UntilWithContext(
```
mfojtik, line 186: @deads2k if I read this right, this will restart the worker on panic, right?

pkg/controller/factory/base_controller.go
```diff
@@ -183,40 +183,53 @@ func (c *baseController) runPeriodicalResync(ctx context.Context, interval time.
...
+
+// degradedPanicHandler will go degraded on failures, then we should catch potential panics and covert them into bad status.
+func (c *baseController) degradedPanicHandler(panicVal interface{}) {
 	if c.syncDegradedClient == nil {
-		return err
+		// if we don't have a client for reporting degraded condition, then let the existing panic handler do the work
+		return
 	}
-	if err != nil {
+	_ = c.reportDegraded(context.TODO(), fmt.Errorf("panic caught:\n%v", panicVal))
```
soltysh, line 214: You can just invoke method, if you don't care about return value :wink:

PR #1052, opened 2021-04-16
add revision reason to status configmaps

deads2k:
we emit an event, but configmaps last long and this one tracks our status.


soltysh (approved):
/lgtm

PR #1053, opened 2021-04-19
add resourceApply for prometheus rule

deads2k:
To handle future us adding alerts to our operator.

/hold pending proof

### DIFF
--- a/pkg/operator/resource/resourceapply/monitoring.go
+++ b/pkg/operator/resource/resourceapply/monitoring.go
@@ -18,7 +18,7 @@ import (
 
 var serviceMonitorGVR = schema.GroupVersionResource{Group: "monitoring.coreos.com", Version: "v1", Resource: "servicemonitors"}
 
-func ensureServiceMonitorSpec(required, existing *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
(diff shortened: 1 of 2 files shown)


deads2k:
proof in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1073

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -86,3 +86,48 @@ func ApplyServiceMonitor(client dynamic.Interface, recorder events.Recorder, req
...
+	}
+
+	if klog.V(4).Enabled() {
+		klog.Infof("PrometheusRule %q changes: %v", namespace+"/"+required.GetName(), JSONPatchNoError(existing, existingCopy))
+	}
+
+	newObj, err := client.Resource(prometheusRuleGVR).Namespace(namespace).Update(context.TODO(), updated, metav1.UpdateOptions{})
+	if err != nil {
+		recorder.Warningf("PrometheusRuleUpdateFailed", "Failed to update PrometheusRule.monitoring.coreos.com/v1: %v", err)
+		return nil, true, err
```
tkashem, line 128: does the `bool` indicate `was updated or created`? I know the caller should not look at the `bool` or  the object if there is an error, but was wondering if it is semantically more correct to return `nil, false, err` here?

PR #1056, opened 2021-04-26
Bug 1947402: csi: add hook to set Deployment replicas according to numbe of master nodes

bertinatto:
This patch adds a function hook that detects the number of master nodes in the cluster (based on the `Deployment.Spec.NodeSelector` field) and sets the `Deployment.Spec.Replicas` field accordingly.

This will allow us to set the number of Deployment replicas based on the number of master nodes available. Also, we set anti-affinity rules and This is useful for mainly two things:

1. Prevent the operator going unavailable when draining nodes where the operand is running on.
2. Support the effort of avoiding Deployment rollouts getting stuck in single-node clusters (this also requires setting `maxSurge=0` in the Deployment strategy).

This PR also adds the possibility to create/update PDBs in the static resource controller. This will be required by CSI driver operators that will use the hook introduced in this PR.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -2,14 +2,12 @@ package csicontrollerset
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/cache"
 
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -38,25 +36,12 @@ type CSIControllerSet struct {
 	csiDriverNodeServiceController       factory.Controller
 	serviceMonitorController             factory.Controller
 
-	preRunCachesSynced []cache.InformerSynced
-	operatorClient     v1helpers.OperatorClient
-	eventRecorder      events.Recorder
+	operatorClient v1helpers.OperatorClient
+	eventRecorder  events.Recorder
 }
 
 // Run starts all controllers initialized in the set.
 func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
-	defer utilruntime.HandleCrash()
-
-	// Create a custom context to sync informers added  with .WithExtraInformers().
-	// This context is not used for individual controllers because factory.Factory
-	// will overwrite the timeout value.
-	cacheSyncCtx, cacheSyncCancel := context.WithTimeout(ctx, defaultCacheSyncTimeout)
-	defer cacheSyncCancel()
-
-	if !cache.WaitForCacheSync(cacheSyncCtx.Done(), c.preRunCachesSynced...) {
-		utilruntime.HandleError(fmt.Errorf("caches did not sync"))
-		return
-	}
 	for _, ctrl := range []factory.Controller{
 		c.logLevelController,
 		c.managementStateController,
@@ -151,18 +136,20 @@ func (c *CSIControllerSet) WithCSIDriverControllerService(
 	file string,
 	kubeClient kubernetes.Interface,
 	namespacedInformerFactory informers.SharedInformerFactory,
-	optionalConfigInformer configinformers.SharedInformerFactory,
+	configInformer configinformers.SharedInformerFactory,
+	optionalInformers []factory.Informer,
 	optionalDeploymentHooks ...csidrivercontrollerservicecontroller.DeploymentHookFunc,
 ) *CSIControllerSet {
 	manifestFile := assetFunc(file)
 	c.csiDriverControllerServiceController = csidrivercontrollerservicecontroller.NewCSIDriverControllerServiceController(
 		name,
 		manifestFile,
+		c.eventRecorder,
 		c.operatorClient,
 		kubeClient,
 		namespacedInformerFactory.Apps().V1().Deployments(),
-		optionalConfigInformer,
-		c.eventRecorder,
+		configInformer,
+		optionalInformers,
 		optionalDeploymentHooks...,
 	)
 	return c
@@ -174,16 +161,18 @@ func (c *CSIControllerSet) WithCSIDriverNodeService(
 	file string,
 	kubeClient kubernetes.Interface,
 	namespacedInformerFactory informers.SharedInformerFactory,
+	optionalInformers []factory.Informer,
 	optionalDaemonSetHooks ...csidrivernodeservicecontroller.DaemonSetHookFunc,
 ) *CSIControllerSet {
 	manifestFile := assetFunc(file)
 	c.csiDriverNodeServiceController = csidrivernodeservicecontroller.NewCSIDriverNodeServiceController(
 		name,
 		manifestFile,
+		c.eventRecorder,
 		c.operatorClient,
 		kubeClient,
 		namespacedInformerFactory.Apps().V1().DaemonSets(),
-		c.eventRecorder,
+		optionalInformers,
 		optionalDaemonSetHooks...,
 	)
 	return c
@@ -206,15 +195,6 @@ func (c *CSIControllerSet) WithServiceMonitorController(
 		c.operatorClient,
 		c.eventRecorder,
 	).WithIgnoreNotFoundOnCreate()
-	return c
-}
-
-// WithExtraInformers adds informers that individual controllers don't wait for. These are typically
-// informers used by hook functions in csidrivercontrollerservicecontroller and csidrivernodeservicecontroller.
-func (c *CSIControllerSet) WithExtraInformers(informers ...cache.SharedIndexInformer) *CSIControllerSet {
-	for i := range informers {
-		c.preRunCachesSynced = append(c.preRunCachesSynced, informers[i].HasSynced)
-	}
 	return c
 }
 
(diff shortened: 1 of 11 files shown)


pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -63,3 +65,20 @@ func WithSecretHashAnnotationHook(
...
+func WithStrategyHook(configinformers configinformers.SharedInformerFactory) DeploymentHookFunc {
+	return func(_ *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		infra, err := configinformers.Config().V1().Infrastructures().Lister().Get(infraConfigName)
+		if err != nil {
+			return err
+		}
+		if infra.Status.ControlPlaneTopology == v1.SingleReplicaTopologyMode {
+			deployment.Spec.Strategy = appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
+			}
```
bertinatto, line 80: We are trying to prevent the deployment rollout getting stuck in single-node clusters. Is setting `deployment.Spec.Strategy=Recreate` the best approach so address this issue?

Or should we [make it RollingUpdate like openshift-oauth-apiserver](https://github.com/openshift/cluster-authentication-operator/blob/master/bindata/oauth-apiserver/deploy.yaml#L12-L20) and set the number of replicas according to the number of master nodes?

@deads2k

deads2k: > Or should we [make it RollingUpdate like openshift-oauth-apiserver](https://github.com/openshift/cluster-authentication-operator/blob/master/bindata/oauth-apiserver/deploy.yaml#L12-L20) and set the number of replicas according to the number of master nodes?
> 
> @deads2k

I prefer the rolling update (if it semantically works for you) and number of replicas.  Now we had to limit surge because of our anti-affinity rules and our interaction with hosts.  if you don't have those contraints, you could choose to surge.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -63,3 +65,23 @@ func WithSecretHashAnnotationHook(
 		return nil
 	}
 }
+
+// WithReplicasHook sets the deployment.Spec.Replicas field according to the number
+// of available nodes. The number of nodes is determined by the node selector
+// specified in the field deployment.Spec.Templates.NodeSelector.
+func WithReplicasHook(nodeLister corev1listers.NodeLister) DeploymentHookFunc {
```
jsafrane, line 72: What's the benefit of Deployment over DaemonSet if we run one per master node?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
jsafrane: It does not work, since most CSI drivers use `hostNetwork: true` and expose ports for metrics & liveness probe. With hostNetwork, these ports become node ports. If we have one pod on each node, a new pod created by rolling upgrade can't find a node with free ports. We will need to use `Recreate` everywhere.

bertinatto: @jsafrane I think that's not true when we set `RollingUpdate` with `maxSurge=0`. Essentially this config should allow the new pod to start only when the old one is stopped.

jsafrane: Ack, tested `maxSurge: 0` and the Deployment  can proceed even with rolling update + single schedulable node.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -63,3 +65,23 @@ func WithSecretHashAnnotationHook(
 		return nil
 	}
 }
+
+// WithReplicasHook sets the deployment.Spec.Replicas field according to the number
+// of available nodes. The number of nodes is determined by the node selector
+// specified in the field deployment.Spec.Templates.NodeSelector.
```
jsafrane, line 71: Please add a note that MaxSurge should be set in the Deployment when node ports or hostNetwork are used.

bertinatto: Done

bertinatto: Done

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -38,25 +36,12 @@ type CSIControllerSet struct {
...
 func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
-	defer utilruntime.HandleCrash()
-
-	// Create a custom context to sync informers added  with .WithExtraInformers().
-	// This context is not used for individual controllers because factory.Factory
-	// will overwrite the timeout value.
-	cacheSyncCtx, cacheSyncCancel := context.WithTimeout(ctx, defaultCacheSyncTimeout)
-	defer cacheSyncCancel()
-
-	if !cache.WaitForCacheSync(cacheSyncCtx.Done(), c.preRunCachesSynced...) {
```
deads2k, line 56: interesting.  pushing it down to the controllers is ok, but do they know to do it?

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -152,8 +151,8 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
 	}
 
 	var clusterID string
-	if c.optionalConfigInformer != nil {
-		infra, err := c.optionalConfigInformer.Config().V1().Infrastructures().Lister().Get(infraConfigName)
+	if c.configInformer != nil {
```
deads2k, line 154: the name no longer says optional and you appear to require during construction.  Can this actually be nil?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
deads2k: > What's the benefit of Deployment over DaemonSet if we run one per master node?

deployments can be evicted and they are shut down on drain.  The shutdown is important enough to warrant using a deployment *for masters*.  If you're running on all workers as well, I would use a daemonset.

You may want to doc when to use one or the other.

pkg/operator/resource/resourceapply/policy.go
```diff
@@ -0,0 +1,43 @@
...
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	existingCopy := existing.DeepCopy()
+
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
```
deads2k, line 31: you need to check the content of the PDB to determine modified so that we'll update the PDB if a user mutates it

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -63,3 +65,26 @@ func WithSecretHashAnnotationHook(
...
+
+// WithReplicasHook sets the deployment.Spec.Replicas field according to the number
+// of available nodes. If the number of available nodes is bigger than one, then the
+// number of replicas will be two. The number of nodes is determined by the node
+// selector specified in the field deployment.Spec.Templates.NodeSelector.
+// When node ports or hostNetwork are used, maxSurge=0 should be set in the
+// Deployment RollingUpdate strategy to prevent the new pod from getting stuck
+// waiting for a node with free ports.
+func WithReplicasHook(nodeLister corev1listers.NodeLister) DeploymentHookFunc {
+	return func(_ *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
```
jsafrane, line 77: Having a second thought about `WithReplicasHook`. Will we have an operator that does **not** want the hook? Would it make sense to have `${MASTER_REPLICAS}` in our yaml files instead?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -63,3 +65,26 @@ func WithSecretHashAnnotationHook(
...
+		nodes, err := nodeLister.List(labels.SelectorFromSet(nodeSelector))
+		if err != nil {
+			return err
+		}
+		replicas := int32(1)
+		if len(nodes) > 1 {
+			replicas = int32(2)
+		}
+		deployment.Spec.Replicas = &replicas
+		return nil
```
jsafrane, line 88: It would be nice to have a common library-go function to get expected nr. of controller-like replicas. In this PR it's "2", but  e.g. in https://github.com/openshift/cluster-csi-snapshot-controller-operator/pull/88 it's all masters (="3").

deads2k:
/approve

I'll leave lgtm with @jsafrane

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: I did that initially, but I moved it to a hook mostly because it's easier to unit test it and because I wanted to enable this code path in a few operators first. 

I think it makes sense to promote it to a placeholder, although I don't think the hook is too bad as well. Can we do this in another PR?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: I can do that, but I don't have a good place in mind (not all controllers should be **2**).

The number of replicas should be evaluated based on the Deployment at hand. For instance, CSI Deployments should be **2**, but scheduler and KCM should be the number of master nodes. Also, IMO vsphere-problem-detector should be **1** because a small downtime isn't critical to the cluster.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: I may have found another reason a hook is better than a placeholder.

It's a chicken and egg issue. I need to replace the `${MASTER_REPLICAS}` string **before** parsing the manifest file (otherwise the parsing fails). However, I'd need to parse the manifest file to get the node selector **before** replacing the string.

A simple solution would be to drop the `${MASTER_REPLICAS}` placeholder and set the replicas regardless, but in that case a hook sounds better IMO.

PR #1057, opened 2021-04-26
Reduce encryption qps

p0lyn0mial:
Synchronizing encryption controllers is expensive because they pull data directly from the servers to get the most recent data.

By default, the controllers resync every 60 seconds. However, tighter loops can be enforced on dependencies. For example, the authentication operator reconciles its resource every 20 seconds.

This PR introduces two wrappers. One for reacting to changes only if an RV has changed and the other one for ignoring updates to nodes since the node's names cannot change.

This allows us to save QPS both on the encrypted and unencrypted clusters.

In the future, we might push it a bit further and avoid QPS entirely when the encryption is disabled.

### DIFF
--- a/pkg/operator/encryption/controllers.go
+++ b/pkg/operator/encryption/controllers.go
@@ -5,6 +5,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/openshift/library-go/pkg/operator/encryption/controllers/migrators"
 
@@ -41,6 +42,9 @@ func NewControllers(
 	// this is fine in terms of performance since these controllers will be idle most of the time
 	// TODO: update the eventHandlers used by the controllers to ignore components that do not match their own
 	encryptionSecretSelector := metav1.ListOptions{LabelSelector: secrets.EncryptionKeySecretsLabel + "=" + component}
+
+	// TODO:
+	operatorClient = newOperatorClientWrapper(operatorClient)
 
 	return &Controllers{
 		controllers: []runner{
@@ -110,4 +114,16 @@ func (c *Controllers) Run(ctx context.Context, workers int) {
 		go con.Run(ctx, workers)
 	}
 	<-ctx.Done()
+}
+
+func newOperatorClientWrapper(operatorClient operatorv1helpers.OperatorClient) operatorv1helpers.OperatorClient {
+	return &operatorClientWrapper{operatorClient}
+}
+
+type operatorClientWrapper struct {
+	operatorv1helpers.OperatorClient
+}
+
+func (ocw *operatorClientWrapper) Informer() cache.SharedIndexInformer {
+	return controllers.NewResourceVersionProtector(ocw.OperatorClient.Informer())
 }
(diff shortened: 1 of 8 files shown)


pkg/operator/encryption/controllers.go
```diff
@@ -42,6 +43,9 @@ func NewControllers(
 	// TODO: update the eventHandlers used by the controllers to ignore components that do not match their own
 	encryptionSecretSelector := metav1.ListOptions{LabelSelector: secrets.EncryptionKeySecretsLabel + "=" + component}
 
+	// TODO:
+	operatorClient = newOperatorClientWrapper(operatorClient)
```
p0lyn0mial, line 47: this might be `authentication-operator` specific and could be moved elsewhere

pkg/operator/encryption/controllers/condition_controller.go
```diff
@@ -58,6 +59,8 @@ func NewConditionController(
 }
 
 func (c *conditionController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+	klog.Info("conditionController sync called")
```
p0lyn0mial, line 62: going to remove those logs

pkg/operator/encryption/deployer/deploynodeprovider.go
```diff
@@ -54,10 +56,27 @@ func (p DeploymentNodeProvider) MasterNodeNames() ([]string, error) {
...
 		p.targetNamespaceDeploymentInformer.Informer().HasSynced,
 		p.nodeInformer.Informer().HasSynced,
 	}
 }
+
+func wrappedNodeEventHandler(delegate cache.ResourceEventHandler) cache.ResourceEventHandler {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) { delegate.OnAdd(obj) },
+		UpdateFunc: func(old, new interface{}) {
+			if _, isNode := new.(*corev1.Node); isNode {
```
p0lyn0mial, line 71: updates to `Status` (`Heartbeats`) trigger reconciliation

pkg/operator/encryption/controllers.go
```diff
@@ -111,3 +115,15 @@ func (c *Controllers) Run(ctx context.Context, workers int) {
 	}
 	<-ctx.Done()
 }
+
+func newOperatorClientWrapper(operatorClient operatorv1helpers.OperatorClient) operatorv1helpers.OperatorClient {
```
sttts, line 119: give it some good name. Why a wrapper?

pkg/operator/encryption/controllers/reactors.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+func NewResourceVersionProtector(delegate cache.SharedIndexInformer) cache.SharedIndexInformer {
```
sttts, line 14: go doc, what is this?

p0lyn0mial:
/close

https://github.com/openshift/library-go/pull/1059 merged

PR #1059, opened 2021-04-27
precondition checker for reducing encryption QPS

p0lyn0mial:
Provides a precondition checker that determines if encryption controllers should synchronize.

Synchronizing encryption controllers is expensive because they pull data directly from the servers to get the most recent data. This PR provides a checker that uses the cache for gathering enough data to determine if there is some work to be done. This helps to avoid sending requests to the API servers if there is no work to do.

TODO:
 - wire the checker
 - implement areServersOnTheSameRevision
 - add unit tests


pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+// PreconditionProvider determines if encryption controllers should synchronise.
+type PreconditionProvider interface {
```
sttts, line 18: Java'esque name.

Interface with one func are odd. Use `type Precondition func() (bool, error)`. Why the extra layer of abstraction?

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+
+	encryptionSecretSelector labels.Selector
+
+	secretLister          corev1listers.SecretNamespaceLister
+	apiServerConfigLister configv1listers.APIServerLister
+}
+
+// NewPreconditionChecker determines if encryption controllers should synchronise.
+// It uses the cache for gathering data to avoid sending requests to the API servers.
+func NewPreconditionChecker(apiServerConfigInformer configv1informers.APIServerInformer, kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces, encryptionSecretSelectorString string) (*PreconditionChecker, error) {
```
sttts, line 34: NewEncryptionEnabledPrecondition

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+	//       having all servers on the same revision
+	//       having desired and actual encryption configuration aligned
+	//       having all resources migrated
+
+	// at this point we know that encryption has been previously enabled on this cluster
+
+	// since rolling out a new version is a slow process, for kas it takes about ~20 min
+	// and the encryption controllers must wait until all servers converge on the same revision before they can act
+	// we have work to do if all api servers are on the same revision
+	if !pc.areServersOnTheSameRevision() {
```
sttts, line 67: why is same revision enough? Couldn't the revision controller be lagging?

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+// it hasn't been enabled when:
+//   the current mode is set to identity
+//   AND the encryption configuration in the managed namespace doesn't exists AND we don't have encryption secrets
+func (pc *PreconditionChecker) encryptionNeverEnabled() (bool, error) {
+	apiServerConfig, err := pc.apiServerConfigLister.Get("cluster")
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return false, err // unknown error
+		}
+		return true, nil
```
sttts, line 84: do we check that the lister is synched?

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil && !errors.IsNotFound(err) {
+		return false, err // unknown error
+	}
+
+	encryptionSecrets, err := pc.secretLister.List(pc.encryptionSecretSelector)
+	if err != nil && !errors.IsNotFound(err) {
+		return false, err // unknown error
+	}
+
+	return encryptionConfiguration == nil && len(encryptionSecrets) == 0, nil
```
sttts, line 100: do we check that the lister is synced?

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (pc *PreconditionChecker) areServersOnTheSameRevision() bool {
+	// TODO
```
sttts, line 104: TODO => WIP PR ?

deads2k, line 104: Yeah, this is critical.

p0lyn0mial:
/hold

for a manual test (https://github.com/openshift/cluster-openshift-apiserver-operator/pull/451)

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// encryptionNeverEnabled checks whether encryption hasn't been enabled on a cluster
+// it hasn't been enabled when:
+//   the current mode is set to identity
+//   AND the encryption configuration in the managed namespace doesn't exists AND we don't have encryption secrets
+func (pc *preconditionChecker) encryptionNeverEnabled() (bool, error) {
+	apiServerConfig, err := pc.apiServerConfigLister.Get("cluster")
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
deads2k, line 80: nit
```go
if apierrors.notFound(err){
  return true, nil
}
if err != nil{
  return false, nil
}

File: pkg/operator/encryption/preconditions.go [hunk 8]
```diff
@@ -0,0 +1,109 @@
...
+			return false, err // unknown error
+		}
+		return true, nil
```
deads2k, line 83: Why special case this?  it seems like if you have no apiserverconfig, you have no work to do because you won't be able to know whether encryption is on.

pkg/operator/encryption/preconditions.go
```diff
@@ -0,0 +1,109 @@
...
+	//       having all resources migrated
+
+	return true, nil // we might have work to do
+}
+
+// encryptionNeverEnabled checks whether encryption hasn't been enabled on a cluster
+// it hasn't been enabled when:
+//   the current mode is set to identity
+//   AND the encryption configuration in the managed namespace doesn't exists AND we don't have encryption secrets
+func (pc *preconditionChecker) encryptionNeverEnabled() (bool, error) {
```
deads2k, line 77: I'm having difficulty parsing the negative logic here.  Can you change it to, `EncryptionHasBeenEnabled`?

deads2k:
This approach will help.  I'm having trouble reading the inverted logic.  It might just be me getting old and tired, but a positively named method would help me I think.

pkg/operator/encryption/preconditions.go
p0lyn0mial: this is exactly what I'm doing here, if there is no `apiserverconfig` I'm returning `true` to indicate encryption wasn never `on`

pkg/operator/encryption/preconditions.go
p0lyn0mial: I could but there is a difference between `never enabled` and `currently disabled` - see `TODO` a few lines earlier.

PR #1060, opened 2021-04-27
WIP: Bug 1950379: resourcesync: allow syncing only specific CM/secret keys

stlaz:
This allows specifying concrete fields of a CM/secret to sync via resourcesyncer.

Proof PR: https://github.com/openshift/cluster-authentication-operator/pull/438

/assign @sttts 

WIP coz:
- [ ] needs unit tests


stlaz:
/retest
the tests appear to be stuck

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -109,14 +113,21 @@ func (c *ResourceSyncController) SyncConfigMap(destination, source ResourceLocat
...
 	// make sure the new rule is picked up
 	c.syncCtx.Queue().Add(c.syncCtx.QueueKey())
 	return nil
 }
 
 func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation) error {
+	return c.SyncSecretWithKeys(destination, source)
+}
+
+func (c *ResourceSyncController) SyncSecretWithKeys(destination, source ResourceLocation, keys ...string) error {
```
p0lyn0mial, line 130: so this is the method you will be using in the operator?

pkg/operator/resource/resourceapply/core.go
```diff
@@ -354,22 +368,31 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 	}
 }
 
+func deleteConfigMapSyncTarget(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, targetNamespace, targetName string) (bool, error) {
+	if _, err := client.ConfigMaps(targetNamespace).Get(context.TODO(), targetName, metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
stlaz: probably not :+1:

pkg/operator/resource/resourceapply/core.go
```diff
@@ -328,24 +329,37 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
...
 	default:
+		if syncedKeys != nil && syncedKeys.Len() > 0 {
+			for sourceKey := range source.Data {
+				if !syncedKeys.Has(sourceKey) {
+					delete(source.Data, sourceKey)
+				}
+			}
+			for sourceKey := range source.BinaryData {
+				if !syncedKeys.Has(sourceKey) {
+					delete(source.Data, sourceKey)
```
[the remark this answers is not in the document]
stlaz: of course I copy pasted this in both the places :facepalm:

pkg/operator/resource/resourceapply/core.go
```diff
@@ -354,22 +368,31 @@ func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorde
 	}
 }
 
+func deleteConfigMapSyncTarget(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, targetNamespace, targetName string) (bool, error) {
```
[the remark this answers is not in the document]
stlaz: `ConfigMapSyncTarget` keeps the relation to the `SyncConfigMap` function. I was wondering whether to keep it in the scope of the function but I am not a fan of assigning functions inside functions

pkg/operator/resource/resourceapply/core.go
```diff
@@ -391,10 +414,44 @@ func SyncSecret(client coreclientv1.SecretsGetter, recorder events.Recorder, sou
 			source.Type = corev1.SecretTypeOpaque
 		}
 
+		if syncedKeys != nil && syncedKeys.Len() > 0 {
```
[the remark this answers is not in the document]
stlaz: syncedKeys is a `sets.String`, I don't think `len(syncedKeys)` would work and I can't use `.Len()` on a nil object

pkg/operator/resourcesynccontroller/resourcesync_controller.go
stlaz: yes

pkg/operator/resource/resourceapply/core.go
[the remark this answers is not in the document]
stlaz: actually, I am not sure `len(syncedKeys)` wouldn't work :thinking:

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -126,7 +137,10 @@ func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation
 
 	c.syncRuleLock.Lock()
 	defer c.syncRuleLock.Unlock()
-	c.secretSyncRules[destination] = source
+	c.secretSyncRules[destination] = ResourceSourceLocation{
```
p0lyn0mial, line 140: then add a new map for the partial resources, ?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -109,14 +113,21 @@ func (c *ResourceSyncController) SyncConfigMap(destination, source ResourceLocat
...
+		SyncedKeys:       sets.NewString(keys...),
+	}
 
 	// make sure the new rule is picked up
 	c.syncCtx.Queue().Add(c.syncCtx.QueueKey())
 	return nil
 }
 
 func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation) error {
+	return c.SyncSecretWithKeys(destination, source)
```
p0lyn0mial, line 127: then store non-partial resources in the old map, ?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
stlaz: I don't think that's necessary

pkg/operator/resourcesynccontroller/resourcesync_controller.go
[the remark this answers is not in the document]
stlaz: I suppose the ease of use is more or less equal, I can go with that

pkg/operator/resource/resourceapply/core.go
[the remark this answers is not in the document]
p0lyn0mial: `sets.String` is a map, no?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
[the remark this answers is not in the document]
stlaz: actually, after trying to put that in code, I believe the original interface is better from the user's perspective

pkg/operator/resource/resourceapply/core.go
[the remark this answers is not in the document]
p0lyn0mial: I would inline just to align to a convention in that file - only exported functions, up to you.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -328,24 +329,37 @@ func ApplySecret(client coreclientv1.SecretsGetter, recorder events.Recorder, re
 }
 
 func SyncConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string, ownerRefs []metav1.OwnerReference) (*corev1.ConfigMap, bool, error) {
+	return SyncPartialConfigMap(client, recorder, sourceNamespace, sourceName, targetNamespace, targetName, nil, ownerRefs)
+}
+
+func SyncPartialConfigMap(client coreclientv1.ConfigMapsGetter, recorder events.Recorder, sourceNamespace, sourceName, targetNamespace, targetName string, syncedKeys sets.String, ownerRefs []metav1.OwnerReference) (*corev1.ConfigMap, bool, error) {
```
p0lyn0mial, line 335: could you rename `syncedKeys` to `sourceKeys` and move it right after `sourceName` param?

p0lyn0mial, line 335: add a `gdoc` please

stlaz: I think the `syncedKeys` name is correct as those keys are going to appear in both the source and the target.

The fact that we only configure that in source of the destination->source map is an implementation detail that I would not like to surface to the function's interface. I can move the argument either to the front or right behind the `source, destination` argument pair, but I would not put in in the middle.

stlaz: added godoc to all four of the sync functions :+1:

pkg/operator/resource/resourceapply/core.go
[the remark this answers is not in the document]
stlaz: it's a type alias for map, yes. I'll change this to just use `len()`. I wonder why they implement `.Len()` if one can use `len()` in a simpler manner

pkg/operator/resource/resourceapply/core.go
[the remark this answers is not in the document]
stlaz: I want the same handling in both the cases where this is used and so I would not inline it. I can move it inside the other sync functions and assign it to a variable to be used where it is used today, if you want.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -171,13 +185,13 @@ func (c *ResourceSyncController) Sync(ctx context.Context, syncCtx factory.SyncC
 			continue
 		}
 
-		_, _, err := resourceapply.SyncConfigMap(c.configMapGetter, syncCtx.Recorder(), source.Namespace, source.Name, destination.Namespace, destination.Name, []metav1.OwnerReference{})
+		_, _, err := resourceapply.SyncPartialConfigMap(c.configMapGetter, syncCtx.Recorder(), source.Namespace, source.Name, destination.Namespace, destination.Name, source.SyncedKeys, []metav1.OwnerReference{})
```
p0lyn0mial, line 188: seeing partial synchronization inside `Sync` method immediately yields a warning sign. Do you think it would make sense to split the code so that we could have two code paths? One for `SyncPartialConfigMap` and the other one for `SyncConfigMap` ?

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -21,3 +31,11 @@ type ResourceSyncer interface {
 	// mirror a deletion from the source.  If the source is a zero object the destination will be deleted.
 	SyncSecret(destination, source ResourceLocation) error
 }
+
+func (rsl *PartialResourceLocation) Equal(other *PartialResourceLocation) bool {
```
p0lyn0mial, line 35: where do we use this method?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
stlaz: > seeing partial synchronization inside Sync method immediately yields a warning sign
I don't understand, partial is just a special case of full

Why split into two separate code paths when the code works exactly the same? We already have a separate code path for secrets and configmaps, even though the actions are more or less equal, why add two other copy-pasted code paths?

I could of course do 
```golang
if len(source.SyncedKeys) > 0 {
    _, _, err := resourceapply.SyncPartialConfigMap(c.configMapGetter, syncCtx.Recorder(), source.Namespace, source.Name, destination.Namespace, destination.Name, source.SyncedKeys, []metav1.OwnerReference{})
} else {
    resourceapply.SyncConfigMap(c.configMapGetter, syncCtx.Recorder(), source.Namespace, source.Name, destination.Namespace, destination.Name, []metav1.OwnerReference{})
}
```
but `resourceapply.SyncConfigMap()` resolves directly to `esourceapply.SyncPartialConfigMap()` with `nil` `syncedKeys`

pkg/operator/resourcesynccontroller/interfaces.go
stlaz: I think it's unused now, let me remove it

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -109,14 +113,21 @@ func (c *ResourceSyncController) SyncConfigMap(destination, source ResourceLocat
 
 	c.syncRuleLock.Lock()
 	defer c.syncRuleLock.Unlock()
-	c.configMapSyncRules[destination] = source
+	c.configMapSyncRules[destination] = PartialResourceLocation{
+		ResourceLocation: source,
+		SyncedKeys:       sets.NewString(keys...),
```
sttts, line 118: `sets.NewString()` gives nil ?

sttts: empty and nil is the same everywhere, isn't it? Then this is fine.

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -10,7 +12,15 @@ type ResourceLocation struct {
 	Provider string `json:"provider,omitempty"`
 }
 
-var emptyResourceLocation = ResourceLocation{}
+type PartialResourceLocation struct {
+	ResourceLocation
+
+	SyncedKeys sets.String `json:"syncedKeys,omitempty"`
```
sttts, line 18: this is odd. Why does a location have the sync keys? This becomes especially strange when looking at the sync rules.

stlaz: I needed a way to store this information within the sync rules (unless you'd like to make a separate map for just the keys).

PR #1061, opened 2021-04-28
WIP: apiserver-encryption: use listers in deployers instead of client

sttts:
We did consistent client reads in the apiserver encryption deployers while eventual consistency with a cache (lister) is good enough.

### DIFF
--- a/pkg/operator/encryption/deployer/revisionedpod.go
+++ b/pkg/operator/encryption/deployer/revisionedpod.go
@@ -1,23 +1,32 @@
 package deployer
 
 import (
-	"context"
 	"fmt"
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
-	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
 	"github.com/openshift/library-go/pkg/operator/encryption/statemachine"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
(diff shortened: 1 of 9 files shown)


pkg/operator/encryption/deployer/revisionedpod.go
```diff
@@ -89,6 +80,10 @@ func NewRevisionLabelPodDeployer(
 // DeployedEncryptionConfigSecret returns the deployed encryption config and whether all
 // instances of the operand have acknowledged it.
 func (d *RevisionLabelPodDeployer) DeployedEncryptionConfigSecret() (secret *corev1.Secret, converged bool, err error) {
+	if !d.HasSynced() {
```
p0lyn0mial, line 83: why do we need an informer here?
why we cannot have two listers instead?

usually, controllers are responsible for checking the caches have been sync before acting

pkg/operator/encryption/deployer/revisionedpod.go
```diff
@@ -97,21 +92,23 @@ func (d *RevisionLabelPodDeployer) DeployedEncryptionConfigSecret() (secret *cor
 		return nil, false, nil
 	}
 
-	// do a live list so we never get confused about what revision we are on
-	apiServerPods, err := d.podClient.List(context.TODO(), metav1.ListOptions{LabelSelector: "apiserver=true"})
+	// do a lister list. This can lag, but a workload controller like deployment can lag too.
+	apiserverPodSelector, _ := labels.Parse("apiserver=true")
```
p0lyn0mial, line 96: check the error

pkg/operator/v1helpers/informers.go
```diff
@@ -48,6 +49,17 @@ func (i kubeInformersForNamespaces) Start(stopCh <-chan struct{}) {
 	}
 }
 
+func (i kubeInformersForNamespaces) WaitForCacheSync(stopCh <-chan struct{}) bool {
```
p0lyn0mial, line 52: where is that method used?

pkg/operator/encryption/deployer/revisionedpod.go
sttts: done

pkg/operator/encryption/deployer/revisionedpod.go
sttts: I don't like to fall over an non-synced informer here, reporting `converged=true` by accident. Better be defensive.

pkg/operator/v1helpers/informers.go
sttts: in tests

PR #1062, opened 2021-04-28
controller: add unit test to prove debug and metrics are served

mfojtik:


pkg/controller/controllercmd/builder.go
```diff
@@ -251,12 +278,13 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
 			return err
 		}
-		serverConfig.Authorization.Authorizer = union.New(
-			// prefix the authorizer with the permissions for metrics scraping which are well known.
-			// openshift RBAC policy will always allow this user to read metrics.
-			hardcodedauthorizer.NewHardCodedMetricsAuthorizer(),
-			serverConfig.Authorization.Authorizer,
-		)
+
+		if b.authorizer != nil {
```
deads2k, line 282: so if it is nil, we get no authorizer?  that seems bad.

pkg/controller/controllercmd/builder.go
```diff
@@ -190,35 +201,51 @@ func (b *ControllerBuilder) WithEventRecorderOptions(options record.CorrelatorOp
 
 // Run starts your controller for you.  It uses leader election if you asked, otherwise it directly calls you
 func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstructured) error {
-	clientConfig, err := b.getClientConfig()
-	if err != nil {
-		return err
+	var clientConfig *rest.Config
+
+	if b.kubeAPIServerConfigFile != nil {
```
deads2k, line 206: this change appears to have broken the automatic in-cluster configuration getting of a client.

deads2k, line 206: this seems like something to include in `b.getClientConfig`

deads2k: > this seems like something to include in `b.getClientConfig`

and in fact it is.

pkg/controller/controllercmd/builder.go
```diff
@@ -190,35 +201,51 @@ func (b *ControllerBuilder) WithEventRecorderOptions(options record.CorrelatorOp
...
 	}
-	controllerRef, err := events.GetControllerReferenceForCurrentPod(kubeClient, namespace, nil)
-	if err != nil {
-		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
+
+	var (
+		kubeClient    kubernetes.Interface
+		controllerRef *corev1.ObjectReference
+	)
+	if clientConfig != nil {
```
deads2k, line 227: if the client-config is nil when you create it, just fail then.

pkg/controller/controllercmd/cmd.go
```diff
@@ -90,9 +90,6 @@ func (c *ControllerCommandConfig) NewCommandWithContext(ctx context.Context) *co
 
 			defer logs.FlushLogs()
 			defer serviceability.BehaviorOnPanic(os.Getenv("OPENSHIFT_ON_PANIC"), c.version)()
-			defer serviceability.Profile(os.Getenv("OPENSHIFT_PROFILE")).Stop()
-
-			serviceability.StartProfiler()
```
deads2k, line 95: why did you kill our profiler?

deads2k:
Seems like what you really want to do is create an e2e test that checks every operator pod to see if it serves debug endpoints.

pkg/controller/controllercmd/builder.go
mfojtik: this is never `nil` except unit test or when somebody generate operator without WithServer (which in that case it does not matter).

pkg/controller/controllercmd/cmd.go
mfojtik: there is already profiler in apiserver :-) so in fact we running 2 profilers today

PR #1064, opened 2021-04-29
bug 1953798: stop doing live gets for optional cert syncer files that don't exist on disk yet

deads2k:
/assign @ihcsim 

This addresses the biggest burn budget offender.


deads2k:
proved in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1120

PR #1065, opened 2021-04-29
bug 1955226: update CRD apply to understand some fields are defaulted

deads2k:
This is causing repeated updated for

I0429 17:11:16.270239       1 apiextensions.go:66] CustomResourceDefinition "apirequestcounts.apiserver.openshift.io" changes: {"spec":{"conversion":null}}


sttts:
These PRs make me sad. This is so wrong to simulate defaulting.

/lgtm

PR #1067, opened 2021-05-05
Add resourceApply for VolumeSnapshotClass

Fedosin:
This commit allows managing VolumeSnapshotClass objects using a dynamic client.

### DIFF
--- a/pkg/operator/resource/resourceapply/volumesnapshotclass.go
+++ b/pkg/operator/resource/resourceapply/volumesnapshotclass.go
@@ -1,0 +1,116 @@
+package resourceapply
+
+import (
+	"context"
+
+	"k8s.io/klog/v2"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+const (
+	VolumeSnapshotClassGroup    = "snapshot.storage.k8s.io"
+	VolumeSnapshotClassVersion  = "v1"
+	VolumeSnapshotClassResource = "volumesnapshotclasses"
+)
+
+var volumeSnapshotClassResourceGVR schema.GroupVersionResource = schema.GroupVersionResource{
+	Group:    VolumeSnapshotClassGroup,
+	Version:  VolumeSnapshotClassVersion,
+	Resource: VolumeSnapshotClassResource,
+}
+
+func ensureGenericVolumeSnapshotClass(required, existing *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
+	var existingCopy *unstructured.Unstructured
+
+	// Apply "parameters"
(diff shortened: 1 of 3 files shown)


pkg/operator/resource/resourceapply/volumesnapshotclass.go
```diff
@@ -0,0 +1,96 @@
...
+	requiredParameters, _, err := unstructured.NestedMap(required.UnstructuredContent(), "parameters")
+	if err != nil {
+		return nil, false, err
+	}
+	existingParameters, _, err := unstructured.NestedMap(existing.UnstructuredContent(), "parameters")
+	if err != nil {
+		return nil, false, err
+	}
+
+	if err := mergo.Merge(&existingParameters, &requiredParameters); err != nil {
```
deads2k, line 41: I suspect you want to just overwrite on any change.  I don't think you want half of your parameters and half a user-corrupted parameter.

pkg/operator/resource/resourceapply/volumesnapshotclass.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, false, err
+	}
+
+	if equality.Semantic.DeepEqual(existingParameters, requiredParameters) {
+		return existing, false, nil
+	}
+
+	existingCopy := existing.DeepCopy()
+	if err := unstructured.SetNestedMap(existingCopy.UnstructuredContent(), existingParameters, "parameters"); err != nil {
```
deads2k, line 50: This was the typo we found in the spec version of this.  You want the requiredParameters here, right?

pkg/operator/resource/resourceapply/volumesnapshotclass.go
jsafrane: I agree, overwrite everything.

pkg/operator/resource/resourceapply/volumesnapshotclass.go
```diff
@@ -0,0 +1,91 @@
...
+			return nil, true, createErr
+		}
+		recorder.Eventf("VolumeSnapshotClassCreated", "Created VolumeSnapshotClass.snapshot.storage.k8s.io/v1 because it was missing")
+		return newObj, true, nil
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	existingCopy := existing.DeepCopy()
```
jsafrane, line 68: Why copy the object here? It gets copied again in `ensureGenericVolumeSnapshotClass`

pkg/operator/resource/resourceapply/volumesnapshotclass.go
```diff
@@ -0,0 +1,91 @@
...
+
+	toUpdate, modified, err := ensureGenericVolumeSnapshotClass(required, existingCopy)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if !modified {
+		return nil, false, nil
```
jsafrane, line 76: it should return `existing` here

pkg/operator/resource/resourceapply/volumesnapshotclass.go
```diff
@@ -0,0 +1,91 @@
...
+	requiredParameters, _, err := unstructured.NestedMap(required.UnstructuredContent(), "parameters")
+	if err != nil {
+		return nil, false, err
+	}
+	existingParameters, _, err := unstructured.NestedMap(existing.UnstructuredContent(), "parameters")
+	if err != nil {
+		return nil, false, err
+	}
+
+	if equality.Semantic.DeepEqual(existingParameters, requiredParameters) {
```
jsafrane, line 40: To be 100% sure you should check all fields, like `DeletionPolicy` and `Driver`

PR #1068, opened 2021-05-06
resourceapply: print apply diff by default

mfojtik:


sttts:
Does this only apply to changes or also to creation? Creation should stay at level 4.

PR #1070, opened 2021-05-06
bug 1957645: fix generic comparison for dynamic resource apply

deads2k:
1. fix the merge function to set the update to required, not existing
2. eliminate the merging of spec
3. provide a way to mimic defaulting to handle the case of defaults set similar to how we had to handle CRD
4. provide a way to ignore differences in spec based on custom equality.  We did this for some required fields in RBAC previously
5. fix test to catch this problem.


deads2k:
proof worked https://github.com/openshift/cluster-kube-apiserver-operator/pull/1127

PR #1073, opened 2021-05-11
Bug 1940286: csi: take into account previous cycle when setting progressing

bertinatto:
We should only set `Progressing=True` when we are rolling out new config. Relying on the number of available/unavailable pods is not enough.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
@@ -198,7 +198,8 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
 		Status: opv1.ConditionFalse,
 	}
 
-	if ok, msg := isProgressing(opStatus, deployment); ok {
+	ok, msg := v1helpers.IsDeploymentProgressing(c.name+opv1.OperatorStatusTypeProgressing, deployment, opStatus)
+	if ok {
 		progressingCondition.Status = opv1.ConditionTrue
 		progressingCondition.Message = msg
 		progressingCondition.Reason = "Deploying"
@@ -212,31 +213,12 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
 
 	_, _, err = v1helpers.UpdateStatus(
 		c.operatorClient,
-		updateStatusFn,
 		v1helpers.UpdateConditionFn(availableCondition),
 		v1helpers.UpdateConditionFn(progressingCondition),
+		updateStatusFn,
 	)
 
 	return err
-}
-
-func isProgressing(status *opv1.OperatorStatus, deployment *appsv1.Deployment) (bool, string) {
-	var deploymentExpectedReplicas int32
-	if deployment.Spec.Replicas != nil {
-		deploymentExpectedReplicas = *deployment.Spec.Replicas
-	}
-
-	switch {
-	case deployment.Generation != deployment.Status.ObservedGeneration:
-		return true, "Waiting for Deployment to act on changes"
-	case deployment.Status.UnavailableReplicas > 0:
-		return true, "Waiting for Deployment to deploy pods"
-	case deployment.Status.UpdatedReplicas < deploymentExpectedReplicas:
-		return true, "Waiting for Deployment to update pods"
-	case deployment.Status.AvailableReplicas < deploymentExpectedReplicas:
-		return true, "Waiting for Deployment to deploy pods"
-	}
-	return false, ""
 }
 
 func replacePlaceholders(manifest []byte, spec *opv1.OperatorSpec, clusterID string) []byte {
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
@@ -163,7 +163,8 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
 		Status: opv1.ConditionFalse,
 	}
 
-	if ok, msg := isProgressing(opStatus, daemonSet); ok {
+	ok, msg := v1helpers.IsDaemonSetProgressing(c.name+opv1.OperatorStatusTypeProgressing, daemonSet, opStatus)
+	if ok {
 		progressingCondition.Status = opv1.ConditionTrue
 		progressingCondition.Message = msg
 		progressingCondition.Reason = "Deploying"
@@ -177,22 +178,12 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
 
 	_, _, err = v1helpers.UpdateStatus(
 		c.operatorClient,
-		updateStatusFn,
 		v1helpers.UpdateConditionFn(availableCondition),
 		v1helpers.UpdateConditionFn(progressingCondition),
+		updateStatusFn,
 	)
 
 	return err
-}
-
-func isProgressing(status *opv1.OperatorStatus, daemonSet *appsv1.DaemonSet) (bool, string) {
-	switch {
-	case daemonSet.Generation != daemonSet.Status.ObservedGeneration:
-		return true, "Waiting for DaemonSet to act on changes"
-	case daemonSet.Status.NumberUnavailable > 0:
-		return true, "Waiting for DaemonSet to deploy node pods"
-	}
-	return false, ""
 }
 
 func replacePlaceholders(manifest []byte, spec *opv1.OperatorSpec) []byte {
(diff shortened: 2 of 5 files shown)


jsafrane (commented):
I like the idea! I would like it even more if I saw unit tests that show how the conditions work now (i.e. that Progressing is kept `False` on some random Deployment / DaemonSet hiccup that does not change the generation).

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -177,21 +177,40 @@ func (c *CSIDriverNodeServiceController) sync(ctx context.Context, syncContext f
...
+
+		// At this point we know two things: we already synced the current generation of the DaemonSet,
+		// but the DaemonSet controller is still rolling out pods. However, we don't know whether those
+		// pods have new config or not, in that case we can't determine whether we are progressing or not.
+		// As a result, we fall-back to the previous condition value and hope that in the next sync
+		// the DaemonSet controller will be already done rolling out daemon pods.
+		lastCondition := v1helpers.FindOperatorCondition(status.Conditions, opv1.OperatorStatusTypeProgressing)
+		if lastCondition == nil || lastCondition != nil && lastCondition.Status == opv1.ConditionTrue {
+			return true, fmt.Sprintf("Waiting for %d node pods", daemonSet.Status.NumberUnavailable)
+		}
```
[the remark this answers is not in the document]
bertinatto: That's the idea! I've revamped the unit tests and added a few more.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -212,30 +212,44 @@ func (c *CSIDriverControllerServiceController) sync(ctx context.Context, syncCon
...
-		return true, "Waiting for Deployment to update pods"
-	case deployment.Status.AvailableReplicas < deploymentExpectedReplicas:
-		return true, "Waiting for Deployment to deploy pods"
+	if deployment.Status.AvailableReplicas < deploymentExpectedReplicas {
+		// Verify if the current deployment generation is the same as the one that this controller
+		// dealt with in the previous sync. If they are not the same, we're definitely rolling
+		// out new config, so we are progressing.
+		lastGeneration := resourcemerge.ExpectedDeploymentGeneration(deployment, status.Generations)
+		if deployment.Generation != lastGeneration {
+			return true, "Waiting for Deployment to deploy node pods"
```
bertinatto, line 239: How can I make these messages more distinct?

jsafrane: IMO the message is OK.

jsafrane:
excellent unit tests!
/lgtm

pkg/operator/v1helpers/helpers.go
```diff
@@ -347,3 +349,65 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+	if deployment.Generation != deployment.Status.ObservedGeneration {
+		return true, "Waiting for Deployment to act on changes"
+	}
+
+	var expectedReplicas int32
+	if deployment.Spec.Replicas != nil {
+		expectedReplicas = *deployment.Spec.Replicas
+	}
+
+	if deployment.Status.AvailableReplicas < expectedReplicas {
```
wking, line 363: A deployment about to roll out changes may be completely available, but still not completely updated.  If we are looking at pod counts instead of the Deployment's `conditions`, I'd expect `updatedReplicas` to come into this.

bertinatto: Thanks! Apparently, `AvailableReplicas` is the total number of replicas in all Deployment's ReplicaSets, *regardless* the current `Deployment.Spec`. So you're right, this check might evaluate to `false` even though new config is rolling out.

On the other hand, `UpdatedReplicas` has the number of replicas of the *newer* ReplicaSet only, the one that represents the current `Deployment.Spec`. That being said, I think we only need to check `UpdatedReplicas` here.

pkg/operator/v1helpers/helpers.go
```diff
@@ -347,3 +349,65 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+		lastGeneration := resourcemerge.ExpectedDeploymentGeneration(deployment, status.Generations)
+		if deployment.Generation != lastGeneration {
+			return true, "Waiting for Deployment to deploy node pods"
+		}
+
+		// At this point we know two things: we already synced the current generation of the deployment,
+		// but the Deployment Controller is still rolling out pods. However, we don't know whether those
+		// pods have new config or not, in that case we can't determine whether we are progressing or not.
+		// As a result, we fall-back to the previous condition value and hope that in the next sync
+		// the Deployment Controller will be already done rolling out pods.
```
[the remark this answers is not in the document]
bertinatto: If the the number of pods is *also* a reliable way of determining the condition, I'd prefer to stick with that to be consistent with the DaemonSet (which doesn't contain conditions as of today).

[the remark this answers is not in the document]
wking: I don't think number of pods gives you access to  `ProgressDeadlineExceeded`.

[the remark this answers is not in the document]
bertinatto: @wking I think I don't need to check for `ProgressDeadlineExceeded`, the code above should handle situations where a Deployment contains the`ProgressDeadlineExceeded` reason, even though it doesn't explicitly checks for that.

pkg/operator/v1helpers/helpers.go
```diff
@@ -347,3 +349,65 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+		// Verify if the current deployment generation is the same as the one that this controller
+		// dealt with in the previous sync. If they are not the same, we're definitely rolling
+		// out new config, so we are progressing.
+		lastGeneration := resourcemerge.ExpectedDeploymentGeneration(deployment, status.Generations)
+		if deployment.Generation != lastGeneration {
+			return true, "Waiting for Deployment to deploy node pods"
+		}
+
+		// At this point we know two things: we already synced the current generation of the deployment,
+		// but the Deployment Controller is still rolling out pods. However, we don't know whether those
```
deads2k, line 373: if the deployment controller is still rolling out pods, why would we want to avoid being progressing?  How often are you scaling up?

pkg/operator/v1helpers/helpers.go
```diff
@@ -347,3 +349,65 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+		// pods have new config or not, in that case we can't determine whether we are progressing or not.
+		// As a result, we fall-back to the previous condition value and hope that in the next sync
+		// the Deployment Controller will be already done rolling out pods.
+		lastCondition := FindOperatorCondition(status.Conditions, conditionName)
+		if lastCondition == nil || lastCondition != nil && lastCondition.Status == operatorv1.ConditionTrue {
+			return true, "Waiting for Deployment to deploy node pods"
+		}
+	}
+
```
deads2k, line 382: don't we also want to check availabile replicas?

pkg/operator/v1helpers/helpers.go
```diff
@@ -347,3 +349,65 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+		// Verify if the current daemonSet generation is the same as the one that this controller
+		// dealt with in the previous sync. If they are not the same, we're definitely rolling
+		// out new config, so we are progressing.
+		lastGeneration := resourcemerge.ExpectedDaemonSetGeneration(daemonSet, status.Generations)
+		if daemonSet.Generation != lastGeneration {
+			return true, fmt.Sprintf("Waiting for %d node pods", daemonSet.Status.NumberUnavailable)
+		}
+
+		// At this point we know two things: we already synced the current generation of the DaemonSet,
+		// but the DaemonSet controller is still rolling out pods. However, we don't know whether those
```
deads2k, line 402: same here.  It's not clear to me why the operand is not progressing if the operand has not achieved a steady state.

pkg/operator/v1helpers/helpers.go
wking: [The bug points out][1] that [the API docs for `Progressing` say][2]:

> Progressing indicates that the operator is actively rolling out new code, propagating config changes, or otherwise moving from one steady state to another.  Operators should not report progressing when they are reconciling a previously known state.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1940286#c0
[2]: https://github.com/openshift/api/blob/8356aa4d4afb94790d3ad58c4debe0e1bdabcbe9/config/v1/types_cluster_operator.go#L147-L151

bertinatto:
/close

After some discussion offline it been decided that `Progressing` should be `True` when operator reacts to changes in the cluster, which makes this PR unnecessary.

The PR https://github.com/openshift/api/pull/935 was opened to change the API documentation and [bug 1940286](https://bugzilla.redhat.com/show_bug.cgi?id=1940286) has been closed.

PR #1075, opened 2021-05-13
staticpod: recreate existing pod manifests instead of updating

mfojtik:


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -337,6 +337,9 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 	// Write secrets, config maps and pod to disk
 	// This does not need timeout, instead we should fail hard when we are not able to write.
 	podFileName := o.PodConfigMapNamePrefix + ".yaml"
+	if err := os.Remove(path.Join(resourceDir, podFileName)); err == nil {
```
p0lyn0mial, line 340: is `resourceDir` our target directory? If yes then it is revisioned and shouldn't contain any data, right?

mfojtik: @p0lyn0mial i think this is for the pod config map... in theory, we don't need to remove the existing file, does not hurt probably?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -348,6 +351,10 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 		return err
 	}
 
+	// remove the existing file to ensure kubelet gets "create" event from inotify watchers
+	if err := os.Remove(path.Join(o.PodManifestDir, podFileName)); err == nil {
```
p0lyn0mial, line 355: we could log a potential error by checking `err != nil && !os.IsNotExist(err)`

p0lyn0mial: since we know there's a potential err we could simply return the error and retry at a later time, wdyt?

mfojtik: @p0lyn0mial right, the pod will retry... and if we can't remove the existing file it is not likely we would be able to write new content :)

PR #1078, opened 2021-05-14
Bug 1960758: New helper for oc adm must-gather / oc debug

smarterclayton:
The ResolveLatestTaggedImage utility method was intended for callers that needed to wait for an image import to succeed (trigger controllers and build controllers that need the image SHA). A number of use cases exist where import is not required, such as oc debug and must-gather where the qualified tag is sufficient. Since the existing utility method has confused callers about its limitations and intended use, create a new utility method intended for the three most common use cases:

1. You need to get the user's *current* desired intent (oc adm release new) and want to error out if the import has errors or isn't
   complete
2. You need to get a pull spec but you don't care if it's out of date (oc debug / oc adm must-gather)
3. You need to get a pull spec and you must have imported the image

Add a new ResolvePullSpecForTag method that debug/must-gather will use.

### DIFF
--- a/pkg/image/imageutil/helpers.go
+++ b/pkg/image/imageutil/helpers.go
@@ -222,6 +222,16 @@ func PrioritizeTags(tags []string) {
 	}
 }
 
+// SpecHasTag returns named tag from image stream's spec and boolean whether one was found.
+func SpecHasTag(stream *imagev1.ImageStream, name string) (imagev1.TagReference, bool) {
+	for _, tag := range stream.Spec.Tags {
+		if tag.Name == name {
+			return tag, true
+		}
+	}
+	return imagev1.TagReference{}, false
+}
+
 // StatusHasTag returns named tag from image stream's status and boolean whether one was found.
 func StatusHasTag(stream *imagev1.ImageStream, name string) (imagev1.NamedTagEventList, bool) {
 	for _, tag := range stream.Status.Tags {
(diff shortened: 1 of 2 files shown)


pkg/image/imageutil/helpers.go
```diff
@@ -286,10 +296,138 @@ func ImageWithMetadataOrDie(image *imagev1.Image) {
...
+// Use cases:
+//
+// * Get the latest pull spec representing user intent (user may be wrong or pull spec may not exist):
+//   - set requireLatest = true so we don't see older tags
+//   - if pullSpec is empty the tag doesn't represent an image
+//   - if isImporting is true the server may be having an issue getting this image or it doesn't exist
+//   - isLatest will always be true
+//
+// * Get the latest valid pull spec from the user (last good value):
+//   - set requireLatest = false so we take the latest
```
wking, line 349: "so we take the latest" -> "so we allow earlier images if the latest has failed or is in progress"?

pkg/image/imageutil/helpers.go
```diff
@@ -286,10 +296,138 @@ func ImageWithMetadataOrDie(image *imagev1.Image) {
...
+// * Get the latest pull spec representing user intent (user may be wrong or pull spec may not exist):
+//   - set requireLatest = true so we only consider the most recent pull/tag/import
+//   - if pullSpec is empty the tag doesn't represent an image
+//   - if isImporting is true the server may be having an issue getting this image or it doesn't exist
+//   - isLatest will always be true
+//
+// * Get the latest valid pull spec from the user (last good value):
+//   - set requireLatest = false so we take the last known valid pull/tag/import
+//   - if pullSpec is empty the tag doesn't represent a valid image
+//   - if isLatest is false, we chose an older image in the history
```
wking, line 351: Do we want to talk about `isImporting` for this case?  I expect that `isImporting` is more likely to be false than in the `requireLatest=true` case, but maybe we can still have `isImporting` false if we have never imported any images for this tag yet?

smarterclayton: isImporting is about the tag, not the value, so it has the same chance of being false in both scenarios.

smarterclayton: If there's an argument to make isImporting about the value that's potentially interesting I guess - since isLatest=false would still communicate importing anyway.  But that doesn't help us get rid of one of the returned booleans, so it seems like a wash.

pkg/image/imageutil/helpers.go
```diff
@@ -286,10 +296,138 @@ func ImageWithMetadataOrDie(image *imagev1.Image) {
...
+//
+func ResolvePullSpecForTag(stream *imagev1.ImageStream, tag string, defaultExternal, requireLatest bool) (pullSpec string, hasStatus, isImporting, isLatest bool, err error) {
+	if len(tag) == 0 {
+		tag = imagev1.DefaultImageTag
+	}
+	status, _ := StatusHasTag(stream, tag)
+	spec, hasSpec := SpecHasTag(stream, tag)
+
+	var event *imagev1.TagEvent
+	var importing bool
```
dmage, line 365: Can you reuse `isImporting`?

pkg/image/imageutil/helpers.go
```diff
@@ -286,10 +296,138 @@ func ImageWithMetadataOrDie(image *imagev1.Image) {
...
+// If the spec tag is still in the process of importing the image or has failed to import the image,
+// the pull spec will be either the previous image in the tag history or the value of the spec tag.
+// Setting requireLatest to true will never return the previous image.
+//
+// If no spec tag is set and no history exists, the pull spec will be the internal image stream repository
+// value plus the tag value (<stream.status.dockerImageRepository>:<tag>). If defaultExternal is true
+// and stream.status.publicDockerImageRepository is not empty, that public repository spec will be
+// returned. If neither field is set ErrNoStreamRepository will be returned.
+//
+// hasStatus is true if the pull spec corresponds to an image in the status tags. hasImported is true
```
dmage, line 336: You don't have `hasImported`.

pkg/image/imageutil/helpers.go
```diff
@@ -286,10 +296,138 @@ func ImageWithMetadataOrDie(image *imagev1.Image) {
...
+// if an import is in progress for this tag. isLatest is true if the returned pull spec matches the
+// user's desired intent (will never be false when requireLatest is true).
```
dmage, line 338: Honestly, I don't understand difference between `hasStatus` and `isLatest`

pkg/image/imageutil/helpers.go
smarterclayton: Probably

pkg/image/imageutil/helpers.go
smarterclayton: I'll make sure they're well clarified.

soltysh: Same comment from my end, to me both `hasStatus` and `isLatest` are equivalent, b/c both are talking about spec matching status, which is what you're saying in explanation as well.

soltysh (approved):
I still wonder about the usefulness of these many return values for `ResolveNewestPullSpecForTag` and `ResolveRecentPullSpecForTag`, but let's try to use it as is and see how it turns out. 

In both cases I'd probably drop `isTagEmpty` and maybe `hasStatus`, but the latter I can see some possible usage. 

/lgtm

PR #1079, opened 2021-05-17
[WIP] add func to allow for dynamic audit policy creation

EmilyM1:
In order to create an audit policy that can filter on a specified group the way we generate the audit policies needs to change to take the custom rule for groups. After the policy accepts the groups this will need to get generated on the fly and server the the upstream operators.


pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	kyaml "k8s.io/apimachinery/pkg/util/yaml"
+	auditconfig "k8s.io/apiserver/pkg/apis/audit/v1"
```
EmilyM1, line 12: @sttts I just wanted to call out that this import is needed for the output of the new `GetAuditPolicy` func. This apiserver was not imported before.

PR #1081, opened 2021-05-18
CSI: Unify assetFuncs for all controllers

Fedosin:
Now there is some inconsistency between CSI controllers, some of which require the asset function to return an error and others to panic immediately. 

Due to the transition to embed module that provides only a function that returns an error[1], we need to unify our interfaces for the controllers.

This commit does this by requiring that all controllers now accept a function that returns an error. For controllers that should have panicked, we do an explicit check in the code.

[1] https://golang.org/pkg/embed/#FS.ReadFile

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -2,6 +2,7 @@ package csicontrollerset
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -101,11 +102,14 @@ func (c *CSIControllerSet) WithStaticResourcesController(
 func (c *CSIControllerSet) WithCredentialsRequestController(
 	name string,
 	operandNamespace string,
-	assetFunc func(string) []byte,
+	assetFunc resourceapply.AssetFunc,
 	file string,
 	dynamicClient dynamic.Interface,
 ) *CSIControllerSet {
-	manifestFile := assetFunc(file)
+	manifestFile, err := assetFunc(file)
+	if err != nil {
+		panic(fmt.Sprintf("asset: Asset(%v): %v", file, err))
+	}
 	c.credentialsRequestController = credentialsrequestcontroller.NewCredentialsRequestController(
 		name,
 		operandNamespace,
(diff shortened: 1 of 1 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -116,11 +116,14 @@ func (c *CSIControllerSet) WithStaticResourcesController(
...
 	operandNamespace string,
-	assetFunc func(string) []byte,
+	assetFunc resourceapply.AssetFunc,
 	file string,
 	dynamicClient dynamic.Interface,
 ) *CSIControllerSet {
-	manifestFile := assetFunc(file)
+	manifestFile, err := assetFunc(file)
+	if err != nil {
+		panic("asset: Asset(" + file + "): " + err.Error())
```
bertinatto, line 125: I think you can just use %q and %v instead of concatenating

bertinatto (commented):
lgtm

pkg/operator/csi/csicontrollerset/csi_controller_set.go
Fedosin: Agree. Initially I copied this part from go-bindata generated code: https://github.com/openshift/openstack-cinder-csi-driver-operator/blob/master/pkg/generated/bindata.go#L1296 to make it the same.

I'll fix it in a minute.

jsafrane:
/lgtm
/assign @deads2k 
for approval

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -101,11 +102,14 @@ func (c *CSIControllerSet) WithStaticResourcesController(
...
 	operandNamespace string,
-	assetFunc func(string) []byte,
+	assetFunc resourceapply.AssetFunc,
 	file string,
 	dynamicClient dynamic.Interface,
 ) *CSIControllerSet {
-	manifestFile := assetFunc(file)
+	manifestFile, err := assetFunc(file)
+	if err != nil {
+		panic(fmt.Sprintf("asset: Asset(%v): %v", file, err))
```
deads2k, line 111: hrm.  I dislike this in my libraries.  it makes testing hard.  How bad is are the ripples for changing the signature to return an error?

Fedosin: @jsafrane @bertinatto what do you think?

deads2k:
/approve

today it panics (I think) and since this is reading bindata and the like, its programming error.

PR #1083, opened 2021-05-20
Bug 1965929: add a CSR approving controller

stlaz:
/assign @deads2k 

----
wip because the unit tests need finishing

### DIFF
--- a/pkg/operator/csr/csr_approver.go
+++ b/pkg/operator/csr/csr_approver.go
@@ -1,0 +1,196 @@
+package webhookcerts
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+
+	certapiv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authentication/serviceaccount"
+	certv1informers "k8s.io/client-go/informers/certificates/v1"
+	certv1client "k8s.io/client-go/kubernetes/typed/certificates/v1"
+	certv1listers "k8s.io/client-go/listers/certificates/v1"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type CSRApprovalDecision string
+
+const (
+	CSRApproved  CSRApprovalDecision = "Approved"
+	CSRDenied    CSRApprovalDecision = "Denied"
+	CSRNoOpinion CSRApprovalDecision = "NoOpinion"
+)
+
+type CSRApprover interface {
+	Approve(csrObj *certapiv1.CertificateSigningRequest, x509CSR *x509.CertificateRequest) (approvalStatus CSRApprovalDecision, denyReason string, err error)
+}
+
+type csrApproverController struct {
+	csrClient certv1client.CertificateSigningRequestInterface
+	csrLister certv1listers.CertificateSigningRequestLister
+
+	csrName     string
+	csrApprover CSRApprover
+}
+
+// NewCSRApproverController returns a controller that is observing the CSR API
+// for a CSR of a given name. If such a CSR exists, it runs the `csrApprover.Approve()`
+// against it and either denies, approves or leaves the CSR.
+func NewCSRApproverController(
+	controllerName string,
+	operatorClient v1helpers.OperatorClient,
+	csrClient certv1client.CertificateSigningRequestInterface,
+	csrInformers certv1informers.CertificateSigningRequestInformer,
+	csrName string,
+	csrApprover CSRApprover,
+	eventsRecorder events.Recorder,
+) factory.Controller {
+	c := &csrApproverController{
+		csrClient:   csrClient,
+		csrLister:   csrInformers.Lister(),
+		csrName:     csrName,
+		csrApprover: csrApprover,
+	}
+
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		WithFilteredEventsInformers(factory.NamesFilter(c.csrName), csrInformers.Informer()).
+		ToController(
+			"WebhookAuthenticatorCertApprover_"+controllerName,
+			eventsRecorder.WithComponentSuffix("webhook-authenticator-cert-approver-"+controllerName),
+		)
+}
+
+func (c *csrApproverController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	csr, err := c.csrLister.Get(c.csrName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	if approved, denied := getCertApprovalCondition(&csr.Status); approved || denied {
+		return nil
+	}
+
+	csrCopy := csr.DeepCopy()
+
+	csrPEM, _ := pem.Decode(csr.Spec.Request)
+	if csrPEM == nil {
+		return fmt.Errorf("failed to PEM-parse the CSR block in .spec.request: no CSRs were found")
+	}
+
+	x509CSR, err := x509.ParseCertificateRequest(csrPEM.Bytes)
+	if err != nil {
+		return fmt.Errorf("failed to parse the CSR bytes: %v", err)
+	}
+
+	if x509CSR.Subject.CommonName == csr.Spec.Username {
+		c.denyCSR(ctx, csrCopy, "IllegitimateRequester", "requester cannot request certificates for themselves", syncCtx.Recorder())
+	}
+
+	csrDecision, denyReason, err := c.csrApprover.Approve(csr, x509CSR)
+	if err != nil {
+		return c.denyCSR(ctx, csrCopy, "CSRApprovingFailed", fmt.Sprintf("there was an error during CSR approval: %v", err), syncCtx.Recorder())
+	}
+
+	switch csrDecision {
+	case CSRDenied:
+		return c.denyCSR(ctx, csrCopy, "CSRDenied", denyReason, syncCtx.Recorder())
+	case CSRApproved:
+		return c.approveCSR(ctx, csrCopy, syncCtx.Recorder())
+	case CSRNoOpinion:
+		fallthrough
+	default:
+		return nil
+	}
+}
+
+func (c *csrApproverController) denyCSR(ctx context.Context, csrCopy *certapiv1.CertificateSigningRequest, reason, message string, eventsRecorder events.Recorder) error {
+	csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
+		certapiv1.CertificateSigningRequestCondition{
+			Type:    certapiv1.CertificateDenied,
+			Status:  corev1.ConditionTrue,
+			Reason:  reason,
+			Message: message,
+		},
+	)
+
+	eventsRecorder.Eventf("CSRDenial", "The CSR %q has been denied: %s - %s", csrCopy.Name, reason, message)
+	_, err := c.csrClient.UpdateApproval(ctx, csrCopy.Name, csrCopy, v1.UpdateOptions{})
+	return err
+}
+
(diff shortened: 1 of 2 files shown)


pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// NewCSRApproverController returns a controller that is observing the CSR API
+// for a CSR of a given name. If such a CSR exists, it runs the `csrApprover` against
+// it and either denies, approves or leaves the CSR.
+func NewCSRApproverController(
+	operatorClient v1helpers.OperatorClient,
+	csrClient certv1client.CertificateSigningRequestInterface,
+	csrInformers certv1informers.CertificateSigningRequestInformer,
+	csrName string,
```
deads2k, line 49: Allow a filter func with a couple "simple" filters
1. label selection
2. names
3. name prefix (generated names)
4. commonname
5. AND filter
6. OR filter

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+
+	csrName     string
+	csrApprover CSRApproveFunc
+}
+
+// NewCSRApproverController returns a controller that is observing the CSR API
+// for a CSR of a given name. If such a CSR exists, it runs the `csrApprover` against
+// it and either denies, approves or leaves the CSR.
+func NewCSRApproverController(
+	operatorClient v1helpers.OperatorClient,
```
deads2k, line 46: followup: we need a no-op operatorclient for cases like prometheus

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+	c := &csrApproverController{
+		csrClient:   csrClient,
+		csrLister:   csrInformers.Lister(),
+		csrName:     csrName,
+		csrApprover: csrApprover,
+	}
+
+	return factory.New().
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
```
deads2k, line 62: follow: need to allow a general failure hook

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+		WithFilteredEventsInformers(factory.NamesFilter(c.csrName), csrInformers.Informer()).
+		ToController("WebhookAuthenticatorCertApprover", eventsRecorder.WithComponentSuffix("webhook-authenticator-cert-approver"))
```
deads2k, line 64: need to pass the name so I can use this twice.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+	csrInformers certv1informers.CertificateSigningRequestInformer,
+	csrName string,
+	csrApprover CSRApproveFunc,
+	eventsRecorder events.Recorder,
+) factory.Controller {
+	c := &csrApproverController{
+		csrClient:   csrClient,
+		csrLister:   csrInformers.Lister(),
+		csrName:     csrName,
+		csrApprover: csrApprover,
```
deads2k, line 57: minor: `IsApproved` or `csrApproverFn` or some such please.  Not a noun name.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+	case CSRApproved:
+		csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
+			certapiv1.CertificateSigningRequestCondition{
+				Type:    certapiv1.CertificateApproved,
+				Status:  corev1.ConditionTrue,
+				Reason:  "AutoApproved",
+				Message: fmt.Sprintf("Auto-approved CSR %q", c.csrName),
+			},
+		)
+		_, err = c.csrClient.UpdateApproval(ctx, c.csrName, csrCopy, v1.UpdateOptions{})
```
deads2k, line 113: please wrap in a function and give us an event.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+	csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
+		certapiv1.CertificateSigningRequestCondition{
+			Type:    certapiv1.CertificateDenied,
+			Status:  corev1.ConditionTrue,
+			Reason:  reason,
+			Message: message,
+		},
+	)
+
+	_, err := c.csrClient.UpdateApproval(ctx, c.csrName, csrCopy, v1.UpdateOptions{})
```
deads2k, line 132: please emit an event in here.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		if c.Type == certapiv1.CertificateDenied {
+			denied = true
+		}
+	}
+	return
+}
+
+// ServiceAccountApprover approves CSRs with a given subject issued by the provided service account
+func ServiceAccountApprover(saName, expectedSubject string) (CSRApproveFunc, error) {
```
deads2k, line 149: minor: allow callers to use namespace and saName

deads2k, line 149: minor, I prefer a struct with these values and an Approve func on it instead of the closure

deads2k:
is this still a WIP?  It's matching my expectations quite well.

stlaz:
> is this still a WIP? It's matching my expectations quite well.

I had some unit tests I wanted to fix before marking it as non-WIP, but should be good to merge with comments addressed (after squashing, perhaps).

Please see if the new modifications are matching your expectations :slightly_smiling_face:

pkg/operator/csr/csr_approver.go
stlaz: I'm now using an interface instead of a function as suggested below so I kept the name

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,272 @@
...
+		csrDecision, denyReason, err := c.csrApprover.Approve(csr, x509CSR)
+		if err != nil {
+			return c.denyCSR(ctx, csrCopy, "CSRApprovingFailed", fmt.Sprintf("there was an error during CSR approval: %v", err), syncCtx.Recorder())
+		}
+
+		switch csrDecision {
+		case CSRDenied:
+			return c.denyCSR(ctx, csrCopy, "CSRDenied", denyReason, syncCtx.Recorder())
+		case CSRApproved:
+			csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
```
slaskawi, line 118: Perhaps it makes sense to extract this into a function as well (as you did with `CSRDenied` case)?

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,272 @@
...
+// ServiceAccountApprover approves CSRs with a given subject issued by the provided service account
+func NewServiceAccountApprover(saNamespace, saName, expectedSubject string) *ServiceAccountApprover {
+	return &ServiceAccountApprover{
+		saName:          serviceaccount.MakeUsername(saNamespace, saName),
+		saGroups:        sets.NewString(append(serviceaccount.MakeGroupNames(saNamespace), "system:authenticated")...),
+		expectedSubject: expectedSubject,
+	}
+}
+
+func (a *ServiceAccountApprover) Approve(csrObj *certapiv1.CertificateSigningRequest, x509CSR *x509.CertificateRequest) (approvalStatus CSRApprovalDecision, denyReason string, err error) {
```
slaskawi, line 183: Perhaps we could use a better name here, e.g. `Validate`? In essence, it's what this method is doing.

slaskawi (approved):
Great work! Only a few nitpicks found - it's up to you if you want to fix them.

pkg/operator/csr/csr_approver.go
stlaz: it gives out an approval decision so I would keep the name

pkg/operator/csr/csr_approver.go
stlaz: it's single use so I did not want to do that, but there's no hurt in doing so :+1:

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,272 @@
...
+	if csrObj == nil || x509CSR == nil {
+		return CSRDenied, "", fmt.Errorf("received a 'nil' CSR")
```
s-urbaniak, line 185: as we are denying here, we should return a deny reason here as well, no?

stlaz: I can add an "Error" deny reason here

pkg/operator/csr/csr_approver.go
slaskawi: Works for me!

slaskawi (approved):

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,275 @@
...
+	csrList, err := c.csrLister.List(labels.Everything())
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	for _, csr := range csrList {
+		if !c.csrFilter.Matches(csr) {
```
deads2k, line 85: something passed in the night.  I'll merge this to have a starting point, but I think we want this filter function on every item we observe from the informer.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,275 @@
...
+	csrCopy.Status.Conditions = append(csrCopy.Status.Conditions,
+		certapiv1.CertificateSigningRequestCondition{
+			Type:    certapiv1.CertificateDenied,
+			Status:  corev1.ConditionTrue,
+			Reason:  reason,
+			Message: message,
+		},
+	)
+
+	eventsRecorder.Eventf("CSRDenial", "The CSR %q has been denied", csrCopy.Name)
```
deads2k, line 140: don't be stingy, list the reason and message too.  CSRs are reaped.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,275 @@
...
+			denied = true
+		}
+	}
+	return
+}
+
+type ServiceAccountApprover struct {
+	saGroups        sets.String // saGroups is the set of groups for the SA expected to have created the CSR
+	saName          string
+	expectedSubject string
```
deads2k, line 174: need to include expected groups too.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,275 @@
...
+func (a *ServiceAccountApprover) Approve(csrObj *certapiv1.CertificateSigningRequest, x509CSR *x509.CertificateRequest) (approvalStatus CSRApprovalDecision, denyReason string, err error) {
+	if csrObj == nil || x509CSR == nil {
+		return CSRDenied, "Error", fmt.Errorf("received a 'nil' CSR")
+	}
+
+	if csrObj.Spec.Username != a.saName {
+		return CSRDenied, fmt.Sprintf("CSR %q was created by an unexpected user: %q", csrObj.Name, csrObj.Spec.Username), nil
+	}
+
+	if csrGroups := sets.NewString(csrObj.Spec.Groups...); !csrGroups.IsSuperset(a.saGroups) {
```
deads2k, line 195: without groups, you need to deny anything specifying a group.

pkg/operator/csr/csr_approver.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (f *RequestCommonNameFilter) Match(csr *certapiv1.CertificateSigningRequest) bool {
+	x509CSR, err := x509.ParseCertificateRequest(csr.Spec.Request)
+	if err != nil {
+		klog.V(4).Infof("failed to parse the CSR .spec.request of %q: %v", csr.Name, err)
+		return false
+	}
+
+	return f.commonNames.Has(x509CSR.Subject.CommonName)
```
deads2k, line 274: difference between subject and and common name filters?  What else does the subject include?

stlaz: There may not be too many uses inside kube for other subject fields, but there's a few other possible fields, typically Organization (o), OrganizationalUnit (ou), Location (l).

Alternatively, we can just match against the string representation of the full subject.

pkg/operator/csr/csr_approver.go
stlaz: I'm going to allow specifying additional groups and make this check and `Equal()`, then

pkg/operator/csr/csr_approver.go
stlaz: make it a key function, I see. Can do in a follow-up, then.

PR #1084, opened 2021-05-20
Alternate lookups

soltysh:
Simplified version of #939 as discussed with @smarterclayton 

/assign @smarterclayton 

The initial take in oc lives in


pkg/image/registryclient/client_mirrored.go
```diff
@@ -123,15 +123,17 @@ func (r *blobMirroredRepository) initialRepos(ctx context.Context) ([]reference.
 		return []reference.DockerImageReference{r.locator.ref}, false, nil
 	}
 
-	r.lock.Lock()
```
smarterclayton: This is concerning - why did you make this change?  This prevents multiple FirstRequests being called, so I didn't expect it to be removed.

pkg/image/registryclient/client.go
```diff
@@ -34,7 +34,7 @@ import (
 type RepositoryRetriever interface {
 	// Repository returns a properly authenticated distribution.Repository for the given registry, repository
 	// name, and insecure toleration behavior.
-	Repository(ctx context.Context, registry *url.URL, repoName string, insecure bool) (distribution.Repository, error)
+	Repository(ctx context.Context, registry *url.URL, repoName string, insecure bool) (RepositoryWithLocation, error)
```
smarterclayton, line 37: Can you link to where this is required (use case in oc)?

pkg/image/registryclient/client_mirrored.go
soltysh: yeah, I think that's coming from the fact that Sally added `errorRepos` below but analyzing that again, it doesn't make sense, since we have higher-level methods responsible for calling `errorRepos`, I'll also update the comments with that information

pkg/image/registryclient/client.go
soltysh: I went through all the possible places and I'm not seeing any particular cases, yet. I'll revert that bit back as well.

sallyom: this was requested by image-registry (@ricardomaraschini ) they use this

sallyom: https://github.com/openshift/library-go/pull/939#discussion_r613257953

ricardomaraschini: Image registry uses the `RepositoryRetriever` interface.

pkg/image/registryclient/client_mirrored.go
sallyom: right, calling errorRepos from initialRepos there was a deadlock-there was limited context with use of FirstRequest, it's simply returned/not used w/ oc & image-registry, so i left the lock in errorRepos since that's where oc required it- wasn't foreseeing multiple FirstRequests - not calling errorRepos & returning the method's ref instead has the same result and is more appropriate.

smarterclayton:
Can you squash the last commit into the previous one?

pkg/image/registryclient/client.go
soltysh: Right, but from what I'm looking around this means keeping the interface untouched which is what I did in my last commit. We can always re-submit that change, if necessary in a followup.

soltysh:
> Can you squash the last commit into the previous one?

Done

PR #1095, opened 2021-06-03
Introduce guard controller for static pods

ravisantoshgudimetla:
Have independent deployment and PDB controllers stitched together to form guard controller

### DIFF
--- a/pkg/operator/resource/resourceapply/apiextensions.go
+++ b/pkg/operator/resource/resourceapply/apiextensions.go
@@ -40,3 +40,15 @@ func ApplyCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1.
 
 	return actual, true, err
 }
+
+func DeleteCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, recorder events.Recorder, required *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, bool, error) {
+	err := client.CustomResourceDefinitions().Delete(ctx, required.Name, metav1.DeleteOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
+		return nil, false, nil
+	}
+	if err != nil {
+		return nil, false, err
+	}
+	reportDeleteEvent(recorder, required, err)
+	return nil, true, nil
+}
--- a/pkg/operator/resource/resourceapply/generic.go
+++ b/pkg/operator/resource/resourceapply/generic.go
@@ -52,6 +52,10 @@ type ApplyResult struct {
 	Error   error
 }
 
+// ConditionalFunction provides needed dependency for a resource on another condition instead of blindly creating
+// a resource. This conditional function can also be used to delete the resource when not needed
+type ConditionalFunction func() bool
+
 type ClientHolder struct {
 	kubeClient          kubernetes.Interface
 	apiExtensionsClient apiextensionsclient.Interface
@@ -212,6 +216,135 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
 				result.Error = fmt.Errorf("missing dynamicClient")
 			} else {
 				result.Result, result.Changed, result.Error = ApplyKnownUnstructured(ctx, clients.dynamicClient, recorder, t)
+			}
+		default:
+			result.Error = fmt.Errorf("unhandled type %T", requiredObj)
+		}
+
+		ret = append(ret, result)
+	}
+
+	return ret
+}
+
+func DeleteAll(ctx context.Context, clients *ClientHolder, recorder events.Recorder, manifests AssetFunc,
+	files ...string) []ApplyResult {
+	ret := []ApplyResult{}
+
+	for _, file := range files {
+		result := ApplyResult{File: file}
+		objBytes, err := manifests(file)
+		if err != nil {
+			result.Error = fmt.Errorf("missing %q: %v", file, err)
+			ret = append(ret, result)
+			continue
+		}
+		requiredObj, err := decode(objBytes)
+		if err != nil {
+			result.Error = fmt.Errorf("cannot decode %q: %v", file, err)
+			ret = append(ret, result)
+			continue
+		}
+		result.Type = fmt.Sprintf("%T", requiredObj)
+		// NOTE: Do not add CR resources into this switch otherwise the protobuf client can cause problems.
+		switch t := requiredObj.(type) {
+		case *corev1.Namespace:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteNamespace(ctx, clients.kubeClient.CoreV1(), recorder, t)
+			}
+		case *corev1.Service:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteService(ctx, clients.kubeClient.CoreV1(), recorder, t)
+			}
+		case *corev1.Pod:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeletePod(ctx, clients.kubeClient.CoreV1(), recorder, t)
+			}
+		case *corev1.ServiceAccount:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteServiceAccount(ctx, clients.kubeClient.CoreV1(), recorder, t)
+			}
+		case *corev1.ConfigMap:
+			client := clients.configMapsGetter()
+			if client == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteConfigMap(ctx, client, recorder, t)
+			}
+		case *corev1.Secret:
+			client := clients.secretsGetter()
+			if client == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteSecret(ctx, client, recorder, t)
+			}
+		case *rbacv1.ClusterRole:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteClusterRole(ctx, clients.kubeClient.RbacV1(), recorder, t)
+			}
+		case *rbacv1.ClusterRoleBinding:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteClusterRoleBinding(ctx, clients.kubeClient.RbacV1(), recorder, t)
+			}
+		case *rbacv1.Role:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteRole(ctx, clients.kubeClient.RbacV1(), recorder, t)
+			}
+		case *rbacv1.RoleBinding:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteRoleBinding(ctx, clients.kubeClient.RbacV1(), recorder, t)
+			}
+		case *policyv1.PodDisruptionBudget:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeletePodDisruptionBudget(ctx, clients.kubeClient.PolicyV1(), recorder, t)
+			}
+		case *apiextensionsv1.CustomResourceDefinition:
+			if clients.apiExtensionsClient == nil {
+				result.Error = fmt.Errorf("missing apiExtensionsClient")
+			} else {
+				_, result.Changed, result.Error = DeleteCustomResourceDefinitionV1(ctx, clients.apiExtensionsClient.ApiextensionsV1(), recorder, t)
+			}
+		case *storagev1.StorageClass:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteStorageClass(ctx, clients.kubeClient.StorageV1(), recorder, t)
+			}
+		case *storagev1.CSIDriver:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteCSIDriver(ctx, clients.kubeClient.StorageV1(), recorder, t)
+			}
+		case *migrationv1alpha1.StorageVersionMigration:
+			if clients.migrationClient == nil {
+				result.Error = fmt.Errorf("missing migrationClient")
+			} else {
+				_, result.Changed, result.Error = DeleteStorageVersionMigration(ctx, clients.migrationClient, recorder, t)
+			}
+		case *unstructured.Unstructured:
+			if clients.dynamicClient == nil {
+				result.Error = fmt.Errorf("missing dynamicClient")
+			} else {
+				_, result.Changed, result.Error = DeleteKnownUnstructured(ctx, clients.dynamicClient, recorder, t)
 			}
 		default:
 			result.Error = fmt.Errorf("unhandled type %T", requiredObj)
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -5,7 +5,9 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/controller/manager"
+	"github.com/openshift/library-go/pkg/operator/deploymentcontroller"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -20,6 +22,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/unsupportedconfigoverridescontroller"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
@@ -54,6 +57,11 @@ type staticPodOperatorControllerBuilder struct {
 
 	// pruning information
 	pruneCommand []string
+	// deploymentManifest is the guard deployment to ensure static pods cannot be evicted
+	deploymentManifest []byte
+	// infraLister is used in guard controller to identify the cluster topology
+	infraLister configv1listers.InfrastructureLister
+
 	// TODO de-dupe this.  I think it's actually a directory name
 	staticPodPrefix string
 }
@@ -83,6 +91,7 @@ type Builder interface {
 	WithCustomInstaller(command []string, installerPodMutationFunc installer.InstallerPodMutationFunc) Builder
 	WithPruning(command []string, staticPodPrefix string) Builder
 	ToControllers() (manager.ControllerManager, error)
+	WithGuardDeployment(deploymentAssetFunc func(string) []byte, deploymentFile string) Builder
 }
 
 func (b *staticPodOperatorControllerBuilder) WithEvents(eventRecorder events.Recorder) Builder {
@@ -135,6 +144,12 @@ func (b *staticPodOperatorControllerBuilder) WithCustomInstaller(command []strin
 func (b *staticPodOperatorControllerBuilder) WithPruning(command []string, staticPodPrefix string) Builder {
 	b.pruneCommand = command
 	b.staticPodPrefix = staticPodPrefix
+	return b
+}
+
+func (b *staticPodOperatorControllerBuilder) WithGuardDeployment(deploymentAssetFunc func(string) []byte,
+	deploymentFile string) Builder {
+	b.deploymentManifest = deploymentAssetFunc(deploymentFile)
 	return b
 }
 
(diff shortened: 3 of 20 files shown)


pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -0,0 +1,199 @@
...
+
+func New(
+	name string,
+	manifest []byte,
+	recorder events.Recorder,
+	operatorClient v1helpers.OperatorClient,
+	kubeClient kubernetes.Interface,
+	deployInformer appsinformersv1.DeploymentInformer,
+	optionalInformers []factory.Informer,
+	optionalManifestHooks []ManifestHookFunc,
```
deads2k, line 65: doc the difference between this and the deploymenthooks.  The construction here is good and I understand why, but be sure to explain it godoc for others.

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -0,0 +1,199 @@
...
+	// TODO: Collapse this into optional deployment hook.
+	optionalManifestHooks []ManifestHookFunc
+	// Optional hook functions to modify the Deployment.
+	// If one of these functions returns an error, the sync
+	// fails indicating the ordinal position of the failed function.
+	// Also, in that scenario the Degraded status is set to True.
+	optionalDeploymentHooks []DeploymentHookFunc
+}
+
+func New(
```
deads2k, line 57: `NewDeploymentController` : the golang recommendations are wrong.

pkg/operator/staticpod/controllers.go
```diff
@@ -138,6 +150,17 @@ func (b *staticPodOperatorControllerBuilder) WithPruning(command []string, stati
 	return b
 }
 
+func (b *staticPodOperatorControllerBuilder) WithGuard(deploymentAssetFunc func(string) []byte, deploymentFile string, pdbAssetFunc func(string) []byte, pdbFile string) Builder {
```
deads2k, line 153: The guard logically requires the `infraLister configv1listers.InfrastructureLister`, so add `infraLister configv1listers.InfrastructureLister` as an arg to this method please.

pkg/operator/staticpod/controllers.go
```diff
@@ -262,6 +285,27 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
...
+	if len(b.deploymentManifest) > 0 && len(b.pdbManifest) > 0 {
+		manager.WithController(pdbcontroller.New(b.staticPodOperatorClient,
+			b.kubeClient,
+			b.kubeInformers,
+			b.eventRecorder,
+			b.infraLister,
+			b.pdbManifest,
+			b.operandNamespace), 1)
+		nodeInformer := b.kubeInformers.InformersFor("").Core().V1().Nodes()
+		manager.WithController(deploymentcontroller.New("static-pod",
```
deads2k, line 299: name as something with guard.  `StaticPodGuardDeployment`?

pkg/operator/pdbcontroller/pdb_controller.go
```diff
@@ -0,0 +1,147 @@
...
+	clusterTopology      configv1.TopologyMode
+	targetNamespace      string
+	pdbManifest          []byte
+	factory              *factory.Factory
+	eventRecorder        events.Recorder
+}
+
+var _ factory.Controller = &Controller{}
+
+func New(operatorClient v1helpers.OperatorClient,
```
deads2k, line 39: this should accept a name too.

deads2k, line 39: Given this `New` method a decent name, the golang recommendation of `New` is nearly always wrong.

deads2k, line 39: I'm ok with this controller as a short term solution if @sttts and  @soltysh want to land in 4.8.  If they prefer waiting until 4.9, please develop the idea of a precondition which can decorate a static resource controller.

`NewStaticResourceController().WithConditionalResource(conditionFunc, manifest, deleteConditionFunc)`.  If the deleteConditionFunc returns true, delete the resource, if the conditionFunc returns true create the resource. Deletion trumps creation.  By default, deletionCondtionFunc is !conditionFunc.

deads2k:
separate out the deploymentcontroller refactor from the rest and get storage to review.  This is the direction I'm thinking of.

pkg/operator/deploymentcontroller/deployment_controller.go
ravisantoshgudimetla: Sure, I explained it [here](https://github.com/openshift/library-go/pull/1095/files#diff-abb679a52a6ea86c466b84d87793ee37682c0e4b6a111234e69cb684d7ec3157R44)

ravisantoshgudimetla (commented):
Split the commits and addressed your comments. 

cc @bertinatto @jsafrane for reviewing f606e021e58fd2f11d250b405c98532e207519a8 and c68f4286f1fac3ae786980d6533a19e185af6931

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -101,187 +61,11 @@ func NewCSIDriverControllerServiceController(
...
-		pairs = append(pairs, []string{"${KUBE_RBAC_PROXY_IMAGE}", kubeRBACProxy}...)
-	}
-
-	// Cluster ID
-	pairs = append(pairs, []string{"${CLUSTER_ID}", clusterID}...)
-
-	// Log level
-	logLevel := loglevel.LogLevelToVerbosity(spec.LogLevel)
-	pairs = append(pairs, []string{"${LOG_LEVEL}", strconv.Itoa(logLevel)}...)
+	optionalInformers = append(optionalInformers, configInformer.Config().V1().Infrastructures().Informer())
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I was initially thinking of moving tests around but I realized that the hooks(both replicas and placeholder hooks are needed by you). The replicas function you have different from the replicas function we want in common controller, so I did not end up moving the tests but I added tests for the hooks I added in deployment.

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -0,0 +1,201 @@
...
+	progressingCondition := opv1.OperatorCondition{
+		Type:   c.name + opv1.OperatorStatusTypeProgressing,
+		Status: opv1.ConditionFalse,
+	}
+
+	if ok, msg := isProgressing(deployment); ok {
+		progressingCondition.Status = opv1.ConditionTrue
+		progressingCondition.Message = msg
+		progressingCondition.Reason = "Deploying"
+	}
```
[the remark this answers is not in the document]
ravisantoshgudimetla: We're planning to land the current PR in 4.9 timeframe. I hope you're fine with it. If not, I'll split this PR into 2
- For the refactor of deployment controller
- One for guard controller
as I want to address David's [comment ](https://github.com/openshift/library-go/pull/1095#discussion_r646860219)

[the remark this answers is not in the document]
jsafrane: no, leave it as it is, it goes the right direction. We can update it when we know what we want (the linked PR has some issues).

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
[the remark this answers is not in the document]
jsafrane: I mean now there is lot of code in `deploymentcontroller` (like the conditions setting) that is currently tested only in `csidrivercontrollerservicecontroller`. This is IMO bad, the code should be tested in package where it's written. Can you please at least *copy* the tests to `deploymentcontroller`, without any hooks, and we can then prune our CSI tests in a subsequent PR?

jsafrane:
/assign @bertinatto 
for CSI details. LGTM from my side.

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -0,0 +1,201 @@
...
+// DeploymentHookFunc is a hook function to modify the Deployment.
+type DeploymentHookFunc func(*opv1.OperatorSpec, *appsv1.Deployment) error
+
+// ManifestHookFunc is a hook function to modify the manifest in raw format.
+type ManifestHookFunc func(*opv1.OperatorSpec, []byte) ([]byte, error)
+
+// DeploymentController is a generic controller that manages a deployment
+
+//
+// <name>Available: indicates that the deployment controller  was successfully deployed and at least one Deployment replica is available.
```
bertinatto, line 34: nit: empty comment, double spaces

pkg/operator/pdbcontroller/pdb_controller.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if haTopologyMode != configv1.HighlyAvailableTopologyMode {
+		return nil
+	}
+
+	if err := pdbc.ensureGuardingPDB(ctx, syncCtx.Recorder()); err != nil {
+		return err
+	}
+
+	if err != nil {
```
bertinatto, line 100: This won't be reached because you're returning `err` above.

Also, I don't think you need to set the Degraded condition yourself, as it's set when you return an error.

pkg/operator/pdbcontroller/pdb_controller.go
```diff
@@ -0,0 +1,152 @@
...
+		}))
+		if updateErr != nil {
+			syncCtx.Recorder().Warning("PDBControllerUpdatingStatus", updateErr.Error())
+		}
+		return err
+	}
+
+	_, _, updateErr := v1helpers.UpdateStatus(pdbc.operatorClient,
+		v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   "PDBControllerDegraded",
```
bertinatto, line 115: Is there any reason you're setting Degraded here? Typically you would only need to set Available.

pkg/operator/pdbcontroller/pdb_controller.go
```diff
@@ -0,0 +1,152 @@
...
+	return updateErr
+}
+
+func (pdbc *PDBController) ensureGuardingPDB(ctx context.Context, recorder events.Recorder) error {
+	guardPDB := resourceread.ReadPodDisruptionBudgetV1OrDie(pdbc.pdbManifest)
+
+	_, _, err := resourceapply.ApplyPodDisruptionBudget(pdbc.kubeClient.PolicyV1(), recorder, guardPDB)
+	if err != nil {
+		klog.Errorf("Failed to verify/apply %s pdb, error %w", guardPDB.Name, err)
+		return err
```
bertinatto, line 128: nit: no need to log the error if we're returning it

bertinatto (commented):
Added a few comments.

I'm not opposed to merging 8ec123cbd3988ccb08cdcc6080163ed4972ef981, but is it possible to use the generic Deployment controller with a brand new controller first?

Also, besides the `CSIDriverControllerServiceController` we also have a very similar controller that deploys a DaemonSet. Will there be a `DaemonSetController`?

ravisantoshgudimetla (commented):
> Added a few comments.
> 
> I'm not opposed to merging [8ec123c](https://github.com/openshift/library-go/commit/8ec123cbd3988ccb08cdcc6080163ed4972ef981), but is it possible to use the generic Deployment controller with a brand new controller first?
> 
> Also, besides the `CSIDriverControllerServiceController` we also have a very similar controller that deploys a DaemonSet. Will there be a `DaemonSetController`?

@jsafrane and @bertinatto - thank you both for the reviews. I have addressed your comments. PTAL.

As discussed over slack, we're in the process of using generic deployment in all control-plane components. But we would like to get this PR merged before we go using it. Yes, we also have plans of having a DSController and Stscontroller in future. So, let us know if you already have a DSController, you can contribute it to library-go directly.

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -52,6 +52,18 @@ type ApplyResult struct {
...
+// ConditionalFunction provides needed dependency for a resource on another condition instead of blindly creating
+// a resource. This conditional function can also be used to delete the resource when not needed
+type ConditionalFunction func() bool
+
+// ResourceConditionalMap maps a file to conditional functions that can be used to create or delete a resource in the
+// given file.
+type ResourceConditionalMap struct {
+	File                  string
+	DeleteConditionalFunc ConditionalFunction
+	CreateConditionalFunc ConditionalFunction
```
jsafrane, line 64: What if both functions return `true`? IMO it would be better to have a single function that returns eg. `ResourceAction`, which can be either `"Create"` or `"Delete"`

ravisantoshgudimetla: https://github.com/openshift/library-go/pull/1095/commits/59ec4d268990a31f4c0604ce2dc2c509ab5cac2c#diff-b91f253f2414046d3d35cc689c04410dfc00c7b8130a953ac59926118f9b8423R119 - Deletion trumps creation.

jsafrane: Yes, I saw it. But what benefit do the two function have over a single function? IMO it makes the code more error prone.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -102,25 +114,40 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
 	return c
 }
 
+// WithConditionalResource makes the controller react to changes in dependent object
+// If the deleteConditionFunc returns true, delete the resource, if the conditionFunc returns true create the resource.
+// Deletion trumps creation. By default, deletionConditionFunc is !conditionFunc.
+// TODO: This can be made multi-resource aware as well.
+func (c *StaticResourceController) WithConditionalResource(createConditionFunc resourceapply.ConditionalFunction,
```
deads2k, line 121: `createConditionFn, deleteConditionFn, files ...string`

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -102,25 +114,40 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
...
+	file string, deleteConditionFunc resourceapply.ConditionalFunction) *StaticResourceController {
+	rcm := resourceapply.ResourceConditionalMap{
```
deads2k, line 123:
```go
conditionalManifests := (read them)
conditionalApplyFn := func() (directResourceResults, error){
  if deletionCondtionFn(){
    return resourceapply.DeleteAll(ctx, c.clients, syncContext.Recorder(), conditionalManifests), nil
  }
  if createConditionFn(){
    return resourceapply.ApplyDirectly(ctx, c.clients, syncContext.Recorder(), conditionalManifests), nil
  }
  return nil, nil
}
c.ConditionalApplyFns = append(conditionalApplyFn)

```

Then you add these below the existing ApplyDirectly in Sync.

ravisantoshgudimetla: @deads2k - I slightly tweaked your suggestion and applied the changes. PTAL when you find time.

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -40,3 +40,20 @@ func ApplyCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1.
...
+func DeleteCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, recorder events.Recorder, required *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, bool, error) {
+	existing, err := client.CustomResourceDefinitions().Get(ctx, required.Name, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, false, nil
+		} else {
+			return nil, false, err
+		}
+	}
+	err = client.CustomResourceDefinitions().Delete(ctx, existing.Name, metav1.DeleteOptions{})
```
deads2k, line 53: you can save a call by simply calling delete and ignoring NotFound.

ravisantoshgudimetla: I believe you don't want me to do a get? I am fine with it.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -219,6 +265,24 @@ func (c StaticResourceController) Sync(ctx context.Context, syncContext factory.
 			continue
 		}
 	}
+	// Handle conditional resources that need to be created or deleted.
+	if c.conditionalResource != nil && len(c.conditionalResource.conditionalFiles) > 0 {
+		var conditionalResourceResults []resourceapply.ApplyResult
+		if c.conditionalResource.shouldDelete {
+			conditionalResourceResults = resourceapply.DeleteAll(ctx, c.clients, syncContext.Recorder(), c.manifests, c.conditionalResource.conditionalFiles...)
+		} else {
```
deads2k, line 273: you shouldn't create unless the create function returns true.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -102,6 +111,43 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
 	return c
 }
 
+func shouldDelete(createConditionalFunc, deleteConditionalFunc resourceapply.ConditionalFunction) bool {
+	shouldDelete := false
+	if createConditionalFunc == nil && deleteConditionalFunc == nil {
+		shouldDelete = false
+	} else {
+		if createConditionalFunc != nil && createConditionalFunc() {
```
deads2k, line 119: delete should only be controlled by the deletionfunc.  Create should only be controlled by the creation func.

if the create and delete func both report true, an error should be reported (do this in sync)

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -219,6 +265,24 @@ func (c StaticResourceController) Sync(ctx context.Context, syncContext factory.
 			continue
 		}
 	}
+	// Handle conditional resources that need to be created or deleted.
+	if c.conditionalResource != nil && len(c.conditionalResource.conditionalFiles) > 0 {
+		var conditionalResourceResults []resourceapply.ApplyResult
+		if c.conditionalResource.shouldDelete {
```
deads2k, line 271:
```go
shouldDelete := false
if shouldDeleteFn != nil && deleteFn(){
  shouldDelete = true
}

shouldCreate != false
if shouldCreateFn != nil && shouldCreateFn(){
  shouldCreate = true
}

if shouldCreate && shouldDelete{
  errors = append(errors, wtf)
}

if shouldDelete{
  deleteAll
} else if shouldCreate{
  createAll
}

accumulateresults.
```

jsafrane:
I have another use case for StaticResourceController removing objects it manages: OLM / optional operators. Here I need StaticResourceController to remove a finalizer on a CR when everything has been removed. I think it is not possible with current callbacks - there is no way how to tell if StaticResourceController has finished removing everything it wanted or not.

Perhaps I miss some context, does static pod controller need a similar signal to tell that PDB has been removed?

ravisantoshgudimetla:
> I have another use case for StaticResourceController removing objects it manages: OLM / optional operators. Here I need StaticResourceController to remove a finalizer on a CR when everything has been removed. I think it is not possible with current callbacks - there is no way how to tell if StaticResourceController has finished removing everything it wanted or not.


The goal in this case is to create, delete static resource based on conditions not the other way around but I am wondering if your use-case can be solved by simple chaining. How do you create CR? Using static resource controller?

jsafrane:
My CR will be created / deleted by user. What do you mean by the "simple chaining"?

pkg/operator/v1helpers/interfaces.go
```diff
@@ -19,7 +19,7 @@ type OperatorClient interface {
 }
 
 type StaticPodOperatorClient interface {
-	OperatorClient
+	OperatorClientWithFinalizers
```
soltysh, line 22: Why this change, I'm not seeing it being used here nor in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/543

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -102,6 +112,46 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
...
+//}
+
+func (c *conditionalResource) shouldDelete() {
+
+}
+
+// WithConditionalResource makes the controller react to changes in dependent object
+// If both createConditional and deleteConditional are true, it will be considered an
+// error during sync.
+func (c *StaticResourceController) WithConditionalResource(createConditionalFunc resourceapply.ConditionalFunction,
```
soltysh, line 145: Maybe `WithConditionalResources`?

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -223,6 +227,135 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
...
+			} else {
+				_, result.Changed, result.Error = DeleteStorageVersionMigration(ctx, clients.migrationClient, recorder, t)
+			}
+		case *unstructured.Unstructured:
+			if clients.dynamicClient == nil {
+				result.Error = fmt.Errorf("missing dynamicClient")
+			} else {
+				_, result.Changed, result.Error = DeleteKnownUnstructured(ctx, clients.dynamicClient, recorder, t)
+			}
+		default:
```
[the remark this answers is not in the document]
ravisantoshgudimetla: To be in sync with `ApplyDirectly` method.

pkg/operator/staticresourcecontroller/static_resource_controller.go
ravisantoshgudimetla: Sure. I changed it.

pkg/operator/v1helpers/interfaces.go
ravisantoshgudimetla: It is being used in https://github.com/openshift/library-go/pull/1095/commits/1df083fadac8328066bb845a4d6049a2870742c5#diff-d0ce926df996bfa572967a0d2a1d011bbfc84cd41e065b957328285647b17cecR293, technically deploymentController started using it and I had to modify the methods to be implemented in staticPodOperatorClient :( but I think in the long run this good as we can provide OperatorClientWithFinalizers for all the staticpodClients.

ravisantoshgudimetla (commented):
@soltysh - I answered your questions. PTAL

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -58,9 +58,19 @@ type StaticResourceController struct {
 
 	eventRecorder events.Recorder
 
-	factory          *factory.Factory
-	restMapper       meta.RESTMapper
-	categoryExpander restmapper.CategoryExpander
+	factory *factory.Factory
+
+	conditionalResource *conditionalResource
```
deads2k, line 63: this should be a slice

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -219,6 +242,36 @@ func (c StaticResourceController) Sync(ctx context.Context, syncContext factory.
...
+		if c.conditionalResource.deleteConditional != nil && c.conditionalResource.deleteConditional() {
+			shouldDelete = true
+		}
+
+		shouldCreate := false
+		if c.conditionalResource.createConditional != nil && c.conditionalResource.createConditional() {
+			shouldCreate = true
+		}
+
+		if shouldCreate && shouldDelete {
```
deads2k, line 258: switch/case please

pkg/operator/staticresourcecontroller/static_resource_controller.go
ravisantoshgudimetla: The [fields](https://github.com/openshift/library-go/pull/1095/files#diff-b91f253f2414046d3d35cc689c04410dfc00c7b8130a953ac59926118f9b8423R71) that this structure is holding is a slice. Do we still need make this slice?

deads2k: based on your current types, `conditionalResources []conditionalResource`.  I should be able to call `WithConditionalResources` multiple times.

ravisantoshgudimetla: Ok. I can make that change.

ravisantoshgudimetla (commented):
@soltysh - Updated and modified the units. PTAL

soltysh (approved):
The changes here looks reasonable, but before merging I'd like to see how this is being used (updated version) in https://github.com/openshift/cluster-kube-controller-manager-operator/pull/543 Especially such that we don't have to introduce yet another pod but rather re-use one of the pre-existing.

PR #1096, opened 2021-06-07
image/registryclient: set insecure and return early in blobMirroredRepository

soltysh:
This is followup to https://github.com/openshift/library-go/pull/1084 fixing:
- setting `insecure` properly for `blobMirroredRepository`
- return proper error if no alternates found
- unit tests for the above two
- replace `golang.org/x/net/context` with built-in `context` package

### DIFF
--- a/pkg/image/registryclient/client_mirrored.go
+++ b/pkg/image/registryclient/client_mirrored.go
@@ -216,6 +216,9 @@ func (r *blobMirroredRepository) alternates(ctx context.Context, fn func(r Repos
 		if err != nil {
 			return err
 		}
+		if len(alternates) == 0 {
+			return attemptErr
+		}
 		if alternateErr := r.attemptRepos(ctx, alternates, fn); alternateErr != nil {
 			return attemptErr
 		}
--- a/pkg/image/registryclient/client_mirrored.go
+++ b/pkg/image/registryclient/client_mirrored.go
@@ -1,6 +1,7 @@
 package registryclient
 
 import (
+	"context"
 	"fmt"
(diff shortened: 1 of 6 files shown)


pkg/image/registryclient/client_mirrored.go
```diff
@@ -216,6 +216,9 @@ func (r *blobMirroredRepository) alternates(ctx context.Context, fn func(r Repos
 		if err != nil {
 			return err
 		}
+		if len(alternates) == 0 {
+			return attemptErr
+		}
```
smarterclayton, line 221: This looks correct but now I want to see a unit test for it.  This is the real loop we need to have unit tests on anyway.

pkg/image/registryclient/client_mirrored.go
```diff
@@ -216,6 +216,9 @@ func (r *blobMirroredRepository) alternates(ctx context.Context, fn func(r Repos
 		if err != nil {
 			return err
 		}
+		if len(alternates) == 0 {
```
soltysh, line 219: @smarterclayton alternatively we could check for nil strategy in https://github.com/openshift/library-go/blob/master/pkg/image/registryclient/client_mirrored.go#L207 similarly to [initialRepos](https://github.com/openshift/library-go/blob/master/pkg/image/registryclient/client_mirrored.go#L122) or [errorRepos](https://github.com/openshift/library-go/blob/master/pkg/image/registryclient/client_mirrored.go#L145)

ricardomaraschini (approved):
/lgtm

PR #1097, opened 2021-06-07
Bug 1879182: make secure- and normal audit policies match

stlaz:
/assign @sttts

We no longer need to be treating the oauth tokens differently as they bear very little information that could be misused.

### DIFF
--- a/pkg/operator/apiserver/audit/testdata/audit-policies-cm-scenario-1.yaml
+++ b/pkg/operator/apiserver/audit/testdata/audit-policies-cm-scenario-1.yaml
@@ -18,11 +18,6 @@ data:
       resources:
       - group: ""
         resources: ["events"]
-    # Don't log oauth tokens as metadata.name is the secret
-    - level: None
-      resources:
-      - group: "oauth.openshift.io"
-        resources: ["oauthaccesstokens", "oauthauthorizetokens"]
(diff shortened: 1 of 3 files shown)


pkg/operator/apiserver/audit/testdata/audit-policies-cm-scenario-1.yaml
```diff
@@ -18,11 +18,6 @@ data:
       resources:
       - group: ""
         resources: ["events"]
-    # Don't log oauth tokens as metadata.name is the secret
-    - level: None
-      resources:
-      - group: "oauth.openshift.io"
-        resources: ["oauthaccesstokens", "oauthauthorizetokens"]
```
EmilyM1, line 25: Can we remove the groups and go to userGroups key without issue?

stlaz: This is an array of specific configurations, I don't think there should be a problem removing an element, if that's what you're asking.

sttts:
/hold

this is the wrong way. Remove the unnecessary policies.

stlaz:
/test unit
that does not seem to be relevant to this PR, interesting for a unit test to fail like that though

PR #1099, opened 2021-06-07
Bug 1961317: Adding static assets to related objects

Elbehery:
This PR adds method to get `RelatedObjects` from `StaticResourceController`

This PR required by https://github.com/openshift/cluster-storage-operator/pull/175

### DIFF
--- a/pkg/operator/staticresourcecontroller/static_resource_controller.go
+++ b/pkg/operator/staticresourcecontroller/static_resource_controller.go
@@ -2,7 +2,9 @@ package staticresourcecontroller
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	configv1 "github.com/openshift/api/config/v1"
 	"strings"
 	"time"
 
@@ -57,7 +59,8 @@ type StaticResourceController struct {
 
 	eventRecorder events.Recorder
 
-	factory *factory.Factory
+	factory    *factory.Factory
+	restMapper meta.RESTMapper
 }
 
 // NewStaticResourceController returns a controller that maintains certain static manifests. Most "normal" types are supported,
@@ -180,6 +183,11 @@ func (c *StaticResourceController) AddInformer(informer cache.SharedIndexInforme
 	return c
 }
 
+func (c *StaticResourceController) AddRESTMapper(mapper meta.RESTMapper) *StaticResourceController {
+	c.restMapper = mapper
+	return c
+}
+
 func (c *StaticResourceController) AddNamespaceInformer(informer cache.SharedIndexInformer, namespaces ...string) *StaticResourceController {
 	c.factory.WithNamespaceInformer(informer, namespaces...)
 	return c
@@ -238,6 +246,53 @@ func (c StaticResourceController) Sync(ctx context.Context, syncContext factory.
 
 func (c *StaticResourceController) Name() string {
 	return "StaticResourceController"
+}
+
+func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference, error) {
+	if c.restMapper == nil {
+		return nil, errors.New("StaticResourceController.restMapper is nil")
+	}
+
+	acc := make([]configv1.ObjectReference, 0)
+	errors := []error{}
+
+	for _, file := range c.files {
+		// parse static asset
+		objBytes, err := c.manifests(file)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		requiredObj, _, err := genericCodec.Decode(objBytes, nil, nil)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		metadata, err := meta.Accessor(requiredObj)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		// map gvk from static asset to gvr
+		gvk := requiredObj.GetObjectKind().GroupVersionKind()
+		mapping, err := c.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		gvr := mapping.Resource
+		acc = append(acc, configv1.ObjectReference{
+			Group:     gvk.Group,
+			Resource:  gvr.Resource,
+			Namespace: metadata.GetNamespace(),
+			Name:      metadata.GetName(),
+		})
+	}
+
+	if len(errors) > 0 {
+		return nil, utilerrors.NewAggregate(errors)
+	}
+	return acc, nil
 }
 
 func (c *StaticResourceController) Run(ctx context.Context, workers int) {
--- a/pkg/operator/staticresourcecontroller/static_resource_controller.go
+++ b/pkg/operator/staticresourcecontroller/static_resource_controller.go
@@ -314,10 +314,7 @@ func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference,
 		})
 	}
 
-	if len(errors) > 0 {
-		return nil, utilerrors.NewAggregate(errors)
-	}
-	return acc, nil
+	return acc, utilerrors.NewAggregate(errors)
 }
 
 func (c *StaticResourceController) Run(ctx context.Context, workers int) {
--- a/pkg/operator/staticresourcecontroller/static_resource_controller.go
+++ b/pkg/operator/staticresourcecontroller/static_resource_controller.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/restmapper"
 	"strings"
 	"time"
 
@@ -59,8 +61,9 @@ type StaticResourceController struct {
 
 	eventRecorder events.Recorder
 
-	factory    *factory.Factory
-	restMapper meta.RESTMapper
+	factory          *factory.Factory
+	restMapper       meta.RESTMapper
+	categoryExpander restmapper.CategoryExpander
 }
 
 // NewStaticResourceController returns a controller that maintains certain static manifests. Most "normal" types are supported,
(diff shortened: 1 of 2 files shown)


pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +247,49 @@ func (c *StaticResourceController) Name() string {
...
 
+func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference, error) {
+	acc := make([]configv1.ObjectReference, 0)
+	errors := []error{}
+
+	for _, file := range c.files {
+		// parse static asset
+		objBytes, err := c.manifests(file)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("missing %q: %v", file, err))
```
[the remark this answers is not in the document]
Elbehery: 👍🏽  I agree, it just the way it was used in during parsing, so I did not want to break the flow :)

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +247,49 @@ func (c *StaticResourceController) Name() string {
 	return "StaticResourceController"
 }
 
+func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference, error) {
+	acc := make([]configv1.ObjectReference, 0)
```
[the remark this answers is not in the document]
Elbehery: done

pkg/operator/staticresourcecontroller/static_resource_controller.go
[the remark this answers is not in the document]
Elbehery: done

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +248,53 @@ func (c *StaticResourceController) Name() string {
...
 
+func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference, error) {
+	if c.restMapper == nil {
+		return nil, errors.New("StaticResourceController.restMapper is nil")
+	}
+
+	acc := make([]configv1.ObjectReference, 0)
+	errors := []error{}
+
+	for _, file := range c.files {
```
deads2k, line 259: I'd like to be smarter.  I think the rules are closer to 
1. if cluster scoped resource, list that exact resource
2. if namespace scoped and included in the "all" alias, just include the namespace.  de-dupe namespaces
3. if namespace and not included in the "all" alias, list that exact resource.

deads2k: See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/restmapper/category_expansion.go#L48-L88 for how to determine "all"

Elbehery: 👍🏽

jsafrane: I think it's not necessary to add namespaces to RelatedObjects - someone must have created the namespace and therefore add it it its RelatedObjects.

(At least in storage operators we have them hardcoded in RelatedObjects).

Elbehery: +1

deads2k: > I think it's not necessary to add namespaces to RelatedObjects - someone must have created the namespace and therefore add it it its RelatedObjects.
> 
> (At least in storage operators we have them hardcoded in RelatedObjects).

The idea is avoid an explosion of individual resources that are not required and slow down collection.  Right now you're iterating through manifests and enumerating all of them.  This will result in entries like
1. ns=foo secret=one
2. ns=foo configmap=two
3. ns=foo serviceaccount=three
4. ns=foo
5. ns=foo rolebinding=four
6. ns=foo servicemonitor=five

What we want is to collapse all that to
1. ns=foo
2. ns=foo servicemonitor=five

Because we know that oc adm inspect will collect the "all" category for ns=foo and collect the individual resources for "all".  And servicemonitor is not in "all" (I don't think).

if you don't inspect all, please at least decimate resources by namespace for 4.8.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +256,69 @@ func (c *StaticResourceController) Name() string {
...
+	if c.restMapper == nil {
+		return nil, errors.New("StaticResourceController.restMapper is nil")
+	}
+
+	if c.categoryExpander == nil {
+		return nil, errors.New("StaticResourceController.categoryExpander is nil")
+	}
+
+	// create lookup for resources in "all" alias
+	grs, _ := c.categoryExpander.Expand("all")
```
jsafrane, line 269: @deads2k does it make sense to return an error here if `Expand` returns `false`? Or how else retry failed expansion?

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +256,69 @@ func (c *StaticResourceController) Name() string {
...
+		mapping, err := c.restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		if err != nil {
+			errors = append(errors, err)
+			continue
+		}
+		gvr := mapping.Resource
+		// filter out resources within "all" alias from result
+		if _, ok := lookup[gvr.GroupResource()]; ok {
+			continue
+		}
```
jsafrane, line 301: You should skip members of `all` *only* for namespaced resources (i.e. `namespace != ""`)

pkg/operator/staticresourcecontroller/static_resource_controller.go
jsafrane: > What we want is to collapse all that to
> 1. ns=foo
> 2. ns=foo servicemonitor=five

I am arguing that we can skip "1. ns=foo", unless the `kind: namespace` is directly in the controller assets.

At least in storage operators we either create namespaces with StaticResourceController, and hence it will be returned in `RelatedObjects()` here as any other non-namespaced object , or CVO creates it from `manifests/` and we have it hardcoded in RelatedObjects in the operator.

IMO StaticResourceController should not report objects it does not create. Something else must have created them and that component is responsible for reporting them.

pkg/operator/staticresourcecontroller/static_resource_controller.go
Elbehery: yes I assumed that since we parse static assets, it should be all namespaced, but I agree should be explicit, on it 👍🏽

pkg/operator/staticresourcecontroller/static_resource_controller.go
deads2k: > @deads2k does it make sense to return an error here if `Expand` returns `false`? Or how else retry failed expansion?

I think the current code is ok.  It may flap briefly, but should settle quickly.  If this fails, your restmapping is likely to fail overall.

If you were trying to harden it, I would probably say that every namespaced GR is logically in "all".

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -240,6 +256,70 @@ func (c *StaticResourceController) Name() string {
...
+		acc = append(acc, configv1.ObjectReference{
+			Group:     gvk.Group,
+			Resource:  gvr.Resource,
+			Namespace: metadata.GetNamespace(),
+			Name:      metadata.GetName(),
+		})
+	}
+
+	if len(errors) > 0 {
+		return nil, utilerrors.NewAggregate(errors)
```
deads2k, line 318: seems like including the related resources you know of in addition to the errors makes sense.  Let the caller decide if some are better than none (spoiler, some is probably better than none).

deads2k:
The change here matches what I was hoping for.  Looks like you are short a test?

/approve
/hold for test

deads2k:
> The change here matches what I was hoping for. Looks like you are short a test?

Specifically, a namespace scoped resource that isn't included in "all" should be present in the returned list.  You aren't likely to notice it's missing until you're missing something from must-gather at a customer site.

pkg/operator/staticresourcecontroller/static_resource_controller.go
Elbehery: I can't get what u mean here tbh

pkg/operator/staticresourcecontroller/static_resource_controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+	assert.ElementsMatch(t, expected, res)
+}
+
+func readBytesFromString(yaml string) ([]byte, error) {
+	data, err := ioutil.ReadAll(strings.NewReader(yaml))
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
```
jsafrane, line 65: If you really want to use file name as a content:
```go
return []byte(yaml), nil
```

pkg/operator/staticresourcecontroller/static_resource_controller.go
deads2k: > IMO StaticResourceController should not report objects it does not create. Something else must have created them and that component is responsible for reporting them.

I can accept that position as a starting point.  In my example, the namespace was created as item 4.

pkg/operator/staticresourcecontroller/static_resource_controller.go
deads2k: return acc, utilerrors.NewAggregate(errors)

pkg/operator/staticresourcecontroller/static_resource_controller_test.go
```diff
@@ -0,0 +1,66 @@
...
+		nil,
+	)
+	files := []string{sa, secret}
+	src := NewStaticResourceController("", readBytesFromString, files, nil, operatorClient, events.NewInMemoryRecorder(""))
+	src = src.AddRESTMapper(restMapper).AddCategoryExpander(expander)
+	res, _ := src.RelatedObjects()
+	assert.ElementsMatch(t, expected, res)
+}
+
+func readBytesFromString(yaml string) ([]byte, error) {
```
deads2k, line 59: for the runes in this test, isn't this simply `[]byte(yaml)` ?

deads2k:
/hold cancel

I'll leave lgtm with @jsafrane

pkg/operator/staticresourcecontroller/static_resource_controller_test.go
Elbehery: make sense 👍🏽

pkg/operator/staticresourcecontroller/static_resource_controller_test.go
Elbehery: +1 👍🏽

PR #1103, opened 2021-06-09
Bug 1963730: writes the dynamic certs atomically

p0lyn0mial:
As of today, the dynamic certificates i.e. `kube-apiserver-certs` are accessed by at least two processes, namely the installer pod and the cert-syncer.
Up until now, there was no coordination between these processes that might have lead to many unexpected errors, like https://bugzilla.redhat.com/show_bug.cgi?id=1963730

This PR writes all certificates in an atomic way by first creating a temporary file, writing the content to it, and then renaming it to the original file. 

`os.Rename` calls `syscall.Rename` which in turn uses the rename syscall (Linux) which provides atomicity (https://man7.org/linux/man-pages/man2/rename.2.html)

The previous attempt didn't work as it would break the DR scripts - https://github.com/openshift/library-go/pull/1098

### DIFF
--- a/pkg/operator/staticpod/file_utils.go
+++ b/pkg/operator/staticpod/file_utils.go
@@ -1,0 +1,36 @@
+package staticpod
+
+import (
+	"os"
+	"path"
+
+	"io/ioutil"
+	"k8s.io/klog/v2"
+)
+
+func WriteFileAtomic(content []byte, filePerms os.FileMode, resource, contentDir, filename string) error {
+	tmpFile, err := writeTemporaryFile(content, filePerms, contentDir, filename)
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("Renaming %s manifest %q to %q ...", resource, tmpFile, path.Join(contentDir, filename))
+	return os.Rename(tmpFile, path.Join(contentDir, filename))
+}
+
+func writeTemporaryFile(content []byte, filePerms os.FileMode, contentDir, filename string) (string, error) {
+	klog.Infof("Creating a temporary file for %q ...", path.Join(contentDir, filename))
+	tmpfile, err := ioutil.TempFile(contentDir, filename)
+	if err != nil {
+		return "", err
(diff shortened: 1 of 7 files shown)


pkg/operator/staticpod/file_utils.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return err
+	}
+
+	klog.Infof("Renaming %s manifest %q to %q ...", resource, tmpFile, path.Join(contentDir, filename))
+	return os.Rename(tmpFile, path.Join(contentDir, filename))
+}
+
+func writeTemporaryFile(content []byte, filePerms os.FileMode, contentDir, filename string) (string, error) {
+	klog.Infof("Creating a temporary file for %q ...", path.Join(contentDir, filename))
```
sttts, line 23: do we really need all this logging? Would streamline the func `WriteFileAtomic` to not output anything.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -408,3 +398,29 @@ func (o *InstallOptions) Run(ctx context.Context) error {
 	recorder.Eventf("StaticPodInstallerCompleted", "Successfully installed revision %s", o.Revision)
 	return nil
 }
+
+func writeConfig(content []byte, atomic bool, contentDir, filename string) error {
```
sttts, line 402: why do we need atomic bool?

p0lyn0mial: because not all files need to be written in that way. The other files are written to a per revision directory, like https://github.com/openshift/library-go/blob/3af072dc814786e3c97f88680db3b9ca6b7cecba/pkg/operator/staticpod/installerpod/cmd.go#L274

we could use `prefixed` arg but that would be a little less explicit.

p0lyn0mial:
/hold

until proof PRs are green

PR #1104, opened 2021-06-09
update leader election defaults so it handles 60s of kube-apiserver communication disruption

deads2k:
found via https://github.com/openshift/origin/pull/26215

We want to handle 60s of communication disruption in all components.

### DIFF
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -84,14 +84,28 @@ func ToConfigMapLeaderElection(clientConfig *rest.Config, config configv1.Leader
 func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace, defaultName string) configv1.LeaderElection {
 	ret := *(&config).DeepCopy()
 
+	// We want to be able to tolerate 60s of kube-apiserver disruption without causing pod restarts.
+	// We want the graceful lease re-acquisition fairly quick to avoid waits on new deployments and other rollouts.
+	// We want a single set of guidance for nearly every lease in openshift.  If you're special, we'll let you know.
+	// 1. clock skew tolerance is leaseDuration-renewDeadline == 30s
+	// 2. kube-apiserver downtime tolerance is == 78s
+	//      lastRetry=floor(renewDeadline/retryPeriod)*retryPeriod == 104
+	//      downtimeTolerance = lastRetry-retryPeriod == 78s
+	// 3. worst non-graceful lease acquisition is leaseDuration+retryPeriod == 163s
+	// 4. worst graceful lease acquisition is retryPeriod == 26s
 	if ret.LeaseDuration.Duration == 0 {
-		ret.LeaseDuration.Duration = 60 * time.Second
+		ret.LeaseDuration.Duration = 137 * time.Second
 	}
+
 	if ret.RenewDeadline.Duration == 0 {
-		ret.RenewDeadline.Duration = 35 * time.Second
+		// this gives 107/26=4 retries and allows for 137-107=30 seconds of clock skew
+		// if the kube-apiserver is unavailable for 60s starting just before t=26 (the first renew),
+		// then we will retry on 26s intervals until t=104 (kube-apiserver came back up at 86), and there will
+		// be 33 seconds of extra time before the lease is lost.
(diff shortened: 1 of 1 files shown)


pkg/config/leaderelection/leaderelection.go
```diff
@@ -85,13 +85,21 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace,
...
+	// these total 30s and then 20% jitter on RetryPeriod.  This gives us two properties
+	//  1. We can span a 60s kube-apiserver communication interruption (30+jitter1 + 30+jitter2) is always at least 60s.
+	//  2. We can always retry at least three times inside of our leaseDuration even in cases of long jitter and long
+	//     waits (black hole) that causes us to bump the RenewDeadline.  In cases of a short renew failure, we may retry
+	//     more often.
 	if ret.RenewDeadline.Duration == 0 {
-		ret.RenewDeadline.Duration = 35 * time.Second
+		// we choose a very long RenewDeadline so that case of very slow etcd are exacerbated by lost leases and
+		// cascading retries.
+		ret.RenewDeadline.Duration = 17 * time.Second
```
smarterclayton, line 99: Note if you do this you only get once chance to hold the lease, so I would expect this to make mots leases fail.  Having LeaseDeadline >>> LeaseDuration makes no sense.  You want to have about 10-20s of slack between RenewDeadline and LeaseDuration.  I.e.:

99 lease, 70 renew, retry period 34s makes some sense (You get two tries)
99 lease, 70 renew, retry period 60s makes sense (we try one time)

pkg/config/leaderelection/leaderelection.go
```diff
@@ -85,13 +85,21 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace,
...
 	}
 	if ret.RetryPeriod.Duration == 0 {
-		ret.RetryPeriod.Duration = 10 * time.Second
+		ret.RetryPeriod.Duration = 13 * time.Second
```
smarterclayton, line 102: Note that a 13s retry period is very short.  Most components would be ok with 30s retry period, 60s renew deadline, and 90s lease.  For a core critical controller 13s is reasonable.

pkg/config/leaderelection/leaderelection.go
```diff
@@ -85,13 +85,19 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace,
 	ret := *(&config).DeepCopy()
 
 	if ret.LeaseDuration.Duration == 0 {
-		ret.LeaseDuration.Duration = 60 * time.Second
+		// account for 20% jitter of the RetryPeriod
+		ret.LeaseDuration.Duration = 99 * time.Second
```
smarterclayton, line 89: this assumes no clock skew, which is probably not totally safe

pkg/config/leaderelection/leaderelection.go
```diff
@@ -84,14 +84,25 @@ func ToConfigMapLeaderElection(clientConfig *rest.Config, config configv1.Leader
 func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace, defaultName string) configv1.LeaderElection {
 	ret := *(&config).DeepCopy()
 
+	// 1. lock skew tolerance is leaseDuration-renewDeadline == 22s
```
smarterclayton, line 87: I might suggest we call this method `FastLeaderElectionDefaulting` or `CriticalLeaderElectionDefaulting`, since these defaults should only be used for the most critical services.  Are there a set of library-go based operators that are not critical that can tolerate 30s / 3 retry leases?

One more question - why are you trying to have 6 retries (i.e. what reason for 6 vs 3)?

smarterclayton:
I appreciate the commit title.

LGTM

pkg/config/leaderelection/leaderelection.go
deads2k: > I might suggest we call this method `FastLeaderElectionDefaulting` or `CriticalLeaderElectionDefaulting`, since these defaults should only be used for the most critical services. Are there a set of library-go based operators that are not critical that can tolerate 30s / 3 retry leases?
> 
> One more question - why are you trying to have 6 retries (i.e. what reason for 6 vs 3)?

Now answered in the code comments.

PR #1105, opened 2021-06-10
csr approver: choose between logging and status updates on errors

stlaz:
Add a noop operator client for generic controllers that are a part
of an operand rather than an operator.

### DIFF
--- a/pkg/operator/csr/csr_approver.go
+++ b/pkg/operator/csr/csr_approver.go
@@ -45,6 +45,9 @@ type csrApproverController struct {
 // NewCSRApproverController returns a controller that is observing the CSR API
 // for a CSR of a given name. If such a CSR exists, it runs the `csrApprover.Approve()`
 // against it and either denies, approves or leaves the CSR.
+//
+// If operatorClient is nil, the controller will log the errors instead of reporting
+// them in an operator status.
 func NewCSRApproverController(
 	controllerName string,
 	operatorClient v1helpers.OperatorClient,
@@ -68,14 +71,18 @@ func NewCSRApproverController(
 		return csrFilter.Matches(csrObj)
 	}
 
-	return factory.New().
+	f := factory.New().
 		WithSync(c.sync).
-		WithSyncDegradedOnError(operatorClient).
-		WithFilteredEventsInformersQueueKeyFunc(factory.ObjectNameToKey, csrFilterConverted, csrInformers.Informer()).
-		ToController(
-			"WebhookAuthenticatorCertApprover_"+controllerName,
-			eventsRecorder.WithComponentSuffix("webhook-authenticator-cert-approver-"+controllerName),
-		)
+		WithFilteredEventsInformersQueueKeyFunc(factory.ObjectNameToKey, csrFilterConverted, csrInformers.Informer())
+
+	if operatorClient != nil {
+		f.WithSyncDegradedOnError(operatorClient)
+	}
+
+	return f.ToController(
+		"WebhookAuthenticatorCertApprover_"+controllerName,
+		eventsRecorder.WithComponentSuffix("webhook-authenticator-cert-approver-"+controllerName),
+	)
 }
 
 func (c *csrApproverController) sync(ctx context.Context, syncCtx factory.SyncContext) error {


pkg/operator/v1helpers/helpers.go
```diff
@@ -21,6 +22,41 @@ import (
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
+var _ OperatorClient = &NoOpOperatorClient{}
```
s-urbaniak, line 25: :+1:

pkg/operator/v1helpers/helpers.go
```diff
@@ -21,6 +22,41 @@ import (
...
+func (c *NoOpOperatorClient) GetObjectMeta() (*metav1.ObjectMeta, error) {
+	return &metav1.ObjectMeta{}, nil
+}
+
+func (c *NoOpOperatorClient) Informer() cache.SharedIndexInformer {
+	return nil
+}
+
+func (c *NoOpOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
+	return &operatorv1.OperatorSpec{}, &operatorv1.OperatorStatus{}, "", nil
```
s-urbaniak, line 49: is there any reason we return empty structs vs nil?

stlaz: it's safer as the error is `nil` and so the controllers might assume (rightfully) that the returned structures exist

slaskawi (approved):

stlaz:
/assign @deads2k 
for approval

sttts:
It's not clear to me why we need this.

pkg/operator/v1helpers/helpers.go
```diff
@@ -21,6 +22,41 @@ import (
...
+}
+
+func (c *NoOpOperatorClient) UpdateOperatorSpec(_ string, _ *operatorv1.OperatorSpec) (*operatorv1.OperatorSpec, string, error) {
+	return &operatorv1.OperatorSpec{}, "", nil
+}
+
+func (c *NoOpOperatorClient) UpdateOperatorStatus(_ string, _ *operatorv1.OperatorStatus) (*operatorv1.OperatorStatus, error) {
```
deads2k, line 56: I'd like to see this method print the attempted state change at info level if conditions are "good" and at warning level with events if conditions are "bad".

The control loops that accept this as an interface expect to report status via this function today.

pkg/operator/v1helpers/helpers.go
```diff
@@ -21,6 +22,41 @@ import (
...
+
+func (c *NoOpOperatorClient) Informer() cache.SharedIndexInformer {
+	return nil
+}
+
+func (c *NoOpOperatorClient) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
+	return &operatorv1.OperatorSpec{}, &operatorv1.OperatorStatus{}, "", nil
+}
+
+func (c *NoOpOperatorClient) UpdateOperatorSpec(_ string, _ *operatorv1.OperatorSpec) (*operatorv1.OperatorSpec, string, error) {
```
deads2k, line 52: if it only accepts status updates, it's ok, but this method should fail imo

sttts:
/hold

I don't think a noop client without persistence implements the interface of an operator client.

slaskawi (approved):
/lgtm

pkg/operator/v1helpers/helpers.go
```diff
@@ -14,13 +14,44 @@ import (
...
+}
+
+func (c *LoggingStatusUpdater) GetOperatorState() (*operatorv1.OperatorSpec, *operatorv1.OperatorStatus, string, error) {
+	return &operatorv1.OperatorSpec{}, &operatorv1.OperatorStatus{
+		Conditions: c.cachedConditions,
+	}, "", nil
+}
+
+func (c *LoggingStatusUpdater) UpdateOperatorStatus(_ string, status *operatorv1.OperatorStatus) (*operatorv1.OperatorStatus, error) {
+	klog.V(2).Info("received a request to update status, diff: %s", diff.ObjectDiff(status, &operatorv1.OperatorStatus{Conditions: c.cachedConditions}))
```
s-urbaniak, line 47: .oO(i wish we could inject the logger at initialization time) ... actually ... does that make sense, potentially just a logging function?

pkg/operator/v1helpers/helpers.go
```diff
@@ -14,13 +14,44 @@ import (
...
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 )
 
+var _ OperatorStatusUpdater = &LoggingStatusUpdater{}
+
+type LoggingStatusUpdater struct {
+	// cachedConditions store the last observed conditions to be returned in the
+	// GetOperatorState method in order to prevent excessive conditions update logging
+	cachedConditions []operatorv1.OperatorCondition
```
s-urbaniak, line 31: we don't have to provide any synchronization for reads/writes? Here, `GetOperatorState` and `UpdateOperatorStatus` must not be used concurrently. We should at least make this clear in the constructor.

stlaz: you're right, we should

pkg/operator/v1helpers/helpers.go
stlaz: `klog` is the logger all our operators use, I believe, and is heavily used in other controllers. I'd maybe do that in a follow-up if such a need arises.

PR #1106, opened 2021-06-10
Bug 1969631: encryption precondition checker must indicate work to do when an operator is degraded

p0lyn0mial:
The precondition checker will run the encryption controllers when the caches haven't been synchronized.
During synchronization, the controllers send live requests to the API server.
A **single** failed request can put a controller into a degraded state and leave it in that state forever. 

Thus the precondition must take into account the status of the operator and let the controller run when it is in a degraded state.

For example, in [this](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-e2e-aws-upgrade/1400652395809083392/artifacts/e2e-aws-upgrade/) CI run the `kas-o` operator acquired a lock at `03:33:24`:

at `03:33:40` the `EncryptionPruneControllerDegraded` went degraded and **never** recovered.

around the same time, a few other controllers (`ResourceSyncControllerDegraded`, `RevisionControllerDegraded`, `TargetConfigControllerDegraded`) went degraded due to the same reason (`etcdserver: request timed out`) but recovered a few seconds later, around `03:33:55`

### DIFF
--- a/pkg/operator/apiserver/controllerset/apiservercontrollerset.go
+++ b/pkg/operator/apiserver/controllerset/apiservercontrollerset.go
@@ -312,12 +312,14 @@ func (cs *APIServerControllerSet) WithoutPruneController() *APIServerControllerS
 
 func (cs *APIServerControllerSet) WithEncryptionControllers(
 	component string,
+	componentClusterOperatorName string,
 	provider controllers.Provider,
 	deployer statemachine.Deployer,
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
/close

closing in favour of https://github.com/openshift/library-go/pull/1108

PR #1107, opened 2021-06-10
factory: allow to specify preconditions

mfojtik:
Preconditions will block the controller run similarly to cache sync. The timeout will bound how long the controller will wait for preconditions to succeed.

This eliminates hacks we do in encryption controllers where we pass precondition function directly to sync() and block.
In general, this would be helpful to other controllers, which have external dependencies on other operators or operator states. For example authentication operator use this extensively.

### DIFF
--- a/pkg/controller/factory/base_controller.go
+++ b/pkg/controller/factory/base_controller.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/robfig/cron"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -37,6 +38,8 @@ type baseController struct {
 	resyncSchedules    []cron.Schedule
 	postStartHooks     []PostStartHook
 	cacheSyncTimeout   time.Duration
+	preconditions      []PreconditionFn
+	exitOnError        func(args ...interface{})
 }
 
 var _ Controller = &baseController{}
@@ -62,6 +65,39 @@ func (s *scheduledJob) Run() {
 	s.queue.Add(DefaultQueueKey)
 }
 
+func waitForPreconditions(ctx context.Context, controllerName string, preconditions ...PreconditionFn) error {
+	if len(preconditions) == 0 {
+		return nil
+	}
+
+	var (
+		wg       sync.WaitGroup
+		errs     []error
+		errsLock sync.Mutex
+	)
+
+	wg.Add(len(preconditions))
+	for _, preconditionFunc := range preconditions {
+		go func(preconditionCtx context.Context, fn PreconditionFn) {
+			defer wg.Done()
+			err := wait.PollImmediateUntil(100*time.Millisecond, func() (bool, error) {
+				return fn(preconditionCtx)
+			}, preconditionCtx.Done())
+			if err != nil {
+				errsLock.Lock()
+				errs = append(errs, err)
+				errsLock.Unlock()
+				return
+			}
+		}(ctx, preconditionFunc)
+	}
+
+	klog.V(2).Infof("Controller %q waiting for %d precondition ...", controllerName, len(preconditions))
+	wg.Wait()
+
+	return kerrors.NewAggregate(errs)
+}
+
 func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheSyncs ...cache.InformerSynced) error {
 	klog.Infof("Waiting for caches to sync for %s", controllerName)
 
@@ -81,6 +117,7 @@ func (c *baseController) Run(ctx context.Context, workers int) {
 	// give caches 10 minutes to sync
 	cacheSyncCtx, cacheSyncCancel := context.WithTimeout(ctx, c.cacheSyncTimeout)
 	defer cacheSyncCancel()
+	cacheSyncStartTime := time.Now()
 	err := waitForNamedCacheSync(c.name, cacheSyncCtx.Done(), c.cachesToSync...)
 	if err != nil {
 		select {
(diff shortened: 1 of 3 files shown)


pkg/controller/factory/base_controller.go
```diff
@@ -74,6 +103,10 @@ func waitForNamedCacheSync(controllerName string, stopCh <-chan struct{}, cacheS
 	return nil
 }
 
+func exitOnError(err error) {
```
sttts, line 106: what's the sense of this func? https://wiki.haskell.org/Eta_conversion

pkg/controller/factory/base_controller.go
```diff
@@ -92,7 +125,24 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
 			// TODO: Consider making this behavior configurable.
-			klog.Exit(err)
+			c.exitOnError(err)
+		}
+	}
+
+	preconditionCtx, preconditionCtxCancel := context.WithTimeout(ctx, c.preconditionsTimeout)
+	defer preconditionCtxCancel()
+	err = waitForPreconditions(c.name, preconditionCtx.Done(), c.preconditions...)
+	if err != nil {
```
sttts, line 135: nit: `if err := ...; err != nil`

pkg/controller/factory/base_controller_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"k8s.io/klog/v2"
```
sttts, line 12: order

pkg/controller/factory/base_controller_test.go
```diff
@@ -41,6 +43,9 @@ func TestBaseController_ExitOneIfCachesWontSync(t *testing.T) {
 	c := &baseController{
 		syncContext:      NewSyncContext("test", eventstesting.NewTestingEventRecorder(t)),
 		cacheSyncTimeout: 1 * time.Second,
+		exitOnError: func(err error) {
+			klog.Exit(err)
+		},
```
sttts, line 48: unnecessary eta expansion

pkg/controller/factory/factory.go
```diff
@@ -83,6 +89,14 @@ func (f *Factory) WithSync(syncFn SyncFunc) *Factory {
 	return f
 }
 
+// WithPreconditions will take a list of precondition functions that have to all succeed in order to start the controller
+// successfully. If the timeout is reached, the controller startup will be terminated.
+func (f *Factory) WithPreconditions(timeout time.Duration, preconditions ...PreconditionFn) *Factory {
+	f.preconditionsTimeout = timeout
+	f.preconditions = preconditions
```
sttts, line 96: append?

pkg/controller/factory/base_controller.go
```diff
@@ -62,6 +65,32 @@ func (s *scheduledJob) Run() {
 	s.queue.Add(DefaultQueueKey)
 }
 
+func waitForPreconditions(controllerName string, stopCh <-chan struct{}, preconditions ...PreconditionFn) error {
+	if len(preconditions) == 0 {
+		return nil
+	}
+	err := wait.PollImmediateUntil(100*time.Millisecond,
```
sttts, line 72: worth to externalize the interval too? I can imagine that something here is expensive.

pkg/controller/factory/base_controller.go
mfojtik: this is for unit testing so I can replace this with something that check this func was called :-)

mfojtik: got rid of eta :-)

pkg/controller/factory/base_controller.go
mfojtik: I reworked this a little.

pkg/controller/factory/base_controller.go
```diff
@@ -37,6 +38,8 @@ type baseController struct {
 	resyncSchedules    []cron.Schedule
 	postStartHooks     []PostStartHook
 	cacheSyncTimeout   time.Duration
+	preconditions      []PreconditionFn
+	exitOnError        func(args ...interface{})
```
sttts, line 42: what is this? Why so untyped?

pkg/controller/factory/base_controller.go
```diff
@@ -62,6 +65,39 @@ func (s *scheduledJob) Run() {
...
+	}
+
+	var (
+		wg       sync.WaitGroup
+		errs     []error
+		errsLock sync.Mutex
+	)
+
+	wg.Add(len(preconditions))
+	for _, preconditionFunc := range preconditions {
```
sttts, line 80: fn

pkg/controller/factory/base_controller.go
```diff
@@ -62,6 +65,39 @@ func (s *scheduledJob) Run() {
...
+		go func(preconditionCtx context.Context, fn PreconditionFn) {
+			defer wg.Done()
+			err := wait.PollImmediateUntil(100*time.Millisecond, func() (bool, error) {
```
sttts, line 83: I still don't like this. If I put a client call in here, this will hotloop (modulo throttling).

pkg/controller/factory/base_controller.go
```diff
@@ -62,6 +65,39 @@ func (s *scheduledJob) Run() {
...
+
+	var (
+		wg       sync.WaitGroup
+		errs     []error
+		errsLock sync.Mutex
+	)
+
+	wg.Add(len(preconditions))
+	for _, preconditionFunc := range preconditions {
+		go func(preconditionCtx context.Context, fn PreconditionFn) {
```
sttts, line 81: ctx

pkg/controller/factory/base_controller.go
```diff
@@ -91,8 +128,25 @@ func (c *baseController) Run(ctx context.Context, workers int) {
...
-			// TODO: Consider making this behavior configurable.
-			klog.Exit(err)
+			c.exitOnError(err)
+		}
+	}
+
+	// for preconditions, use the cache sync timeout minus the time we were waiting for caches for sync.
+	preconditionTimeout := c.cacheSyncTimeout - time.Now().Sub(cacheSyncStartTime)
+
+	preconditionCtx, preconditionCtxCancel := context.WithTimeout(ctx, preconditionTimeout)
```
sttts, line 138: you like long var names

pkg/controller/factory/base_controller.go
mfojtik: compatible with klog.Exit

pkg/controller/factory/base_controller.go
mfojtik: the 100ms should prevent hotloop, no? also client will throttle as well...

pkg/controller/factory/base_controller.go
mfojtik: no, it can collide with upper level `ctx`

pkg/controller/factory/base_controller.go
mfojtik: yeah, they are cheap and describe what is this context for :)

PR #1108, opened 2021-06-10
Bug 1969631: encryption controllers wait for all informers before running the sync loops

p0lyn0mial:
Before this change, the precondition checker could run the encryption controllers when the caches haven't been synchronized.
During synchronization, the controllers send live requests to the API server.
A **single** failed request can put a controller into a degraded state and leave it in that state forever. 

Thus the precondition function MUST be called after the caches were synchronized. This PR ensures that.

For example, in [this](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-e2e-aws-upgrade/1400652395809083392/artifacts/e2e-aws-upgrade/) CI run the `kas-o` operator acquired a lock at `03:33:24`:

at `03:33:40` the `EncryptionPruneControllerDegraded` went degraded and **never** recovered.

around the same time, a few other controllers (`ResourceSyncControllerDegraded`, `RevisionControllerDegraded`, `TargetConfigControllerDegraded`) went degraded due to the same reason (`etcdserver: request timed out`) but recovered a few seconds later, around `03:33:55`


pkg/operator/encryption/controllers/condition_controller.go
```diff
@@ -56,6 +58,7 @@ func NewConditionController(
 	return factory.New().WithInformers(
 		kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
 		operatorClient.Informer(),
+		apiServerConfigInformer.Informer(),
```
sttts, line 61: where is that used? (next person will factor this out because apparently it is not referenced anywhere).

p0lyn0mial: do you want me to add comment?

p0lyn0mial:
/hold cancel

added the comment

pkg/operator/encryption/controllers/condition_controller.go
p0lyn0mial: done

pkg/operator/encryption/preconditions.go
```diff
@@ -3,17 +3,13 @@ package encryption
...
-	"k8s.io/klog/v2"
-
-	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
 	"github.com/openshift/library-go/pkg/operator/encryption/state"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
stlaz, line 12: nit: wrong order

stlaz:
/hold cancel
had some questions, those were answered in Slack

PR #1109, opened 2021-06-14
WIP - Implementing CredentialStoreFactory

ricardomaraschini:


ricardomaraschini:
/assign @dmage 

Could you please take an initial look into this one?

pkg/image/registryclient/client.go
```diff
@@ -373,17 +385,22 @@ func (c *Context) cachedTransport(rt http.RoundTripper, host string, scopes []au
 		scopes = append(scopes, stringScope(s))
 	}
 
+	creds := c.Credentials
+	if c.CredentialsFactory != nil {
+		creds = c.CredentialsFactory.CredentialStoreFor(ref.AsRepository().String())
+	}
```
dmage, line 391: you need to rework cachedTransports so that it respects new credentials

ricardomaraschini: I don't think I need to refactor it. `Scopes` (repository and operation) and the registry `host` are taken into account when evaluating if we are going to reuse or not a transport. In any case I have developed an unit test where I assure we are trying the right authentications even when mirror from one repository into another inside the same registry. Is there any corner case you see this failing? ptal.

PR #1111, opened 2021-06-17
makes the apiservice controller more resilient to failures

p0lyn0mial:
previously a single failed request to the discovery endpoint of a service could make an operator report Availiable=false

to increase resiliency and avoid frequent changes in Availability we are going to send 5 requests in parallel and deemed a service as unavailable only when we didn't get at least 2 successful.

### DIFF
--- a/pkg/operator/apiserver/controller/apiservice/apigroup.go
+++ b/pkg/operator/apiserver/controller/apiservice/apigroup.go
@@ -4,11 +4,16 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"sync"
+	"time"
 
-	"github.com/openshift/library-go/pkg/operator/events"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
proof PR https://github.com/openshift/library-go/pull/1111

pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -65,16 +70,60 @@ func newEndpointPrecondition(kubeInformers kubeinformers.SharedInformerFactory)
...
+	var errs = []error{}
+	for resTuple := range resultsCh {
+		if resTuple.status != http.StatusOK {
+			errs = append(errs, fmt.Errorf("an attempt failed with statusCode = %v, err = %v", resTuple.status, resTuple.err))
+			continue
+		}
+		successfulRequests++
+	}
+
+	if successfulRequests < 2 {
```
sttts, line 125: this needs a comment. Why 2 and not 3?

p0lyn0mial: done

PR #1112, opened 2021-06-17
fixes a makefile target to run an integration test for checking encryption

p0lyn0mial:
it looks like the test was never exercised in CI

### DIFF
--- a/test/e2e-encryption/encryption_test.go
+++ b/test/e2e-encryption/encryption_test.go
@@ -121,8 +121,9 @@ func TestEncryptionIntegration(tt *testing.T) {
 		{Group: "operator.openshift.io", Resource: "kubeschedulers"},
 	})
 
-	controllers := encryption.NewControllers(
+	controllers, err := encryption.NewControllers(
 		component,
+		[]string{},
 		provider,
 		deployer,
 		migrator,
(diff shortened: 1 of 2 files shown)


ravisantoshgudimetla (commented):
I noticed this error when I loaded the project in GoLand too and was about to provide a fix. Thank you for fixing it @p0lyn0mial 

/lgtm

/hold 

for auth folks to give final approval

test/e2e-encryption/encryption_test.go
```diff
@@ -360,7 +374,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 	deployer.DeleteOperandConfig()
 	waitForConfigs(
 		// 7 is migrated and hence only one needed, but we rotate through identity
-		"kubeapiservers.operator.openshift.io=identity,aescbc:7;kubeschedulers.operator.openshift.io=identity,aescbc:7",
+		"kubeapiservers.operator.openshift.io=identity,aescbc:7,aescbc:5;kubeschedulers.operator.openshift.io=identity,aescbc:7,aescbc:5",
```
p0lyn0mial, line 377: since `aescbc:5` wasn't deleted it should be on the list, right?

PR #1114, opened 2021-06-18
apiserver/audit: split policy config map manifest into policy rule slices

sttts:
Preparation for `customRules` implementation.


EmilyM1:
Huge overhaul!
We now will now longer need the static files for anything?

PR #1116, opened 2021-06-21
WIP Enable extension of the staticpod installer pod

marun:
This is in support of enabling rollback to the previous revision if the new revision proves unhealthy.


marun:
Not intended for merge, just posting wip stuff in aid of the eventual enhancement.

/cc @sttts @p0lyn0mial

PR #1117, opened 2021-06-22
DeploymentController: Do not overwrite manifest

jsafrane:
When applying hooks, make sure not to overwrite the original manifest. Some hooks may change the content later.

Example: a hook that replaces `${LOG_LEVEL}` with `operator.spec.logLevel`. As user modifies the `logLevel`, the controller + its hooks should replace `${LOG_LEVEL}` with the *current* value, not with the value that was *available on the first sync*.


soltysh (approved):
/lgtm

ravisantoshgudimetla (approved):
/lgtm

PR #1120, opened 2021-06-23
OTA-260: add duplicate manifest error check

jottofar:
Ensure each manifest contains the Kubernetes required fields apiVersion, kind, and metadata.name. Use those fields along with the optional field metadata.namespace to uniquely identify each manifest. Generate an error for any duplicate manifests, i.e. any with the same unique id.

Version is removed from apiVersion to produce Group which is what is actually used to form the above mentioned unique id. Some resources, e.g. ConfigMap, have an apiVersion with only a version identifier, e.g. `v1`. In these cases the Group is set to `""` as opposed to `v1`.

See [OTA-260](https://issues.redhat.com/browse/OTA-260).

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -15,23 +15,50 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+// resourceId uniquely identifies a Kubernetes resource.
+// It is used to identify any duplicate resources within
+// a given set of manifests.
+type resourceId struct {
+	// Group identifies a set of API resources exposed together.
+	// +optional
+	Group string
+	// Kind is the name of a particular object schema.
+	Kind string
+	// Name, sometimes used with the optional Namespace, helps uniquely identify an object.
+	Name string
+	// Namespace helps uniquely identify an object.
+	// +optional
+	Namespace string
+}
+
 // Manifest stores Kubernetes object in Raw from a file.
-// It stores the GroupVersionKind for the manifest.
-// Raw and Obj should always be kept in sync such that
-// each provides the same data but in different formats.
-// To ensure Raw and Obj are always in sync, they should not
-// be set directly but rather only be set by calling
-// either method ManifestsFromFiles or ParseManifests.
+// It stores the id and the GroupVersionKind for
+// the manifest. Raw and Obj should always be kept in sync
+// such that each provides the same data but in different
+// formats. To ensure Raw and Obj are always in sync, they
+// should not be set directly but rather only be set by
+// calling either method ManifestsFromFiles or
+// ParseManifests.
 type Manifest struct {
 	// OriginalFilename is set to the filename this manifest was loaded from.
-	// It is not guaranteed to be set or be unique, but we will set it when
-	// loading from disk to provide better debuggability.
+	// It is not guaranteed to be set or be unique, but will be set when
+	// loading from disk to provide a better debug capability.
 	OriginalFilename string
+
+	id resourceId
 
 	Raw []byte
 	GVK schema.GroupVersionKind
 
 	Obj *unstructured.Unstructured
+}
+
+func (r resourceId) String() string {
+	if len(r.Namespace) == 0 {
+		return fmt.Sprintf("Group: %q Kind: %q Name: %q", r.Group, r.Kind, r.Name)
+	} else {
+		return fmt.Sprintf("Group: %q Kind: %q Namespace: %q Name: %q", r.Group, r.Kind, r.Namespace, r.Name)
+	}
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface for the Manifest
@@ -60,16 +87,25 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	if !ok {
 		return fmt.Errorf("expected manifest to decode into *unstructured.Unstructured, got %T", ud)
 	}
-
 	m.GVK = ud.GroupVersionKind()
 	m.Obj = ud
-	return nil
+	m.id = resourceId{
+		Group:     m.GVK.Group,
+		Kind:      m.GVK.Kind,
+		Namespace: m.Obj.GetNamespace(),
+		Name:      m.Obj.GetName(),
+	}
+	return validateResourceId(m.id)
 }
 
 // ManifestsFromFiles reads files and returns Manifests in the same order.
-// files should be list of absolute paths for the manifests on disk.
+// 'files' should be list of absolute paths for the manifests on disk. An
+// error is returned for each manifest that defines a duplicate resource
+// as compared to other manifests defined within the 'files' list.
+// Duplicate resources have the same group, kind, name, and namespace.
 func ManifestsFromFiles(files []string) ([]Manifest, error) {
 	var manifests []Manifest
+	ids := make(map[resourceId]bool)
 	var errs []error
 	for _, file := range files {
 		file, err := os.Open(file)
@@ -86,6 +122,10 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
 		}
 		for _, m := range ms {
 			m.OriginalFilename = filepath.Base(file.Name())
+			err = addIfNotDuplicateResource(m, ids)
+			if err != nil {
+				errs = append(errs, errors.Wrapf(err, "File %s contains", file.Name()))
+			}
 		}
 		manifests = append(manifests, ms...)
 	}
@@ -99,8 +139,10 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
 }
 
 // ParseManifests parses a YAML or JSON document that may contain one or more
-// kubernetes resources.
+// kubernetes resources. An error is returned if the input cannot be parsed
+// or contains a duplicate resource.
 func ParseManifests(r io.Reader) ([]Manifest, error) {
+	theseIds := make(map[resourceId]bool)
 	d := yaml.NewYAMLOrJSONDecoder(r, 1024)
 	var manifests []Manifest
 	for {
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -3,9 +3,11 @@ package manifest
 import (
 	"bytes"
 	"fmt"
+	goyaml "gopkg.in/yaml.v2"
```
wking, line 6: nit: non-stdlib package, so ideally import goes down in the non-stdlib block below.

pkg/manifest/manifest.go
```diff
@@ -15,25 +17,60 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+// ResourceIdYaml is used to parse out the information
+// that uniquely identifies a Kubernetes resource.
+type ResourceIdYaml struct {
```
wking, line 22: Does this need to be a public type?

pkg/manifest/manifest.go
```diff
@@ -15,25 +17,60 @@ import (
...
+}
+
+// ResourceId uniquely identifies a Kubernetes resource.
+// It is used to identify any duplicate resources within
+// a given set of manifests.
+type ResourceId struct {
+	Group     string
+	Kind      string
+	Name      string
+	Namespace string
```
wking, line 38: nit: copy a sentence or two of godocs for these properties from [`ObjectReference`][1] or similar?

[1]: https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference

pkg/manifest/manifest.go
```diff
@@ -15,25 +17,60 @@ import (
...
 
+func (r ResourceId) fullName() string {
+	if len(r.Namespace) == 0 {
+		return r.Name
+	}
+	return r.Namespace + "/" + r.Name
+}
+
+func (r ResourceId) String() string {
+	return fmt.Sprintf("Group: \"%s\" Kind: \"%s\" Name: \"%s\"", r.Group, r.Kind, r.fullName())
```
wking, line 71: nit: Instead of the `fullName` collapse, I'd rather use:

```go
func (r ResourceId) String() string {
  if len(r.Namespace) == 0 {
    return fmt.Sprintf("Group: %q Kind: %q Name: %q", r.Group, r.Kind, r.Name)
  } else {
    return fmt.Sprintf("Group: %q Kind: %q Namespace: %q Name: %q", r.Group, r.Kind, r.Namespace, r.Name)
  }
}
```

or some such, so readers don't have to mentally split on the slash.  And while I prefer `%q` if we want quotes, I'm not actually sure we need quotes on these, because they're already covered by domain-name restrictions.  So I'm fine with `%q`, and I'm fine with `%s`, but I'd rather avoid `"%s"`.

pkg/manifest/manifest.go
```diff
@@ -99,22 +148,77 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
...
 
 // ParseManifests parses a YAML or JSON document that may contain one or more
-// kubernetes resources.
-func ParseManifests(r io.Reader) ([]Manifest, error) {
+// kubernetes resources. Parsing and structure related errors are returned in
+// the first error parameter and cause an immediate return. Duplicate resource
+// errors are returned in the error list. Processing continues after a
+// duplicate resource error allowing all duplicates to be identified in the
+// given input.
+func ParseManifests(r io.Reader) ([]Manifest, error, []error) {
```
wking, line 156: I'm not sure we want the separate return value just for duplicates.  Failing fast if/when we detect duplicates might miss some entries if we are lousy with dups, but ideally the only dups are those in the in-flight PR being tested.

pkg/manifest/manifest.go
```diff
@@ -99,22 +148,77 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
...
 }
+
+func unmarshalResourceId(manifestBytes []byte) (ResourceId, error) {
+	var resourceId ResourceId
+	var yamlId ResourceIdYaml
+	if err := goyaml.Unmarshal(manifestBytes, &yamlId); err != nil {
+		return resourceId, errors.Wrapf(err, "unable to unmarshall resource id fields")
+	}
+	if !validReourceId(yamlId) {
+		return resourceId, errors.New("Kubernetes objects must contain required fields apiVersion, kind, and metadata.name")
```
wking, line 192: It's probably more sustainable and easier to debug to use `validateResourceId(id ResourceIdYaml) error` that complains directly about the missing property.  Also, might be worth linking [these docs][1], either in the error message or a Go comment, to back up the requirement.

[1]: https://github.com/kubernetes/community/blob/2f26b93380f504e4c11effdae1eee7bbe6648d5b/contributors/devel/sig-architecture/api-conventions.md#resources

pkg/manifest/manifest.go
```diff
@@ -99,22 +148,77 @@ func ManifestsFromFiles(files []string) ([]Manifest, error) {
...
+				return manifests, nil, dupErrors
 			}
-			return manifests, errors.Wrapf(err, "error parsing")
+			return manifests, errors.Wrapf(err, "error parsing"), dupErrors
 		}
 		m.Raw = bytes.TrimSpace(m.Raw)
 		if len(m.Raw) == 0 || bytes.Equal(m.Raw, []byte("null")) {
 			continue
 		}
+		if resourceId, err := unmarshalResourceId(m.Raw); err != nil {
```
wking, line 173: We probably don't have to unmarshal again.   Doesn't the `Decode` a few lines up call `*Manifest.UnmarshalJSON(...)` indirectly?

pkg/manifest/manifest.go
```diff
@@ -60,16 +97,24 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	if !ok {
 		return fmt.Errorf("expected manifest to decode into *unstructured.Unstructured, got %T", ud)
 	}
-
+	if resourceId, err := unmarshalResourceId(m.Raw); err != nil {
```
wking, line 100: I feel like we can skip the YAML unmarshal with something like:

```go
m.GVK = ud.GroupVersionKind()
m.Obj = ud
m.resourceId = resourceId{
  Group: m.GVK.Group,
  Kind: m.GVK.Kind,
  Namespace: m.Obj.GetNamespace(),
  Name: m.Obj.GetName(),
}
return validateResourceId(m.resourceId)
```

pkg/manifest/manifest_test.go
```diff
@@ -155,6 +161,100 @@ data:
 
 }
 
+func TestParseManifestsDuplicates(t *testing.T) {
+	tests := []struct {
+		name             string
+		raw              string
+		wantNumDupErrors int
+		wantErr          bool
```
wking, line 169: a boolean leaves a lot of room for "actual error message doesn't look like the one I expected when writing the tests".  I prefer an `expectedError string` with handling that is either a direct string match ([example][1]) or a regexp match ([example][2]).

[1]: https://github.com/openshift/cluster-version-operator/blob/9c0b0b446ed201999c7cbae8b3d6ade24f531353/pkg/cvo/cvo_test.go#L3359-L3367
[2]: https://github.com/openshift/installer/blob/c9171cfa4017771d696bbad8b70e6c2edaa244f4/pkg/types/validation/installconfig_test.go#L1274-L1277

pkg/manifest/manifest.go
jottofar: How about this?

pkg/manifest/manifest.go
jottofar: Done

pkg/manifest/manifest.go
jottofar: Yeah, we're not losing much by not doing it and it makes it much easier to understand. Done.

pkg/manifest/manifest.go
jottofar: I added a method comment with a kube link that I thought explained it well. And since Group could be empty I left it off. Also turns out that the existing unmarshall captures `kind` related issues but not an empty `name`.

pkg/manifest/manifest.go
jottofar: Duh. It does and I had forgotten it worked that way.

pkg/manifest/manifest.go
jottofar: Yes

pkg/manifest/manifest_test.go
jottofar: This is changed and tightened up.

pkg/manifest/manifest.go
```diff
@@ -115,6 +158,34 @@ func ParseManifests(r io.Reader) ([]Manifest, error) {
...
+// validateResourceId ensures the id contains the required fields per
+// https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields.
+func validateResourceId(id ResourceId) error {
+	if id.Kind == "" || id.Name == "" {
+		return fmt.Errorf("Resource with fields %s must contain kubernetes required fields kind and name", id)
+	}
+	return nil
+}
+
+func truncateVersion(apiVersion string) string {
```
wking, line 177: You can drop this, now that you've removed its consumers.

pkg/manifest/manifest_test.go
```diff
@@ -155,6 +161,76 @@ data:
...
+  name: a-config
+  namespace: default
+---
+apiVersion: extensions/v2
+kind: ConfigMap
+metadata:
+  name: a-config
+  namespace: default
+`,
+		wantErr: "duplicate resource:",
```
wking, line 210: should we grow this to ensure the explanation of the duplicate includes the description we want?  Looks like it will be `duplicate resource: (Group: "extensions" Kind: "ConfigMap" Namespace: "default" Name: "a-config")`

pkg/manifest/manifest_test.go
```diff
@@ -155,6 +161,76 @@ data:
...
+	}}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			_, err := ParseManifests(strings.NewReader(test.raw))
+			if err == nil {
+				if len(test.wantErr) != 0 {
+					t.Fatalf("Expected an error and got none")
+				}
+			} else {
+				if len(test.wantErr) == 0 {
```
wking, line 220: nit: you can drop one level of nesting with:

```go
if err == nil {
  ...your existing no-error handling...
} else if len(test.wantErr) == 0 {
  t.Fatalf("Got unexpected error: %v", err)
} else if !strings.Contains(err.Error(), test.wantErr) {
  t.Fatalf("Got incorrect error. Wanted it to contain %s but got %s", test.wantErr, err.Error())
}
```

pkg/manifest/manifest.go
jottofar: yep

pkg/manifest/manifest_test.go
jottofar: done

pkg/manifest/manifest_test.go
jottofar: agree

wking (approved):
/lgtm

PR #1123, opened 2021-06-30
apiserver: implement audit.customRules

sttts:
This implements the policy algorithm described in https://github.com/openshift/enhancements/pull/792.

- [x] unit tests


pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -83,11 +83,26 @@ func GetAuditPolicy(audit configv1.Audit) (*auditv1.Policy, error) {
 	p := basePolicy.DeepCopy()
 	p.Name = string(audit.Profile)
 
-	extraRules, ok := profileRules[audit.Profile]
+	for _, cr := range audit.CustomRules {
+		profileRules, ok := profileRules[audit.Profile]
```
p0lyn0mial, line 87: shouldn't you get the profile from the custom rule?

like:
`profileRules, ok := profileRules[cr.Profile]`

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -83,11 +83,26 @@ func GetAuditPolicy(audit configv1.Audit) (*auditv1.Policy, error) {
...
+		if !ok {
+			return nil, fmt.Errorf("unknown audit profile %q", audit.Profile)
```
p0lyn0mial, line 89: could you change the error msg to indicate that a profile wasn't found in a custom rule?

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -83,11 +83,26 @@ func GetAuditPolicy(audit configv1.Audit) (*auditv1.Policy, error) {
...
+		}
+
+		groupRules := make([]auditv1.PolicyRule, len(profileRules))
+		for i, r := range profileRules {
+			r.DeepCopyInto(&groupRules[i])
+			groupRules[i].UserGroups = []string{cr.Group}
```
p0lyn0mial, line 95: what if in the future we decide to have a rule that already sets `UserGroups` ?
we could append instead of overwriting

p0lyn0mial: ah, there's a unit test for that `TestNoUserGroups`

pkg/operator/apiserver/audit/audit_policies.go
p0lyn0mial: how this error will be displayed to an end-user?

pkg/operator/apiserver/audit/audit_policies.go
sttts: yep

pkg/operator/apiserver/audit/audit_policies.go
sttts: This error is never shown to the user. I.e. it should not happen if we get the API validation right.

p0lyn0mial: will we have a validation webhook for the audit policies?

pkg/operator/apiserver/audit/audit_policies.go
sttts: See my rant about missing conjunction in Slack channel. These profiles for group can never have pre-defined groups. We made a mistake in API design long ago in upstream.

pkg/operator/apiserver/audit/audit_policies.go
```diff
@@ -83,11 +83,30 @@ func GetAuditPolicy(audit configv1.Audit) (*auditv1.Policy, error) {
...
-	extraRules, ok := profileRules[audit.Profile]
+	for _, cr := range audit.CustomRules {
+		rules, ok := profileRules[cr.Profile]
+		if !ok {
+			return nil, fmt.Errorf("unknown audit profile %q in customRules for group %q", cr.Profile, cr.Group)
+		}
+
+		groupRules := make([]auditv1.PolicyRule, len(rules))
+		for i, r := range rules {
+			r.DeepCopyInto(&groupRules[i])
```
p0lyn0mial, line 94: why do we need a deep copy?

pkg/operator/apiserver/audit/audit_policies.go
sttts: done via openapi.

pkg/operator/apiserver/audit/audit_policies.go
sttts: we modify it. Technically we only change the shallow copy. But I don't like to allow any modification like that without prior deep copy, at least in non-performance-critical code paths.

pkg/operator/apiserver/audit/audit_policies_test.go
```diff
@@ -27,6 +36,82 @@ func TestDefaultPolicy(t *testing.T) {
...
+	for _, scenario := range scenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			// act
+			policy, err := GetAuditPolicy(scenario.config)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// validate
+			if len(scenario.goldenFile) > 0 && len(scenario.goldenFile) > 0 {
```
p0lyn0mial, line 97: `if len(scenario.goldenFile) > 0 {` should be enough :)

sttts: better twice to be sure

sttts: fixed

pkg/operator/apiserver/audit/audit_policies.go
p0lyn0mial: agreed.

PR #1124, opened 2021-07-02
Bug 1978193: fix CSR requester controller possible infinite crashloop and expose cert validation errors

stlaz:
/assign @deads2k

### DIFF
--- a/pkg/operator/csr/cert_controller.go
+++ b/pkg/operator/csr/cert_controller.go
@@ -102,7 +102,11 @@ func NewClientCertificateController(
 	spokeCoreClient corev1client.CoreV1Interface,
 	recorder events.Recorder,
 	controllerName string,
-) factory.Controller {
+) (factory.Controller, error) {
+	if len(csrOption.ObjectMeta.Name) > 0 {
+		return nil, fmt.Errorf("the CSR controller does not allow specifying static names for the CSRs")
+	}
+
 	c := clientCertificateController{
 		ClientCertOption: clientCertOption,
 		CSROption:        csrOption,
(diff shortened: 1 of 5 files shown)


pkg/operator/csr/certificate.go
```diff
@@ -11,45 +11,42 @@ import (
...
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 	certutil "k8s.io/client-go/util/cert"
-	"k8s.io/klog/v2"
 )
 
 // IsCertificateValid return true if
 // 1) All certs in client certificate are not expired.
 // 2) At least one cert matches the given subject if specified
-func IsCertificateValid(certData []byte, subject *pkix.Name) (bool, error) {
+func IsCertificateValid(certData []byte, subject *pkix.Name) error {
```
deads2k, line 19: this change seems ok.

pkg/operator/csr/cert_controller.go
```diff
@@ -284,6 +284,13 @@ func (c *clientCertificateController) createCSR(ctx context.Context) (string, er
 		},
 	}
 
+	// prevent CSR creation failure on static CSR names
+	if csrStaticName := c.ObjectMeta.Name; len(csrStaticName) > 0 {
```
deads2k, line 288: Instead of handling this, I think we should fail early if this is present.  Generated names are better for CSRs than fixed names.

1. CSRs are temporary. They are reaped some time after reaching a terminal state.
2. because they are temporary, it means a fixed name doesn't allow you use existence as a signal
3. fixed names mean that someone else could "steal" your name, leading conflicts.  Generated names do not.

stlaz: Indeed, I agree

slaskawi (approved):
/lgtm

praveenkumar:
ping @slaskawi / @deads2k can you also provide `approve` label also since it is still issue for CRC side?

stlaz:
/hold
we need to fix this in order to disallow static names for CSRs instead, single names would be an antipattern

PR #1125, opened 2021-07-02
clear encryption conditions when there is no work to be done

p0lyn0mial:
It turns out that the new precondition checker added in 4.8 might prevent the encryption controllers from running and clearing the stale conditions.
This can happen especially on upgrades from 4.7 to 4.8.

This PR changes the encryption controllers to clear stale conditions when the preconditions are not fulfilled.


pkg/operator/encryption/controllers/condition_controller.go
```diff
@@ -63,8 +63,21 @@ func NewConditionController(
 	).ResyncEvery(time.Minute).WithSync(c.sync).ToController("EncryptionConditionController", eventRecorder.WithComponentSuffix("encryption-condition-controller"))
 }
 
-func (c *conditionController) sync(ctx context.Context, syncContext factory.SyncContext) error {
+func (c *conditionController) sync(_ context.Context, _ factory.SyncContext) (err error) {
+	condition := &operatorv1.OperatorCondition{Type: "Encrypted", Status: operatorv1.ConditionFalse}
```
sttts, line 67: keep name `cond` and the patch gets much smaller.

p0lyn0mial: done

PR #1128, opened 2021-07-06
add possibility to modify ControllerBuilder in ControllerCommandConfig

atiratree:
It is hard to customize the `StartController` without a possibility to modify the builder. Otherwise it is necessary to copy paste a lot of code: https://github.com/openshift/cluster-policy-controller/pull/65/commits/a94cb90b7397b26d1fc72c2ffaa29d27d41a541e. 

This change would be helpful for https://github.com/openshift/cluster-policy-controller/pull/65/

### DIFF
--- a/pkg/controller/controllercmd/cmd.go
+++ b/pkg/controller/controllercmd/cmd.go
@@ -39,6 +39,8 @@ type ControllerCommandConfig struct {
 
 	basicFlags *ControllerFlags
 
+	controllerBuilderModifier ControllerBuilderModifier
+
 	// DisableServing disables serving metrics, debug and health checks and so on.
 	DisableServing bool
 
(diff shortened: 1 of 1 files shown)


pkg/controller/controllercmd/cmd.go
```diff
@@ -39,23 +39,42 @@ type ControllerCommandConfig struct {
...
+	ConfigContent      []byte
+}
+
+// ControllerBuilderModifier can be used to ControllerBuilder
+type ControllerBuilderModifier func(builder *ControllerBuilder, options ControllerBuilderOptions) (*ControllerBuilder, error)
+
 // NewControllerConfig returns a new ControllerCommandConfig which can be used to wire up all the boiler plate of a controller
 // TODO add more methods around wiring health checks and the like
-func NewControllerCommandConfig(componentName string, version version.Info, startFunc StartFunc) *ControllerCommandConfig {
+func NewControllerCommandConfig(componentName string, version version.Info, startFunc StartFunc, controllerBuilderModifier ControllerBuilderModifier) *ControllerCommandConfig {
```
atiratree, line 68: is it okay to change the function header?

pkg/controller/controllercmd/cmd.go
```diff
@@ -39,23 +39,42 @@ type ControllerCommandConfig struct {
...
+
 	// DisableServing disables serving metrics, debug and health checks and so on.
 	DisableServing bool
 
 	// DisableLeaderElection allows leader election to be suspended
 	DisableLeaderElection bool
 }
 
+// ControllerBuilderOptions contain data needed for ControllerBuilder modification
+type ControllerBuilderOptions struct {
```
soltysh, line 52: This seems a bit excessive, I'd imagine tweaking only the bits you require see https://github.com/openshift/cluster-policy-controller/commit/a94cb90b7397b26d1fc72c2ffaa29d27d41a541e#r53164748 where I'm pointing only the necessity to translate from one type of config to another, wdyt?

soltysh (changes requested):

atiratree:
after a discussion, the default config should be enough for our use case

PR #1129, opened 2021-07-12
Updates for Cert Manager

slaskawi:
https://issues.redhat.com/browse/AUTH-5

This Pull Request introduces CRD and Admission Registration types into the StaticResourceController

### DIFF
--- a/pkg/operator/resource/resourceapply/generic.go
+++ b/pkg/operator/resource/resourceapply/generic.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
@@ -23,6 +24,7 @@ import (
 	migrationclient "sigs.k8s.io/kube-storage-version-migrator/pkg/clients/clientset"
 
 	"github.com/openshift/api"
+
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourceapply/generic.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	admissionRegistrationv1 "k8s.io/api/admissionregistration/v1"
```
stlaz, line 7: nit: we usually keep these lower-cased

slaskawi: Fixed

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -195,6 +197,18 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
 			} else {
 				result.Result, result.Changed, result.Error = ApplyStorageClass(ctx, clients.kubeClient.StorageV1(), recorder, t)
 			}
+		case *admissionregistrationv1.ValidatingWebhookConfiguration:
+			if clients.kubeClient == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				result.Result, result.Changed, result.Error = ApplyValidatingWebhookConfiguration(ctx, clients.kubeClient.AdmissionregistrationV1(), recorder, t, -1)
```
bertinatto, line 204: @slaskawi this `-1` for `expectedGeneration` means that every time this function is called (i.e., during the static controller sync), the resource will be updated in the apiserver. Is that what we want?

Coincidentally I was working on this and I was thinking about this this to solve this issue:

https://github.com/openshift/library-go/compare/master...bertinatto:webhook-config-static?expand=1#

bertinatto: Also, I believe this won't work without registering the scheme in the `init()` function above.

CC @stlaz

slaskawi: > Also, I believe this won't work without registering the scheme in the init() function above.

@bertinatto I believe you're right. Let me send an update in a few minutes.

>  this -1 for expectedGeneration means that every time this function is called (i.e., during the static controller sync), the resource will be updated in the apiserver. Is that what we want?
> 
> Coincidentally I was working on this and I was thinking about this this to solve this issue:
>
> https://github.com/openshift/library-go/compare/master...bertinatto:webhook-config-static?expand=1#

Yes, I was aware of that. Unfortunately I couldn't come up with anything better. 

So I decided to do a tradeoff here and always update the resource. Otherwise, a user could modify it and we would never revert his changes.

Perhaps @stlaz has a better idea how to solve this?

slaskawi: PR added: https://github.com/openshift/library-go/pull/1135

gnufied: PR#1135 does not fixes the issue @bertinatto pointed out though right?

slaskawi: No. The code as it is now always updates performs an update.

PR #1130, opened 2021-07-13
introduces a startup monitor

p0lyn0mial:
The main purpose of the new app is to provide a fallback mechanism that will make the operand to revert to the last healthy revision when the new revision fails.

The fallback mechanism is opt-in by the operator, and kube-apiserver and etcd operators (and potentially kcm and ks) have to consult the deployment topology in the infrastructure resource to decide.

Co-authored-by: Dr. Stefan Schimanski <stefan.schimanski@gmail.com>
Co-authored-by: Lukasz Szaszkiewicz <lukasz.szaszkiewicz@gmail.com>

For more details see: openshift/enhancements#833


pkg/operator/staticpod/startupmonitor/cmd.go
```diff
@@ -0,0 +1,67 @@
...
+	"k8s.io/klog/v2"
+)
+
+type Options struct {
+	// FallbackTimeout specifies a timeout after which the monitor starts the fall back procedure
+	FallbackTimeout time.Duration
+
+	// IsTargetHealthy defines a function that abstracts away assessing operand's health condition.
+	// This is the extention point for the operators to provide a custom health function for their operands
+	IsTargetHealthy func() bool
```
p0lyn0mial, line 21: this could be an interface with `Start, Stop, IsHealthy` methods

pkg/operator/staticpod/startupmonitor/assets/startup-monitor-pod.yaml
```diff
@@ -0,0 +1,56 @@
...
+        requests:
+          memory: 50Mi
+          cpu: 5m
+      securityContext:
+        privileged: true
+  hostNetwork: true
+  terminationGracePeriodSeconds: 5
+  priorityClassName: system-node-critical
+  tolerations:
+    - operator: "Exists"
```
p0lyn0mial, line 46: do we need it?

pkg/operator/staticpod/startupmonitor/assets/startup-monitor-pod.yaml
```diff
@@ -0,0 +1,56 @@
...
+  containers:
+    - name: startup-monitor
+      image: {{.OperatorImage}}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command: ["cluster-kube-apiserver-operator", "startup-monitor"] # todo: make it dynamic
+      args:
+        - {{.Verbosity}}
+        - --fallback-timeout-duration=300s
+        - --target-name=kube-apiserver
```
sttts, line 18: `{{.TargetName}}`

pkg/operator/staticpod/startupmonitor/assets/startup-monitor-pod.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - --manifests-dir=/etc/kubernetes/manifests
+        - --resource-dir=/etc/kubernetes/static-pod-resources
+        - --revision=REVISION
```
sttts, line 21: we really still have `REVISION` and not `{{.Revision}}` ?

pkg/operator/staticpod/startupmonitor/assets/startup-monitor-pod.yaml
sttts: no idea. Better remove it instead of cargo culting :)

pkg/operator/staticpod/startupmonitor/assets/startup-monitor-pod.yaml
p0lyn0mial: yeah, the `REVISION` is replaced by the `installer-pod`

p0lyn0mial: at some point we should clean up :)

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+)
+
+// HealthChecker a contract between the startup monitor and operators
+// allows for initialization and an operand's health condition assessment
+type HealthChecker interface {
+	// Start is used to initialize and start the underlying health monitor
+	Start(context.Context) error
```
sttts, line 26: Start or Run? I.e. non-blocking or blocking?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// IsTargetHealthy defines a function that abstracts away assessing operand's health condition.
+	// the provided functions should be async and cheap in a sense that it shouldn't assess the target
+	// only read the current state.
+	// mainly because we acquire a lock on each sync.
+	IsTargetHealthy() (healthy bool, reason string, err error)
```
sttts, line 32: What is err? Is it nil when unhealthy? Where do we get a human readible message from?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	"k8s.io/klog/v2"
+)
+
+// HealthChecker a contract between the startup monitor and operators
+// allows for initialization and an operand's health condition assessment
+type HealthChecker interface {
+	// Start is used to initialize and start the underlying health monitor
+	Start(context.Context) error
+
+	// IsTargetHealthy defines a function that abstracts away assessing operand's health condition.
```
sttts, line 28: is health = ready?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	// the provided functions should be async and cheap in a sense that it shouldn't assess the target
```
sttts, line 29: why should it be cheap and async?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
p0lyn0mial: it should be non-blocking

pkg/operator/staticpod/startupmonitor/startup_monitor.go
p0lyn0mial: yes, we can rename to `IsTargetReady`

pkg/operator/staticpod/startupmonitor/startup_monitor.go
p0lyn0mial: because we hold a lock

sttts: but then health must be indepedent of locking. Is it?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
p0lyn0mial: I was thinking that the reason would be `fallback-reason` and the `err` could be `fallback-message`

```
   - `startup-monitor.static-pods.openshift.io/fallback-reason: MachineReadableReason`
   - `startup-monitor.static-pods.openshift.io/fallback-message: "Human readable, possibly multiline message"`
```

sttts: err is normally a real error. Stroustrup said somewhere: don't use error flow for non-error cases.

p0lyn0mial: do you think we need `IsTargetHealthy() (healthy bool, reason string, msg string, err error)` ?

p0lyn0mial: or just `IsTargetHealthy() (reason string, msg string, err error)` ?

sttts: would keep the bool, but that's just taste.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+
+func (m *monitor) Run(ctx context.Context) error {
+	klog.Infof("Starting the startup monitor with Interval = %v, Timeout = %v", m.probeInterval, m.timeout)
+	defer klog.Info("Shutting down the startup monitor")
+
+	if err := m.healthChecker.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start the health checker for %s, due to %v", m.targetName, err)
+	}
+
+	wait.Until(m.syncErrorWrapper, m.probeInterval, ctx.Done())
```
sttts, line 94: wait.Until has a fixed interval for the start of the next round, i.e. if sync takes longer than probeInterval, this will hotloop. There is a wait variant that has a constant delay between sync end and next sync.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	return nil
+}
+
+func (m *monitor) syncErrorWrapper() {
+	if err := m.sync(); err != nil {
+		klog.Error(err)
```
sttts, line 100: inline this in the wait.Until call above.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
sttts: And shouldn't this be `wait.PollSomeThing` with a bool return value?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+			return err
+		}
+		return nil
+	}
+	if err != nil {
+		klog.Infof("failed to assess health condition due to err = %v, reason = %v", err, reason)
+	}
+
+	// check if we reached the timeout
+	if time.Now().After(m.monitorTimeStamp.Add(m.timeout)) {
```
sttts, line 147: this is better done through the context timeout externally.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	if m.monitorTimeStamp.IsZero() {
+		m.monitorTimeStamp = time.Now()
+	}
+
+	// first check if the target is healthy
+	// note that we will always reconcile on transient errors
+	// before starting the fall back procedure
+	healthy, reason, err := m.healthChecker.IsTargetHealthy()
```
sttts, line 134: make this synchronous

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	manifestsPath string
+
+	// staticPodResourcesPath points to the directory that holds all files supporting the static pod manifests
+	staticPodResourcesPath string
+
+	// healthChecker is used to get the current operand's health condition
+	healthChecker HealthChecker
+
+	// records the time the monitor has started assessing operand's health condition
+	monitorTimeStamp time.Time
```
sttts, line 72: just use the outer context with a timeout.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+
+func (m *monitor) createLastKnowGoodRevisionAndDestroy() error {
+	// step 0: rm the previous last good known revision if exists
+	// step 1: create last known good revision
+	if err := m.createLastKnowGoodRevisionFor(m.revision, true); err != nil {
+		return err
+	}
+
+	// step 2: commit suicide
+	return m.io.Remove(path.Join(m.manifestsPath, fmt.Sprintf("%s-startup-monitor.yaml", m.targetName)))
```
sttts, line 169: would move all this logic independent of the readiness monitor out of this file, and call it in the outer Run func of the command.

sttts: Also the suicide should happen on both case: timeout and ready operand.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+}
+
+// TODO: pruner|installer: protect the linked revision
+func (m *monitor) fallbackToPreviousRevision() error {
+	// step 0: if the last known good revision doesn't exist
+	//         find a previous revision to work with
```
sttts, line 175: write which is chosen. Latest one != current?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+			return nil
+		}
+
+		targetManifestForPrevRevExists, err := m.fileExists(m.targetManifestPathFor(prevRev))
+		if err != nil {
+			return err // retry, a transient err
+		}
+		if !targetManifestForPrevRevExists {
+			klog.Info("Unable to roll back because a manifest %q hasn't been found for the previous revision %d", m.targetManifestPathFor(prevRev), prevRev)
+			// TODO: commit suicide ? this seems to be fatal
```
sttts, line 199: it is, and it's best we can do, i.e. let the current revision active and hope.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		// step 1: create the last known good revision file
+		if err := m.createLastKnowGoodRevisionFor(prevRev, false); err != nil {
+			return err
+		}
+	}
+
+	// step 2: if the last known good revision exits and we got here
+	//         that could mean that:
```
sttts, line 210: could mean of

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+			return nil
+		}
+
+		// step 1: create the last known good revision file
+		if err := m.createLastKnowGoodRevisionFor(prevRev, false); err != nil {
+			return err
+		}
+	}
+
+	// step 2: if the last known good revision exits and we got here
```
sttts, line 209: exists or exits?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+		return err
+	}
+
+	lastKnownGoodPodBytes := []byte(resourceread.WritePodV1OrDie(lastKnownGoodPod))
+	klog.Infof("Writing a static pod manifest %q \n%s", path.Join(rootTargetManifestPath), lastKnownGoodPodBytes)
+	if err := m.io.WriteFile(path.Join(rootTargetManifestPath), lastKnownGoodPodBytes, 0644); err != nil {
+		return err
+	}
+
+	// TODO: commit suicide ?
```
sttts, line 246: yes

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+		} else if exists {
+			if err := m.io.Remove(m.lastKnownGoodManifestDstPath()); err != nil {
+				return err
+			}
+			klog.Info("Removed existing last known good revision manifest %s", m.lastKnownGoodManifestDstPath())
+		}
+	}
+
+	// step 1: create last known good revision
+	if err := m.io.Symlink(revisionedTargetManifestPath, m.lastKnownGoodManifestDstPath()); err != nil {
```
sttts, line 266: isn't there a force variant for this? I always thought symlink update is atomic.

pkg/operator/staticpod/startupmonitor/startup_monitor.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	if len(allRevisions) < 2 {
+		return 0, false, nil
+	}
+	sort.IntSlice(allRevisions).Sort()
+	return allRevisions[len(allRevisions)-2], true, nil
+}
+
+func (m *monitor) fileExists(filepath string) (bool, error) {
```
sttts, line 350: do we need this?

pkg/operator/staticpod/startupmonitor/startup_monitor.go
sttts: fixed

pkg/operator/staticpod/startupmonitor/startup_monitor.go
sttts: done

sttts:
Needs squashing.

Then lgtm.

/approve

PR #1131, opened 2021-07-13
introduces FLock for supporting file locking to coordinate work between processes

p0lyn0mial:


p0lyn0mial:
will be used in https://github.com/openshift/library-go/pull/1130

pkg/operator/staticpod/flock.go
```diff
@@ -0,0 +1,95 @@
...
+// TryLock tries to acquire an exclusive lock on a file
+// until it succeeds, an error is returned or the timeout is reached.
+//
+// The callers MUST call the corresponding Unlock method to release the lock and associated resources except when an error
+// is returned (including the timeout).
+//
+// This method is safe for concurrent access.
+//
+// Note the given timeout shouldn't be less than 1 second.
+func (f *FLock) TryLock(timeout time.Duration) error {
```
sttts, line 37: not a context?

p0lyn0mial: we could but then the usage becomes
```
ctx, cancel := context.WithTimeout()
defer cancel()
flock.TryLock(ctx)
```

vs

`flock.TryLock(time.Second * 5)`

PR #1133, opened 2021-07-13
changes the installer-pod to write optional pods

p0lyn0mial:
up until now, the installer pod expected only a single pod manifest in the configmap (the cm is specified by the --pod flag)

since we are planning to add a startup monitor for assessing operand's health check we are extending the installer
to take into account optional keys that might contain other pods.

the new pods will be written to the same locations as the main pod (revisioned, manifests).
the name will be derived from the name filed and suffixed with "-pod.yaml"

xref: https://github.com/openshift/library-go/pull/1130

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -3,7 +3,6 @@ package installerpod
 import (
 	"context"
 	"fmt"
-	"github.com/openshift/library-go/pkg/operator/staticpod"
 	"io/ioutil"
 	"os"
 	"path"
@@ -26,6 +25,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	"github.com/openshift/library-go/pkg/operator/resource/retry"
+	"github.com/openshift/library-go/pkg/operator/staticpod"
 )
 
 type InstallOptions struct {
@@ -288,8 +288,8 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
 		}
 	}
 
-	// Gather pod yaml from config map
-	var rawPodBytes string
+	// Gather the config map that holds pods to be installed
+	var podsConfigMap *corev1.ConfigMap
 
 	err := retry.RetryOnConnectionErrors(ctx, func(ctx context.Context) (bool, error) {
 		klog.Infof("Getting pod configmaps/%s -n %s", o.nameFor(o.PodConfigMapNamePrefix), o.Namespace)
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
untested

/assign @sttts

p0lyn0mial:
proof PR https://github.com/openshift/cluster-kube-apiserver-operator/pull/1175

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -300,55 +300,37 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
-		return err
-	}
-	klog.Infof("Writing static pod manifest %q ...\n%s", path.Join(o.PodManifestDir, podFileName), finalPodBytes)
-	if err := ioutil.WriteFile(path.Join(o.PodManifestDir, podFileName), []byte(finalPodBytes), 0644); err != nil {
-		return err
-	}
+	// then write the required pod and all optional
+	// the key must contain pod.yaml to be considered
+	for rawPodKey, rawPod := range podsConfigMap.Data {
+		if !strings.Contains(rawPodKey, "pod.yaml") {
```
sttts, line 319: What else do you expect here?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -300,55 +300,37 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
+			continue
+		}
+		var podName string
+		if rawPodKey == "pod.yaml" {
+			// this is our main pod, assign the previous name for compatibility
+			podName = o.PodConfigMapNamePrefix + ".yaml"
+		}
```
sttts, line 326: can't we have a fully qualified name as key? e.g. kube-apiserver-pod.yaml.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -300,55 +300,37 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
-	if err := ioutil.WriteFile(path.Join(o.PodManifestDir, podFileName), []byte(finalPodBytes), 0644); err != nil {
-		return err
-	}
+	// then write the required pod and all optional
+	// the key must contain pod.yaml to be considered
+	for rawPodKey, rawPod := range podsConfigMap.Data {
+		if !strings.Contains(rawPodKey, "pod.yaml") {
+			continue
+		}
+		var podName string
```
sttts, line 322: this is not the podName, but the manifestFileName

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -300,55 +300,37 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
+		if rawPodKey == "pod.yaml" {
+			// this is our main pod, assign the previous name for compatibility
+			podName = o.PodConfigMapNamePrefix + ".yaml"
+		}
 
+		klog.Infof("Writing a pod under %q key \n%s", rawPodKey, rawPod)
+		err := o.writePod([]byte(rawPod), podName, resourceDir)
```
sttts, line 329: nit: dir, manifestFileName

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -300,55 +300,37 @@ func (o *InstallOptions) copyContent(ctx context.Context) error {
...
-	} else if !os.IsNotExist(err) {
-		return err
-	}
-	klog.Infof("Writing static pod manifest %q ...\n%s", path.Join(o.PodManifestDir, podFileName), finalPodBytes)
-	if err := ioutil.WriteFile(path.Join(o.PodManifestDir, podFileName), []byte(finalPodBytes), 0644); err != nil {
-		return err
-	}
+	// then write the required pod and all optional
+	// the key must contain pod.yaml to be considered
+	for rawPodKey, rawPod := range podsConfigMap.Data {
```
sttts, line 318: does order matter?

p0lyn0mial: I was thinking about the same, I can make it stable.
At the moment it doesn't matter, the `startup-monitor` will wait until an operand matches its current revision.

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: true :)

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: oh, sure, we could, good idea, will change.

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: as of today the `cm` contains a few additional keys, like `forceRedeploymentReason`

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: I think that this is a two-step process, first, we need to merge this PR, and then we need to update `kas-o`, right ?

pkg/operator/staticpod/installerpod/cmd.go
sttts: I also think kubelet is not deterministic in any way. So I think we are fine.

pkg/operator/staticpod/installerpod/cmd.go
sttts: would check for suffix `-pod.yaml`.

pkg/operator/staticpod/installerpod/cmd.go
sttts: probably, yes.

p0lyn0mial: ok, I will prepare a follow up PRs

PR #1134, opened 2021-07-14
make component owner reference customizable in controllercmd

atiratree:
This is needed to lower RBAC privileges (fetching pod definition) for controller that know the owner or want to set themselves as the owner.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openshift/library-go/pkg/config/serving"
 	"github.com/openshift/library-go/pkg/controller/fileobserver"
 	"github.com/openshift/library-go/pkg/operator/events"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/version"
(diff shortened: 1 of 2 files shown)


soltysh (approved):
/lgtm

ravisantoshgudimetla (approved):
/lgtm

pkg/controller/controllercmd/builder.go
```diff
@@ -188,6 +190,12 @@ func (b *ControllerBuilder) WithEventRecorderOptions(options record.CorrelatorOp
 	return b
 }
 
+// WithComponentOwnerReference overrides controller reference resolution for event recording
+func (b *ControllerBuilder) WithComponentOwnerReference(reference corev1.ObjectReference) *ControllerBuilder {
```
mfojtik, line 194: nit: *corev1.ObjectReference

pkg/controller/controllercmd/cmd.go
```diff
@@ -281,5 +290,9 @@ func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
 		builder = builder.WithServer(config.ServingInfo, config.Authentication, config.Authorization)
 	}
 
+	if c.ComponentOwnerReference != nil {
+		builder = builder.WithComponentOwnerReference(*c.ComponentOwnerReference)
```
mfojtik, line 294: then you don't have to check this

pkg/controller/controllercmd/builder.go
atiratree: I wanted to prevent the mutation from outside, but I suppose the implementation will be easier with a pointer

atiratree:
@mfojtik fixed

PR #1136, opened 2021-07-15
encryption condition controller doesn't reset previously set conditon

p0lyn0mial:
the condition controller is used to report the current state of the encryption: EncryptionCompleted, EncryptionInProgress or DecryptionCompleted

unfortunately, the previously set condition might have been cleared due to a bug in the controller every time the API server was deployed.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1982398

### DIFF
--- a/pkg/operator/encryption/controllers/condition_controller.go
+++ b/pkg/operator/encryption/controllers/condition_controller.go
@@ -84,6 +84,8 @@ func (c *conditionController) sync(_ context.Context, _ factory.SyncContext) (er
 	encryptedGRs := c.provider.EncryptedGRs()
 	currentConfig, desiredState, foundSecrets, transitioningReason, err := statemachine.GetEncryptionConfigAndState(c.deployer, c.secretClient, c.encryptionSecretSelector, encryptedGRs)
 	if err != nil || len(transitioningReason) > 0 {
+		// do not update the encryption condition (cond). Note: progressing is set elsewhere.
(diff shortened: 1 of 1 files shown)


pkg/operator/encryption/controllers/condition_controller.go
```diff
@@ -84,6 +84,9 @@ func (c *conditionController) sync(_ context.Context, _ factory.SyncContext) (er
 	encryptedGRs := c.provider.EncryptedGRs()
 	currentConfig, desiredState, foundSecrets, transitioningReason, err := statemachine.GetEncryptionConfigAndState(c.deployer, c.secretClient, c.encryptionSecretSelector, encryptedGRs)
 	if err != nil || len(transitioningReason) > 0 {
+		// do not reset previously set condition
+		// transitioningReason means APIServerRevisionNotConverged
+		cond = nil
```
sttts, line 89: cannot follow here, how it solves the BZ.

What you did here is: don't update conditions when we are progressing.

p0lyn0mial: don't overwrite previously set condtion

p0lyn0mial: like `EncryptionInProgress`

p0lyn0mial: or any other

PR #1137, opened 2021-07-19
Bug 1986003: Bump k8s.io to 1.22.0-rc.0

soltysh:


soltysh:
Manually adding approved label since this is pretty standard operation and the PR was opened long enough that people are aware.

PR #1138, opened 2021-07-20
Custom CA Certs for Route helpers

slaskawi:
https://issues.redhat.com/browse/AUTH-14

This Pull Request introduces a library helpers for dealing with routes as well as custom CA Certificates for Routes (introduced in https://github.com/openshift/cluster-authentication-operator/pull/430).

Most of the bits have been copy-pasted with slight modifications. ﻿


pkg/operator/resource/resourceapply/route.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	modified := resourcemerge.BoolPtr(false)
+	existingCopy := existing.DeepCopy()
+	resourcemerge.EnsureObjectMeta(modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+
+	// this guarantees that route.Spec.Host is set to the current canonical host
+	if *modified || !equality.Semantic.DeepEqual(existingCopy.Spec, required.Spec) {
+		if klog.V(4).Enabled() {
+			klog.Infof("Route %q changes: %s", existing.Name, JSONPatchNoError(existing, existingCopy))
```
stlaz, line 36: this might leak private key stored in that route

stlaz: `JSONPatchNoError(existing, existingCopy)`

shouldn't this compare existing and required?

pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -8,7 +8,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
 	kvalidation "k8s.io/apimachinery/pkg/util/validation"
-	field "k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apimachinery/pkg/util/validation/field"
```
stlaz, line 11: keep as is

pkg/route/routeapihelpers/routeapihelpers_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	field "k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apimachinery/pkg/util/validation/field"
```
stlaz, line 12: keep as is or move to a separate commit

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
```diff
@@ -0,0 +1,133 @@
+package routecomponenthelpers
```
stlaz, line 1: is this a 1:1 copy?

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
```diff
@@ -0,0 +1,133 @@
...
+	"k8s.io/client-go/util/retry"
+
+	"github.com/openshift/library-go/pkg/certs"
+)
+
+// RouteComponentControllerHelper aggregates necessary interfaces for Custom Certs for routes functions.
+type RouteComponentControllerHelper struct {
+	secretLister  corev1listers.SecretLister
+	ingressClient configsetterv1.IngressInterface
+}
```
stlaz, line 24: keep as two separate clients in function arguments

stlaz (commented):
We should also move the whole `CustomRouteController` here

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
slaskawi: No, this part is new. I had to create a construct for Listers and Interfaces. Nevertheless once I move the controller, it will remove this.

pkg/route/routeapihelpers/routeapihelpers_test.go
slaskawi: I'll leave it as it was.

pkg/route/routeapihelpers/routeapihelpers.go
slaskawi: :+1:

pkg/operator/resource/resourceapply/route.go
slaskawi: :man_facepalming: Of course!

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
slaskawi: Will do. I'll move the whole controller here and we'll see what we can unify.

pkg/certs/util.go
```diff
@@ -68,3 +71,60 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
 	}
 	return strings.Join(output, "\n")
 }
+
+func ValidatePrivateKey(pemKey []byte) []error {
+	if len(pemKey) == 0 {
+		return []error{fmt.Errorf("required private key is empty")}
+	}
+	if _, err := keyutil.ParsePrivateKeyPEM(pemKey); err != nil {
+		return []error{fmt.Errorf("failed to parse the private key")}
```
awgreene, line 80: It might be worthwhile to surface the error returned by the `ParsePrivateKeyPEM` function.

pkg/certs/util.go
```diff
@@ -68,3 +71,60 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
...
+
+	certs, certErrs := parseCerts(pem)
+	errs = append(errs, certErrs...)
+
+	if numCerts := len(certs); numCerts == 0 {
+		return append(errs, fmt.Errorf("expected at least one server certificate"))
+	}
+
+	return errs
+}
```
awgreene, line 96:
```suggestion
func ValidateServerCert(pem []byte) []error {
	certs, errs := parseCerts(pem)
	if len(errs) != 0 {
	     return errs
	 }
	 if len(certs) == 0 {
	      return []error{fmt.Errorf("expected at least one server certificate")}
	 }
	 return nil
}
```

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
```diff
@@ -0,0 +1,133 @@
...
+func UpdateComponentRouteStatus(ctx context.Context, routeComponentHelper *RouteComponentControllerHelper, componentRoute *configv1.ComponentRouteStatus) (bool, error) {
+	// Override the timestamps
+	now := metav1.Now()
+
+	// Create a copy for comparison and remove transaction times
+	componentRouteCopy := componentRoute.DeepCopy()
+	setLastTransactionTime(componentRouteCopy, now)
+
+	updated := false
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
awgreene, line 45: Is it standard for the library to retry on conflicts?

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
```diff
@@ -0,0 +1,133 @@
...
+	return updated, err
+}
+
+func setLastTransactionTime(componentRoute *configv1.ComponentRouteStatus, now metav1.Time) {
+	for i := range componentRoute.Conditions {
+		componentRoute.Conditions[i].LastTransitionTime = now
+	}
+}
+
+// ConfigureRoute configures route.Spec.Host along with
```
awgreene, line 80: This is an incomplete sentence.

awgreene (changes requested):
Nice work @slaskawi, a few nits.

pkg/route/routecomponenthelpers/routecomponentcontrollerhelpers.go
slaskawi: After I moved the controller into the Library Go, this comment probably is no longer applicable. Resolving.

pkg/certs/util.go
slaskawi: Fixed thanks!

pkg/certs/util.go
slaskawi: Done

pkg/operator/resource/resourceapply/route.go
slaskawi: Done

slaskawi: Replaced with a custom function that masks TLS keys.

slaskawi:
Hey @spadgett @jhadvig, 

@stlaz mentioned you were working on a similar (if not the same) controller on the UI side. Perhaps you could re-use this one? Would you mind to review this PR?

Thanks!

pkg/certs/util.go
```diff
@@ -68,3 +71,58 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
...
 }
+
+func ValidatePrivateKey(pemKey []byte) []error {
+	if len(pemKey) == 0 {
+		return []error{fmt.Errorf("required private key is empty")}
+	}
+	if _, err := keyutil.ParsePrivateKeyPEM(pemKey); err != nil {
+		return []error{fmt.Errorf("failed to parse the private key, %w", err)}
+	}
+	return []error{}
```
s-urbaniak, line 82: nit: for consistency let's `return nil`.

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		},
+	}
+	conditions := checkErrorsConfiguringCustomRoute(customRouteErrors)
+	if conditions == nil {
+		conditions = checkIngressURI(ingressConfig, route)
+		if conditions == nil {
+			conditions = checkRouteAvailablity(c.secretLister, ingressConfig, route)
+		}
+	}
+	componentRoute.Conditions = ensureDefaultConditions(conditions)
```
deads2k, line 220: In our experience, this construction is chronically buggy.  The sync method where you do work shouldn't set the status because when it fails (and it will fail), it doesn't keep the status up to date.  If you want to wrap your main sync method in something else *and* be sure to handle panics in it, you can try to set both conditions.

It seems slightly more reasonable to update the controller factory bit to always handle this.

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		return err
+	}
+
+	// update ingressConfig status
+	if err = c.updateIngressConfigStatus(ctx, ingressConfigCopy, nil); err != nil {
+		return err
+	}
+
+	// sync the secret
+	return c.syncSecret(secretName)
```
deads2k, line 127: looks pretty weird to do work *after* you've set the condition above.

pkg/operator/customroute/custom_route_conditions.go
```diff
@@ -0,0 +1,205 @@
...
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+func ensureDefaultConditions(conditions []metav1.Condition) []metav1.Condition {
+	for _, conditionType := range []string{"Progressing", "Degraded"} {
+		condition := findCondition(conditions, conditionType)
+		if condition == nil {
+			conditions = append(conditions, metav1.Condition{
+				LastTransitionTime: metav1.Now(),
+				Type:               conditionType,
```
deads2k, line 29: this looks weird/incorrect.  Conditions on operator resources are suffixed with Progressing and Degraded, but they shouldn't actually *be* progressing or degraded.

stlaz: these won't be conditions in an operator resource but in the `ingress.config` object's status

pkg/operator/customroute/custom_route_conditions.go
```diff
@@ -0,0 +1,205 @@
...
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			TLSClientConfig: &tls.Config{
+				RootCAs: rootCAs,
+			},
+		},
+	}
+
+	// Make a request to the endpoint, expect a 403
```
stlaz, line 155: nit: looks like an outdated comment

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		if len(errors) != 0 {
+			return fmt.Errorf("error validating secret %s/%s: %v", "openshift-config", secretName, errors)
+		}
+	}
+	return nil
+}
+
+func (c *customRouteController) updateIngressConfigStatus(ctx context.Context, ingressConfig *configv1.Ingress, customRouteErrors []error) error {
+	// update ingressConfig status
+	// TODO: consider using c.targetRoute. Nevertheless the downside is that we might have a stale reference?
```
stlaz, line 192: the FLP me wants `targetRoute` to be an argument of this function

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	if err = c.updateIngressConfigStatus(ctx, ingressConfigCopy, nil); err != nil {
+		return err
+	}
+
+	// sync the secret
+	return c.syncSecret(secretName)
+}
+
+func (c *customRouteController) getOAuthRouteAndSecretName(ingressConfig *configv1.Ingress) (*routev1.Route, string, []error) {
+	route := c.targetRoute
```
stlaz, line 131: should we use bindata for this instead?

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (c *customRouteController) syncSecret(secretName string) error {
+	source := resourcesynccontroller.ResourceLocation{}
+	if secretName != "" {
+		source = resourcesynccontroller.ResourceLocation{Namespace: "openshift-config", Name: secretName}
+	}
+
+	destination := resourcesynccontroller.ResourceLocation{Namespace: c.destSecret.Namespace, Name: c.destSecret.Name}
+
+	return c.resourceSyncer.SyncSecret(destination, source)
```
stlaz, line 233: this should use https://github.com/openshift/library-go/blob/14aa5b749d8ddbae509ea97c988228934dbbd3db/pkg/operator/resource/resourceapply/core.go#L387 instead

pkg/operator/resource/resourceapply/json_patch_helpers.go
```diff
@@ -68,3 +69,26 @@ func JSONPatchSecretNoError(original, modified *corev1.Secret) string {
...
+	safeModified := modified.DeepCopy()
+	safeOriginal := original.DeepCopy()
+
+	if safeOriginal.Spec.TLS.Key != "" {
+		safeOriginal.Spec.TLS.Key = "MASKED"
+	}
+
+	if safeModified.Spec.TLS.Key != "" {
+		safeModified.Spec.TLS.Key = "MASKED"
+	}
```
stlaz, line 91: compare them and somehow mark them as changed if they are different

pkg/certs/util.go
slaskawi: Actually, that `nil` from `ValidateServerCert` is inconsistent. Let me fix that one instead of this.

pkg/operator/customroute/custom_route_conditions.go
slaskawi: Just echoing, what @stlaz said. This is the `Ingress.Status.ComponentRoutes[].Conditions[]` from `github.com/openshift/api/config/v1/types_ingress.go`

pkg/operator/customroute/custom_route_conditions.go
slaskawi: Removed.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Syncing secret mechanism uses `resourcesynccontroller.ResourceSyncer`. The syncer validates if source or destination namespaces and names are set and then syncs Secrets/ConfigMaps asynchronously.

Because of this design, the only way (at least that I know) to know if anything goes wrong is to observe the `Conditions` in the Operator's status. Checking the error here wouldn't buy us anything. 

If you know any better way to do it - please let me know. Perhaps, this piece of code could be improved.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Actually no, we can't do it here. 

If we want this code to be reusable (by the Console Team), we need to obtain the route from the Operator that will be using this code. In our case, this will be `openshift-authentication/oauth-openshift`, but in the Console's case it will be something different. 

By the way, the library-go doesn't seem to ship bindata at all. But that's a side note.

pkg/operator/customroute/custom_route_controller.go
slaskawi: I checked this piece of code again and the comment is wrong. We can't use the `c.targetRoute` as we extract the `route.Spec.Host`. This can be set either by a user or automatically by the Route Controller. We always need to grab a fresh reference here...

Erasing the comment.

pkg/operator/customroute/custom_route_controller.go
slaskawi: This makes a lot of sense. Let me divide the code in the `sync` method into 2 sections. One of them will be creating the Route and requesting the Secret sync and the other will handle the status updates. 

It's worth to mention that this approach minimizes the chances of getting Conditions out of sync. Nevertheless, if a bug happens in setting the Conditions, you will never know that (unless you analyze the logs).

slaskawi: During the refactoring I also changed the semantics a bit. I will need to verify the integration tests once we're happy with this PR.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Actually, it's pretty clever and I'd like to leave it :)

Doing it the way it is implemented we automatically handles deletes. If the `secretName` is empty, we'll create an empty `source` and pass it into the `resourceSyncer`. If the source is empty, the `resourceSyncer` is clever enough to delete the destination too. If I used the method you sent me, I'd need to duplicate this piece of code.

pkg/operator/resource/resourceapply/json_patch_helpers.go
slaskawi: Done

slaskawi:
Updated the code along with the comments. @stlaz @s-urbaniak @deads2k Please have another look in a free slot.

pkg/operator/customroute/custom_route_controller.go
stlaz: > If we want this code to be reusable (by the Console Team), we need to obtain the route from the Operator that will be using this code.

...which could be done by using bindata, it's very likely that the `route` here will actually be loaded from bindata while starting this operator.

See https://github.com/openshift/console-operator/blob/c92d52d4a83afb77e6b586d37f1f0dfd6d891df2/pkg/console/subresource/route/route.go#L143-L152, they're using bindata already. I was helping them a little to move to it.

> By the way, the library-go doesn't seem to ship bindata at all. But that's a side note.

That's not entirely true, there are subpackages that ship their own bindata, like `pkg/operator/apiserver/audit/bindata` for example. Also, there are other controllers that are capable of handling bindata, like https://github.com/openshift/library-go/blob/14aa5b749d8ddbae509ea97c988228934dbbd3db/pkg/operator/staticresourcecontroller/static_resource_controller.go#L68

pkg/operator/customroute/custom_route_controller.go
stlaz: fair enough :+1:

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+
+		route.Spec.Host = hostname
+	}
+
+	return route, secretName, nil
+}
+
+func (c *customRouteController) validateCustomTLSSecret(secretName string) error {
+	if secretName != "" {
```
stlaz, line 157:
```golang
if len(secretName) == 0 {
    return nil
}
... the rest here...
```

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (c *customRouteController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	ingressConfig, err := c.ingressLister.Get("cluster")
+	if err != nil {
+		return err
+	}
+
+	ingressConfigCopy := ingressConfig.DeepCopy()
+
+	errors := c.syncResources(ctx, syncCtx, ingressConfigCopy)
```
stlaz, line 104: nit: the rest of the code usually uses `errs` for error slices

stlaz (commented):
nits, otherwise lgtm

pkg/operator/customroute/custom_route_controller.go
slaskawi: Done.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Just to educate me - what's the difference between those two?

slaskawi: Fixed.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Resolving.

pkg/operator/customroute/custom_route_controller.go
slaskawi: I'm still not entirely convinced. What if one of the operators (either Console or CAO) moves to Go Embed and the other doesn't? 

I'm still in favor of pushing the responsibility of obtaining the Route to the caller. The only improvement that I can think of here is to make the Route parameter in the constructor a function pointer. Then, the caller could put there whatever he wants. Let me know if you'd like me to implement this change.

slaskawi:
Updated and rebased. Ready for the next round @s-urbaniak @stlaz

pkg/certs/util.go
s-urbaniak: If you want to return empty slices, returning nil is more idiomatic. You avoid memory allocation.

pkg/operator/customroute/custom_route_controller.go
stlaz: The difference is that you don't have to deal with the indentation for the rest of the function, therefore it's considered more clean.

pkg/operator/customroute/custom_route_controller.go
stlaz: I wouldn't mind forcing Go embed, but for the time being the library-go is using bindata for this purpose on other places so I assume the move will be unite for these at certain point.

The reason I don't like this is that it's unnecessary state that's being kept in a somewhat global variable and is therefore prone to side effects. I've done my part in functional programming and so I consider this ugly.

That being said, I can live with it.

slaskawi: As agreed to put the route into the bindata. If the Console Operator requires something else - we'll adapt.

pkg/certs/util.go
slaskawi: Done

slaskawi:
Comments addressed

pkg/certs/util.go
```diff
@@ -68,3 +71,58 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
...
+	if len(pemKey) == 0 {
+		return []error{fmt.Errorf("required private key is empty")}
+	}
+	if _, err := keyutil.ParsePrivateKeyPEM(pemKey); err != nil {
+		return []error{fmt.Errorf("failed to parse the private key, %w", err)}
+	}
+	return nil
+}
+
+func ValidateServerCertDate(pem []byte) []error {
```
sttts, line 85: godoc. This is not obvious what it does. Maybe `ValidateServerCertValidity`?

slaskawi: Added. I also inlined the `parse` function as this was the only place it was used.

pkg/operator/customroute/custom_route_conditions.go
stlaz: I still see the comment :neutral_face:

pkg/operator/customroute/custom_route_controller.go
stlaz: Is this still the same?

pkg/certs/util.go
```diff
@@ -68,3 +71,51 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
 	}
 	return strings.Join(output, "\n")
 }
+
+func ValidatePrivateKeyFormat(pemKey []byte) []error {
```
stlaz, line 75: Wasn't there supposed to be a godoc here as well?

stlaz (commented):
Are we missing a push of the latest changes?

pkg/certs/util.go
slaskawi: Added

slaskawi:
> Is this still the same?

@stlaz Yes, but the type is different. Now it's a `*routev1.Route`, which gets pulled from bindata as you asked.

> Are we missing a push of the latest changes?

@stlaz Nope, everything seems to be up to date. What else do you miss from this PR?

pkg/operator/customroute/custom_route_controller.go
slaskawi: Nope, it's not, see https://github.com/openshift/library-go/pull/1138#issuecomment-910024433

stlaz:
> Yes, but the type is different. Now it's a *routev1.Route, which gets pulled from bindata as you asked.

I expected it to be passed as an argument, not to stay a controller-local variable, but I don't mind the current change.
> Nope, everything seems to be up to date. What else do you miss from this PR?

I still see the outdated comment that you mentioned was gone :neutral_face:

s-urbaniak:
/retitle Custom CA Certs for Route helpers

pkg/operator/customroute/manifests/route.yaml
```diff
@@ -0,0 +1,21 @@
+# emulates server-side defaulting as in https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apis/route/configv1listers/defaults.go
```
stlaz, line 1: this does not belong to library-go but to the CAO

stlaz (changes requested):

pkg/operator/customroute/manifests/route.yaml
slaskawi: Changed to a function call, see https://github.com/openshift/cluster-authentication-operator/pull/478/files#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R445

pkg/operator/customroute/custom_route_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	controller := &customRouteController{
+		destSecret: types.NamespacedName{
+			Namespace: destSecretNamespace,
+			Name:      destSecretName,
+		},
+		componentRoute: types.NamespacedName{
+			Namespace: componentRouteNamespace,
+			Name:      componentRouteName,
+		},
+		targetRoute:    targetRouteResourceGetter(),
```
s-urbaniak, line 78: I might have missed some discussion points and not a blocker for merging, but why not pass `*routev1.Route` directly?

slaskawi: We went through a long path with @stlaz about this. I initially implement it this way. Then we switched to `bindata`. Finally, we switched into a function pointer, so that the caller can use either `bindata`, Go Embed or whatever he/she likes.

@stlaz wanted to keep it here to avoid side effects.

s-urbaniak: leaving the final word to @stlaz but since the getter func is only used in the constructor I don't see a reason why we shouldn't pass the full reference and document that mutations happen in this controller or consequently use `AssetFunc` as is done in staticresourcecontroller.

stlaz: use a combination of the `AssetFunc` and `resourceread.ReadRouteV1OrDie`

pkg/certs/util.go
```diff
@@ -68,3 +71,52 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
...
+
+	if len(pemCerts) == 0 {
+		return []error{fmt.Errorf("required certificate is empty")}
+	}
+
+	var errs []error
+	now := time.Now()
+
+	cert, rest := pem.Decode(pemCerts)
+	for ; cert != nil; cert, rest = pem.Decode(rest) {
```
stlaz, line 98: nit: could be one line

pkg/certs/util.go
```diff
@@ -68,3 +71,52 @@ func CertificateBundleToString(bundle []*x509.Certificate) string {
...
+			continue
+		}
+
+		certs = append(certs, parsed)
+	}
+
+	if len(errs) != 0 {
+		return errs
+	}
+	return nil
```
stlaz, line 121: can't we just return errs?

stlaz (changes requested):

pkg/certs/util.go
slaskawi: @s-urbaniak asked me to return `nil` here as a more idiomatic way of returning errors. See https://github.com/openshift/library-go/pull/1138#discussion_r680882743.

I'm leaving this as is.

pkg/certs/util.go
slaskawi: I prefer to leave this as is.

pkg/operator/customroute/custom_route_controller.go
slaskawi: Done

PR #1142, opened 2021-07-21
operators: wire context properly to operator controllers

mfojtik:


pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -101,7 +101,7 @@ func checkDiscoveryForAPIService(restclient rest.Interface, apiService *apiregis
 			defer wg.Done()
 			defer utilruntime.HandleCrash()
 
-			ctx, ctxCancelFn = context.WithTimeout(context.TODO(), 25*time.Second)
+			ctx, ctxCancelFn = context.WithTimeout(context.Background(), 25*time.Second)
```
sttts, line 104: why no ctx from outside? IMO background is wrong for go routines with an wg.Wait just behind in code.

pkg/operator/events/eventstesting/recorder_wrapper.go
```diff
@@ -11,6 +12,10 @@ type EventRecorder struct {
 	testingEventRecorder *TestingEventRecorder
 }
 
+func (e *EventRecorder) WithContext(ctx context.Context) events.Recorder {
+	return e
+}
+
```
sttts, line 18: what is this?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apiserver/pkg/server"
+
```
sttts, line 13: order

pkg/operator/apiserver/controller/apiservice/apigroup.go
mfojtik: fixed, wired context from the outside

pkg/operator/events/eventstesting/recorder_wrapper.go
mfojtik: this is a "eventtesting" wrapper, I need to to meet the interface, but in testing we never use the context, so no need to store it.

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: fixed.

PR #1143, opened 2021-07-21
factory: add AddRateLimitedAfter to sync context queue

sttts:
Controller might have to requeue the key, but after a minimum time while taking into account the rate limited. This PR exposes a `AddRateLimitedAfter` function from the sync context queue.

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -18,7 +18,8 @@ import (
 // the sync to be triggered.
 type syncContext struct {
 	eventRecorder events.Recorder
-	queue         workqueue.RateLimitingInterface
+	rl            workqueue.RateLimiter
+	queue         addRateLimitedAfterWorkqueue
 	queueKey      string
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/factory/controller_context.go
```diff
@@ -107,3 +111,17 @@ func namespaceChecker(interestingNamespaces []string) func(obj interface{}) bool
 		return false
 	}
 }
+
+// rateLimitedAddAfterWorkqueue is like workqueue.RateLimitingInterface, but takes the
```
mfojtik, line 115: nit: i would move this to separate file, contextually this has nothing to do with context, it is new rate limiter implementation :-)

pkg/controller/factory/controller_context.go
```diff
@@ -107,3 +111,17 @@ func namespaceChecker(interestingNamespaces []string) func(obj interface{}) bool
...
+// rate limiter into account on AddAfter.
+type rateLimitedAddAfterWorkqueue struct {
+	workqueue.RateLimitingInterface
+	rateLimiter workqueue.RateLimiter
+}
+
+func (q rateLimitedAddAfterWorkqueue) AddAfter(item interface{}, duration time.Duration) {
```
mfojtik, line 122: nit: would explain what this does comparing to default in godoc

pkg/controller/factory/controller_context.go
sttts: done

pkg/controller/factory/controller_context.go
sttts: done

pkg/controller/factory/controller_context.go
```diff
@@ -18,21 +18,24 @@ import (
...
 	return syncContext{
-		queue:         workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), name),
+		rl:            rl,
+		queue:         addRateLimitedAfterWorkqueue{workqueue.NewNamedRateLimitingQueue(rl, name), rl},
 		eventRecorder: recorder.WithComponentSuffix(strings.ToLower(name)),
 	}
 }
 
-func (c syncContext) Queue() workqueue.RateLimitingInterface {
+func (c syncContext) Queue() RateLimitedAddAfterWorkqueue {
```
mfojtik, line 38: this will still make it default to your queue right? I mean all those 100+ controllers we run will switch to this after we bump :-)

sttts: it won't. The method is new.

PR #1145, opened 2021-07-22
staticpod/installer: get rid of sleep in sync loop

sttts:
We had a time.Sleep in the installer controller sync loop to delay progress such that load balancer have time to notice ready pods. This is awkward. There is the sync queue, just return and requeue after the given time.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -339,26 +339,23 @@ func (c *InstallerController) timeToWaitBeforeInstallingNextPod(ctx context.Cont
 
 // manageInstallationPods takes care of creating content for the static pods to install.
 // returns whether or not requeue and if an error happened when updating status.  Normally it updates status itself.
-func (c *InstallerController) manageInstallationPods(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, originalOperatorStatus *operatorv1.StaticPodOperatorStatus) (bool, error) {
+func (c *InstallerController) manageInstallationPods(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, originalOperatorStatus *operatorv1.StaticPodOperatorStatus) (bool, time.Duration, error) {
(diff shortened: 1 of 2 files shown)


aojea:
LGTM
how do you run an e2e job against the PRs in this repo?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -339,26 +339,23 @@ func (c *InstallerController) timeToWaitBeforeInstallingNextPod(ctx context.Cont
...
 	}
 
 	// determine the amount of time to delay before creating the next installer pod.  We delay to avoid an LB outage (see godoc on minReadySeconds)
 	sleepTime := c.timeToWaitBeforeInstallingNextPod(ctx, operatorStatus.NodeStatuses)
 	if sleepTime > 0 {
-		select {
-		case <-ctx.Done():
-		case <-time.After(sleepTime):
-		}
+		return true, sleepTime, nil
```
p0lyn0mial, line 358: nit: rename `sleepTime` to `retryAfter` ?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -979,9 +976,10 @@ func (c InstallerController) Sync(ctx context.Context, syncCtx factory.SyncConte
 
 	// Only manage installation pods when all required certs are present.
 	if err == nil {
-		requeue, syncErr := c.manageInstallationPods(ctx, operatorSpec, operatorStatus)
+		requeue, after, syncErr := c.manageInstallationPods(ctx, operatorSpec, operatorStatus)
```
p0lyn0mial, line 979: could we drop `requeue` param and requeue when `after.IsNotZero` ?

p0lyn0mial: it looks like `0` is a valid input to `AddAfter` it just means requeue immediately.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: done

PR #1146, opened 2021-07-22
staticpod/startupmonitor: update nodeStatus after readiness

sttts:
There must not be a different understanding what readiness mean between the operator and the startup-monitor. To avoid race, we move the readiness update of the nodeStatus to the startup-monitor.

### DIFF
--- a/pkg/operator/staticpod/startupmonitor/cmd.go
+++ b/pkg/operator/staticpod/startupmonitor/cmd.go
@@ -31,6 +31,9 @@ type Options struct {
 	// Revision identifier for this particular installation instance
 	Revision int
 
+	// NodeName as used to update the right nodeStatus struct in the static pod operator resource
+	NodeName string
+
 	// FallbackTimeout specifies a timeout after which the monitor starts the fall back procedure
 	FallbackTimeout time.Duration
 
@@ -83,7 +86,8 @@ func NewCommand(check ReadinessChecker) *cobra.Command {
 				withRevision(o.Revision).
 				withManifestPath(o.ManifestDir).
 				withStaticPodResourcesPath(o.ResourceDir).
-				withTargetName(o.TargetName)
+				withTargetName(o.TargetName).
+				withNodeName(o.NodeName)
 
 			if len(o.KubeConfig) > 0 {
 				clientConfig, err := client.GetKubeConfigOrInClusterConfig(o.KubeConfig, nil)
@@ -125,6 +129,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.ManifestDir, "manifests-dir", o.ManifestDir, "directory for the static pod manifest")
 	fs.StringVar(&o.TargetName, "target-name", o.TargetName, "identifies operand used to construct the final file name when reading the current and previous manifests")
 	fs.StringVar(&o.ResourceDir, "installer-lock-file", o.InstallerLockFile, "file path for the installer flock based lock file")
+	fs.StringVar(&o.NodeName, "node-name", o.NodeName, "the name of the node as used in the static pod operator resource")
 }
 
 func (o *Options) Validate() error {
@@ -139,6 +144,9 @@ func (o *Options) Validate() error {
 	}
 	if len(o.TargetName) == 0 {
 		return fmt.Errorf("--target-name is required")
+	}
+	if len(o.NodeName) == 0 {
+		return fmt.Errorf("--node-name is required")
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


pkg/operator/staticpod/startupmonitor/fallback.go
```diff
@@ -128,7 +136,49 @@ func (f *staticPodFallback) fallbackToPreviousRevision(reason, message string) e
...
+	}
+
+	// the startup-monitor is authorative to signal readiness of a revision. The operator
+	// is waiting for this to happen. Otherwise, the operator could assume readiness while
+	// the startup-monitor falls back, leading to an awkward situation.
+	// Note that this will retry forever, with a backoff.
+	return retry.OnError(retry.DefaultRetry, func(error) bool { return true }, func() error {
+		_, status, rv, err := f.operatorClient.GetStaticPodOperatorState()
+		if err != nil {
+			klog.Errorf("Failed to update static pod operator status.nodeStatus: %v", err)
```
p0lyn0mial, line 154: `Failed to get...`

pkg/operator/staticpod/startupmonitor/cmd.go
```diff
@@ -95,6 +100,12 @@ func NewCommand(check ReadinessChecker) *cobra.Command {
 				if c, ok := o.Check.(WantsRestConfig); ok {
 					c.SetRestConfig(restConfig)
 				}
+
+				operatorClient, err := newOperatorClient(restConfig)
```
p0lyn0mial, line 104: maybe at this point, we should make `o.KubeConfig` required?

pkg/operator/staticpod/startupmonitor/fallback.go
```diff
@@ -128,7 +136,49 @@ func (f *staticPodFallback) fallbackToPreviousRevision(reason, message string) e
...
+	if f.operatorClient == nil || len(f.nodeName) == 0 {
+		return nil
+	}
+
+	// the startup-monitor is authorative to signal readiness of a revision. The operator
+	// is waiting for this to happen. Otherwise, the operator could assume readiness while
+	// the startup-monitor falls back, leading to an awkward situation.
+	// Note that this will retry forever, with a backoff.
+	return retry.OnError(retry.DefaultRetry, func(error) bool { return true }, func() error {
+		_, status, rv, err := f.operatorClient.GetStaticPodOperatorState()
```
p0lyn0mial, line 152: will `status` come from a live GET or from the cache ?

pkg/operator/staticpod/startupmonitor/fallback.go
sttts: fixed

pkg/operator/staticpod/startupmonitor/cmd.go
sttts: done

pkg/operator/staticpod/startupmonitor/fallback.go
sttts: live get. It has to. Chance is high that it works on first try.

pkg/operator/staticpod/startupmonitor/cmd.go
```diff
@@ -79,11 +83,12 @@ func NewCommand(check ReadinessChecker) *cobra.Command {
...
-			fb := newFallback().
+			fb := newStaticPodFallback().
 				withRevision(o.Revision).
 				withManifestPath(o.ManifestDir).
 				withStaticPodResourcesPath(o.ResourceDir).
-				withTargetName(o.TargetName)
+				withTargetName(o.TargetName).
+				withNodeName(o.NodeName)
 
 			if len(o.KubeConfig) > 0 {
```
p0lyn0mial, line 93: since `KubeConfig` is required we don't need that `if` clause anymore

pkg/operator/staticpod/startupmonitor/fallback.go
```diff
@@ -128,7 +136,52 @@ func (f *staticPodFallback) fallbackToPreviousRevision(reason, message string) e
 }
 
 func (f *staticPodFallback) markRevisionGood() error {
-	return f.createLastKnowGoodRevisionFor(f.revision, true)
+	if err := f.createLastKnowGoodRevisionFor(f.revision, true); err != nil {
+		return err
+	}
+
+	if f.operatorClient == nil || len(f.nodeName) == 0 {
```
p0lyn0mial, line 143: since both, kubeconfig and the node name are required we could get rid of this as well

pkg/operator/staticpod/startupmonitor/cmd.go
sttts: fixed

sttts: But there could be a in-cluster config. Either has to work. So kubeconfig is not required if in-cluster config is provided.

pkg/operator/staticpod/startupmonitor/fallback.go
sttts: fixed

PR #1150, opened 2021-07-23
staticpod/installer: fix backoff of installers

sttts:
The DeepEqual clause for updating the operator status returned from `Sync`, after `lastFailedCount` had been increased. On next `Sync` we ran directly into new state computation and installer pod creation. The delay much further down was never effective. This PR moved the installer backoff delay up.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -88,8 +88,9 @@ type InstallerController struct {
 
 	installerPodMutationFns []InstallerPodMutationFunc
 
-	factory *factory.Factory
-	clock   clock.Clock
+	factory          *factory.Factory
+	clock            clock.Clock
+	installerBackOff func(count int) time.Duration
 }
 
 // InstallerPodMutationFunc is a function that has a chance at changing the installer pod before it is created
@@ -160,6 +161,7 @@ func NewInstallerController(
 
 		installerPodImageFn: getInstallerPodImageFromEnv,
 		clock:               clock.RealClock{},
+		installerBackOff:    backOffDuration(10*time.Second, 1.5, 10*time.Minute),
 	}
 
 	c.ownerRefsFn = c.setOwnerRefs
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -372,6 +374,15 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 
 		// if we are in a transition, check to see whether our installer pod completed
 		if currNodeState.TargetRevision > currNodeState.CurrentRevision {
+			if currNodeState.LastFailedRevision == currNodeState.TargetRevision && currNodeState.LastFailedTime != nil && !currNodeState.LastFailedTime.IsZero() {
+				delay := c.installerBackOff(currNodeState.LastFailedCount)
+				earliestRetry := currNodeState.LastFailedTime.Add(delay)
+				if !c.now().After(earliestRetry) {
+					klog.V(4).Infof("Backing off node %s installer retry %d until %v", currNodeState.NodeName, currNodeState.LastFailedCount+1, earliestRetry)
+					return true, earliestRetry.Sub(c.now()), nil
```
aojea, line 382: so:
1. we obtain the delay based on the backoff function and the  `currNodeState.LastFailedCount`
2. we don't retry until delay sec after `currNodeState.LastFailedTime`

my questions is what happens if current clock and `currNodeState.LastFailedTime`  are not in sync, i.e. edge case, controller restarts and node clock goes crazy and goes back in time or DST change, , ...

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -160,6 +161,7 @@ func NewInstallerController(
 
 		installerPodImageFn: getInstallerPodImageFromEnv,
 		clock:               clock.RealClock{},
+		installerBackOff:    backOffDuration(10*time.Second, 1.5, 10*time.Minute),
```
p0lyn0mial, line 164: it means we will retry ~10 times, which seems reasonable

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: we don't support that case. Clocks not in sync break OpenShift and Kubernetes.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: Yes. Unchanged btw.

We start with 10, then 15, then 22.5 up to 10min.

sttts:
Unit test flake. TestClusterOperatorLoggingController

/retest

PR #1151, opened 2021-07-23
staticpod/installer: simplify state machine by removing fall-through

sttts:


pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -1127,6 +1127,8 @@ func TestCreateInstallerPodMultiNode(t *testing.T) {
 					NodeName:           "test-node-1",
 					CurrentRevision:    1,
 					LastFailedRevision: 2,
+					LastFailedCount:    1,
```
sttts, line 1130: we didn't have `lastFailedCount` in 4.6 and older. But every cluster has been upgraded many times since then, so we can expect to have it.

sttts: Compare the added test below which is the same as this (modified) one, but without count and revision. So this is NOT a test change, but I am extending coverage.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -433,14 +433,19 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
...
+				NodeName:        currNodeState.NodeName,
+				CurrentRevision: currNodeState.CurrentRevision,
+				TargetRevision:  revisionToStart,
+			}
+		} else {
+			newCurrNodeState = currNodeState.DeepCopy()
+			newCurrNodeState.TargetRevision = revisionToStart
+			if newCurrNodeState.LastFailedCount == 0 {
+				newCurrNodeState.LastFailedCount = 1
+			}
```
sttts: this logic is questionable, probably wrong. The following commit get rid of it. Here, I only rewrite.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -646,6 +639,7 @@ func (c *InstallerController) newNodeStateForInstallInProgress(ctx context.Conte
 		return &operatorv1.NodeStatus{
 			NodeName:        currNodeState.NodeName,
 			CurrentRevision: currNodeState.CurrentRevision,
+			TargetRevision:  latestRevisionAvailable,
```
sttts, line 642: this line came from the former fall-through in line 423.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -687,16 +681,14 @@ func (c *InstallerController) newNodeStateForInstallInProgress(ctx context.Conte
...
-			ret.TargetRevision = 0
-			ret.LastFailedRevision = 0
-			ret.LastFailedTime = nil
-			ret.LastFailedCount = 0
-			ret.LastFailedRevisionErrors = nil
-			return ret, false, staticPodReason, nil
+			return &operatorv1.NodeStatus{
+				NodeName:        currNodeState.NodeName,
+				CurrentRevision: rev,
+			}, false, staticPodReason, nil
```
sttts, line 691: @p0lyn0mial this is the code that will be done by the startup-monitor in SNO (next PR).

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -317,7 +317,7 @@ func TestNewNodeStateForInstallInProgress(t *testing.T) {
 			&operatorv1.NodeStatus{
 				NodeName:        "test-node-1",
 				CurrentRevision: 0,
-				TargetRevision:  0,
+				TargetRevision:  2,
```
sttts, line 320: this comes from https://github.com/openshift/library-go/pull/1151/commits/db266aea60807d3b8678f469dfbcc2a04a988d2f#r675510545 because now the node state func sets target revision when there is a revision pending. Before the sync func with the fallthrough did this and we stayed at zero target version here in the tests.

Same for all 4 cases here and below.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -414,17 +417,14 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
...
 				return false, 0, nil // no requeue because UpdateStaticPodStatus triggers an external event anyway
-			} else {
-				klog.V(2).Infof("%q is in transition to %d, but has not made progress because %s", currNodeState.NodeName, currNodeState.TargetRevision, reasonWithBlame(reason))
 			}
 
-			// We want to retry the installer pod by deleting and then rekicking. Also we don't set LastFailedRevision.
-			if !installerPodFailed {
-				break
-			}
-			klog.Infof("Will retry %q for revision %d for the %s time because %s", currNodeState.NodeName, currNodeState.TargetRevision, nthTimeOr1st(newCurrNodeState.LastFailedCount), reason)
```
sttts, line 425: this is the fall-through the commit is about. After this PR we return.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -608,54 +604,52 @@ func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1
...
+
 	installerPodName := getInstallerPodName(currNodeState.TargetRevision, currNodeState.NodeName, previouslyFailedLatestRevisionPods)
 	installerPod, err := c.podsGetter.Pods(c.targetNamespace).Get(ctx, installerPodName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		// installer pod has disappeared before we saw it's termination state. Retry like if it had never existed.
-		c.eventRecorder.Warning("InstallerPodDisappeared", err.Error())
-		return currNodeState, true, fmt.Sprintf("pod %s disappeared", installerPodName), nil
-	}
-	if err != nil {
+		if !pendingNewRevision {
```
p0lyn0mial, line 618: ok, if a pod disappeared and there is a new rev just continue

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -608,54 +604,52 @@ func setAvailableProgressingNodeInstallerFailingConditions(newStatus *operatorv1
...
 		}
 
-		ret.LastFailedRevision = 0
-		ret.LastFailedTime = nil
-		ret.LastFailedCount = 0
-		ret.TargetRevision = 0
-		ret.LastFailedRevisionErrors = nil
-
-		return ret, false, "new revision pending", nil
+		return &operatorv1.NodeStatus{
```
p0lyn0mial, line 639: yeah, this will update the spec.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -433,14 +433,10 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
...
 
-		newCurrNodeState := currNodeState.DeepCopy()
-		newCurrNodeState.TargetRevision = revisionToStart
-		if newCurrNodeState.LastFailedRevision != revisionToStart {
-			newCurrNodeState.LastFailedRevisionErrors = nil
-			newCurrNodeState.LastFailedCount = 0
-			newCurrNodeState.LastFailedTime = nil
-		} else if newCurrNodeState.LastFailedCount == 0 {
-			newCurrNodeState.LastFailedCount = 1
+		newCurrNodeState := &operatorv1.NodeStatus{
```
p0lyn0mial, line 436: is it okay to clean `LastFailedRevisionErrors` because a failed revision can never reach this code?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -396,6 +396,9 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 			if err != nil {
 				return true, 0, err
 			}
+			if installerPodFailed {
+				klog.Infof("Will retry %q for revision %d for the %s time because %s", currNodeState.NodeName, currNodeState.TargetRevision, nthTimeOr1st(newCurrNodeState.LastFailedCount), reason)
```
p0lyn0mial, line 400: except when we fail to update the spec a few lines down :)

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: yes. It's what we always did.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: Well, the will is still correct.

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: The revision is a new one. We already wiped out all lastFailed* fields before, part of the API.

PR #1157, opened 2021-07-23
adds startupMonitorPodConditionController for monitoring the lifecycle of the startup monitor pod

p0lyn0mial:
xref: openshift/enhancements#833

### DIFF
--- a/pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
+++ b/pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
@@ -1,0 +1,151 @@
+package startupmonitorcondition
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+// startupMonitorPodConditionController monitors the startup monitor pod and reports unwanted state/condition/phase as a degraded condition
+type startupMonitorPodConditionController struct {
+	operatorClient operatorv1helpers.OperatorClient
+
+	targetName string
+	podLister  corev1listers.PodNamespaceLister
+
+	startupMonitorEnabledFn func() bool
+}
+
+// New returns a controller for monitoring the lifecycle of the startup monitor pod
+func New(targetNamespace string,
+	targetName string,
+	operatorClient operatorv1helpers.OperatorClient,
+	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
+	startupMonitorEnabledFn func() bool,
+	eventRecorder events.Recorder) factory.Controller {
+	fd := &startupMonitorPodConditionController{
+		operatorClient:          operatorClient,
+		targetName:              targetName,
+		podLister:               kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Pods().Lister().Pods(targetNamespace),
+		startupMonitorEnabledFn: startupMonitorEnabledFn,
+	}
+	return factory.New().WithSync(fd.sync).ResyncEvery(6*time.Minute).WithInformers(kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Pods().Informer()).ToController("StartupMonitorPodCondition", eventRecorder)
+}
+
+func (fd *startupMonitorPodConditionController) sync(_ context.Context, _ factory.SyncContext) (err error) {
+	startupPodDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodDegraded", Status: operatorv1.ConditionFalse}
+	startupPodContainerExcessiveRestartsDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerExcessiveRestartsDegraded", Status: operatorv1.ConditionFalse}
+
+	defer func() {
+		if err == nil {
+			if _, _, updateError := operatorv1helpers.UpdateStatus(fd.operatorClient,
+				operatorv1helpers.UpdateConditionFn(*startupPodDegraded),
+				operatorv1helpers.UpdateConditionFn(*startupPodContainerExcessiveRestartsDegraded)); updateError != nil {
+				err = updateError
+			}
+		}
+	}()
+
+	// in practice we rely on operators to provide
+	// a condition for checking we are running on a single node cluster
+	if !fd.startupMonitorEnabledFn() {
+		return nil
+	}
+
+	monitorPod, err := fd.podLister.Get(fmt.Sprintf("%s-startup-monitor", fd.targetName))
+	if errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err // unknown error
+	}
+
+	// sets the degraded condition when the pod has been in pending for more than 2 minutes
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,150 @@
...
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	corev1listers "k8s.io/client-go/listers/core/v1"
```
sttts, line 15: order

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,150 @@
...
+		startupMonitorEnabledFn: startupMonitorEnabledFn,
+	}
+	return factory.New().WithSync(fd.sync).ResyncEvery(6*time.Minute).WithInformers(kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Pods().Informer()).ToController("StartupMonitorPodCondition", eventRecorder)
+}
+
+func (fd *startupMonitorPodConditionController) sync(_ context.Context, _ factory.SyncContext) (err error) {
+	startupPodPendingDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodPendingDegraded", Status: operatorv1.ConditionFalse}
+	startupPodContainerWaitingDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerWaitingDegraded", Status: operatorv1.ConditionFalse}
+	startupPodFailedContainerDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodFailedContainerDegraded", Status: operatorv1.ConditionFalse}
+	startupPodContainerExcessiveRestartsDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerExcessiveRestartsDegraded", Status: operatorv1.ConditionFalse}
```
sttts, line 48: these are many conditions. I don't think we want one for each symptom, but rather use the reason and message to distinguish.

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (fd *startupMonitorPodConditionController) sync(_ context.Context, _ factory.SyncContext) (err error) {
+	startupPodDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodDegraded", Status: operatorv1.ConditionFalse}
+	startupPodContainerExcessiveRestartsDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerExcessiveRestartsDegraded", Status: operatorv1.ConditionFalse}
+
+	defer func() {
+		if err == nil {
+			if _, _, updateError := operatorv1helpers.UpdateStatus(fd.operatorClient,
+				operatorv1helpers.UpdateConditionFn(*startupPodDegraded),
+				operatorv1helpers.UpdateConditionFn(*startupPodContainerExcessiveRestartsDegraded)); updateError != nil {
```
sttts, line 52: do we have a variant of `UpdateConditionFn` that also remove the condition if it is not needed?

sttts: i.e. if `startupMonitorEnabledFn()` is false

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,146 @@
...
+
+	withDegradedWhenInPendingPhaseFor(monitorPod, 2*time.Minute, startupPodDegraded)
+	withDegradedWhenHasFailedContainer(monitorPod, startupPodDegraded)
+	withDegradedWhenHasExcessiveRestarts(monitorPod, 2, startupPodContainerExcessiveRestartsDegraded)
+
+	return nil
+}
+
+func withDegradedWhenInPendingPhaseFor(monitorPod *corev1.Pod, maxPendingDuration time.Duration, startupPodDegraded *operatorv1.OperatorCondition) {
+	if monitorPod.Status.Phase != corev1.PodPending || monitorPod.Status.StartTime == nil || time.Now().Sub(monitorPod.Status.StartTime.Time) < maxPendingDuration {
```
sttts, line 79: nit add a oneliner comment above each block to make reading the flow easy

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,146 @@
...
+	startupPodDegraded.Status = operatorv1.ConditionTrue
+	pendingPodReason, pendingPodMessage := "Unknown", "unknown"
+	if len(monitorPod.Status.Reason) > 0 {
+		pendingPodReason = monitorPod.Status.Reason
+	}
+	if len(monitorPod.Status.Message) > 0 {
+		pendingPodMessage = monitorPod.Status.Message
+	}
+	startupPodDegraded.Reason = pendingPodReason
+	startupPodDegraded.Message = fmt.Sprintf("the pod %v has been in %v phase for %v due to %v, the max tolerated value is %v", monitorPod.Name, corev1.PodPending, time.Now().Sub(monitorPod.Status.StartTime.Time), pendingPodMessage, maxPendingDuration)
```
sttts, line 92: don't use durations in messages. Messages must be constant.

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -0,0 +1,146 @@
...
+	monitorPod, err := fd.podLister.Get(fmt.Sprintf("%s-startup-monitor", fd.targetName))
+	if errors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return err // unknown error
+	}
+
+	withDegradedWhenInPendingPhaseFor(monitorPod, 2*time.Minute, startupPodDegraded)
+	withDegradedWhenHasFailedContainer(monitorPod, startupPodDegraded)
+	withDegradedWhenHasExcessiveRestarts(monitorPod, 2, startupPodContainerExcessiveRestartsDegraded)
```
aojea, line 73: is it possible that `withDegradedWhenHasExcessiveRestarts` shadows some of the previous?

aojea: I mean, should any of `withDegradedWhen*`return immediately once if finds a problem?

p0lyn0mial: I don't think so, hence a new condition. The previous condition will be set when a pod failed with a non-zero exit code or has been pending for more than 2 minutes.

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
p0lyn0mial: I am not aware of any.

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
aojea: actually, the time in the message will be different than the one used in the `if`

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
p0lyn0mial: done

PR #1165, opened 2021-07-28
startup-monitor: always try to delete the symlink before creating it

p0lyn0mial:
the existing `fileExists` method checks only the target file and creating a symlink fails when it already exists

### DIFF
--- a/pkg/operator/staticpod/startupmonitor/fallback.go
+++ b/pkg/operator/staticpod/startupmonitor/fallback.go
@@ -62,7 +62,8 @@ func (f *staticPodFallback) fallbackToPreviousRevision(reason, message string) e
 	// step 0: if the last known good revision doesn't exist
 	//         find latest previous revision to work with
 	//         return in case no revision has been found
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/startupmonitor/fallback.go
```diff
@@ -182,24 +183,20 @@ func (f *staticPodFallback) markRevisionGood(ctx context.Context) error {
...
-	if strict {
-		if exists, err := f.fileExists(f.lastKnownGoodManifestDstPath()); err != nil {
-			return err
-		} else if exists {
-			if err := f.io.Remove(f.lastKnownGoodManifestDstPath()); err != nil {
-				return err
-			}
-			klog.Info("Removed existing last known good revision manifest %s", f.lastKnownGoodManifestDstPath())
-		}
+	// step 0: always try to delete the symlink otherwise creation will fail
```
p0lyn0mial, line 189: it should not happen but I think it is better to make progress . . .

PR #1166, opened 2021-07-28
startup-monitor: releases the lock on each iteration to prevent fight with the installer

p0lyn0mial:
additionally, it doesn't fall back when the target revision doesn't match the expected one.
this usually means something bad happened, in that case, we should report an error (the startup controller will put an operator into a degraded state)

### DIFF
--- a/pkg/operator/staticpod/startupmonitor/monitor.go
+++ b/pkg/operator/staticpod/startupmonitor/monitor.go
@@ -106,23 +106,29 @@ func (m *monitor) Run(ctx context.Context, installerLock Locker) (ready bool, re
 	// we use wait.Until because it uses sliding interval, wait.Poll does not.
 	withTimeoutCtx, cancel := context.WithTimeout(ctx, m.timeout-m.probeInterval) // last iteration is done manually to have control over the Unlock
 	wait.UntilWithContext(withTimeoutCtx, func(ctx context.Context) {
-		defer installerLock.Unlock()
+		defer func() {
+			if !lastReady {
+				installerLock.Unlock()
+			}
+		}()
 		iteration(ctx)
 		if lastReady {
 			cancel()
 		}
 	}, m.probeInterval)
 
+	// keep the lock (for suicide) and return
+	if lastReady {
+		return true, "", "", nil
+	}
+
+	// at this point we are either not ready or we got an error
 	// do last iteration without calling Unlock()
 	iteration(ctx)
 	if lastError != nil {
 		// release the lock since we are exiting anyway
 		installerLock.Unlock()
 		return false, lastReason, lastMessage, lastError
-	}
-
-	if lastReady {
-		return true, "", "", nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/startupmonitor/monitor.go
```diff
@@ -106,18 +106,22 @@ func (m *monitor) Run(ctx context.Context, installerLock Locker) (ready bool, re
 	// we use wait.Until because it uses sliding interval, wait.Poll does not.
 	withTimeoutCtx, cancel := context.WithTimeout(ctx, m.timeout-m.probeInterval) // last iteration is done manually to have control over the Unlock
 	wait.UntilWithContext(withTimeoutCtx, func(ctx context.Context) {
-		defer func() {
-			if lastReady {
-				installerLock.Unlock()
-			}
-		}()
+		defer installerLock.Unlock()
```
sttts, line 109: no, this breaks the contract.

sttts: Only on error or timeout we are allowed to release the lock. If the target is ready here in the loop, we have to keep the lock kept.

pkg/operator/staticpod/startupmonitor/monitor.go
```diff
@@ -106,18 +106,22 @@ func (m *monitor) Run(ctx context.Context, installerLock Locker) (ready bool, re
...
+	// do last iteration without calling Unlock()
+	// note that you cannot logically have lastReady = true and lastError != nil
 	if !lastReady {
-		iteration(ctx) // do last iteration without Unlock
+		iteration(ctx)
+		if lastError != nil {
+			// release the lock since we are exiting anyway
+			installerLock.Unlock()
+			return false, lastReason, lastMessage, lastError
+		}
```
sttts, line 124: this case was missing :+1:

Can we test it?

pkg/operator/staticpod/startupmonitor/monitor.go
```diff
@@ -143,7 +147,8 @@ func (m *monitor) isReady(ctx context.Context) (ready bool, reason string, messa
 		return false, "", "", err
 	}
 	if m.revision != currentTargetRevision {
-		return false, "UnexpectedRevision", fmt.Sprintf("expected revision %d, found %d", m.revision, currentTargetRevision), nil
+		// return an error so that we won't fallback when we fail to monitor unexpected target
+		return false, "", "", fmt.Errorf("expected revision %d, found %d", m.revision, currentTargetRevision)
```
sttts, line 151: good point

pkg/operator/staticpod/startupmonitor/monitor.go
p0lyn0mial: fixed.

pkg/operator/staticpod/startupmonitor/monitor.go
p0lyn0mial: added unit tests

PR #1167, opened 2021-07-29
staticpod/prune: replace status state machine w/ n revs before current/target/failed/latest

sttts:
The origin state machine was complicated, had races, spread over 3 controller, and can be – without loss of value – replaced with this approximation:

- prune everything before the last excluded revisions while excluded means:
  - `failedLimit` many revisions before every targetRevision
  - `max(failedLimit,succeededLimit)` many revisions before every lastAvailableRevisions
  - `failedLimit` many revisions before every lastFailedRevisions
  - `succeddedLimit` many revisions before every currentRevision

PR #1170, opened 2021-07-30
staticpod/installer: do not wipe nodeStatus.lastFailed* when going ready

sttts:
This makes https://github.com/openshift/library-go/pull/1167 approximation of the failedLimit better.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -470,7 +470,7 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 				}
 			}
 
-			if err := c.ensureInstallerPod(ctx, currNodeState.NodeName, operatorSpec, currNodeState.TargetRevision, currNodeState.LastFailedCount+currNodeState.LastFallbackCount); err != nil {
+			if err := c.ensureInstallerPod(ctx, operatorSpec, currNodeState); err != nil {
 				c.eventRecorder.Warningf("InstallerPodFailed", "Failed to create installer pod for revision %d count %d on node %q: %v",
 					currNodeState.TargetRevision, currNodeState.NodeName, currNodeState.LastFailedCount, err)
 				// if a newer revision is pending, continue, so we retry later with the latest available revision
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -745,10 +735,10 @@ func (c *InstallerController) newNodeStateForInstallInProgress(ctx context.Conte
 				if currNodeState.TargetRevision > currNodeState.CurrentRevision {
 					rev = currNodeState.TargetRevision
 				}
-				return &operatorv1.NodeStatus{
-					NodeName:        currNodeState.NodeName,
-					CurrentRevision: rev,
-				}, false, staticPodReason, nil
+				ret := currNodeState.DeepCopy()
+				ret.CurrentRevision = rev
+				ret.TargetRevision = 0
```
p0lyn0mial, line 740: ok, since the static pod is ready.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -717,24 +711,20 @@ func (c *InstallerController) newNodeStateForInstallInProgress(ctx context.Conte
...
-				// increase failed count, but not in startup-monitor case where this is done when the fallback pod is observed below in staticPodStateFallback,
-				ret.LastFailedRevision = currNodeState.TargetRevision
-				ret.LastFailedReason = nodeStatusOperandFailedReason
-				if ret.LastFailedTime == nil || !ret.LastFailedTime.Time.Equal(ts) {
-					// avoid increasing the count continuously by comparing the fail time
-					failTime := metav1.NewTime(ts)
-					ret.LastFailedTime = &failTime
-					ret.LastFailedCount++
-				}
+				return currNodeState, false, fmt.Sprintf("operand pod failed: %v", strings.Join(staticPodErrors, "\n")), nil
```
p0lyn0mial, line 715: I think we miss `staticPodErrors`, previously `ret.LastFailedRevisionErrors = staticPodErrors`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1091,6 +1086,17 @@ func (c InstallerController) Sync(ctx context.Context, syncCtx factory.SyncConte
 	return err
 }
 
+func deepCopyNodeStatusWithoutOldFailedState(ns *operatorv1.NodeStatus) *operatorv1.NodeStatus {
+	if ns.TargetRevision == 0 || ns.TargetRevision != ns.LastFailedRevision {
+		return &operatorv1.NodeStatus{
```
p0lyn0mial, line 1091: are we going to overwrite the failed state in this case?

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: intentional. Setting the error without the count is odd, no?

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: yes, that the "we don't track history of fails" part. I call this func when we set a new set of `lastFailed*` fields.

PR #1181, opened 2021-08-16
[wip] pkg/operator/staticpod: stop setting static pod UID

hexfusion:
The kubelet has fixed its handling of graceful termination in static pods in 1.22. The existence of the UID can result in an unexpected failure case if the static pod spec is mutated but the UID remains unchanged. A possible resolution for this corner case is to stop setting the UID for static pods. This PR explores that scenario.

Manual revert of changes introduced in https://github.com/openshift/library-go/pull/920

/hold

Signed-off-by: Sam Batschelet <sbatsche@redhat.com>


hexfusion:
proof https://github.com/openshift/cluster-etcd-operator/pull/641

PR #1182, opened 2021-08-17
deflake TestStaticPodFallbackConditionController

aojea:
The order of pods yields different results, however, the test only
look for one possible outcome.
Instead of complicating the test to deal with all the different
possible returns, we can use a custom podlister that returns
the pods in a deterministic order.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

Tested with 
```
go test -v ./pkg/operator/staticpod/controller/staticpodfallback -run TestStaticPodFallbackConditionController -count 1 -c
$ stress ./staticpodfallback.test 
5s: 4406 runs so far, 0 failures
10s: 8767 runs so far, 0 failures
```

### DIFF
--- a/pkg/operator/staticpod/controller/staticpodfallback/static_pod_fallback_condition_test.go
+++ b/pkg/operator/staticpod/controller/staticpodfallback/static_pod_fallback_condition_test.go
@@ -2,6 +2,7 @@ package staticpodfallback
 
 import (
 	"fmt"
+	"sort"
 	"testing"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/staticpodfallback/static_pod_fallback_condition_test.go
```diff
@@ -173,3 +175,38 @@ func newPod(phase corev1.PodPhase, ready corev1.ConditionStatus, revision, name
 
 	return &pod
 }
+
+// we need a deterministic pod lister for tests
+// otherwise the order of the pods will yield different results
+type orderedPodNamespaceLister struct {
+	podLister corev1listers.PodNamespaceLister
```
sttts, line 182: why not deeply nest without name? Then you don't need the Get func

sttts:
/lgtm
/approve

This not only applies to the unit test, but potentially also stops condition flapping in a real cluster.

pkg/operator/staticpod/controller/staticpodfallback/static_pod_fallback_condition_test.go
aojea: First interface ive seen :)

aojea:
> This not only applies to the unit test, but potentially also stops condition flapping in a real cluster.

Should I add it to the real controller too?

PR #1184, opened 2021-08-23
golang serving and client trust behavior

deads2k:


pkg/crypto/cert_accept_reject_test.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// strip the intermediate signers which are included
+	servingCertConfig.Certs = servingCertConfig.Certs[:1]
+	servingPEMCertByte, servingPEMKeyBytes, err := servingCertConfig.GetPEMBytes()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Log(string(servingPEMCertByte))
```
slaskawi, line 72: We should probably remove this or print other certs out as well. I'm fine with either way.

s-urbaniak, line 72: leftover?

slaskawi (approved):
/lgtm

PR #1185, opened 2021-08-23
encryption test should not give up early

tkashem:


tkashem:
/hold

this is a hack and for test purpose only

test/library/encryption/helpers.go
```diff
@@ -162,6 +162,13 @@ func WaitForNextMigratedKey(t testing.TB, kubeClient kubernetes.Interface, prevK
 	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
 		currentKeyMeta, err := GetLastKeyMeta(kubeClient, namespace, labelSelector)
```
p0lyn0mial, line 163: why could extend the timeout passed to the `onErrorWithTimeout` function (inside `GetLastKeyMeta`) - it already understands 429. I think the max downtime on an SNO cluster is 60s. So we could set it to 120s

`err := onErrorWithTimeout(wait.ForeverTestTimeout, retry.DefaultBackoff, transientAPIError, func() (err error) {`

PR #1186, opened 2021-08-24
Improve zombie process detection performance

saschagrunert:
We now rely on `/proc/[pid]/stat` for parsing the state of a process
rather than the line-by-line parsing of `/proc/[pid]/status`. Tests have
been added as well.

cc @haircommander @mrunalp


saschagrunert:
PTAL @smarterclayton @deads2k @sttts

PR #1187, opened 2021-08-24
Merge ownerrefs in mergeObject

qiujian16:


pkg/operator/resource/resourcemerge/object_merger.go
```diff
@@ -187,3 +189,51 @@ func SetMapStringStringIfSet(modified *bool, existing *map[string]string, requir
...
+func MergeOwnerRefs(modified *bool, existing *[]metav1.OwnerReference, required []metav1.OwnerReference) {
+	if *existing == nil {
+		*existing = []metav1.OwnerReference{}
+	}
+
+	for _, o := range required {
+		actualUID := o.UID
+		removeOwner := false
+
+		// if "required" ownerRefss contains an owner.UID with "-" as suffix, remove that
```
deads2k, line 202: nit ownerRefs

pkg/operator/resource/resourcemerge/object_merger.go
```diff
@@ -187,3 +189,51 @@ func SetMapStringStringIfSet(modified *bool, existing *map[string]string, requir
...
+		// ownerRef from the existing ownerRefs instead of replacing the value
```
deads2k, line 203: note that this is the same format as `kubectl annotate` and `kubectl label`

pkg/operator/resource/resourcemerge/object_merger.go
```diff
@@ -187,3 +189,51 @@ func SetMapStringStringIfSet(modified *bool, existing *map[string]string, requir
...
+		if strings.HasSuffix(string(actualUID), "-") {
+			removeOwner = true
+			actualUID = types.UID(strings.TrimRight(string(actualUID), "-"))
+		}
+
+		existedIndex := 0
+
+		for existedIndex < len(*existing) {
+			if actualUID == (*existing)[existedIndex].UID {
```
deads2k, line 212: I think I prefer matching on apiversion (the group part only), kind, name.  That allows removal of an ownerreference without knowing the original UID.

PR #1192, opened 2021-08-26
csi: Replace leader-election parameters

jsafrane:
CSI controller pods contain many sidecars, which run leader election with the default setting (timeout after 15 seconds). Replace
`${LEADER_ELECTION_LEASE_DURATION}`, `${LEADER_ELECTION_RENEW_DEADLINE}` and `${LEADER_ELECTION_RETRY_PERIOD}` with correct OpenShift's values to make the sidecars resilient to API server disruption during update.

It's up to an operator to put these placeholders to their yaml files to benefit from this replacement.

This is a prerequisite for https://bugzilla.redhat.com/show_bug.cgi?id=1990662 - we need to update leader election in all CSI driver operators.

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
@@ -1,6 +1,8 @@
 package csidrivercontrollerservicecontroller
 
 import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/config/leaderelection"
 	appsinformersv1 "k8s.io/client-go/informers/apps/v1"
 	"k8s.io/client-go/kubernetes"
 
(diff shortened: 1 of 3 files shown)


pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -66,6 +72,12 @@ func NewCSIDriverControllerServiceController(
 	optionalInformers = append(optionalInformers, configInformer.Config().V1().Infrastructures().Informer())
 	var optionalManifestHooks []dc.ManifestHookFunc
 	optionalManifestHooks = append(optionalManifestHooks, WithPlaceholdersHook(configInformer))
+	optionalManifestHooks = append(optionalManifestHooks,
+		WithLeaderElectionReplacerHook(
+			leaderelection.LeaderElectionDefaulting(
```
sttts, line 77: ugly indention, hard to read

sttts: if you moved it into a local variable `libraryGoLeaderElectionsDefaults` is would be trivial to read. Now I had to check two functions to understand.

jsafrane: Reworked to a variable

PR #1196, opened 2021-09-01
startup-monitor: allows for logging to a log file in addition to stderr

p0lyn0mial:
since kubelet doesn't respect the restart policy for static pods
we need to store the logs in a file for future troubleshooting

this PR introduces a new flag that allows specifying a log file.
the file will be rotated on a monthly basis and we will retain only the last 3 files.

### DIFF
--- a/pkg/operator/staticpod/startupmonitor/cmd.go
+++ b/pkg/operator/staticpod/startupmonitor/cmd.go
@@ -10,6 +10,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"gopkg.in/natefinch/lumberjack.v2"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
@@ -59,6 +60,9 @@ type Options struct {
 	// every iteration of the probe interval with this lock taken.
 	InstallerLockFile string
 
+	// LogFile is the file the logs are written.
+	LogFile string
+
 	// Check is the readiness step.
 	Check ReadinessChecker
 }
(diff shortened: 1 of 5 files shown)


p0lyn0mial:
proof PR https://github.com/openshift/cluster-kube-apiserver-operator/pull/1219

pkg/operator/staticpod/startupmonitor/cmd.go
```diff
@@ -165,6 +173,19 @@ func (o *Options) Validate() error {
...
+func setupLogger(logFilePath string) {
+	if len(logFilePath) == 0 {
+		return
+	}
+	klog.SetOutput(&lumberjack.Logger{
+		Filename:   logFilePath,
+		MaxSize:    100, // large files break some editors
+		MaxBackups: 3,
+		MaxAge:     28,   // if something goes wrong we will be called out early, retain logs for ~ a month, just in case
+		Compress:   true, // determines if the rotated log files should be compressed (i.e. gzip)
```
sttts, line 185: do these setting match those of kube-apiserver?

p0lyn0mial: except the `Compress` field. `watch-termination` binary [uses](https://github.com/openshift/kubernetes/blob/master/cmd/watch-termination/main.go#L255):
```
l := &lumberjack.Logger{
			Filename:   w.fn,
			MaxSize:    100,
			MaxBackups: 3,
			MaxAge:     28,
			Compress:   false,
		}
```

p0lyn0mial: it means that the rotated logs files will be compressed, which should be fine, right ?

sttts: would not compress. Not super important. But 3x100 is not a lot. Let's just leave it uncompressed.

p0lyn0mial: updated to reflect what we use in `watch-termination`

PR #391, opened 2019-05-03
installerpod: save secrets 0600

sttts:
Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/455

PR #1202, opened 2021-09-08
Bug 1977730: change write storage chmod

EmilyM1:
This is for a Bug 1977730. Cert were being given 600 here, but Kubernetes prefers 644.
https://kubernetes.io/docs/concepts/configuration/secret/#secret-files-permissions

### DIFF
--- a/pkg/operator/staticpod/certsyncpod/certsync_controller.go
+++ b/pkg/operator/staticpod/certsyncpod/certsync_controller.go
@@ -264,7 +264,7 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
 			}
 
 			klog.Infof("Writing secret manifest %q ...", fullFilename)
-			if err := staticpod.WriteFileAtomic(content, 0644, fullFilename); err != nil {
+			if err := staticpod.WriteFileAtomic(content, 0600, fullFilename); err != nil {
 				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed writing file for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
 				errors = append(errors, err)
 				continue


deads2k:
this was done in response to a customer comment/request.  Is something functionally broken here?

/hold

EmilyM1:
@deads2k. Yes - multiple customer ask.

s-urbaniak:
/retitle Bug 1977730: change write storage chmod

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -448,7 +448,7 @@ func writeConfig(content []byte, fullFilename string) error {
 func writeSecret(content []byte, fullFilename string) error {
 	klog.Infof("Writing secret manifest %q ...", fullFilename)
 
-	filePerms := os.FileMode(0600)
```
deads2k, line 451: I don't think this should change.  I'm pretty sure I explicitly made this 0600.

EmilyM1: @deads2k Git blame says it's Polynomial that made it 600 on 6/8/21

s-urbaniak: @EmilyM1 hmm, I did some more history digging and found the original change being done here: https://github.com/openshift/library-go/pull/391/files. So this has already been changed from 0644 to 0600 to limit read/write to the root user.

Looking at https://bugzilla.redhat.com/show_bug.cgi?id=1977730 I think there is a different ask: `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/user-serving-cert-000/tls.crt` and `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/user-serving-cert-000/tls.key` currently have 0644 permissions, the expected permissions should be 0600.

I am not sure if those assets are handled in this `writeSecret` function as well, but if not probably a different spot needs a fix to set permissions to 0600 on user-provided serving certs and keys.

stlaz: I don't see why we would want 0644, definitely not for TLS keys.

EmilyM1: @s-urbaniak  @stlaz:  when keys are placed in a secret yaml -- isn't it actually k8's itself that sets the permissions on this file?
I am no longer sure if this is something we can fix.
The closest I could see of a way to change it was our writeSecret func.

EmilyM1: Updated to 0600 per discussions

deads2k:
this looks like a change in the right direction

/hold cancel

EmilyM1:
@deads2k can we tag/approve it?

PR #1203, opened 2021-09-09
Bug 2001856: Avoid spurious MissingVersion events on the static pod controller

aojea:
If the controller can't get a Pod from the API, typically because of
network problems of apiserver unavailable, we can't differentiate
if there are no images on the pods or if we couldn't  obtain them.

Only generate the event if we are sure the image was missing.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

### DIFF
--- a/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
+++ b/pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
@@ -8,6 +8,7 @@ import (
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
@@ -80,13 +81,17 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
 	errs := []error{}
 	failingErrorCount := 0
 	images := sets.NewString()
+	podsFound := false
 	for _, node := range originalOperatorStatus.NodeStatuses {
 		pod, err := c.podsGetter.Pods(c.targetNamespace).Get(ctx, mirrorPodNameForNode(c.staticPodName, node.NodeName), metav1.GetOptions{})
 		if err != nil {
-			errs = append(errs, err)
-			failingErrorCount++
+			if !apierrors.IsNotFound(err) {
+				errs = append(errs, err)
+				failingErrorCount++
+			}
 			continue
 		}
+		podsFound = true
 		images.Insert(pod.Spec.Containers[0].Image)
 
 		for i, containerStatus := range pod.Status.ContainerStatuses {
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -80,13 +80,15 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
 	errs := []error{}
 	failingErrorCount := 0
 	images := sets.NewString()
+	podsFound := false
 	for _, node := range originalOperatorStatus.NodeStatuses {
 		pod, err := c.podsGetter.Pods(c.targetNamespace).Get(ctx, mirrorPodNameForNode(c.staticPodName, node.NodeName), metav1.GetOptions{})
 		if err != nil {
 			errs = append(errs, err)
 			failingErrorCount++
```
sttts, line 88: we should distinguish between NotFound and other errors. 404 is not a reason to degrade, see logic below

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -126,13 +128,12 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
...
-		// if NodeStatuses is still empty, we are most probably in bootstrapping phase and this controller races with the
-		// installer controller. Hence, ignore that case. It will settle.
-
-		if len(originalOperatorStatus.NodeStatuses) > 0 {
-			syncCtx.Recorder().Warningf("MissingVersion", "no image found for operand pod")
-		}
+	// if NodeStatuses is still empty, we are most probably in bootstrapping phase and this controller races with the
+	// installer controller.
+	// If we can't Get any Pod from the apiserver most probably there is a network problem.
+	// Hence, ignore that case. It will settle.
```
sttts, line 134: the argument is not about network problem, but about having found at least one pod, but it lacks the env var for the image. This network comment is more confusing. Would just say that "// we found pods, but none has an image".

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
aojea: done

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
aojea: yep, actually I have to handle all the len(images) == 0 cases here otherwise it hits the default clause.

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -127,10 +132,8 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
 
 	switch {
 	case len(images) == 0:
-		// if NodeStatuses is still empty, we are most probably in bootstrapping phase and this controller races with the
-		// installer controller. Hence, ignore that case. It will settle.
-
-		if len(originalOperatorStatus.NodeStatuses) > 0 {
+		// only generate an event if we found pods, but none has an image.
```
sttts, line 135: wondering what the case "we found pods, some have an image, some don't" means. A pod without image is strange.

aojea: yeah
```
cat emptypod.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: emptypod
  namespace: default
$ kubectl apply -f emptypod.yaml -v2
The Pod "emptypod" is invalid: spec.containers: Required value
```

aojea: and
```
apiVersion: v1
kind: Pod
metadata:
  name: testemt
  namespace: default
spec:
  containers:
  - image: ""
    imagePullPolicy: Always
    name: test2
kubectl apply -f emptypod.yaml -v2
The Pod "testemt" is invalid: spec.containers[0].image: Required value
```

PR #1204, opened 2021-09-09
csi: add hook that annotates a DaemonSet with a secret's hash

bertinatto:
This is required by Azure Disk and Azure File operators.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivernodeservicecontroller/helpers.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/helpers.go
@@ -1,12 +1,16 @@
 package csidrivernodeservicecontroller
 
 import (
+	"crypto/sha256"
+	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/operator/csi/csidrivernodeservicecontroller/helpers.go
```diff
@@ -23,3 +27,38 @@ func WithObservedProxyDaemonSetHook() DaemonSetHookFunc {
 		return err
 	}
 }
+
+// WithSecretHashAnnotationHook creates a DaemonSet hook that annotates a DaemonSet with a secret's hash.
+func WithSecretHashAnnotationHook(
```
bertinatto, line 32: Already in use at https://github.com/openshift/azure-disk-csi-driver-operator/blob/master/pkg/operator/starter.go#L190

PR #1205, opened 2021-09-10
Bug 1858760: allow injecting authorizationService URL into Credentials

atiratree:
resolution in registryclient request authorization

- idea behing this: `CredentialStore` will implement `AuthorizationServiceRegistryMappingConsumer` interface to obtain authorizationService URL for mapping it to registry URL

- this is required to prevent hacky alternatives for obtaining authorizationService URL when retrieving tokens ([example](https://github.com/openshift/oc/pull/924/files#diff-592af2d02373e6118684e6622400bc1cf96c4cd51ae91098ad54ba9cf2354431R62))

### DIFF
--- a/pkg/image/registryclient/authorization_service_registry_mapping.go
+++ b/pkg/image/registryclient/authorization_service_registry_mapping.go
@@ -1,0 +1,35 @@
+package registryclient
+
+import (
+	"net/http"
+
+	"github.com/docker/distribution/registry/client/auth"
+)
+
+type AuthorizationServiceRegistryMappingConsumer interface {
+	AcceptAuthorizationServiceRegistryMapping(authorizationService, registry string)
+}
+
+type authorizationServiceHandler struct {
+	credentialStore auth.CredentialStore
+}
+
+func (*authorizationServiceHandler) Scheme() string {
+	return "bearer"
+}
+
+func (bh *authorizationServiceHandler) AuthorizeRequest(_ *http.Request, params map[string]string) error {
+	if consumer, ok := bh.credentialStore.(AuthorizationServiceRegistryMappingConsumer); ok {
+		// look for challenge params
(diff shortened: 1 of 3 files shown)


soltysh (approved):
The changes here looks reasonable to me, but since we're touching auth I'd like additional pair of eyes (actually 2) to look at it.

/assign @dmage 
from image pov
/assign @stlaz 
from auth pov

dmage:
@atiratree https://github.com/openshift/oc/blob/538ee42d982c248a704a106a60aba40c513da8b1/pkg/cli/image/manifest/dockercredentials/credentials.go#L55 - this line is incorrect, for every image reference you should have a different CredentialStore. You shouldn't workaround this implementation, instead you can use CredentialStoreFactory.

atiratree:
> @atiratree https://github.com/openshift/oc/blob/538ee42d982c248a704a106a60aba40c513da8b1/pkg/cli/image/manifest/dockercredentials/credentials.go#L55 - this line is incorrect, for every image reference you should have a different CredentialStore. You shouldn't workaround this implementation, instead you can use CredentialStoreFactory.

@dmage Thanks for highlighting that.
- we still had an old version of library-go where CredentialStoreFactory wasn't included yet
- I created a PR which adds a support for that, but I am still not sure of the right approach. Can you please take a look? https://github.com/openshift/oc/pull/947

Nevertheless I think this PR does not depend on the CredentialStoreFactory enhancement as it solves a different thing. Support for obtaining authorization service URL before calling `Basic()` on the CredentialStore is still needed as the current docker code does not support such behaviour: https://github.com/openshift/library-go/blob/master/vendor/github.com/docker/distribution/registry/client/auth/session.go#L422

atiratree:
closing in favor of a new implementation in https://github.com/openshift/oc/pull/924

PR #1209, opened 2021-09-13
Use inclusive language in observe_featuregates.go

adambkaplan:
Replace the use of "Whitelist" with "AllowList," and "Blacklist" with
"DenyList." This is in accordance with the Inclusive Naming Initiative's
recommendations for these terms. See
https://inclusivenaming.org/language/word-list/#whitelist-blacklist.


adambkaplan:
/remove-lifecycle stale

Will revisit after 2022

PR #1211, opened 2021-09-14
etcd observer: explicitly reports reserved address block as an error

p0lyn0mial:
As of today, the following blocks are reserved: 2001:db8::/32 and 192.0.2.0/24.
The usage of these IPs will be reported as a degraded condition (ConfigObservationDegraded)
Hopefully, this will make troubleshooting easier. In the past, it caused confusion as we didn't report any issues.

Note that we preserve backward compatibility since `kas-o` doesn't allow for usage of reserved IPs.
However, we still need to update `kas-o` because currently, it will silently drop reserved addresses

### DIFF
--- a/pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
+++ b/pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
@@ -106,6 +106,13 @@ func innerObserveStorageURLs(fallbackObserver fallBackObserverFn, alwaysAppendLo
 			errs = append(errs, ipErr)
 			continue
 		}
+
+		// explicitly report reserved IPs to the user
+		if strings.HasPrefix(ip.String(), "192.0.2.") || strings.HasPrefix(ip.String(), "2001:db8:") {
+			errs = append(errs, fmt.Errorf("configmap %s/%s: address %v is reserved and not supported", EtcdEndpointNamespace, etcdEndpointName, ip.String()))
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
```diff
@@ -106,6 +106,13 @@ func innerObserveStorageURLs(fallbackObserver fallBackObserverFn, alwaysAppendLo
 			errs = append(errs, ipErr)
 			continue
 		}
+
+		// explicitly report reserved IPs to the user
+		if strings.HasPrefix(ip.String(), "192.0.2.") || strings.HasPrefix(ip.String(), "2001:db8:") {
+			errs = append(errs, fmt.Errorf("configmaps/%s in the %s namespace: %v is reserved, please use a different address", etcdEndpointName, EtcdEndpointNamespace, ip.String()))
```
hexfusion, line 112: optional change.

`configmap %s/%s: address %v is reserved and not supported`

hexfusion:
/approve
/lgtm
/hold

Feel free to unhold

pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
p0lyn0mial: done

p0lyn0mial:
@hexfusion updated the PR. Please retag.

PR #1217, opened 2021-09-16
csi: add fbertina and jsafrane as approves

bertinatto:
This would avoid getting the Storage Team blocked.

/assign @mfojtik


jsafrane:
/lgtm
:-D

PR #1219, opened 2021-09-22
bug 2006945: extend hardcoded restmapper for cluster-bootstrap to avoid crashlooping bootstrap kube-apiserver

deads2k:
found on aws-ovn, 4.9 installation

1. bootstrap-kube-apiserver starts, readyz=false
2. cluster-bootstrap starts creating resources, but discovery does not return a full list
3. without a mapping for CRD, the rolebindingrestriction resource isn't created
4. without rolebindingrestrictions defined, the bootstrap-kube-apiserver poststarthook for initial rolebindings cannot create (blocked by admission)
5. the poststarthook fails, preventing initial rolebindings from being created causing a crashloop
6. additionally, permissions granted in our rendered artifacts are not added, this includes permissions for CSR creation for nodes
6. kubelets trying to join the cluster lack RBAC permissions for the initial CSR and anything before the kubelet client
7. pods trying to start may lack RBAC bound SCC permissions
8. operator pods that rely on visibility to config CRDs and CRs are delayed significantly by the lack of both

This avoids those conditions by hardcoding the RESTMappings that appeared to be missing and hopefully breaking the racy loop

### DIFF
--- a/pkg/client/openshiftrestmapper/hardcoded_restmapper.go
+++ b/pkg/client/openshiftrestmapper/hardcoded_restmapper.go
@@ -70,6 +70,96 @@ var defaultRESTMappings = []meta.RESTMapping{
 		Scope:            meta.RESTScopeRoot,
 		Resource:         schema.GroupVersionResource{Group: "security.openshift.io", Version: "v1", Resource: "securitycontextconstraints"},
 	},
+	// This is created so that cluster-bootstrap can always map customresourcedefinitions, RBAC, machine resources so that CRDs and
+	// permissions are always created quickly.  We observed discovery not including these on AWS OVN installations and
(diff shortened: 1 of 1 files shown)


deads2k:
/hold

needs proof PR in cluster-bootstrap

deads2k:
proof PR is https://github.com/openshift/cluster-bootstrap/pull/64

dgoodwin:
Possibly related to https://bugzilla.redhat.com/show_bug.cgi?id=2005967

dgoodwin:
Bug filed for this specific issue: https://bugzilla.redhat.com/show_bug.cgi?id=2006945

deads2k:
proof worked

/hold cancel

PR #1222, opened 2021-10-06
CertsFromPEM: avoid mutating the input byte slice

stlaz:
CertsFromPEM might have possibly been causing mutations of its
input when using pem.Decode. Avoid that by deep-copying.

---

Currently, this is the only spot which CAO uses that seems suspicious with regards to https://bugzilla.redhat.com/show_bug.cgi?id=1935654.

cc @s-urbaniak @sttts


stlaz:
/close
proved to be wrong hypothesis

PR #1226, opened 2021-10-08
Bug 2008539: Try another registry if blob is not found

dmage:
ServeBlob uses only the first connected registry, but it does not check
that the registry has needed data. If the blob is absent of this
registry or the registry rejects the request, the client can safely make
another attempt with another registry.

### DIFF
--- a/pkg/image/registryclient/client_mirrored.go
+++ b/pkg/image/registryclient/client_mirrored.go
@@ -2,12 +2,16 @@ package registryclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
 	"sync"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/registry/api/errcode"
+	"github.com/docker/distribution/registry/client"
+	"github.com/docker/distribution/registry/client/auth"
 	"github.com/opencontainers/go-digest"
 	"github.com/openshift/library-go/pkg/image/reference"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


xiuwang:
Since cluster-bot can't launch cluster from library-go repo, I just reviewed the code.
I will test this bug after pr merged.

pkg/image/registryclient/client_mirrored.go
```diff
@@ -436,6 +468,16 @@ func (f blobMirroredBlobstore) Open(ctx context.Context, dgst digest.Digest) (di
...
 		rsc, err = r.Blobs(ctx).Open(ctx, dgst)
+		if err != nil {
+			klog.V(5).Infof("open %s from %s: %v", dgst, r.Named(), err)
+			return err
+		}
+
+		// Distribution's implementation of Open doesn't send any requests to
+		// the registry. We need the reader to send a request to see if the
+		// registry can serve the blob.
+		_, err = rsc.Read([]byte{})
```
smarterclayton, line 479: Yikes.  What are the performance characteristic changes to this?

This changes every trivial serve blob into a stat?

dmage: In most cases it won't causes additional HTTP requests.

https://github.com/distribution/distribution/blob/1563384b69df9376389fe45ce949173a6383770a/registry/client/transport/http_reader.go#L66-L93

It forces the client to establish an HTTP connection and this connection will be reused by next reads. It won't be reused only if the caller decides to Seek before Read.

(Btw, we increased the number of requests by https://github.com/openshift/library-go/pull/1084, see https://github.com/openshift/image-registry/pull/278/commits/d7eeb86d675641dfcda4ba2dab0ef2f3d8e8e7b7; perhaps we need to check why)

dmage: (`r.Blobs(ctx)` creates a BlobService with an internal cache, as we recreate BlobService on every call, it does not cache anything)

dmage: (https://github.com/openshift/library-go/compare/master...dmage:cache-blobstore addresses the problem with the extra request, but as it is not related to the BZ, perhaps it should be a follow-up PR?)

pkg/image/registryclient/client_mirrored.go
```diff
@@ -197,7 +215,18 @@ func (r *blobMirroredRepository) attemptFirstConnectedRepo(ctx context.Context,
 			}
 			continue
 		}
-		return fn(repo)
+		if err := fn(repo); err != nil {
+			if !isRequestError(err) {
```
soltysh, line 219: Why in those particular cases we would return instead of falling back to other registries?

dmage:
@soltysh @smarterclayton the PR is ready for the second review, ptal

dmage:
@soltysh @smarterclayton please take a look and tag it

dmage:
We need to update this library in `image-registry` so that the image-registry can use all mirrors.
We can update this library in `oc` so that it can use all mirrors and has better error messages when it mirrors images.

soltysh (approved):
/lgtm
@smarterclayton has approval rights

deads2k:
> We need to update this library in `image-registry` so that the image-registry can use all mirrors.
> We can update this library in `oc` so that it can use all mirrors and has better error messages when it mirrors images.

this sounds reasonable to me.  The additional call in error cases looks like it degrades the already degraded cases.

/approve

PR #1227, opened 2021-10-12
add greater intertia for workloadDeploymentDegraded

deads2k:
See the code comment for some details.

ok, here's a theory.  Based https://coreos.slack.com/archives/C02CZNQHGN8/p1633973677227800 we had no OS upgrade for this micro and based on logs there was no networking update.  This means that the authentication operator had a pod that ended up backing off the scheduler upgrade.
The CVO decides upgrade success based on each component reaching A=t, p=f once.  So the authentication operator made it past the post before MCO started rolling out.  Because the deployment didn't change, the operator didn't go p=t.  Because it had 2/3 available pods, it went d=t and the inertia against only lasted 2 minutes.  The CVO called upgrade complete,but the auth and openshift-apiserver operators were both d=t.
Because those deployments now have PDBs that prevent going non-HA, I think we can now relax the degraded interia constraint on available pods.  I suggest that we can make intertia for APIServerDeploymentDegraded to 30 minutes to accommodate upgrades.  I think https://github.com/openshift/library-go/blob/master/pkg/operator/status/inertia.go#L49 allows individual conditions to be specified, so we can be narrow on just the APIServerDeploymentDegraded , which appears to be focused on available replicas.
That will also improve our "authentication goes degraded during upgrade".
If this this theory is correct, the next networking, kubelet, or OS change will unstick the payload

### DIFF
--- a/pkg/operator/apiserver/controllerset/apiservercontrollerset.go
+++ b/pkg/operator/apiserver/controllerset/apiservercontrollerset.go
@@ -3,6 +3,8 @@ package apiservercontrollerset
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -71,6 +73,8 @@ type APIServerControllerSet struct {
 	operatorClient v1helpers.OperatorClient
 	eventRecorder  events.Recorder
 
+	workloadConditionsPrefix string
+
 	apiServiceController            controllerWrapper
 	auditPolicyController           controllerWrapper
 	clusterOperatorStatusController controllerWrapper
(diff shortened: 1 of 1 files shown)


pkg/operator/apiserver/controllerset/apiservercontrollerset.go
```diff
@@ -137,9 +141,39 @@ func (cs *APIServerControllerSet) WithClusterOperatorStatusController(
...
+			WithDegradedInertia(status.MustNewInertia(
+				2*time.Minute,
+				// the workload deployment degraded condition happens when the number of available replicas does not
+				// match the desired state.  Assuming a PDB is present to ensure that we do not optionally dip below
+				// HA (not enforced here, but probably should be in the staticresourcecontroller check), we can safely
+				// allow being down a single replica for a considerable period of time.  Doing it here allows the
+				// detailed operator state to be fully correct (degraded instantly), while allowing graceful resolution
+				// in the majority of cases.
+				// The primary disadvantage is that we will no longer see signal in these operators when a single node
+				// does not quickly come back.  I actually caught this because there appears to be a significant scheduling
```
sttts, line 166: don't use first person. Reads odd.

pkg/operator/apiserver/controllerset/apiservercontrollerset.go
```diff
@@ -137,9 +141,39 @@ func (cs *APIServerControllerSet) WithClusterOperatorStatusController(
...
+		cs.clusterOperatorStatusController.controller = cs.clusterOperatorStatusController.controller.(*status.StatusSyncer).
+			WithDegradedInertia(status.MustNewInertia(
+				2*time.Minute,
+				// the workload deployment degraded condition happens when the number of available replicas does not
+				// match the desired state.  Assuming a PDB is present to ensure that we do not optionally dip below
+				// HA (not enforced here, but probably should be in the staticresourcecontroller check), we can safely
+				// allow being down a single replica for a considerable period of time.  Doing it here allows the
+				// detailed operator state to be fully correct (degraded instantly), while allowing graceful resolution
+				// in the majority of cases.
+				// The primary disadvantage is that we will no longer see signal in these operators when a single node
```
mfojtik, line 165: @deads2k according to @ravisantoshgudimetla we don't install PDB in SNO.

s-urbaniak:
> looks like auth has a second deployment: OAuthServerDeploymentDegraded which also doesn't surge. Likely for similar audit reasons.
hm, oauth-server afaik doesn't use PDBs and doesn't have an audit log right now.

pkg/operator/apiserver/controllerset/apiservercontrollerset.go
ravisantoshgudimetla: Correction: We don't intend to. I believe the PRs for creating PDBs for non-static components by @ingvagabund  did not take into account the cluster topology which might be a problem for future

PR #1231, opened 2021-10-17
pkg/operator/staticpod/controller/installer: add support for WithNodeFilter

hexfusion:
Today static pods are rolled out to every control-plane node in the cluster. For some operands such as etcd we desire more precise control of the nodes which an installer pod is scheduled. With the addition of `WithNodeFilter` the operator can explicitly prescribe given multiple new nodes which node will be scaled up first or limit the number of nodes that the static pod is installed to.

Signed-off-by: Sam Batschelet <sbatsche@redhat.com>

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -94,6 +94,8 @@ type InstallerController struct {
 
 	startupMonitorEnabled func() (bool, error)
 
+	nodeFilterFn func(ctx context.Context) (map[string]bool, error)
+
 	factory          *factory.Factory
 	clock            clock.Clock
 	installerBackOff func(count int) time.Duration
@@ -110,6 +112,13 @@ func (c *InstallerController) WithInstallerPodMutationFn(installerPodMutationFn 
 
 func (c *InstallerController) WithMinReadyDuration(minReadyDuration time.Duration) *InstallerController {
 	c.minReadyDuration = minReadyDuration
+	return c
+}
+
+// WithNodeFilter sets a map of all control plane nodes called on every sync with a bool
+// representing if the node was included for the revision.
+func (c *InstallerController) WithNodeFilter(nodeFilterFn func(ctx context.Context) (map[string]bool, error)) *InstallerController {
+	c.nodeFilterFn = nodeFilterFn
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -384,6 +392,36 @@ func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodS
 	return 0, reason, nil
 }
 
+// nodeStatusesSelectedForRevision returns a list of nodeStatuses which have been selected for the next revision.
+func nodeStatusesSelectedForRevision(ctx context.Context, nodeSelectorFunc func(ctx context.Context) (map[string]bool, error), nodeStatuses []operatorv1.NodeStatus) ([]operatorv1.NodeStatus, error) {
```
hexfusion, line 396: nodeSelectorFn

hexfusion:
prrof https://github.com/openshift/cluster-etcd-operator/pull/704

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -384,6 +393,42 @@ func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodS
...
+	if nodeFilterFn == nil {
+		return nodeStatuses, nil
+	}
+
+	nodeSelectorHostnameMap, err := nodeFilterFn(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if len(nodeSelectorHostnameMap) == 0 {
+		return nodeStatuses, nil
```
hasbro17, line 407: I can't exactly think of when the [etcd-operator's filter](https://github.com/openshift/cluster-etcd-operator/blob/7951acdaf4a8ebcffc5cfbf31e5f5bd184b5285d/pkg/operator/staticpodnodefilter/staticpodnodefilter.go#L13) would return an empty node map (maybe some node list error not caught by the filter), but if for whatever reason it does, do we want to resort to the default of no-filter, or error out saying the filter is invalid?
```suggestion
	if len(nodeSelectorHostnameMap) == 0 {
		return nil, fmt.Errorf("invalid node filter: the node map cannot be empty")
```

hasbro17: And we are error-ing out below as well when the node map is not fully populated too. So maybe just the same error.

hasbro17 (commented):
SGTM except one nit.

pkg/operator/staticpod/controller/installer/installer_controller.go
hexfusion: good catch!

PR #1235, opened 2021-10-25
Improve performance by caching src/dest when applying resource manifests

jerpeter1:
If we see that neither the manifest being applied (by caching a hash of the manifest),
or the existing resource in the cluster (by caching the resourceVersion) has
changed since the manifest was last updated, applying the resource can be
skipped, since it won't actually do anything.

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -3,22 +3,154 @@ package resourceapply
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
+	"errors"
 	"fmt"
+	"io"
 	"sort"
 	"strings"
-
-	"k8s.io/klog/v2"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	coreclientv1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
+
+type CachedVersionKey struct {
+	name, kind, namespace string
+}
+
+// record of resource metadata used to determine if its safe to return early from an ApplyFoo
+// resourceHash is an ms5 hash of the required in an ApplyFoo that is computed in case the input changes
+// resourceVersion is the received resourceVersion from the apiserver in response to an update that is comparable to the GET
+type CachedResource struct {
+	resourceHash, resourceVersion string
+}
+
+func getResourceMetadata(obj runtime.Object) (string, string, string, string, error) {
+	var name, kind, namespace, resourceHash string
+	switch r := obj.(type) {
+	case *corev1.Namespace:
+		name = r.Name
+		kind = r.Kind
+	case *corev1.Service:
+		name = r.Name
+		kind = r.Kind
+		namespace = r.Namespace
+	case *corev1.Pod:
+		name = r.Name
+		kind = r.Kind
+		namespace = r.Namespace
+	case *corev1.ServiceAccount:
+		name = r.Name
+		kind = r.Kind
+		namespace = r.Namespace
+	case *corev1.ConfigMap:
+		name = r.Name
+		kind = r.Kind
+		namespace = r.Namespace
+	case *corev1.Secret:
+		name = r.Name
+		kind = r.Kind
+		namespace = r.Namespace
+	default:
+		return name, kind, namespace, resourceHash, errors.New("attempted to access metadata for unsupported type")
+	}
+	resourceHash = hashOfResourceStruct(obj)
+	return name, kind, namespace, resourceHash, nil
+}
+
+func getResourceVersion(obj runtime.Object) (string, error) {
+	var resourceVersion string
+	switch r := obj.(type) {
+	case *corev1.Namespace:
+		resourceVersion = r.ResourceVersion
+	case *corev1.Service:
+		resourceVersion = r.ResourceVersion
+	case *corev1.Pod:
+		resourceVersion = r.ResourceVersion
+	case *corev1.ServiceAccount:
+		resourceVersion = r.ResourceVersion
+	case *corev1.ConfigMap:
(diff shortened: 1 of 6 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
+type CachedVersionKey struct {
+	name, kind, namespace string
+}
+
+type CachedResource struct {
+	resourceHash, resourceVersion string
+}
+
+// map of name,kind to resourceHash (of last applied resource) : resourceVersion(when resource was applied)
+var cachedResourceVersions map[CachedVersionKey]CachedResource = make(map[CachedVersionKey]CachedResource)
```
deads2k, line 33: no global state please.  I'd rather either
1. break everyone and change the methods
2. create new methods that accept a cache or use a method receiver struct to track it and have the staticresource controller maintain them.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
+type CachedVersionKey struct {
+	name, kind, namespace string
+}
+
+type CachedResource struct {
+	resourceHash, resourceVersion string
```
deads2k, line 29: for future readers can you godoc these.  I think that `resourcehash` is the hash of the `required` in an ApplyFoo that is computed in case the input changes and `resourceVersion` is the received resourceVersion from the apiserver in response to an update that is comparable to the GET.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
+}
+
+// map of name,kind to resourceHash (of last applied resource) : resourceVersion(when resource was applied)
+var cachedResourceVersions map[CachedVersionKey]CachedResource = make(map[CachedVersionKey]CachedResource)
+
+func updateCachedResourceVersion(name string, kind string, namespace string, resourceHash string, resourceVersion string) {
```
deads2k, line 35: if this took a `runtime.Object` instead of a `resourceVersion`, you'd be able to simply call `updateCachedResourceVersion` and have it clear on nil (or do nothing) and avoid all the nil checks in callers.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
+}
+
+func clearCachedResourceVersion() {
+	cachedResourceVersions = make(map[CachedVersionKey]CachedResource)
+}
+
+// if the manifest applied is the same as the last time this resource was
+// updated, and the resourceVersion also hasn't changed, we don't need to
+// actually do anything
+func checkCachedResourceVersion(name string, kind string, namespace string, resourceHash string, resourceVersion string) bool {
```
deads2k, line 51: `hasResourceMutated` or `resourceChanged` or `mayNeedUpdate`?

pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
+		hashMatch = cached.resourceHash == resourceHash
+		if versionMatch && hashMatch {
+			klog.V(7).Infof("found matching resourceVersion & manifest hash")
+			return true
+		}
+	}
+	return false
+}
+
+func hashOfResourceStruct(o interface{}) string {
```
deads2k, line 64: please note that this only works if the string representation has changed.  Not all changes result in changed ToString functions.  But it's a pretty narrow edge.

pkg/operator/resource/resourceapply/core.go
```diff
@@ -3,42 +3,102 @@ package resourceapply
...
+	h := md5.New()
+	io.WriteString(h, oString)
+	rval := fmt.Sprintf("%x", h.Sum(nil))
+	return rval
+}
+
 // ApplyNamespace merges objectmeta, does not worry about anything else
 func ApplyNamespace(ctx context.Context, client coreclientv1.NamespacesGetter, recorder events.Recorder, required *corev1.Namespace) (*corev1.Namespace, bool, error) {
 	existing, err := client.Namespaces().Get(ctx, required.Name, metav1.GetOptions{})
+	resourceHash := hashOfResourceStruct(required)
```
deads2k, line 75: While technically more efficient, this requires callers to compute instead of simply passing their `required` to `checkCachedResourceVersion` and `updateCachedResourceVersion`.  My preference is to pass the runtime.Objects around and know that if we actually issued an update, it was significantly more expensive than re-computing the hash.

pkg/operator/resource/resourceapply/core.go
jerpeter1: I chose option 2. Since there are so many users of resourceapply.ApplyFoo functions, I decided to create new functions that accept a cache and use those new functions from static resource controller, buta t some point I'd like to go back and add something that lets everyone take advantage of the enhanced performance.

pkg/operator/resource/resourceapply/core.go
jerpeter1: Done

pkg/operator/resource/resourceapply/core.go
jerpeter1: Done

pkg/operator/resource/resourceapply/core.go
jerpeter1: I went with `safeToSkipApply`

pkg/operator/resource/resourceapply/core.go
jerpeter1: I think with the reduced scoped of the last refactor, that this won't come up. I couldn't get find this edge case in my testing - under what circumstances would the string representation of a Resource from the apiserver not change when the Resource has changed?

pkg/operator/resource/resourceapply/core.go
jerpeter1: Done. It is less performant than the previous implementation (since we're calculating the hash of `required` several times from the same ApplyFoo, but it's still an improvement from the prior implementation.

pkg/operator/resource/resourceapply/core.go
deads2k: > I think with the reduced scoped of the last refactor, that this won't come up. I couldn't get find this edge case in my testing - under what circumstances would the string representation of a Resource from the apiserver not change when the Resource has changed?

nil versus empty is pretty classic.  It has come up a few times when handling fuzzing tests.  I'm ok with the gap as long as it is described.

jerpeter1: Added description

PR #1238, opened 2021-10-27
Guard controller

ingvagabund:
Deploy guarding pods with a PDB which will make sure a node does not drain a node until at most one replica of an operand is unavailable. Each guarding pod checks healthz status of an operand on the same node.

The guard controller repeatedly checks available master nodes and for each one renders an unmanaged guard pod. Each guard pod is responsible for checking if corresponding operand (operand pod running on the same node as the guard pod) is ready (through checking the `/healthz` endpoint). The guard uses an https probe to check readiness of its operand. At the same time a pdb with minAvailable set to `len(masters)-1` is rendered. Each time the number of master nodes changes, the `minAvailable` field is updated accordingly. 

Outstanding facts:
- guard pod has its nodename set directly (in case the KS is down)
- the Guard controller can delete pods (in case image/probe host changes)
- the Guard controller requires create/delete condition checkers (in case the number of master nodes is scaled to 1 to remove the PDB)

Applied and tested in:
- https://github.com/openshift/cluster-kube-scheduler-operator/pull/373
- https://github.com/openshift/cluster-kube-controller-manager-operator/pull/568

### DIFF
--- a/pkg/operator/staticpod/controller/guard/bindata/bindata.go
+++ b/pkg/operator/staticpod/controller/guard/bindata/bindata.go
@@ -1,0 +1,313 @@
+// Code generated for package bindata by go-bindata DO NOT EDIT. (@generated)
+// sources:
+// pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
+// pkg/operator/staticpod/controller/guard/manifests/pdb.yaml
+package bindata
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+type asset struct {
+	bytes []byte
+	info  os.FileInfo
+}
+
+type bindataFileInfo struct {
+	name    string
+	size    int64
+	mode    os.FileMode
+	modTime time.Time
+}
+
+// Name return file name
+func (fi bindataFileInfo) Name() string {
+	return fi.name
+}
+
+// Size return file size
+func (fi bindataFileInfo) Size() int64 {
+	return fi.size
+}
+
+// Mode return file mode
+func (fi bindataFileInfo) Mode() os.FileMode {
+	return fi.mode
+}
+
+// Mode return file modify time
+func (fi bindataFileInfo) ModTime() time.Time {
+	return fi.modTime
+}
+
+// IsDir return file whether a directory
+func (fi bindataFileInfo) IsDir() bool {
+	return fi.mode&os.ModeDir != 0
+}
+
+// Sys return file is sys mode
+func (fi bindataFileInfo) Sys() interface{} {
+	return nil
+}
+
+var _pkgOperatorStaticpodControllerGuardManifestsGuardPodYaml = []byte(`apiVersion: v1
+kind: Pod
+metadata:
+  namespace: # Value set by operator
+  name: # Value set by operator
+  labels:
+    app: guard
+  ownerReferences: # Value set by operator
+spec:
+  affinity: # Value set by operator
+  priorityClassName: "system-cluster-critical"
+  terminationGracePeriodSeconds: 3
+  tolerations:
+    - key: node-role.kubernetes.io/master
+      effect: NoSchedule
+      operator: Exists
+    - key: node.kubernetes.io/not-ready
+      effect: NoExecute
+      operator: Exists
+    - key: node.kubernetes.io/unreachable
+      effect: NoExecute
+      operator: Exists
+    - key: node-role.kubernetes.io/etcd
+      operator: Exists
+      effect: NoSchedule
+  containers:
+    - name: guard
+      image: # Value set by operator
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/bash
+      args:
+        - -c
+        - |
+          # properly handle TERM and exit as soon as it is signaled
+          set -euo pipefail
+          trap 'jobs -p | xargs -r kill; exit 0' TERM
+          sleep infinity & wait
+      readinessProbe:
+        failureThreshold: 3
+        httpGet:
+          host: # Value set by operator
+          path: healthz
+          port: # Value set by operator
+          scheme: HTTPS
+        periodSeconds: 5
+        successThreshold: 1
+        timeoutSeconds: 5
+      resources:
+        requests:
+          cpu: 10m
+          memory: 5Mi
+`)
+
+func pkgOperatorStaticpodControllerGuardManifestsGuardPodYamlBytes() ([]byte, error) {
+	return _pkgOperatorStaticpodControllerGuardManifestsGuardPodYaml, nil
+}
+
+func pkgOperatorStaticpodControllerGuardManifestsGuardPodYaml() (*asset, error) {
+	bytes, err := pkgOperatorStaticpodControllerGuardManifestsGuardPodYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
+	a := &asset{bytes: bytes, info: info}
+	return a, nil
+}
+
+var _pkgOperatorStaticpodControllerGuardManifestsPdbYaml = []byte(`apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: # Value set by operator
+  namespace: # Value set by operator
+spec:
+  minAvailable: 0 # Value set by operator
+  selector:
+    matchLabels:
+      app: guard
+`)
+
+func pkgOperatorStaticpodControllerGuardManifestsPdbYamlBytes() ([]byte, error) {
+	return _pkgOperatorStaticpodControllerGuardManifestsPdbYaml, nil
+}
+
+func pkgOperatorStaticpodControllerGuardManifestsPdbYaml() (*asset, error) {
+	bytes, err := pkgOperatorStaticpodControllerGuardManifestsPdbYamlBytes()
+	if err != nil {
+		return nil, err
+	}
+
+	info := bindataFileInfo{name: "pkg/operator/staticpod/controller/guard/manifests/pdb.yaml", size: 0, mode: os.FileMode(0), modTime: time.Unix(0, 0)}
+	a := &asset{bytes: bytes, info: info}
+	return a, nil
+}
+
+// Asset loads and returns the asset for the given name.
+// It returns an error if the asset could not be found or
+// could not be loaded.
+func Asset(name string) ([]byte, error) {
+	cannonicalName := strings.Replace(name, "\\", "/", -1)
+	if f, ok := _bindata[cannonicalName]; ok {
+		a, err := f()
+		if err != nil {
+			return nil, fmt.Errorf("Asset %s can't read by error: %v", name, err)
+		}
+		return a.bytes, nil
+	}
+	return nil, fmt.Errorf("Asset %s not found", name)
+}
+
+// MustAsset is like Asset but panics when Asset would return an error.
+// It simplifies safe initialization of global variables.
+func MustAsset(name string) []byte {
+	a, err := Asset(name)
+	if err != nil {
+		panic("asset: Asset(" + name + "): " + err.Error())
+	}
+
+	return a
+}
+
+// AssetInfo loads and returns the asset info for the given name.
+// It returns an error if the asset could not be found or
+// could not be loaded.
+func AssetInfo(name string) (os.FileInfo, error) {
+	cannonicalName := strings.Replace(name, "\\", "/", -1)
+	if f, ok := _bindata[cannonicalName]; ok {
+		a, err := f()
+		if err != nil {
+			return nil, fmt.Errorf("AssetInfo %s can't read by error: %v", name, err)
+		}
+		return a.info, nil
+	}
+	return nil, fmt.Errorf("AssetInfo %s not found", name)
+}
+
+// AssetNames returns the names of the assets.
+func AssetNames() []string {
+	names := make([]string, 0, len(_bindata))
+	for name := range _bindata {
+		names = append(names, name)
+	}
+	return names
+}
+
+// _bindata is a table, holding each asset generator, mapped to its name.
+var _bindata = map[string]func() (*asset, error){
+	"pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml": pkgOperatorStaticpodControllerGuardManifestsGuardPodYaml,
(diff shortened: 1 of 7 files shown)


pkg/operator/resource/resourceapply/generic.go
```diff
@@ -238,6 +242,135 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
...
+			} else {
+				_, result.Changed, result.Error = DeleteCSIDriver(ctx, clients.kubeClient.StorageV1(), recorder, t)
+			}
+		case *migrationv1alpha1.StorageVersionMigration:
+			if clients.migrationClient == nil {
+				result.Error = fmt.Errorf("missing migrationClient")
+			} else {
+				_, result.Changed, result.Error = DeleteStorageVersionMigration(ctx, clients.migrationClient, recorder, t)
+			}
+		case *unstructured.Unstructured:
```
deads2k, line 358: note that in combination with a RESTMapper, we'd be able to delete any manfiest we can read with a one to one mapping of resources to kind.

Doesn't have to be in this PR, but seems like a valuable future thing.

deads2k:
move `Allow resource deletion based on a certain condition ` to its own commit so I can merge it

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+		pdb := resourceread.ReadPodDisruptionBudgetV1OrDie(bindata.MustAsset(filepath.Join("pkg/operator/staticpod/controller/guard", "manifests/pdb.yaml")))
+		pdb.ObjectMeta.Name = getGuardPDBName(c.podResourcePrefix)
+		pdb.ObjectMeta.Namespace = c.targetNamespace
+		if len(nodes) > 1 {
+			minAvailable := intstr.FromInt(len(nodes) - 1)
+			pdb.Spec.MinAvailable = &minAvailable
+		}
+
+		_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
```
deads2k, line 160: The PDB portion of this could be applied using our static resource controller and a `WithConditionalResource`, right?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+		podGetter:             podGetter,
+		pdbGetter:             pdbGetter,
+		installerPodImageFn:   getInstallerPodImageFromEnv,
+		createConditionalFunc: createConditionalFunc,
+		deleteConditionalFunc: deleteConditionalFunc,
+	}
+
+	return factory.New().WithInformers(
+		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		kubeInformersForTargetNamespace.Core().V1().Nodes().Informer(),
```
deads2k, line 76: one of these is wrong.  nodes are clusterscoped, so namespace scoping doesn't work.  and you want to be sure you only watch pods in your namespace.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+func getGuardPodName(prefix, nodeName string) string {
+	return fmt.Sprintf("%s-guard-%s", prefix, nodeName)
+}
+
+func getGuardPDBName(prefix string) string {
+	return fmt.Sprintf("%s-guard-pdb", prefix)
+}
+
+func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
deads2k, line 92: sync methods need to be zero-API call in the no-action case.  This requires lister usage for PDBs, nodes, and pods for instance.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func NewGuardController(
+	targetNamespace, podResourcePrefix string,
+	operatorName string,
+	readyzPort string,
+	kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	kubeInformersClusterScoped informers.SharedInformerFactory,
+	podGetter corev1client.PodsGetter,
+	pdbGetter policyclientv1.PodDisruptionBudgetsGetter,
```
deads2k, line 56: is this passed a caching GET client?

pkg/operator/staticpod/controller/guard/guard_controller.go
deads2k: > The PDB portion of this could be applied using our static resource controller and a `WithConditionalResource`, right?

I see, it's dynamic. nevermind then.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+		pdbGetter:             pdbGetter,
+		installerPodImageFn:   getInstallerPodImageFromEnv,
+		createConditionalFunc: createConditionalFunc,
+		deleteConditionalFunc: deleteConditionalFunc,
+	}
+
+	return factory.New().WithInformers(
+		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		kubeInformersForTargetNamespace.Core().V1().Nodes().Informer(),
+	).WithSync(c.sync).ToController("GuardController", eventRecorder)
```
deads2k, line 77: `WithSyncDegradedOnError`?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,255 @@
...
+		selector, err := labels.NewRequirement("node-role.kubernetes.io/master", selection.Equals, []string{""})
+		if err != nil {
+			panic(err)
+		}
+		nodes, err := c.nodeLister.List(labels.NewSelector().Add(*selector))
+		if err != nil {
+			return err
+		}
+
+		pods, err := c.podGetter.Pods(c.targetNamespace).List(ctx, metav1.ListOptions{
```
deads2k, line 143: lister

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: The same that's passed to other staticpod controllers

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Can you more clarify on what falls under the no-action case? Do you mean in order to decide if it is the no-action case the code needs to call c.createConditionalFunc/c.deleteConditionalFunc?

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: `kubeInformersForTargetNamespace.Core().V1().Nodes().Informer()` replaced with `kubeInformersClusterScoped.Core().V1().Nodes().Informer()`

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Done

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Done

pkg/operator/staticpod/controller/guard/guard_controller.go
deads2k: > Can you more clarify on what falls under the no-action case? Do you mean in order to decide if it is the no-action case the code needs to call c.createConditionalFunc/c.deleteConditionalFunc?

in the delete case, you issue a delete because you do not use a lister first to see if there is work to do.

pkg/operator/staticpod/controller/guard/bindata/bindata.go
```diff
@@ -0,0 +1,313 @@
+// Code generated for package bindata by go-bindata DO NOT EDIT. (@generated)
```
soltysh, line 1: Oh my, we got rid of go-bindata from all the operators but library-go is still left with one :disappointed: we'll should fix that...

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,254 @@
...
+
+		operands := map[string]*corev1.Pod{}
+		for _, pod := range pods {
+			operands[pod.Spec.NodeName] = pod
+		}
+
+		for _, node := range nodes {
+			if _, exists := operands[node.Name]; !exists {
+				klog.Errorf("Missing operand on node %v", node.Name)
+				errs = append(errs, fmt.Errorf("Missing operand on node %v", node.Name))
```
soltysh, line 174: During reboots or similar problems we'll be throwing a lot of errors, I don't think that's the intention. Shouldn't we filter out NotReady nodes earlier in the node query? IIRC installer ignores not ready nodes.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: I see what you mean. Both PDB cases updated. Both Apply/Delete PDB are preceded by checking the the PDB needs an update/exists.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,254 @@
...
+		podResourcePrefix:     podResourcePrefix,
+		operatorName:          operatorName,
+		readyzPort:            readyzPort,
+		nodeLister:            kubeInformersClusterScoped.Core().V1().Nodes().Lister(),
+		podLister:             kubeInformersForTargetNamespace.Core().V1().Pods().Lister(),
+		podGetter:             podGetter,
+		pdbGetter:             pdbGetter,
+		installerPodImageFn:   getInstallerPodImageFromEnv,
+		createConditionalFunc: createConditionalFunc,
+		deleteConditionalFunc: deleteConditionalFunc,
```
soltysh, line 75: Looking at https://github.com/openshift/cluster-kube-scheduler-operator/pull/373 I see this is only used for SNO check, any other particular reason you want to split into create and delete condition? If that's only SNO, why not having a single check?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,254 @@
...
+
+			if operands[node.Name].Status.PodIP == "" {
+				klog.Errorf("Missing PodIP in operand %v on node %v", operands[node.Name].Name, node.Name)
+				errs = append(errs, fmt.Errorf("Missing PodIP in operand %v on node %v", operands[node.Name].Name, node.Name))
+				continue
+			}
+
+			klog.V(5).Infof("Rendering guard pod for operand %v on node %v", operands[node.Name].Name, node.Name)
+
+			pod := resourceread.ReadPodV1OrDie(bindata.MustAsset(filepath.Join("pkg/operator/staticpod/controller/guard", "manifests/guard-pod.yaml")))
```
soltysh, line 186: This seems like a lot of unnecessary reads, why not just read it once when we create the controller and then just update the internal representation as needed and apply.

soltysh (commented):
Left some questions, in general this looks good.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Since create is not always of !delete (not delete)

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Good point. In case the operand pod does not exist, I will skip in case the underlying node is not ready. That might help reduce the number of errors. Also, once a node condition changes, the sync loop gets triggered again so no need to return an error here.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: `bindata.MustAsset` already has the internal representation as it is. So it's just a convenient wrapper around it.

pkg/operator/staticpod/controller/guard/bindata/bindata.go
ingvagabund: That will require more changes. Better to do it in another PR.

soltysh: Definitely a separate PR, just an observation :wink:

pkg/operator/staticpod/controller/guard/guard_controller.go
soltysh: Right, but returning `bool, errror` tuple is an option you also mention there.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,254 @@
...
+				klog.Errorf("Unable to apply pod %v changes: %v", pod.Name, err)
+				errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+			}
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func WithIsSNOCheck(infraLister configv1listers.InfrastructureLister, neg bool) resourceapply.ConditionalFunction {
```
soltysh, line 236: Do we really need `neg` parameter to negate the result, it's not that hard to add `!` before the invocation, is it?

soltysh (approved):
Two nits about `ConditionalFunc`, but those can be either addressed in followup or I can be overruled by majority :wink: 
/hold

/lgtm

soltysh (approved):
/lgtm
/hold cancel

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -0,0 +1,291 @@
...
+			// The readyz port as string type is expected to be convertible into int!!!
+			readyzPort, err := strconv.Atoi(c.readyzPort)
+			if err != nil {
+				panic(err)
+			}
+			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromInt(readyzPort)
+
+			actual, err := c.podGetter.Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
+			if err == nil {
+				// Delete the pod so it can be re-created. ApplyPod only updates the metadata part of the manifests, ignores the rest
```
deads2k, line 245: ApplyPod shoudl be fixed.

deads2k:
nothing I'll block on

/approve

ingvagabund:
Something changed meantime
```
pkg/operator/staticpod/controller/guard/guard_controller_test.go:379:7: unknown field 'Handler' in struct literal of type "k8s.io/api/core/v1".Probe
FAIL	github.com/openshift/library-go/pkg/operator/staticpod/controller/guard [build failed]
```

ingvagabund:
`Handler` renamed to `ProbeHandler`

soltysh (approved):
/lgtm

PR #1239, opened 2021-11-02
Remove AdmissionTimeout and use context timeout

ardaguclu:
This PR removes deprecated AdmissionTimeout struct and uses context timeout
in pluginHandlerWithTimeout.

Thanks to that, consumers can get admission plugin whose timeout value
is wrapped in context via `admissiontimeout.WithTimeout`.

### DIFF
--- a/pkg/apiserver/admission/admissiontimeout/decorator.go
+++ b/pkg/apiserver/admission/admissiontimeout/decorator.go
@@ -1,12 +1,18 @@
 package admissiontimeout
 
 import (
+	"time"
+
 	"k8s.io/apiserver/pkg/admission"
 )
 
+const defaultAdmissionTimeout = 13 * time.Second
+
+// WithTimeout decorates admission plugin with timeout
 func WithTimeout(admissionPlugin admission.Interface, name string) admission.Interface {
 	return pluginHandlerWithTimeout{
-		name:            name,
-		admissionPlugin: admissionPlugin,
+		name:             name,
+		admissionPlugin:  admissionPlugin,
+		admissionTimeout: defaultAdmissionTimeout,
 	}
 }
--- a/pkg/apiserver/admission/admissiontimeout/decorator.go
+++ b/pkg/apiserver/admission/admissiontimeout/decorator.go
@@ -1,22 +1,12 @@
 package admissiontimeout
 
 import (
-	"time"
-
 	"k8s.io/apiserver/pkg/admission"
 )
 
-// AdmissionTimeout provides a decorator that will fail an admission plugin after a certain amount of time
-//
-// DEPRECATED: use the context of the admission handler instead.
-type AdmissionTimeout struct {
-	Timeout time.Duration
-}
-
-func (d AdmissionTimeout) WithTimeout(admissionPlugin admission.Interface, name string) admission.Interface {
+func WithTimeout(admissionPlugin admission.Interface, name string) admission.Interface {
 	return pluginHandlerWithTimeout{
 		name:            name,
 		admissionPlugin: admissionPlugin,
-		timeout:         d.Timeout,
 	}
 }
--- a/pkg/apiserver/admission/admissiontimeout/timeoutadmission.go
+++ b/pkg/apiserver/admission/admissiontimeout/timeoutadmission.go
@@ -25,6 +25,7 @@ func (p pluginHandlerWithTimeout) Handles(operation admission.Operation) bool {
 
 func (p pluginHandlerWithTimeout) Admit(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
 	admissionCtx, cancelFn := context.WithTimeout(ctx, p.admissionTimeout)
+	deadline := getTimeoutFromContext(admissionCtx)
 	defer cancelFn()
 
 	mutatingHandler, ok := p.admissionPlugin.(admission.MutationInterface)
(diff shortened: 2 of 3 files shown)


pkg/apiserver/admission/admissiontimeout/timeoutadmission_test.go
```diff
@@ -35,13 +34,11 @@ func TestTimeoutAdmission(t *testing.T) {
...
 		name string
 
-		timeout         time.Duration
 		admissionPlugin func() (admit admitFunc, stopCh chan struct{})
 		expectedError   string
 	}{
 		{
-			name:    "stops on time",
-			timeout: 50 * time.Millisecond,
+			name: "stops on time",
```
tkashem, line 41: because we are passing `context.TODO()`, this test blocks for `13s`, right? If so, I think we should reduce test duration, maybe we can still keep `timeout         time.Duration` as a member variable of `pluginHandlerWithTimeout` so we can manipulate it from a test.

ardaguclu: Thanks @tkashem, you are right. I changed it to be configurable.

pkg/apiserver/admission/admissiontimeout/timeoutadmission.go
```diff
@@ -34,18 +37,21 @@ func (p pluginHandlerWithTimeout) Admit(ctx context.Context, a admission.Attribu
...
+		admissionErr = mutatingHandler.Admit(admissionCtx, a, o)
 	}()
 
 	select {
 	case <-admissionDone:
 		return admissionErr
-	case <-time.After(p.timeout):
-		return errors.NewInternalError(fmt.Errorf("admission plugin %q failed to complete mutation in %v", p.name, p.timeout))
+	case <-admissionCtx.Done():
+		return errors.NewInternalError(fmt.Errorf("admission plugin %q failed to complete mutation in %v", p.name, p.admissionTimeout))
```
tkashem, line 47: so i just realized that contexts are hierarchical - so if the parent context has a timeout of 5s and the child has a timeout of 13s, the child will timeout as soon as the parent times out after 5s.  so including `p.admissionTimeout` in the message may be misleading now.

we can  either remove the `admissionTimeout` from the error message or print the actual deadline of the context. thoughts?

ardaguclu: Indeed you are right https://github.com/golang/go/blob/6fefb7f9f3b632bdd0c3997ecc5b1096a5077cdf/src/context/context.go#L440. I learnt something new today.

ardaguclu: That information also concludes that nobody will not be able to increase admission timeout and will be able to decrease it.

ardaguclu: I think, it would be beneficial showing the timeout in error message. But it needs subtraction and some calculation because parent context has `time.Time` instead duration

tkashem:
cc @aojea

pkg/apiserver/admission/admissiontimeout/timeoutadmission_test.go
```diff
@@ -81,13 +81,16 @@ func TestTimeoutAdmission(t *testing.T) {
 			defer close(stopCh)
 
 			fakePlugin := dummyAdmit{admitFn: admitFn}
-			decorator := AdmissionTimeout{Timeout: test.timeout}
-			decoratedPlugin := decorator.WithTimeout(fakePlugin, "fake-name")
+			p := WithTimeout(fakePlugin, "fake-name").(pluginHandlerWithTimeout)
+			p.admissionTimeout = test.timeout
+			decorator := func(p pluginHandlerWithTimeout) admission.Interface {
+				return p
+			}(p)
```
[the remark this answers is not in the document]
ardaguclu: Good idea, thanks @aojea

pkg/apiserver/admission/admissiontimeout/timeoutadmission.go
aojea: the problem I've found is that this will race on the error message if the handler implements the context too, because it should err with a "context cancelled" error. However, we need to keep this out of the goroutine for handlers that doesn't implement context cancellation, what I did is to wait a few milliseconds more for the later

https://github.com/openshift/library-go/pull/1244/files#diff-bfc004e22663aa56dcd945b47c1776b0beebfdf2eb176044cf6130faebc63a4c

pkg/apiserver/admission/admissiontimeout/decorator.go
```diff
@@ -6,17 +6,13 @@ import (
 	"k8s.io/apiserver/pkg/admission"
 )
 
-// AdmissionTimeout provides a decorator that will fail an admission plugin after a certain amount of time
-//
-// DEPRECATED: use the context of the admission handler instead.
-type AdmissionTimeout struct {
```
aojea, line 12: why not keep this backwards compatible? 
otherwise , revendoring will need to change code in the consumers

aojea: I'm not familiar with this code, whatever you and @tkashem decide LGTM

pkg/apiserver/admission/admissiontimeout/timeoutadmission.go
aojea: I don't have any inconvenience :)

ardaguclu: Thanks :)

ardaguclu:
It would be better to remove deprecated field in another time.

/close

ardaguclu:
@tkashem @dgrisonnet could you PTAL this one?. Thanks.

ardaguclu:
There are merge conflicts. I'm closing this PR for future implementations.

/close

PR #1240, opened 2021-11-02
`pkg/route`: Add `ValidateSubdomain`

Miciah:
#### `pkg/route`: `ValidateHost`: Use `bool` for parameter

* `pkg/route/routeapihelpers/routeapihelpers.go` (`ValidateHost`): Change the `allowNonCompliant` parameter type from `string` to `bool`.
* `pkg/route/routeapihelpers/routeapihelpers_test.go` (`TestValidateRoute`): Rename...
(`TestValidateHost`): ...to this.  Update argument type.

#### `pkg/route`: Add `ValidateSubdomain`

* `pkg/route/routeapihelpers/routeapihelpers.go` (`ValidateSubdomain`): New function.  Validate that the given value is a valid DNS subdomain per RFC 1123.
* `pkg/route/routeapihelpers/routeapihelpers_test.go` (`TestValidateSubdomain`): Verify that `ValidateSubdomain` behaves as expected.

### DIFF
--- a/pkg/route/routeapihelpers/routeapihelpers.go
+++ b/pkg/route/routeapihelpers/routeapihelpers.go
@@ -4,6 +4,7 @@ package routeapihelpers
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/route/routeapihelpers/routeapihelpers.go
```diff
@@ -47,18 +48,16 @@ func IngressURI(route *routev1.Route, host string) (*url.URL, *routev1.RouteIngr
...
-
-	if allowNonCompliant == "true" {
+// Unless allowNonCompliant is true, host name must have at least two labels,
+// with each label no more than 63 characters from the set of alphanumeric
+// characters, '-' or '.', and must start and end with an alphanumeric
+// character. A trailing dot is allowed.  The total host name length must be no
+// more than 253 characters.  If allowNonCompliant is true, it uses a smaller
+// set of conditions from IsDNS1123Subdomain, e.g. character set as described
+// above, and total host name length must be no more than 253 characters.
+func ValidateHost(host string, allowNonCompliant bool, hostPath *field.Path) field.ErrorList {
```
candita, line 58: I think the annotation that was formerly used for this parameter is still set to a string, just from:
https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apis/route/validation/validation.go#L64

I couldn't find any references in documentation though.

Miciah: Semantically, `ValidateHost` really takes a Boolean value.  The fact that the caller may be deriving the value from an annotation is something that the caller can deal with, by converting the string annotation value into a Boolean value before calling `ValidateHost`.  In some cases, there is no annotation, and the caller just needs to be able to pass in a Boolean value.  For these reasons, I think that making the parameter a Boolean and requiring the caller to convert a string value (if that's what it has) into Boolean value for `ValidateHost` makes this function and its semantics cleaner.  

This isn't really related to `ValidateSubdomain`, though, so I can remove the change if you prefer.

PR #1241, opened 2021-11-03
Allow resource deletion based on a certain condition

ingvagabund:
With introduction of various topologies, some resources such as PDBs
have no use in single node installations. Other resources may lose
their purpose when certain condition holds. E.g. AWS machine controller
in absence of all nodes provisioned in AWS in a hybrid installation.

This PR adds Delete functions for various resources an defines a ConditionalFunction which can be used by controllers to conditionally delete/create a resource.


deads2k:
I reviewed in another PR.  merging the extracted easy bits

/lgtm

PR #1242, opened 2021-11-03
Feat: sno leader election config

eggfoobar:
Leader election in SNO clusters have limited resources where requests can put un-needed pressure on the `kube-apiserver`. This PR increases the retry period and lease duration on leader election and exposes some helper functions to determine topology and SNO leader election defaults, on operator start up.

### DIFF
--- a/pkg/config/clusterstatus/clusterstatus.go
+++ b/pkg/config/clusterstatus/clusterstatus.go
@@ -2,6 +2,7 @@ package clusterstatus
 
 import (
 	"context"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
 	openshiftcorev1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -21,13 +22,8 @@ func GetClusterInfraStatus(ctx context.Context, restClient *rest.Config) (*confi
 	if err != nil {
 		return nil, err
 	}
-	return &infra.Status, nil
-}
-
-func GetClusterInfraStatusOrDie(ctx context.Context, restClient *rest.Config) *configv1.InfrastructureStatus {
-	infra, err := GetClusterInfraStatus(ctx, restClient)
-	if err != nil {
-		panic(err)
+	if infra == nil {
+		return nil, fmt.Errorf("getting resource Infrastructure (name: %s) succeeded but object was nil", infraResourceName)
 	}
-	return infra
+	return &infra.Status, nil
 }
--- a/pkg/config/clusterstatus/clusterstatus.go
+++ b/pkg/config/clusterstatus/clusterstatus.go
@@ -1,0 +1,33 @@
+package clusterstatus
+
+import (
+	"context"
+
+	configv1 "github.com/openshift/api/config/v1"
+	openshiftcorev1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/rest"
+)
+
+const infraResourceName = "cluster"
+
+func GetClusterInfraStatus(ctx context.Context, restClient *rest.Config) (*configv1.InfrastructureStatus, error) {
+	client, err := openshiftcorev1.NewForConfig(restClient)
+	if err != nil {
+		return nil, err
+	}
+	infra, err := client.Infrastructures().Get(ctx, infraResourceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return &infra.Status, nil
+}
+
+func GetClusterInfraStatusOrDie(ctx context.Context, restClient *rest.Config) *configv1.InfrastructureStatus {
+	infra, err := GetClusterInfraStatus(ctx, restClient)
+	if err != nil {
+		panic(err)
+	}
+	return infra
+}
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -124,3 +124,31 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace, 
 	}
 	return ret
 }
+
+// LeaderElectionSNOConfig uses the formula derived in LeaderElectionDefaulting with increased
+// retry period and lease duration for SNO clusters that have limited resources.
+// This method does not respect the passed in LeaderElection config and the returned object will have values
+// that are overridden with SNO environments in mind.
+// This method should only be called when running in an SNO Cluster.
+func LeaderElectionSNOConfig(config configv1.LeaderElection) configv1.LeaderElection {
+
+	// We want to make sure we respect a 30s clock skew as well as a 4 retry attempt with out making
+	// leader election ineffectual while still having some small performance gain by limiting calls against
+	// the api server.
+
+	// 1. clock skew tolerance is leaseDuration-renewDeadline == 30s
+	// 2. kube-apiserver downtime tolerance is == 180s
+	//      lastRetry=floor(renewDeadline/retryPeriod)*retryPeriod == 240
+	//      downtimeTolerance = lastRetry-retryPeriod == 180s
+	// 3. worst non-graceful lease acquisition is leaseDuration+retryPeriod == 330s
+	// 4. worst graceful lease acquisition is retryPeriod == 60s
+
+	ret := *(&config).DeepCopy()
+	// 270-240 = 30s of clock skew tolerance
+	ret.LeaseDuration.Duration = 270 * time.Second
+	// 240/60 = 4 retries attempts before leader is lost.
+	ret.RenewDeadline.Duration = 240 * time.Second
+	// With 60s retry config we aim to maintain 30s of clock skew as well as 4 retry attempts.
+	ret.RetryPeriod.Duration = 60 * time.Second
+	return ret
+}
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -13,6 +13,7 @@ import (
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"github.com/openshift/library-go/pkg/authorization/hardcodedauthorizer"
 	"github.com/openshift/library-go/pkg/config/client"
+	"github.com/openshift/library-go/pkg/config/clusterstatus"
 	"github.com/openshift/library-go/pkg/config/configdefaults"
 	leaderelectionconverter "github.com/openshift/library-go/pkg/config/leaderelection"
 	"github.com/openshift/library-go/pkg/config/serving"
@@ -89,6 +90,10 @@ type ControllerBuilder struct {
 	// This stub exists for unit test where we can check if the graceful termination work properly.
 	// Default function will klog.Warning(args) and os.Exit(1).
 	nonZeroExitFn func(args ...interface{})
+
+	// Keep track if we defaulted leader election, used to make sure we don't stomp on the users intent for leader election
+	// We use this flag to determine at runtime if we can alter leader election for SNO configurations
+	userExplicitlySetLeaderElectionValues bool
 }
 
 // NewController returns a builder struct for constructing the command you want to run
@@ -141,6 +146,11 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	if leaderElection.Disable {
 		return b
 	}
+
+	// Set flag that SNO leader election configs can be used since user provided no timing configs
+	b.userExplicitlySetLeaderElectionValues = leaderElection.LeaseDuration.Duration != 0 ||
+		leaderElection.RenewDeadline.Duration != 0 ||
+		leaderElection.RetryPeriod.Duration != 0
 
 	defaulted := leaderelectionconverter.LeaderElectionDefaulting(leaderElection, defaultNamespace, defaultName)
 	b.leaderElection = &defaulted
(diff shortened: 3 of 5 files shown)


pkg/config/leaderelection/leaderelection.go
```diff
@@ -124,3 +124,13 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace,
 	}
 	return ret
 }
+
+// LeaderElectionSNOConfig uses the formula derived in LeaderElectionDefaulting with increased
+// retry period and lease duration for SNO clusters that have limited resources
+func LeaderElectionSNOConfig(config configv1.LeaderElection) configv1.LeaderElection {
```
deads2k, line 130: Can you run the math on these numbers to produce 
1. clock skew tolerance
2. kube-apiserver downtime tolerance
3. worst non-graceful lease acquisition
4. worst graceful lease acquisition

And add a comment to the code please.

eggfoobar:
To give a bit more context. Some of the numbers I was seeing are pasted below along with the queries ran against prometheus. The methodology was to create two 4.9 SNO-DU clusters (leader-election-new and leader-election-oc), both had CVO scaled down, and the only modification was that `leader-election-new` was running `11` operators that were rebuilt with the modified `library-go` running with the new leader-election numbers. I was seeing some small reductions, so I figured if we extrapolate that to the rest of the core operators we would see better gains in SNO-DU clusters, where we have very limited resources (currently management workloads are pinned to 2 cores).

Things I tried to verify was the downtime support for APIServer in SNO clusters described in this [bz](https://bugzilla.redhat.com/show_bug.cgi?id=1984730), rolling out APIServer didn't cause any restarts on the operators. Similarly rolling out the operator didn't cause any restarts and the operands behaved correctly with out any restarts as well.

### Queries Ran

#### 95th Quantile Over Time (CPU Usage) - API Server

| Unchanged          | Changed           | Delta   |
| ------------------ | ----------------- | ------- |
| 478.90296182259607 | 462.6560489574072 | `-3.39` |

```sh
quantile_over_time(
  .95,
  sum(
    rate(container_cpu_usage_seconds_total{
      container!="POD",
      pod=~"kube-apiserver-leader-election-oc-49|kube-apiserver-leader-election-new-49",
      namespace="openshift-kube-apiserver"}[5m]
    )
  )[6h:]
) * 1000
```

#### Average Over Time (CPU Usage) - API Server

| Unchanged          | Changed           | Delta   |
| ------------------ | ----------------- | ------- |
| 429.08792083196755 | 395.3039245479713 | `-7.87` |

```sh
avg_over_time(
  sum(
    rate(container_cpu_usage_seconds_total{
      container!="POD",
      pod=~"kube-apiserver-leader-election-oc-49|kube-apiserver-leader-election-new-49",
      namespace="openshift-kube-apiserver"}[5m]
    )
  )[6h:]
) * 1000

```

#### 95th Quantile Over Time (CPU Usage) - Operators

| Unchanged          | Changed           | Delta   |
| ------------------ | ----------------- | ------- |
| 192.51093951592614 | 174.9938822722223 | `-9.10` |

```sh
quantile_over_time(
  .95,
  sum(
    rate(container_cpu_usage_seconds_total{
      container!="POD",
      namespace=~"openshift-apiserver-operator|openshift-authentication-operator|openshift-cluster-storage-operator|openshift-console-operator|openshift-controller-manager-operator|openshift-etcd-operator|openshift-kube-apiserver-operator|openshift-kube-controller-manager-operator|openshift-kube-scheduler-operator|openshift-kube-storage-version-migrator-operator|openshift-service-ca-operator"
      }[5m]
    )
  )[6h:]
) * 1000
```

#### Average Over Time (CPU Usage) - Operators

| Unchanged          | Changed            | Delta   |
| ------------------ | ------------------ | ------- |
| 173.11949276437636 | 159.50446507412303 | `-7.86` |

```sh
avg_over_time(
  sum(
    rate(container_cpu_usage_seconds_total{
      container!="POD",
      namespace=~"openshift-apiserver-operator|openshift-authentication-operator|openshift-cluster-storage-operator|openshift-console-operator|openshift-controller-manager-operator|openshift-etcd-operator|openshift-kube-apiserver-operator|openshift-kube-controller-manager-operator|openshift-kube-scheduler-operator|openshift-kube-storage-version-migrator-operator|openshift-service-ca-operator"
      }[5m]
    )
  )[6h:]
) * 1000
```

pkg/config/leaderelection/leaderelection.go
eggfoobar: Absolutely, added.

deads2k:
Summarizing the overall benefit to an SNO cluster over X timeframe will be helpful for weighing the tradeoff of slightly lower CPU cost at the expense of latency of lease reacquisition.

eggfoobar:
Sure thing, a quick summary with some new numbers. I ran the clusters under load. I used [kube-burner](https://github.com/cloud-bulldozer/kube-burner) to spin up some traffic against the API Server and just have some workloads in the cluster. The latest numbers show a pretty consistent picture of a small reduction in CPU consumption by the Changed Operators. On average by increasing leader election to 60s I'm seeing a reduction of `1-2 millicores` per operator, extrapolating that to the 30 operators we run we could see `30millicore` reduction on on average on the low end and `60millicore` on the high end. That is a very small reduction but the gain is meaningful in environments where there is significant reduction in resources. Some DU configurations for example are pinning management workloads to 2 cores. This gives them very little breathing room and any measurable reduction is useful to provide more resources for Crio and Kubelet to operate. This does mean we reduce the responsiveness of our operators during rollouts, but the gain would mean that if need be we can support another operator in those limited environments.

In short, with the current numbers, we would have a worst non-graceful case of `330s` before the new operator takes leader and a best graceful case of `60s` which would mean a latency in lease reacquisition. The upside is that this would afford resource starved deployments some more breathing room in SNO environments (all be it small), which would offer Kubelet and Crio more room or potentially allow running another operator.

As always with things of this nature, another pair of eyes would be fantastic on the numbers, queries and the methodology. Any input is greatly appreciated.

pkg/config/leaderelection/leaderelection.go
```diff
@@ -124,3 +124,31 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace,
...
 }
+
+// LeaderElectionSNOConfig uses the formula derived in LeaderElectionDefaulting with increased
+// retry period and lease duration for SNO clusters that have limited resources.
+// This method does not respect the passed in LeaderElection config and the returned object will have values
+// that are overridden with SNO environments in mind.
+// This method should only be called when running in an SNO Cluster.
+func LeaderElectionSNOConfig(config configv1.LeaderElection) configv1.LeaderElection {
+
+	// We want to make sure we respect a 30s clock skew as well as a 4 retry attempt with out making
```
hexfusion, line 135: Can you provide color into the 30s skew and why this duration was chosen? In the case of SNO where we are essentially talking about a single clock on the control plane, it's not directly apparent where the skew would come from in regards to a lease.

eggfoobar: That is a great call out, the idea there was to keep that mostly intact as it's not clear to me if there will be SNO clusters that might have a worker machine and what that arrangement might be in terms of where operators might fall. Figured to err on the side of caution and keep with the set convention.

pkg/controller/controllercmd/builder.go
```diff
@@ -304,6 +305,15 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
 	}
 
+	infraStatus, err := clusterstatus.GetClusterInfraStatus(ctx, clientConfig)
+	if err != nil {
+		eventRecorder.Warningf("ClusterInfrastructureStatus", "Failed to get cluster infrastructure status because of %s", err)
+		klog.Warningf("unable to get cluster infrastructure status for leader election (falling back to default): %v", err)
+	} else if infraStatus.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
+		snoLeaderElection := leaderelectionconverter.LeaderElectionSNOConfig(*b.leaderElection)
+		b.leaderElection = &snoLeaderElection
+	}
```
soltysh, line 315: Shouldn't that block move to `WithleaderElection` since it's possible that LE is disabled and this code can accidentally turn it on if we're not careful enough.

soltysh (commented):
One nit, otherwise this lgtm

pkg/controller/controllercmd/builder.go
eggfoobar: That's a good point, the reason why it's there is because we won't know the topology we're in until runtime, the function currently won't alter the `disable` flag in LE, but to make it more robust I changed the if statement to be `if topology = SingleReplica AND LE is not disabled` as well as some tests to ensure behavior, do think that's a good enough safeguard against accidentally turnning it on?

soltysh (approved):
/lgtm
deferring to @deads2k for approval

pkg/controller/controllercmd/builder.go
```diff
@@ -304,6 +305,15 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		return nil
 	}
 
+	infraStatus, err := clusterstatus.GetClusterInfraStatus(ctx, clientConfig)
+	if err != nil {
+		eventRecorder.Warningf("ClusterInfrastructureStatus", "Failed to get cluster infrastructure status because of %s", err)
+		klog.Warningf("unable to get cluster infrastructure status for leader election (falling back to default): %v", err)
+	} else if infraStatus.ControlPlaneTopology == configv1.SingleReplicaTopologyMode && !b.leaderElection.Disable {
+		snoLeaderElection := leaderelectionconverter.LeaderElectionSNOConfig(*b.leaderElection)
+		b.leaderElection = &snoLeaderElection
```
deads2k, line 314: this is stomping the code configured leader election if someone specified their actual values.  Only in `WithleaderElection` do you have sufficient information to know what the user provided so you don't overwrite it.

How you resolve this is up to you/@soltysh, but you cannot unconditionally stomp the user's intent.

pkg/controller/controllercmd/builder.go
```diff
@@ -304,6 +305,15 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		return nil
 	}
 
+	infraStatus, err := clusterstatus.GetClusterInfraStatus(ctx, clientConfig)
+	if err != nil {
+		eventRecorder.Warningf("ClusterInfrastructureStatus", "Failed to get cluster infrastructure status because of %s", err)
+		klog.Warningf("unable to get cluster infrastructure status for leader election (falling back to default): %v", err)
+	} else if infraStatus.ControlPlaneTopology == configv1.SingleReplicaTopologyMode && !b.leaderElection.Disable {
```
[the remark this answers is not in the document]
eggfoobar: thank you! fixed

pkg/controller/controllercmd/builder.go
eggfoobar: I altered this a bit, I added a flag approach to keep it straight forward and introduce less things. The idea being if the user provides no timing configs in the  `WithleaderElection` then it's safe to assume we can change those configs, so we fetch the infraStatus to see if we should alter the timings for SNO. @soltysh What do you think about this approach?

soltysh: > How you resolve this is up to you/@soltysh, but you cannot unconditionally stomp the user's intent.

@deads2k that was also my suggestion here: https://github.com/openshift/library-go/pull/1242/files#r754222946

eggfoobar: With the recent [change](https://github.com/openshift/library-go/pull/1242/files#diff-e78bc214728945a843063f3f0d995c03dd7490a6956486055cc0ce4b965dd806R151 ) @soltysh , I'm using the 0 values in the timings to ensure that if the user has decided to have some customizations in the timings then we won't alter the timings during runtime. That should make sure we don't stomp on the users intent, and if no values were provided, then we can query the cluster and change the values if SNO is detected. That's going to be triggered [here](https://github.com/openshift/library-go/pull/1242/files#diff-e78bc214728945a843063f3f0d995c03dd7490a6956486055cc0ce4b965dd806R317)

pkg/config/clusterstatus/clusterstatus.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if infra == nil {
+		return nil, fmt.Errorf("getting resource Infrastructure (name: %s) succeeded but object was nil", infraResourceName)
+	}
+	return &infra.Status, nil
+}
+
+func GetClusterInfraStatusOrDie(ctx context.Context, restClient *rest.Config) *configv1.InfrastructureStatus {
```
deads2k, line 31: this function shouldn't exist

pkg/controller/controllercmd/builder.go
deads2k: This code is still in a spot that does not have access to the user's intent.  @soltysh's comment looks unaddressed to me.

pkg/controller/controllercmd/builder.go
```diff
@@ -142,6 +147,11 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 		return b
 	}
 
+	// Set flag that SNO leader election configs can be used since user provided no timing configs
+	b.canUseSNOLeaderElection = (leaderElection.LeaseDuration.Duration == 0 &&
```
deads2k, line 151: This is actually `userExplicitlySetLeaderElectionValues`

pkg/controller/controllercmd/builder.go
```diff
@@ -304,6 +314,17 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
 		return nil
 	}
 
+	if b.canUseSNOLeaderElection {
```
deads2k, line 317: turning this into `if !b,userExplicitlySetLeaderElectionValues`

pkg/controller/controllercmd/builder.go
```diff
@@ -304,6 +314,17 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+		infraStatus, err := clusterstatus.GetClusterInfraStatus(ctx, clientConfig)
+		if err != nil || infraStatus == nil {
+			eventRecorder.Warningf("ClusterInfrastructureStatus", "Failed to get cluster infrastructure status because of %s", err)
+			klog.Warningf("unable to get cluster infrastructure status for leader election (falling back to default): %v", err)
```
deads2k, line 321: event and warning should match and I think

`klog.Warningf("unable to get cluster infrastructure status, using HA cluster values for leader election: %v", err)`

pkg/config/clusterstatus/clusterstatus.go
eggfoobar: sounds good, removed function

pkg/controller/controllercmd/builder.go
eggfoobar: done, updated logic

pkg/controller/controllercmd/builder.go
eggfoobar: updated if statement

pkg/controller/controllercmd/builder.go
eggfoobar: I like that message too, updated logging and event messages to match

deads2k:
/approve

please squash before merge.

deads2k:
diff of diffs is clean https://github.com/openshift/library-go/compare/aa75fd24bdd8b2b1c798a1237bc27364a16bfac0..2612981f3019479805ac8448b997266fc07a236a

/lgtm

PR #1243, opened 2021-11-04
Add conditional resource synchronization to the resource sync controller

dgrisonnet:
Controllers behind optional features such as the encryption controller
which is disabled by default for all OpenShift apiservers, have no way
to synchronize secrets/configmaps conditionally. For instance, the
encryption controller tried to synchronize its encryption-config very
often even though encryption is disabled. To mitigate that, some
controllers have some logic to set the SyncResource source
conditionally. An empty source meaning that the sync controller will
skip syncing the object. Since this is very hacky, it would be best to
have preconditions that need to be fulfilled before synchronizing
resources.

To improve that, the ResourceSyncer interface was extended to
add SyncConfigMapCondtionally and SyncSecretConditionally
functions. This allows adding conditional syncs per resource to the
original resource sync controller instead of adding a new controller
dedicated for that.

Related Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1982726
Cause of the bug: the static pod deployer always sync the encryption config secret even when the feature is disabled: https://github.com/openshift/library-go/blob/d429589a7691e57aedcf382d251e6e0ebb4ed0bb/pkg/operator/encryption/deployer/revisionedpod.go#L72-L77
Solution: sync conditionnaly with a controller

Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1250

### DIFF
--- a/pkg/operator/encryption/controllers.go
+++ b/pkg/operator/encryption/controllers.go
@@ -2,11 +2,14 @@ package encryption
 
 import (
 	"context"
+	"fmt"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	"github.com/openshift/library-go/pkg/operator/encryption/controllers/migrators"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
+	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
@@ -35,6 +38,7 @@ func NewControllers(
 	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
 	secretsClient corev1.SecretsGetter,
 	eventRecorder events.Recorder,
+	resourceSyncer *resourcesynccontroller.ResourceSyncController,
 ) (*Controllers, error) {
 	// avoid using the CachedSecretGetter as we need strong guarantees that our encryptionSecretSelector works
 	// otherwise we could see secrets from a different component (which will break our keyID invariants)
@@ -45,6 +49,17 @@ func NewControllers(
 	encryptionEnabledChecker, err := newEncryptionEnabledPrecondition(apiServerInformer.Lister(), kubeInformersForNamespaces, encryptionSecretSelector.LabelSelector, component)
 	if err != nil {
 		return nil, err
+	}
+
+	// for testing resourceSyncer might be nil
+	if resourceSyncer != nil {
+		if err := resourceSyncer.SyncSecretConditionally(
+			resourcesynccontroller.ResourceLocation{Namespace: component, Name: encryptionconfig.EncryptionConfSecretName},
+			resourcesynccontroller.ResourceLocation{Namespace: "openshift-config-managed", Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, component)},
+			encryptionEnabledChecker.PreconditionFulfilled,
+		); err != nil {
+			return nil, err
+		}
 	}
 
 	return &Controllers{
--- a/pkg/operator/encryption/deployer/revisionedpod.go
+++ b/pkg/operator/encryption/deployer/revisionedpod.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptionconfig"
 	"github.com/openshift/library-go/pkg/operator/encryption/statemachine"
-	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
@@ -53,30 +52,17 @@ var (
 // a label storing the deployed encryption config revision, like the pods created
 // by the staticpod controllers.
 //
-// It syns the encryption-config-<targetNamespace> from openshift-config-managed
-// namespace to the target namespace as encryption-config. From there it is
-// revisioned and deployed to the static pods. The last deployed encryption
-// config is read from encryption-config-<revision>.
-//
-// For testing, resourceSyncer might be nil.
+// It revisiones and deployes the synchronized encryption-config from the
+// operator namespace to the static pods. The last deployed encryption config is
+// read from encryption-config-<revision>.
 func NewRevisionLabelPodDeployer(
 	revisionLabel string,
 	targetNamespace string,
 	namespaceInformers operatorv1helpers.KubeInformersForNamespaces,
-	resourceSyncer resourcesynccontroller.ResourceSyncer,
 	podClient corev1client.PodsGetter,
 	secretClient corev1client.SecretsGetter,
 	nodeProvider MasterNodeProvider,
 ) (*RevisionLabelPodDeployer, error) {
-	if resourceSyncer != nil {
-		if err := resourceSyncer.SyncSecret(
-			resourcesynccontroller.ResourceLocation{Namespace: targetNamespace, Name: encryptionconfig.EncryptionConfSecretName},
-			resourcesynccontroller.ResourceLocation{Namespace: "openshift-config-managed", Name: fmt.Sprintf("%s-%s", encryptionconfig.EncryptionConfSecretName, targetNamespace)},
-		); err != nil {
-			return nil, err
-		}
-	}
-
 	return &RevisionLabelPodDeployer{
 		podClient:                podClient.Pods(targetNamespace),
 		secretClient:             secretClient.Secrets(targetNamespace),
--- a/pkg/operator/resourcesynccontroller/interfaces.go
+++ b/pkg/operator/resourcesynccontroller/interfaces.go
@@ -12,9 +12,16 @@ type ResourceLocation struct {
 	Provider string `json:"provider,omitempty"`
 }
 
+// PreconditionsFulfilled is a function that indicates whether all prerequisites
+// are met and a resource can be synced.
+type preconditionsFulfilled func() (bool, error)
+
+func alwaysFulfilledPreconditions() (bool, error) { return true, nil }
+
 type syncRuleSource struct {
 	ResourceLocation
-	syncedKeys sets.String // defines the set of keys to sync from source to dest
+	syncedKeys               sets.String            // defines the set of keys to sync from source to dest
+	preconditionsFulfilledFn preconditionsFulfilled // preconditions to fulfill before syncing the resource
 }
 
 type syncRules map[ResourceLocation]syncRuleSource
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -100,10 +100,20 @@ func (c *ResourceSyncController) Name() string {
 }
 
 func (c *ResourceSyncController) SyncConfigMap(destination, source ResourceLocation) error {
-	return c.SyncPartialConfigMap(destination, source)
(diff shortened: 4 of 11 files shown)


pkg/operator/encryption/controllers.go
```diff
@@ -34,6 +37,7 @@ func NewControllers(
 	apiServerInformer configv1informers.APIServerInformer,
 	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
 	secretsClient corev1.SecretsGetter,
+	configmapsClient corev1.ConfigMapsGetter,
```
dgrisonnet, line 40: We need to add a configmap client when creating the resource syncer even though we don't need to sync configmaps in the encryption controller. Any suggestion on how this could be improved?

pkg/operator/resourcesynccontroller/conditional_resourcesync_controller.go
```diff
@@ -0,0 +1,83 @@
...
+	*ResourceSyncController
+
+	preconditionsFulfilledFn preconditionsFulfilled
+}
+
+var _ ResourceSyncer = &ConditionalResourceSyncController{}
+var _ factory.Controller = &ConditionalResourceSyncController{}
+
+// NewConditionalResourceSyncController creates NewConditionalResourceSyncController.
+func NewConditionalResourceSyncController(
```
sttts, line 32: couldn't this also be by resource and not with another controller? How do both implementation compare?

pkg/operator/encryption/deployer/revisionedpod.go
```diff
@@ -53,30 +52,17 @@ var (
...
 func NewRevisionLabelPodDeployer(
 	revisionLabel string,
 	targetNamespace string,
 	namespaceInformers operatorv1helpers.KubeInformersForNamespaces,
-	resourceSyncer resourcesynccontroller.ResourceSyncer,
 	podClient corev1client.PodsGetter,
 	secretClient corev1client.SecretsGetter,
 	nodeProvider MasterNodeProvider,
 ) (*RevisionLabelPodDeployer, error) {
-	if resourceSyncer != nil {
```
p0lyn0mial, line 71: I think that we could rely on the fact that all encryption controllers accept a `PreconditionFulfilled` function that dictates usage of a deployer only when the function returns true (all preconditions are met). 

We could change the deployer to lazily initialize `resourceSyncer` instead of adding a new controller. Does it make sense?

pkg/operator/resourcesynccontroller/conditional_resourcesync_controller.go
dgrisonnet: I think it would make sense to have by resource instead of defining one global condition for the whole controller. Going this route there wouldn't be much difference between both controllers except that one adds a condition per resource, so I think it would make sense to extend the original resource sync controller instead of adding a new one.

pkg/operator/encryption/deployer/revisionedpod.go
p0lyn0mial: changing `resourceSyncer.SyncSecret` to accept a precondition function also sounds good to me.

pkg/operator/resourcesynccontroller/conditional_resourcesync_controller.go
dgrisonnet: I updated the PR to extend the ResourceSyncer interface and add the `SyncResourceConditionally` functions. PTAL @sttts @p0lyn0mial

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -25,10 +32,22 @@ var (
...
+	SyncConfigMapConditionally(destination, source ResourceLocation, precopreconditionsFulfilledFn PreconditionsFulfilled) error
+	// SyncSecret indicates that a secret should be copied from the source to the
+	// destination.  It will also mirror a deletion from the source.  If the
+	// source is a zero object the destination will be deleted.
 	SyncSecret(destination, source ResourceLocation) error
+	// SyncSecretCondtionally conditionally indicates that a secret should be
+	// copied from the source to the destination.  It will also mirror a deletion
+	// from the source.  If the source is a zero object the destination will be
+	// deleted.
+	SyncSecretConditionally(destination, source ResourceLocation, precopreconditionsFulfilledFn PreconditionsFulfilled) error
```
p0lyn0mial, line 52: Would it make sense to pass a precondition function during the `NewResourceSyncController` construction? 

We wouldn't have to change the interface, and the precondition would apply to all resources.

pkg/operator/encryption/controllers.go
```diff
@@ -35,6 +38,7 @@ func NewControllers(
 	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
 	secretsClient corev1.SecretsGetter,
 	eventRecorder events.Recorder,
+	resourceSyncer resourcesynccontroller.ResourceSyncer,
```
p0lyn0mial, line 41: could we pass the concrete obj not an interface so that we don't have to change the `ResourceSyncer` interface?

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -124,10 +129,14 @@ func (c *ResourceSyncController) SyncPartialConfigMap(destination ResourceLocati
...
-	return c.SyncPartialSecret(destination, source)
+	return c.SyncPartialSecret(destination, source, noPreconditions)
+}
+
+func (c *ResourceSyncController) SyncSecretConditionally(destination, source ResourceLocation, preconditionsFulfilledFn PreconditionsFulfilled) error {
+	return c.SyncPartialSecret(destination, source, preconditionsFulfilledFn)
 }
 
-func (c *ResourceSyncController) SyncPartialSecret(destination, source ResourceLocation, keys ...string) error {
+func (c *ResourceSyncController) SyncPartialSecret(destination, source ResourceLocation, preconditionsFulfilledFn PreconditionsFulfilled, keys ...string) error {
```
p0lyn0mial, line 139: this is a breaking change, wondering if there is a way to avoid it?

pkg/operator/resourcesynccontroller/interfaces.go
dgrisonnet: If we do that, we won't be able to add preconditions per resource, so we would have to create a new resource sync controller inside of the encryption controller instead of reusing the common one.

pkg/operator/encryption/controllers.go
dgrisonnet: Done

pkg/operator/resourcesynccontroller/resourcesync_controller.go
dgrisonnet: Good catch, I change that and kept the initial signature of the function and used an intermediate function to do the computation: 
https://github.com/openshift/library-go/pull/1243/files#diff-c72f483f7c5a6fb8b8c216cb67361320b8c81865eede67417512622c452f8b9eR114

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -12,9 +12,16 @@ type ResourceLocation struct {
 	Provider string `json:"provider,omitempty"`
 }
 
+// PreconditionsFulfilled is a function that indicates whether all prerequisites
+// are met and the controller can Sync.
```
p0lyn0mial, line 16: nit: `and a resource can be synced`

pkg/operator/resourcesynccontroller/interfaces.go
```diff
@@ -12,9 +12,16 @@ type ResourceLocation struct {
...
+type preconditionsFulfilled func() (bool, error)
+
+func noPreconditions() (bool, error) { return true, nil }
```
p0lyn0mial, line 19: nit: `alwaysFulfilledPreconditions`

pkg/operator/resourcesynccontroller/resourcesync_controller_test.go
```diff
@@ -249,6 +249,86 @@ func TestSyncConfigMap(t *testing.T) {
...
+					ManagementState: operatorv1.Managed,
+				},
+				&operatorv1.OperatorStatus{},
+				nil,
+			)
+			eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), "test-operator", &corev1.ObjectReference{})
+
+			c := NewResourceSyncController(
+				fakeStaticPodOperatorClient,
+				v1helpers.NewFakeKubeInformersForNamespaces(map[string]informers.SharedInformerFactory{
```
p0lyn0mial, line 293: nit: I think you could create `ResourceSyncController` directly and avoid passing/creating informers

pkg/operator/encryption/controllers.go
```diff
@@ -47,6 +51,17 @@ func NewControllers(
 		return nil, err
 	}
 
+	// for testing resourceSyncer might be nil
+	if resourceSyncer != nil {
+		if err := resourceSyncer.SyncSecretConditionally(
```
p0lyn0mial, line 56: the previous location was also okay.

dgrisonnet: we don't have all the information needed by the precondition in the deployer. We would have to pass new parameters just for that so I think it is better to just move the logic to the encryption controller.

pkg/operator/resourcesynccontroller/interfaces.go
dgrisonnet: done

pkg/operator/resourcesynccontroller/interfaces.go
dgrisonnet: done

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -124,10 +133,18 @@ func (c *ResourceSyncController) SyncPartialConfigMap(destination ResourceLocati
...
 func (c *ResourceSyncController) SyncSecret(destination, source ResourceLocation) error {
-	return c.SyncPartialSecret(destination, source)
+	return c.syncSecret(destination, source, noPreconditions)
 }
 
 func (c *ResourceSyncController) SyncPartialSecret(destination, source ResourceLocation, keys ...string) error {
+	return c.syncSecret(destination, source, noPreconditions, keys...)
+}
+
+func (c *ResourceSyncController) SyncSecretConditionally(destination, source ResourceLocation, preconditionsFulfilledFn preconditionsFulfilled) error {
```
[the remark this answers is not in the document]
dgrisonnet: done

pkg/operator/resourcesynccontroller/resourcesync_controller_test.go
dgrisonnet: we can't because we need to watch the source and destination namespaces to synchronise resources: https://github.com/openshift/library-go/blob/master/pkg/operator/resourcesynccontroller/resourcesync_controller.go#L131-L136

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -100,10 +100,20 @@ func (c *ResourceSyncController) Name() string {
...
 
 func (c *ResourceSyncController) SyncPartialConfigMap(destination ResourceLocation, source ResourceLocation, keys ...string) error {
+	return c.syncConfigMap(destination, source, alwaysFulfilledPreconditions, keys...)
+}
+
+// SyncConfigMapConditionally adds a new configmap that the resource sync
+// controller will synchronise if the given precondition is fulfilled.
+func (c *ResourceSyncController) SyncConfigMapConditionally(destination, source ResourceLocation, preconditionsFulfilledFn preconditionsFulfilled) error {
+	return c.syncConfigMap(destination, source, preconditionsFulfilledFn)
+}
```
sttts, line 114: what happens if I want a partial configmap conditionally? Is there a better pattern?

p0lyn0mial: atm it is not supported :) we can get back to it once there is a use case

PR #1244, opened 2021-11-08
pass down context timeout on admission timeout plugin

aojea:
the admission timeout plugin was not cascading down the timeout to the plugins, with the risk of leaking goroutines.


aojea:
dup of https://github.com/openshift/library-go/pull/1239

PR #1245, opened 2021-11-08
installerpod: skip installer pod UID generation when running on newer kubelets

mfojtik:
Ff the node is >=4.9, the pod UID does not need to be set in the file. Setting it to static UID might trigger a rare case where kubelet does not properly handle graceful shutdown and might lead to a case when regular pods go from "running" to "pending" state.

xref: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/26572/pull-ci-openshift-origin-master-e2e-gcp-upgrade/1456620463206699008

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/blang/semver"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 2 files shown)


deads2k:
This looks plausible to me and per the node team should resolve the direct pending to running state change for our static pods.  My intuition suggests that this fully separates previous pod from current pod, so the file locks are likely to become critical.

@sttts did we get the audit log separation to happen and the port conflict resolved?

/assign @sttts

sttts:
> @sttts did we get the audit log separation to happen and the port conflict resolved?

We have a lock in place to stop harm. The author of the lock and this PR are equal.

sttts:
/hold

Please add a PR description. Why do we need this PR.

PR #1249, opened 2021-11-16
pkg: fix apiservice discovery context panic

dgrisonnet:
Fix panic seen in https://github.com/openshift/cluster-openshift-apiserver-operator/pull/483#issuecomment-970532608, after pulling the changes from https://github.com/openshift/library-go/pull/1142.

### DIFF
--- a/pkg/operator/apiserver/controller/apiservice/apigroup.go
+++ b/pkg/operator/apiserver/controller/apiservice/apigroup.go
@@ -101,7 +101,7 @@ func checkDiscoveryForAPIService(ctx context.Context, restclient rest.Interface,
 			defer wg.Done()
 			defer utilruntime.HandleCrash()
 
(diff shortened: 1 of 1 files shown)


pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -101,7 +101,7 @@ func checkDiscoveryForAPIService(ctx context.Context, restclient rest.Interface,
 			defer wg.Done()
 			defer utilruntime.HandleCrash()
 
-			discoveryCtx, ctxCancelFn = context.WithTimeout(discoveryCtx, 25*time.Second)
+			discoveryCtx, ctxCancelFn = context.WithTimeout(ctx, 25*time.Second)
```
p0lyn0mial, line 104: a typo, `discoveryCtx` is always `nil`. thanks for fixing!

PR #1251, opened 2021-11-18
WIP: staticpods: make installer controller avoid down nodes after a day

sttts:
Addressing stuck, expired certificates after 30d of one master's downtime.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -27,11 +27,14 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/clock"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -286,36 +292,58 @@ type staticPodStateFunc func(ctx context.Context, nodeName string) (state static
 // - not updating
 // - ready
 // - at the revision claimed in CurrentRevision.
-func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodStateFunc, nodes []operatorv1.NodeStatus) (int, string, error) {
-	if len(nodes) == 0 {
-		return 0, "", fmt.Errorf("nodes array cannot be empty")
+func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodStateFunc, nodeStatuses []operatorv1.NodeStatus, nodes []*corev1.Node) (int, string, error) {
```
p0lyn0mial, line 295: could we document the order in which the nodes will be processed?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -286,36 +292,58 @@ type staticPodStateFunc func(ctx context.Context, nodeName string) (state static
...
+		}
+		nodeStatusesUpFirst = append(nodeStatusesUpFirst, status)
 	}
 
-	// find upgrading node as this will be the first to start new revision (to minimize number of down nodes)
-	for i := range nodes {
-		if nodes[i].TargetRevision != 0 {
-			reason := fmt.Sprintf("node %s is progressing towards %d", nodes[i].NodeName, nodes[i].TargetRevision)
+	// find upgrading node as this will be the first to start new revision (to minimize number of down nodeStatuses)
+	for i := range nodeStatuses {
```
p0lyn0mial, line 323: shouldn't you use `nodeStatusesUpFirst` ?

p0lyn0mial:
We should also understand where did we fail in case one node was down. 
Was it here https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L351 ?

pkg/operator/staticpod/controller/installer/installer_controller.go
sttts: done

PR #1255, opened 2021-11-26
bug 2026352: staticpod pruner: check if the cert directory exists

ingvagabund:
When a new master node is provisioned, there's no /etc/kubernetes/static-pod-resources/kube-scheduler-certs directory existing on the node yet. Given there's no sync between when a pruner pod and an installer pod gets to run, it may happen the pruner pod gets to run before the installer pod. In which case the certs directory is missing. Leading to a panic since the pruner expects the certs directory to exists. The directory gets eventually created but the panic is not user friendly.

### DIFF
--- a/pkg/operator/staticpod/prune/cmd.go
+++ b/pkg/operator/staticpod/prune/cmd.go
@@ -128,6 +128,13 @@ func (o *PruneOptions) Run() error {
 		return nil
 	}
 
+	// If the cert dir does not exist, do nothing.
+	// The dir will get eventually created by an installer pod.
+	if _, err := os.Stat(path.Join(o.ResourceDir, o.CertDir)); os.IsNotExist(err) {
+		klog.Infof("Skipping %s as it does not exist", path.Join(o.ResourceDir, o.CertDir))
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/prune/cmd.go
```diff
@@ -128,6 +128,12 @@ func (o *PruneOptions) Run() error {
 		return nil
 	}
 
+	// If the cert dir does not exist, do nothing.
+	// The dir will get eventually created by an installer pod.
+	if _, err := os.Stat(path.Join(o.ResourceDir, o.CertDir)); os.IsNotExist(err) {
```
p0lyn0mial, line 133: I think that `filepath.Walk` doesn't panic, check out - https://go.dev/play/p/FH3D3rPcQO7

Do you have a stacktrace?

ingvagabund: It's not the filepath.Walk which panics. It is https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/prune/cmd.go#L48. If unchecked the Run() returns an error for non-existing directory. Whereas, it's better to return nil when the directory does not exist.

PR #1256, opened 2021-12-01
resourceapply: don't require generation for applying *WebhookConfiguration resources

bertinatto:
Currently, whoever uses the `staticresourcecontroller` to roll out webhook configuration objects will get this test failure in CI:

```
[sig-arch] events should not repeat pathologically 

1 events happened too frequently  event happened 585 times, something is wrong: ns/openshift-cluster-csi-drivers deployment/vmware-vsphere-csi-driver-operator - reason/ValidatingWebhookConfigurationUpdated Updated ValidatingWebhookConfiguration.admissionregistration.k8s.io/validation.csi.vsphere.vmware.com because it changed
```

I see 2 ways of fixing this:

1. Special-case negative values for the expected generation, so that [this](https://github.com/openshift/library-go/blob/c58f31ee6d3a5b97360175fa073a7eec6ccd1b79/pkg/operator/resource/resourceapply/generic.go#L215) won't constantly update the object.

2. Remove the expected generation argument altogether from the `Apply[Validating,Mutating]WebhookConfiguration()` functions. This might break some users ([here](https://github.com/openshift/cluster-baremetal-operator/blob/91c6c71b6f8219be8a7bd431d71a43afbfe16dfb/provisioning/baremetal_webhook.go#L42) and [here](https://github.com/openshift/machine-api-operator/blob/8525494f83c66da81c96bdf635f168bd7f3ff086/pkg/operator/sync.go#L196-L224)).

Edit: we chose option 2 above (see https://github.com/openshift/library-go/pull/1256#issuecomment-984581830).

### DIFF
--- a/pkg/operator/resource/resourceapply/admissionregistration.go
+++ b/pkg/operator/resource/resourceapply/admissionregistration.go
@@ -21,7 +21,7 @@ import (
 // and an update performed if the mutatingwebhookconfiguration spec and metadata differ from
 // the previously required spec and metadata based on generation change.
 func ApplyMutatingWebhookConfiguration(ctx context.Context, client admissionregistrationclientv1.MutatingWebhookConfigurationsGetter, recorder events.Recorder,
-	requiredOriginal *admissionregistrationv1.MutatingWebhookConfiguration, expectedGeneration int64) (*admissionregistrationv1.MutatingWebhookConfiguration, bool, error) {
+	requiredOriginal *admissionregistrationv1.MutatingWebhookConfiguration) (*admissionregistrationv1.MutatingWebhookConfiguration, bool, error) {
 
 	if requiredOriginal == nil {
 		return nil, false, fmt.Errorf("Unexpected nil instead of an object")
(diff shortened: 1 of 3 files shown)


bertinatto:
I want to start a discussion about fixing this: https://github.com/openshift/library-go/pull/1129#discussion_r669819849

@slaskawi @deads2k can you chime in? Please check the PR description for possible approaches.

deads2k:
That test is startlingly useful! :)

`expectedGeneration` was added for applying deployments and daemonsets to handle cases where defaulting changed objects.  I suggest removing it from the apply*admission calls and letting the callers work out fixing it when they rebase.

bertinatto:
/test 

> That test is startlingly useful! :)
> 
> `expectedGeneration` was added for applying deployments and daemonsets to handle cases where defaulting changed objects. I suggest removing it from the apply*admission calls and letting the callers work out fixing it when they rebase.

Done, thanks!

deads2k:
We used generations for deployments and daemonsets to handle interaction with other defaulting and controller actors, but it doesn't seem required for admission webhooks.

/lgtm
/approve

PR #1261, opened 2021-12-06
bug 2029470: installerpod: do not copy resources if a target directory with a higher revision already exists

ingvagabund:
It is possible to get to a situation when an installer pod of a given revision (e.g. 6) is requested to be deleted and new installer pod of a higher revision (e.g. 7) is created. Yet, the container engine does not delete the installer container of revision 6 before installer container of revision 7 gets to run and completes. In which case resources of revision 7 are replaced by resources of revision 6 even though the installer of revision 7 completes successfuly. The installer controller does not expects this situation to happen and indefinitely waits until the revision 7 static pod appears. However, this condition never becomes true.


ingvagabund:
/hold
before it gets tested

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -428,6 +430,50 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
 
+	// Check if there's a target resource directory with higher revision.
+	// In some cases an installer container with a lower revision can start
+	// to run after an installer with a higher revision finishes.
+	// In which case the higher revision resources gets rewritten by lower
+	// revision resources. See https://bugzilla.redhat.com/show_bug.cgi?id=2028595.
+	higherRevisionFound := false
+	targetRevision, err := strconv.Atoi(o.Revision)
+	if err != nil {
+		return fmt.Errorf("Unable to convert revision into an integer: %v", err)
```
sttts, line 441: lower case

ingvagabund:
/hold cancel
Tests going green in https://github.com/openshift/cluster-kube-scheduler-operator/pull/385 or flaking

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -398,6 +400,56 @@ func (o *InstallOptions) substituteSecret(obj *corev1.Secret) *corev1.Secret {
...
+	// Make sure at most one installer carries this check
+	installerLockFile := path.Join(o.ResourceDir, o.PodConfigMapNamePrefix+".lock")
+	installerLock := flock.New(installerLockFile)
+	klog.Infof("acquiring an exclusive lock on a %s", installerLockFile)
+	if err := installerLock.Lock(ctx); err != nil {
+		return fmt.Errorf("failed to acquire an exclusive lock on %s, due to %v", installerLockFile, err), false
+	}
+	defer installerLock.Unlock()
+
+	if err := filepath.Walk(o.ResourceDir, func(filePath string, info os.FileInfo, err error) error {
```
sttts, line 419: how does a lock around the directory walker protect concurrent directory+file creation? Seems to be racy.

ingvagabund: You need to walk through the directory before you can create it. The lock is release after you know whether you can create the directory or not. So when two installers are running in parallel, only one of them is checking for the revision at the given time.

Unless you get to a situation where installer 6 and 7 are running at the same time. Installer 6 takes the precedence when checking for the revision. At which point it gets green. After which the revision 7 installer checks the revision and gets green as well since no matter when the installer 6 creates the directory, installer 7 will always have higher revision. At which point revision 6 installer might take a longer and actually rewrite the revision 7 resources.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -428,6 +430,64 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+
+		if revision > targetRevision {
+			klog.Infof("Found a %v target resource directory with higher revision than requested (%v). Skipping.", path.Join(o.ResourceDir, info.Name()), o.Revision)
+			higherRevisionFound = true
+		}
+
+		return nil
+	}); err != nil {
+		return fmt.Errorf("unable to process target resource directories: %v", err)
+	}
```
sttts, line 483: wouldn't this make a nice function? Would be much easier to read.

ingvagabund:
/hold
until https://github.com/openshift/cluster-kube-scheduler-operator/pull/385 goes green and gets checked

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -860,23 +860,16 @@ func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSp
...
 		fmt.Sprintf("--namespace=%s", pod.Namespace),
 		fmt.Sprintf("--pod=%s", c.configMaps[0].Name),
 		fmt.Sprintf("--resource-dir=%s", hostResourceDirDir),
 		fmt.Sprintf("--pod-manifest-dir=%s", hostPodManifestDir),
-	}
-	if withStartupMonitorSupport {
-		args = append(args, fmt.Sprintf("--pod-manifests-lock-file=%s", fmt.Sprintf("/var/lock/%s-installer.lock", c.staticPodName)))
+		// The installer has to work exclusively so a check for higher revision and copying
+		// resources to a node are an atomic operation. See https://bugzilla.redhat.com/show_bug.cgi?id=2028595.
+		fmt.Sprintf("--pod-manifests-lock-file=%s", fmt.Sprintf("/var/lock/%s-installer.lock", c.staticPodName)),
```
sttts, line 872: not sure this makes sense. Why disable to the startup monitor logic?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -398,6 +388,47 @@ func (o *InstallOptions) substituteSecret(obj *corev1.Secret) *corev1.Secret {
 	return ret
 }
 
+func (o *InstallOptions) checkForHigherRevision(ctx context.Context) (error, bool) {
```
rphillips, line 391: standard practice is to have the error on the right hand side (bool, error).

ingvagabund:
Closing in favor of https://github.com/openshift/library-go/pull/1302

PR #1264, opened 2021-12-08
Add DialContext configuration override

ardaguclu:
library-go GetClientConfig&GetKubeConfigOrInClusterConfig functions do not
support customizing DialContext settings like Timeout, KeepAlive, etc. Due to that reason
some other repositories use their own GetClientConfig function,
ie. https://github.com/openshift/origin/blob/a9f946043f69428491745ab517d0c23a5b4c2cdd/test/extended/util/client.go#L862

This PR provides a way to customize DialContext settings to unify
GetClientConfig implementation spread across multiple repositories.

### DIFF
--- a/pkg/config/client/client_config.go
+++ b/pkg/config/client/client_config.go
@@ -2,9 +2,10 @@ package client
 
 import (
 	"io/ioutil"
+	"net/http"
+
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"net/http"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/network"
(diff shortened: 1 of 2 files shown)


ardaguclu:
@p0lyn0mial what do you think about this proposal?, am I missing something or is this the right way to go?, thanks.

p0lyn0mial:
I don't feel comfortable with changing `Timeout` and/or `KeepAlive`, there were carefully chosen so that we can recover from a stuck TCP connection quickly. Have a look at https://github.com/openshift/library-go/pull/926

If for some reason you need to change some Transport level setting perhaps you could simply provide your own `Dialer` with the options you want. 

You could simply add it to `config.WrapTransport` function before creating a client. I think that would work and wouldn't require changing the default config. Does it make sense?

ardaguclu:
As @p0lyn0mial stated in comment, there is already simpler method to override DialContext and there is no need to provide a new field. Thus, I'm closing this PR.

/close

PR #1259, opened 2021-12-02
Bump k8s to 1.23.0

damemi:
Part of https://github.com/openshift/kubernetes/pull/1076

PR #1265, opened 2021-12-08
Rebase 1.23

atiratree:
builds on top of https://github.com/openshift/library-go/pull/1259 and fixes logging for NewCommands


pkg/operator/staticpod/certsyncpod/certsync_cmd.go
```diff
@@ -55,6 +66,8 @@ func NewCertSyncControllerCommand(configmaps, secrets []installer.UnrevisionedRe
 	cmd.Flags().StringVar(&o.KubeConfigFile, "kubeconfig", o.KubeConfigFile, "Location of the master configuration file to run from.")
```
atiratree, line 66: Can I encapsulate these in a new options struct?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -83,7 +88,13 @@ func NewInstaller(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "installer",
 		Short: "Install static pod and related resources",
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			// Activate logging as soon as possible, after that
+			// show flags with the final logging configuration.
+			if err := o.logs.ValidateAndApply(); err != nil {
+				return err
+			}
```
atiratree, line 96: I saw that in most of the places we are not calling `defer logs.FlushLogs()`; only in `NewControllerCommandConfig`. Can we expect the consumers will remember to call this? Can I add it in the missing places?

pkg/controller/controllercmd/cmd.go
```diff
@@ -39,6 +39,7 @@ type ControllerCommandConfig struct {
 	version       version.Info
 
 	basicFlags *ControllerFlags
+	logs       *logs.Options
```
p0lyn0mial, line 42: is this a mandatory change? why do we have to change it?

atiratree: yes, it is. klog global flags are removed in 1.23 (https://github.com/kubernetes/enhancements/issues/2845) and need to be registered so we don't get these: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1269/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-operator-disruptive-single-node/1468226993714630656

pkg/operator/staticpod/installerpod/cmd.go
atiratree: although now I see that flushes were removed as well https://github.com/kubernetes/kubernetes/pull/105076/files#diff-186fff86092e78dc06e5f5943eeed276427a565d276e976cdc36939988592c3cL46, but it seems to me from looking at the code this could be skipped in last 5s window : https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/config/v1alpha1/defaults.go#L126

pkg/controller/controllercmd/cmd.go
```diff
@@ -83,10 +85,14 @@ func (c *ControllerCommandConfig) NewCommand() *cobra.Command {
...
 func (c *ControllerCommandConfig) NewCommandWithContext(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			// boiler plate for the "normal" command
 			rand.Seed(time.Now().UTC().UnixNano())
-			logs.InitLogs()
+
+			// Activate logging as soon as possible.
+			if err := c.logs.ValidateAndApply(); err != nil {
```
soltysh, line 93: Why not using the same pattern as below with `klog.Fatal`? The result is identical, we fail.

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
soltysh: What for?

pkg/operator/staticpod/installerpod/cmd.go
soltysh: Yes, I'd add Flush everywhere.

pkg/controller/controllercmd/cmd.go
soltysh: :facepalm: of course we don't have klog setup, so print error and `os.Exit(1)`

pkg/operator/staticpod/installerpod/cmd.go
soltysh: That flush moved to `cli.Run`: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/cli/run.go#L39

pkg/operator/staticpod/certsyncpod/certsync_cmd.go
atiratree: for separation of concern and using similar approach to other places

pkg/controller/controllercmd/cmd.go
```diff
@@ -58,6 +59,7 @@ func NewControllerCommandConfig(componentName string, version version.Info, star
 		version:       version,
 
 		basicFlags: NewControllerFlags(),
```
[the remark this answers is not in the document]
atiratree: ack, let's keep the current approach

pkg/operator/staticpod/installerpod/cmd.go
atiratree: thanks, makes sense now :)

pkg/controller/controllercmd/cmd.go
atiratree: reverted back to `Run`

atiratree:
resolved feedback

atiratree:
rebased on https://github.com/openshift/library-go/pull/1259 - now with v0.23.0

pkg/controller/controllercmd/cmd.go
atiratree: posted a small explanation here: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/581#pullrequestreview-826829435

atiratree:
closing in favor of a simpler default approach (please see https://github.com/openshift/cluster-kube-controller-manager-operator/pull/581). This approach is only useful  for custom solution

pkg/controller/controllercmd/cmd.go
atiratree: fortunately took a bad turn and it is not mandatory change but an optional one :)

PR #1246, opened 2021-11-15
[poc] add conditional resources to the static resource controller

deads2k:
@ravisantoshgudimetla had a PR very similar to this with some other stuff layered on top.
@ingvagabund had a PR that has a specific implementation of this for PDBs

I think this is a basis for PDB creation on some topologies and not others by using conditionals.

/assign @ingvagabund

PR #1266, opened 2021-12-08
APIServerControllerSet: allow to set conditional resources

ingvagabund:
With introduction of conditional resources in https://github.com/openshift/library-go/pull/1246, `APIServerControllerSet.WithStaticResourcesController` needs to be extended to allow to set the conditional resources as well.


ingvagabund:
/hold
until https://github.com/openshift/cluster-authentication-operator/pull/514 tests pass

deads2k:
I'm ok with this.  I'm also ok with having unconditional files and conditional files

/lgtm

PR #1267, opened 2021-12-09
oauth: move RequestingToken to library-go

stlaz:
RequestingTokens is used in multiple places (oc, e2e tests in origin, and, soon, e2e tests in the authentication operator), we should therefore move it to a library that's shared among these sources to prevent the need of copy-pasting this code.

cc @deads2k @s-urbaniak @soltysh


pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+	"os"
+	"regexp"
+	"strings"
+
+	"k8s.io/klog/v2"
+)
+
+// BasicAuthNoUsernameError when basic authentication challenge handling was attempted
+// but the required username was not provided from the command line options
+type BasicAuthNoUsernameError struct{}
```
s-urbaniak, line 17: why does this needs to be fully qualified type?

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+		Host:             host,
+		Reader:           reader,
+		Writer:           writer,
+		passwordPrompter: passwordPrompter,
+		Username:         username,
+		Password:         password,
+	}
+}
+
+type BasicChallengeHandler struct {
```
s-urbaniak, line 46: this is missing godoc

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+	// prompted tracks whether this handler has already prompted for a username and/or password.
+	prompted bool
+}
+
+func (c *BasicChallengeHandler) CanHandle(headers http.Header) bool {
+	isBasic, _ := basicRealm(headers)
+	return isBasic
+}
+func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.Header) (http.Header, bool, error) {
+	if c.prompted {
```
s-urbaniak, line 74: please add a note in godoc that this is not guarded, hence unsafe for concurrent use.

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+		klog.V(2).Info("already prompted for challenge, won't prompt again")
```
s-urbaniak, line 75: isn't V(2) too coarse grained?

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, false, nil
+	}
+	if c.handled {
+		klog.V(2).Info("already handled basic challenge")
```
s-urbaniak, line 79: dito re:granularity

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Reader is used to prompt for username/password. If nil, no prompting is done
+	Reader io.Reader
+	// Writer is used to output prompts. If nil, stdout is used
+	Writer io.Writer
+
+	// passwordPrompter is used to retrieve the password as a string
+	passwordPrompter PasswordPrompter
+
+	// Username is the username to use when challenged. If empty, a prompt is issued to a non-nil Reader
```
s-urbaniak, line 58: the code seems to behave differently. If `len(Username) == 0` we are returning a `NewBasicAuthNoUsernameError()`.

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+	// no realm
+	regexp.MustCompile(`(?i)^\s*basic(?:\s+|$)`),
+}
+
+// basicRealm returns true if a header indicates a basic auth challenge,
+// and the realm if one exists.
+func basicRealm(headers http.Header) (bool, string) {
+	for _, challengeHeader := range headers[http.CanonicalHeaderKey("WWW-Authenticate")] {
+		for _, r := range basicRegexes {
+			if matches := r.FindStringSubmatch(challengeHeader); matches != nil {
```
s-urbaniak, line 146: let's reduce indentation:
```
if matches := r.FindStringSubmatch(challengeHeader); matches == nil {
  continue
}
...
```

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	responseHeaders := http.Header{}
+	responseHeaders.Set("Authorization", getBasicHeader(username, password))
+	// remember so we don't re-handle non-interactively
+	c.handled = true
+	return responseHeaders, true, nil
+}
+func (c *BasicChallengeHandler) CompleteChallenge(requestURL string, headers http.Header) error {
+	return nil
```
s-urbaniak, line 123: is this implementing some interface? If so, which one?

pkg/oauth/tokenrequest/basicauth_test.go
```diff
@@ -0,0 +1,347 @@
...
+			t.Run(fmt.Sprintf("%s[%d]", k, i), func(t *testing.T) {
+				out := &bytes.Buffer{}
+				tc.Handler.Writer = out
+
+				canHandle := tc.Handler.CanHandle(challenge.Headers)
+				if canHandle != challenge.ExpectedCanHandle {
+					t.Errorf("%s: %d: Expected CanHandle=%v, got %v", k, i, challenge.ExpectedCanHandle, canHandle)
+				}
+
+				if canHandle {
```
s-urbaniak, line 189: let's reduce indentation:
```
if !canHandle {
  continue
}
```

pkg/oauth/tokenrequest/basicauth_test.go
```diff
@@ -0,0 +1,347 @@
...
+
+		"realm containing comma": {
+			Headers: http.Header{
+				WWW_AUTHENTICATE: []string{`basic realm="Foo, bar", otherparam="value"`},
+			},
+			ExpectedBasic: true,
+			ExpectedRealm: `Foo, bar`,
+		},
+
+		// TODO: additional forms to support
```
s-urbaniak, line 333: wouldn't this be a good timing to add these?

pkg/oauth/tokenrequest/negotiator_gssapi.go
```diff
@@ -0,0 +1,198 @@
...
+	if err := g.cred.Release(); err != nil {
+		errs = append(errs, convertGSSAPIError(err))
+	}
+	if err := g.lib.Unload(); err != nil {
+		errs = append(errs, convertGSSAPIError(err))
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func (g *gssapiNegotiator) loadLib() (*gssapi.Lib, error) {
```
s-urbaniak, line 158: oh dear :see_no_evil:

pkg/oauth/tokenrequest/basicauth.go
stlaz: so that you are able to distinguish the error outside of the library

pkg/oauth/tokenrequest/basicauth.go
stlaz: this was originally purposed for `oc` but I agree that V(4) will be better for a library function

pkg/oauth/tokenrequest/basicauth.go
stlaz: `ChallengeHandler`, added that to the struct's godoc

pkg/oauth/tokenrequest/basicauth.go
stlaz: That would add two lines and would have to to carry `matches` out of the scope of the condition, I think the current implementation is fine given how simple and short it is.

pkg/oauth/tokenrequest/basicauth_test.go
stlaz: this does not seem to be about the tests but about the general support of these forms, I don't think we should extend the functionality as a part of this PR

pkg/oauth/tokenrequest/basicauth.go
s-urbaniak: ack, just a suggestion to consider other options as well, see https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully. The "Assert errors for behaviour, not type" is a good one.

this seems fine for now though.

pkg/oauth/tokenrequest/basicauth.go
s-urbaniak, line 48: the struct (and constructor) still do not have a summary godoc.

go.mod
```diff
@@ -18,6 +20,7 @@ require (
 	github.com/openshift/api v0.0.0-20211209135129-c58d9f695577
 	github.com/openshift/build-machinery-go v0.0.0-20210806203541-4ea9b6da3a37
 	github.com/openshift/client-go v0.0.0-20211209144617-7385dd6338e3
+	github.com/openshift/gssapi v0.0.0-20161010215902-5fb4217df13b
```
soltysh, line 23: This library requires static bindings to gssapi library which are one of the biggest pains in oc preventing us from upgrading rhel or cross-building binaries for OS-es other than linux. Is there any potential danger for other consumers?

pkg/oauth/tokenrequest/basicauth.go
```diff
@@ -0,0 +1,168 @@
...
+					return true, matches[1]
+				}
+				// No realm, but still basic
+				return true, ""
+			}
+		}
+	}
+	return false, ""
+}
+func getBasicHeader(username, password string) string {
```
soltysh, line 166: Nit: missing empty line before new function

pkg/oauth/tokenrequest/negotiator_gssapi.go
soltysh: yeah, that's the bit I told you we should investigate long-term to replace with a go-based solution

soltysh (commented):
Have you considered splitting the request_token from gssapi and sspi implementations such that library-go has only interfaces defined for handlers, but only oc (I don't know of any consumer of these) have the gssapi and sspi implementations.

soltysh (approved):
/lgtm

mfojtik:
adding label manually... the only change there requiring top-level approver is go.mod modification.

PR #1268, opened 2021-12-14
tweak implementation of resourcecache

deads2k:
builds on https://github.com/openshift/library-go/pull/1235 to improve code-reuse

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -3,479 +3,105 @@ package resourceapply
 import (
 	"bytes"
 	"context"
-	"crypto/md5"
-	"errors"
 	"fmt"
-	"io"
 	"sort"
 	"strings"
 
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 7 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -231,11 +331,29 @@ func ApplyConfigMap(ctx context.Context, client coreclientv1.ConfigMapsGetter, r
...
 	}
 	reportUpdateEvent(recorder, required, err, details)
+	cache.UpdateCachedResourceMetadata(required, actual)
 	return actual, true, err
 }
 
 // ApplySecret merges objectmeta, requires data
-func ApplySecret(ctx context.Context, client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret) (*corev1.Secret, bool, error) {
+func ApplySecretImproved(ctx context.Context, client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret, cache ResourceCache) (*corev1.Secret, bool, error) {
+	existing, err := client.Secrets(requiredInput.Namespace).Get(ctx, requiredInput.Name, metav1.GetOptions{})
```
deads2k, line 340: here it is

deads2k: @jerpeter1 I should have noticed, but the diff was messy before my commit on top.  you cannot create until after you manipulate the stringdata below.

PR #1269, opened 2021-12-15
operator/invalidcertscontroller: initial commit

s-urbaniak:
This introduces "invalid certs controller". This controller uses Prometheus to detect invalid certificates in the cluster.

See https://github.com/openshift/enhancements/pull/980 for more details.

/cc @sttts @stlaz


pkg/operator/invalidcertscontroller/invalid_certs_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
+)
+
+type invalidCertsController struct {
+	operatorClient          v1helpers.OperatorClient
+	prometheusQuery         string
```
s-urbaniak, line 26: not sure if leaving this as an arbitrary string is a great idea. however, the underlying prometheus queries might vary a lot.

pkg/operator/invalidcertscontroller/invalid_certs_controller.go
```diff
@@ -0,0 +1,145 @@
...
+func newInvalidCertsCondition(invalidCertsCount float64) operatorv1.OperatorCondition {
+	condition := operatorv1.OperatorCondition{
+		Type:   "InvalidCertsUpgradeable",
+		Status: operatorv1.ConditionTrue,
+	}
+
+	if invalidCertsCount >= 1.0 {
+		condition.Status = operatorv1.ConditionFalse
+		condition.Reason = "InvalidCertsDetected"
+		condition.Message = fmt.Sprintf("%d invalid certificates were detected", int(invalidCertsCount))
```
stlaz, line 90: this needs to be self-explanatory - what's an invalid cert?

s-urbaniak: the underlying query might have different semantics on what "invalid" means. suggestion: how about adding some context string in the constructor?

s-urbaniak: for our use case, "invalid" means "common name is used vs. SAN fields".

stlaz: If you wanted to have the controller more generic, you'd have to allow for specifying custom checks that each generate their own condition.

The controller as of now is very specific to a single problem so you don't need to worry about adding context strings in the constructor and may just as well hard-code it here.

s-urbaniak: you have a good point, let's make this controller then very specific to the "CN field used vs. SAN field" issue. I wonder if the underlying query should be more opinioated then, @dgrisonnet do you maybe have opinions?

dgrisonnet: I personally really like the idea of keeping it generic because if we want to take into account other errors in the future, we will have to write a new controller which will most likely have a very similar logic. With that in mind, what do you think about:
- Decouple the Prometheus query part and create a PrometheusQuery controller whose sole role would be to execute queries and return values
- Use PrometheusQuery controllers in the invalidCertsController to get specific information about invalid certs and update status depending on the query and the values returned by the PrometheusQuery controllers.

That would allow extending the invalidCertsController with new checks whenever we want.

pkg/operator/invalidcertscontroller/invalid_certs_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	prometheusapi "github.com/prometheus/client_golang/api"
+	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
soltysh, line 21: Nit: if you want to keep imports ordered: builtin, 3-rd party, k8s, openshift :wink:

pkg/operator/invalidcertscontroller/invalid_certs_controller.go
soltysh: I'm hesitant on this approach, what kind of metric consumers are meant to look after? I only know the one from kube-apiserver, shouldn't that controller be defined only there? If you're thinking about other consumers I'm not sure if they have that metric available, I was rather considering this particular controller to look over pre-defined certificates and checking for their correctness which is what that kube-apiserver metric does, no?

s-urbaniak: this is also supposed to be used in cluster-authentication-operator and a yet-to-be-introduced metric in oauth-server, see https://github.com/openshift/enhancements/blob/9225dbdfa07d7479f412e6d2c5f275f484def5b0/enhancements/authentication/invalid-certs.md#oauth-server

s-urbaniak: Additionally, we can replace the metric controller in etcd-operator, see https://github.com/openshift/cluster-etcd-operator/tree/master/pkg/operator/metriccontroller.

soltysh: Gotcha, that makes sens, thx :)

pkg/operator/metricscontroller/metrics_controller.go
```diff
@@ -0,0 +1,109 @@
...
+	"k8s.io/client-go/transport"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type MetricsSyncFunc func(ctx context.Context, controllerContext factory.SyncContext, promClient prometheusv1.API) error
+
+type invalidCertsController struct {
```
soltysh, line 25: Nit: metricsController to match the new approach

soltysh (approved):
/lgtm
modulo one nit which can be fixed before final tag

soltysh:
/hold 
to get the nit fixed

pkg/operator/metricscontroller/legacy_cn_certs_sync.go
```diff
@@ -0,0 +1,58 @@
...
+func newInvalidCertsCondition(invalidCertsCount float64) operatorv1.OperatorCondition {
+	condition := operatorv1.OperatorCondition{
+		Type:   "InvalidCertsUpgradeable",
+		Status: operatorv1.ConditionTrue,
+	}
+
+	if invalidCertsCount >= 1.0 {
+		condition.Status = operatorv1.ConditionFalse
+		condition.Reason = "InvalidCertsDetected"
+		condition.Message = fmt.Sprintf("%d certificates using CN field as host name detected", int(invalidCertsCount))
```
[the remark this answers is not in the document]
s-urbaniak: fixed

pkg/operator/metricscontroller/metrics_controller.go
s-urbaniak: fixed

s-urbaniak:
thanks everybody for the reviews. Setting up a proof PR against kube-apiserver-operator.

pkg/operator/invalidcertscontroller/invalid_certs_controller.go
s-urbaniak: fixed

pkg/operator/invalidcertscontroller/invalid_certs_controller.go
s-urbaniak: it seems we converged against consensus, now reflected in code.

pkg/operator/metricscontroller/metrics_controller.go
```diff
@@ -0,0 +1,109 @@
...
+	return factory.New().
+		WithSync(c.sync).
+		ResyncEvery(time.Minute).
+		ToController("MetricsController", recorder.WithComponentSuffix("metrics-controller"))
+}
+
+func (c *metricsController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	client, closer, err := c.newPrometheusClientFunc()
+	if err != nil {
+		return fmt.Errorf("error instantiating prometheus client: %w", err)
```
mfojtik, line 53: is this event worthy?

s-urbaniak: yes i think this is worth an event. these would catch hiccups in prometheus client instantiation and query execution.

pkg/operator/metricscontroller/legacy_cn_certs_sync.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// NewLegacyCNCertsMetricsSyncFunc creates a metrics sync function that executes the given query
+// and interprets the result value as the count of invalid certificates containing CN fields used as host names.
+// If the value is >=1 the `InvalidCertsDetected` condition state will be set.
+//
+// The supported prometheus query result types are:
+// - vector: only the first sample of the vector will be used for evaluation.
+// - scalar: the returned value will be used.
+func NewLegacyCNCertsMetricsSyncFunc(query string, operatorClient v1helpers.OperatorClient) MetricsSyncFunc {
```
deads2k, line 23: this doesn't look very useful to expose.

I think people wanted a, "pass a cert or a secret along with a servername and determine if it was valid or not"

pkg/operator/metricscontroller/metrics_controller.go
```diff
@@ -0,0 +1,109 @@
...
+		newPrometheusClientFunc: func() (prometheusv1.API, idleConnectionCloser, error) {
+			return newInClusterPrometheusClient(serviceCAPath)
+		},
+		metricsSyncFunc: metricsSyncFunc,
+	}
+
+	return factory.New().
+		WithSync(c.sync).
+		ResyncEvery(time.Minute).
+		ToController("MetricsController", recorder.WithComponentSuffix("metrics-controller"))
```
deads2k, line 47: need a better name that can be flavored or set by a name argument.

deads2k:
lgtm modulo my comments.

pkg/operator/metricscontroller/legacy_cn_certs_sync.go
s-urbaniak: This is supposed to leverage the metric exposed via https://github.com/kubernetes/kubernetes/pull/95396.

s-urbaniak: And a metric we are introducing in oauth-server: https://github.com/openshift/enhancements/blob/9225dbdfa07d7479f412e6d2c5f275f484def5b0/enhancements/authentication/invalid-certs.md#oauth-server

s-urbaniak: Having said that, the idea to fetch secrets directly in the operator seems viable as well. Let me check if we can access all of this in the respective operators.

pkg/operator/metricscontroller/metrics_controller.go
s-urbaniak: Indeed, good suggestion.

pkg/operator/metricscontroller/legacy_cn_certs_sync.go
s-urbaniak: @stlaz as far as i know, the actual target server certificates are not available as secrets and/or configmaps, are they?

s-urbaniak: alternatively we could, redundantly implement the metrics sync function in the corresponding operators as well, as they have intimate knowledge over the underlying query and their return type, hence the implementation would vary slightly as well.

(in case they don't expose the server cert via config/secret)

pkg/operator/metricscontroller/metrics_controller.go
s-urbaniak: fixed

pkg/operator/metricscontroller/metrics_controller.go
s-urbaniak: fixed

s-urbaniak: side note: event generation should be done within the metrics sync func itself. we warn if prometheus client fails to instantiate.

pkg/operator/metricscontroller/legacy_cn_certs_sync.go
stlaz: >  as far as i know, the actual target server certificates are not available as secrets and/or configmaps, are they?

In case of the metrics, they aren't. You would only know which CA bundle to use to connect to the webhook/aggregated API, but that does not help you much because CA bundles are irrelevant to the SAN checks.

What I think David is trying to point out is that it would be useful to have a validation function you can use in places where the server certificate would in fact appear in secrets/configmaps (=> config).

s-urbaniak: > validation function you can use in places where the server certificate would in fact appear in secrets/configmaps (=> config).

agreed, however i want to make this as a separate PR.

deads2k: I'm not against using a metric  in your controller (though I probably would not have chose that).  I'm saying that the request for an exported library function was about inspecting a cert, not making a prom query that assumes other code has inspected.

s-urbaniak: As clarified out of band: this is not supposed to replace the idea of introducing functions or a controller to validate certificates explicitly. This functionality will be provided as a separate PR.

pkg/operator/metricscontroller/metrics_controller.go
```diff
@@ -0,0 +1,114 @@
...
+		WithSync(c.sync).
+		ResyncEvery(time.Minute).
+		ToController(name, recorder)
+}
+
+func (c *metricsController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	client, closer, err := c.newPrometheusClientFunc()
+	if err != nil {
+		err = fmt.Errorf("error instantiating prometheus client: %w", err)
+		c.recorder.Warning("PrometheusClientFailed", err.Error())
```
stlaz, line 58: did you mean to continue with improperly instantiated client?

s-urbaniak: indeed, this is a mistake, fixed.

PR #1270, opened 2021-12-16
use upstream client-go metrics

aojea:
We were using our own copy of the client-go metrics, that diverged from
the upstream ones, actually one of them was deprecated.


aojea:
/assign @sttts @tkashem @d-ramakri 

This will bring the client-go metrics to all the Openshift components that  use library-go and keep us aligned with upstream

pkg/controller/metrics/client_metrics.go
```diff
@@ -1,97 +0,0 @@
...
-
-	"github.com/prometheus/client_golang/prometheus"
-)
-
-var (
-	// requestLatency is a Prometheus Summary metric type partitioned by
-	// "verb" and "url" labels. It is used for the rest client latency metrics.
-	requestLatency = k8smetrics.NewHistogramVec(
-		&k8smetrics.HistogramOpts{
-			Name:    "rest_client_request_latency_seconds",
```
aojea, line 19: this metric was deprecated, now is rest_client_request_duration_seconds

aojea:
/assign @dgrisonnet 
autocomplete things :)

PR #1271, opened 2021-12-16
csi: add hook to inject custom CA bundle

bertinatto:
/assign @deads2k

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	corev1 "k8s.io/client-go/informers/core/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/v1helpers/helpers.go
```diff
@@ -383,3 +383,42 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
 
 	return nil
 }
+
+func InjectTrustedCAIntoContainers(podSpec *corev1.PodSpec, configMapName string, containerNames []string) error {
+	podSpec.Volumes = append(podSpec.Volumes, corev1.Volume{
+		Name: "trusted-ca-bundle",
```
deads2k, line 389: `non-standard-root-system-trust-ca-bundle`

pkg/operator/v1helpers/helpers.go
```diff
@@ -383,3 +383,42 @@ func InjectObservedProxyIntoContainers(podSpec *corev1.PodSpec, containerNames [
...
+		VolumeSource: corev1.VolumeSource{
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: configMapName,
+				},
+				Items: []corev1.KeyToPath{
+					{Key: "ca-bundle.crt", Path: "tls-ca-bundle.pem"},
```
deads2k, line 396: does the file format match?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -48,6 +49,28 @@ func WithObservedProxyDeploymentHook() dc.DeploymentHookFunc {
...
 }
 
+func WithTrustedCADeploymentHook(
+	configMapNamespace string,
+	configMapName string,
+	configMapInformer corev1.ConfigMapInformer,
+) dc.DeploymentHookFunc {
+	return func(_ *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		_, err := configMapInformer.Lister().ConfigMaps(configMapNamespace).Get(configMapName)
+		if apierrors.IsNotFound(err) {
```
deads2k, line 59: also require the "ca-bundle.crt" key?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -48,6 +49,28 @@ func WithObservedProxyDeploymentHook() dc.DeploymentHookFunc {
...
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("failed to get ConfigMap %s/%s: %v", configMapNamespace, configMapName, err)
+		}
+		containerNamesString := deployment.Annotations["config.openshift.io/inject-trusted-ca"]
```
deads2k, line 65: when the content changes, what manages the rolling restart?

bertinatto: This controller resyncs when the CM changes (and this function hook is called again).

pkg/operator/v1helpers/helpers.go
bertinatto: They should match, I got these file names from here: https://docs.openshift.com/container-platform/4.9/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -48,6 +49,32 @@ func WithObservedProxyDeploymentHook() dc.DeploymentHookFunc {
...
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("failed to get ConfigMap %s/%s: %v", configMapNamespace, configMapName, err)
+		}
+		_, ok := cm.Data["ca-bundle.crt"]
+		if !ok {
+			return nil
+		}
+		containerNamesString := deployment.Annotations["config.openshift.io/inject-trusted-ca-bundle"]
```
bertinatto, line 69: The Cluster Network Operator uses `(...)trusted-cabundle` (without `-`) to choose ConfigMaps to inject the custom bundle. Do I need to use the same?

bertinatto:
CC @openshift/storage

deads2k:
> @deads2k deads2k 4 days ago
when the content changes, what manages the rolling restart?

> @bertinatto bertinatto 4 days ago
This controller resyncs when the CM changes (and this function hook is called again).

When the content of the configmap changes, but is still present, the deployment and daemonset do not change, but the process running in each pod likely needs to restart in response.  I don't think this code is doing that.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
deads2k: > Search other repos and be consistent. This would also be a reason to force the name of the config map you inject to be a name I don't prefer. I didn't think about this being the same as the CVO annotation.

Actually, I don't see this there.   I see this for configmaps: https://github.com/openshift/library-go/blob/d48a1fb9b7c2c89f5914cc670a4056bcccdd0a34/pkg/operator/resource/resourceapply/core.go#L277

but nothing injecting into pods. Where does the pod injection come from?  Is it fully novel in this code?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: OK, so I hashed the whole CM and added an annotation to the deployment. When the CM content changes, the deployment will be rolled out again.

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: As per offline discussion, named it `config.openshift.io/inject-proxy-cabundle` (to be consistent with `config.openshift.io/inject-proxy`).

deads2k:
/approve

leaving lgtm to someone else in storage.

PR #1272, opened 2021-12-16
Ensure that `required` is always fixed up by ApplySecret

jerpeter1:
This is to fix TestEnsureNodeKubeconfigs in cluster-kube-apiserver's unit tests (I'm not sure if the test being broken is actually important or not, but we can re-address that later)

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -337,16 +337,10 @@ func ApplyConfigMapImproved(ctx context.Context, client coreclientv1.ConfigMapsG
 
 // ApplySecret merges objectmeta, requires data
 func ApplySecretImproved(ctx context.Context, client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret, cache ResourceCache) (*corev1.Secret, bool, error) {
+	// copy the stringData to data.  Error on a data content conflict inside required.  This is usually a bug.
+
 	existing, err := client.Secrets(requiredInput.Namespace).Get(ctx, requiredInput.Name, metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		requiredCopy := requiredInput.DeepCopy()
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -337,23 +337,6 @@ func ApplyConfigMapImproved(ctx context.Context, client coreclientv1.ConfigMapsG
 
 // ApplySecret merges objectmeta, requires data
 func ApplySecretImproved(ctx context.Context, client coreclientv1.SecretsGetter, recorder events.Recorder, requiredInput *corev1.Secret, cache ResourceCache) (*corev1.Secret, bool, error) {
-	existing, err := client.Secrets(requiredInput.Namespace).Get(ctx, requiredInput.Name, metav1.GetOptions{})
```
deads2k, line 340:
```go
existing, err := 
if err != nil && !apierrors.IsNotFound(err){
    return nil, false, err
}
if cache.SafeToSkipApply(requiredInput, existing) {
	return existing, false, nil
}

// do the required deep copy and string data stuff

		requiredCopy := requiredInput.DeepCopy()
		actual, err := client.Secrets(requiredCopy.Namespace).
			Create(ctx, resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*corev1.Secret), metav1.CreateOptions{})
		reportCreateEvent(recorder, requiredCopy, err)
		cache.UpdateCachedResourceMetadata(requiredInput, actual)
		return actual, true, err
	}

// make sure all UpdateCachedResourceMetadata use requiredInput and not required
```

and you keep your "skip deep copy" behavior, right?

jerpeter1: I just tested your suggestion, but I had to make a small change to get the unit tests to pass (`requiredCopy := required.DeepCopy()` instead of `requiredCopy := requiredInput.DeepCopy()`

PR #1273, opened 2021-12-20
Add nil checks to resource cache

jerpeter1:
ApplSecret is a caching client so sometimes will return a nil from a GET, which causes
the embedded ObjectMeta to be nil.


deads2k:
golang will make an old man out of me.

/lgtm

PR #1274, opened 2021-12-21
Add precondition to the StaticResourceController

slaskawi:
https://issues.redhat.com/browse/AUTH-136

This Pull Request introduces requirements into the StaticResourceController required by https://github.com/openshift/cert-manager-operator/pull/29

### DIFF
--- a/pkg/operator/staticresourcecontroller/static_resource_controller.go
+++ b/pkg/operator/staticresourcecontroller/static_resource_controller.go
@@ -51,10 +51,17 @@ func init() {
 	utilruntime.Must(admissionregistrationv1.AddToScheme(genericScheme))
 }
 
+// StaticResourcesPreconditionsFuncType checks if the precondition is met (is true) and then proceeds with the sync.
+// When the requirement is not met, the controller reports degraded status.
+//
+// In case, the returned ready flag is false, a proper error with a valid description is recommended.
+type StaticResourcesPreconditionsFuncType func(ctx context.Context) (bool, error)
+
 type StaticResourceController struct {
 	name                   string
 	manifests              []conditionalManifests
 	ignoreNotFoundOnCreate bool
+	preconfitions          []StaticResourcesPreconditionsFuncType
 
 	operatorClient v1helpers.OperatorClient
 	clients        *resourceapply.ClientHolder
@@ -99,6 +106,8 @@ func NewStaticResourceController(
 		operatorClient: operatorClient,
 		clients:        clients,
 
+		preconfitions: []StaticResourcesPreconditionsFuncType{defaultStaticResourcesPreconditionsFunc},
+
 		eventRecorder: eventRecorder.WithComponentSuffix(strings.ToLower(name)),
 
 		factory:          factory.New().WithInformers(operatorClient.Informer()).ResyncEvery(1 * time.Minute),
@@ -116,6 +125,18 @@ func NewStaticResourceController(
 // NotFound errors are reported in <name>Degraded condition, but with Degraded=false.
 func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceController {
 	c.ignoreNotFoundOnCreate = true
+	return c
+}
+
+// WithPrecondition adds a precondition, which blocks the sync method from being executed. Preconditions might be chained using:
+//  WithPrecondition(a).WithPrecondition(b).WithPrecondition(c).
+// If any of the preconditions is false, the sync will result in an error.
+//
+// The requirement parameter should follow the convention described in the StaticResourcesPreconditionsFuncType.
+//
+// When the requirement is not met, the controller reports degraded status.
+func (c *StaticResourceController) WithPrecondition(precondition StaticResourcesPreconditionsFuncType) *StaticResourceController {
+	c.preconfitions = append(c.preconfitions, precondition)
 	return c
 }
 
@@ -262,6 +283,28 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
 		return nil
 	}
 
+	for _, precondition := range c.preconfitions {
+		ready, err := precondition(ctx)
+		// We don't care about the other preconditions, we just stop on the first one.
+		if !ready {
+			var message string
+			if err != nil {
+				message = err.Error()
+			} else {
+				message = "the operator didn't specify what preconditions are missing"
+			}
+			if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Type:    fmt.Sprintf("%sDegraded", c.name),
+				Status:  operatorv1.ConditionTrue,
+				Reason:  "PreconditionNotReady",
+				Message: message,
+			})); updateErr != nil {
+				return updateErr
+			}
+			return err
+		}
+	}
+
 	errors := []error{}
 	var notFoundErrorsCount int
 	for _, conditionalManifest := range c.manifests {
@@ -392,4 +435,8 @@ func (c *StaticResourceController) RelatedObjects() ([]configv1.ObjectReference,
 
 func (c *StaticResourceController) Run(ctx context.Context, workers int) {
 	c.factory.WithSync(c.Sync).ToController(c.Name(), c.eventRecorder).Run(ctx, workers)
+}
+
+func defaultStaticResourcesPreconditionsFunc(_ context.Context) (bool, error) {
+	return true, nil
 }


pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -119,6 +126,12 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
 	return c
 }
 
+// WithPreconfitionFunc adds a pre-condtion function, which is executed in the sync function.
+func (c *StaticResourceController) WithPreconfitionFunc(precondition StaticResourcesPreconditionFuncType) *StaticResourceController {
```
sttts, line 130: what if the return value changes, will objects be removed?

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -265,6 +278,28 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+			})); updateErr != nil {
+				return updateErr
+			}
+			return err
+		}
+		if !ready {
+			if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "PreconditionNotReady",
+				Message: "PreconditionNotReady",
```
sttts, line 296: message should be human readable

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -265,6 +278,28 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+				Status:  operatorv1.ConditionFalse,
+				Reason:  "ErrorCheckingPrecondition",
+				Message: err.Error(),
+			})); updateErr != nil {
+				return updateErr
+			}
+			return err
+		}
+		if !ready {
+			if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
```
sttts, line 293: I don't see a place where the condition is set to status=true

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -265,6 +278,28 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
 	errors := []error{}
 	var notFoundErrorsCount int
 	for _, conditionalManifest := range c.manifests {
+		ready, err := c.precondition(ctx, conditionalManifest.files)
+		if err != nil {
+			if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
```
sttts, line 283: does this call have an internal comparison or will it continuously update?

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -265,6 +278,28 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
 	errors := []error{}
 	var notFoundErrorsCount int
 	for _, conditionalManifest := range c.manifests {
+		ready, err := c.precondition(ctx, conditionalManifest.files)
```
stlaz, line 281: I would expect the precondition to be only run once and have the rest of the logic be handled by the conditional manifest methods `shouldCreateFn()` and `shouldDeleteFn()`

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: No. This only blocks the `sync` from running.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: In this case, we can not tell anything more than the precondition is not ready.

This statement is a safety net if someone doesn't follow the convention described in `StaticResourcesPreconditionFuncType` and did not return an error (with proper description):

> // StaticResourcesPreconditionFuncType checks if precondition is true and then proceeds with the reconciliation loop.
> // In case, the returned ready flag is false, a proper error with a valid description is required.

In such a case, you will fallback here and at this point, we only know that precondition was set to false with no error message.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: No, definitely not!  The precondition is expected to change over time. 

The contract for `shouldCreateFn` evaluates if something should be created once and acts upon it. The precondition may change and once it does, you should proceed with the reconciliation loop. Giving you an example - the Cert Manager Operator blocks the reconciliation loop **until** you remove any existing (like IBM's) Operator. Once you do - it will go ahead and create everything for you.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: It does this continuously. I was copy-pasting this part from `apiservice_controller.go` BTW.

pkg/operator/staticresourcecontroller/static_resource_controller.go
stlaz: Ah, I see, I was unclear - by "run once" I meant it should not be run in a for-cycle but prior to it.

slaskawi: aaaah... yeah, I was thinking about this. However, `StaticResourcesPreconditionFuncType` takes `files []string` as an argument and I'd need aggregate a list of files from all `[]conditionalManifests`. So it's bad this or another way, I'd say.

Do you have any advise, which one is better?

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: Great spot! I added: `Type:    fmt.Sprintf("%sDegraded", c.name),`. The update is at the bottom of the `sync`.

pkg/operator/staticresourcecontroller/static_resource_controller.go
stlaz: I wouldn't expect the precondition to be dependent on a list of files. I would actually expect the terms of the precondition to be constant when you're creating the controller (e.g. "there is no other instance of a cert manager operator in the cluster" - no dependency on any files).

I see that in you're PR you're actually doing exactly that - https://github.com/openshift/cert-manager-operator/pull/29/files#diff-e11833beee2c6440e2fdea89051a509fb504d11388c3ea7bbfeb64ea8f93f772R24 - the string slice is ignored.

slaskawi: Ok, agreed.

pkg/operator/staticresourcecontroller/static_resource_controller.go
sttts: the comment still does not specify what it does.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -262,6 +275,30 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+			return updateErr
+		}
+		return err
+	}
+	if !ready {
+		if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    fmt.Sprintf("%sDegraded", c.name),
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "PreconditionNotReady",
+			Message: "PreconditionNotReady",
```
sttts, line 295: this needs a human readable string, not CamlCase

pkg/operator/staticresourcecontroller/static_resource_controller.go
stlaz: LGTM

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -262,6 +275,30 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+		})); updateErr != nil {
+			return updateErr
+		}
+		return err
+	}
```
sttts, line 300: Will this degraded condition go away ever? Don't see a false-case.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -262,6 +275,30 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+			Reason:  "ErrorCheckingPrecondition",
+			Message: err.Error(),
+		})); updateErr != nil {
+			return updateErr
+		}
+		return err
+	}
+	if !ready {
+		if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    fmt.Sprintf("%sDegraded", c.name),
```
sttts, line 292: I would not expect an operator to go degraded because of some precondition. A precondition elsewhere is something that disabled or delays a controllers to run, but has no visibility to outside.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: Updated with `// WithPreconditionFunc adds a pre-condition function, which can block the sync method from being executed.`. Should be better now.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: Fixed

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: This flips back here: https://github.com/openshift/library-go/blob/master/pkg/operator/staticresourcecontroller/static_resource_controller.go#L298

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: I added a bit more details to the status update block. Should be more readable now.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: I checked what other controllers are doing and it seems the `workload.go` reports degraded as well. Here's the relevant part:

```
	if !preconditionsReady {
		var message string
		for _, err := range errs {
			message = message + err.Error() + "\n"
		}
		if len(message) == 0 {
			message = "the operator didn't specify what preconditions are missing"
		}

		// we are degraded, not available and we are not progressing

		deploymentDegradedCondition.Status = operatorv1.ConditionTrue
		deploymentDegradedCondition.Reason = "PreconditionNotFulfilled"
		deploymentDegradedCondition.Message = message

		deploymentAvailableCondition.Status = operatorv1.ConditionFalse
		deploymentAvailableCondition.Reason = "PreconditionNotFulfilled"

		deploymentProgressingCondition.Status = operatorv1.ConditionFalse
		deploymentProgressingCondition.Reason = "PreconditionNotFulfilled"

		return kerrors.NewAggregate(errs)
	}
```

Source: https://github.com/openshift/library-go/blob/a467dafcef6988afbcce43400c339d7ca68a3298/pkg/operator/apiserver/controller/workload/workload.go#L196-L218

So for symmetry reasons, I'd suggest to leave this as is.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -119,6 +128,15 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
 	return c
 }
 
+// WithRequirement adds a requirement function, which blocks the sync method from being executed.
+// The precondition parameter should follow the convention described in the StaticResourcesRequirementFuncType.
+//
+// When the requirement is not met, the controller reports degraded status.
+func (c *StaticResourceController) WithRequirement(requirement StaticResourcesRequirementFuncType) *StaticResourceController {
+	c.requirement = requirement
```
sttts, line 136: would have expected a slice and appending here.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -262,6 +280,36 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+			Message: err.Error(),
+		})); updateErr != nil {
+			return updateErr
+		}
+		return err
+	}
+	if !ready {
+		var message string
+		if err != nil {
+			message = "the operator didn't specify what preconditions are missing"
```
sttts: how can we get here?

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -262,6 +280,36 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
...
+		} else {
+			message = err.Error()
```
sttts, line 300: err is nil

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: I believe this is fine. The method signature is `WithRequirement` (singular). This means we have only one requirement here. If that was plural, than I would agree with you - it should be a slice.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: My mistake - fixed.

pkg/operator/staticresourcecontroller/static_resource_controller.go
slaskawi: Simplified this part and fixed.

pkg/operator/staticresourcecontroller/static_resource_controller.go
sttts: I disagree. `c.WithRequirement(x).WithRequirement(y)`. That's a common pattern.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -119,6 +128,15 @@ func (c *StaticResourceController) WithIgnoreNotFoundOnCreate() *StaticResourceC
 	return c
 }
 
+// WithPrecondition sets a precondition, which blocks the sync method from being executed.
+// The requirement parameter should follow the convention described in the StaticResourcesPreconditionsFuncType.
+//
+// When the requirement is not met, the controller reports degraded status.
+func (c *StaticResourceController) WithPrecondition(precondition StaticResourcesPreconditionsFuncType) *StaticResourceController {
+	c.precondition = precondition
```
sttts, line 136: the pervious comment still applies here: you can call `WithPrecondition` multiple times and it should append.

PR #1275, opened 2021-12-22
Bug 2037274: [crypto,transport]/missing_san: verify invalid certificates

s-urbaniak:
This introduces helper methods to verify invalid certificates
having missing SAN fields.

Additionally, this introduces a roundtripper decorator
which increases a metric if invalid certificates without a SAN field
are detected

See https://github.com/openshift/enhancements/pull/980

Note: this functionality is against 4.9 only as this is the last release having tolerant behavior against missing SAN certs.

/cc @stlaz

### DIFF
--- a/pkg/transport/missing_san_roundtripper.go
+++ b/pkg/transport/missing_san_roundtripper.go
@@ -1,0 +1,43 @@
+package transport
+
+import (
+	"net/http"
+
+	"github.com/openshift/library-go/pkg/crypto"
+)
+
+type roundTripperFunc func(req *http.Request) (*http.Response, error)
+
+func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return rt(r)
+}
+
+type counter interface {
+	Inc()
+}
+
+// NewMissingSANRoundTripper returns a round tripper that increases the given counter
+// if the wrapped roundtripper returns an error indicating usage of legacy CN fields
+// or if the wrapped roundtripper response leaf certificate includes no SAN field.
+//
+// The counter is compatible both with native Prometheus and k8s metrics types.
+func NewMissingSANRoundTripper(rt http.RoundTripper, c counter) http.RoundTripper {
+	return roundTripperFunc(func(req *http.Request) (resp *http.Response, err error) {
+		resp, err = rt.RoundTrip(req)
+		if crypto.IsHostnameError(err) {
+			c.Inc()
+			return
+		}
+
+		if resp == nil || resp.TLS == nil || len(resp.TLS.PeerCertificates) == 0 {
+			return
+		}
+
+		// The first element is the leaf certificate.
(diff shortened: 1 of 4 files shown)


pkg/transport/missing_san_roundtripper.go
```diff
@@ -0,0 +1,44 @@
...
+			return
+		}
+
+		if resp == nil || resp.TLS == nil || len(resp.TLS.PeerCertificates) == 0 {
+			return
+		}
+
+		// The first element is the leaf certificate that the connection is verified against.
+		// Hence, this is the certificate we have to check SANs against.
+		if !crypto.CertHasSAN(resp.TLS.PeerCertificates[0]) {
```
ibihim, line 38: Would it be better to create a `const` and avoid a **magic number**?

```Go
const leafCert = 0
```

Comments age faster than code.

stlaz: I would cut the comment to just `// The first element is the leaf certificate` instead. That's the documented API. The rest seems obvious.

stlaz:
This is attempting to merge to `master` but I don't think we need this functionality there.

s-urbaniak:
@stlaz afaik we can't directly submit a PR against 4.9 without having a PR against master.

pkg/transport/missing_san_roundtripper.go
s-urbaniak: yes, the first element is documented in go's api directly, i am agnostic to the `const` but idiomatically this is not a "magic" number, it signifies the first element.

s-urbaniak: fixed the comment

s-urbaniak:
actually, we can, let's skip `master` and target 4.9 directly.

pkg/transport/missing_san_roundtripper.go
```diff
@@ -0,0 +1,43 @@
...
+
+type roundTripperFunc func(req *http.Request) (*http.Response, error)
+
+func (rt roundTripperFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return rt(r)
+}
+
+type counter interface {
+	Inc()
+}
```
stlaz, line 17: how is the counter going to be used? Shouldn't it be public since `NewMissingSANRoundTripper()` is public and accepts it as its argument?

s-urbaniak: We don't want to expose the interface as public (it's not an API). We want consumers to dependency-inject a counter implementation that is compatible with this signature. This is the case for both native prometheus counters and k8s counters.

Example: https://github.com/openshift/oauth-server/pull/95/commits/12d2d1d34cdd4230c6be1458aa13d2e1f54e29c0#diff-26eb62ac396048ba3dc28ae5764caf68710f82c7956d15af24dfef8542c7fbc8R755

PR #1276, opened 2021-12-22
Bug 2000276: EncryptionStateControllerDegraded: failed to get converged static pod revision

p0lyn0mial:
It turns out that when a node dies or is disconnected from the rest of the cluster, Kubernetes (kubelet) applies a policy for setting the phase of all Pods on the lost node to Failed.
Pods in the Failed state cannot transition back to Running state.
In addition, failed Pods are note GC'ed. The API objects remain in the cluster's API until a human or controller process explicitly removes them.
Thus we need to tolerate pods in Succeeded or Failed phase only if their revision <= the running pods. Otherwise, we won't make any progress.

See the following for more:
 https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
 https://bugzilla.redhat.com/show_bug.cgi?id=2000276

### DIFF
--- a/pkg/operator/encryption/deployer/revisionedpod.go
+++ b/pkg/operator/encryption/deployer/revisionedpod.go
@@ -141,7 +141,13 @@ func (d *RevisionLabelPodDeployer) HasSynced() bool {
 //   1. All running pods are ready and at the same revision
 //   2. All master nodes have a running pod
 //   3. There are no pending or unknown pods
-//   4. All succeeded and failed pods have revisions that are before the running pods
+//   4. We tolerate pods in Succeeded or Failed phase only if their revision <= the running pods
+//      It turns out that when a node dies or is disconnected from the rest of the cluster, Kubernetes (kubelet) applies a policy for setting the phase of all Pods on the lost node to Failed.
+//      Pods in the Failed state cannot transition back to Running state.
+//      In addition, failed Pods are note GC'ed. The API objects remain in the cluster's API until a human or controller process explicitly removes them.
+//      See the following for more:
+//        https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+//        https://bugzilla.redhat.com/show_bug.cgi?id=2000276
 // Once a converged revision has been determined, it can be used to determine
 // what encryption config state has been successfully observed by the API servers.
 // It assumes that podClient is doing live lookups against the cluster state.
@@ -164,10 +170,6 @@ func getAPIServerRevisionOfAllInstances(revisionLabel string, nodes []string, ap
 	revision, _ := goodRevisions.PopAny()
 	if len(revision) == 0 {
 		revision = "0"
-	}
-
-	if failingRevisions.Has(revision) {
-		return "", fmt.Errorf("api server revision %s has both running and failed pods", revision)
 	}
 
 	// make sure all expected nodes are there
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/deployer/revisionedpod_test.go
```diff
@@ -132,7 +160,7 @@ func TestCategorizePods(t *testing.T) {
 				t.Errorf("categorizePods() gotProgressing = %v, want %v", gotProgressing, tt.wantCategorizeProgressing)
 			}
 
-			if err != nil {
+			if err == nil {
```
p0lyn0mial, line 163: I think this was wrong

pkg/operator/encryption/deployer/revisionedpod.go
```diff
@@ -141,7 +141,13 @@ func (d *RevisionLabelPodDeployer) HasSynced() bool {
...
 //   2. All master nodes have a running pod
 //   3. There are no pending or unknown pods
-//   4. All succeeded and failed pods have revisions that are before the running pods
+//   4. We tolerate pods in Succeeded or Failed phase only if their revision <= the running pods
+//      It turns out that when a node dies or is disconnected from the rest of the cluster, Kubernetes (kubelet) applies a policy for setting the phase of all Pods on the lost node to Failed.
+//      Pods in the Failed state cannot transition back to Running state.
+//      In addition, failed Pods are note GC'ed. The API objects remain in the cluster's API until a human or controller process explicitly removes them.
+//      See the following for more:
+//        https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+//        https://bugzilla.redhat.com/show_bug.cgi?id=2000276
```
sttts, line 150: the BZ is about static pods. The pod life-cycle is about non-static pods, isn't it?

p0lyn0mial: the title is wrong then. The description points that there were `pods` in `InvalidNodeInfo` state in the `openshift-apiserver` ns

pkg/operator/encryption/deployer/revisionedpod_test.go
dgrisonnet: I agree

pkg/operator/encryption/deployer/revisionedpod_test.go
```diff
@@ -132,7 +160,7 @@ func TestCategorizePods(t *testing.T) {
...
 				rev, err := getAPIServerRevisionOfAllInstances("revision", tt.nodes, tt.pods)
 				if (err != nil) != tt.wantCategorizeErr {
```
dgrisonnet, line 165:
```suggestion
				if (err != nil) != wantGetAPIServerRevisionOfAllInstancesError {
```

dgrisonnet (commented):
The fix makes sense to me.

As for the unit tests, I am not sure how much effort it would take, but it would be great to split the current `TestCategorizePods` into two different tests:
- TestCategorizePods
- TestGetAPIServerRevisionOfAllInstances

The current approach where a call to `getAPIServerRevisionOfAllInstances` was added alongside some additional checks goes against unit testing principles as we are testing two different things in one test. As we are seeing with this PR, it is confusing and prone to errors. I wouldn't block this PR with this refactor, but it would be great to at least have a follow-up PR improving that.

dgrisonnet:
I will address the concerns I raised in a follow-up PR.

/lgtm

PR #1278, opened 2022-01-06
Bug 2037274: pkg/security/ldapclient: add configuration for connection verification

s-urbaniak:
This adds the possibility to add connection verification in the ldap client configuration.

See https://github.com/openshift/enhancements/pull/980 for context.

### DIFF
--- a/pkg/security/ldapclient/client.go
+++ b/pkg/security/ldapclient/client.go
@@ -13,6 +13,11 @@ import (
 
 // NewLDAPClientConfig returns a new LDAP client config
 func NewLDAPClientConfig(URL, bindDN, bindPassword, CA string, insecure bool) (Config, error) {
+	return NewLDAPClientConfigWithTLSConnectionVerification(URL, bindDN, bindPassword, CA, insecure, nil)
+}
+
+// NewLDAPClientConfig returns a new LDAP client config
+func NewLDAPClientConfigWithTLSConnectionVerification(URL, bindDN, bindPassword, CA string, insecure bool, verifyFn func(tls.ConnectionState) error) (Config, error) {
 	url, err := ldaputil.ParseURL(URL)
 	if err != nil {
 		return nil, fmt.Errorf("Error parsing URL: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/security/ldapclient/client.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 // NewLDAPClientConfig returns a new LDAP client config
-func NewLDAPClientConfig(URL, bindDN, bindPassword, CA string, insecure bool) (Config, error) {
+func NewLDAPClientConfig(URL, bindDN, bindPassword, CA string, insecure bool) (*ldapClientConfig, error) {
```
ibihim, line 15: It is idiomatic (["accept interfaces, return struct"](https://en.m.wikipedia.org/wiki/Robustness_principle)). But I remember a linter message claiming that returning unexposed types sucks.
Does it make a difference if the struct is exported or not? All the properties are private.

s-urbaniak: This is a controversial topic, even @sttts commented way back in https://github.com/golang/lint/issues/210 (see https://github.com/golang/lint/issues/210#issuecomment-339244863). The biggest complaint is that the type is not embeddable however an interface implied from it is.

ibihim: I think it is better than before.

stlaz: Alternatively you can do:
```go
NewLDAPClientConfig(...) (Config, error) {
    return NewLDAPClientConfigWithTLSConnectionVerification(..., nil)
}
```

s-urbaniak: I am fine with whatever proposal, i am good with a separate constructor modulo the bigger diff. but since admittedly returning private types is not something common in the k8s code base i think @stlaz's proposal does not look alien to readers of existing code.

s-urbaniak: fixed with the additional constructor.

sttts:
The change should also be in master.

s-urbaniak:
@sttts as suggested by @stlaz these changes (amongst others) are not supposed to be in master. These helper functions help to identify legacy certificates in 4.9/go1.16. In master these functions have no value.

sttts:
/hold cancel

The OWNER file changes were copied to master.

PR #1281, opened 2022-01-10
bug 2005901: staticpod builder: initialize operand label selector separately

ingvagabund:
Not every controller consumes operand pod label selector when the StartMonitor is enabled.
Also, the operand pod label selector is now required by the guard
controller since not every operator has the static pod name
equal to the app label. E.g.

| operand | label | static pod name |
| ------- | ----- | --------------- |
| kube-scheduler | app=openshift-kube-scheduler | openshift-kube-scheduler |
| kube-controller-manager | app=kube-controller-manager | kube-controller-manager |
| kube-apiserver | app=openshift-kube-apiserver | kube-apiserver |

In case of KS and KCM both app label and the static pod name are equal.
Whereas in the case of KA, the app label differs from the static pod
name. Thus, it is more practical to provide the operand pod label selector rather than rely on a specific label.

Required by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1275.

### DIFF
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -88,11 +88,12 @@ func NewBuilder(
 type Builder interface {
 	WithEvents(eventRecorder events.Recorder) Builder
 	WithVersioning(operandName string, versionRecorder status.VersionGetter) Builder
+	WithOperandPodLabelSelector(labelSelector labels.Selector) Builder
 	WithRevisionedResources(operandNamespace, staticPodName string, revisionConfigMaps, revisionSecrets []revisioncontroller.RevisionResource) Builder
 	WithUnrevisionedCerts(certDir string, certConfigMaps, certSecrets []installer.UnrevisionedResource) Builder
 	WithInstaller(command []string) Builder
 	WithMinReadyDuration(minReadyDuration time.Duration) Builder
-	WithStartupMonitor(enabledStartupMonitor func() (bool, error), operandPodLabelSelector labels.Selector) Builder
+	WithStartupMonitor(enabledStartupMonitor func() (bool, error)) Builder
 
 	// WithCustomInstaller allows mutating the installer pod definition just before
 	// the installer pod is created for a revision.
@@ -110,6 +111,11 @@ func (b *staticPodOperatorControllerBuilder) WithEvents(eventRecorder events.Rec
 func (b *staticPodOperatorControllerBuilder) WithVersioning(operandName string, versionRecorder status.VersionGetter) Builder {
 	b.operandName = operandName
 	b.versionRecorder = versionRecorder
+	return b
+}
+
+func (b *staticPodOperatorControllerBuilder) WithOperandPodLabelSelector(labelSelector labels.Selector) Builder {
+	b.operandPodLabelSelector = labelSelector
 	return b
 }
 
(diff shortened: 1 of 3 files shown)


ingvagabund:
/hold
waiting until tests in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1275 go green

pkg/operator/staticpod/controllers.go
```diff
@@ -88,11 +88,12 @@ func NewBuilder(
 type Builder interface {
 	WithEvents(eventRecorder events.Recorder) Builder
 	WithVersioning(operandName string, versionRecorder status.VersionGetter) Builder
+	WithOperandPodLabelSelector(labelSelector labels.Selector) Builder
 	WithRevisionedResources(operandNamespace, staticPodName string, revisionConfigMaps, revisionSecrets []revisioncontroller.RevisionResource) Builder
 	WithUnrevisionedCerts(certDir string, certConfigMaps, certSecrets []installer.UnrevisionedResource) Builder
 	WithInstaller(command []string) Builder
 	WithMinReadyDuration(minReadyDuration time.Duration) Builder
-	WithStartupMonitor(enabledStartupMonitor func() (bool, error), operandPodLabelSelector labels.Selector) Builder
+	WithStartupMonitor(enabledStartupMonitor func() (bool, error)) Builder
```
p0lyn0mial, line 96: so it will become:

```
staticPodControllers, err := staticpod.NewBuilder(operatorClient, kubeClient, kubeInformersForNamespaces).
...
WithStartupMonitor(startupmonitorreadiness.IsStartupMonitorEnabledFunction(configInformers.Config().V1().Infrastructures().Lister(), operatorClient)).
WithOperandPodLabelSelector(labels.Set{"apiserver": "true"}.AsSelector())
```

pkg/operator/staticpod/controllers.go
```diff
@@ -285,14 +290,18 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
...
-	manager.WithController(staticpodfallback.New(
-		b.operandNamespace,
-		b.operandPodLabelSelector,
-		b.staticPodOperatorClient,
-		b.kubeInformers,
-		b.enableStartMonitor,
-		b.eventRecorder,
-	), 1)
+	if b.operandPodLabelSelector.Empty() {
+		eventRecorder.Warning("OperandPodLabelMissing", "not enough information provided, not all functionality is present")
```
p0lyn0mial, line 294: I think we should return an error since we need a way to find the operand.

p0lyn0mial: actually, we could do without the selector. At the moment it is used only in the `staticPodFallbackConditionController`. In the worst-case scenario, we would loop over all pods in a given namespace.

pkg/operator/staticpod/controllers.go
ingvagabund: Yes.

pkg/operator/staticpod/controllers.go
ingvagabund: As I understand the `staticPodFallbackConditionController` is narrowly focused on the static pods/operands. So it needs to ignore any other pod. It could incorrectly set the degraded condition otherwise.

p0lyn0mial: not sure why do we want to fire an event in that case? It will make troubleshooting harder. We should either return an error or let it fall through (previous behavior).

ingvagabund:
@p0lyn0mial I like the idea of removing the code. Unfortunately, the guard controller needs to target only the operands and ignore everything else. I will not oppose removing the pod label selector from the `staticPodFallbackConditionController` as I am not fully familiar with the expected functionality and whether the restriction can be relaxed or replaced by what is already wired to the controller. Nevertheless, the guard controller is mirroring the operand pods and it needs to know exactly the difference as each operand pod can have any combination of labels.

I might rely on the operand prefix and match the pods based on the prefix. However, using a pod label selector is more intuitive.

pkg/operator/staticpod/controllers.go
ingvagabund: I do not consider the controller to be crucial. I'd rather set no condition than set it incorrectly. Though, I might be wrong here.

PR #1282, opened 2022-01-11
Bug 2037856: use 'configmapsleases' based leader election

tkashem:
See https://github.com/kubernetes/kubernetes/issues/107454 for details on how to migrate to "leases" leader election.


ravisantoshgudimetla (commented):
/lgtm

Any proof PR in a repo with tests passing?

PR #1283, opened 2022-01-11
Bug 2037274: metrics controller

s-urbaniak:
This introduces "metrics controller".
It uses Prometheus to execute queries in regular intervals.

This is a cherry-pick of https://github.com/openshift/library-go/pull/1269.


s-urbaniak:
Reassigning to the 4.9-only bugzilla. This should speed up the process and we want to gather all fixes in the 4.9-only bugzilla anyways.

PR #1284, opened 2022-01-12
Bug 2031839: metricscontroller: handle empty vector query result

s-urbaniak:
This fixes an empty vector result which can happy during early bootstrap phases, recognized during proof PR https://github.com/openshift/cluster-kube-apiserver-operator/pull/1274.

/cc @stlaz

### DIFF
--- a/pkg/operator/metricscontroller/legacy_cn_certs_sync.go
+++ b/pkg/operator/metricscontroller/legacy_cn_certs_sync.go
@@ -30,7 +30,11 @@ func NewLegacyCNCertsMetricsSyncFunc(query string, operatorClient v1helpers.Oper
 		var value model.SampleValue
 		switch result.Type() {
 		case model.ValVector:
-			value = result.(model.Vector)[0].Value
+			vec := result.(model.Vector)
+			if len(vec) == 0 {
+				return fmt.Errorf("empty vector result from query: %q", query)
+			}
+			value = vec[0].Value
 		case model.ValScalar:
 			value = result.(*model.Scalar).Value
 		default:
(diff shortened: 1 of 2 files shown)


pkg/operator/metricscontroller/legacy_cn_certs_sync.go
```diff
@@ -30,7 +31,11 @@ func NewLegacyCNCertsMetricsSyncFunc(query string, operatorClient v1helpers.Oper
 		var value model.SampleValue
 		switch result.Type() {
 		case model.ValVector:
-			value = result.(model.Vector)[0].Value
+			vec := result.(model.Vector)
+			if len(vec) == 0 {
+				return errors.New("query returned empty vector")
```
stlaz, line 36: Why is this an error? Why can't the vector just be empty?

You should use `fmt.Errorf` instead of `errors.New` as in the rest of the code

s-urbaniak: It is an error as we have no data available and cannot interpret it correctly. Similar to here https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/metriccontroller/fsync_controller.go#L64 there is little we can do.

Here, concretely, it is a transient error. This controller doesn't set a degraded status but rather retries during the next sync.

s-urbaniak: why should we use `fmt.Errorf` here? there is no parameter to be set.

dgrisonnet: I think returning an error is the correct behavior here since an empty result is unexpected and most likely means that no timeserie like the following one exist in Prometheus yet:
```
apiserver_webhooks_x509_missing_san_total{job="apiserver",namespace="default"}
```
That can happen for various reasons unrelated to this particular controller, so in the case where it isn't able to execute the query or get a proper result, I would consider that an error since the controller can't proceed with providing a correct status.

stlaz: I see, I did not have this kind of understanding, in my mind this was just that the metrics was uninitialized which would mean that having the `value` be 0 would be correct in my mind.

PR #1285, opened 2022-01-12
rely on client interfaces for prometheus client

deads2k:
This is required to clean up origin test prometheus usage.


tkashem:
/lgtm

cc @dgrisonnet

PR #1287, opened 2022-01-13
bug 2005901: guard controller: create the pdb if it does not exist

ingvagabund:
The guard controller is missing the code that creates a PDB. Also, do not render the guard pod if its underlying node is not schedulable (tainted with `NoSchedule:node.kubernetes.io/unschedulable`).

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -181,25 +182,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			pdb.Spec.MinAvailable = &minAvailable
 		}
 
-		// List the pdb from the cache in case it exists and there's nothing to update
-		// so no Get request is executed.
-		pdbs, err := c.pdbLister.PodDisruptionBudgets(c.targetNamespace).List(labels.Everything())
-		if err != nil {
-			klog.Errorf("Unable to list PodDisruptionBudgets: %v", err)
-			return err
-		}
-
-		for _, pdbItem := range pdbs {
-			if pdbItem.Name == pdb.Name {
-				if pdbItem.Spec.MinAvailable != pdb.Spec.MinAvailable {
-					_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
-					if err != nil {
-						klog.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
-						return err
-					}
+		pdbObj, err := c.pdbLister.PodDisruptionBudgets(pdb.Namespace).Get(pdb.Name)
+		if err == nil {
+			if pdbObj.Spec.MinAvailable != pdb.Spec.MinAvailable {
+				_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
+				if err != nil {
+					klog.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
+					return fmt.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
 				}
-				break
 			}
+		} else if errors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
+			if err != nil {
+				klog.Errorf("Unable to create PodDisruptionBudget: %v", err)
+				return fmt.Errorf("Unable to create PodDisruptionBudget: %v", err)
+			}
+		} else {
+			klog.Errorf("Unable to get PodDisruptionBudget: %v", err)
+			return err
 		}
 
 		operands := map[string]*corev1.Pod{}
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -23,7 +23,7 @@ import (
 	"k8s.io/klog/v2"
 
 	configv1 "github.com/openshift/api/config/v1"
-	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	configv1informers "github.com/openshift/client-go/config/informers/externalversions/config/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
@@ -48,7 +48,7 @@ type GuardController struct {
 
 	// installerPodImageFn returns the image name for the installer pod
 	installerPodImageFn   func() string
-	createConditionalFunc func() (bool, error)
+	createConditionalFunc func() (bool, bool, error)
 }
 
 func NewGuardController(
@@ -63,7 +63,7 @@ func NewGuardController(
 	podGetter corev1client.PodsGetter,
 	pdbGetter policyclientv1.PodDisruptionBudgetsGetter,
 	eventRecorder events.Recorder,
-	createConditionalFunc func() (bool, error),
+	createConditionalFunc func() (bool, bool, error),
 ) factory.Controller {
 	c := &GuardController{
 		targetNamespace:         targetNamespace,
@@ -112,12 +112,17 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 	klog.V(5).Info("Syncing guards")
 
 	if c.createConditionalFunc == nil {
-		return fmt.Errorf("create conditional not set")
+		return fmt.Errorf("createConditionalFunc not set")
 	}
 
-	shouldCreate, err := c.createConditionalFunc()
+	shouldCreate, precheckSucceeded, err := c.createConditionalFunc()
 	if err != nil {
-		return fmt.Errorf("create conditional returns an error: %v", err)
+		return fmt.Errorf("createConditionalFunc returns an error: %v", err)
+	}
+
+	if !precheckSucceeded {
+		klog.V(4).Infof("create conditional precheck did not succeed, skipping")
+		return nil
 	}
 
 	errs := []error{}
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -189,8 +189,10 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			return err
 		}
 
+		found := false
 		for _, pdbItem := range pdbs {
```
p0lyn0mial, line 193: how many `pdb` do we expect to get?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -201,6 +203,13 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 				break
 			}
 		}
+		if !found {
+			_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
+			if err != nil {
+				klog.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
```
p0lyn0mial, line 209: I think that the errors returned from the sync method are reported by the base controller.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -280,9 +289,13 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 	return utilerrors.NewAggregate(errs)
 }
 
-func IsSNOCheckFnc(infraLister configv1listers.InfrastructureLister) func() (bool, error) {
+// IsSNOCheckFnc creates a function that checks if the topology is SNO
+func IsSNOCheckFnc(infraInformer configv1informers.InfrastructureInformer) func() (bool, error) {
 	return func() (bool, error) {
-		infraData, err := infraLister.Get("cluster")
+		if !infraInformer.Informer().HasSynced() {
+			return false, fmt.Errorf("infrastructures.config.openshift.io informer has not synced yet")
```
p0lyn0mial, line 296: not sure if we want to return an error here, especially it will be logged by the higher-order functions/component. After all an informer is not synced immediately.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: The base controller does not log the line number from the guard_controller.go which is useful for quick location of origin of the error message.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Would you prefer to klog.Infof and return `false, nil` instead?

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: mhm, but the name must be unique therefore we should only get/expect a single PDB, am I right?

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: how about `return fmt.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)` ?

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: I would `return false, nil`. I would expect an informer to log failed synchronization attempts to the API server, wdyt?

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: The reason to call `List()` and not `Get` is to avoid unnecessary communication with the apiserver. To my understand `List` only reads the cache and `Get` makes a direct call to the apiserver.

p0lyn0mial:
```
// PodDisruptionBudgetNamespaceLister helps list and get PodDisruptionBudgets.
// All objects returned here must be treated as read-only.
type PodDisruptionBudgetNamespaceLister interface {
	// List lists all PodDisruptionBudgets in the indexer for a given namespace.
	// Objects returned here must be treated as read-only.
	List(selector labels.Selector) (ret []*v1.PodDisruptionBudget, err error)
	// Get retrieves the PodDisruptionBudget from the indexer for a given namespace and name.
	// Objects returned here must be treated as read-only.
	Get(name string) (*v1.PodDisruptionBudget, error)

}
```

:)

p0lyn0mial: willing to refactor?

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Improving the error message sounds good. Though, I would like to keep the `klog.Errof` there so it's easier to locate the line number in the file.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: `nil` error means the function successfully retrieved the `cluster` object and determined if the topology is SNO or not. Returning `nil` when the informer cache is not synced changes the semantics.

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: fine by me.

ingvagabund: Done

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Done

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -107,16 +108,30 @@ func nodeConditionFinder(status *corev1.NodeStatus, condType corev1.NodeConditio
...
 func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	klog.V(5).Info("Syncing guards")
 
 	if c.createConditionalFunc == nil {
-		return fmt.Errorf("create conditional not set")
+		return fmt.Errorf("createConditionalFunc not set")
 	}
 
-	shouldCreate, err := c.createConditionalFunc()
+	shouldCreate, unknown, err := c.createConditionalFunc()
```
p0lyn0mial, line 127: `notReady`? `notSynced` ?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -107,16 +108,30 @@ func nodeConditionFinder(status *corev1.NodeStatus, condType corev1.NodeConditio
...
 	if err != nil {
-		return fmt.Errorf("create conditional returns an error: %v", err)
+		return fmt.Errorf("createConditionalFunc returns an error: %v", err)
+	}
+
+	if unknown {
+		klog.Infof("create conditional result unknown, skipping")
```
p0lyn0mial, line 133: I would increase the log level here

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -181,25 +196,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
-					}
+		pdbObj, err := c.pdbLister.PodDisruptionBudgets(pdb.Namespace).Get(pdb.Name)
+		if err == nil {
+			if pdbObj.Spec.MinAvailable != pdb.Spec.MinAvailable {
+				_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
+				if err != nil {
+					klog.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
+					return fmt.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
 				}
-				break
```
p0lyn0mial, line 201: okay, so previously we might have created the pdb but not the pods, right?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -208,6 +222,12 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		}
 
 		for _, node := range nodes {
+			// Check whether the node is schedulable
+			if !isNodeSchedulable(node) {
```
p0lyn0mial, line 226: so, this is the fix for https://bugzilla.redhat.com/show_bug.cgi?id=2038481, right?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -181,25 +196,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+			if pdbObj.Spec.MinAvailable != pdb.Spec.MinAvailable {
+				_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
+				if err != nil {
+					klog.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
+					return fmt.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
 				}
-				break
 			}
+		} else if errors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
```
p0lyn0mial, line 209: is this the first time we actually create the a pdb? does it mean we previously didn't create it ?  (we didn't handle `IsNotFound`)

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -107,16 +108,30 @@ func nodeConditionFinder(status *corev1.NodeStatus, condType corev1.NodeConditio
 	return nil
 }
 
+func isNodeSchedulable(node *corev1.Node) bool {
```
p0lyn0mial, line 111: I hope this is the only and the best way to validate a node is schedulable :)

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -280,16 +300,22 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+	// is, unknown, error
+	return func() (bool, bool, error) {
+		if !infraInformer.Informer().HasSynced() {
+			// Do not return transient error
+			return false, true, nil
+		}
+		infraData, err := infraInformer.Lister().Get("cluster")
 		if err != nil {
-			return false, fmt.Errorf("Unable to list infrastructures.config.openshift.io/cluster object, unable to determine topology mode")
+			return false, true, fmt.Errorf("Unable to list infrastructures.config.openshift.io/cluster object, unable to determine topology mode")
```
p0lyn0mial, line 313: shouldn't we return `false` here ?

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: I can not guarantee this is the best way. `oc adm cordon NODE` taints a node with `NoSchedule:node.kubernetes.io/unschedulable` taint so checking for the taint is a minimal condition. `isNodeSchedulable` name might not be the best name in this context. What about renaming the function to `nodeHasUnschedulableTaint`?

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: This is the first time the pdb is created. Previously, there was a loop which assumed the pdb already exists. The case of creating the actual pdb was omitted.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Yes, it is. A node is cordoned before it is drained. Without this check the guard pod gets recreated right after it gets evicted from a node.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: When the `err` is not nil, the second returned value has no use so it can have arbitrary value. Given the isSNO condition is known only when the infrastructure object has its status and the topology properly set, returning `true` is more suitable.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: The `createConditionalFunc` method is a black box for the guard controller. Being ready or synced is too specific without a context.  What about `conditionIncomplete`?

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: +1 for renaming `nodeHasUnschedulableTaint`

PR #1289, opened 2022-01-18
Bug 2037274: pkg/operator/metricscontroller: add prefix, support vector results

s-urbaniak:
Currently, NewLegacyCNCertsMetricsSyncFunc cares about a single result from vectors.
By adding support for handling all results from vectors we can enhance status messaging
for better debugability of faulty certificates.

Additionally, this adds a conditionPrefix to support multiple metrics controllers.

### DIFF
--- a/pkg/operator/metricscontroller/legacy_cn_certs_sync.go
+++ b/pkg/operator/metricscontroller/legacy_cn_certs_sync.go
@@ -3,6 +3,7 @@ package metricscontroller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	prometheusv1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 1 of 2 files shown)


pkg/operator/metricscontroller/legacy_cn_certs_sync.go
```diff
@@ -20,42 +21,54 @@ import (
...
-		Type:   "InvalidCertsUpgradeable",
+		Type:   conditionPrefix + "InvalidCertsUpgradeable",
 		Status: operatorv1.ConditionTrue,
 	}
 
-	if invalidCertsCount >= 1.0 {
+	var invalidSANs []string
+	for _, sample := range samples {
+		if sample.Value > 0 {
+			invalidSANs = append(invalidSANs, fmt.Sprintf("%s => %s", sample.Metric, sample.Value))
```
stlaz, line 64: `sample.Metric` is `nil` if the orig result was of `ValScalar`. Is there a better way to init that struct based on our knowledge? Also, isn't `sample.Metric` a map? What will such output look like? Should it be `"%v => %s"` maybe?

s-urbaniak: `%s` will gracefully print `{}` in this case, see the unit test at: https://github.com/openshift/library-go/blob/7127a1be248594d05807f87be265309c7d6b0315/pkg/operator/metricscontroller/legacy_cn_certs_sync_test.go#L136-L147

Here concretely it will print `server certificates without SAN detected: {} => 10` which is how prometheus outputs metrics without labels which applies to scalars.

stlaz: when will that happen and how am I as an administrator supposed to know what it means?

stlaz: wouldn't it make more sense for the message to read something like "server certificates without SAN were detected in webhooks and aggregated APIs"?

s-urbaniak: This is a synthetic test case for scalars. In your concrete case of webhooks and/or aggregated APIs the message will be:
```
server certificates without SAN detected: {type="webhooks"} => 123, {type="aggregated"} => 456
```

s-urbaniak: in the case of oauth it will be something like:
```
server certificates without SAN detected: {provider="OpenID"} => 123, {provider="LDAP"} => 456
```

stlaz: ok, my proposal is therefore:
1. do not support scalars
2. don't print the `%s => %s` in case of scalars

Also, shouldn't the message be: `times server certificates without SAN were detected: ` instead of what it currently is? IIRC the metrics increases when a server certificate was spotted not for each unique server certificate.

s-urbaniak: > how am I as an administrator supposed to know what it means?

This question rather refers to the semantics of a concrete playbook. We can indicate the resolution in the text message here of course. How about something like:

```
Server certificates without SAN detected: {provider="OpenID"} => 123, {provider="LDAP"} => 456. These have to be replaced to include SAN fields and not rely on the CN field to be valid.
```

s-urbaniak: plus, instead of the arrow we can make it an equals so it becomes more obvious:
```
Server certificates without SAN detected: {provider="OpenID"} = 123, {provider="LDAP"} = 456. These have to be replaced to include SAN fields and not rely on the CN field to be valid.
```

s-urbaniak: Agreed, let's remove the support for scalars as well.

PR #1291, opened 2022-01-19
Apply ConfigMaps with Injected Keys

adambkaplan:
Add a general purpose function to apply ConfigMaps with data that can be
injected into a specified key. This allows operators to reconcile
ConfigMaps that can have data injected by other controllers. Injected
keys are used by components that require the cluster's global trust
bundle or require the Service CA certificate.

This change preserves the existing ability for the global trust bundle
to be injected into a ConfigMap via the
"config.openshift.io/inject-trusted-cabundle": "true" label, and then be
reconciled using ApplyConfigMap.


adambkaplan:
The new `ApplyInjectableConfigMap` method can be used by operators that need the Service CA injected into a ConfigMap. There are several CVO operators which do this - see https://docs.openshift.com/container-platform/4.9/security/certificate_types_descriptions/service-ca-certificates.html

PR #1299, opened 2022-01-28
staticpodcontroller: refactor ToControllers() method to be less error prone

p0lyn0mial:
in particular:
 run the monitor controllers only when explicitly requested
 allow the monitor and the guard controllers to report required parameters

### DIFF
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -80,7 +80,6 @@ func NewBuilder(
 		staticPodOperatorClient: staticPodOperatorClient,
 		kubeClient:              kubeClient,
 		kubeInformers:           kubeInformers,
-		enableStartMonitor:      func() (bool, error) { return false, nil },
 	}
 }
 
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
in response to address the issue encountered in https://github.com/openshift/cluster-etcd-operator/pull/729

pkg/operator/staticpod/controllers.go
```diff
@@ -281,26 +278,28 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
...
+			b.kubeInformers,
+			b.enableStartMonitor,
+			eventRecorder,
+		), 1)
 
-	if b.operandPodLabelSelector.Empty() {
-		errs = append(errs, fmt.Errorf("missing OperandPodLabelSelector when running StaticPodFallbackConditionController; cannot proceed"))
-	} else {
-		manager.WithController(staticpodfallback.New(
+		if staticPodFallbackController, err := staticpodfallback.New(
```
dgrisonnet, line 291: Having this controller under the `enableStartMonitor`gate feels a bit odd since all the other controllers have their own parameters in the builder that they can be gated on. I can see why we could want it under `enableStartMonitor` since it goes along with the monitoring, but for consistency purposes, it might be better to have it separated from the startup monitor controller.

dgrisonnet: Had an offline discussion with @p0lyn0mial who told me that these controllers were under the same gate because they were part of the same story.

PR #1300, opened 2022-01-31
wait for other installers to finish

deads2k:
this resolves the problem of rapid revisions and it resolves the problem about rapid revisions causing overwrite by old revisions..

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -6,8 +6,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/blang/semver"
 	"github.com/davecgh/go-spew/spew"
@@ -413,6 +417,10 @@ func (o *InstallOptions) Run(ctx context.Context) error {
 		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
 	}
 
+	if err := o.waitForOtherInstallerRevisionsToSettle(ctx); err != nil {
+		return err
+	}
+
 	err = retry.RetryOnConnectionErrors(ctx, func(context.Context) (bool, error) {
 		version, err := o.kubeletVersion(ctx)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,121 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+		}
+		if len(installerPods) == 0 {
+			return false, fmt.Errorf("no installer pods found")
+		}
+		latestRevision, err := getRevisionOfPod(installerPods[len(installerPods)-1])
+		if err != nil {
+			return false, err
+		}
+		if latestRevision > int(currRevision) {
+			return false, fmt.Errorf("more recent revision present on node: thisRevision=%v, moreRecentRevision=%v", currRevision, latestRevision)
```
deads2k, line 471: if a newer revision is placed on the node while an older revision is in this state, then the older revision quits here.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,121 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+		}
+
+		// if any container is not terminated, we need to wait.  There should only ever be a single installer pod
+		// that could be run
+		for _, pod := range installerPods {
+			for _, container := range pod.Status.ContainerStatuses {
+				if container.State.Terminated == nil {
+					return false, nil
```
deads2k, line 479: if a newer revision is placed on the node after an older revision gets past the `waitForOtherInstallerRevisionsToSettle` method, then the newer revision waits here until the kubelet confirms the older revision is finished.

if the newer revision takes too long and the controller deletes it and retries, the same check is run again.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,130 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+			// skip this revision
+			podRevision, err := getRevisionOfPod(pod)
+			if err != nil {
+				return false, err
+			}
+			if podRevision == currRevision {
+				continue
+			}
+			for _, container := range pod.Status.ContainerStatuses {
+				if container.State.Terminated == nil {
```
ingvagabund, line 487: To related the code changes to https://bugzilla.redhat.com/show_bug.cgi?id=2028595#c1:
```
08:50:14.597891 - installer-6-ip-10-0-129-204.us-east-2.compute.internal created
08:50:14.915568 - installer-6-ip-10-0-129-204.us-east-2.compute.internal waiting for its sandbox
08:50:20.195360 - installer-6-ip-10-0-129-204.us-east-2.compute.internal requested to be deleted
08:50:24.001152 - installer-7-ip-10-0-129-204.us-east-2.compute.internal created
08:50:28.014171 - installer-7-ip-10-0-129-204.us-east-2.compute.internal waiting for its sandbox
08:50:25.997697 - installer-7-ip-10-0-129-204.us-east-2.compute.internal installer container started to run
08:50:27.013251 - installer-7-ip-10-0-129-204.us-east-2.compute.internal installer container stopped running/got deleted
08:50:47.250319 - installer-6-ip-10-0-129-204.us-east-2.compute.internal installer container started to run
08:50:48.265838 - installer-6-ip-10-0-129-204.us-east-2.compute.internal installer container stopped running
```

it may take at least 34s from creating an installer pod until it stops running. In this case the installer rev-7 will have to wait for installer rev-6 for 24s.

deads2k:
proof in https://github.com/openshift/cluster-etcd-operator/pull/736

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,130 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+			return false, nil
+		}
+		if len(installerPods) == 0 {
+			return false, fmt.Errorf("no installer pods found")
+		}
+		latestRevision, err := getRevisionOfPod(installerPods[len(installerPods)-1])
+		if err != nil {
+			return false, err
+		}
+		if latestRevision > currRevision {
```
ingvagabund, line 471: Assuming an installer pod with lower revision does not get created after an installer pod with higher revision, lower revision installer pod will never get through this point.

This condition might cause the installation for a given master node to get stuck. Assuming there are installer rev-6 and rev-7 created, rev-6 gets to run before rev-7 (the usual case) but the rev-7 for some reason never starts (e.g. kubelet failing to create a sandbox or other dependent components like secrets). Now the installer rev-6 never gets through this, the installer rev-7 never starts and the installed static pod never gets updated. The installer controller will hang and the operator goes into Degraded=true state.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,130 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+		for _, pod := range installerPods {
+			// skip this revision
+			podRevision, err := getRevisionOfPod(pod)
+			if err != nil {
+				return false, err
+			}
+			if podRevision == currRevision {
+				continue
+			}
+			for _, container := range pod.Status.ContainerStatuses {
```
ingvagabund, line 486: Could `pod.Status.ContainerStatuses` be an empty list before the kubelet updates the field? Worth checking:
```
if len(pod.Status.ContainerStatuses) == 0 {
	return false, nil
}
```

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -438,6 +446,130 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+		if err != nil {
+			return false, err
+		}
+		if latestRevision > currRevision {
+			return false, fmt.Errorf("more recent revision present on node: thisRevision=%v, moreRecentRevision=%v", currRevision, latestRevision)
+		}
+
+		// if any container is not terminated, we need to wait.  There should only ever be a single installer pod
+		// that could be run
+		for _, pod := range installerPods {
```
tkashem, line 477: `getInstallerPodsOnThisNode` returns the installer pod that's running this loop, shouldn't it cause an infinite wait here?

pkg/operator/staticpod/installerpod/cmd.go
deads2k: > it may take at least 34s from creating an installer pod until it stops running. In this case the installer rev-7 will have to wait for installer rev-6 for 24s.

for each installer pod that is created, that is true. If this fails, we will create another install pod.  Waiting for 24s looks correct, I made the wait 60s as I recall.

pkg/operator/staticpod/installerpod/cmd.go
deads2k: rephrasing the timeline to be sure I understand it.
1. installer-6 created
2. installer-7 created
3. installer-6 running
4. installer-6 exits because it's not the latest (this is good)
5. kubelet never starts installer-7 for some reason
6. operator goes degraded.

I think this situation is expected and ok.

pkg/operator/staticpod/installerpod/cmd.go
deads2k: > Could pod.Status.ContainerStatuses be an empty list before the kubelet updates the field? Worth checking:

Sure, will update.

PR #1306, opened 2022-02-08
Bug 2027976: move prune controller and prune pod responsibilities to the installer

atiratree:
to have all logic for creating/deletion at the same place

- installer pod now prunes old resources after succesful installation.
- installer controller now prunes revision-status ConfigMaps when syncing
- remove WithPruning from the static pod builder as it is now done by installer by default

This will prevent race conditions in pruner x installer. and also simplify the use case of a pruner (no additional management of a pod). Basically the installer will cleanup after itself each time it runs. 

@soltysh @sttts


atiratree:
rebased

atiratree:
@soltysh @ravisantoshgudimetla can you please take a look?

PR #1307, opened 2022-02-09
Introduce pre-process function for resources

slaskawi:
This Pull Request introduces a pre-process function to the resources. This gives a caller an opportunity to alter them after decoding. 

Proof PR: https://github.com/openshift/cert-manager-operator/pull/44

### DIFF
--- a/pkg/operator/resource/resourceapply/generic.go
+++ b/pkg/operator/resource/resourceapply/generic.go
@@ -45,6 +45,10 @@ func init() {
 	utilruntime.Must(policyv1.AddToScheme(genericScheme))
 }
 
+// PreprocessResourcesFuncType gives the caller an opportunity to preprocess decoded resources
+// before applying them.
+type PreprocessResourcesFuncType func(ctx context.Context, object runtime.Object) (runtime.Object, error)
+
 type AssetFunc func(name string) ([]byte, error)
 
 type ApplyResult struct {
@@ -102,6 +106,12 @@ func (c *ClientHolder) WithMigrationClient(client migrationclient.Interface) *Cl
 
 // ApplyDirectly applies the given manifest files to API server.
 func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, files ...string) []ApplyResult {
+	return ApplyDirectlyWithPreprocess(ctx, clients, recorder, cache, manifests, defaultPreprocessFunc(), files...)
+}
+
+// ApplyDirectlyWithPreprocess applies the given manifest files to API server (similarly to ApplyDirectly)
+// but also respects the pre-process function.
+func ApplyDirectlyWithPreprocess(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, preprocessFunction func(ctx context.Context, object runtime.Object) (runtime.Object, error), files ...string) []ApplyResult {
 	ret := []ApplyResult{}
 
 	for _, file := range files {
@@ -115,6 +125,12 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
 		requiredObj, err := decode(objBytes)
 		if err != nil {
 			result.Error = fmt.Errorf("cannot decode %q: %v", file, err)
+			ret = append(ret, result)
+			continue
+		}
+		requiredObj, err = preprocessFunction(ctx, requiredObj)
+		if err != nil {
+			result.Error = fmt.Errorf("cannot preprocess %q: %v", file, err)
 			ret = append(ret, result)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


slaskawi:
@sttts @s-urbaniak @stlaz Can I ask for a review?

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -371,6 +392,16 @@ func DeleteAll(ctx context.Context, clients *ClientHolder, recorder events.Recor
 	return ret
 }
 
+func DeleteAll(ctx context.Context, clients *ClientHolder, recorder events.Recorder, manifests AssetFunc, files ...string) []ApplyResult {
+	return DeleteAllWithPreprocess(ctx, clients, recorder, manifests, defaultPreprocessFunc(), files...)
+}
+
+func defaultPreprocessFunc() func(ctx context.Context, object runtime.Object) (runtime.Object, error) {
```
s-urbaniak, line 399: naming suggestion: `nopPreprocess`

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -45,6 +45,10 @@ func init() {
 	utilruntime.Must(policyv1.AddToScheme(genericScheme))
 }
 
+// PreprocessResourcesFuncType gives the caller an opportunity to preprocess decoded resources
+// before applying them.
+type PreprocessResourcesFuncType func(ctx context.Context, object runtime.Object) (runtime.Object, error)
```
s-urbaniak, line 50: naming suggestion: `PreprocessFunc`

s-urbaniak (changes requested):
a few naming suggestions, and generally: let's add unit tests for the introduced preprocessing.

pkg/operator/resource/resourceapply/generic.go
slaskawi: I'd like to leave it as is. It's symmetric with `PreprocessResourcesFuncType` in `static_resource_controller.go`.

pkg/operator/resource/resourceapply/generic.go
slaskawi: I'd like to leave it as is. It's symmetric with `defaultPreprocessFunc` in `static_resource_controller.go`.

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -242,8 +258,7 @@ func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.R
 	return ret
 }
 
-func DeleteAll(ctx context.Context, clients *ClientHolder, recorder events.Recorder, manifests AssetFunc,
-	files ...string) []ApplyResult {
+func DeleteAllWithPreprocess(ctx context.Context, clients *ClientHolder, recorder events.Recorder, manifests AssetFunc, preprocessFunction func(ctx context.Context, object runtime.Object) (runtime.Object, error), files ...string) []ApplyResult {
```
s-urbaniak, line 261: same here, shouldn't this be:
```
func DeleteAllWithPreprocess(ctx context.Context, clients *ClientHolder, recorder events.Recorder, manifests AssetFunc, preprocessFunction PreprocessResourcesFuncType, files ...string) []ApplyResult {

```

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -102,6 +106,12 @@ func (c *ClientHolder) WithMigrationClient(client migrationclient.Interface) *Cl
 
 // ApplyDirectly applies the given manifest files to API server.
 func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, files ...string) []ApplyResult {
+	return ApplyDirectlyWithPreprocess(ctx, clients, recorder, cache, manifests, defaultPreprocessFunc(), files...)
+}
+
+// ApplyDirectlyWithPreprocess applies the given manifest files to API server (similarly to ApplyDirectly)
+// but also respects the pre-process function.
+func ApplyDirectlyWithPreprocess(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, preprocessFunction func(ctx context.Context, object runtime.Object) (runtime.Object, error), files ...string) []ApplyResult {
```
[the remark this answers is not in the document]
slaskawi: It depends...

We cannot freely use `PreprocessResourcesFuncType` freely here. If we did, the `static_resource_controller.go` would need to reuse this function type to implement `defaultStaticResourcesPreconditionsFunc` and we'd introduce a direct dependency between those two files. 

So the question is whether we want to introduce this dependency or not. At the moment, there is no such dependency and I wanted to keep it that way. The `static_resource_controller.go` can define its own `StaticResourcesPreconditionsFuncType` and as long as signatures match - it's fine with the compiler. 

Let me know if you still want it or not.

slaskawi:
Rebased.

PR #1308, opened 2022-02-10
factory: log warning when controller does not specify periodic resync interval

mfojtik:
This is a cheap check for whether the controllers created by the factory have the periodical resync interval specified. Without periodic resync, controllers only reconcile when informers emit an event. This makes them dependent on working informers.
Manual periodic resync make then reconcile independently from informers. This is useful when some informers are missing or not registered or some informers get stuck.

### DIFF
--- a/pkg/controller/factory/factory.go
+++ b/pkg/controller/factory/factory.go
@@ -5,14 +5,13 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/events"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 	"github.com/robfig/cron"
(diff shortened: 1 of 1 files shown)


pkg/controller/factory/factory.go
```diff
@@ -181,6 +181,13 @@ func (f *Factory) ResyncSchedule(schedules ...string) *Factory {
 	return f
 }
 
+// WithoutPeriodicResync will disable the warning emitted to log if a controller is started without specifying periodical resync.
+// This means the controller will only reconcile when an informer emits event.
+func (f *Factory) WithoutPeriodicResync() *Factory {
```
tkashem, line 186: `WithResyncWarningDisabled` or something along that line?

PR #1313, opened 2022-02-11
WIP: installerpod: wait for the static pod manifest to be mirrored to API server before terminating

mfojtik:
@deads2k I still wonder what is wrong with this approach... 

Given the timings, kubelet should notice the static pod manifest change pretty fast, so this should extend the time needed to install new revision by a few seconds. However, if this gets stale for some reason, the event here might be valuable because it will tell us exactly WHEN the copying of manifests started and how long we waited before timing out.

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -63,6 +63,8 @@ type InstallOptions struct {
 
 	// StaticPodManifestsLockFile used to coordinate work between multiple processes when writing static pod manifests
 	StaticPodManifestsLockFile string
+	StaticPodName              string
+	StaticPodNamespace         string
 
 	PodMutationFns []PodMutationFunc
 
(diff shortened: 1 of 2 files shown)


mfojtik:
proof: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/604

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -442,7 +463,13 @@ func (o *InstallOptions) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to copy: %v", err)
 	}
 
-	recorder.Eventf("StaticPodInstallerCompleted", "Successfully installed revision %s", o.Revision)
+	waitStartedAt := time.Now()
+	if err := o.waitForPodMirror(ctx); err != nil {
```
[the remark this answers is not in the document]
p0lyn0mial: I think it might be different for different operands, for example, kas needs more than 200 seconds to show up.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -442,7 +463,13 @@ func (o *InstallOptions) Run(ctx context.Context) error {
...
+		recorder.Warningf("StaticPodInstallerMirrorFailed", "Failed to observe mirrored static pod %q revision %s in API server on node %q after %s of waiting: %v", o.StaticPodName, o.Revision, o.NodeName, time.Now().Sub(waitStartedAt), err)
```
p0lyn0mial, line 468: it won't put an operator into a degraded condition

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -211,6 +213,25 @@ func (o *InstallOptions) kubeletVersion(ctx context.Context) (string, error) {
 	return node.Status.NodeInfo.KubeletVersion, nil
 }
 
+// waitForPodMirror will wait until the installer pod see that the static pod it installed is mirrored by kubelet to API server.
+// It takes context that should be bound to the overall timeout we give static pod installer to complete (120s by default).
+func (o *InstallOptions) waitForPodMirror(ctx context.Context) error {
```
p0lyn0mial, line 218: shouldn't the installer pod install a pod and not monitor it :) ?

pkg/operator/staticpod/installerpod/cmd.go
mfojtik: it could, we have event observer controller that can make operator degraded if we want to.

pkg/operator/staticpod/installerpod/cmd.go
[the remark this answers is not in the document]
mfojtik: @p0lyn0mial this is not about waiting for "ready" pod, it is about waiting for "pod was mirrored by kubelet" ... I set it to 120s for now, it means if kubelet has not seen this pod for 2 minutes and has not mirrored it, then we have problem :-)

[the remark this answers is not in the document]
p0lyn0mial: mhm, but still you need to know for how long you need to wait.
In addition to that this case (`pod was mirrored by kubelet`) seems to already be covered by the installer-controller, checkout out

https://github.com/openshift/library-go/blob/2079b9a0825c65d933ad22792eda9582090c00d7/pkg/operator/staticpod/controller/installer/installer_controller.go#L738

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: would that be deterministic? I always thought events are not quartered to be delivered

PR #1316, opened 2022-02-11
Bug 2053582: Track static pod lifecycle

dgrisonnet:
Effort based on https://github.com/openshift/cluster-kube-apiserver-operator/pull/1317 and including the following gists written by @p0lyn0mial to address a comment regarding the termination grace period:
- https://gist.github.com/p0lyn0mial/363f5f1c04c1ff6b607d8358a123dc53
- https://gist.github.com/p0lyn0mial/abd14f30fe1f83b849e5f7f9f9811275

### DIFF
--- a/pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
+++ b/pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
@@ -1,0 +1,268 @@
+package missingstaticpodcontroller
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	corelisterv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/library-go/pkg/operator/staticpod/internal"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type missingStaticPodController struct {
+	operatorClient                    v1helpers.StaticPodOperatorClient
+	podListerForTargetNamespace       corelisterv1.PodNamespaceLister
+	configMapListerForTargetNamespace corelisterv1.ConfigMapNamespaceLister
+	targetNamespace                   string
+	operandName                       string
+
+	lastEventEmissionPerNode lastEventEmissionPerNode
+}
+
+type lastEventEmissionPerNode map[string]struct {
+	revision  int
+	timestamp time.Time
+}
+
+// New checks the latest static pods pods and uses the installer to get the
+// latest revision.  If the installer pod was successful, and if it has been
+// longer than the terminationGracePeriodSeconds+120seconds since the installer
+// pod completed successfully, and if the static pod is not at the correct
+// revision, this controller will go degraded.  It will also emit an event for
+// detection in CI.
+func New(
+	operatorClient v1helpers.StaticPodOperatorClient,
+	kubeInformersForTargetNamespace informers.SharedInformerFactory,
+	eventRecorder events.Recorder,
+	targetNamespace string,
+	staticPodName string,
+) factory.Controller {
+	c := &missingStaticPodController{
+		operatorClient:                    operatorClient,
+		podListerForTargetNamespace:       kubeInformersForTargetNamespace.Core().V1().Pods().Lister().Pods(targetNamespace),
+		configMapListerForTargetNamespace: kubeInformersForTargetNamespace.Core().V1().ConfigMaps().Lister().ConfigMaps(targetNamespace),
+		targetNamespace:                   targetNamespace,
+		operandName:                       staticPodName,
+		lastEventEmissionPerNode:          make(lastEventEmissionPerNode),
+	}
+	return factory.New().
+		ResyncEvery(time.Minute).
+		WithInformers(
+			operatorClient.Informer(),
+			kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		).
+		WithBareInformers(
+			kubeInformersForTargetNamespace.Core().V1().ConfigMaps().Informer(),
+		).
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
+		ToController("MissingStaticPodController", eventRecorder)
+}
+
+func (c *missingStaticPodController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	_, originalOperatorStatus, _, err := c.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return err
+	}
+
+	installerPods, err := c.podListerForTargetNamespace.List(labels.SelectorFromSet(labels.Set{"app": "installer"}))
+	if err != nil {
+		return err
+	}
+
+	// get the most recent installer pod for each node
+	latestInstallerPodsByNode, err := getMostRecentInstallerPodByNode(installerPods)
+	if err != nil {
+		return err
+	}
+
+	errors := make([]string, 0, len(latestInstallerPodsByNode))
+	for node, latestInstallerPodOnNode := range latestInstallerPodsByNode {
+		installerPodRevision, err := internal.GetRevisionOfPod(latestInstallerPodOnNode)
+		if err != nil {
+			// we expect every installer pod to have a installerPodRevision in its name, unexpected error here
+			return fmt.Errorf("failed to get installerPodRevision for installer pod %q - %w", latestInstallerPodOnNode.Name, err)
+		}
+
+		finishedAt, ok := installerPodFinishedAt(latestInstallerPodOnNode)
+		if !ok {
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+
+		gracePeriod, err := c.getStaticPodTerminationGracePeriodSecondsForRevision(installerPodRevision)
+		if err != nil {
+			return err
+		}
+		threshold := gracePeriod + 120*time.Second
+
+		staticPodRevisionOnThisNode := getStaticPodCurrentRevision(node, originalOperatorStatus)
+		if staticPodRevisionOnThisNode == 0 {
+			// skip the bootstraping phase
+			continue
+		}
+
+		if time.Since(finishedAt) > threshold &&
+			staticPodRevisionOnThisNode < installerPodRevision {
+			// if we are here:
+			//  a: the latest installer pod successfully completed at finishedAt
+			//  b: it has been more than 'terminationGracePeriodSeconds' + 120s since the
+			//     installer pod has completed at finishedAt
+			//  c. the static pod is not at correct installerPodRevision yet
+			// then all of the above conditions are true, so we should produce an event
+
+			// only emit an event if for a particular node:
+			// - this is the first time we see a failure
+			// - this is a failure for new revision
+			// - the previously reported event was at least 30 min ago
+			lastEventEmission, found := c.lastEventEmissionPerNode[node]
+			if !found || installerPodRevision != lastEventEmission.revision || time.Since(lastEventEmission.timestamp) > 30*time.Minute {
+				syncCtx.Recorder().Eventf("MissingStaticPod", "static pod lifecycle failure - static pod: %q in namespace: %q for revision: %d on node: %q didn't show up, waited: %v",
+					c.operandName, c.targetNamespace, installerPodRevision, node, threshold)
+
+				lastEventEmission.revision = installerPodRevision
+				lastEventEmission.timestamp = time.Now()
+				c.lastEventEmissionPerNode[node] = lastEventEmission
+			}
+
+			errors = append(errors, fmt.Sprintf("static pod lifecycle failure - static pod: %q in namespace: %q for revision: %d on node: %q didn't show up, waited: %v",
+				c.operandName, c.targetNamespace, installerPodRevision, node, threshold))
+		}
+	}
+
+	if len(errors) > 0 {
+		return fmt.Errorf(strings.Join(errors, "\n"))
+	}
+
+	return nil
+}
+
+// getStaticPodTerminationGracePeriodSecondsForRevision reads the static pod manifest from a configmap
+// in the target namespace and returns the value of terminationGracePeriodSeconds.
+// In case no value was provided for the static pod it returns a default value of 30 seconds.
+func (c *missingStaticPodController) getStaticPodTerminationGracePeriodSecondsForRevision(revision int) (time.Duration, error) {
+	staticPodKeyName := "pod.yaml"
+	staticPodConfigMapName := fmt.Sprintf("%s-pod-%d", c.operandName, revision)
+
+	staticPodConfigMap, err := c.configMapListerForTargetNamespace.Get(staticPodConfigMapName)
+	if err != nil {
+		return 0, err
+	}
+
+	rawStaticPodManifest, hasPodKey := staticPodConfigMap.Data[staticPodKeyName]
+	if !hasPodKey {
+		return 0, fmt.Errorf("didn't find required key %q in cm: %s/%s", staticPodKeyName, staticPodConfigMapName, c.targetNamespace)
+	}
+
+	serializedStaticPod, err := resourceread.ReadPodV1([]byte(rawStaticPodManifest))
+	if err != nil {
+		return 0, err
+	}
+
+	if serializedStaticPod.Spec.TerminationGracePeriodSeconds == nil {
+		klog.V(6).Infof("optional field: %s.spec.terminationGracePeriodSeconds was not specified in cm: %s, returning default value of 30s", staticPodKeyName, staticPodConfigMapName)
+		return 30 * time.Second, nil
+	}
+
+	return time.Duration(*serializedStaticPod.Spec.TerminationGracePeriodSeconds) * time.Second, nil
+}
+
+// TODO: consider the following logic to get the static pods revision:
+// - list all pods
+// - filter only static pods, those with annotation: `kubernetes.io/config.mirror`
+// - there will only be one for each master
+// - determine the node based on .spec.nodeName
+// - get the revision from label: `revision`
+// This would allow reducing the delay and remove the dependency on the node
+// status.
+func getStaticPodCurrentRevision(node string, status *operatorv1.StaticPodOperatorStatus) int {
+	var nodeStatus *operatorv1.NodeStatus
+	for i := range status.NodeStatuses {
+		if status.NodeStatuses[i].NodeName == node {
+			nodeStatus = &status.NodeStatuses[i]
+			break
+		}
+	}
+
+	if nodeStatus == nil {
+		return 0
+	}
+	return int(nodeStatus.CurrentRevision)
+}
+
+func getMostRecentInstallerPodByNode(pods []*corev1.Pod) (map[string]*corev1.Pod, error) {
+	mostRecentInstallerPodByNode := map[string]*corev1.Pod{}
+	byNodes, err := getInstallerPodsByNode(pods)
+	if err != nil {
+		return nil, err
+	}
+
+	for node, installerPodsOnThisNode := range byNodes {
+		if len(installerPodsOnThisNode) == 0 {
+			continue
+		}
+		sort.Sort(internal.ByRevision(installerPodsOnThisNode))
+		mostRecentInstallerPodByNode[node] = installerPodsOnThisNode[len(installerPodsOnThisNode)-1]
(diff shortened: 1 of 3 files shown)


dgrisonnet:
proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1319

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+
+		staticPodRevisionOnThisNode := getStaticPodCurrentRevision(node, originalOperatorStatus)
+		gracePeriod, err := c.getStaticPodTerminationGracePeriodSecondsForRevision(staticPodRevisionOnThisNode)
+		if err != nil {
+			return err
+		}
+		threshold := time.Duration(gracePeriod) + 10*time.Second
```
dgrisonnet, line 99: Should we consider increasing the margin? 10 seconds seems a bit low, increasing it to 1 minute might be safer and we shouldn't miss any error since once a static pod is missing, we can expect it not to appear for a while.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
```
p0lyn0mial, line 11: nit: please move it down to the `openshift` group.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+type missingStaticPodController struct {
+	operatorClient                    v1helpers.StaticPodOperatorClient
+	podLister                         corelisterv1.PodLister
```
p0lyn0mial, line 27: nit: `podListerForTargetNamespace`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+
+		finishedAt, ok := installerPodFinishedAt(latestInstallerPodOnNode)
+		if !ok {
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+
+		staticPodRevisionOnThisNode := getStaticPodCurrentRevision(node, originalOperatorStatus)
+		gracePeriod, err := c.getStaticPodTerminationGracePeriodSecondsForRevision(staticPodRevisionOnThisNode)
```
p0lyn0mial, line 95: I think this should be for `installerPodRevision` not for `staticPodRevisionOnThisNode`, am I right?
We are interested in knowing `gracePeriod` for a pod being installed not the previous one

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		return fmt.Errorf(strings.Join(errors, "\n"))
+	}
+
+	return nil
+}
+
+// getStaticPodTerminationGracePeriodSecondsForRevision reads the static pod manifest from a configmap
+// in the target namespace and returns the value of terminationGracePeriodSeconds
+// in case no value was provided for the static pod it returns a default value of 30 seconds
+func (c *missingStaticPodController) getStaticPodTerminationGracePeriodSecondsForRevision(revision int) (int64, error) {
```
p0lyn0mial, line 127: nit: now i see that it could return `(time.Duration, error)`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+		operandName:                       staticPodName,
+	}
+	return factory.New().
+		ResyncEvery(time.Minute).
+		WithInformers(
+			operatorClient.Informer(),
+			kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		).
+		WithSync(c.sync).
+		WithSyncDegradedOnError(operatorClient).
```
p0lyn0mial, line 59: I think it will end up being reported as `Type:MissingStaticPodControllerDegraded, Reason: SyncError` which is okay

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
p0lyn0mial: let's wait until the proof pr for `kas-o` finishes

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+			// if we are here:
+			//  a: the latest installer pod successfully completed at finishedAt
+			//  b: it has been more than 'terminationGracePeriodSeconds' + 10s since the
+			//     installer pod has completed at finishedAt
+			//  c. the static pod is not at correct installerPodRevision yet
+			// then all of the above conditions are true, so we should go degraded
+			syncCtx.Recorder().Eventf("MissingStaticPod", "static pod lifecycle failure - installer pod: %q with revision: %d completed at: %s, static pod revision: %d",
+				latestInstallerPodOnNode.Name, installerPodRevision, finishedAt, staticPodRevisionOnThisNode)
+
+			errors = append(errors, fmt.Sprintf("static pod lifecycle failure - installer pod: %q with revision: %d completed at: %s, static pod revision: %d",
```
p0lyn0mial, line 112: do we want to report the `threshold` value? (same for the event)

p0lyn0mial: oh and maybe, slightly change the msg so that we know why it failed i.e. `static pod for revision: %d didn't show up, waited: %v`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+		if time.Since(finishedAt) > threshold &&
+			staticPodRevisionOnThisNode < installerPodRevision {
+			// if we are here:
+			//  a: the latest installer pod successfully completed at finishedAt
+			//  b: it has been more than 'terminationGracePeriodSeconds' + 10s since the
+			//     installer pod has completed at finishedAt
+			//  c. the static pod is not at correct installerPodRevision yet
+			// then all of the above conditions are true, so we should go degraded
+			syncCtx.Recorder().Eventf("MissingStaticPod", "static pod lifecycle failure - installer pod: %q with revision: %d completed at: %s, static pod revision: %d",
```
p0lyn0mial, line 109: I think it can produce many events, at least per every `sync` loop. I think we should find a way to avoid it.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+type missingStaticPodController struct {
+	operatorClient                    v1helpers.StaticPodOperatorClient
+	podLister                         corelisterv1.PodLister
+	configMapListerForTargetNamespace corelisterv1.ConfigMapNamespaceLister
+	targetNamespace                   string
+	operandName                       string
+}
+
+// New checks the latest installer pods.  If the installer pod was successful, and if it has
```
p0lyn0mial, line 33: I think it checks the static pods, it uses the installer to get the latest revision.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,263 @@
...
+// been longer than the terminationGracePeriodSeconds+10seconds since the installer pod
+// completed successfully, and if the static pod is not at the correct revision, this
+// controller will go degraded.
+// It will also emit an event for detection in CI,
```
p0lyn0mial, line 37: nit: rm `,` at the end.

tkashem:
/retitle Bug 2053582: Track static pod lifecycle

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
tkashem, line 59: we have decided to drop it for now - we are this close to the release and a new controller that degrades a cluster and blocks upgrade - false positives could trigger many customer cases in the real world

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
tkashem: yes, `installerPodRevision`. because that's the desired version we want the static pod to be at

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
dgrisonnet: this is actually the way goimports is sorting the imports. First you have go std libs, then external library and finally local libraries.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
dgrisonnet: at maximum the controller will be responsible for 1620 active events with 9 created per minute:
```
3 (kas, etcd, kcm) * 3 (replicas) * 180 (event ttl)
```
event ttl: https://github.com/openshift/cluster-kube-apiserver-operator/blob/69a9edba92634e68[…]57c77aa063d889fdf30bed/bindata/assets/config/defaultconfig.yaml

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
dgrisonnet: good catch, added a test case to cover that

tkashem:
cc @deads2k

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	c := &missingStaticPodController{
+		operatorClient:                    operatorClient,
+		podListerForTargetNamespace:       kubeInformersForTargetNamespace.Core().V1().Pods().Lister().Pods(targetNamespace),
+		configMapListerForTargetNamespace: kubeInformersForTargetNamespace.Core().V1().ConfigMaps().Lister().ConfigMaps(targetNamespace),
+		targetNamespace:                   targetNamespace,
+		operandName:                       staticPodName,
+	}
+	return factory.New().
+		ResyncEvery(time.Minute).
+		WithInformers(
```
p0lyn0mial, line 55: add `configMapListerForTargetNamespace` we need to make sure it has synced before we start using it. If you don't want to react to changes then add it to `WithBareInformers`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,265 @@
...
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+
+		gracePeriod, err := c.getStaticPodTerminationGracePeriodSecondsForRevision(installerPodRevision)
+		if err != nil {
+			return err
+		}
+		threshold := gracePeriod + 10*time.Second
+		staticPodRevisionOnThisNode := getStaticPodCurrentRevision(node, originalOperatorStatus)
```
p0lyn0mial, line 102: I think we need to skip when `staticPodRevisionOnThisNode == 0`, this can happen during bootstrapping phase.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,265 @@
...
+		if time.Since(finishedAt) > threshold &&
+			staticPodRevisionOnThisNode < installerPodRevision {
+			// if we are here:
+			//  a: the latest installer pod successfully completed at finishedAt
+			//  b: it has been more than 'terminationGracePeriodSeconds' + 10s since the
+			//     installer pod has completed at finishedAt
+			//  c. the static pod is not at correct installerPodRevision yet
+			// then all of the above conditions are true, so we should produce an event
+			syncCtx.Recorder().Eventf("MissingStaticPod", "static pod lifecycle failure - static pod: %q in namespace: %q for revision: %d didn't show up, waited: %v",
```
p0lyn0mial, line 111: I think we miss info about a `node`. For example:

`static pod lifecycle failure, static pod: %q in namespace: %q for revision: %d on node: %v didn't show up, waited: %v`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
p0lyn0mial: on an sno cluster, in case of a fallback (currentRevision < mostRecentRevision), it can produce events forever.

dgrisonnet: that statement should be valid for any type of cluster not only sno no? Also, even if events will be produced indefinitely, their retention will be very short so it should have a limited impact on the clusters

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,274 @@
...
+		if !ok {
+			// either it's in the process of running, or it ran into an error
+			continue
+		}
+
+		gracePeriod, err := c.getStaticPodTerminationGracePeriodSecondsForRevision(installerPodRevision)
+		if err != nil {
+			return err
+		}
+		threshold := gracePeriod + 10*time.Second
```
p0lyn0mial, line 104: it looks like `10 s` might be too tight. I have just looked over [this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/605/pull-ci-openshift-cluster-kube-controller-manager-operator-master-e2e-aws-ccm/1493177597461270528) run and it looks like a kcm pod was reported after `102`s, here is a timeline

at `11:40:39` the installer-9-ip-10-0-151-209.ec2.internal finished (Exit code 0)

at `11:40:52` the node agent started deleting the previous instance `"Trying to delete pod" pod="openshift-kube-controller-manager/kube-controller-manager-ip-10-0-151-209.ec2.internal" podUID=1007ded5-acb3-4f6b-8a62-6d08f7a910b9`

at `11:41:49` the node agent reported the new instance as running `"Pod status updated" pod="openshift-kube-controller-manager/kube-controller-manager-ip-10-0-151-209.ec2.internal" status=Running`

at `11:42:01` the installer controller reported `CurrentNode=9` `Updated node "ip-10-0-151-209.ec2.internal" from revision 7 to 9 because static pod is ready`

p0lyn0mial: similarly [in ](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-controller-manager-operator/605/pull-ci-openshift-cluster-kube-controller-manager-operator-master-e2e-aws-operator/1493177597528379392/artifacts/e2e-aws-operator/gather-must-gather/artifacts/), a kcm pod was reported after `105 s`:

the installer-pod for revision `6` finished at `2022-02-14T11:30:18Z` and 
the kcm pod was reported at `11:32:03` `Updated node "ip-10-0-231-223.us-east-2.compute.internal" from revision 5 to 6 because static pod is ready`

soltysh: @p0lyn0mial it looks like kcmo does not set `TerminationGracePeriodSeconds` so for those cases we should default to 30s, just like specified in https://github.com/kubernetes/kubernetes/blob/b591acca57f918404f35b8905021f2bda0e42866/staging/src/k8s.io/api/core/v1/types.go#L2722

soltysh: I see that's being handled there, properly, nvmd.

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,274 @@
...
+		if len(nodeName) == 0 {
+			return nil, fmt.Errorf("node name for installer pod %q is empty", pod.Name)
+		}
+		byNodes[nodeName] = append(byNodes[nodeName], pod)
+	}
+
+	return byNodes, nil
+}
+
+type byRevision []*corev1.Pod
```
soltysh, line 216: We're copying code from https://github.com/openshift/library-go/blob/18f172cff934d86b5fe70f0f26d101c97a6121e9/pkg/operator/staticpod/installerpod/cmd.go#L553-L583 why not creating a package with just that?

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
deads2k, line 104: 60s is fine with me.

deads2k: 120s is ok

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -28,6 +28,9 @@ type missingStaticPodController struct {
 	configMapListerForTargetNamespace corelisterv1.ConfigMapNamespaceLister
 	targetNamespace                   string
 	operandName                       string
+
+	lastEventEmittedForRevision int
```
p0lyn0mial, line 32: I think should keep a state by `node`, i.e. `map[string]time.Time`

p0lyn0mial: plus some clean-up logic that removes entries for new and/or successful revisions

p0lyn0mial: so maybe we need to keep by `node and revision`

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
dgrisonnet: good point, I opened https://github.com/openshift/library-go/pull/1317

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+type lastEventEmissionPerNode map[string]struct {
+	revision  int
+	timestamp time.Time
+}
+
+// New checks the latest static pods pods and uses the installer to get the
+// latest revision.  If the installer pod was successful, and if it has been
+// longer than the terminationGracePeriodSeconds+120seconds since the installer
```
tkashem, line 42: nit: `terminationGracePeriodSeconds+120 seconds

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		return fmt.Errorf(strings.Join(errors, "\n"))
+	}
+
+	return nil
+}
+
+// getStaticPodTerminationGracePeriodSecondsForRevision reads the static pod manifest from a configmap
+// in the target namespace and returns the value of terminationGracePeriodSeconds.
+// In case no value was provided for the static pod it returns a default value of 30 seconds.
+func (c *missingStaticPodController) getStaticPodTerminationGracePeriodSecondsForRevision(revision int) (time.Duration, error) {
```
tkashem, line 156: can we add a TODO with David's suggestion for the future? https://github.com/openshift/library-go/pull/1316/files#r805889436

pkg/operator/staticpod/controller/missingstaticpod/missing_static_pod_controller.go
dgrisonnet: that is already what is written no?

tkashem: yes, `120seconds`-> `120 seconds` :)

tkashem:
/lgtm
/approved

/hold 
(once the proof PRs have a few greens on the runs, we can cancel the hold)

tkashem:
(results in proof PRs look good)

/hold cancel

PR #1317, opened 2022-02-14
Add package shared between all staticpod controllers

dgrisonnet:
Refactor some of the code used by staticpod controllers to make it reusable by all the controllers as per https://github.com/openshift/library-go/pull/1316#discussion_r805877420.

### DIFF
--- a/pkg/operator/staticpod/internal/utils.go
+++ b/pkg/operator/staticpod/internal/utils.go
@@ -1,0 +1,41 @@
+package internal
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type ByRevision []*corev1.Pod
+
+func (s ByRevision) Len() int {
+	return len(s)
+}
+func (s ByRevision) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+func (s ByRevision) Less(i, j int) bool {
+	jRevision, err := GetRevisionOfPod(s[j])
+	if err != nil {
+		return true
+	}
+	iRevision, err := GetRevisionOfPod(s[i])
+	if err != nil {
+		return false
+	}
+	return iRevision < jRevision
+}
+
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/utils_test.go
```diff
@@ -0,0 +1,80 @@
...
+			name:        "installer pod with bad revision",
+			pod:         &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "installer-a-node-1"}},
+			expectError: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			revision, err := GetRevisionOfPod(tc.pod)
+			if err != nil != tc.expectError {
```
p0lyn0mial, line 72: `if err != nil && err != tc.expectError`

p0lyn0mial: ah, `expectError` is `boolean`

pkg/operator/staticpod/internal/utils.go
```diff
@@ -0,0 +1,41 @@
+package internal
```
p0lyn0mial, line 1: it will allow us to keep it private for packages outside of `staticpod`

pkg/operator/staticpod/internal/utils_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			revision, err := GetRevisionOfPod(tc.pod)
+			if err != nil != tc.expectError {
+				t.Fatalf("expected error to have occured to be %t, got: %t, err: %v", tc.expectError, err != nil, err)
+			}
+			if err == nil && revision != tc.expectedRevision {
+				t.Fatalf("expected revision to be %d, got: %d", tc.expectedRevision, revision)
+			}
```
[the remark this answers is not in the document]
p0lyn0mial: seems it will be handled by `if err != nil != tc.expectError {`

PR #1324, opened 2022-02-28
Recreate CSIDriver on change

jsafrane:
`CSIDriver.Spec` is immutable after creation. Some of its fields should not be changed, for example changing `attachRequired` from `true` to `false` when a volume is being attached may result in orphaned VolumeAttachment objects.

But most of its fields can be changed without any issues. We need to change `fsGroupPolicy` from default (empty) to `File` in all CSI drivers that are based on a block file to prevent fsGroup issues like https://bugzilla.redhat.com/show_bug.cgi?id=2058626.

Also, remove support for CSIDriver v1beta1, nobody uses it by now.

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -2,6 +2,7 @@ package resourceapply
 
 import (
 	"context"
+	"fmt"
 
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -57,7 +58,17 @@ func ApplyStorageClass(ctx context.Context, client storageclientv1.StorageClasse
 }
 
 // ApplyCSIDriver merges objectmeta, does not worry about anything else
-func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter, recorder events.Recorder, required *storagev1.CSIDriver) (*storagev1.CSIDriver, bool, error) {
+func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter, recorder events.Recorder, requiredOriginal *storagev1.CSIDriver) (*storagev1.CSIDriver, bool, error) {
+
+	required := requiredOriginal.DeepCopy()
+	if required.Annotations == nil {
+		required.Annotations = map[string]string{}
+	}
+	err := SetSpecHashAnnotation(&required.ObjectMeta, required.Spec)
+	if err != nil {
+		return nil, false, err
+	}
+
 	existing, err := client.CSIDrivers().Get(ctx, required.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		requiredCopy := required.DeepCopy()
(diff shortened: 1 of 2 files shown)


jsafrane:
/retest

flake in TestControllerWithQueueFunction ?

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -103,21 +80,38 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
...
+		reportUpdateEvent(recorder, required, err)
+		return actual, true, err
+	}
+
+	existingCopy.Spec = required.Spec
+	existingCopy.ObjectMeta.ResourceVersion = ""
+	// Spec is read-only after creation. Delete and re-create the object
+	err = client.CSIDrivers().Delete(ctx, existingCopy.Name, metav1.DeleteOptions{})
+	reportDeleteEvent(recorder, existingCopy, err, "Deleting CSIDriver to re-create it with updated parameters")
+	if err != nil && !apierrors.IsNotFound(err) {
```
gnufied, line 110: Do we need to poll for `CSIDriver` object to be actually removed from API server before recreating the new CSIDriver object?

jsafrane: In case the CSIDriver can't be deleted from any reason (like a finalizer), `Create` below will fail and the operator will retry in the next sync.

I can see that other Apply functions do the same: https://github.com/openshift/library-go/blob/a3f30cb5faf9e47565ff1a6c177592b75aa9f9c1/pkg/operator/resource/resourceapply/core.go#L431-L438


This will result in some errors and the corresponding operator Degraded, if it takes too long time / too many attemtps, but IMO if we poll until a CSIDriver is deleted, it would lead to the same Degraded condition - `ApplyCSIDriver` should report an error if the old CSIDriver exists after a timeout.

jsafrane: I updated the code to return something else than a bare `AlreadyExists` from `Create()` below in case there is a finalizer or the deletion did not proceed from any other reason.

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -103,21 +81,46 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
...
+	if err != nil && apierrors.IsAlreadyExists(err) {
+		// Delete() few lines above did not really delete the object,
+		// the API server is probably waiting for a finalizer removal or so.
+		// Report an error, but something else than "Already exists", because
+		// that would be very confusing - Apply failed because the object
+		// already exists???
+		err = fmt.Errorf("failed to re-create CSIDriver object %s, waiting for the original object to be deleted", existingCopy.Name)
+	}
+	reportCreateEvent(recorder, existingCopy, err)
 	return actual, true, err
```
bertinatto, line 124: hm, what if `err` above is something else than `AlreadyExists`? I think this should return `false` instead

jsafrane: It returned `true` before this PR, In addition, `Delete()` before succeeded, therefore would consider that "something has changed" and `true` is correct.

PR #1327, opened 2022-03-14
Bugzilla 2063342: audit/manifests: exclude tokenreviews and tokenrequests from audit logs

s-urbaniak:
/cc @sttts @tkashem @stlaz

### DIFF
--- a/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
+++ b/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
@@ -4,8 +4,8 @@
   - group: "route.openshift.io"
     resources: ["routes"]
   - resources: ["secrets"]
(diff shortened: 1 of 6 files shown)


pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
```diff
@@ -4,8 +4,8 @@
   - group: "route.openshift.io"
     resources: ["routes"]
   - resources: ["secrets"]
-- level: Metadata
```
s-urbaniak, line 7: @sttts was this just cosmetic, do you remember?

s-urbaniak: confirmed OOB this is cosmentic

PR #1333, opened 2022-03-16
Adding method from cvo for capabilities

LalatenduMohanty:
Moving include() method from CVO

as the code will be reused in OC CLI and CVO. The include() method currently exists in https://github.com/openshift/cluster-version-operator/blob/master/pkg/payload/payload.go#L221 and will be replaced by the code from library-go in future.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -6,13 +6,20 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	CapabilityAnnotation  = "capability.openshift.io/name"
+	DefaultClusterProfile = "self-managed-high-availability"
 )
 
 // resourceId uniquely identifies a Kubernetes resource.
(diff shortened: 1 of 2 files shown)


jottofar:
With the proposed change to not filter on capabilities at load time but rather at apply time I changed `include` to return the capabilities from a given manifest or an error if an unknown capability is found. That then also replaces method ` CheckResourceEnablement` with `GetResourceCapabilities`. Like so:

x-special/nautilus-clipboard
copy
file:///home/jottofar/tmp/tmp.go

LalatenduMohanty:
> With the proposed change to not filter on capabilities at load time but rather at apply time I changed `include` to return the capabilities from a given manifest or an error if an unknown capability is found. That then also replaces method ` CheckResourceEnablement` with `GetResourceCapabilities`

I checked https://github.com/openshift/cluster-version-operator/pull/754/ and we should move ahead with this PR merge and once https://github.com/openshift/cluster-version-operator/pull/754/ is merged we can move the changes to this repo or even doing the changes directly in this repo. It depends on on the timing of merge of this PR. I am optimistic that we can get this PR merged asap.

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+		return fmt.Errorf("%s=%s", excludeAnnotation, v)
+	}
+
+	featureGateAnnotation := "release.openshift.io/feature-gate"
+	featureGateAnnotationValue, featureGateAnnotationExists := annotations[featureGateAnnotation]
+	if featureGateAnnotationValue == "TechPreviewNoUpgrade" && !includeTechPreview {
+		return fmt.Errorf("tech-preview excluded, and %s=%s", featureGateAnnotation, featureGateAnnotationValue)
+	}
+	// never include the manifest if the feature-gate annotation is outside of allowed values (only TechPreviewNoUpgrade is currently allowed)
+	if featureGateAnnotationExists && featureGateAnnotationValue != "TechPreviewNoUpgrade" {
```
wking, line 131: Worth using `string(configv1.TechPreviewNoUpgrade)` here, to avoid the local string?

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	return validateResourceId(m.id)
 }
 
+func (m *Manifest) Include(excludeIdentifier string, includeTechPreview bool, profile string, capabilities ClusterCapabilities) error {
```
wking, line 113: If we want to use a single "should we include this?" method both for load-time filtering (cluster profile, etc.) and apply-time filtering (capabilities, etc.), we should make these arguments pointers, and internally only filter with the non-`nil` terms.

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+
+	return CheckResourceEnablement(annotations, capabilities)
+}
+
+// CheckResourceEnablement, given resource annotations and defined cluster capabilities, checks if the capability
+// annotation exists. If so, each capability name is validated against the known set of capabilities. Each valid
+// capability is then checked if it is disabled. If any invalid capabilities are found an error is returned listing
+// all invalid capabilities. Otherwise, if any disabled capabilities are found an error is returned listing all
+// disabled capabilities.
+func CheckResourceEnablement(annotations map[string]string, capabilities ClusterCapabilities) error {
```
wking, line 150: This does not need to be part of the public API (the public aspect is `Include`, this is just an internal helper).  Can we rename to `checkResourceEnablement`?

pkg/manifest/manifest.go
LalatenduMohanty: Agree. Will fix it

pkg/manifest/manifest.go
LalatenduMohanty: yes, will fix it

LalatenduMohanty: Fixed in https://github.com/openshift/library-go/commit/d2d9b0ce7cccf99dfd5903d246b0e619dc764096

pkg/manifest/manifest.go
LalatenduMohanty: Fixed in https://github.com/openshift/library-go/commit/d2d9b0ce7cccf99dfd5903d246b0e619dc764096

pkg/manifest/manifest.go
LalatenduMohanty: Fixed in https://github.com/openshift/library-go/commit/e35ac895df110ebbd4157d2e853f4715dd501ad6

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
 }
 
+func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview bool, profile *string, capabilities *ClusterCapabilities) error {
+
+	annotations := m.Obj.GetAnnotations()
+	if annotations == nil {
+		return fmt.Errorf("no annotations")
+	}
+
+	excludeAnnotation := fmt.Sprintf("exclude.release.openshift.io/%s", *excludeIdentifier)
```
wking, line 120: you need to wrap all this in an `if excludeIdentifier != nil { ... }` block to avoid panicking if `excludeIdentifier` is nil.  Same for the other pointer arguments.

pkg/manifest/manifest.go
```diff
@@ -31,6 +38,11 @@ type resourceId struct {
 	Namespace string
 }
 
+type ClusterCapabilities struct {
```
wking, line 41: We want to drop this and use the API type instead.

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	return validateResourceId(m.id)
 }
 
+func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview bool, profile *string, capabilities *ClusterCapabilities) error {
```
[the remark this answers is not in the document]
wking: Discussing in our team meeting, we want `*configv1.ClusterVersionCapabilitiesStatus` as the arg here, so the library API doesn't need to have to have opinions about a similar, alternative type.  The alternative map-based type can remain internal to the CVO.

pkg/manifest/manifest.go
```diff
@@ -98,6 +110,69 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+	if val, ok := annotations[profileAnnotation]; ok && val != "true" {
+		return fmt.Errorf("unrecognized value %s=%s", profileAnnotation, val)
+	} else if !ok {
+		return fmt.Errorf("%s unset", profileAnnotation)
+	}
+
+	return checkResourceEnablement(annotations, capabilities)
+}
+
+// CheckResourceEnablement, given resource annotations and defined cluster capabilities, checks if the capability
```
wking, line 145: nit: `checkResourceEnablement` , to match the now-internal function name

LalatenduMohanty: fixed in https://github.com/openshift/library-go/commit/f529590c5da24451b978dd972aade0a4bc3d40bd

pkg/manifest/manifest.go
LalatenduMohanty: fixed in https://github.com/openshift/library-go/commit/f529590c5da24451b978dd972aade0a4bc3d40bd

pkg/manifest/manifest.go
LalatenduMohanty: fixed in https://github.com/openshift/library-go/commit/f529590c5da24451b978dd972aade0a4bc3d40bd

pkg/manifest/manifest.go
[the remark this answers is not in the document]
LalatenduMohanty: fixed in https://github.com/openshift/library-go/commit/f529590c5da24451b978dd972aade0a4bc3d40bd

pkg/manifest/manifest.go
```diff
@@ -98,6 +105,100 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+	annotations := m.Obj.GetAnnotations()
+	if annotations == nil {
+		return fmt.Errorf("no annotations")
+	}
+
+	var excludeAnnotation string
+	if excludeIdentifier != nil {
+		excludeAnnotation = fmt.Sprintf("exclude.release.openshift.io/%s", *excludeIdentifier)
+	}
+	if v := annotations[excludeAnnotation]; v == "true" {
```
wking, line 123: unlikely to have an empty-string annotation, but this whole check should be inside the `nil` guard:

```go
if excludeIdentifier != nil {
  excludeAnnotation := fmt.Sprintf("exclude.release.openshift.io/%s", *excludeIdentifier)
  if v := annotations[excludeAnnotation]; v == "true" {
    return fmt.Errorf("%s=%s", excludeAnnotation, v)
  }
}
```

And same for `profile` below.

pkg/manifest/manifest.go
```diff
@@ -98,6 +105,100 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+		profileAnnotation = fmt.Sprintf("include.release.openshift.io/%s", *profile)
+	}
+
+	if val, ok := annotations[profileAnnotation]; ok && val != "true" {
+		return fmt.Errorf("unrecognized value %s=%s", profileAnnotation, val)
+	} else if !ok {
+		return fmt.Errorf("%s unset", profileAnnotation)
+	}
+
+	// If there is no capabilities defined in a release then we do not need to check presence of capabilities in the manifest
```
wking, line 148: This is really "the caller didn't ask us to filter by capability", which is fairly orthogonal to whether there are capabilities defined in the release.  And we already cover the meaning of `nil` arguments in English in the functions godocs, so I expect we can just drop this line.  Also, just a comment, so if you don't want to drop or rephrase, I don't have a problem landing the PR with this line unchanged.

pkg/manifest/manifest.go
LalatenduMohanty: I agree. Fixed in https://github.com/openshift/library-go/commit/096d671148e65ca7b31a78b56dfbd2f15fd9ab94

pkg/manifest/manifest.go
LalatenduMohanty: It is not big deal , but the comment will help the code readability. Lets keep it.

wking (approved):
/lgtm

jottofar:
/hold
This is not the `include` interface my CVO PR needs. Why not use what I have defined in my PR for `include` and `checkResourceEnablement` that returns the resource's validated capabilities?

PR #1335, opened 2022-03-18
Bug: 2039256 Validate a hostname the same as openshift-apiserver

candita:
ValidateHost checks that a route's host name satisfies OpenShift DNS requirements,
with the assumption that the caller is not passing in an empty host name.
ValidateHost mimics the host validation in the validateRoute method in the
[openshift-apiserver/pkg/route/apis/route/validation/validation package.](https://github.com/openshift/openshift-apiserver/blob/db22c038b9184b4ae823bba3f57b669526999a2b/pkg/route/apis/route/validation/validation.go#L58-L72)

The host name length must be no more than 253 characters.  The only characters
allowed in host name are alphanumeric characters, '-' or '.', and it must start
and end with an alphanumeric character. A trailing dot is NOT allowed.
If the allowNonCompliant annotation is not set to true, the host name must in
addition consist of one or more labels, with each label no more than 63 characters
from the character set described above, and each label must start and end with an
alphanumeric character.

### DIFF
--- a/pkg/route/routeapihelpers/routeapihelpers.go
+++ b/pkg/route/routeapihelpers/routeapihelpers.go
@@ -4,6 +4,7 @@ package routeapihelpers
 import (
 	"fmt"
 	"net/url"
+	"strings"
 
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


rfredette:
Sorry it took me so long to get back to you on this
/lgtm

PR #1336, opened 2022-03-28
don't include CA in server cert

matthyx:
Related to https://github.com/openshift/service-ca-operator/issues/188


matthyx:
@smarterclayton can you have a look?

matthyx:
/lifecycle frozen
@sttts can you have a look pls?

PR #1337, opened 2022-03-28
WIP: add container crashing detector

mfojtik:


dgrisonnet:
Continuing our discussion here. You mentioned that we couldn't use the existing [KubePodCrashLooping](https://github.com/openshift/cluster-monitoring-operator/blob/a582608a29511aabc2267889c8f4a2e497e44bc7/assets/control-plane/prometheus-rule.yaml#L15-L24) alert because the `kube_pod_container_status_waiting_reason` metric was incorrect and we would need to use the terminated status instead. Since this information is also present as a metric, is there anything that prevents us from fixing the alert instead of writing a whole new controller?

PR #1340, opened 2022-03-28
Bug 2063953: extend rest mapping with job definition

tkashem:


tkashem:
once this merges, open a PR in o/k 4.8 ref: https://github.com/openshift/kubernetes/pull/1233

PR #1342, opened 2022-03-29
pkg/mainifest: change includeTechPreview to pointer

jottofar:
to allow easy no-o of teach-preview check.

### DIFF
--- a/pkg/manifest/manifest_test.go
+++ b/pkg/manifest/manifest_test.go
@@ -711,7 +711,7 @@ func Test_include(t *testing.T) {
 					},
 				},
 			}
-			err := m.Include(&tt.exclude, tt.includeTechPreview, &tt.profile, &tt.caps)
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest_test.go
```diff
@@ -711,7 +711,7 @@ func Test_include(t *testing.T) {
 					},
 				},
 			}
-			err := m.Include(&tt.exclude, tt.includeTechPreview, &tt.profile, &tt.caps)
```
LalatenduMohanty, line 714: Lets add an unit test case  when tt.includeTechPreview bool is nil. It is not a super important one but IMO we should add one to cover the use case.

LalatenduMohanty (changes requested):

LalatenduMohanty (approved):
/lgtm

pkg/manifest/manifest_test.go
LalatenduMohanty: Never mind. I have added lgtm

PR #1343, opened 2022-03-30
Adding unit tests to cover nil pointer cases of Include()

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

### DIFF
--- a/pkg/manifest/manifest_test.go
+++ b/pkg/manifest/manifest_test.go
@@ -587,39 +587,62 @@ func setupTestFS(t *testing.T, d dir) (string, func() error) {
 }
 
 func Test_include(t *testing.T) {
+	identifier := "identifier"
+	defaultClusterProfile := "self-managed-high-availability"
+	singleNodeProfile := "single-node"
+	trueBol := true
+	falseBol := false
+
 	tests := []struct {
 		name               string
-		exclude            string
-		includeTechPreview bool
-		profile            string
+		exclude            *string
+		includeTechPreview *bool
+		profile            *string
 		annotations        map[string]interface{}
-		caps               configv1.ClusterVersionCapabilitiesStatus
+		caps               *configv1.ClusterVersionCapabilitiesStatus
 
 		expected error
 	}{
 		{
 			name:    "exclusion identifier set",
-			exclude: "identifier",
-			profile: DefaultClusterProfile,
+			exclude: &identifier,
+			profile: &defaultClusterProfile,
 			annotations: map[string]interface{}{
 				"exclude.release.openshift.io/identifier":                     "true",
 				"include.release.openshift.io/self-managed-high-availability": "true"},
 			expected: fmt.Errorf("exclude.release.openshift.io/identifier=true"),
 		},
 		{
+			name:    "exclusion identifier set with no capability",
+			exclude: &identifier,
+			profile: &defaultClusterProfile,
+			annotations: map[string]interface{}{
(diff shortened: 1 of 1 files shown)


pkg/manifest/manifest_test.go
```diff
@@ -587,39 +587,62 @@ func setupTestFS(t *testing.T, d dir) (string, func() error) {
...
 		},
 		{
 			name:        "profile selection works",
-			profile:     "single-node",
+			profile:     &singleNodeProfile,
 			annotations: map[string]interface{}{"include.release.openshift.io/self-managed-high-availability": "true"},
 			expected:    fmt.Errorf("include.release.openshift.io/single-node unset"),
 		},
+		{
+			name:        "No profile",
```
jottofar, line 632: Change this to "all nil", which I think it is, and cover all nil's with one case. Or add a case for each one set to nil.

pkg/manifest/manifest_test.go
```diff
@@ -587,39 +587,62 @@ func setupTestFS(t *testing.T, d dir) (string, func() error) {
...
-		profile            string
+		exclude            *string
+		includeTechPreview *bool
+		profile            *string
 		annotations        map[string]interface{}
-		caps               configv1.ClusterVersionCapabilitiesStatus
+		caps               *configv1.ClusterVersionCapabilitiesStatus
 
 		expected error
 	}{
```
jottofar, line 605: add "no annotations" test

pkg/manifest/manifest_test.go
```diff
@@ -656,42 +680,42 @@ func Test_include(t *testing.T) {
...
 			annotations: map[string]interface{}{},
 			expected:    fmt.Errorf("include.release.openshift.io/self-managed-high-availability unset"),
 		},
 		{
 			name:        "default profile selection excludes with no annotation",
-			profile:     DefaultClusterProfile,
+			profile:     &defaultClusterProfile,
 			annotations: nil,
 			expected:    fmt.Errorf("no annotations"),
 		},
```
jottofar, line 692: add "no capability anno" test

pkg/manifest/manifest_test.go
LalatenduMohanty: I will add an all nil use case which will be equivalent to no annotation test case. The include method reacts differently when each of the input is nil so I have handled each use case differently.

pkg/manifest/manifest_test.go
LalatenduMohanty: Now that capability is pointer i.e. `caps    *configv1.ClusterVersionCapabilitiesStatus` , when ever we not passing any capability it is passed as `nil`.

pkg/manifest/manifest_test.go
LalatenduMohanty: Fixed https://github.com/openshift/library-go/commit/7ae641e6db34e5fad1456c5954b03f2b51292cae

pkg/manifest/manifest_test.go
LalatenduMohanty: Fixed in https://github.com/openshift/library-go/commit/7ae641e6db34e5fad1456c5954b03f2b51292cae

pkg/manifest/manifest_test.go
jottofar: Right, but I was referring to something more explicit but it's not needed.

PR #1345, opened 2022-04-01
add storage class controller

RomanBednar:
Adding a storage class controller. This controller should be used by CSI driver operators and as a result provide users the possibility of changing their default storage class.

In short the controller can take a storage class asset file which has default StorageClass annotation set to true - this indicates we want to create this specific storage class as default, but only if:

1) there are no other storage classes defined as default
2) there is no storage class with matching name (if there is just copy its annotation)

All other cases should result in applying non-default storage class regardless of the annotation set in asset.

Here is an example of the storage class asset I used for testing:

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: gp2-csi
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
parameters:
  type: gp2
  encrypted: "true"
provisioner: ebs.csi.aws.com
reclaimPolicy: "Delete"
volumeBindingMode: WaitForFirstConsumer
allowVolumeExpansion: true
```

And here is a couple of tests I tried manually:

 1) Setup: no prior default SC
    Action: CSI operator SC is created during sync
    Expected result: operator SC is applied as default

```
    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:false]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   <none>

    $ oc delete sc/gp2-csi
    storageclass.storage.k8s.io "gp2-csi" deleted

    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:false]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:true]
    gp3-csi   <none>
```

2)  Setup: prior default SC exists
    Action: CSI operator SC is created during sync
    Expected result: operator SC is applied as non default

```
    $ oc patch sc gp2 -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
    storageclass.storage.k8s.io/gp2 patched

    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   <none>

    $ oc delete sc/gp2-csi
    storageclass.storage.k8s.io "gp2-csi" deleted

    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   <none>
```

3)  Setup: prior non default SC exists and matches csi operator SC name
    Action: operator SC is reapplied during sync
    Expected result: reapplying does not override default SC annotation to "true"

```
    $ oc patch sc gp2-csi -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
storageclass.storage.k8s.io/gp2-csi patched

    # Wait for sync and check the gp2-csi is still non-default
    $ sleep 70 && oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:false]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   <none>
```

4)  Setup: prior default SC exists and matches csi operator SC name
    Action: operator SC is reapplied during sync
    Expected result: reapplying does not override existing SC annotation back to false - two SCs are kept as default
     
     *This is not a supported configuration but we have to allow this in order to keep behavior consistent with other resource controllers.*
    
```
    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   <none>
    
    $ oc patch sc gp2-csi -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
    storageclass.storage.k8s.io/gp2-csi patched
    
    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:true]
    gp3-csi   <none>
```

5)  Setup: prior default SC exists and matches csi operator SC name and has empty value
    Action: operator SC is reapplied during sync
    Expected result: reapplying does not override existing SC annotation
    
```
    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:false]
    gp3-csi   map[storageclass.kubernetes.io/is-default-class:false]
    
    $ oc patch sc gp2-csi -p '{"metadata":{"annotations":{"storageclass.kubernetes.io/is-default-class":""}}}'
    storageclass.storage.k8s.io/gp2-csi patched
    
    $ oc get sc -o "custom-columns=name:metadata.name,default:metadata.annotations"
    name      default
    gp2       map[storageclass.kubernetes.io/is-default-class:true]
    gp2-csi   map[storageclass.kubernetes.io/is-default-class:]
    gp3-csi   map[storageclass.kubernetes.io/is-default-class:false]

```

### DIFF
--- a/pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go
+++ b/pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go
@@ -1,0 +1,146 @@
+package csistorageclasscontroller
+
+import (
+	"context"
+	operatorapi "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/storage/v1"
+	"k8s.io/klog/v2"
+	"time"
+)
+
+const (
+	defaultScAnnotationKey = "storageclass.kubernetes.io/is-default-class"
+)
+
+// This Controller deploys a StorageClass provided by CSI driver operator
+// and decides if this StorageClass should be applied as default - if requested.
+// If operator wants to request it's StorageClass to be created as default,
+// the asset file provided to this controller must have defaultScAnnotationKey set to "true".
+// Based on the current StorageClasses in the cluster the controller can decide not
+// to deploy given StorageClass as default if there already is any existing default.
+// When the controller detects there already is a StorageClass with a same name it
+// just copies the default StorageClass annotation from the existing one to prevent
+// overwriting value that user might have manually changed.
+// If the asset file does not have defaultScAnnotationKey set at all, controller
+// just skips any checks and modifications and applies the StorageClass as it is.
+// It produces following Conditions:
+// StorageClassControllerDegraded - failed to apply StorageClass provided
+type CSIStorageClassController struct {
+	name               string
+	assetFunc          resourceapply.AssetFunc
+	file               string
+	kubeClient         kubernetes.Interface
+	storageClassLister v1.StorageClassLister
+	operatorClient     v1helpers.OperatorClient
+	eventRecorder      events.Recorder
+}
+
+func NewCSIStorageClassController(
+	name string,
+	assetFunc resourceapply.AssetFunc,
+	file string,
+	kubeClient kubernetes.Interface,
+	informerFactory informers.SharedInformerFactory,
+	operatorClient v1helpers.OperatorClient,
+	eventRecorder events.Recorder) factory.Controller {
+	c := &CSIStorageClassController{
+		name:               name,
+		assetFunc:          assetFunc,
+		file:               file,
+		kubeClient:         kubeClient,
+		storageClassLister: informerFactory.Storage().V1().StorageClasses().Lister(),
+		operatorClient:     operatorClient,
+		eventRecorder:      eventRecorder,
+	}
+
+	return factory.New().WithSync(
+		c.Sync,
+	).ResyncEvery(
+		time.Minute,
+	).WithSyncDegradedOnError(
+		operatorClient,
+	).WithInformers(
+		operatorClient.Informer(),
+		informerFactory.Storage().V1().StorageClasses().Informer(),
+	).ToController(
+		"StorageClassController",
+		eventRecorder,
+	)
+}
+
+func (c *CSIStorageClassController) Sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	klog.V(4).Infof("StorageClassController sync started")
+	defer klog.V(4).Infof("StorageClassController sync finished")
+
+	opSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return err
+	}
+	if opSpec.ManagementState != operatorapi.Managed {
+		return nil
+	}
+
+	syncErr := c.syncStorageClass(ctx)
+
+	return syncErr
+}
+
+func (c *CSIStorageClassController) syncStorageClass(ctx context.Context) error {
+	expectedScBytes, err := c.assetFunc(c.file)
+	if err != nil {
+		return err
+	}
+
+	expectedSC := resourceread.ReadStorageClassV1OrDie(expectedScBytes)
+
+	existingSCs, err := c.storageClassLister.List(labels.Everything())
+	if err != nil {
+		klog.V(2).Infof("could not list StorageClass objects")
+		return err
+	}
+
+	defaultSCCount := 0
+	annotationKeyPresent := false
+	// Skip the default SC annotation check if it's not in the input/expectedSC.
+	if expectedSC.Annotations != nil && expectedSC.Annotations[defaultScAnnotationKey] != "" {
+		for _, sc := range existingSCs {
+			if sc.Annotations[defaultScAnnotationKey] == "true" && sc.Name != expectedSC.Name {
+				defaultSCCount++
+			}
+			if sc.Name == expectedSC.Name && sc.Annotations != nil {
+				// There already is an SC with same name we intend to apply, copy its annotation.
+				if val, ok := sc.Annotations[defaultScAnnotationKey]; ok {
+					expectedSC.Annotations[defaultScAnnotationKey] = val
+					annotationKeyPresent = true // If there is a key, we need to preserve it, whatever the value is.
+				} else {
+					annotationKeyPresent = false
+				}
+			}
+		}
+		// There already is a default, and it's not set on the SC we intend to apply. Also, if there is any value for
+		// defaultScAnnotationKey do not overwrite it (empty string is a correct value).
+		if defaultSCCount > 0 && !annotationKeyPresent {
+			expectedSC.Annotations[defaultScAnnotationKey] = "false"
+		}
+	}
+
+	_, _, err = resourceapply.ApplyStorageClass(ctx, c.kubeClient.StorageV1(), c.eventRecorder, expectedSC)
+
+	return err
+}
+
+// UpdateConditionFunc returns a func to update a condition.
+func removeConditionFn(condType string) v1helpers.UpdateStatusFunc {
+	return func(oldStatus *operatorapi.OperatorStatus) error {
+		v1helpers.RemoveOperatorCondition(&oldStatus.Conditions, condType)
+		return nil
+	}
+}
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/csi/csiconfigobservercontroller"
 	"github.com/openshift/library-go/pkg/operator/csi/csidrivercontrollerservicecontroller"
 	"github.com/openshift/library-go/pkg/operator/csi/csidrivernodeservicecontroller"
+	"github.com/openshift/library-go/pkg/operator/csi/csistorageclasscontroller"
 	"github.com/openshift/library-go/pkg/operator/deploymentcontroller"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
@@ -38,6 +39,7 @@ type CSIControllerSet struct {
 	csiDriverControllerServiceController factory.Controller
 	csiDriverNodeServiceController       factory.Controller
 	serviceMonitorController             factory.Controller
+	csiStorageclassController            factory.Controller
 
 	operatorClient v1helpers.OperatorClientWithFinalizers
 	eventRecorder  events.Recorder
@@ -54,6 +56,7 @@ func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
 		c.csiDriverControllerServiceController,
 		c.csiDriverNodeServiceController,
 		c.serviceMonitorController,
+		c.csiStorageclassController,
 	} {
 		if ctrl == nil {
 			continue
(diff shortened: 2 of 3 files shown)


pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,153 @@
...
+			v1helpers.UpdateConditionFn(degradedCnd),
+		)
+		return updateErr
+	}
+
+	if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient,
+		removeConditionFn(conditionsPrefix+operatorapi.OperatorStatusTypeDegraded),
+	); updateErr != nil {
+		return errutil.NewAggregate([]error{syncErr, updateErr})
+	}
```
jsafrane, line 109: This looks unnecessary, as `WithSyncDegradedOnError` will already add a very similar processing of errors returned from `Sync`. In `Sync`, just return `err` or `nil` and the DegradedOnError handled will add / remove degraded condition automatically.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	existingSCs, err := c.storageClassLister.List(labels.Everything())
+	if err != nil {
+		klog.V(2).Infof("could not list StorageClass objects")
+		return err
+	}
+
+	defaultScExists := false
+	defaultScCount := 0
+	for _, sc := range existingSCs {
+		if sc.Annotations[defaultScAnnotationKey] == "true" {
```
jsafrane, line 131: you need to check `sc.Name` to exclude the `expectedSC` itself from the count.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	expectedSC := resourceread.ReadStorageClassV1OrDie(expectedScBytes)
+
+	existingSCs, err := c.storageClassLister.List(labels.Everything())
+	if err != nil {
+		klog.V(2).Infof("could not list StorageClass objects")
+		return err
+	}
+
+	defaultScExists := false
+	defaultScCount := 0
```
jsafrane, line 129: `defaultScExists` seems to be useless, as you can always check `defaultScCount > 0`

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: I see, thanks. Removed.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: Fixed.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: Fixed.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, sc := range existingSCs {
+		if sc.Annotations[defaultScAnnotationKey] == "true" && sc.Name != expectedSC.Name {
+			defaultScCount++
+		}
+	}
+
+	if defaultScCount > 0 {
+		klog.V(2).Infof("default StorageClass already defined in cluster, %v StorageClass will be applied as non-default", expectedSC.Name)
+		expectedSC.Annotations["storageclass.kubernetes.io/is-default-class"] = "false"
+	}
```
jsafrane, line 118: I did not test it in a real cluster, what happens when user does not want any default storage class in the cluster? From the code above it seems the controller will mark its own SC as default. It should do so only when *creating* the SC, it should not overwrite the annotation when user set it manually to `false`.

RomanBednar: I just tested it, and yeah I did not realize we need to differentiate creation from re-apply. I've changed the code a bit to fix this and added a unit test for this case. Good catch.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	operatorClient     v1helpers.OperatorClient
+	eventRecorder      events.Recorder
+}
+
+func NewCSIDefaultStorageClassController(
+	name string,
+	assetFunc resourceapply.AssetFunc,
+	file string,
+	kubeClient kubernetes.Interface,
+	InformerFactory informers.SharedInformerFactory,
```
jsafrane, line 49: nit: lowercase `informerFactory`

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	expectedSC := resourceread.ReadStorageClassV1OrDie(expectedScBytes)
+
+	existingSCs, err := c.storageClassLister.List(labels.Everything())
+	if err != nil {
+		klog.V(2).Infof("could not list StorageClass objects")
+		return err
+	}
+
+	defaultScCount := 0
+	expectedScNonDefaultExists := false
```
jsafrane, line 109: nit: `defaultSCCount`, `expectedSCNonDefaultExists` (SC is an abbreviation -> all uppercase, unless it's the first word in an identifier)

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// This Controller deploys a StorageClass provided by CSI driver operator
+// and decides if this StorageClass should be applied as default - if requested.
+// If operator wants to request it's StorageClass to be created as default,
+// the asset file provided to this controller must have defaultScAnnotationKey set to "true".
+// Based on the current StorageClasses in the cluster the controller can decide not
+// to deploy given StorageClass as default if there is already an existing default.
+// Also, the controller will emit alerts if more than one StorageClass defined
+// as default exists - this is not a recommended setup.
```
jsafrane, line 31: Please remove this, it's for a subsequent PR.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller_test.go
```diff
@@ -0,0 +1,322 @@
...
+			for name, expectedSC := range expectedSCs {
+				actualSC, found := actualSCs[name]
+				if !found {
+					t.Errorf("Expected StorageClass not found: %s", name)
+					continue
+				}
+				if !equality.Semantic.DeepEqual(expectedSC, actualSC) {
+					t.Errorf("Unexpected StorageClass %+v content:\n%s", name, cmp.Diff(expectedSC, actualSC))
+				}
+				delete(expectedSCs, name)
```
jsafrane, line 310: Why do you delete the SC? It almost seems that you wanted to check if the list is empty after the loop.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller_test.go
```diff
@@ -0,0 +1,322 @@
...
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			// Initialize
+			ctx := newTestContext(test, t)
+
+			// Act
+			_ = ctx.controller.Sync(context.TODO(), factory.NewSyncContext(controllerName, events.NewInMemoryRecorder(operandName)))
```
jsafrane, line 287: please check for error

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,135 @@
...
+			defaultScCount++
+		}
+		if sc.Annotations[defaultScAnnotationKey] == "false" && sc.Name == expectedSC.Name {
+			expectedScNonDefaultExists = true
+		}
+	}
+
+	// Change expected SC to non-default if there already is a default or expected SC is already present as non-default.
+	if defaultScCount > 0 || expectedScNonDefaultExists {
+		expectedSC.Annotations["storageclass.kubernetes.io/is-default-class"] = "false"
```
jsafrane, line 121: `expectedSC.Annotations` can be nil

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
```diff
@@ -0,0 +1,135 @@
...
+		if sc.Annotations[defaultScAnnotationKey] == "true" && sc.Name != expectedSC.Name {
+			defaultScCount++
+		}
+		if sc.Annotations[defaultScAnnotationKey] == "false" && sc.Name == expectedSC.Name {
+			expectedScNonDefaultExists = true
+		}
+	}
+
+	// Change expected SC to non-default if there already is a default or expected SC is already present as non-default.
+	if defaultScCount > 0 || expectedScNonDefaultExists {
```
jsafrane, line 120: I still think it's important to distinguish if `expectedSC` already exists or not. If it exists and it's default, its annotation will be set to `false` when user creates another default SC. Which is quite surprising.

In other word, if `expectedSC` exists -> don't touch its annotations (i.e. copy `is-default-class` from the existing one). 
If `expectedSC` does not exist, adjust its default annotation as below and create it.

RomanBednar: Good point, changed the code a bit and added another example test in the description to demonstrate this.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: Added checks for it, also I think we can skip any checks or modifications of this annotation is the source/asset file does not have it.

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller_test.go
RomanBednar: done

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller_test.go
RomanBednar: not needed - removing

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: done

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: done

pkg/operator/csi/csidefaultstorageclasscontroller/csi_default_storageclass_controller.go
RomanBednar: done

RomanBednar:
Changing name from "default storage class controller" to just "storage class controller", using the word "default" here can be misleading and does not match the behavior of the controller well.

pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go
```diff
@@ -0,0 +1,148 @@
...
+	var existingSCAnnotation string
+
+	// Skip the default SC annotation check if it's not in the input/expectedSC.
+	if expectedSC.Annotations != nil && expectedSC.Annotations[defaultScAnnotationKey] != "" {
+		for _, sc := range existingSCs {
+			if sc.Annotations[defaultScAnnotationKey] == "true" && sc.Name != expectedSC.Name {
+				defaultSCCount++
+			}
+			if sc.Name == expectedSC.Name && sc.Annotations != nil {
+				existingSCAnnotation = sc.Annotations[defaultScAnnotationKey]
```
jsafrane, line 120: If the annotation exists and has empty value (`""`), `existingSCAnnotation` will be `""` and `expectedSC.Annotations[defaultScAnnotationKey]` won't be cleared at `if existingSCAnnotation` below.

RomanBednar: Right...let's add even more unit tests then. Posted a fix now.

PR #1346, opened 2022-04-06
pkg/manifest: Add methods to support capability implicit enablement

jottofar:
to check if manifest is enabled and to return manifest's capabilities.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -60,12 +61,20 @@ type Manifest struct {
 	Obj *unstructured.Unstructured
 }
 
+func (r resourceId) equal(id resourceId) bool {
+	return reflect.DeepEqual(r, id)
+}
+
 func (r resourceId) String() string {
 	if len(r.Namespace) == 0 {
 		return fmt.Sprintf("Group: %q Kind: %q Name: %q", r.Group, r.Kind, r.Name)
 	} else {
 		return fmt.Sprintf("Group: %q Kind: %q Namespace: %q Name: %q", r.Group, r.Kind, r.Namespace, r.Name)
 	}
+}
+
+func (m Manifest) SameResourceID(manifest Manifest) bool {
+	return m.id.equal(manifest.id)
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface for the Manifest
@@ -105,7 +114,7 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	return validateResourceId(m.id)
 }
 
-// Include returns an error if the manifest fails an inclusion filter and should not be excluded from futher
+// Include returns an error if the manifest fails an inclusion filter and should be excluded from further
 // processing by cluster version operator. Pointer arguments can be set nil to avoid excluding based on that
 // filter. For example, setting profile non-nil and capabilities nil will return an error if the manifest's
 // profile does not match, but will never return an error about capability issues.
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -60,6 +61,10 @@ type Manifest struct {
 	Obj *unstructured.Unstructured
 }
 
+func (r *resourceId) Equal(id *resourceId) bool {
```
wking, line 64: I have mixed feelings about a public method on a private type.  If this needs to be public, can we make `ResourceID` public?

pkg/manifest/manifest.go
```diff
@@ -151,17 +156,78 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
 	return nil
 }
 
+// ManifestEnabled returns whether the manifest is enabled given the defined cluster capabilities.
+func (m *Manifest) ManifestEnabled(capabilities configv1.ClusterVersionCapabilitiesStatus) bool {
```
LalatenduMohanty, line 160: s/ManifestEnabled/IsManifestEnabled/

pkg/manifest/manifest.go
```diff
@@ -151,17 +156,78 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+	return enabled
+}
+
+// GetManifestCapabilities returns the manifest's capabilities.
+func (m *Manifest) GetManifestCapabilities() []configv1.ClusterVersionCapability {
+	annotations := m.Obj.GetAnnotations()
+	if annotations == nil {
+		return nil
+	}
+	return getManifestCapabilities(annotations)
```
LalatenduMohanty, line 183: You can write 
```
if annonations != nil {
   return getManifestCapabilities(annotations)
}

File: pkg/manifest/manifest.go [hunk 4]
```diff
@@ -151,17 +156,78 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+		allCaps[i] = configv1.ClusterVersionCapability(c)
+	}
+	return allCaps
+}
+
+func getCapabilityProperties(capability configv1.ClusterVersionCapability,
+	capabilities configv1.ClusterVersionCapabilitiesStatus) (isKnown, isEnabled bool) {
+
+	isKnown = false
+	isEnabled = false
```
LalatenduMohanty, line 204: default value of a bool variable is "false" , I wonder if we really need to assign false to them

pkg/manifest/manifest.go
```diff
@@ -151,17 +156,78 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+	}
+	for _, enabledCapability := range capabilities.EnabledCapabilities {
+		if capability == enabledCapability {
+			isEnabled = true
+			break
+		}
+
+	}
+	return true, isEnabled
+}
```
LalatenduMohanty, line 222: I think we should break this function in to two different functions i.e. one for known capabilities and another for enabled capabilities.

LalatenduMohanty (changes requested):

pkg/manifest/manifest.go
LalatenduMohanty: That way it will be easier to maintain

pkg/manifest/manifest.go
```diff
@@ -151,17 +156,78 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
 	return nil
 }
 
+// ManifestEnabled returns whether the manifest is enabled given the defined cluster capabilities.
```
LalatenduMohanty, line 159: We should move `func checkResourceEnablement` just after `func (m *Manifest) Include` and the new methods can be inserted after that. As `checkResourceEnablement` only used in the include method. Just a nit pick, not something blocking this PR from merging

pkg/manifest/manifest.go
jottofar: I don't see an issue. The private type merely says nothing outside the package can or needs to reference the type. 

I do need to make `Manifest.id` public though.

pkg/manifest/manifest.go
jottofar: I can do that but normally lint would bitch about not having a return at the end of a function.

pkg/manifest/manifest.go
jottofar: You are correct. Also on reading up a bit on `named parameters` seems more appropriate to have one bare return at the end of the function for readability so I may do that as well.

pkg/manifest/manifest.go
jottofar: Yeah, I agree. But this method can just go away if I get rid of `ManifestEnabled`.

pkg/manifest/manifest.go
jottofar: method removed

LalatenduMohanty (approved):
/lgtm

PR #1348, opened 2022-04-07
Bug 2057495: Re-create StorageClass to change immutable parameters

jsafrane:
Some StorageClass fields are immutable after creation, therefore we need to delete and re-create the StorageClass if an operator needs to change these fields.

cc @openshift/storage 

This will help with Alibaba CSI driver, where we want to add `volumeSizeAutoAvailable: true` to the default storage class.


pkg/operator/resource/resourceapply/storage.go
```diff
@@ -51,12 +51,50 @@ func ApplyStorageClass(ctx context.Context, client storageclientv1.StorageClasse
...
+func storageClassNeedsRecreate(oldSC, newSC *storagev1.StorageClass) bool {
+	// Based on kubernetes/kubernetes/pkg/apis/storage/validation/validation.go,
+	// these fields are immutable.
+	if !equality.Semantic.DeepEqual(oldSC.Parameters, newSC.Parameters) {
+		return true
+	}
+	if !equality.Semantic.DeepEqual(oldSC.Provisioner, newSC.Provisioner) {
+		return true
+	}
+	if !equality.Semantic.DeepEqual(oldSC.ReclaimPolicy, newSC.ReclaimPolicy) {
```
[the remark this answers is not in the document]
jsafrane: Added direct compare, but with some `nil` checks (mostly for unit tests). Not sure it looks better :-).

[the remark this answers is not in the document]
jsafrane: And added unit test for each parameter.

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -51,12 +51,57 @@ func ApplyStorageClass(ctx context.Context, client storageclientv1.StorageClasse
...
+			return existing, false, err
+		}
+		actual, err := client.StorageClasses().Create(ctx, requiredCopy, metav1.CreateOptions{})
+		if err != nil && apierrors.IsAlreadyExists(err) {
+			// Delete() few lines above did not really delete the object,
+			// the API server is probably waiting for a finalizer removal or so.
+			// Report an error, but something else than "Already exists", because
+			// that would be very confusing - Apply failed because the object
+			// already exists???
+			err = fmt.Errorf("failed to re-create CSIDriver object %s, waiting for the original object to be deleted", existingCopy.Name)
```
bertinatto, line 68: s/CSIDriver/StorageClass/g :smile:

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -51,12 +51,57 @@ func ApplyStorageClass(ctx context.Context, client storageclientv1.StorageClasse
...
+		}
+		reportCreateEvent(recorder, actual, err)
+		return actual, true, err
+	}
```
bertinatto, line 72: nit: not required, but IMO it'd good to prepend something like "failed to re-created StorageClass:" to any errors returned in this block. This would add more context to the user and avoid any confusion in case something goes awry.

bertinatto (commented):
Just a comment and an optional suggestion, otherwise LGTM.

It's a shame we have to re-create instead of updating, but the StorageClass immutable fields don't leave us many choices.

/assign @deads2k 
for approval and validation

pkg/operator/resource/resourceapply/storage.go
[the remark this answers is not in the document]
bertinatto: Yeah, now that I see it I regret I suggested it :smile: but it's OK either way

pkg/operator/resource/resourceapply/storage.go
jsafrane: fixed

pkg/operator/resource/resourceapply/storage.go
jsafrane: Added the context as suggested. And I added it also to ApplyCSIDriver below, to keep both re-create blocks in sync.

PR #1350, opened 2022-04-11
adds a unit test for arePodsOnTheSameRevision method

p0lyn0mial:
This simple test proves that the method works against a single `pod`.
That means this method works on an SNO cluster.

The `arePodsOnTheSameRevision` method is used in `WaitForPodsToStabilizeOnTheSameRevision` which is exposed to various tests that need stable APIs.

### DIFF
--- a/test/library/pod_same_revision_test.go
+++ b/test/library/pod_same_revision_test.go
@@ -1,0 +1,61 @@
+package library
+
+import (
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
merged in https://github.com/openshift/library-go/pull/1362
/close

PR #1351, opened 2022-04-12
controller factory: allow enqueuing multiple objects in eventHandler

stlaz:
Allow enqueuing multiple or no keys while handling an object in an eventHandler. Useful when handling multiple types of objects in a single controller and object of one type needs to be expanded in multiple keys representing the object of the default handled type.

/cc @s-urbaniak 
/cc @sttts 
/assign @mfojtik

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -45,7 +45,7 @@ func (c syncContext) Recorder() events.Recorder {
 }
 
 // eventHandler provides default event handler that is added to an informers passed to controller factory.
-func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeyFunc, filter EventFilterFunc) cache.ResourceEventHandler {
+func (c syncContext) eventHandler(queueKeysFunc ObjectQueueKeysFunc, filter EventFilterFunc) cache.ResourceEventHandler {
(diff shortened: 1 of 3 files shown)


pkg/controller/factory/controller_context.go
```diff
@@ -45,35 +45,36 @@ func (c syncContext) Recorder() events.Recorder {
 }
 
 // eventHandler provides default event handler that is added to an informers passed to controller factory.
-func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeyFunc, filter EventFilterFunc) cache.ResourceEventHandler {
+func (c syncContext) eventHandler(queueKeyFunc ObjectQueueKeysFunc, filter EventFilterFunc) cache.ResourceEventHandler {
```
mfojtik, line 48: nit: s/queueKeyFunc/queueKeysFunc/

pkg/controller/factory/factory.go
```diff
@@ -18,6 +18,10 @@ import (
 // DefaultQueueKey is the queue key used for string trigger based controllers.
 const DefaultQueueKey = "key"
 
+func DefaultQueueKeyFunc(_ runtime.Object) []string {
```
mfojtik, line 21: nit: DefaultQueueKeysFunc

mfojtik: and godoc please

PR #1353, opened 2022-04-13
Update owners in subdirectories based on their past commit history

soltysh:
Generated from output from:
```
for i in ./pkg/*; do echo $i; git shortlog -s -n --no-merges $i; done
```
/assign @mfojtik @s-urbaniak @deads2k @tkashem


cgwalters:
Nice,
/approve
/lgtm

PR #1358, opened 2022-04-21
Bug 2071998: pkg/verify: Expose underlying signature errors

wking:
Sometimes we fail to verify because we find signatures we don't trust.  Sometimes we fail to find signatures at all, either because the backing stores have no signatures, or because we fail to contact the backing stores.  With this commit, we pass all that useful debugging context on to the caller.  Not in the string, because we don't want to overwhelm all callers, but they can get it via [`Unwrap`][1] so they can show the details where they have space to do so.

[1]: https://pkg.go.dev/errors#Unwrap

### DIFF
--- a/pkg/verify/verify.go
+++ b/pkg/verify/verify.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"golang.org/x/crypto/openpgp"
+	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/verify/store"
@@ -41,6 +42,19 @@ type Interface interface {
 
 	// AddStore adds additional stores for signature verification.
 	AddStore(additionalStore store.Store)
+}
+
+type wrapError struct {
+	msg string
+	err error
+}
+
+func (e *wrapError) Error() string {
+	return e.msg
+}
+
+func (e *wrapError) Unwrap() error {
+	return e.err
 }
 
 type rejectVerifier struct{}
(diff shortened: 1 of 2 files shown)


pkg/verify/verify.go
```diff
@@ -193,16 +211,21 @@ func (v *releaseVerifier) Verify(ctx context.Context, releaseDigest string) erro
...
-			for k := range remaining {
-				klog.Infof("Unable to verify %s against keyring %s", releaseDigest, k)
-			}
+		remainingKeyRings := make([]string, 0, len(remaining))
+		for k := range remaining {
+			remainingKeyRings = append(remainingKeyRings, k)
 		}
-		return fmt.Errorf("unable to locate a valid signature for one or more sources")
+		err := &wrapError{
+			msg: fmt.Sprintf("unable to verify %s against keyrings: %s", releaseDigest, strings.Join(remainingKeyRings, ", ")),
```
LalatenduMohanty, line 224: Looks like we are logging the keys too. I am wondering if it is fine from security point of view>

LalatenduMohanty (changes requested):

pkg/verify/verify.go
```diff
@@ -43,6 +44,19 @@ type Interface interface {
 	AddStore(additionalStore store.Store)
 }
 
+type wrapError struct {
+	msg string
+	err error
+}
+
+func (e *wrapError) Error() string {
```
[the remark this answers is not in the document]
wking: `wrapError` is [from Go][1], because I wanted `fmt.Sprintf("...%w")`'s `Unwrap` support, but without formatting the whole, possibly long aggregate error into the message.  We need to use `Error` and `Unwrap` to match the `error` interface.

[1]: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/fmt/errors.go;l=32-43;drc=refs%2Ftags%2Fgo1.18.1

pkg/verify/verify.go
wking: we're logging the keyring name, which is already public in [our GitHub source][1].  The only secret keys here are internal to Red Hat's build-time release signing.  By the time we get out to clusters verifying signatures, it's all public names and public keys.

[1]: https://github.com/openshift/cluster-update-keys/blob/master/manifests.rhel/0000_90_cluster-update-keys_configmap.yaml#L6

LalatenduMohanty (approved):
/lgtm

PR #1359, opened 2022-04-25
OpenStack uses External CCM by default

mdbooth:
This will need to be merged carefully in a coordinated manner.

Need pre-approval on each of the following before merging:
- [x] https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/186
- [ ] https://github.com/openshift/machine-config-operator/pull/3113
- [ ] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/617
- [ ] https://github.com/openshift/cluster-kube-apiserver-operator/pull/1342

These PRs *must* be merged in the following order:

1. CCCMO: nothing changes because CCCMO gates on a KCMO condition.
1. KCMO: KCM will be deployed without the cloud controllers.

   CCCMO will deploy OpenStack CCM, which will run the cloud controllers, including the node controller.
   
   All kubelets are still running with legacy cloud provider until MCO is upgraded. Any new kubelets are deployed with legacy cloud provider for the same reason. kubelet will not create Nodes with TaintExternalCloudProvider, so node controller sync loop safely ignores them.
   
   Kubelet with legacy cloud provider updates node addresses. CCM node-controller *also* updates node addresses. We have not hit a problem with this in practise during upgrade testing.
   
   New in 1.24, cloud-provider (and therefore OpenStack CCM) has the same filtering behaviour as kubelet when kubelet was passed a --node-ip argument. We always set this argument for OpenStack deployments. The implementation of OpenStack's NodeAddresses() is vastly different in downstream OpenShift to the implementation in OpenStack CCM. However, due to filtering we expect the resulting Node.Addresses to be the same. If the resulting Node.Addresses are not the same in some unforeseen configuration we can still be sure that, although Node.Addresses will flap, there will only be a single InternalIP, it will be listed first, and it will be the same for both controllers.
1. MCO: Updated machines will roll out kubelet with external cloud provider. Newly created kubelets will create tainted Nodes which will be initialised by the CCM. Existing kubelets will stop updating their own node addresses, leaving this to CCM, which was already doing it.
1. KAO: no ordering requirements.

/hold


mdbooth:
The unit test failure looks like an infra flake?

/test unit

elmiko (commented):
/lgtm

mdbooth:
We have now deferred the feature to 4.12, so this is obsolete.

PR #1362, opened 2022-05-02
updates signature of WaitForAPIServerToStabilizeOnTheSameRevision function

p0lyn0mial:
so that it:
 - accepts an object that implements `LoggingT` interface for logging purposes. That allows for using this method from ginkgo and the standard library's testing framework.
 - returns an error, that allows the caller of this method to make a decision on how to log the error or/and end the test.

### DIFF
--- a/test/library/library.go
+++ b/test/library/library.go
@@ -18,6 +18,10 @@ var (
 	WaitPollTimeout  = 10 * time.Minute
 )
 
+type LoggingT interface {
+	Logf(format string, args ...interface{})
+}
+
 // GenerateNameForTest generates a name of the form `prefix + test name + random string` that
 // can be used as a resource name. Convert the result to lowercase to use as a dns label.
 func GenerateNameForTest(t *testing.T, prefix string) string {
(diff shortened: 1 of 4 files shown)


test/library/pod_same_revision_test.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestArePodsOnTheSameRevision(t *testing.T) {
```
p0lyn0mial, line 12: plus a simple unit test that was missing.

test/library/library.go
```diff
@@ -18,6 +18,10 @@ var (
 	WaitPollTimeout  = 10 * time.Minute
 )
 
+type TestingT interface {
+	Logf(format string, args ...interface{})
```
s-urbaniak, line 22: nit: this interface doesn't implement all of testing.T, just logging, maybe naming could be differently.

p0lyn0mial: done.

PR #1363, opened 2022-05-02
Stop credential controller from syncing in manual mode

gnufied:
We need this to fix syncing of credentials in manual mode.

cc @openshift/storage


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -110,6 +111,7 @@ func (c *CSIControllerSet) WithCredentialsRequestController(
 	assetFunc resourceapply.AssetFunc,
 	file string,
 	dynamicClient dynamic.Interface,
+	typedVersionedOperatorClient operatorv1client.Interface,
```
gnufied, line 114: except efs operator, I did not find a single use of this function in any other openshift project and hence I figured breaking this function interface should be an acceptable tradeoff. cc @bertinatto

bertinatto: Yeah, I think this controller is not used anywhere else.

bertinatto (commented):
/lgtm

pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
```diff
@@ -69,9 +78,19 @@ func (c CredentialsRequestController) sync(ctx context.Context, syncContext fact
 	if apierrors.IsNotFound(err) {
 		return nil
 	}
+
 	if err != nil {
 		return err
 	}
+	clusterCloudCredential, err := c.typedVersionedOperatorClient.OperatorV1().CloudCredentials().Get(ctx, clusterCloudCredentialName, metav1.GetOptions{})
```
deads2k, line 85: this needs to be from a lister so that on a standard sync loop you do you not hit the kube-apiserver.

Once this is wired from a lister, that looks fine.

gnufied: okay... though we do not have a lister for `CredentialRequest` type - so it was hitting kube-apiserver anyways on every standard sync.I think I could fix this in a follow up or this PR.

gnufied:
@deads2k moved the code to use lister for `CloudCredential` objects. PTAL.

pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
```diff
@@ -41,6 +47,7 @@ func NewCredentialsRequestController(
 	manifest []byte,
 	dynamicClient dynamic.Interface,
 	operatorClient v1helpers.OperatorClientWithFinalizers,
+	operatorLister operatorv1lister.CloudCredentialLister,
```
deads2k, line 50: You'll want to wait for the informer to sync and possibly fire on state changes.

PR #1364, opened 2022-05-04
Worker Latency Profiles: Config Observe Fn and Latency Profile Controller

swghosh:
- Config observer for [Worker Latency Profiles](https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md) intended to update the different KCM/KApiS arguments on the basis of latency profile values found in [config.openshift.io/v1/node](https://github.com/openshift/api/blob/master/config/v1/types_node.go) resource.
- LatencyController to sync rollout status with the help of respective operator client that invokes the functionality

PS:
[1] https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r857345809
[2] https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r857347647
[3] https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611#pullrequestreview-954535665

Signed-off-by: Swarup Ghosh <swghosh@redhat.com>


pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+		}()
+
+		listers := genericListers.(NodeLister)
+		configNode, err := listers.NodeLister().Get("cluster")
+		// we got an error so without the node object we are not able to determine worker latency profile
+		if err != nil {
+			// if config/v1/node/cluster object is not found, that can be treated as a non-error case
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("nodes.config.openshift.io/cluster not found")
```
p0lyn0mial, line 58: since it can be treated as a `non-error` case it will spam the logs, I'd remove it.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+	// eg. observedConfig.extendedArguments.node-monitor-grace-period
+	// is []string{"extendedArguments", "node-monitor-grace-period"}
+	ArgConfigNestedPath []string
+	// value that is to be set for the given argument
+	ArgValue string
+}
+
+// CreateLatencyProfileObserverFn is used to create ObserveConfigFunc that can be used with an configobservation controller to trigger
+// changes to different arg val pairs in observedConfig.* fields and update them on the basis of current worker latency profile
+func CreateLatencyProfileObserverFn(latencyProfileValMap map[configv1.WorkerLatencyProfileType][]LatencyProfileArgVal, conditionFn func() (bool, error)) configobserver.ObserveConfigFunc {
```
p0lyn0mial, line 28: nit `NewLatencyProfileObserver`

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+			// log that latency profile is empty
+			klog.Infof("worker latency profile is not set in nodes.config.openshift.io/cluster")
+			return existingConfig, errs
+		}
+
+		// get the arg val pairs for currently set latency profile
+		profileArgValPairs, ok := latencyProfileValMap[configNode.Spec.WorkerLatencyProfile]
+		if !ok {
+			// log that this is an unsupported latency profile
+			klog.Warningf("unsupported worker latency profile found in nodes.config.openshift.io/cluster Spec.WorkerLatencyProfile = %v", configNode.Spec.WorkerLatencyProfile)
```
p0lyn0mial, line 90: I would document that the list of allowed values is defined on the API server and we can only accept new values during an upgrade. In that case, we tolerate skew until a new version of the operator will be installed.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	configMapClient         corev1client.ConfigMapsGetter
+	configNodeLister        listerv1.NodeLister
+	latencyProfileArgValMap map[apiconfigv1.WorkerLatencyProfileType][]nodeobserver.LatencyProfileArgVal
+}
+
+func NewLatencyProfileController(
+	operatorClient v1helpers.StaticPodOperatorClient,
+	targetNamespace string,
+	latencyProfileArgValMap map[apiconfigv1.WorkerLatencyProfileType][]nodeobserver.LatencyProfileArgVal,
+	configClient configv1.ConfigV1Interface,
```
p0lyn0mial, line 55: why do we need this client?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	nodeInformer configv1informers.NodeInformer,
+	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces,
+	kubeClient kubernetes.Interface,
```
p0lyn0mial, line 58: why do we need this client?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	eventRecorder events.Recorder,
+) factory.Controller {
+
+	ret := &LatencyProfileController{
+		operatorClient:          operatorClient,
+		targetNamespace:         targetNamespace,
+		latencyProfileArgValMap: latencyProfileArgValMap,
+		configClient:            configClient,
+		configMapClient:         v1helpers.CachedConfigMapGetter(kubeClient.CoreV1(), kubeInformersForNamespaces),
```
p0lyn0mial, line 67: won't `kubeInformersForNamespaces.ConfigMapLister().ConfigMaps(targetNamespace)` do?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (c *LatencyProfileController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+
+	// Collect the current latency profile
+	configNodeObj, err := c.configNodeLister.Get("cluster")
+
+	// if config/v1/node/cluster object is not found this controller should do nothing
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warningf("nodes.config.openshift.io/cluster not found")
```
p0lyn0mial, line 94: the node obj is always present, I think we could just `return nil`

pkg/operator/latencyprofilecontroller/controller_status.go
```diff
@@ -0,0 +1,51 @@
...
+const (
+	// set of reasons used for updating status
+	reasonLatencyProfileUpdated         = "ProfileUpdated"
+	reasonLatencyProfileUpdateTriggered = "ProfileUpdateTriggered"
+	reasonLatencyProfileEmpty           = "ProfileEmpty"
+
+	wlpPrefix = "WorkerLatencyProfile"
+)
+
+func (c *LatencyProfileController) updateStatus(ctx context.Context, reason, message string, progressing, complete bool) (bool, error) {
```
p0lyn0mial, line 19: I would move to `controller.go` file

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		return err
+	}
+
+	// in case of empty workerlatency profile, set status Complete=False and pre-empt sync
+	if configNodeObj.Spec.WorkerLatencyProfile == "" {
+		// TODO: in case worker latency profile is transitioned
+		// from Default/Medium/Low to ""
+		// check that the required args should vanish
+		_, err = c.updateStatus(ctx, reasonLatencyProfileEmpty, "latency profile not set on cluster", false, true)
```
p0lyn0mial, line 105: are we `completed` when the profile wasn't set?

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+
+	_, operatorStatus, _, err := c.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return err
+	}
+
+	// Collect the unique set of revisions for all nodes in cluster
+	revisionMap := map[int32]struct{}{}
+	uniqueRevisions := []int32{}
+	for _, nodeStatus := range operatorStatus.NodeStatuses {
```
p0lyn0mial, line 117: could we change the code to wait until all nodes are on the same revision and only then check the actual config?

we could set the status to progressing while waiting for all nodes to converge.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if revisionsHaveSynced {
+		_, err = c.updateStatus(ctx, reasonLatencyProfileUpdated, "all static pod revision(s) have updated latency profile", false, true)
+	} else {
+		_, err = c.updateStatus(ctx, reasonLatencyProfileUpdateTriggered, "one or more static pod revision(s) are updating latency profile", true, false)
+	}
+	return err
+}
+
+func configMatchProfileArguments(configMap *corev1.ConfigMap, desiredArgValPairs []nodeobserver.LatencyProfileArgVal) (bool, error) {
```
p0lyn0mial, line 160: I think this function should be provided by the operator.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+		defer func() {
+			// Prune the observed config so that it only contains fields specific to this observer
+			ret = configobserver.Pruned(ret, uniqueNestedPaths...)
+
+			// in case of existingConfig and currently observedConfig has mismatch in fields that are
+			// being monitored eg. apiServerArguments.default-not-ready-toleration-seconds,
+			// extendedArguments.node-monitor-grace-period, etc.
+			prevRet := configobserver.Pruned(existingConfig, uniqueNestedPaths...)
+			if !reflect.DeepEqual(prevRet, ret) {
+				eventRecorder.Eventf("LatencyProfileUpdated", "latency profile on cluster was updated, observed-config = %v", ret)
```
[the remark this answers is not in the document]
swghosh: I actually added it in favour of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611#discussion_r859554873.

Also, if you check the https://github.com/openshift/library-go/blob/5e6227194fbeb42769809e7026625dd1190d4ab4/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go#L69-L71 they're firing a similar event (similarly as an extra to the `ObservedConfigChanged` event as you mentioned).

It might be helpful as well for the users to know when the respective operator has enacted upon the newly applied latency profile.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Added it in favour of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611#discussion_r859543150.

Should it not be fine to see that in the logs upon every re-sync (though there would be too many repeated entries, I am not entirely sure how we could remove repetition)?

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Updated!

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Refactored, to use it that way.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Adding that client was erroneous/extra/left back when I copied over old code. 
Removed it now.

pkg/operator/latencyprofilecontroller/controller_status.go
swghosh: Moved the `updateStatus` func to controller.go.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+		// typically when WorkerLatencyProfileProgressing status (i.e. sprofile rollout is already in progressing state)
+		if !cond {
+			// log that latency profile couldn't be updated due to falsy condition
+			klog.Warningf("latency profile config observer did not update observed config due to unsatisified pre-requisite")
+			return existingConfig, errs
+		}
+
+		if configNode.Spec.WorkerLatencyProfile == "" {
+			// log that latency profile is empty
+			klog.Infof("worker latency profile is not set in nodes.config.openshift.io/cluster")
```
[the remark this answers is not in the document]
swghosh: Added it as it was similar to https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611#discussion_r859543150, but not absolutely necessary.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: We could always keep such a function in the operator but what I found was that it would lead to code repetition in cluster-kube-apiserver-operator and cluster-kube-controller-manager-operator codebase atleast. Not sure if wanna repeat the same logic in both of the operators.

Moreover, I see that this function is kinda generic as the intended behavior for this function is to compare content present in installercontroller generated config maps (storing configs about different active revisions) against known values.

pkg/operator/configobserver/node/observe_latency_profile.go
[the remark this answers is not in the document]
swghosh: Actually, we had specific reason in mind behind adding the `conditionFn` to config observer function.

We would want to gate the observer to not update the latency profile arg val pairs in case all the control plane nodes have not achieved the state desired by the current latency profile first. For eg. in case a user is changing the spec from Default to Medium latency profile and then again to the Low profile in a short duration, we would have to reject that update for sometime until the control plane nodes stabilize and we are on the same setting otherwise we could end up breaking the entire cluster. Please refer more about this in [cluster stability analysis which is explained in detail in our worker latency profile EP](https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md#cluster-stability-analysis). The same is applicable for both Kube API Server as well as Kube Controller Manager arguments (also for Kubelet), because a mismatch of profiles would essentially result in some nodes having args applied from one latency profile and other nodes having a different profile args which would hamper cluster stability to a great extent.

The plan here was to pass a `conditionFn` in the respective operator that would return if the current operator is in "WorkerLatencyProfileProgressing" state or not. If it's progressing, the `conditionFn` should be `false` and guard against any profile updates, whereas if it's not progressing (i.e. when completed rollout) the func should return `true` allowing the config observer to post new values.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+
+		cond, err := conditionFn()
+		if err != nil {
+			errs = append(errs, err)
+			return existingConfig, errs
+		}
+
+		// condition is not satisfied, so return existing config back without any changes
+		// eg. conditionFn is to check if rollouts are in progressing state
+		// typically when WorkerLatencyProfileProgressing status (i.e. sprofile rollout is already in progressing state)
```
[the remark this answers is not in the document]
swghosh: Contd. in thread: https://github.com/openshift/library-go/pull/1364#discussion_r866256288.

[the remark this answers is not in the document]
swghosh: We also might want to refer to the (x+1, ..., x+n) case referred in: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r854092530.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: I believe this is a continuation of: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1328#discussion_r855284876?

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Similar: https://github.com/openshift/library-go/pull/1364#discussion_r866145737

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,233 @@
...
+	).ResyncEvery(time.Minute).WithSync(ret.sync).WithSyncDegradedOnError(operatorClient).ToController(
+		"WorkerLatencyProfile",
+		eventRecorder.WithComponentSuffix("latency-profile-controller"),
+	)
+}
+
+func (c *LatencyProfileController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+
+	// Collect the current latency profile
+	configNodeObj, err := c.configNodeLister.Get("cluster")
```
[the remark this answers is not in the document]
swghosh: Sounds good to me. Lemme try this approach and code it up, to test it through.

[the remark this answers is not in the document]
p0lyn0mial: np, hth, just remember that the current state is operator specific, we could code it as a function provided by the operators.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,138 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+// LatencyProfileArgVal is used to set arg value pairs for each worker latency profile
+type LatencyProfileArgVal struct {
```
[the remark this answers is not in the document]
swghosh: Done.

pkg/operator/configobserver/node/observe_latency_profile.go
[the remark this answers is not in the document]
swghosh: Contd. discussion: https://github.com/openshift/library-go/pull/1364#discussion_r866696610, about gating the config observe func whether to rollout new observation values / not.

pkg/operator/latencyprofilecontroller/controller.go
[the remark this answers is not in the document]
swghosh: Yeah.

- Controller: So using `type MatchProfileRevisionConfigsFunc func(profile configv1.WorkerLatencyProfileType, revisions []int32) (match bool, err error)` function signature as controller init inputs such that operator specific profile matching logic across different latency profile revisions can be independent.

- Config Observer: using the same `conditionFn func() (bool, error)` which can hold operator specific logic

pkg/operator/latencyprofilecontroller/controller.go
swghosh: (please nvm, this is slightly contd. from https://github.com/openshift/library-go/pull/1364#discussion_r867704531)

Decoupled the logic for checking revision matching from the actual controller (as it should be operator specific), moved this functionality over to `pkg/operator/latencyprofilecontroller/helpers` by crafting a more generic implementation that returns a `MatchProfileRevisionConfigsFunc` which can be passed onto latencyprofilecontroller at init time.

Essentially, would like to invoke `helpers.NewInstallerRevisionConfigMatcher(...)` in the operator(s) (KAO, KCMO in our case), the output from which could be passed to our controller.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: As per our discussion, we can use current revisions to get the list of unique revisions that are active and we can pass the same to `controller.matchRevisionsFn()` which would take care of operator-specific logic for checking if revisions have synced / not & update status accordingly.

swghosh: More context, just in case:
https://github.com/openshift/library-go/pull/1364#discussion_r867704531, https://github.com/openshift/library-go/pull/1364#discussion_r867723045.

pkg/operator/configobserver/node/observe_latency_profile.go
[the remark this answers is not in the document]
atiratree: I would vote for simplifying it if we have one event already. And if the users want to react to these changes they can always watch the conditions which is better API wise I think.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+// LatencyConfigProfileTuple is used to set arg value pairs for each worker latency profile
+type LatencyConfigProfileTuple struct {
+	// path to required argument in observed config
+	// eg. extendedArguments.node-monitor-grace-period from observed config
+	// is []string{"extendedArguments", "node-monitor-grace-period"}
+	ConfigPath   []string
+	ProfileValue map[configv1.WorkerLatencyProfileType]string
```
atiratree, line 21: can we name this better please? It is not a single value

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// NewLatencyProfileObserver is used to create ObserveConfigFunc that can be used with an configobservation controller to trigger
+// changes to different arg val pairs in observedConfig.* fields and update them on the basis of current worker latency profile
+func NewLatencyProfileObserver(latencyConfigs []LatencyConfigProfileTuple, conditionFn func() (bool, error)) configobserver.ObserveConfigFunc {
```
atiratree, line 26: can we name and document the conditionFn better?

pkg/operator/latencyprofilecontroller/controller.go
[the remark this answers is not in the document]
atiratree: Shouldn't we at least try to persist the old value when gating the updates? I think we are causing following problems with such a simple gating.

let's start with following state

```
node1: low
node2: low
node3: low
```

update to medium

```
node1: low
node2: medium
node3: medium
```

update some other value

```
node1: low
node2: medium
node3: default
```

In this case we would get 3 different latency profiles and that is not desirable from what I understand. Also other combinations are possible.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	completedCondition := operatorv1.OperatorCondition{
+		Type:   wlpPrefix + "Complete",
+		Status: operatorv1.ConditionUnknown,
+	}
+
+	if progressing {
+		progressingCondition.Status = operatorv1.ConditionTrue
+		progressingCondition.Message = message
+		progressingCondition.Reason = reason
```
atiratree, line 155: this is hard to guess from the function header. Can we have `progressingReason` and `completedReason` in there? and pass for example also `"AsExpected"` accordingly.

atiratree: although I feel `"AsExpected"` could be named better for each state and tailored to each condition.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,174 @@
...
+	return err
+}
+
+func (c *LatencyProfileController) updateStatus(ctx context.Context, reason, message string, progressing, complete bool) (bool, error) {
+	progressingCondition := operatorv1.OperatorCondition{
+		Type:   wlpPrefix + "Progressing",
+		Status: operatorv1.ConditionUnknown,
+	}
+	completedCondition := operatorv1.OperatorCondition{
+		Type:   wlpPrefix + "Complete",
```
atiratree, line 148: Personally I would prefer having these both as constants. It is easier to search for these in code.

pkg/operator/latencyprofilecontroller/controller.go
atiratree: reminder about the TODO

```
// TODO: in case worker latency profile is transitioned
// from Default/Medium/Low to ""
```

pkg/operator/latencyprofilecontroller/controller.go
atiratree: Maybe we might want to update the conditions in case it is removed as well?

pkg/operator/latencyprofilecontroller/helpers/helpers.go
```diff
@@ -0,0 +1,102 @@
...
+				return false, nil
+			}
+		}
+		return true, nil
+	}
+}
+
+func configMatchProfileArguments(
+	configMap *corev1.ConfigMap,
+	latencyConfigs *[]nodeobserver.LatencyConfigProfileTuple,
```
atiratree, line 57: why do we need a pointer to a slice? Isn't slice enough?

swghosh: A slice would just be fine, I passed pointer to a slice to just save a bit of memory (makes sense? not also sure if it'd), as I guess we're mostly passing around the same contents if `latencyConfigs []nodeobserver.LatencyConfigProfileTuple` here and there, everywhere.

Also, you could assume that `latencyConfigs *[]nodeobserver.LatencyConfigProfileTuple` would be one const-like varibale per operator. So 1 for KCMO, 1 for KAO.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: What name do you suggest @atiratree? I'd go away and update the doc first.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Updated it to: `ConfigProfileValues map[configv1.WorkerLatencyProfileType]string` instead.

pkg/operator/configobserver/node/observe_latency_profile.go
atiratree: something like `ShouldSuppressConfigUpdatesFn()`? But I would rather revisit this if we decide there should be a different implementation after discussing https://github.com/openshift/library-go/pull/1364#discussion_r868325737

pkg/operator/latencyprofilecontroller/helpers/helpers.go
atiratree: Passing a slice is cheap as it is basically a pointer as well

pkg/operator/configobserver/node/observe_latency_profile.go
atiratree: to me ProfileConfigValues sounds better, although I guess there exist even a better name

swghosh: Let's keep it at `ProfileConfigValues` for now, I updated the struct!

pkg/operator/configobserver/node/observe_latency_profile.go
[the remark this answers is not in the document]
swghosh: Done, removed the extra event.

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,110 @@
...
+	observeLatencyProfileFn := func(
```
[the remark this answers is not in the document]
swghosh: Refactored the code accordingly to used struct and return named functions.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Updated it to: `ShouldSuppressConfigUpdatesFunc`.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Using constants now: "WorkerLatencyProfileProgressing", "WorkerLatencyProfilleComplete".

pkg/operator/latencyprofilecontroller/helpers/helpers.go
swghosh: Oh cool, I used slice directly now.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Updated (removed warningf).

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Updated (removed warningf).

pkg/operator/configobserver/node/observe_latency_profile.go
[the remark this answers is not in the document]
swghosh: Removed it.

pkg/operator/latencyprofilecontroller/controller.go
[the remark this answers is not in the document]
swghosh: The current implementation (as per https://github.com/openshift/library-go/pull/1364#discussion_r866696610) i.e. when the config observer is invoked as follows is able to tackle the race condition which we were initially facing.
```
nodeobserver.NewLatencyProfileObserver(
	latencyConfigs,
	nodeobserver.NewSuppressConfigUpdateUntilSameProfileFunc(
		operatorClient, configMapLister, latencyConfigs,
	),
),
```

Essentially, we would want to suppress further config updates (i.e. further updates to the reqd. arg vals) even if the user changed the latency profile on the cluster until the last latency profile is attained completely by all replicas of operand for current operator as this helps to ensure cluster stability.

@atiratree the observer when invoked in the said manner should not result in `node1: low, node2: medium, node: default` atleast. Also, in transition a mix of two latency profiles eg. `node1: low, node2: medium, node3: medium` is fine and our cluster would still remain stable until replicas are not in 2+ unique profiles. All we want is that our config observer should send new values only once it's confirmed that a single latency profile have been achieved which helps in stability.

Plus, regarding your suggestion on keeping track of the internal state and persist it somewhere before allowing jump to another profile is a reasonable ask, but me and @p0lyn0mial after significant brainstorming found it even harder to keep track of internal states and avoid race, until the fact that this idea just worked out:
- compare the current latency profile among replicas
- allow the observer to transition (desired -> current) to a new value only if the comparison checked out okay
It is simpler in implementation and does what we wanna do quite well, essentially instead of comparing internal state with desired state we're now just saying that our config observer will not update the values until of all of the replicas are on the same internal state (i.e. have same arg val pairs irrespective of which profile which values).

@p0lyn0mial I tried out the entire approach on a real cluster yesterday, seems like it works out quite well, I tried to check for different profile updates. Eg. I tried switching latency profile from `Default` -> `Medium` and then again to `Default` in a very short while, the args for Default were updated back again only after all the replicas achieved the revision which contained medium args. So the config observer got suppressed for the period until all replicas where able to achieve medium profile.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Updated function headers.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Documented in comments wherever relevant.
```
// Note: In case new latency profiles are added in the future in openshift/api
// this could break cluster upgrades although we're not passing an error here
// to ensure that ConfigObservationController doesn't reach degraded state.
```

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (s *revisionDiffProfileSuppressor) shouldSuppressConfigUpdates() (suppress bool, err error) {
+	_, operatorStatus, _, err := s.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return false, err
+	}
+
+	// Collect the unique set of revisions for all nodes in cluster
+	revisionMap := map[int32]struct{}{}
```
atiratree, line 167: why `struct{}`? , `bool` is simpler to use. You can also use `sets` if you want.

swghosh: Changed it to simpler `bool`.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (c *LatencyProfileController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+
+	// Collect the current latency profile
+	configNodeObj, err := c.configNodeLister.Get("cluster")
+
+	// if config/v1/node/cluster object is not found this controller should do nothing
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
```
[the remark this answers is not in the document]
swghosh: Updated the conditions accordingly, thanks.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Also used similar `bool` in 
```
revisionMap := make(map[int32]bool)
```
in controller.go line 114.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Contd. in https://github.com/openshift/library-go/pull/1364/#discussion_r870361860.

pkg/operator/latencyprofilecontroller/controller.go
swghosh: It's taken care in the newest implementation anyway.

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,184 @@
...
+			ctx,
+			false, true, // not progressing, complete=True
+			"AsExpected", reasonLatencyProfileUpdated,
+			"all static pod revision(s) have updated latency profile",
+		)
+	} else {
+		_, err = c.updateStatus(
+			ctx,
+			true, false, // progressing=True, not complete
+			reasonLatencyProfileUpdateTriggered, "AsExpected",
```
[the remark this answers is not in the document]
swghosh: So I added another reason "ProfileRollingOut" does that seem apt? 
@atiratree you can PTAL once on all the 3 status update conditions.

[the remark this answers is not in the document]
atiratree: I am not sure now, it seems we have a too verbose terminology atm: 

```
	reasonLatencyProfileUpdated         = "ProfileUpdated"
	reasonLatencyProfileUpdateTriggered = "ProfileUpdateTriggered"
	reasonLatencyProfileEmpty           = "ProfileEmpty"
	reasonLatencyProfileSynced          = "ProfileSynced"
	reasonLatencyProfileRollingOut      = "ProfileRollingOut"
```

@p0lyn0mial what is your opinion on this?

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,205 @@
...
+	_, operatorStatus, _, err := s.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return false, err
+	}
+
+	// Collect the unique set of revisions for all nodes in cluster
+	uniqueRevisionMap := make(map[int32]bool)
+	for _, nodeStatus := range operatorStatus.NodeStatuses {
+		revision := nodeStatus.CurrentRevision
+		if _, ok := uniqueRevisionMap[revision]; !ok {
```
atiratree, line 170:
```suggestion
		uniqueRevisionMap[nodeStatus.CurrentRevision] = true
```

should simplify it

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,205 @@
...
+		configMapLister: configMapLister,
+		usedConfigPaths: usedConfigPaths,
+	}
+
+	// creates an actual conditional fn that can be invoked each time config observer
+	// intends to update observed config
+	return ret.shouldSuppressConfigUpdates
+}
+
+func (s *revisionDiffProfileSuppressor) shouldSuppressConfigUpdates() (suppress bool, err error) {
```
soltysh, line 160: I'm missing test cases ensuring this suppressor works as expected.

pkg/operator/latencyprofilecontroller/helpers/helpers.go
```diff
@@ -0,0 +1,115 @@
+package helpers
```
soltysh, line 1: Put these revision matcher files inside `pkg/operator/latencyprofilecontroller` no need to multiple packages. Call it `revisionconfigmatcher.go` explicitly.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Done :D

pkg/operator/latencyprofilecontroller/controller.go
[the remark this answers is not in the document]
atiratree: @swghosh I am sorry for ringing false alarms, I simply misunderstood how ConfigObserver controller works. Now that I have tested it properly, the feature and profile switching seems to work well. I was not able to achieve more than 2 simultaneous profiles.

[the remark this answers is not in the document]
swghosh: No problem @atiratree, thanks for discussing it through and glad that the functionality is well tested now to achieve desired outcome. :)

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		// prune the current config with only paths that is supposed to be monitored
+		// and keep comparing with the last config, in case of mis match, suppress
+		currentConfigPruned := configobserver.Pruned(currentConfig, s.usedConfigPaths...)
+		if lastConfig != nil && !reflect.DeepEqual(lastConfig, currentConfigPruned) {
+			// suppress=true: when config values don't match
+			return true, nil
+		}
+		lastConfig = currentConfigPruned
```
p0lyn0mial, line 198: mhm, I think we should compare against `operatorSpec.ObservedConfig.Raw` besides we should narrow the comparison down only to some specific arguments. Right now we are comparing the entire configuration. Does it make sense?

p0lyn0mial: In fact we should merge `operatorSpec.ObservedConfig.Raw` and `operatorSpec.UnsupportedConfigOverrides.Raw` and compare against it.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Added test cases for it.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: It could be an issue handling `UnsupportedConfigOverrides` in case someone sets the arg on the cluster for the ones which KCM/KApiS use for latency profiles.

> am okay to tell that the operator is trying forever as well... The cluster is in that case in unsupported mode anyway.

This might be helpful: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/611#discussion_r857868140

pkg/operator/latencyprofilecontroller/controller.go
```diff
@@ -0,0 +1,180 @@
...
+
+		// for configmaps of operator client target namespace
+		kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().ConfigMaps().Informer(),
+	).ResyncEvery(time.Minute).WithSync(ret.sync).WithSyncDegradedOnError(operatorClient).ToController(
+		"WorkerLatencyProfile",
+		eventRecorder.WithComponentSuffix("latency-profile-controller"),
+	)
+}
+
+func (c *LatencyProfileController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
p0lyn0mial, line 89: are we missing some unit tests for this controller?

pkg/operator/latencyprofilecontroller/controller.go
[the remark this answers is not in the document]
swghosh: As discussed with @p0lyn0mial we can use the same set of reasons for both complete and progressing.

pkg/operator/latencyprofilecontroller/controller.go
p0lyn0mial: follow-up is fine

pkg/operator/configobserver/node/observe_latency_profile.go
atiratree: +1 for merging with `UnsupportedConfigOverrides`

pkg/operator/latencyprofilecontroller/controller.go
swghosh: Thank you.

pkg/operator/configobserver/node/observe_latency_profile.go
swghosh: Yeah, we moving ahead with merging! (added the change)

pkg/operator/configobserver/node/observe_latency_profile.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (s *revisionDiffProfileSuppressor) shouldSuppressConfigUpdates() (suppress bool, err error) {
+	operatorSpec, operatorStatus, _, err := s.operatorClient.GetStaticPodOperatorState()
+	if err != nil {
+		return false, err
+	}
+
+	mergedConfigRaw, err := resourcemerge.MergeProcessConfig(
+		map[string]resourcemerge.MergeFunc{},
```
atiratree, line 167: you could pass a nil here instead

swghosh: Updated.

soltysh (approved):
/lgtm
/approve

PR #1367, opened 2022-05-10
Fix ApplyService to detect new Service spec

jsafrane:
When an operator needs to update `service.spec`, `ApplyService` should apply it. We detect change of the spec by using an annotation with the old `service.spec` required by the operator and comparing it with the current operator's request.

Keep the existing heuristics of user changes (`spec.type`, `spec.selector`) in place.

This code expects that the API server restores any immutable ``service.spec` fields, like `clusterIP`. My observation is that the API server restores them when processing the update.

This is related to https://bugzilla.redhat.com/show_bug.cgi?id=2076637 - we want to add new `service.spec.ports` to an existing service during cluster update.

cc @deads2k @openshift/storage

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -124,10 +124,17 @@ func ApplyNamespaceImproved(ctx context.Context, client coreclientv1.NamespacesG
 	return actual, true, err
 }
 
-// ApplyService merges objectmeta and requires
-// TODO, since this cannot determine whether changes are due to legitimate actors (api server) or illegitimate ones (users), we cannot update
+// ApplyService merges objectmeta and requires.
+// It detects changes in `required`, i.e. an operator needs .spec changes and overwrites existing .spec with those.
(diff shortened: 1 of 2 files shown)


jsafrane:
/retest

I don't see any failed unit test, just nozero exit code

```
INFO[2022-05-10T12:29:50Z] set -o pipefail; go test  -race -json ./pkg/... | gotest2junit > /logs/artifacts/junit_report.xml
SKIP: github.com/openshift/library-go/pkg/operator/loglevel TestClusterOperatorLoggingController
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/golang/test-unit.mk:12: test-unit] Error 1
{"component":"entrypoint","error":"wrapped process failed: exit status 2","file":"k8s.io/test-infra/prow/entrypoint/run.go:80","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"[20](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_library-go/1367/pull-ci-openshift-library-go-master-unit/1524002337901776896#1:build-log.txt%3A20)22-05-10T12:29:43Z"}
```

PR #1368, opened 2022-05-11
Bug 2084331: pkg/manifest/manifest.go: Fix checkResourceEnablement

jottofar:
so a resource is filtered out if any of the capabilities are disabled.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -171,10 +171,10 @@ func checkResourceEnablement(annotations map[string]string, capabilities *config
 	unknownCaps := make([]string, 0, numCaps)
 	disabledCaps := make([]string, 0, numCaps)
 
-	var isKnownCap bool = false
-	var isEnabledCap bool = false
-
 	for _, c := range caps {
+		var isKnownCap bool
+		var isEnabledCap bool
+
 		for _, knownCapability := range capabilities.KnownCapabilities {
 			if c == knownCapability {
 				isKnownCap = true
--- a/pkg/manifest/manifest_test.go
+++ b/pkg/manifest/manifest_test.go
@@ -691,16 +691,14 @@ func Test_include(t *testing.T) {
 			expected:    fmt.Errorf("no annotations"),
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/manifest/manifest_test.go
```diff
@@ -720,6 +720,41 @@ func Test_include(t *testing.T) {
 				EnabledCapabilities: []configv1.ClusterVersionCapability{"cap1"},
 			},
 		},
+		{
+			name:    "multiple capabilities, 1 unknown, works",
```
wking, line 724: nit: Why "works"?  Although I guess you have `enabled capability works` in the previous entry, so maybe the pattern is just to end all test-case names with "works".

pkg/manifest/manifest_test.go
```diff
@@ -720,6 +720,41 @@ func Test_include(t *testing.T) {
...
+			profile: &defaultClusterProfile,
```
wking, line 725: nit: since #1343's pivot to pointer, we can leave unset all the properties that aren't the one whose filter we're exercising.

pkg/manifest/manifest.go
```diff
@@ -171,10 +171,13 @@ func checkResourceEnablement(annotations map[string]string, capabilities *config
 	unknownCaps := make([]string, 0, numCaps)
 	disabledCaps := make([]string, 0, numCaps)
 
-	var isKnownCap bool = false
-	var isEnabledCap bool = false
+	var isKnownCap bool
+	var isEnabledCap bool
```
wking, line 175: nit: `is*Cap` are only used within the loop, so you can shift the whole `var ...` or `isKnownCap := false` or whatever into the loop and drop the `var` declarations here.

wking (approved):
/bugzilla refresh
/lgtm
/hold

Left a few comments inline, but nothing that I think needs to block merging.  Ping me for a fresh LGTM if you decide to adjust anything, or pull the hold if you're fine as things stand.

LalatenduMohanty (approved):
/lgtm

PR #1371, opened 2022-05-19
Bug 2071998: pkg/verify: Expose store error details, especially slow access

wking:
Extending #1358 with more details and a test-case that confirms we will get timestamps and URIs when a HTTP request times out or is canceled.  Bunch of fiddly pivots here, so it's a number of small commits.  See the individual commit messages for a description of why I'm making the pivot that I'm making in that commit.

### DIFF
--- a/pkg/verify/store/configmap/configmap.go
+++ b/pkg/verify/store/configmap/configmap.go
@@ -122,6 +122,12 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
 				}
 			}
 		}
+		if done, err := fn(ctx, nil, fmt.Errorf("prefix %s in config map %s: %w", prefix, cm.ObjectMeta.Name, store.ErrNotFound)); err != nil || done {
+			return err
+		}
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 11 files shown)


pkg/verify/store/configmap/configmap.go
```diff
@@ -122,6 +122,12 @@ func (s *Store) Signatures(ctx context.Context, name string, digest string, fn s
 				}
 			}
 		}
+		if done, err := fn(ctx, nil, fmt.Errorf("prefix %s in config map %s: %w", prefix, cm.ObjectMeta.Name, store.ErrNotFound)); err != nil || done {
```
jottofar, line 125: This is confusing to me. It appears that all it's going to do is log the error `prefix...` but why? At this point it's not clear to me that there's been an error.

wking: b594c76c8 pivots `Signatures` from:

> Not finding any acceptable signatures is not an error; it is up to the caller to handle that case.

to:

> Not finding additional signatures should result in a callback call with an error wrapping ErrNotFound, to allow the caller to figure out when and why the store was unable to find a signature.  When a store has several lookup mechanisms, this may result in several callback calls with different ErrNotFound.  Signatures itself should return nil in this case, because eventually running out of signatures is an expected part of any invocation where the callback calls never return done=true.

So here I'm saying:

> I've exhausted this particular ConfigMap's signatures, and you still aren't happy.  `ErrNotFound` (wrapped up with some context about who I am), to let you know that I'm done.  Maybe I'll have another ConfigMap with more signatures in a bit.  Or maybe not.  But the `ErrNotFound` will make it easy for your to track/timestamp my progress through these ConfigMaps.

PR #1376, opened 2022-06-14
Migrate to go-ldap v3

yselkowitz:
The last go-ldap v2 release was in 2017.

Once merged, consuming components will need to update their own usage of go-ldap to v3 simultaneously when updating library-go.  Similarly, without this, such components are unable to update to go-ldap v3 themselves.

Keeping as a draft until 4.12 development opens.


yselkowitz:
This allows (but also forces) components to upgrade go-ldap from v2 (2017) to v3 (current).
/assign @deads2k @sttts

yselkowitz:
> This allows (but also forces) components to upgrade go-ldap from v2 (2017) to v3 (current). /assign @deads2k @sttts

Ping?

yselkowitz:
#1378 is a duplicate but has gained some traction where this has not; closing.

PR #1377, opened 2022-06-15
Bug 2096691: Add custom hooks to CSIStorageClassController

jsafrane:
An operator now has possibility to modify a StorageClass before it's applied by the SC controller.

Alibaba CSI driver operator will use it to dynamically inject `ResourceGroupID` to the default SC to provision PVs in the same resource group as where are all the other cluster resources like VMs and their root disks.


bertinatto (commented):
/lgtm

PR #1378, opened 2022-06-23
Bug 1879980: Bump go-ldap from 2.5.1 to 3.4.3

ardaguclu:
go-ldap 2.5.1 version is pretty old version and does not include the [fix](https://github.com/go-ldap/ldap/pull/307) for referenced bug. This PR bumps go-ldap module from 2.5.1 to 3.4.3 which includes fixes.


ardaguclu:
I'll push bump PR in 4.12 instead 4.11.
/hold
/cc @soltysh

ardaguclu:
/hold cancel
@ibihim @stlaz could you please take a look at this bump?. If we merge this bump, I'll re-vendor oc and bump this package also in there.

soltysh (approved):
In general this lgtm, but I'd like you to verify potential impact this might have on other consumers of this particular functionality (I only found oc and origin repos). Sync with @stlaz, I will want his review on this, and make sure oc is working as expected. The bump looks significant, although I'm fully supportive of updating our libraries. 

/lgtm
/approve
/hold

/assign @stlaz

stlaz:
@ardaguclu @yselkowitz Can you please summarize the main changes between v2 and v3 and point out the most important things that we need to be careful in regards of security?

Also, split the non-generated changes into a separate commit.

yselkowitz:
I don't think the change between v2 and v3 is actually that big.  The bottom line is that v2 is quite old and no longer maintained, so in terms of security, that itself should be a reason to migrate.

ardaguclu:
Only manual change did by me is this https://github.com/openshift/library-go/pull/1378/files#diff-da349abf5799c7e7dae830ae47b0630fbfcadb09eb4642bfba50d2b17d58d9deR103-R122. The rest is autogenerated and import path changes.

I agree with @yselkowitz. v2 is no longer maintained and it might be better to migrate to v3 in terms of security? What do you think @stlaz?

stlaz:
Keeping the hold until this is vendored to `oc` and `oauth-server`

stlaz:
/unhold
appears to be working in a happy path

PR #1379, opened 2022-06-27
fix: should judge whether updateErr is nil

dk-lockdown:
same as title.


kikisdeliveryservice (commented):
I'm not super familiar with this repo but seems like a reasonable fix.  In the worst case this would be using err from line 114. :grimacing:  

Similar issue on line 146 where it's referencing a generic err instead of createErr https://github.com/openshift/library-go/blob/84d884f4c9f6d5f3f38c4f4e6bf3305143f23c4b/pkg/operator/status/status_controller.go#L146

/ok-to-test

dk-lockdown:
If there is no status update permission, the update operation will always fail, but using err judgment will make the program continue to execute. I think  line 146 should use `createErr instead of `err`, otherwise, `syncCtx.Recorder()` will not record the correct error output

kikisdeliveryservice (commented):
/lgtm

soltysh (approved):
/approve

PR #1380, opened 2022-06-30
BZ_2065028: remove P=f on degrading operator

ibihim:
## What

An operator that is `progressing=true` for a long time should stay in progressing state (on top of getting `degraded=true`).

## Why

```
Progressing Timeline in Minutes
[--------|--------]
0   ^   15   ^  45  ^
     A         B        C

A is progressing=true, degraded=false
B is progressing=false, degraded=false
C is progressing=false, degraded=true (inertia logic set to 30 minutes for CAO)
```

## Ref

[Request Header IDP Flaky Test](https://bugzilla.redhat.com/show_bug.cgi?id=2065028)
[30 minutes](https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controllerset/apiservercontrollerset.go#L160-L165=)

Signed-off-by: Krzysztof Ostrowski <kostrows@redhat.com>


ibihim:
@stlaz, @deads2k PTAL

PR #1382, opened 2022-07-25
Enable external cloud provider for OpenStack

MaysaMacedo:
Adds support for OpenStack as an external cloud provider.


MaysaMacedo:
It will be handled by another PR

PR #1387, opened 2022-08-16
update manifest selection to allow Default and TechPreview for releas…

deads2k:
…e.openshift.io/feature-gate

This will allow us to have manifests that are used in TechPreview and alternative manifests for the same resource to be used for the Default featureset.

You can set .metadata.annotations["release.openshift.io/feature-set"]="Default,LatencySensitive,IPv6DualStackNoUpgrade" and in another manifest .metadata.annotations["release.openshift.io/feature-set"]="TechPreviewNoUpgrade,CustomNoUpgrade" to select which manifests should be active depending on featureset.

This does make it more expensive to add FeatureSets.  This is not a bad thing, FeatureSets should be expensive.

/assign @wking

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -9,6 +9,8 @@ import (
 	"reflect"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -21,7 +23,20 @@ import (
 const (
 	CapabilityAnnotation  = "capability.openshift.io/name"
 	DefaultClusterProfile = "self-managed-high-availability"
+	featureSetAnnotation  = "release.openshift.io/feature-set"
 )
+
+var knownFeatureSets = sets.String{}
+
+func init() {
+	for featureSet := range configv1.FeatureSets {
+		if len(featureSet) == 0 {
+			knownFeatureSets.Insert("Default")
+			continue
+		}
+		knownFeatureSets.Insert(string(featureSet))
+	}
+}
 
 // resourceId uniquely identifies a Kubernetes resource.
 // It is used to identify any duplicate resources within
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -114,11 +116,46 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	return validateResourceId(m.id)
 }
 
+const featureSetAnnotation = "release.openshift.io/feature-set"
```
wking, line 119: nit: can we shift this into the `const (` block at the top of the file?

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,46 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+
+var knownFeatureSets = sets.String{}
+
+func init() {
```
wking, line 123: Globals and the `init` package-singleton are also something I'd expect to find at the top of the file. But 🤷 search is cheap, so if feels better to you to keep them here, I don't mind.

pkg/manifest/manifest.go
```diff
@@ -132,15 +169,17 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+		}
+		manifestFeatureSets, manifestSpecifiesFeatureSets, err := getFeatureSets(annotations)
+		if err != nil {
+			return err
 		}
-		// never include the manifest if the feature-gate annotation is outside of allowed values (only TechPreviewNoUpgrade is currently allowed)
-		if featureGateAnnotationExists && featureGateAnnotationValue != string(configv1.TechPreviewNoUpgrade) {
-			return fmt.Errorf("unrecognized value %s=%s", featureGateAnnotation, featureGateAnnotationValue)
+		if manifestSpecifiesFeatureSets && !manifestFeatureSets.Has(requiredAnnotationValue) {
+			return fmt.Errorf("%q is required, and %s=%s", requiredAnnotationValue, featureSetAnnotation, strings.Join(manifestFeatureSets.List(), ","))
```
wking, line 182: nit: all of this could be pushed down into a `checkFeatureSets`, like:

```go
if  requiredFeatureSet != nil {
  err := checkFeatureSets(annotations, requiredFeatureSet)
  if err != nil {
    return err
  }
}
```

which is the pattern we use today for `checkResourceEnablement`.  To implement, you could shift some of this wrapping stuff over into your existing `getFeatureSets` and then rename and shuffle the function signature.

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,46 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+			continue
+		}
+		knownFeatureSets.Insert(string(featureSet))
+	}
+}
+
+func getFeatureSets(annotations map[string]string) (sets.String, bool, error) {
+	ret := sets.String{}
+	specified := false
+	for _, featureSetAnnotation := range []string{"release.openshift.io/feature-gate", featureSetAnnotation} {
```
wking, line 136: add a comment that we are deprecating `release.openshift.io/feature-gate` to be dropped as soon as possible, so I have something unambiguous to link when I try to push folks to modernize annotations? ;)

pkg/manifest/manifest_test.go
```diff
@@ -660,23 +676,62 @@ func Test_include(t *testing.T) {
...
 			name:               "incorrect techpreview value is not excluded if techpreview off",
-			includeTechPreview: &falseBol,
+			requiredFeatureSet: utilpointer.String(""),
 			profile:            &defaultClusterProfile,
 			annotations: map[string]interface{}{
 				"include.release.openshift.io/self-managed-high-availability": "true",
 				"release.openshift.io/feature-gate":                           "Other",
 			},
-			expected: fmt.Errorf("unrecognized value release.openshift.io/feature-gate=Other"),
+			expected: fmt.Errorf("unrecognized value \"Other\" in release.openshift.io/feature-gate=Other; known values are: CustomNoUpgrade,Default,IPv6DualStackNoUpgrade,LatencySensitive,TechPreviewNoUpgrade"),
```
wking, line 685: nit: you might want to pivot `expected` to a regular expression, like [this][1], so folks don't have to come in and bump this when they add or remove feature sets.  On the other hand, folks shouldn't be adding or removing feature sets all that often.

[1]: https://github.com/openshift/library-go/blob/b89722b6e9b5229a53af8906237a50f4b0cbb900/pkg/verify/store/serial/serial_test.go#L71

pkg/manifest/manifest_test.go
```diff
@@ -660,23 +676,62 @@ func Test_include(t *testing.T) {
...
+		},
+		{
+			name:               "correct techpreview value is excluded if techpreview off using feature-set",
```
wking, line 698: how does this differ from `correct techpreview value is excluded if techpreview off`?  Ah, that one uses the `release.openshift.io/feature-gate` annotation.  Can we add that qualifier to the existing test-case name `... using feature-gate`?

pkg/manifest/manifest_test.go
deads2k: > On the other hand, folks shouldn't be adding or removing feature sets all that often.

This

wking (approved):
/lgtm

PR #1388, opened 2022-08-16
Ldap: Respect ALL_PROXY when dialing

alvaroaleman:
This change adds downstream ldapDial and ldapDialTLS funcs that are
identical to ldap.Dial{,tls} except that they use a proxy dialer that
respects the ALL_PROXY environment variable to use a socks5 proxy from
there if set.

This is needed for Hypershift, where the oauth server runs in the
management cluster but the ldap server is only accessible from the guest
VPC. In that situation, the traffic from the oauth server needs to be
routed through a proxy into konnektivity which then routes it onto an
agent that runs on workers in the guest vpc and from there it can reach
its final destination.

/assign @stlaz 
Ref https://issues.redhat.com/browse/HOSTEDCP-421

### DIFF
--- a/pkg/security/ldapclient/client.go
+++ b/pkg/security/ldapclient/client.go
@@ -1,11 +1,13 @@
 package ldapclient
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"net"
 
 	"github.com/openshift/library-go/pkg/security/ldaputil"
+	"golang.org/x/net/proxy"
 	"k8s.io/client-go/util/cert"
 
 	"github.com/go-ldap/ldap/v3"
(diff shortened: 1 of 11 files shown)


stlaz:
/hold
this needs an enhancement

csrwng:
@stlaz @deads2k what is the enhancement exactly for? I don't think we're ready to add support for ALL_PROXY to the entire product, or even expose it in standalone OCP. This is for the narrow use case of oauth server running in HyperShift.

alvaroaleman:
> this needs an enhancement

Update to the enhancement is here: https://github.com/openshift/enhancements/pull/1223

pkg/security/ldapclient/client.go
```diff
@@ -57,6 +59,45 @@ type ldapClientConfig struct {
 // ldapClientConfig is an Config
 var _ Config = &ldapClientConfig{}
 
+var proxyDialer = proxy.FromEnvironment().(proxy.ContextDialer)
```
stlaz, line 62: no global vars, please

pkg/security/ldapclient/client.go
```diff
@@ -77,7 +118,7 @@ func (l *ldapClientConfig) Connect() (ldap.Client, error) {
 
 	switch l.scheme {
 	case ldaputil.SchemeLDAP:
-		con, err := ldap.Dial("tcp", l.host)
+		con, err := ldapDial("tcp", l.host)
```
stlaz, line 121: We should synchronize with @ardaguclu on his https://github.com/openshift/library-go/pull/1378 and use https://github.com/go-ldap/ldap/blob/master/conn.go#L120 instead of replacing the package's dial.

stlaz (changes requested):

pkg/security/ldapclient/client.go
alvaroaleman: Looked again and this isn't actually possible, because `DialWithDialer` takes  a concrete`*net.DIaler` and not an interface, so we can't pass the proxy dialer there. Also, it is only used in `DialURL`, which we don't use

pkg/security/ldapclient/client.go
```diff
@@ -57,6 +59,43 @@ type ldapClientConfig struct {
 // ldapClientConfig is an Config
 var _ Config = &ldapClientConfig{}
 
+// ldapDial is the same as ldap.Dial except that it uses a proxyDialer that respects
```
stlaz, line 62: Could we have also point out in the docs of both these functions that while the upstream allows supplying a custom dialer by `ldap.DialWithDialer`, this cannot currently be used because it specifically requires the `*net.Dialer` struct?

pkg/security/ldapclient/client.go
```diff
@@ -98,7 +137,7 @@ func (l *ldapClientConfig) Connect() (ldap.Client, error) {
 		return con, nil
 
 	case ldaputil.SchemeLDAPS:
-		return ldap.DialTLS("tcp", l.host, tlsConfig)
+		return ldapDialTLS("tcp", l.host, tlsConfig)
```
stlaz, line 140: I would only ask you to provide a `tcpdump` of the communication from the oauth-server pod to an encrypted LDAP server.

We need to be super-careful about our own supplied logic here. While we are testing LDAP IdP, I don't think we test it with a server that would refuse non-TLS connections.

Please see the following test, it might help you set up the IDP: https://github.com/openshift/origin/blob/804b1c878eb827068ba558d8940979e426298103/test/extended/oauth/ldap.go#L19-L20 (you might use this test to set up the server) -

stlaz (commented):
The changes are agreeable. Unfortunately Golang stdlib does not really think much about custom `Dialers` being supplied directly, while these are developed in the `x` libraries. See [tls.DialWithDialer](https://pkg.go.dev/crypto/tls#DialWithDialer) which I believe is the reason the go-ldap library `DialWithDialer` cannot expect anything better than `*net.Dialer`

pkg/security/ldapclient/client.go
alvaroaleman: https://gist.github.com/alvaroaleman/2e35dcc2640192b8d7e8daabfcfb49be

Top one is with the insecure option set, bottom one without. Both are captured during a login attempt using the IDP

soltysh (approved):
/approve

PR #1389, opened 2022-08-18
[wip] make the static resource controller handle featureset annotations

deads2k:
This allows static resources to be labelled the same way as CVO resources.

/hold

holding until I add tests

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -9,6 +9,8 @@ import (
 	"reflect"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


pkg/manifest/manifest.go
```diff
@@ -114,11 +116,62 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
 	return validateResourceId(m.id)
 }
 
+const featureSetAnnotation = "release.openshift.io/feature-set"
```
wking, line 119: nit: can we shift this into the `const (` block at the top of the file?

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,62 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+
+var knownFeatureSets = sets.String{}
+
+func init() {
```
wking, line 123: Globals and the `init` package-singleton are also something I'd expect to find at the top of the file.  But :shrug: search is cheap, so if feels better to you to keep them here, I don't mind.

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,62 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+	for featureSet := range configv1.FeatureSets {
+		if len(featureSet) == 0 {
+			knownFeatureSets.Insert("Default")
+			continue
+		}
+		knownFeatureSets.Insert(string(featureSet))
+	}
+}
+
+func GetFeatureSetsFromAnnotations(annotations map[string]string) (sets.String, bool, error) {
```
wking, line 133: This could be a method on `Manifest`.  And may not need to be public?

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,62 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+				}
+			}
+			ret.Insert(featureSetAnnotationValues...)
+		}
+	}
+
+	return ret, specified, nil
+}
+
+func ResourceSatisfiesFeatureSetRequirement(requiredFeatureSet string, annotations map[string]string) (bool, error) {
```
wking, line 154: This could be a method on Manifest. And may not need to be public?

pkg/manifest/manifest.go
```diff
@@ -132,15 +185,14 @@ func (m *Manifest) Include(excludeIdentifier *string, includeTechPreview *bool,
...
+		shouldInclude, err := ResourceSatisfiesFeatureSetRequirement(*requiredFeatureSet, annotations)
+		if err != nil {
+			return err
 		}
-		// never include the manifest if the feature-gate annotation is outside of allowed values (only TechPreviewNoUpgrade is currently allowed)
-		if featureGateAnnotationExists && featureGateAnnotationValue != string(configv1.TechPreviewNoUpgrade) {
-			return fmt.Errorf("unrecognized value %s=%s", featureGateAnnotation, featureGateAnnotationValue)
+		if !shouldInclude {
+			manifestFeatureSets, _, _ := GetFeatureSetsFromAnnotations(annotations) // best effort for message
+			return fmt.Errorf("%q is required, and %s=%s", *requiredFeatureSet, featureSetAnnotation, strings.Join(manifestFeatureSets.List(), ","))
```
wking, line 195: you shouldn't need to handle this error message in `Include`.  If `ResourceSatisfiesFeatureSetRequirement` (or [a successor method][1]) doesn't think the manifest is suitable, it can construct the error, and `Include` can blindly bubble that error up.  Basically, no need to distinguish `shouldInclude, err`; you can just return `err`, with all the different categories of "we weren't sure we could include this" lumped into `err` and passed up the chain.

[1]: https://github.com/openshift/library-go/pull/1389#discussion_r950599219

pkg/manifest/manifest.go
```diff
@@ -114,11 +116,62 @@ func (m *Manifest) UnmarshalJSON(in []byte) error {
...
+			continue
+		}
+		knownFeatureSets.Insert(string(featureSet))
+	}
+}
+
+func GetFeatureSetsFromAnnotations(annotations map[string]string) (sets.String, bool, error) {
+	ret := sets.String{}
+	specified := false
+	for _, featureSetAnnotation := range []string{"release.openshift.io/feature-gate", featureSetAnnotation} {
```
wking, line 136: add a comment that we are deprecating `release.openshift.io/feature-gate` to be dropped as soon as possible, so I have something unambiguous to link when I try to push folks to modernize annotations? ;)

PR #1392, opened 2022-08-26
Fix Services informer in StaticResourceController

jsafrane:
Use services informer instead of namespaces when the controller is about to apply a Service.


deads2k:
eek!

/lgtm

PR #1395, opened 2022-09-12
Installer static controller: be more forgiving and consider only recent installer pods

knelasevero:
In case all the latest revision installer pods are running and an older revision installer pod is in Pending state (e.g. the container runtime unable to delete the pod), do not change the InstallerPodContainerWaitingDegraded condition to True.

Based on https://github.com/openshift/library-go/pull/1230 (just rebased master into here - and changed how maps were being declared/assigned)

I think this is already an acceptable solution. Please let me know if a different approach would be better.

Bugs:

https://bugzilla.redhat.com/show_bug.cgi?id=2009646
https://bugzilla.redhat.com/show_bug.cgi?id=1986385

### DIFF
--- a/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
+++ b/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
@@ -3,6 +3,7 @@ package installerstate
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -59,6 +60,14 @@ var degradedConditionNames = []string{
 	"InstallerPodNetworkingDegraded",
 }
 
+func installerNameToRevision(name string) (int, error) {
+	parts := strings.Split(name, "-")
+	if len(parts) < 2 {
+		return 0, fmt.Errorf("Installer name %v is invalid, missing revision number", name)
+	}
+	return strconv.Atoi(parts[1])
+}
+
 func (c *InstallerStateController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	pods, err := c.podsGetter.Pods(c.targetNamespace).List(ctx, metav1.ListOptions{
 		LabelSelector: labels.SelectorFromSet(labels.Set{"app": "installer"}).String(),
@@ -67,19 +76,41 @@ func (c *InstallerStateController) sync(ctx context.Context, syncCtx factory.Syn
 		return err
 	}
 
+	masterRevisions := make(map[string][]*v1.Pod)
+	installerHighestRunningRevision := make(map[string]int)
+	for _, pod := range pods.Items {
+		p := pod
+		masterRevisions[pod.Spec.NodeName] = append(masterRevisions[pod.Spec.NodeName], &p)
+	}
+	// find the highest revision of a non-pending pod on each master node
+	for masterNode, pods := range masterRevisions {
+		maxRunningRev := 0
+		for _, pod := range pods {
+			if pod.Status.Phase != v1.PodPending || pod.Status.StartTime == nil {
+				rev, err := installerNameToRevision(pod.Name)
+				if err != nil {
+					return err
+				}
+				if rev > maxRunningRev {
+					maxRunningRev = rev
+				}
+			}
+		}
+		installerHighestRunningRevision[masterNode] = maxRunningRev
+	}
+
 	// collect all startingObjects that are in pending state for longer than maxToleratedPodPendingDuration
 	pendingPods := []*v1.Pod{}
 	for _, pod := range pods.Items {
 		if pod.Status.Phase != v1.PodPending || pod.Status.StartTime == nil {
 			continue
 		}
-		if c.timeNowFn().Sub(pod.Status.StartTime.Time) >= maxToleratedPodPendingDuration {
+		if rev, _ := installerNameToRevision(pod.Name); c.timeNowFn().Sub(pod.Status.StartTime.Time) >= maxToleratedPodPendingDuration && rev >= installerHighestRunningRevision[pod.Spec.NodeName] {
 			pendingPods = append(pendingPods, pod.DeepCopy())
 		}
 	}
 
-	// in theory, there should never be two installer startingObjects pending as we don't roll new installer pod
-	// until the previous/existing pod has finished its job.
+	// handle pending installer pods conditions
 	foundConditions := []operatorv1.OperatorCondition{}
 	foundConditions = append(foundConditions, c.handlePendingInstallerPods(syncCtx.Recorder(), pendingPods)...)
 
--- a/pkg/operator/staticpod/controller/installerstate/installer_state_controller_test.go
+++ b/pkg/operator/staticpod/controller/installerstate/installer_state_controller_test.go
@@ -147,6 +147,35 @@ func TestInstallerStateController(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "should report false when pending startingObjects are not the latest ones",
+			startingObjects: []runtime.Object{
+				newInstallerPod("installer-1", func(status *corev1.PodStatus) {
+					status.Phase = corev1.PodPending
+					status.Reason = "PendingReason"
+					status.Message = "PendingMessage"
+					status.StartTime = &metav1.Time{Time: time.Now().Add(-(maxToleratedPodPendingDuration + 5*time.Minute))}
+					status.ContainerStatuses = append(status.ContainerStatuses, corev1.ContainerStatus{Name: "test", State: corev1.ContainerState{Waiting: &corev1.ContainerStateWaiting{
+						Reason:  "PodInitializing",
+						Message: "initializing error",
+					}}})
+				}),
+				newInstallerPod("installer-2", func(status *corev1.PodStatus) {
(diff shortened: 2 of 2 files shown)


pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -67,13 +76,36 @@ func (c *InstallerStateController) sync(ctx context.Context, syncCtx factory.Syn
...
+	installerHighestRunningRevision := make(map[string]int)
+	for _, pod := range pods.Items {
+		p := pod
+		masterRevisions[pod.Spec.NodeName] = append(masterRevisions[pod.Spec.NodeName], &p)
+	}
+	// find the highest revision of a non-pending pod on each master node
+	for masterNode, pods := range masterRevisions {
+		maxRunningRev := 0
+		for _, pod := range pods {
+			if pod.Status.Phase != v1.PodPending || pod.Status.StartTime == nil {
```
ingvagabund, line 89: I wonder whether `pod.Status.Phase != v1.PodPending` makes `pod.Status.StartTime == nil` always true.

ingvagabund:
@knelasevero also worth extending tests in https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installerstate/installer_state_controller_test.go with few cases. At least those mentioned in the BZs.

pkg/operator/staticpod/controller/installerstate/installer_state_controller_test.go
```diff
@@ -147,6 +147,35 @@ func TestInstallerStateController(t *testing.T) {
...
+				}),
+			},
+			evalConditions: func(t *testing.T, conditions []operatorv1.OperatorCondition) {
+				podPendingCondition := v1helpers.FindOperatorCondition(conditions, "InstallerPodPendingDegraded")
+				if podPendingCondition.Status != operatorv1.ConditionFalse {
+					t.Errorf("expected InstallerPodPendingDegraded condition to be False")
+				}
+				podContainerWaitingCondition := v1helpers.FindOperatorCondition(conditions, "InstallerPodContainerWaitingDegraded")
+				if podContainerWaitingCondition.Status != operatorv1.ConditionFalse {
+					t.Errorf("expected InstallerPodPendingDegraded condition to be False")
```
ingvagabund, line 175: s/InstallerPodPendingDegraded/InstallerPodContainerWaitingDegraded

dgrisonnet:
I don't think this approach will resolve the problem in its entirety. What we are seeing in the bugs is that there can be multiple pending installer pods on a node at the same time, which is not something we would expect since theoretically a new installer pod should only be created when the previous one has been completed.
In practice, this is not the case as we can see in the bugs, so I think we should controllers logic to make sure that actions are taken to prevent having multiple pending installer pods on the same node.

In the installerstate controller there is a notion of pods being pending for too long, but we don't have it in the installer controller. 
https://github.com/openshift/library-go/blob/73d5a4a828656ce283fcc3097d6dafc73e13019c/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go#L25

From the installer controller point of view, as long as the pod is running and is not failing, it won't take any action towards it so we might want to consolidate the controller logic to take into account pods that are stuck in pending state.

ingvagabund:
> a new installer pod should only be created when the previous one has been completed.

Combination of container runtime being slow and change of the operator configuration can produce such a state. I have seen issues where installer-5 container was created but slow. Meantime installer-6 container was crated and finished. Writing revision-6 manifests on a disk. After which installer-5 container finished and rewritten the files on the disk to revision-5 version. Which we partially fixed. Though, depending on the container runtime it might take longer than anticipated to wait for the previous container to complete. I recall the decision was made to create a new installer rather than wait for completion to speed up the process and avoid getting stuck.

dgrisonnet:
My main concern with the bug reported was that installer pods can be stuck in pending state forever. I overlooked your comment in https://bugzilla.redhat.com/show_bug.cgi?id=2009646#c2, but it seems that this scenario is only possible when something wrong happens with kubelet. In the case of this BZ, kubelet tried to delete and recreate the pod because it couldn't start the sandbox, but it failed for unknown reasons. It would have been worth digging further into that, but the must-gather is not available anymore. In practice when that happens, the installer controller could theoretically recreate the installer pod which is the direction I wanted to take, but since this scenario happens because of a bug in the runtime, I don't think we should handle it at the controller level anymore.

Also, it is only problematic when this scenario occurs on the latest installer pod and a new revision isn't created after a while. But, this will be covered by this PR since it will put the controller in a degraded state and then it would be up to the user to either recreate the installer pod or a new revision.

tl;dr: I am fine with the approach from this PR

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -67,13 +76,36 @@ func (c *InstallerStateController) sync(ctx context.Context, syncCtx factory.Syn
...
+	}
+
 	// collect all startingObjects that are in pending state for longer than maxToleratedPodPendingDuration
 	pendingPods := []*v1.Pod{}
 	for _, pod := range pods.Items {
 		if pod.Status.Phase != v1.PodPending || pod.Status.StartTime == nil {
 			continue
 		}
-		if c.timeNowFn().Sub(pod.Status.StartTime.Time) >= maxToleratedPodPendingDuration {
+		if rev, _ := installerNameToRevision(pod.Name); c.timeNowFn().Sub(pod.Status.StartTime.Time) >= maxToleratedPodPendingDuration && rev >= installerHighestRunningRevision[pod.Spec.NodeName] {
```
dgrisonnet, line 108: The way you are building your list of pending pods is that you are taking all the pending pods that are more recent than the latest completed installer pod, but I don't think we need to report the status of all the pending pods, only the latest one should be enough since users will most likely only be interested about what happened to the latest revision.

For that purpose, it might be better to just build a map of the latest installer pod (no matter its state) per node and then pass it on to this loop that builds a list of pods that are in a pending state for longer than maxToleratedPodPendingDuration.

dgrisonnet (commented):
While you are at it, would you mind also updating this comment? https://github.com/openshift/library-go/blob/7772048e4447df0c11894beb2f3eca5878d1d1dd/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go#L81-L82

As per https://github.com/openshift/library-go/pull/1395#issuecomment-1256257601, this is not correct anymore and is quite confusing.

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
ingvagabund: > only the latest one for each node should be enough since users will most likely only be interested about what happened to the latest revision.

Good point. The latest installer pod invalidates all pending installer pods of a lower revision.

> For that purpose, it might be better to just build a map of the latest installer pod (no matter its state) per node and then pass it on to this loop that builds a list of pods that are in a pending state for longer than maxToleratedPodPendingDuration.

A completed pod is a base for deciding whether there are resources of a given revision successfully written to a master node. Existence of two or more pending pods with a higher revision than the latest completed revision might signify there's someone wrong happening on a master node. Question is whether this case is important for reporting. When the installer_state_controller reports Degraded state, there will always be at least one pending installer pod. Each pending pod is reported through a [separate condition](https://github.com/knelasevero/library-go/blob/702c72f4495d8e74dec40601b4520d27f53b7e6f/pkg/operator/staticpod/controller/installerstate/installer_state_controller.go#L197). Will reporting two or more pending pods for the same master node help in debugging? Will it provide some additional value to an admin? If not, let's simplify the logic and report only pending pods of the latest revision. I wonder whether the current/original implementation took this reasoning into consideration. When a must-gather is collected all the pending pods are collected as well. Though, all the conditions are logged by an operator. So having the additional pods reported in the conditions might give us more detailed historical background of what was happening to the installer pods after they eventually either completed or were deleted but no longer collected by must-gather.

dgrisonnet: > Though, all the conditions are logged by an operator. So having the additional pods reported in the conditions might give us a more detailed historical background of what was happening to the installer pods after they eventually either completed or were deleted but no longer collected by must-gather.

I can buy that argument, having historical background makes sense. It might not end up being useful for debugging, but at least it could for example help us track down when slowness first started to appear.

dgrisonnet:
/lgtm

@tkashem can you also have a look at this PR and tag it if you are fine with the change?

pkg/operator/staticpod/controller/installerstate/installer_state_controller.go
```diff
@@ -67,19 +76,41 @@ func (c *InstallerStateController) sync(ctx context.Context, syncCtx factory.Syn
...
 			pendingPods = append(pendingPods, pod.DeepCopy())
 		}
 	}
 
-	// in theory, there should never be two installer startingObjects pending as we don't roll new installer pod
-	// until the previous/existing pod has finished its job.
+	// handle pending installer pods conditions
```
knelasevero, line 113: @dgrisonnet I changed the comment that you mentioned before. Let me know if you think any other phrasing or removing it would be better

ingvagabund:
@knelasevero would you please squash the commits?

PR #1397, opened 2022-09-14
AUTH-296: crypto: add a CA{} method for creating subCAs, too

stlaz:
While working on microshift, I noticed that this method was missing.

/assign @deads2k 
/cc @ibihim

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -693,6 +693,43 @@ func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA) 
 	return signerConfig, nil
 }
 
+func (ca *CA) EnsureSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, bool, error) {
+	if subCA, err := GetCA(certFile, keyFile, serialFile); err == nil {
+		return subCA, false, err
+	}
+	subCA, err := ca.MakeAndWriteSubCA(certFile, keyFile, serialFile, name, expireDays)
+	return subCA, true, err
+}
+
+func (ca *CA) MakeAndWriteSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, error) {
+	klog.V(4).Infof("Generating sub-CA certificate in %s, key in %s, serial in %s", certFile, keyFile, serialFile)
+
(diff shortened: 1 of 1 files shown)


deads2k:
Required I assume?  The PR looks reasonable to me

/approve
/assign @ibihim 

@ibihim owns lgtm

stlaz:
> Required I assume? The PR looks reasonable to me

It's going to make our life in Microshift a bit easier, it makes sense to include this method from the perspective of the code I added there.

ibihim (approved):
Let me check

pkg/crypto/crypto.go
```diff
@@ -693,6 +693,43 @@ func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA)
 	return signerConfig, nil
 }
 
+func (ca *CA) EnsureSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, bool, error) {
```
ibihim, line 696: Nit: It would be nice to have a comment on an exported method.

pkg/crypto/crypto.go
```diff
@@ -693,6 +693,43 @@ func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA)
...
+	if subCA, err := GetCA(certFile, keyFile, serialFile); err == nil {
+		return subCA, false, err
+	}
+	subCA, err := ca.MakeAndWriteSubCA(certFile, keyFile, serialFile, name, expireDays)
+	return subCA, true, err
+}
+
+func (ca *CA) MakeAndWriteSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, error) {
```
ibihim, line 704: Nit: It would be nice to have a comment on an exported method.

pkg/crypto/crypto.go
```diff
@@ -693,6 +693,43 @@ func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA)
 	return signerConfig, nil
 }
 
+func (ca *CA) EnsureSubCA(certFile, keyFile, serialFile, name string, expireDays int) (*CA, bool, error) {
+	if subCA, err := GetCA(certFile, keyFile, serialFile); err == nil {
+		return subCA, false, err
```
ibihim, line 698: What is the point of the boolean value?

PR #1398, opened 2022-09-22
crypto: fix the EnsureSubCA function failing to read the serial.txt file

stlaz:
The EnsureSubCA function would fail if the serial.txt file did not exist before it's called.

This PR fixes the above, adds unit tests for the Ensure* functions, and removes the deprecated ioutil module use from the crypto package.


ibihim (approved):
/lgtm

tests, comments 😍

PR #1399, opened 2022-09-27
Required for oc bump to kubectl 1.25, ldaptestclient missing interface methods for ldap.v3.Client

deejross:
Attempting to update oc dependencies results in an error in library-go for the ldap.v3 version bump due to missing methods in an interface.


ardaguclu:
As far as I can see, interface changed was happened for 3.4.4. But currently we are using 3.4.3. Maybe we can handle this change in 3.4.4 go-ldap bump PR?

soltysh (changes requested):
/hold
this is only needed when you bump ldap to v.3.4.4 which is ok to do, but in a separate PR and you need to bump it also in this repo

deejross:
Closing this since we aren't ready for ldap v3.4.4 yet

PR #1401, opened 2022-09-30
Bug 2101520: test connectivity before retry getting extension-apiserver-authentication

bertinatto:
This patch re-works [1] to avoid calling *.ApplytTo() multiple times, as that effectively start controllers multiple times and leak watchers/informers.

[1] https://github.com/bertinatto/library-go/commit/f4a305ba6337ea11bcff7d87ef4394c49e9e136a

/assign @deads2k 
CC @openshift/storage


pkg/config/serving/server.go
```diff
@@ -45,15 +47,19 @@ func ToServerConfig(ctx context.Context, servingInfo configv1.HTTPServingInfo, a
 		// In some cases the API server can return connection refused when getting the "extension-apiserver-authentication"
 		// config map.
 		err := wait.PollImmediateUntil(1*time.Second, func() (done bool, err error) {
-			lastApplyErr = authenticationOptions.ApplyTo(&config.Authentication, config.SecureServing, config.OpenAPIConfig)
-			if lastApplyErr != nil {
-				klog.V(4).Infof("Error initializing delegating authentication (will retry): %v", lastApplyErr)
+			if _, lastGetErr = kubeClient.CoreV1().Namespaces().Get(pollCtx, "default", metav1.GetOptions{}); lastGetErr != nil {
```
deads2k, line 50: realized that control-plane operators have access to see ns/default, but most operators may only have access to their current or operand namespace.  Do we require one of those to construct an event recorder (almost certainly yes)?  Let's use that value for this namespace get.

bertinatto: Done

bertinatto: hm, this could still be a problem: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-csi-snapshot-controller-operator/129/pull-ci-openshift-cluster-csi-snapshot-controller-operator-master-e2e-gcp-csi/1575937853483913216/artifacts/e2e-gcp-csi/gather-extra/artifacts/pods/openshift-cluster-storage-operator_csi-snapshot-controller-operator-5c74844449-7r2jc_csi-snapshot-controller-operator.log

bertinatto:
/hold
Fetching the operand's namespace could still be a problem in certain cases: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-csi-snapshot-controller-operator/129/pull-ci-openshift-cluster-csi-snapshot-controller-operator-master-e2e-gcp-csi/1575937853483913216/artifacts/e2e-gcp-csi/gather-extra/artifacts/pods/openshift-cluster-storage-operator_csi-snapshot-controller-operator-5c74844449-7r2jc_csi-snapshot-controller-operator.log

bertinatto:
/hold cancel
Using leases instead of namespaces.

PR #1404, opened 2022-10-11
OCPBUGS-1904: add new ConditionalStaticResourcesController

bertinatto:
This controller can be used to conditionally to deploy resources that depend on a runtime check. With this we can check if the a `VolumeSnapshotClass` CRD exists before deploying the CR.

While at it, fix the `StaticResourceController.Name()` method to return the controller name that was provided by the user. With this patch log entries will show the correct controller name.

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -35,6 +35,7 @@ type CSIControllerSet struct {
 	logLevelController                   factory.Controller
 	managementStateController            factory.Controller
 	staticResourcesController            factory.Controller
+	conditionalStaticResourcesController factory.Controller
 	credentialsRequestController         factory.Controller
 	csiConfigObserverController          factory.Controller
 	csiDriverControllerServiceController factory.Controller
@@ -52,6 +53,7 @@ func (c *CSIControllerSet) Run(ctx context.Context, workers int) {
 		c.logLevelController,
 		c.managementStateController,
 		c.staticResourcesController,
+		c.conditionalStaticResourcesController,
 		c.credentialsRequestController,
 		c.csiConfigObserverController,
 		c.csiDriverControllerServiceController,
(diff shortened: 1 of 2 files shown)


pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -368,7 +368,7 @@ func (c *StaticResourceController) Sync(ctx context.Context, syncContext factory
 }
 
 func (c *StaticResourceController) Name() string {
-	return "StaticResourceController"
+	return c.name
```
bertinatto, line 371: See  PR description. Now log entries will show the correct controller name.

bertinatto: @mfojtik @soltysh, could you approve?

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -84,6 +86,32 @@ func (c *CSIControllerSet) WithManagementStateController(operandName string, sup
...
+		manifests,
+		[]string{},
+		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithDynamicClient(dynamicClient),
+		c.operatorClient,
+		c.eventRecorder,
+	).WithConditionalResources(
+		manifests,
+		files,
+		shouldCreateFnArg,
+		shouldDeleteFnArg,
```
gnufied, line 110: Should we always supply an implicit `false` function here, rather than forcing callers to specify a dummy function which returns `false` ?

bertinatto: IMO there are cases where users might not want to specify a dummy function here. I think it's worth leaving this open.

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -84,6 +86,32 @@ func (c *CSIControllerSet) WithManagementStateController(operandName string, sup
...
+func (c *CSIControllerSet) WithConditionalStaticResourcesController(
+	name string,
+	kubeClient kubernetes.Interface,
+	dynamicClient dynamic.Interface,
+	kubeInformersForNamespace v1helpers.KubeInformersForNamespaces,
+	manifests resourceapply.AssetFunc,
+	files []string,
+	shouldCreateFnArg, shouldDeleteFnArg resourceapply.ConditionalFunction,
+) *CSIControllerSet {
+	c.conditionalStaticResourcesController = staticresourcecontroller.NewStaticResourceController(
```
deads2k, line 99: you either want this to be a slice or you want a way to stop people from calling it multple times.

jsafrane: Yeah, I'd appreciate if  CSIControllerSet had `[]Controller` instead of fields. Worth some refactoring in the next release.

bertinatto: IIRC the original idea was to prevent users from creating duplicate controllers

PR #1406, opened 2022-10-12
OCPBUGS-2125: pkg/manifest/manifest.go: add option to allow unknown capabilities

jottofar:
during capabilities filtering. When set to true a manifest will not be excluded simply because it contains an unknown capability. This is necessary to allow updates to an OCP version containing newly defined capabilities.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -172,6 +172,17 @@ func checkFeatureSets(requiredFeatureSet string, annotations map[string]string) 
 // filter. For example, setting profile non-nil and capabilities nil will return an error if the manifest's
 // profile does not match, but will never return an error about capability issues.
 func (m *Manifest) Include(excludeIdentifier *string, requiredFeatureSet *string, profile *string, capabilities *configv1.ClusterVersionCapabilitiesStatus) error {
+	return m.IncludeAllowUnknownCapabilities(excludeIdentifier, requiredFeatureSet, profile, capabilities, false)
+}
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -203,35 +214,40 @@ func (m *Manifest) Include(excludeIdentifier *string, requiredFeatureSet *string
...
 	disabledCaps := make([]string, 0, numCaps)
 
 	for _, c := range caps {
-		var isKnownCap bool
-		var isEnabledCap bool
 
-		for _, knownCapability := range capabilities.KnownCapabilities {
-			if c == knownCapability {
-				isKnownCap = true
+		if !allowUnknownCapabilities {
```
wking, line 237: I'm fine as you have it, but a chattier option would be to leave this `unknownCaps` population in place, and later on do something like:

```go
if len(unknownCaps) > 0 {
	if allowUnknownCapabilities {
		klog.V(2).Infof("allowing unrecognized capability names: %s", strings.Join(unknownCaps, ", "))
	} else {
		return fmt.Errorf("unrecognized capability names: %s", strings.Join(unknownCaps, ", "))
	}
}
```

Although maybe to make that more usable, you'd need to make this helper a method:

```go
func (m *Manifest) checkResourceEnablement(...)
```

so you could format `m` into the log line?

jottofar: Prefer to leave it as is

PR #1407, opened 2022-10-12
Copy route validation and defaulting bits from openshift-apiserver.

benluddy:
This is an almost-direct copy of the route validation and hostname assignment logic from openshift-apiserver, which has been refactored to operate on the v1 types from openshift/api instead of the internal types from openshift/openshift-apiserver. From library-go, one implementation will be sharable between openshift-apiserver and kube-apiserver to support OpenShift variants that serve the route API via CRD.

### DIFF
--- a/pkg/route/hostassignment/assignment.go
+++ b/pkg/route/hostassignment/assignment.go
@@ -1,0 +1,197 @@
+package hostassignment
+
+import (
+	"context"
+	"fmt"
+
+	authorizationv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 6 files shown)


benluddy:
/hold

For openshift-apiserver 1.25 bump.

deads2k:
/lgtm
/approve
/hold

This cannot merge until there is a proof PR that demonstrates openshift-apiserver using this branch and passing CI.

benluddy:
@deads2k The openshift-apiserver kube bump went in and the proof PR (https://github.com/openshift/openshift-apiserver/pull/328) for this move is green. PTAL when you have a chance.

PR #1409, opened 2022-10-12
OCPBUGS-2293: pkg/manifest/manifest.go: add option to allow unknown capabilities

jottofar:
during capabilities filtering. When set to true a manifest will not be excluded simply because it contains an unknown capability. This is necessary to allow updates to an OCP version containing newly defined capabilities.


wking:
I've added a Jira link for OCPBUGS-2293 `depends on` OCPBUGS-2125, which may address `but no dependents were found`.

/jira refresh

wking (approved):
/lgtm

PR #1410, opened 2022-10-14
STOR-1055: Add CSIDriver verification check in library-go

dobsonj:
https://issues.redhat.com/browse/STOR-1055
Enhancement: https://github.com/openshift/enhancements/blob/master/enhancements/storage/csi-inline-vol-security.md
Related PR's:
https://github.com/openshift/kubernetes/pull/1384
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1385
https://github.com/openshift/csi-driver-shared-resource-operator/pull/59
/hold for those 3 PR's to merge first
/cc @openshift/storage

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -15,6 +15,11 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
 )
 
+const (
+	// Label on the CSIDriver to declare the driver's effective pod security profile
+	csiInlineVolProfileLabel = "security.openshift.io/csi-ephemeral-volume-profile"
+)
+
 // ApplyStorageClass merges objectmeta, tries to write everything else
 func ApplyStorageClass(ctx context.Context, client storageclientv1.StorageClassesGetter, recorder events.Recorder, required *storagev1.StorageClass) (*storagev1.StorageClass, bool,
 	error) {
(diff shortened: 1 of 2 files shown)


dobsonj:
Tested this change with and without the label in https://github.com/openshift/csi-driver-shared-resource-operator/pull/60

dobsonj:
/assign @soltysh
for `/approve`
this PR is only on hold until the other 3 PR's from the description merge first, so we do not enforce this check prematurely.

This should not be user-noticable change, it's meant to get CI to scream loudly at us if we try to ship a CSI driver without the new required label. Tested in https://github.com/openshift/csi-driver-shared-resource-operator/pull/60 to make sure it fails without the label and passes with the label.
/label docs-approved
/label px-approved
/label qe-approved

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -115,6 +120,10 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
 	if err != nil {
 		return nil, false, err
 	}
+	err = validateRequiredCSIDriverLabels(required)
```
soltysh, line 123: Nit: I like when those are written like so:
```golang
if err = validateRequiredCSIDriverLabels(required); err != nil {
```
it makes them more readable and can potentially save from shadowing errors too.

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -173,6 +182,24 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
 	return actual, true, err
 }
 
+func validateRequiredCSIDriverLabels(required *storagev1.CSIDriver) error {
+	supportsEphemeralVolumes := false
+	for _, mode := range required.Spec.VolumeLifecycleModes {
+		if mode == storagev1.VolumeLifecycleEphemeral {
+			supportsEphemeralVolumes = true
```
soltysh, line 189: once you found it you could break the loop to save a few cycles, no?

soltysh (approved):
Minor nits that can be addressed in followup
/approve

pkg/operator/resource/resourceapply/storage.go
dobsonj: yes, updated

pkg/operator/resource/resourceapply/storage.go
dobsonj: done

dobsonj:
/unhold
everything else is merged

PR #1422, opened 2022-11-10
USHIFT-573 Regenerate client certificate on subject change

mangelajo:
MicroShift uses this library to generate certificates, including the kubelet
certificates.

The subject part of the certificate is used to authenticate the kubelet and
limit the permissions on what can access. On a host name, the client
certificate was not being regenerated, so the kubelet could not create a
new entry for the new hostname, and blocked on a loop.

This commit forces the regeneration of the client certificate when the
subject field on the disk file differs.

https://issues.redhat.com/browse/USHIFT-573

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"sync"
@@ -828,13 +829,35 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.String, lifetime time.Dur
 }
 
 func (ca *CA) EnsureClientCertificate(certFile, keyFile string, u user.Info, expireDays int) (*TLSCertificateConfig, bool, error) {
-	certConfig, err := GetTLSCertificateConfig(certFile, keyFile)
+	certConfig, err := GetClientCertificate(certFile, keyFile, u)
 	if err != nil {
 		certConfig, err = ca.MakeClientCertificate(certFile, keyFile, u, expireDays)
 		return certConfig, true, err // true indicates we wrote the files.
 	}
-
 	return certConfig, false, nil
+}
+
+func GetClientCertificate(certFile, keyFile string, u user.Info) (*TLSCertificateConfig, error) {
+	certConfig, err := GetTLSCertificateConfig(certFile, keyFile)
+	if err != nil {
+		return nil, err
+	}
+
+	if subject := certConfig.Certs[0].Subject; subjectChanged(subject, userToSubject(u)) {
+		return nil, fmt.Errorf("existing client certificate in %s was issued for a different Subject (%s)",
+			certFile, subject)
+	}
+
+	return certConfig, nil
+}
+
+func subjectChanged(existing, expected pkix.Name) bool {
+	sort.Strings(existing.Organization)
+	sort.Strings(expected.Organization)
+
+	return existing.CommonName != expected.CommonName ||
+		existing.SerialNumber != expected.SerialNumber ||
+		!reflect.DeepEqual(existing.Organization, expected.Organization)
 }
 
 func (ca *CA) MakeClientCertificate(certFile, keyFile string, u user.Info, expireDays int) (*TLSCertificateConfig, error) {
--- a/pkg/crypto/crypto_test.go
+++ b/pkg/crypto/crypto_test.go
@@ -458,4 +458,69 @@ func TestCertGeneration(t *testing.T) {
 	require.Equal(t, "testclient", clientCert.Certs[0].Subject.CommonName)
 	require.Equal(t, []string{"testclients"}, clientCert.Certs[0].Subject.Organization)
 	require.Equal(t, ca.Config.Certs[0].SubjectKeyId, clientCert.Certs[0].AuthorityKeyId)
+
+	// ensure the new client cert signed by the root CA exists but gets regenerated because the Subject changes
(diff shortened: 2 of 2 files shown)


mangelajo:
@stlaz ^
This will need a backport to 4.12 how do we mark for that?

mangelajo:
Unit tests seem to be failing but unrelated to my change, 
```
[root@rhel8shift-2 library-go]# go test -mod=vendor -race ./pkg/crypto
ok  	github.com/openshift/library-go/pkg/crypto	(cached)
```

pkg/crypto/crypto.go
```diff
@@ -829,14 +829,24 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.String, lifetime time.Dur
 
 func (ca *CA) EnsureClientCertificate(certFile, keyFile string, u user.Info, expireDays int) (*TLSCertificateConfig, bool, error) {
 	certConfig, err := GetTLSCertificateConfig(certFile, keyFile)
-	if err != nil {
+	if err != nil || subjectChanged(u, certConfig) {
```
stlaz, line 832: I believe we should follow the pattern from `EnsureServerCertificate`, which would mean:
1. create a `func GetClientCertificate(certfile, keyfile string, u user.Info) (*TLSCertificateConfig, error)
2. write the client cert equivalent of https://github.com/openshift/library-go/blob/master/pkg/crypto/crypto.go#L759-L768
3. call it here instead of `GetTLSCertificateConfig()`

stlaz (changes requested):

pkg/crypto/crypto_test.go
```diff
@@ -454,8 +454,14 @@ func TestCertGeneration(t *testing.T) {
...
+	clientCert, created, err = ca.EnsureClientCertificate(clientCertFile, clientKeyFile, &user.DefaultInfo{Name: "testclient2", Groups: []string{"testclients"}}, 2)
+	require.NoError(t, err)
+	require.NotNil(t, clientCert)
+	require.True(t, created)
+
 	require.Len(t, clientCert.Certs, 1) // we don't need to include the whole chain, unlike in server certs
-	require.Equal(t, "testclient", clientCert.Certs[0].Subject.CommonName)
+	require.Equal(t, "testclient2", clientCert.Certs[0].Subject.CommonName)
 	require.Equal(t, []string{"testclients"}, clientCert.Certs[0].Subject.Organization)
 	require.Equal(t, ca.Config.Certs[0].SubjectKeyId, clientCert.Certs[0].AuthorityKeyId)
```
stlaz, line 466: duplicate these checks for both the cases

pkg/crypto/crypto.go
mangelajo: ack, checking to match the pattern

pkg/crypto/crypto.go
```diff
@@ -828,15 +828,32 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.String, lifetime time.Dur
...
+	certConfig, err := GetTLSCertificateConfig(certFile, keyFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// Validate if the existing certificate subject has changed, if it has we need to regenerate the certificate
+	// The Subject field is converted to String (RFC 2253 Distinguished Names syntax) because disk
+	// encoding does not match 100% to struct encoding and DeepCompare fails when it should match
+	cert := certConfig.Certs[0]
+	if cert.Subject.String() != userToSubject(u).String() {
```
stlaz, line 849: After reading the internets I'd rather if we just compared the three aspects that we care about than rely on the implementation of `String().`

I think the best way would be to:
1. compare `CommonName` and `SerialNumber` of the subjects
2. order `Organization` fields of both and compare them

stlaz (changes requested):

pkg/crypto/crypto.go
mangelajo: Ack I'll update the pr

pkg/crypto/crypto.go
```diff
@@ -828,15 +828,32 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.String, lifetime time.Dur
...
+	certConfig, err := ca.GetClientCertificate(certFile, keyFile, u)
 	if err != nil {
 		certConfig, err = ca.MakeClientCertificate(certFile, keyFile, u, expireDays)
 		return certConfig, true, err // true indicates we wrote the files.
 	}
-
 	return certConfig, false, nil
 }
 
+func (ca *CA) GetClientCertificate(certFile, keyFile string, u user.Info) (*TLSCertificateConfig, error) {
```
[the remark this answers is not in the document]
mangelajo: Oops right

pkg/crypto/crypto.go
mangelajo: Done

mangelajo:
@stlaz hopefully it may look better now!, thanks for the review time

pkg/crypto/crypto.go
```diff
@@ -828,15 +829,37 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.String, lifetime time.Dur
...
+
+	if subject := certConfig.Certs[0].Subject; subjectChanged(subject, userToSubject(u)) {
+		return nil, fmt.Errorf("existing client certificate in %s was issued for a different Subject (%s)",
+			certFile, subject)
+	}
+
+	return certConfig, nil
+}
+
+func subjectChanged(existing, expected pkix.Name) bool {
```
stlaz, line 854: could we also unit test this function, please?

mangelajo: Done.

PR #1428, opened 2022-11-22
move the featureset detection for rendering after substitution so the…

deads2k:
… yaml is valid

found while working through https://github.com/openshift/cluster-kube-apiserver-operator/pull/1409

### DIFF
--- a/pkg/assets/assets.go
+++ b/pkg/assets/assets.go
@@ -32,7 +32,7 @@ type Assets []Asset
 
 // New walks through a directory recursively and renders each file as asset. Only those files
 // are rendered that make all predicates true.
-func New(dir string, data interface{}, predicates ...FileInfoPredicate) (Assets, error) {
+func New(dir string, data interface{}, manifestPredicates []FileContentsPredicate, predicates ...FileInfoPredicate) (Assets, error) {
 	files, err := LoadFilesRecursively(dir, predicates...)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/assets/assets.go
```diff
@@ -43,7 +43,20 @@ func New(dir string, data interface{}, predicates ...FileInfoPredicate) (Assets,
...
+		}
+
+		skipManifest := false
+		for _, manifestPredicate := range manifestPredicates {
+			shouldInclude, err := manifestPredicate(a.Data)
+			if err != nil {
+				return nil, fmt.Errorf("failed to check manifest filter %q: %v", path, err)
+			}
+			if !shouldInclude {
+				skipManifest = true
```
sanchezl, line 56: Consider `break` here, subsequent predicates will not change `skipManifest` again.

pkg/assets/assets.go
```diff
@@ -106,24 +119,22 @@ func assetFromTemplate(name string, tb []byte, data interface{}) (*Asset, error)
 
 type FileInfoPredicate func(path string, info os.FileInfo) (bool, error)
 
+type ManifestPredicate func(manifest []byte) (bool, error)
```
sanchezl, line 122: Consider something like `FileContentsPredicate` to make it clearer for future users.

sanchezl (approved):
/lgtm

PR #1430, opened 2022-11-23
OCPBUGS-1341: Remove stale pod network connectivity checks

pperiyasamy:
This commit makes relevant changes into synchronization logic of connectivity check controller to ensure old pod network connectivity check objects are removed from the cluster.


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -145,56 +146,70 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
 		return err
 	}
 
+	existingChecks, err := c.operatorcontrolplaneClient.ControlplaneV1alpha1().PodNetworkConnectivityChecks(c.namespace).List(ctx, metav1.ListOptions{})
```
deads2k, line 149: no live calls outside of enabled.  If you need this outside enabled, then wire up a lister

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -145,56 +146,70 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
 		}
-		syncContext.Recorder().Eventf("EndpointCheckCreated", "Created %s because it was missing.", resourcehelper.FormatResourceForCLIWithNamespace(check))
+		syncContext.Recorder().Eventf("EndpointCheckCreated", "Created %s because it was missing.", resourcehelper.FormatResourceForCLIWithNamespace(newCheck))
 	}
 
 	// TODO for checks which longer exist, mark them as completed
 
-	// TODO reap old connectivity checks
+	// reap old connectivity checks
+	for _, existingCheck := range existingChecks.Items {
```
deads2k, line 202: This prevents an external actor from creating a new connectivity check and managing it differently. I suggest adding an annotation or label to indicate, "managed by this operator" and then only deleting those items managed by this operator.

Since this is an alpha API, I'll allow an upgrade that leaves old content around or deletes old content as you wish.  But to delete old content, you'll need a two stage migration where the first stage labels everything in the namespace in 4.n-1 .

pperiyasamy: sure, now this would update old check objects with `managedBy` label, but this won't delete old stale objects during migration. hope that's ok.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
pperiyasamy: lister wire up is done now.

pperiyasamy: @deads2k lister wireup is done properly now with a dependent PR in [cno](https://github.com/openshift/cluster-network-operator/pull/1649) and checks cleanup seems to be working fine. please have a look.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -26,10 +30,16 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+const (
+	managedByLabelKey   = "managedBy"
```
deads2k, line 34: needs to be qualified.  "networking.openshift.io/managedBy"

mfojtik:
this looks good, @tkashem can you please review?

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -160,41 +178,69 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
-			if equality.Semantic.DeepEqual(existing.Spec, check.Spec) {
+			if value, ok := existing.Labels[managedByLabelKey]; ok &&
+				value == managedByLabelValue && equality.Semantic.DeepEqual(existing.Spec, newCheck.Spec) {
 				// already exists, no changes, skip
 				continue
 			}
 			updated := existing.DeepCopy()
-			updated.Spec = *check.Spec.DeepCopy()
+			updated.Spec = *newCheck.Spec.DeepCopy()
+			updated.Labels = newCheck.Labels
```
tkashem, line 211: what if  the `existing` object has other labels, should we add the new label `managedByLabelKey` to `Labels` instead? that seems a safer operation.

tkashem (commented):
looks good to me, except for a minor comment

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
pperiyasamy: yes, that's correct. changed it now.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -160,41 +178,66 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
 		return err
 	}
 
 	pnccClient := c.operatorcontrolplaneClient.ControlplaneV1alpha1().PodNetworkConnectivityChecks(c.namespace)
-	for _, check := range checks {
-		existing, err := pnccClient.Get(ctx, check.Name, metav1.GetOptions{})
+	newCheckNames := sets.NewString()
+	for _, newCheck := range newChecks {
+		newCheckNames.Insert(newCheck.Name)
+		existing, err := pnccClient.Get(ctx, newCheck.Name, metav1.GetOptions{})
```
deads2k, line 198: why live over cached?

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -160,41 +178,66 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
-			if equality.Semantic.DeepEqual(existing.Spec, check.Spec) {
+			if value, ok := existing.Labels[managedByLabelKey]; ok &&
+				value == managedByLabelValue && equality.Semantic.DeepEqual(existing.Spec, newCheck.Spec) {
 				// already exists, no changes, skip
 				continue
 			}
 			updated := existing.DeepCopy()
-			updated.Spec = *check.Spec.DeepCopy()
+			updated.Spec = *newCheck.Spec.DeepCopy()
+			updated = setWithManagedByLabel(updated)
```
deads2k, line 207: This looks like a perfect use-case for a server-side apply that blindly applies the label in question.  @tkashem at some point the org has to learn, is this the moment?  If not, perhaps this is a good spot for the first migration?  Doing a server-side apply would eliminate the need to do the initial GET and would do an upsert for you.

deads2k:
/approve
/hold

hold in case this is the moment for server side apply.  If you two agree this isn't the right spot, I'm ok deferring, but someone has to go first and this usage looked perfect.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
tkashem: the server-side-apply is orthogonal to what this PR is doing, may be a separate PR that uses server-side-apply for connectivity checker?

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -251,3 +294,17 @@ func (c *connectivityCheckController) enabled(operatorSpec *operatorv1.OperatorS
...
+
+func setWithManagedByLabel(check *v1alpha1.PodNetworkConnectivityCheck) *v1alpha1.PodNetworkConnectivityCheck {
+	if check == nil {
+		return check
+	}
+	if check.Labels == nil {
+		check.Labels = make(map[string]string)
+	}
+	check.Labels[managedByLabelKey] = managedByLabelValue
+	check.ManagedFields = []metav1.ManagedFieldsEntry{{Manager: managedByLabelValue, Operation: metav1.ManagedFieldsOperationApply,
```
tkashem, line 306: I don't think you have to set the `ManagedFields` explicitly, it's managed by the apiserver. 
 https://kubernetes.io/docs/reference/using-api/server-side-apply/#field-management

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -26,10 +30,16 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+const (
+	managedByLabelKey   = "networking.openshift.io/managedBy"
+	managedByLabelValue = "oc-connectivity-check-controller"
```
tkashem, line 35: I think for OpenShift we can adopt a naming convention for the field managers, so it's easier to make sense from the names.
I would recommend that the prefix should be the controller name, the name that appears in log/event so its easier to identify.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -160,41 +178,66 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
...
 		if err == nil {
-			if equality.Semantic.DeepEqual(existing.Spec, check.Spec) {
+			if equality.Semantic.DeepEqual(existing.Spec, newCheck.Spec) {
 				// already exists, no changes, skip
 				continue
 			}
 			updated := existing.DeepCopy()
-			updated.Spec = *check.Spec.DeepCopy()
+			updated.Spec = *newCheck.Spec.DeepCopy()
 			_, err := pnccClient.Update(ctx, updated, metav1.UpdateOptions{})
```
tkashem, line 208: It's the same controller, I think we should use `Apply` here, an example: https://github.com/kubernetes/kubernetes/pull/110173/files

tkashem (commented):
how do we want to handle the conflicts? https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts

dgrisonnet:
As discussed on slack, it will be simpler if the SSA changes were done in a separate PR so I am removing the initial hold.

/lgtm
/unhold

PR #1432, opened 2022-11-24
USHIFT-573 Regenerate client certificate on subject change

mangelajo:
MicroShift uses this library to generate certificates, including the kubelet certificates.

The subject part of the certificate is used to authenticate the kubelet and limit the permissions on what can access. On a host name, the client certificate was not being regenerated, so the kubelet could not create a new entry for the new hostname, and blocked on a loop.

This commit forces the regeneration of the client certificate when the subject field on the disk file differs.

(cherry picked from commit bf4ec512c632fc58304927448bac6c34d744d882)

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


mangelajo:
@stlaz here is the backport

benluddy:
/retitle USHIFT-573: Regenerate client certificate on subject change

mangelajo:
@benluddy this is still asking for a bugzilla. OCPBUGS created. Is it possible to merge now?

PR #1447, opened 2023-01-03
[wip] Apply example

deads2k:
POC of modification to https://github.com/openshift/library-go/pull/1430

looks like an output helper needs adjusting.


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -90,7 +89,7 @@ type connectivityCheckController struct {
 	enabledByDefault bool
 }
 
-type PodNetworkConnectivityCheckFunc func(ctx context.Context, syncContext factory.SyncContext) ([]*v1alpha1.PodNetworkConnectivityCheck, error)
+type PodNetworkConnectivityCheckFunc func(ctx context.Context, syncContext factory.SyncContext) ([]*applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration, error)
```
dgrisonnet, line 92: wouldn't it be easier to integrate if we were to take care of extracting the apply config from the PodNetworkConnectivityCheck object in the controller?

deads2k: > wouldn't it be easier to integrate if we were to take care of extracting the apply config from the PodNetworkConnectivityCheck object in the controller?

That will not handle a case of an explicit clearing of a particular field.  To express, "this field must be empty", you cannot use a struct.  That's one of the primary failings of understanding from controller-runtime as well.

deads2k: If the networking team is ok with never being able to unset a field, I suppose I'm ok creating that restriction if that's what they want, but I doubt that's what they intend in this case.

pperiyasamy: I think connectivity check controller in [CNO](https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/connectivitycheck/connectivity_check_controller.go#L116) either generates same PodNetworkConnectivityCheck object or new PodNetworkConnectivityCheck object depending on target endpoint resides on same node or moved into different node. That way i hope apply config makes sense. I'll update CNO [PR](https://github.com/openshift/cluster-network-operator/pull/1649) with apply config to see of this works or not.

pperiyasamy:
@deads2k can you fix controller for unit tests and [connectivitycheckcontroller/connectivity_check_template.go](https://github.com/openshift/library-go/blob/master/pkg/operator/connectivitycheckcontroller/connectivity_check_template.go) file with apply config as well ?

dgrisonnet:
I don't think David is planning to actually drive this effort. This PR is just a POC to show how the migration to SSA would look like. @pperiyasamy could you take the changes suggested here and and integrate them to the codebase in a separate PR?

pperiyasamy:
> I don't think David is planning to actually drive this effort. This PR is just a POC to show how the migration to SSA would look like. @pperiyasamy could you take the changes suggested here and and integrate them to the codebase in a separate PR?

sure @dgrisonnet , I will work on this.

Issue #1178, opened 2021-08-09
Migrate LeaderElection to Lease API


aojea:
Upstream has done it for the controller manager and scheduler
https://github.com/kubernetes/kubernetes/issues/80289

should we do it here too?

https://github.com/openshift/library-go/blob/965974e0af3f9fb43ac3b33bc9e67c1cbfd16a0a/pkg/config/leaderelection/leaderelection.go#L53

it comes with some upgrade problems thought

deezy30:
[adamsexotics32]()

mprahl:
This PR doesn't change the default but it makes the leader election resource lock type configurable:
https://github.com/openshift/library-go/pull/1448

PR #1448, opened 2023-01-06
Allow the leader election resource lock type to be configurable

mprahl:
The controllercmd package always utilized a configmapsleases resource lock for leader election, but some client may entirely want to move to leases. This commit keeps the default to configmapsleases but allows it to be configurable to whatever the
k8s.io/client-go/tools/leaderelection/resourcelock package allows.

Related issues:
https://github.com/openshift/library-go/issues/1178
https://issues.redhat.com/browse/ACM-2653

### DIFF
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -29,8 +29,20 @@ import (
 // See https://github.com/kubernetes/kubernetes/issues/107454 for
 // details on how to migrate to "leases" leader election.
 // Don't forget the callbacks!
-// TODO: In the next version we should switch to using "leases"
 func ToLeaderElectionWithConfigmapLease(clientConfig *rest.Config, config configv1.LeaderElection, component, identity string) (leaderelection.LeaderElectionConfig, error) {
+	return toLeaderElection(clientConfig, config, component, identity, resourcelock.ConfigMapsLeasesResourceLock)
+}
+
+// ToLeaderElectionWithLease returns a "leases" based leader
+// election config that you just need to fill in the Callback for.
+// Don't forget the callbacks!
+func ToLeaderElectionWithLease(clientConfig *rest.Config, config configv1.LeaderElection, component, identity string) (leaderelection.LeaderElectionConfig, error) {
+	return toLeaderElection(clientConfig, config, component, identity, resourcelock.LeasesResourceLock)
+}
+
+// toLeaderElection returns a leader election config that you just need to fill in the Callback for. The input
+// resourceLock must be a value from the k8s.io/client-go/tools/leaderelection/resourcelock package.
+func toLeaderElection(clientConfig *rest.Config, config configv1.LeaderElection, component, identity, resourceLock string) (leaderelection.LeaderElectionConfig, error) {
 	kubeClient, err := kubernetes.NewForConfig(clientConfig)
 	if err != nil {
 		return leaderelection.LeaderElectionConfig{}, err
@@ -57,7 +69,7 @@ func ToLeaderElectionWithConfigmapLease(clientConfig *rest.Config, config config
 	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
 	eventRecorder := eventBroadcaster.NewRecorder(clientgoscheme.Scheme, corev1.EventSource{Component: component})
 	rl, err := resourcelock.New(
-		resourcelock.ConfigMapsLeasesResourceLock,
+		resourceLock,
 		config.Namespace,
 		config.Name,
 		kubeClient.CoreV1(),
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -2,6 +2,7 @@ package controllercmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/component-base/metrics"
 	"k8s.io/component-base/metrics/legacyregistry"
@@ -59,19 +61,29 @@ type ControllerContext struct {
 	OperatorNamespace string
 }
 
+type leaderElectionResourceLock string
+
+const (
+	LeasesResourceLock           leaderElectionResourceLock = resourcelock.LeasesResourceLock
+	ConfigMapsLeasesResourceLock leaderElectionResourceLock = resourcelock.ConfigMapsLeasesResourceLock
+)
+
+var ErrInvalidResourceLock = errors.New("the leader election resource lock is not supported")
+
 // defaultObserverInterval specifies the default interval that file observer will do rehash the files it watches and react to any changes
 // in those files.
 var defaultObserverInterval = 5 * time.Second
 
 // ControllerBuilder allows the construction of an controller in optional pieces.
 type ControllerBuilder struct {
-	kubeAPIServerConfigFile *string
-	clientOverrides         *client.ClientConnectionOverrides
-	leaderElection          *configv1.LeaderElection
-	fileObserver            fileobserver.Observer
-	fileObserverReactorFn   func(file string, action fileobserver.ActionType) error
-	eventRecorderOptions    record.CorrelatorOptions
-	componentOwnerReference *corev1.ObjectReference
+	kubeAPIServerConfigFile    *string
+	clientOverrides            *client.ClientConnectionOverrides
+	leaderElection             *configv1.LeaderElection
+	leaderElectionResourceLock leaderElectionResourceLock
+	fileObserver               fileobserver.Observer
+	fileObserverReactorFn      func(file string, action fileobserver.ActionType) error
+	eventRecorderOptions       record.CorrelatorOptions
+	componentOwnerReference    *corev1.ObjectReference
 
 	startFunc          StartFunc
 	componentName      string
@@ -154,6 +166,14 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 
 	defaulted := leaderelectionconverter.LeaderElectionDefaulting(leaderElection, defaultNamespace, defaultName)
 	b.leaderElection = &defaulted
+	return b
+}
+
+// WithLeaderElectionResourceLock sets the resource lock. If not set, controllercmd.ConfigMapsLeasesResourceLock will
+// be used for backwards compatibility.
+func (b *ControllerBuilder) WithLeaderElectionResourceLock(resourceLock leaderElectionResourceLock) *ControllerBuilder {
+	b.leaderElectionResourceLock = resourceLock
+
 	return b
 }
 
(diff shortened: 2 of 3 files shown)


mprahl:
Could I please get a review on this PR?

pkg/config/leaderelection/leaderelection.go
```diff
@@ -29,8 +29,14 @@ import (
...
 // Don't forget the callbacks!
-// TODO: In the next version we should switch to using "leases"
 func ToLeaderElectionWithConfigmapLease(clientConfig *rest.Config, config configv1.LeaderElection, component, identity string) (leaderelection.LeaderElectionConfig, error) {
+	return ToLeaderElection(clientConfig, config, component, identity, resourcelock.ConfigMapsLeasesResourceLock)
+}
+
+// ToLeaderElection returns a leader election config that you just need to fill in the Callback for. The input
+// resourceLock must be a value from the k8s.io/client-go/tools/leaderelection/resourcelock package.
+// Don't forget the callbacks!
+func ToLeaderElection(clientConfig *rest.Config, config configv1.LeaderElection, component, identity, resourceLock string) (leaderelection.LeaderElectionConfig, error) {
```
dgrisonnet, line 39: The Endpoint and ConfigMap locks are not supported by upstream since Kubernetes v1.24 so the only locking mechanisms that should be supported are lease, endpointlease and configmaplease. Since there is hardly a way to limit if any string is allowed to be passed, it would be better to write two functions:
- ToLeaderElectionWithEndpointLease
- ToLeaderElectionWithLease

And make ToLeaderElection internal to prevent people from using unsupported features.

pkg/controller/controllercmd/builder.go
```diff
@@ -157,6 +159,14 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	return b
 }
 
+// WithLeaderElectionResourceLock sets the resource lock based on the options in the
+// k8s.io/client-go/tools/leaderelection/resourcelock package. A safe default is configmapsleases.
+func (b *ControllerBuilder) WithLeaderElectionResourceLock(resourceLock string) *ControllerBuilder {
+	b.leaderElectionResourceLock = resourceLock
+
+	return b
+}
```
dgrisonnet, line 168: As mentioned in my other comment, we shouldn't allow every type of lock since some of them are not supported anymore.

pkg/config/leaderelection/leaderelection.go
mprahl: I disagree. I don't think this library should limit the lease options based on what is supported or unsupported in the latest Kubernetes version. If the user of the library is running against an older version of Kubernetes (e.g. OpenShift 3.11), they wouldn't be able to use the lease API.

Also, rather than try to keep up with new lease options in client-go as client-go gets updated, I'd rather just rely on its validation here rather than add further restrictions in this library:
https://github.com/kubernetes/client-go/blob/e6998df99e84a57bfe500bf6fa96e9a724b1698c/tools/leaderelection/resourcelock/interface.go#L165-L211

mprahl:
@dgrisonnet could you please provide your thoughts on my last comment?

pkg/config/leaderelection/leaderelection.go
dgrisonnet: @tkashem do you have an opinion on that since you work on the initial move from ConfigMap to ConfigmapLease in https://github.com/openshift/library-go/pull/1282.

pkg/controller/controllercmd/builder.go
```diff
@@ -328,8 +349,19 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+	// default to configmapsleases for leader election
+	if b.leaderElectionResourceLock == "" || b.leaderElectionResourceLock == ConfigMapsLeasesResourceLock {
+		leaderElection, err = leaderelectionconverter.ToLeaderElectionWithConfigmapLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
+	} else if b.leaderElectionResourceLock == LeasesResourceLock {
+		leaderElection, err = leaderelectionconverter.ToLeaderElectionWithLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
+	} else if b.leaderElectionResourceLock == EndpointsLeasesResourceLock {
+		leaderElection, err = leaderelectionconverter.ToLeaderElectionWithEndpointLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
+	} else {
+		return fmt.Errorf("%w: %s", ErrInvalidResourceLock, b.leaderElectionResourceLock)
+	}
```
dgrisonnet, line 363: it would be cleaner to have a switch here

mprahl: Good suggestion. Thanks!

pkg/config/leaderelection/leaderelection.go
tkashem: > I disagree. I don't think this library should limit the lease options based on what is supported or unsupported in the latest Kubernetes version

library-go can be a bit opinionated, in the interest of convergence, ideally we want OpenShift operators to start using lease (i can't see a reason why a user would have opinion to use configmap vs endpoint here, the leader election feature on the surface is invariant no matter what the underlying lock is)


Can you help us understand your actual use case here?

mprahl: @tkashem my main goal is to exclusively use the lease API for leader election, but having support for configmapsleases allows consumers of the library to update to the latest version of the library without making the switch just yet. OpenShift can later decide when to change the default and/or remove the configmapsleases option.

I personally don't need the endpoint lease option, but was asked by @dgrisonnet to add it in a comment above. I'll remove it now that the library attempts to be opinionated.

mprahl:
Okay, the PR is updated.

pkg/config/leaderelection/leaderelection.go
dgrisonnet: My initial comment was meant to reduce the scope of resources to only the ones that are supported today, but it is fine with me if we just want to add leases now. The things I didn't want was to make it possible to use leader election with configmaps and secrets as these are not supported anymore.

mprahl:
@dgrisonnet is this good to merge then?

pkg/config/leaderelection/leaderelection.go
```diff
@@ -29,8 +29,20 @@ import (
 // See https://github.com/kubernetes/kubernetes/issues/107454 for
 // details on how to migrate to "leases" leader election.
 // Don't forget the callbacks!
-// TODO: In the next version we should switch to using "leases"
```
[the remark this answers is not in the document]
mprahl: @dgrisonnet and @tkashem I moved the TODO to where the logic for the default resource lock is made. Is this okay with you?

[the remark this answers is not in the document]
tkashem: Yes, actually, that was the goal

we want the following:
- A: a brand new operator being written should us `lease` (preferably by default)
- B: existing operators who are currently using `configmapsleases` should use `lease` (preferably by default)
- C: existing operators who have not yet upgraded libray-go for a while (still using `configmaps`) should start using `configmapsleases` as soon as they upgrade to the latest library-go (we can't have these operators use `lease` by default, it will break their upgrade flow)

I am concerned about `C` - if we don't have any existing operators in `C` category today then we can safely rename `ToLeaderElectionWithConfigmapLease` to `ToLeaderElectionWithLease` and have `lease` be the default option. 
So I would like to know if `C` is empty today - it can be possible today to count C, maybe by having a test in CI?

or is there a way we can break (static, compile time) these operators in C when they upgrade library-go so they can opt-in? I would like us to be able to avoid these changes if possible, especially if C is empty.

pkg/controller/controllercmd/builder.go
```diff
@@ -157,6 +169,14 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	return b
 }
 
+// WithLeaderElectionResourceLock sets the resource lock. If not set, controllercmd.ConfigMapsLeasesResourceLock will
+// be used for backwards compatibility.
+func (b *ControllerBuilder) WithLeaderElectionResourceLock(resourceLock leaderElectionResourceLock) *ControllerBuilder {
```
tkashem, line 174: instead of using the lock as a parameter, can we have two methods?
```
WithLeaderElectionLeaseResourceLock() *ControllerBuilder
WithLeaderElectionConfigmapLeaseResourceLock() *ControllerBuilder
```

this way we don't have to have validation around resource lock names and in future we can simply remove `WithLeaderElectionConfigmapLeaseResourceLock`

pkg/config/leaderelection/leaderelection.go
```diff
@@ -29,8 +29,20 @@ import (
...
 func ToLeaderElectionWithConfigmapLease(clientConfig *rest.Config, config configv1.LeaderElection, component, identity string) (leaderelection.LeaderElectionConfig, error) {
+	return toLeaderElection(clientConfig, config, component, identity, resourcelock.ConfigMapsLeasesResourceLock)
+}
+
+// ToLeaderElectionWithLease returns a "leases" based leader
+// election config that you just need to fill in the Callback for.
+// Don't forget the callbacks!
+func ToLeaderElectionWithLease(clientConfig *rest.Config, config configv1.LeaderElection, component, identity string) (leaderelection.LeaderElectionConfig, error) {
```
tkashem, line 39: we can remove `ToLeaderElectionWithLease` and restore `toLeaderElection` to `ToLeaderElection`

pkg/controller/controllercmd/builder.go
```diff
@@ -328,8 +348,19 @@ func (b *ControllerBuilder) Run(ctx context.Context, config *unstructured.Unstru
...
+	// TODO: In the next version we should switch to using "leases" by default
+	case "", ConfigMapsLeasesResourceLock:
+		leaderElection, err = leaderelectionconverter.ToLeaderElectionWithConfigmapLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
+	case LeasesResourceLock:
+		leaderElection, err = leaderelectionconverter.ToLeaderElectionWithLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
+	default:
+		return fmt.Errorf("%w: %s", ErrInvalidResourceLock, b.leaderElectionResourceLock)
+	}
 
-	leaderElection, err := leaderelectionconverter.ToLeaderElectionWithConfigmapLease(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity)
```
tkashem, line 332: use proposed `eaderelectionconverter.ToLeaderElection(leaderConfig, *b.leaderElection, b.componentName, b.instanceIdentity, b.leaderElectionResourceLock`

we can get rid of the switch here

pkg/controller/controllercmd/cmd.go
```diff
@@ -43,6 +44,10 @@ type ControllerCommandConfig struct {
 	// DisableServing disables serving metrics, debug and health checks and so on.
 	DisableServing bool
 
+	// LeaderElectionResourceLock sets the resource lock. If not set, controllercmd.ConfigMapsLeasesResourceLock will
+	// be used for backwards compatibility.
+	LeaderElectionResourceLock leaderElectionResourceLock
```
tkashem, line 49: we need to know whether it's `lease` or `configmapsleases`, so can we make it a boolean
```
// DEPRECATED: at some point everything should uses lease by default
OverrideConfigmapsLeasesResourceLock bool
```

This will allow new operators to opt in for `lease`

pkg/controller/controllercmd/cmd.go
```diff
@@ -282,6 +288,7 @@ func (c *ControllerCommandConfig) StartController(ctx context.Context) error {
 		WithKubeConfigFile(c.basicFlags.KubeConfigFile, nil).
 		WithComponentNamespace(c.basicFlags.Namespace).
 		WithLeaderElection(config.LeaderElection, c.basicFlags.Namespace, c.componentName+"-lock").
+		WithLeaderElectionResourceLock(c.LeaderElectionResourceLock).
```
tkashem, line 291: replace it with
```
if c.OverrideConfigmapsLeasesResourceLock {
     builder = builder. WithLeaderElectionConfigmapLeaseResourceLock() 
} else {
    builder = builder. WithLeaderElectionLeaseResourceLock() 
}
```

pkg/config/leaderelection/leaderelection.go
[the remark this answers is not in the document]
deads2k: Leases are not 100% reliable today.  It's been a while since the switch I think, I'd simply switch.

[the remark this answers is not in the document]
tkashem: @mprahl can you revise the pr to switch to `lease`? (disregard `C`, that means no need to make anything configurable)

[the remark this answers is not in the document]
mprahl: @tkashem in ACM we would like to still have the option for configmapsleases if that's okay. I'd make `lease` the default though.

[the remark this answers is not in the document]
tkashem: @mprahl  yeah, sure, go ahead. (just curious, you have an operator that is currently using configmaps?)

mprahl:
Closing in favor of:
https://github.com/openshift/library-go/commit/caaf57df1ccae32c0df21021eff272f06cbbfe91

PR #1450, opened 2023-01-09
Extend check controller with server side apply

pperiyasamy:
This extends connectivity check controller's sync function with server side apply so that check objects can be applied directly on the K8s without need for checking existing object fields and calling update function.

### DIFF
--- a/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
+++ b/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
+	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/api/operatorcontrolplane/v1alpha1"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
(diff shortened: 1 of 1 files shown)


pperiyasamy:
> @deads2k @tkashem @dgrisonnet

This is tested with required changes in connectivity check controller in cno https://github.com/openshift/cluster-network-operator/pull/1649, it seems to be working perfectly fine. I can also verify conflict and ownership transfer cases with `kubectl apply` and `kubectl edit `operations.

pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -186,11 +196,39 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
 		}
 	}
 
-	newChecks, err := c.podNetworkConnectivityCheckFn(ctx, syncContext)
+	var newCheckNames sets.String
+	if c.podNetworkConnectivityCheckFn != nil {
+		newCheckNames, err = c.handlePodNetworkConnectivityCheckFn(ctx, syncContext)
```
deads2k, line 201: any plan to retire this function once the transition is made?  I prefer to have a single path if possible.

deads2k:
At some point a lister may be necessary to reduce QPS, but this is a significant improvement over the current logic

/lgtm

PR #1453, opened 2023-01-10
expose method for finding featuregates

deads2k:
Needed to cleanly incorporate featuregate handling.


deads2k:
just exposure, I"m going for it.

PR #1454, opened 2023-01-13
MULTIARCH-3212: Add region mapping config for PowerVS cloud

dharaneeshvrd:
This is to keep the region mapping of PowerVS in common package, so all the components can access.

Region mapping is required to derive the location for VPC or COS from PowerVS region, since PowerVS follows different location notation, this mapping is required. With this mapping, user need to pass PowerVS region alone and other resource like VPC's region can be derived from this mapping.

As per @JoelSpeed's suggestion [here](https://github.com/openshift/api/pull/1383#issuecomment-1377039920) raised this PR.

Fixes https://issues.redhat.com/browse/MULTIARCH-3212

### DIFF
--- a/pkg/config/cloud/powervs/powervs_regions.go
+++ b/pkg/config/cloud/powervs/powervs_regions.go
@@ -1,0 +1,102 @@
+package powervs
+
+import (
+	"fmt"
+)
+
+// Since there is no API to query these, we have to hard-code them here.
+
+// Region describes resources associated with a region in Power VS.
+// We're using a few items from the IBM Cloud VPC offering. The region names
+// for VPC are different so another function of this is to correlate those.
+type Region struct {
+	Description string
+	VPCRegion   string
+	Zones       []string
+}
+
+// Regions provides a mapping between Power VS and IBM Cloud VPC regions.
+var Regions = map[string]Region{
+	"dal": {
+		Description: "Dallas, USA",
+		VPCRegion:   "us-south",
+		Zones:       []string{"dal12"},
+	},
+	"eu-de": {
+		Description: "Frankfurt, Germany",
+		VPCRegion:   "eu-de",
+		Zones: []string{
+			"eu-de-1",
+			"eu-de-2",
+		},
+	},
+	"lon": {
+		Description: "London, UK.",
+		VPCRegion:   "eu-gb",
+		Zones: []string{
+			"lon04",
+			"lon06",
+		},
+	},
+	"mon": {
+		Description: "Montreal, Canada",
+		VPCRegion:   "ca-tor",
+		Zones:       []string{"mon01"},
+	},
+	"osa": {
+		Description: "Osaka, Japan",
+		VPCRegion:   "jp-osa",
+		Zones:       []string{"osa21"},
+	},
+	"syd": {
+		Description: "Sydney, Australia",
+		VPCRegion:   "au-syd",
+		Zones: []string{
+			"syd04",
+			"syd05",
+		},
+	},
+	"sao": {
+		Description: "São Paulo, Brazil",
+		VPCRegion:   "br-sao",
+		Zones:       []string{"sao01"},
+	},
+	"tok": {
+		Description: "Tokyo, Japan",
+		VPCRegion:   "jp-tok",
+		Zones:       []string{"tok04"},
+	},
+	"us-east": {
+		Description: "Washington DC, USA",
+		VPCRegion:   "us-east",
+		Zones:       []string{"us-east"},
+	},
+}
+
+// VPCRegionForPowerVSRegion returns the VPC region for the specified PowerVS region.
+func VPCRegionForPowerVSRegion(region string) (string, error) {
+	if r, ok := Regions[region]; ok {
+		return r.VPCRegion, nil
+	}
+
+	return "", fmt.Errorf("VPC region corresponding to a PowerVS region %s not found ", region)
+}
+
+// RegionShortNames returns the list of region names
+func RegionShortNames() []string {
+	keys := []string{}
+	for r := range Regions {
+		keys = append(keys, r)
+	}
+	return keys
+}
+
+// ValidateVPCRegion validates that given VPC region is known/tested.
+func ValidateVPCRegion(region string) bool {
+	for r := range Regions {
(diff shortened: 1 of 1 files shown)


pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,111 @@
...
+// ValidateVPCRegion validates that given VPC region is known/tested.
+func ValidateVPCRegion(region string) bool {
+	found := false
+	for r := range Regions {
+		if region == Regions[r].VPCRegion {
+			found = true
+			break
+		}
+	}
+	return found
```
mkumatag, line 110:
```suggestion
	for r := range Regions {
		if region == Regions[r].VPCRegion {
			return true
		}
	}
	return false
```

mkumatag (changes requested):
few minor changes and also mention the original location of code from where we are moving this code

pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,111 @@
...
+
+// RegionShortNames returns the list of region names
+func RegionShortNames() []string {
+	keys := make([]string, len(Regions))
+	i := 0
+	for r := range Regions {
+		keys[i] = r
+		i++
+	}
+	return keys
```
mkumatag, line 98:
```suggestion
	keys := []string{}
	for r := range Regions {
		keys = append(keys, r)
	}
	return keys
```

dharaneeshvrd:
Currently this mapping is placed inside installer repo here - https://github.com/openshift/installer/blob/master/pkg/types/powervs/powervs_regions.go
Once it moved here, we can remove from installer.

dharaneeshvrd:
@mkumatag Addressed your comments.
@clnperez @Prashanth684
Could you please review this PR when you get a chance?

dharaneeshvrd:
@soltysh @sttts Could you please take a look at this PR?

pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,107 @@
...
+			"syd04",
+			"syd05",
+		},
+	},
+	"sao": {
+		Description: "São Paulo, Brazil",
+		VPCRegion:   "br-sao",
+		Zones:       []string{"sao01"},
+	},
+	"tor": {
```
clnperez, line 64: can you delete this element?

clnperez (changes requested):

pkg/config/cloud/powervs/powervs_regions.go
dharaneeshvrd: Removed tor region.

pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,102 @@
...
+// Region describes resources associated with a region in Power VS.
+// We're using a few items from the IBM Cloud VPC offering. The region names
+// for VPC are different so another function of this is to correlate those.
+type Region struct {
+	Description string
+	VPCRegion   string
+	Zones       []string
+}
+
+// Regions holds the regions for IBM Power VS, and descriptions used during the survey.
```
mjturek, line 18: nit: "the survey" is an IPI term. I change this to:

"Regions provides a mapping between Power VS and IBM Cloud VPC regions."

pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ValidateVPCRegion validates that given VPC region is known/tested.
+func ValidateVPCRegion(region string) bool {
+	for r := range Regions {
+		if region == Regions[r].VPCRegion {
+			return true
+		}
+	}
+	return false
+}
```
mjturek, line 102: I know these utility functions are useful for IPI, but do they belong in this repo? I'm personally fine with it but would like to hear from others.

pkg/config/cloud/powervs/powervs_regions.go
dharaneeshvrd: Fixed.

pkg/config/cloud/powervs/powervs_regions.go
dharaneeshvrd: Need to use `func VPCRegionForPowerVSRegion()` in image registry changes. Need to decide on remaining funcs. I am fine with it too on keeping them here.

dharaneeshvrd:
@sttts @soltysh
Could you please take a look at this PR and provide your comments?

pkg/config/cloud/powervs/powervs_regions.go
```diff
@@ -0,0 +1,102 @@
...
+// We're using a few items from the IBM Cloud VPC offering. The region names
+// for VPC are different so another function of this is to correlate those.
+type Region struct {
+	Description string
+	VPCRegion   string
+	Zones       []string
+}
+
+// Regions provides a mapping between Power VS and IBM Cloud VPC regions.
+var Regions = map[string]Region{
```
soltysh, line 19: This looks like configuration data, not something we'd want to keep in a repository. 
/hold

soltysh (changes requested):

pkg/config/cloud/powervs/powervs_regions.go
mjturek: @soltysh For some context this is a mapping between two different cloud services in IBM Cloud. Currently this is in the openshift-installer, but we found that it would need to be referenced by other projects as well. It was then suggested to put it here in a comment in this PR https://github.com/openshift/api/pull/1383

If something like this doesn't belong here, could you suggest where it should go?

clnperez:
@mjturek i have forgotten where this discussion landed. should we close this?

dharaneeshvrd:
I think We can close this one.
@clnperez As per RH team's suggestion we decided to keep the region mapping in IBM's repo here https://github.com/ppc64le-cloud/powervs-utils/blob/main/region.go since it does not have any other major functionality, we thought of keeping it here so we won't need to worry about dependency it would bring when we update a new region.
Yet to change the cluster-image-registry-operator to use region mapping from here.
@mjturek If need openshift/installer can also use this mapping I believe.

PR #1455, opened 2023-01-13
OCPBUGS-4551: [release-4.11] guard controller: set an explicit hostname to avoid name collisions

ingvagabund:
Backporting https://github.com/openshift/library-go/pull/1427


soltysh (approved):
/lgtm
/approve

PR #1456, opened 2023-01-17
STOR-947: allow separate evaluation of storage class state

dobsonj:
https://issues.redhat.com/browse/STOR-947
In some cases like https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/126 the operator needs a way to check the StorageClassState first, take some action on it, then apply the SC at the end. This PR refactors the StorageClassStateEvaluator to make it possible to do this while only checking the StorageClassState once per sync.
/cc @openshift/storage


dobsonj:
/assign @bertinatto 
for approval

PR #1457, opened 2023-01-19
WRKLDS-649: Guard controller: set the readiness probe endpoint explicitly

ingvagabund:
KS and KCM does not publish /readyz endpoint, only /healthz. On the other hand KA publishes both /healthz and /readyz. The /readyz endpoint is more suitable for checking the readiness.

Thus, the guard controller needs to be allow to explicitly set the readiness probe endpoint.
/hold

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -37,6 +37,7 @@ type GuardController struct {
 	targetNamespace, podResourcePrefix string
 	operatorName                       string
 	readyzPort                         string
+	readyzEndpoint                     string
 	operandPodLabelSelector            labels.Selector
 
 	nodeLister corelisterv1.NodeLister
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -77,6 +79,7 @@ func NewGuardController(
 		podResourcePrefix:       podResourcePrefix,
 		operatorName:            operatorName,
 		readyzPort:              readyzPort,
+		readyzEndpoint:          readyzEndpoint,
```
deads2k, line 82: this lgtm.  Is it stuck on something in particular?

deads2k:
/approve

so this can merge once someone is happy with the shape.

pkg/operator/staticpod/controller/guard/guard_controller.go
ingvagabund: Just preempted by higher priority tasks

ingvagabund:
Proof PRs:
- https://github.com/openshift/cluster-kube-scheduler-operator/pull/459
- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1437

The KA guards are posting ready when reaching to the /readyz endpoint. KS stays unchanged.

soltysh (approved):
/lgtm

PR #1459, opened 2023-01-27
STOR-1128: Export code that sets CSI StorageClass as the default one

bertinatto:
We need this for the vSphere CSI Driver Operator, where there is a different StorageClass controller.

CC @openshift/storage


bertinatto:
/assign @dobsonj
Who recently touched this code.

PR #1462, opened 2023-02-07
leaderelection: make election client verbose

mfojtik:
This makes the leader election client-go client super verbose (printing DNS timing, connection errors, etc.).
In practical life, the leader election client ought to be the first client to indicate a networking problem in the cluster (it is unable to get the config map), so having extra debugging information in logs to investigate the client error is helpful.


mfojtik:
proof: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/696

PR #1466, opened 2023-02-13
Avoid unneeded apply requests to the API server

pperiyasamy:
Most of the times connectivity check controller produces same check apply object that already exists in K8s, but it blindly applies same change to the api server. So this avoids no-op apply operation by comparing new object with existing ones.

### DIFF
--- a/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
+++ b/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
@@ -200,7 +200,15 @@ func (c *connectivityCheckController) Sync(ctx context.Context, syncContext fact
 	if c.podNetworkConnectivityCheckFn != nil {
 		newCheckNames, err = c.handlePodNetworkConnectivityCheckFn(ctx, syncContext)
 	} else if c.podNetworkConnectivityCheckApplyFn != nil {
-		newCheckNames, err = c.handlePodNetworkConnectivityCheckApplyFn(ctx, syncContext)
+		var existingChecksApply []*applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration
+		for _, checks := range existingChecks {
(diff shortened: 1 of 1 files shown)


pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go
```diff
@@ -274,6 +283,9 @@ func (c *connectivityCheckController) handlePodNetworkConnectivityCheckApplyFn(c
 	for _, newCheck := range newChecks {
 		newCheckNames.Insert(*newCheck.Name)
 		newCheck.WithLabels(map[string]string{managedByLabelKey: managedByLabelValue})
+		if c.exists(existingChecksApply, newCheck) {
```
deads2k, line 286: this is the right spot, but for apply you want to know, "will this resource change when I apply".  That isn't the same as semantic equals.  For instance, someone setting labels and annotations doesn't matter.

pperiyasamy: Doesn't [this](https://github.com/openshift/library-go/pull/1466/files#diff-0410af2e751ae8570096acf650682f1fdc74977172ebf688cf40083be6e95a7fR205) pickup only relevant fields managed by this controller and using semantic equals achieve "will this resource change when I apply" ?

deads2k: Perhaps a unit test will be instructive, but I think that will take the current object and create the applyconfiguration that would produce it.  Meaning that if the object via the API contains label[foo]=bar and the labels in the desiredCheck are empty, there will always be a "not equal".

A unit test could probably show if this was actually the case

pperiyasamy: oh! yes, i see it, applyconfiguration need not have fields set for which no modifications needed. so equal check is invalid here.

PR #1469, opened 2023-02-16
API-1509: Enable aes-gcm encryption provider

dgrisonnet:
proof PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1446

PR #1470, opened 2023-02-17
Fix test-e2e-encryption rule

dgrisonnet:
The test-e2e-encryption rule was broken which caused the `e2e-aws-encryption` CI job to not do anything. For instance, if we look at the [job logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_library-go/1469/pull-ci-openshift-library-go-master-e2e-aws-encryption/1626266910360342528/artifacts/e2e-aws-encryption/test/build-log.txt), the `make test-e2e-encryption` invocation is yielding the following error:
```
make: Nothing to be done for 'test-e2e-encryption'.
```

The intended way to run these test seem to call the `test-unit` rule from build-machinery: https://github.com/openshift/build-machinery-go/blob/bd58a901bbfeb5f63f027cfaa35de4863bb535fb/make/targets/golang/test-unit.mk#L5-L13. And that is what we are doing today in cluster-kube-apiserver-operator: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Makefile#L39-L60.

Alongside this fix, I also added the flags that we are setting in the kas-operator repo as they seem to be a good fit for this test aswell.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ verify-podnetworkconnectivitychecks:
 	$(MAKE) -C pkg/operator/connectivitycheckcontroller verify
 
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
+test-e2e-encryption: GO_TEST_FLAGS += -v
+test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
+test-e2e-encryption: GO_TEST_FLAGS += -p 1
+test-e2e-encryption: test-unit
 .PHONY: test-e2e-encryption
 
 test-e2e-monitoring:
--- a/test/e2e-encryption/encryption_test.go
+++ b/test/e2e-encryption/encryption_test.go
@@ -382,7 +382,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 }
 
 const encryptionTestOperatorCRD = `
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: encryptiontests.operator.openshift.io
--- a/test/e2e-encryption/encryption_test.go
+++ b/test/e2e-encryption/encryption_test.go
@@ -26,12 +26,14 @@ import (
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/clock"
 	"sigs.k8s.io/yaml"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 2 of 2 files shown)


dgrisonnet:
As explained in https://github.com/openshift/library-go/pull/1469#issuecomment-1434696095, this test seems to have been broken in CI since the time it was added, because the Makefile rule wasn't doing anything.

tkashem:
/lgtm
/approve

Thanks !

dgrisonnet:
There was a compilation error in the test, I fixed it in the latest commit.

/retest

Makefile
```diff
@@ -24,6 +24,10 @@ verify-podnetworkconnectivitychecks:
 	$(MAKE) -C pkg/operator/connectivitycheckcontroller verify
 
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
+test-e2e-encryption: GO_TEST_FLAGS += -v
+test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
```
p0lyn0mial, line 28: 4h for a single unit test is quite a long timeout :)

Makefile
```diff
@@ -24,6 +24,10 @@ verify-podnetworkconnectivitychecks:
...
+test-e2e-encryption: GO_TEST_FLAGS += -p 1
```
p0lyn0mial, line 29: is this required ?

test/e2e-encryption/encryption_test.go
```diff
@@ -628,3 +638,34 @@ func (p *provider) EncryptedGRs() []schema.GroupResource {
 func (p *provider) ShouldRunEncryptionControllers() (bool, error) {
 	return true, nil
 }
+
+func extractOperatorSpec(obj *unstructured.Unstructured, fieldManager string) (*applyoperatorv1.OperatorSpecApplyConfiguration, error) {
+	castObj := &operatorv1.OpenShiftAPIServer{}
```
p0lyn0mial, line 643: isn't the test using some fake resource ?

Makefile
dgrisonnet: This is mostly copied from https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Makefile#L45-L48.

And there nothing unit about this test besides the fact that we invoke "test-unit" from build-machinery. The test creates 7 different keys: https://github.com/openshift/library-go/blob/master/test/e2e-encryption/encryption_test.go#L366-L371 which should result in at least 21 revisions if we include the migrations, which can take a while :sweat_smile:

Makefile
dgrisonnet: Not right now since there is only one test in the package, but it can be a great protection in the future if we add another test as it would prevent running different encryption changes in parallel against the same cluster.

Based on that, do you think we should keep it?

test/e2e-encryption/encryption_test.go
dgrisonnet: correct :man_facepalming:, I blindly copy-pasted these and didn't check

Issue #1472, opened 2023-02-22
resourcesynccontroller miss key configmaps


lance5890:
when resourcesynccontroller  miss key configmap in CombineCABundleConfigMaps, and err is notFound, it will continue and override this failure, this will cause the caller func generate wrong configgurations
https://github.com/openshift/library-go/blob/81bb04eedffd559bb9d50351dcada9de497182ad/pkg/operator/resourcesynccontroller/core.go#L17

lance5890:
@ingvagabund @atiratree @deads2k @soltysh

lance5890:
This PR has closed abnormally , I will propose a new PR later

PR #191, opened 2019-01-17
combine cabundle configmaps

deads2k:
This is the logic we use the kube-apiserver to combine ca-bundle configmaps

/assign @mfojtik

PR #1473, opened 2023-02-22
Notice the notFound err

lance5890:
**1 bug description**
when resourcesynccontroller miss key configmap in CombineCABundleConfigMaps, and err is notFound, it will continue and ignore this failure, this will cause the caller func generate wrong configurations

**2 related issues**:

**more information, please read the following issues**
[# issue 702 openshift/cluster-kube-controller-manager-operator kube-controller-manager rootCA miss key conf](https://github.com/openshift/cluster-kube-controller-manager-operator/issues/702)

[# issue 1472 github.com/openshift/library-go  missing key configmap](https://github.com/openshift/library-go/issues/1472)

### DIFF
--- a/pkg/operator/resourcesynccontroller/core.go
+++ b/pkg/operator/resourcesynccontroller/core.go
@@ -18,9 +18,6 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
-			continue
-		}
 		if err != nil {
 			return nil, err
 		}


lance5890:
@ingvagabund @atiratree @deads2k @soltysh

lance5890:
@tkashem

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -18,9 +18,6 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
 	certificates := []*x509.Certificate{}
 	for _, input := range inputConfigMaps {
 		inputConfigMap, err := lister.ConfigMaps(input.Namespace).Get(input.Name)
-		if apierrors.IsNotFound(err) {
-			continue
-		}
```
ingvagabund, line 23: The configmap may be present but with empty "ca-bundle.crt" data key. Which will end up in the same situation as non-existing CM.

@deads2k @mfojtik given you created/approved the PR introducing the function (https://github.com/openshift/library-go/pull/191). Was the assumption here either of the CMs can be ignored if not present? The current implementation gives impression of AnyOf rather than AllOf.

PR #1477, opened 2023-03-02
Fix key migration never completing in encryption tests

dgrisonnet:
There is a bug in the TestEncryptionTypeAESCBC test where the first
migration will never be counted as completed.
This is because the test checks the completion of the migration based on
the fact that the migrated resources for the previous key are the same
as the migrated resources of the current (desired) key. However, when
going from no key to aescbc for example, the first key doesn't have any
migrated resources since it didn't encrypt anything and as such the test
becomes stuck until a new migration is requested which in most test
cases never happens.


dinhxuanvu (commented):
/lgtm

PR #1481, opened 2023-03-02
Use leases for leader election

tkashem:
more info here: https://github.com/kubernetes/kubernetes/issues/107454


soltysh (approved):
/lgtm
/approve

PR #1487, opened 2023-03-14
make leader election option "LeaseDuration/RenewDeadline/RetryPeriod" configurable

ldpliu:
In this [pr](https://github.com/openshift/library-go/pull/1104/files), we set the default value for `LeaseDuration/RenewDeadline/RetryPeriod`. 
But sometimes, the default config is not suitable for every component, so we should make these options are configurable.

### DIFF
--- a/pkg/config/leaderelection/leaderelection.go
+++ b/pkg/config/leaderelection/leaderelection.go
@@ -3,6 +3,7 @@ package leaderelection
 import (
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"strings"
 	"time"
@@ -116,6 +117,16 @@ func LeaderElectionDefaulting(config configv1.LeaderElection, defaultNamespace, 
 	if ret.RetryPeriod.Duration == 0 {
 		ret.RetryPeriod.Duration = 26 * time.Second
 	}
+
+	retryTimes := int(math.Floor(float64(ret.RenewDeadline.Duration / ret.RetryPeriod.Duration)))
+	klog.Infof("The leader election gives %v retries and allows for %v of clock skew. The kube-apiserver downtime tolerance is %vs. Worst non-graceful lease acquisition is %v. Worst graceful lease acquisition is %v.",
+		retryTimes,
+		ret.LeaseDuration.Duration-ret.RenewDeadline.Duration,
+		(retryTimes-1)*(int(ret.RetryPeriod.Duration.Seconds())),
+		ret.LeaseDuration.Duration+ret.RetryPeriod.Duration,
+		ret.RetryPeriod,
+	)
+
 	if len(ret.Namespace) == 0 {
 		if len(defaultNamespace) > 0 {
 			ret.Namespace = defaultNamespace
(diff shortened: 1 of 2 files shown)


xuezhaojun:
> 

I'm a little concerned about adding these logs, because the calculation comment in the origin default catches a very specific case where the kube-apiserver can have 60s network disruption and the user of `library-go` want to tolerate it.

The reason why we want to customize these 3 parameters is we have a **different** case where we care about the load that the leader election process created than disruption tolerance. And there will be other cases cares about other aspects in the future.

So from a user's perspective, the logs of kube-apiserver network issue tolerance, graceful lease duration and so are not very useful information and a little confusing.

xuezhaojun:
> I'm ok with this in principle, but I'd like to see a klog statement that prints out the calculation that appears in this comment in the original default:
> 
> 1. clock skew tolerance is leaseDuration-renewDeadline == ??s
> 2. kube-apiserver downtime tolerance is == ??s
>    lastRetry=floor(renewDeadline/retryPeriod)*retryPeriod == ??s
>    downtimeTolerance = lastRetry-retryPeriod == ??s
> 3. worst non-graceful lease acquisition is leaseDuration+retryPeriod == ??s
> 4. worst graceful lease acquisition is retryPeriod == 26s
> 5. this gives 107/26=4 retries and allows for 137-107=30 seconds of clock skew
>    if the kube-apiserver is unavailable for 60s starting just before t=26 (the first renew),
>    then we will retry on 26s intervals until t=104 (kube-apiserver came back up at 86), and there will
>    be 33 seconds of extra time before the lease is lost.
> 
> where the numbers are correct for the values that are provided. It's a long log line (or multiple log lines), but it really helps explain what will happen at runtime.

I'm a little concerned about adding these logs, because the calculation comment in the origin default catches a very specific case where the kube-apiserver can have 60s network disruption and the user of `library-go` want to tolerate it.

The reason why we want to customize these 3 parameters is we have a **different** case where we care about the load that the leader election process created than disruption tolerance. And there will be other cases cares about other aspects in the future.

So from a user's perspective, the logs of kube-apiserver network issue tolerance, graceful lease duration and so are not very useful information and a little confusing.

xuezhaojun:
Also the "clock skew tolerance" part seems not calculated correctly in `library-go`, from the `leader-election` pkg comment: https://github.com/kubernetes/client-go/blob/1517ffb8d37c99e6a3a2842bcdee0aa271f0332b/tools/leaderelection/leaderelection.go#L27 It should be tolerant to arbitrary clock skew.

deads2k:
I think the calculations can be done independent of the desired tolerance.  The tolerance is an output of the calculations.

deads2k:
> It should be tolerant to arbitrary clock skew.

Only if it remains constant.  When skew shifts (the rate it refers to), it fails.

xuezhaojun:
> tolerance

Agree， but tolerance is not everyone's focus,  for example, in the case "load" matters, I would more like to see `loadPerHour = 3600/retryPeriod` included in the calculation result.

PR #1489, opened 2023-03-22
OCPBUGS-8691: Allow CSI driver operators to override nodeSelector on HyperShift

jsafrane:
`ControlPlaneTopologyHook` is currently always the last in CSIControllerSet and sets `nodeSelector=nil` on HyperShift. This does not allow CSI driver operators that run in Hypershift's hosted control plane to set their own `nodeSelector` that would match HostedControlPlane CR.

Inspired by https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/93

Set the hook as the first one, so subsequent hooks can add their own node selector.


bertinatto (commented):
/lgtm

jsafrane:
TestClusterOperatorLoggingController flaked
/retest

PR #1490, opened 2023-03-22
add ability to detect vsphere csi migration

elmiko:
Due to changes in upstream kubernetes, the vSphere in-tree cloud provider will need to be used in situations where the in-tree storage is still in use. Because OpenShift 4.13 is enabling CSI migration by default for vSphere, there will need to be a regression method for the CCM as well. This change transforms the IsCloudProviderExternal function to detect this condition.

references:
https://groups.google.com/g/kubernetes-sig-storage/c/wvJpNDBS9vU 
https://github.com/kubernetes/kubernetes/pull/116342 
https://issues.redhat.com/browse/STOR-1265

### DIFF
--- a/pkg/cloudprovider/external.go
+++ b/pkg/cloudprovider/external.go
@@ -22,6 +22,11 @@ func IsCloudProviderExternal(platformStatus *configv1.PlatformStatus, featureGat
 		return false, fmt.Errorf("platformStatus is required")
 	}
 	switch platformStatus.Type {
+	case configv1.VSpherePlatformType:
+		// vSphere has a special condition whereby CCM reversion must be possible in 4.13 OpenShift if the in-tree storage driver is enabled
+		disabled, err := isCSIMigrationvSphereDisabled(featureGate)
+		// we only want to use the in-tree cloud-provider when the migration is also disabled
+		return !disabled, err
 	case configv1.GCPPlatformType:
 		// Platforms that are external based on feature gate presence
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/external.go
```diff
@@ -48,18 +52,53 @@ func isExternalFeatureGateEnabled(featureGate *configv1.FeatureGate) (bool, erro
...
+		// If no featureGate is present, then the user hasn't opted in to the in-tree vSphere driver
+		return false, nil
+	}
+
+	enabledFeatureGates, disabledFeatureGates, err := getEnabledDisabledFeatureGates(featureGate)
+	if err != nil {
+		return false, err
+	}
+
+	return !disabledFeatureGates.Has(configv1.InTreeVSphereVolumes) && enabledFeatureGates.Has(configv1.InTreeVSphereVolumes), nil
```
dobsonj, line 85: `InTreeVSphereVolumes` is the FeatureSet that disables the `CSIMigrationvSphere` FeatureGate [here](https://github.com/openshift/api/blob/ead9036371d0f77aff3dc6f3cdbaba081a774626/config/v1/types_feature.go#L132), so this line should check for `CSIMigrationvSphere` instead.

elmiko: ack, thanks!

elmiko: so, just to be clear here, the `CSIMigrationvSphere` needs to be in the disabled list?

PR #1491, opened 2023-03-23
OCPBUGS-10887: usercache: fix race when groups are removed from user informer

mfojtik:


stlaz:
/retitle OCPBUGS-4482: usercache: fix race when groups are removed from user informer

PR #1496, opened 2023-04-06
OCPCLOUD-2010: Add external platform type support in IsCloudProviderExternal

adriengentil:
This change enables the external platform when external platform type is
set and when cloud controller manager mode is set to External.

Since cloud controller manager mode is controlled from the
`PlatformSpec` struct, I had to change the signature of
`IsCloudProviderExternal` to take the `Infrastructure` struct in
parameter.

PR on API side: https://github.com/openshift/api/pull/1434


pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go
```diff
@@ -194,6 +194,7 @@ func GetPlatformName(platformType configv1.PlatformType, recorder events.Recorde
 	case configv1.KubevirtPlatformType:
 	case configv1.AlibabaCloudPlatformType:
 	case configv1.PowerVSPlatformType:
+	case configv1.ExternalPlatformType:
```
adriengentil, line 197: not sure if it's relevant?

pkg/cloudprovider/external.go
```diff
@@ -35,12 +36,27 @@ func IsCloudProviderExternal(platformStatus *configv1.PlatformStatus, featureGat
...
 		return true, nil
+	case configv1.ExternalPlatformType:
+		return isExternalPlatformEnabled(infrastructure)
 	default:
 		// Platforms that do not have external cloud providers implemented
 		return false, nil
 	}
 }
 
+func isExternalPlatformEnabled(infrastructure *configv1.Infrastructure) (bool, error) {
```
elmiko, line 47: since this is also checking the ccm status perhaps we could name it `isExternalPlatformCCMEnabled`, wdyt?

elmiko (commented):
i think this generally looks good, i'd like to talk with @JoelSpeed about changing the way that `IsCloudProviderExternal` works, and if we should persist some of the external platform information into the status field.

pkg/cloudprovider/external.go
adriengentil: it makes sense, I'll update it :+1:

JoelSpeed (commented):
Changing signatures is fun, have we searched through OpenShift to see how many usages there are of this function? I know of at least CCMO, KCMO, KASO and MCO usage, how many other places will need fixing?

adriengentil:
> Changing signatures is fun, have we searched through OpenShift to see how many usages there are of this function? I know of at least CCMO, KCMO, KASO and MCO usage, how many other places will need fixing?

I see usages in MCO, KCMO and CCMO:
https://github.com/search?q=org%3Aopenshift+IsCloudProviderExternal+NOT+repo%3Aopenshift%2Flibrary-go+language%3AGo&type=Code&l=Go

at least for these ones, we need to re-vendor `library-go` them as part of the task. I don't it used in KASO.

elmiko (commented):
this is looking good to me
/lgtm

elmiko:
oops, just noticed it's in draft

elmiko (commented):
this is lgtm, just need the revendor after the api change lands

elmiko:
uh oh, looks like an issue with the unit test

/lgtm cancel

adriengentil:
Looks like the error in the unit test is unrelated to my change :thinking: 
```{  set -o pipefail; go test  -race -json ./pkg/... | gotest2junit > /logs/artifacts/junit_report.xml
SKIP: github.com/openshift/library-go/pkg/operator/loglevel TestClusterOperatorLoggingController
make: *** [vendor/github.com/openshift/build-machinery-go/make/targets/golang/test-unit.mk:12: test-unit] Error 1```

**Comment by elmiko** [comment]
ack, thanks @adriengentil , i will check back in a bit, going to grab some food

File: pkg/cloudprovider/external.go [hunk 3]
```diff
@@ -59,6 +64,23 @@ func isAzureStackHub(platformStatus *configv1.PlatformStatus) bool {
 	return platformStatus.Azure != nil && platformStatus.Azure.CloudName == configv1.AzureStackCloud
 }
 
+func isExternalPlatformCCMEnabled(platformStatus *configv1.PlatformStatus, featureGateAccessor featuregates.FeatureGateAccess) (bool, error) {
+	featureEnabled, err := isExternalFeatureGateEnabled(featureGateAccessor, ExternalCloudProviderFeature, ExternalCloudProviderFeatureExternal)
+	if err != nil || !featureEnabled {
+		return featureEnabled, err
+	}
+
+	if platformStatus.External == nil {
```
soltysh, line 73: Can this be:
```golang
if platformStatus == nil || platformStatus.External == nil {
```

soltysh (commented):
One nit and you're good to go

pkg/cloudprovider/external.go
adriengentil: Done!

soltysh (approved):
/lgtm
/approve

PR #1498, opened 2023-04-14
Add a flag to StatusController to remove unknown versions

jsafrane:
Allow StatusController to remove unknown versions. It's up to the caller to ensure that provided `VersionGetter` has all versions during the first `Sync()`.

Related to https://issues.redhat.com/browse/OCPBUGS-11359

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -48,6 +48,8 @@ type StatusSyncer struct {
 	controllerFactory *factory.Factory
 	recorder          events.Recorder
 	degradedInertia   Inertia
+
+	removeUnusedVersions bool
 }
 
 var _ factory.Controller = &StatusSyncer{}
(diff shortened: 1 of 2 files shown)


jsafrane:
/assign @deads2k 
you added this todo comment: https://github.com/openshift/library-go/blob/a704a5774e20501fcc70fed8885034fabc266ad7/pkg/operator/status/status_controller.go#L197-L205

deads2k:
/approve

Be careful when you use this to avoid flapping.

jsafrane:
cluster-storage-operator has just one item in `versions`, so it should be fine for us. I understand it may be dangerous for other operators.

pkg/operator/status/status_controller.go
```diff
@@ -217,6 +219,32 @@ func (c StatusSyncer) Sync(ctx context.Context, syncCtx factory.SyncContext) err
...
 }
 
+func (c *StatusSyncer) syncStatusVersions(clusterOperatorObj *configv1.ClusterOperator, syncCtx factory.SyncContext) {
+	versions := c.versionGetter.GetVersions()
+	// Add new versions from versionGetter to status
+	for operand, version := range versions {
+		previousVersion := operatorv1helpers.SetOperandVersion(&clusterOperatorObj.Status.Versions, configv1.OperandVersion{Name: operand, Version: version})
+		if previousVersion != version {
+			// having this message will give us a marker in events when the operator updated compared to when the operand is updated
+			syncCtx.Recorder().Eventf("OperatorVersionChanged", "clusteroperator/%s version %q changed from %q to %q", c.clusterOperatorName, operand, previousVersion, version)
```
RomanBednar, line 229: Wouldn't that create kind of a duplicate event? If we change `clusterOperatorObj.Status` here and continue sync loop, then we create `"OperatorStatusChanged"` event on (L218) which will show the version diff too?

PR #1500, opened 2023-04-14
OCPBUGS-2765: don't log jwt tokens

ibihim:
# What

Don't audit-log jwt tokens (pre-SHA'ed).

# Why

- the **kube-apiserver** logs `serviceaccounts/token`.
- The **oauth-apiserver** logs `tokenreviews` from `oauth.openshift.io`.
  - I couldn't find `tokenrequests` from `oauth.openshift.io`, but added it anyway.

# Note

I couldn't find any references to `tokenreviews` from `oauth.openshift.io`.
Should we be worried about that?

### kube-apiserver

```yaml
User: &{system:node:$NODE_NAME [system:nodes system:authenticated] map[]}
Verb: create
IsReadOnly: false
Namespace: openshift-cluster-csi-drivers
Resource: serviceaccounts
Subresource: token
Name: gcp-pd-csi-driver-operator
APIGroup: 
APIVersion: v1
IsResourceRequest: true
Path: /api/v1/namespaces/openshift-cluster-csi-drivers/serviceaccounts/gcp-pd-csi-driver-operator/token
```

### oauth-apiserver

```yaml
User: &{system:serviceaccount:openshift-oauth-apiserver:openshift-authenticator  [system:authenticated] map[]}
Verb: create
IsReadOnly: false
Namespace: 
Resource: tokenreviews
Subresource: 
Name: 
APIGroup: oauth.openshift.io
APIVersion: v1
IsResourceRequest: true
Path: /apis/oauth.openshift.io/v1/tokenreviews
```

### DIFF
--- a/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
+++ b/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
@@ -3,10 +3,10 @@
   resources:
   - group: "route.openshift.io"
     resources: ["routes", "routes/status"]
-  - resources: ["secrets"]
+  - resources: ["secrets", "serviceaccounts/token"]
   - group: "authentication.k8s.io"
     resources: ["tokenreviews", "tokenrequests"]
   - group: "oauth.openshift.io"
-    resources: ["oauthclients"]
+    resources: ["oauthclients", "tokenreviews"]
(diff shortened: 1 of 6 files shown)


ibihim:
PTAL @openshift/openshift-team-auth

stlaz (changes requested):

PR #1504, opened 2023-04-24
tokenrequest: add webconsoleurl to basicauth challenger

stlaz:
This is to sync the lib-go code with what's in oc.

xref https://github.com/openshift/oc/pull/991

### DIFF
--- a/pkg/oauth/tokenrequest/challengehandlers/basicauth.go
+++ b/pkg/oauth/tokenrequest/challengehandlers/basicauth.go
@@ -26,7 +26,7 @@ func NewBasicAuthNoUsernameError() error {
 }
 
 func NewBasicChallengeHandler(
-	host string,
+	host, consoleURL string,
 	reader io.Reader,
 	writer io.Writer,
 	passwordPrompter PasswordPrompter,
@@ -34,6 +34,7 @@ func NewBasicChallengeHandler(
 ) *BasicChallengeHandler {
 	return &BasicChallengeHandler{
 		Host:             host,
+		WebConsoleURL:    consoleURL,
 		Reader:           reader,
 		Writer:           writer,
 		passwordPrompter: passwordPrompter,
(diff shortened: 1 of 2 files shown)


pkg/oauth/tokenrequest/challengehandlers/basicauth.go
```diff
@@ -107,6 +111,10 @@ func (c *BasicChallengeHandler) HandleChallenge(requestURL string, headers http.
 			w = os.Stdout
 		}
 
+		if len(c.WebConsoleURL) > 0 {
+			fmt.Fprintf(w, "Console URL: %s/console\n", c.WebConsoleURL)
```
ardaguclu, line 115: In oc, `Host` field is used for printing this message https://github.com/openshift/oc/blob/a65a29a57e4d213df00e2ee1fa3641dc19912e3b/pkg/helpers/tokencmd/basicauth.go#L97 and as far as I can see `Host` is already defined. Is this new field `WebConsoleURL` different than `Host`?

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -149,15 +149,15 @@ func TestRequestToken(t *testing.T) {
 
 		// Prompting basic handler
 		"prompting basic handler, no challenge, success": {
-			Handler: challengehandlers.NewBasicChallengeHandler("", bytes.NewBufferString("mypassword\n"), nil, &testPasswordPrompter{}, "myuser", ""),
+			Handler: challengehandlers.NewBasicChallengeHandler("", "", bytes.NewBufferString("mypassword\n"), nil, &testPasswordPrompter{}, "myuser", ""),
```
ardaguclu, line 152: The reason of not adding a test case for `WebConsoleURL` is that we are not injecting`writer` in unit tests?

stlaz: yes, we don't test the output here

pkg/oauth/tokenrequest/challengehandlers/basicauth.go
stlaz: the `Host` field is used along with an addition of the `/console` path. The code you refer to supplies the whole ClusterVersionGetter to the challengers, and copy-pastes all the logic.

My proposal is to do the cluster version eval once _outside_ the challengers, and these would only print the console URL if supplied.

As such, I am replacing the clusterVersionGetter with just a string

PR #1510, opened 2023-04-26
update for more precise types

deads2k:
update for https://github.com/openshift/api/pull/1448 so we don't get a vendoring trap.


JoelSpeed (commented):
Once the API is repinned, LGTM

PR #1515, opened 2023-05-01
tweak rendering reader for easier consumption

deads2k:
need to hold for proof.

/hold


deads2k:
proof installs

/hold cancel

PR #1518, opened 2023-05-02
[WIP] Add secret monitor for Routes

thejasn:
Related: [CFE-866](https://issues.redhat.com//browse/CFE-866)
Implements: https://github.com/openshift/enhancements/pull/1307


p0lyn0mial:
Hi @thejasn, I would like to get some context before I start reviewing this PR (or reading the KEP). 
Could you please explain to me why we need this functionality in the library-go ?

pkg/route/monitor/monitor.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+type objectKey struct {
+	namespace string
+	name      string
+	uid       types.UID
+}
+
+type Object struct {
+	RefCount int
```
deads2k, line 17: This seems more suitable as a property tracked by the caller.

pkg/route/monitor/monitor.go
```diff
@@ -0,0 +1,78 @@
...
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/cache"
+)
+
+type objectKey struct {
+	namespace string
+	name      string
+	uid       types.UID
```
deads2k, line 13: top-to-bottom, I'm surprised to see UID.

pkg/route/secret/secret_monitor.go
```diff
@@ -0,0 +1,211 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/library-go/pkg/route/monitor"
+)
+
+type listObjectFunc func(string, metav1.ListOptions) (runtime.Object, error)
+type watchObjectFunc func(string, metav1.ListOptions) (watch.Interface, error)
+
+// Monitor manages Kubernetes secrets. This includes retrieving
+// secrets or registering/unregistering them via Routes.
+type Monitor interface {
```
deads2k, line 33: I think we need something like

```go
type SecretEventHandlerRegistration interface{
  ResourceEventHandlerRegistration

  UID() string
}

type SecretMonitor interface{
  AddEventHandler(namespace, name string, handler ResourceEventHandler) (SecretEventHandlerRegistration, error)
  
  RemoveEventHandler(SecretEventHandlerRegistration) error
}
```

This will allow an eventHandler for every route (overlapping secrets).  It's not perfectly efficient, but it is easy.  And handler passed in could wire back to a route to be queued in a different controller.  Need a link to your calling code to see how to wire that back in.

We might also need a way to get a list of all known eventHandlers to diff against the current state (re-list after an outage), but if the caller correctly handles deletes we may not need it.  It is a potential spot for a leak.

deads2k: See the 1.27 SharedInformer interface for where I got inspiration and how can you wire it up through a sharedinformer versus a regular informer.

thejasn: So I understand your suggestion as, we need an abstraction between the secret watching logic and And the caller (the component that interfaces with Routes) in form of (do correct me if I understood it incorrectly)
```go
type SecretMonitor interface{
  AddEventHandler(namespace, name string, handler ResourceEventHandler) (SecretEventHandlerRegistration, error)
  
  RemoveEventHandler(SecretEventHandlerRegistration) error
}
```

> See the 1.27 SharedInformer interface for where I got inspiration and how can you wire it up through a sharedinformer versus a regular informer.

Are you also suggesting I replace the use of the regular informer in the implementation with a sharedinformer?

pkg/route/monitor/monitor.go
thejasn: Yeah I would agree, will rework this bit.

deads2k:
opened https://github.com/openshift/library-go/pull/1554 to adjust the secretmonitor.  see if it makes sense.

PR #1527, opened 2023-05-14
change missing static pod controller by latest installer revision

lance5890:
[issue 1055 static pod controller degraded wrong](https://github.com/openshift/cluster-etcd-operator/issues/1055)


lance5890:
@Elbehery @tjungblu

PR #1529, opened 2023-05-16
IR-371: upgrade distribution

flavianmissi:
I'm currently upgrading openshift/image-registry to distribution v3, and because openshift/image-registry uses types from library-go, we need to upgrade to distribution v3 here as well.


flavianmissi:
Hi @soltysh, in a private conversation @dmage mentioned that we might want to investigate upgrading distribution in the kubernetes repository before we upgrade it here.

The thing is that v3 is not an official distribution release yet, so I'm thinking it might be hard to sell the upgrade in the kubernetes repo. The only reason we are upgrading library-go (and more importantly openshift/image-registry) is because we want a feature that is on the main branch and it doesn't really make sense to backport (apart from it being difficult).

The kubernetes use of distribution is for image references only, and that code has not changed in any relevant way in the last 8 years.

Long story short, I think we are okay with kubernetes using v2 and us v3.

Can you help us make a more informed decision?
Thanks!

soltysh (commented):
The changes in this PR lgtm, but since this is being used at minimum by:
- openshift-apiserver
- oc
before merging this I'd like to see teams owning these repos to confirm these changes

dinhxuanvu (commented):
/lgtm

dinhxuanvu (commented):
/lgtm

flavianmissi:
Hi @soltysh, we have gotten LGTMs from oc and openshift-apiserver folks so we're good to proceed with reviews now.
Thanks!

soltysh (approved):
/approve

PR #1532, opened 2023-05-18
notice the notFoundErr in resourcesynccontroller

lance5890:
**1 bug description**
when resourcesynccontroller miss key configmap in CombineCABundleConfigMaps, and err is notFound, it will continue and ignore this failure, this will cause the caller func generate wrong configurations

**2 related issues**:

**more information, please read the following issues**
[# issue 702 openshift/cluster-kube-controller-manager-operator kube-controller-manager rootCA miss key conf](https://github.com/openshift/cluster-kube-controller-manager-operator/issues/702)

[# issue 1472 github.com/openshift/library-go  missing key configmap](https://github.com/openshift/library-go/issues/1472)

releted to CKA-o rollout too much times
https://github.com/openshift/cluster-kube-apiserver-operator/issues/1504

https://github.com/openshift/cluster-kube-apiserver-operator/issues/1509

[# closed PR 1473](https://github.com/openshift/library-go/pull/1473)

### DIFF
--- a/pkg/operator/resourcesynccontroller/core.go
+++ b/pkg/operator/resourcesynccontroller/core.go
@@ -6,7 +6,6 @@ import (
 	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/cert"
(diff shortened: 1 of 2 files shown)


lance5890:
@ingvagabund @atiratree @deads2k @soltysh

lance5890:
The configmap may be present but with empty "ca-bundle.crt" data key. Which will end up in the same situation as non-existing CM.

@deads2k @mfojtik given you created/approved the PR introducing the function (https://github.com/openshift/library-go/pull/191). Was the assumption here either of the CMs can be ignored if not present? The current implementation gives impression of AnyOf can be present rather than AllOf.

PR #1534, opened 2023-05-30
WRKLDS-728: controller/apiservice: allow to disable/delete an API service resources

ingvagabund:
SSIA

Needed by https://github.com/openshift/cluster-openshift-apiserver-operator/pull/532

### DIFF
--- a/pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
+++ b/pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
@@ -7,13 +7,16 @@ import (
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/errors"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	apiregistrationv1client "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/typed/apiregistration/v1"
 	apiregistrationinformers "k8s.io/kube-aggregator/pkg/client/informers/externalversions"
+	apiregistrationv1lister "k8s.io/kube-aggregator/pkg/client/listers/apiregistration/v1"
 
 	operatorsv1 "github.com/openshift/api/operator/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -23,17 +26,21 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
-type GetAPIServicesToMangeFunc func() ([]*apiregistrationv1.APIService, error)
+// GetAPIServicesToMangeFunc provides list of enabled and disabled managed APIService items.
+// Both lists need to always contain all the managed APIServices so the controller
+// can avoid reconciling user-created/unmanaged objects.
+type GetAPIServicesToMangeFunc func() (enabled []*apiregistrationv1.APIService, disabled []*apiregistrationv1.APIService, err error)
 type apiServicesPreconditionFuncType func([]*apiregistrationv1.APIService) (bool, error)
 
 type APIServiceController struct {
 	getAPIServicesToManageFn GetAPIServicesToMangeFunc
-	// precondition must return true before the apiservices will be created
-	precondition apiServicesPreconditionFuncType
+	// preconditionForEnabledAPIServices must return true before the apiservices will be created
+	preconditionForEnabledAPIServices apiServicesPreconditionFuncType
 
 	operatorClient          v1helpers.OperatorClient
 	kubeClient              kubernetes.Interface
 	apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface
+	apiservicelister        apiregistrationv1lister.APIServiceLister
 }
 
 func NewAPIServiceController(
@@ -47,11 +54,12 @@ func NewAPIServiceController(
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &APIServiceController{
-		precondition:             newEndpointPrecondition(kubeInformersForOperandNamespace),
-		getAPIServicesToManageFn: getAPIServicesToManageFunc,
+		preconditionForEnabledAPIServices: newEndpointPrecondition(kubeInformersForOperandNamespace),
+		getAPIServicesToManageFn:          getAPIServicesToManageFunc,
 
 		operatorClient:          operatorClient,
 		apiregistrationv1Client: apiregistrationv1Client,
+		apiservicelister:        apiregistrationInformers.Apiregistration().V1().APIServices().Lister(),
 		kubeClient:              kubeClient,
 	}
 
@@ -60,6 +68,86 @@ func NewAPIServiceController(
 		kubeInformersForOperandNamespace.Core().V1().Endpoints().Informer(),
 		apiregistrationInformers.Apiregistration().V1().APIServices().Informer(),
 	).ToController("APIServiceController_"+name, eventRecorder.WithComponentSuffix("apiservice-"+name+"-controller"))
+}
+
+func (c *APIServiceController) updateOperatorStatus(
+	ctx context.Context,
+	syncDisabledAPIServicesErr error,
+	preconditionReadyErr error,
+	preconditionsReady bool,
+	syncEnabledAPIServicesErr error,
+) (err error) {
+	errs := []error{}
+	conditionAPIServicesDegraded := operatorv1.OperatorCondition{
+		Type:   "APIServicesDegraded",
+		Status: operatorv1.ConditionFalse,
+	}
+	conditionAPIServicesAvailable := operatorv1.OperatorCondition{
+		Type:   "APIServicesAvailable",
+		Status: operatorv1.ConditionTrue,
+	}
+
+	if syncDisabledAPIServicesErr != nil || preconditionReadyErr != nil || syncEnabledAPIServicesErr != nil {
+		// a closed context indicates that the process has been requested to shutdown
+		// in that case we might have failed to check availability of the downstream servers due to the context being closed
+		// in that case don't report the failure to avoid false positives and changing the condition of the operator
+		// the next process will perform the checks immediately after the startup
+		select {
+		case <-ctx.Done():
+			if syncDisabledAPIServicesErr != nil {
+				errs = append(errs, fmt.Errorf("failed to delete disabled APIs: %v", syncDisabledAPIServicesErr))
+			}
+			if preconditionReadyErr != nil {
+				errs = append(errs, fmt.Errorf("failed to check precondition for enabled APIs: %v", preconditionReadyErr))
+			}
+			if syncEnabledAPIServicesErr != nil {
+				errs = append(errs, fmt.Errorf("failed to reconcile enabled APIs: %v", syncEnabledAPIServicesErr))
+			}
+			nerr := fmt.Errorf("the operator is shutting down, skipping updating conditions, err = %v", errors.NewAggregate(errs))
+			return nerr
+		default:
+		}
+	}
+
+	defer func() {
+		updates := []v1helpers.UpdateStatusFunc{
+			v1helpers.UpdateConditionFn(conditionAPIServicesDegraded),
+			v1helpers.UpdateConditionFn(conditionAPIServicesAvailable),
+		}
+
+		if _, _, updateError := v1helpers.UpdateStatus(ctx, c.operatorClient, updates...); updateError != nil {
+			// overrides error returned through 'return <ERROR>' statement
+			err = updateError
+		}
+	}()
+
+	if syncDisabledAPIServicesErr != nil {
+		conditionAPIServicesDegraded.Status = operatorv1.ConditionTrue
+		conditionAPIServicesDegraded.Reason = "DisabledAPIServicesPresent"
+		conditionAPIServicesDegraded.Message = syncDisabledAPIServicesErr.Error()
+		errs = append(errs, syncDisabledAPIServicesErr)
+	}
+
+	if preconditionReadyErr != nil {
+		conditionAPIServicesAvailable.Status = operatorv1.ConditionFalse
+		conditionAPIServicesAvailable.Reason = "ErrorCheckingPrecondition"
+		conditionAPIServicesAvailable.Message = preconditionReadyErr.Error()
+		errs = append(errs, preconditionReadyErr)
+	} else if !preconditionsReady {
+		conditionAPIServicesAvailable.Status = operatorv1.ConditionFalse
+		conditionAPIServicesAvailable.Reason = "PreconditionNotReady"
+		conditionAPIServicesAvailable.Message = "PreconditionNotReady"
+		return errors.NewAggregate(errs)
+	}
+
+	if syncEnabledAPIServicesErr != nil {
+		conditionAPIServicesAvailable.Status = operatorv1.ConditionFalse
+		conditionAPIServicesAvailable.Reason = "Error"
+		conditionAPIServicesAvailable.Message = syncEnabledAPIServicesErr.Error()
+		return errors.NewAggregate(append(errs, syncEnabledAPIServicesErr))
+	}
+
+	return errors.NewAggregate(errs)
 }
 
 func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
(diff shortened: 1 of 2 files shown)


pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -23,8 +24,20 @@ import (
...
+	// Any disabled APIService is removed from a cluster.
+	Enabled bool
+	// APIService stores an actual APIService
+	APIService *apiregistrationv1.APIService
+}
+
+// GetAPIServicesToMangeFunc provides a full list of managed APIService items.
+// The list needs to always contain all the managed APIServices so only managed
+// objects are removed without removing user-created/unmanaged objects.
+type GetAPIServicesToMangeFunc func() ([]ManagedAPIService, error)
```
p0lyn0mial, line 39: why this function couldn't return services that are enabled :) ?

p0lyn0mial:
also the title of this pr is a bit misleading for me, are we actually going to delete something?

pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -25,11 +25,14 @@ func newEndpointPrecondition(kubeInformers kubeinformers.SharedInformerFactory)
...
 
-	return func(apiServices []*apiregistrationv1.APIService) (bool, error) {
+	return func(apiServices []ManagedAPIService) (bool, error) {
 
 		coordinates := []coordinate{}
-		for _, apiService := range apiServices {
-			curr := coordinate{namespace: apiService.Spec.Service.Namespace, name: apiService.Spec.Service.Name}
+		for _, managedAPIService := range apiServices {
+			if !managedAPIService.Enabled {
+				continue
```
p0lyn0mial, line 33: since we are skipping disabled services I would simply return only enabled services in `GetAPIServicesToMangeFunc`

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: how will you know if an api service is not enabled ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -23,8 +24,20 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
-type GetAPIServicesToMangeFunc func() ([]*apiregistrationv1.APIService, error)
-type apiServicesPreconditionFuncType func([]*apiregistrationv1.APIService) (bool, error)
+// ManagedAPIService keeps expected state of a single APIService object
+type ManagedAPIService struct {
+	// Enabled signals whether an APIService is enabled.
+	// Any disabled APIService is removed from a cluster.
+	Enabled bool
```
p0lyn0mial, line 31: will this field be used by all clients or only for some?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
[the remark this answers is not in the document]
p0lyn0mial: or changing the signature of this method to return `enabled []*apiregistrationv1.APIService, disabled []*apiregistrationv1.APIService, error)`

ingvagabund: > why this function couldn't return services that are enabled :) ?

The controller needs a list of APIServices that are disabled. So they can be deleted if present.

> how will you know if an api service is not enabled ?

Configurable by anyone who consumes the APIService controller. Atm, the OA operator constructs the list based on enabled capabilities.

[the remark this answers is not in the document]
ingvagabund: That would be an alternative. In any case, still a breaking change.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Once such struct is created, it's intended to be used only by the controller.

pkg/operator/apiserver/controller/apiservice/apigroup.go
ingvagabund: Both `newEndpointPrecondition` and `checkDiscoveryForByAPIServices` can be probably given only a list of enabled services.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: An alternative would be to provide a filter function that if not provide will include every api service. On the other hand such a filter will get invoked all over again instead of constructing a list of enabled/disabled service only once at the beginning.

p0lyn0mial: OK, since there are only two operators using this controller I would:

1. Change the signature of the method to `GetAPIServicesToMangeFunc func() (enabled []*apiregistrationv1.APIService, disabled []*apiregistrationv1.APIService, error)`
2. Rename `precondition` to `preconditionForEnabledAPIServices`
3. Split the `sync` method into two flows, one for reconciling `enabled` services and the second for reconciling `disabled` services.

Does it make sense ?

ingvagabund: Done. It looks better now.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -118,7 +122,7 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
 		return err
 	}
 
-	err = c.syncAPIServices(ctx, apiServices, syncCtx.Recorder())
+	err = c.syncAPIServices(ctx, enabledApiServices, disabledApiServices, syncCtx.Recorder())
```
p0lyn0mial, line 125: How about adding a separate method for reconciling `disalbedApiServices` and calling it before  `preconditionForEnabledAPIServices` (line 101)?

Having two separate methods would allow us to reconcile `enabledApiServices` and `disabledApiServices` independently. I think we always want to reconcile both and aggregate the errors. What do you think ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -152,11 +156,18 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
...
 }
 
-func (c *APIServiceController) syncAPIServices(ctx context.Context, apiServices []*apiregistrationv1.APIService, recorder events.Recorder) error {
+func (c *APIServiceController) syncAPIServices(ctx context.Context, enabledApiServices, disabledApiServices []*apiregistrationv1.APIService, recorder events.Recorder) error {
 	errs := []error{}
 	var availableConditionMessages []string
 
-	for _, apiService := range apiServices {
+	for _, apiService := range disabledApiServices {
+		if err := c.apiregistrationv1Client.APIServices().Delete(ctx, apiService.Name, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
p0lyn0mial, line 164: I think we should get an APIService from the `cache` first as that would allow us to put less pressure on the server. 
We should issue a live request only when we know the resource must be deleted.
What do you think?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: Thanks, I have added a few more comments/questions, PTAL.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -23,13 +24,16 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
-type GetAPIServicesToMangeFunc func() ([]*apiregistrationv1.APIService, error)
+// GetAPIServicesToMangeFunc provides a two list of managed APIService items.
+// Both lists need to always contain all the managed APIServices so only managed
+// objects are removed without removing user-created/unmanaged objects.
```
p0lyn0mial, line 29: what do you mean by `without removing user-created/unmanaged objects` ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Updated

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Updated

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Comment updated.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -23,17 +25,21 @@ import (
...
 	getAPIServicesToManageFn GetAPIServicesToMangeFunc
-	// precondition must return true before the apiservices will be created
-	precondition apiServicesPreconditionFuncType
+	// preconditionForEnabledAPIServices must return true before the apiservices will be created
+	preconditionForEnabledAPIServices apiServicesPreconditionFuncType
 
 	operatorClient          v1helpers.OperatorClient
 	kubeClient              kubernetes.Interface
 	apiregistrationv1Client apiregistrationv1client.ApiregistrationV1Interface
+	apiservicelister        apiregistrationv1lister.APIServiceLister
```
p0lyn0mial, line 42: isn't that strange that this controller didn't use `apiservicelister` ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -74,13 +81,17 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
...
-		apiServices, err := c.getAPIServicesToManageFn()
+		enabledApiServices, disabledApiServices, err := c.getAPIServicesToManageFn()
 		if err != nil {
 			errs = append(errs, err)
 			return errors.NewAggregate(errs)
 		}
-		for _, apiService := range apiServices {
-			if err := c.apiregistrationv1Client.APIServices().Delete(ctx, apiService.Name, metav1.DeleteOptions{}); err != nil {
+		for _, apiService := range append(enabledApiServices, disabledApiServices...) {
+			if _, err := c.apiservicelister.Get(apiService.Name); err == nil {
```
p0lyn0mial, line 90: how about simply calling `syncDisabledAPIServices` with combined `enabledApiServices, disabledApiServices` ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -152,11 +168,28 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
 	return err
 }
 
-func (c *APIServiceController) syncAPIServices(ctx context.Context, apiServices []*apiregistrationv1.APIService, recorder events.Recorder) error {
+func (c *APIServiceController) syncDisabledAPIServices(ctx context.Context, apiServices []*apiregistrationv1.APIService) error {
 	errs := []error{}
-	var availableConditionMessages []string
 
 	for _, apiService := range apiServices {
+		if _, err := c.apiservicelister.Get(apiService.Name); err == nil {
```
p0lyn0mial, line 175: should we also consider a case in which a service has been already marked for deletion but hasn't been deleted?
(I think that would require checking the `DeletionTimestamp` field)

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -90,11 +101,14 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
...
-	apiServices, err := c.getAPIServicesToManageFn()
+	enabledApiServices, disabledApiServices, err := c.getAPIServicesToManageFn()
 	if err != nil {
 		return err
 	}
-	ready, err := c.precondition(apiServices)
+
+	disErr := c.syncDisabledAPIServices(ctx, disabledApiServices)
+
+	ready, err := c.preconditionForEnabledAPIServices(enabledApiServices)
```
p0lyn0mial, line 111: what if both `syncDisabledAPIServices` and `preconditionForEnabledAPIServices` return some error?
do we want to report errors from `syncDisabledAPIServices` ?
i think the errors will be ignored as it is right now.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Done

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: That should be idempotent. Deleting an object that's in the process of deletion should be a no-op. While `Get` should still return the object without an error.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: If `syncDisabledAPIServices` returns an error, there are two cases that can happen:
- `preconditionForEnabledAPIServices` returns `err != nil` or `!ready` in which case the `APIServicesAvailable` condition gets reported as false. Making error returned by `syncDisabledAPIServices` kinda obsolete since the `APIServicesAvailable` is already false.
- `preconditionForEnabledAPIServices` returns `err == nil` and `ready` in which case the error returned by `syncDisabledAPIServices` gets checked. Producing `APIServicesAvailable` condition set to false. Just with a different error string.

p0lyn0mial: Does setting `APIServicesAvailable` when `syncDisabledAPIServices returns some errors` make sense ?

If all services were available and we failed to remove one service would you expect to find the error in `APIServicesAvailable` ? Maybe we should report it as a separate condition? (Especially that we already made distinction between enabled/disabled on the API level)

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: It is idempotent but we will hammer the API unnecessary. Does it make sense?

ingvagabund: I agree this will reduce the toll. On the other hand I wonder if checking the `DeletionTimestamp` is something we already do in other controller or whether this is a novel improvement. If the latter, I prefer to refactor the code more and implement generic methods like DeleteOnlyIfPresent. Probably automatically generated for all our APIs. So we don't re-invent the check all over the place. What do you think?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: The current implementation does not make the reported condition worse. Right now it will get reported with `Reason: Error`.  Whether that happens before or after `preconditionForEnabledAPIServices` makes no difference. Failing to remove or create an APIservice object falls under the same category of an error. If we decide to distinguish between error related to removing and creating an APIService, we should probably introduce a new condition for both `Failed to create` and `Failed to check discovery`.

ingvagabund: I wonder whether there's a value in reporting non-nil error for `return c.syncDisabledAPIServices(ctx, append(enabledApiServices, disabledApiServices...))` under:
```
	switch operatorConfigSpec.ManagementState {
	case operatorsv1.Managed:
	case operatorsv1.Unmanaged:
		return nil
	case operatorsv1.Removed:
		enabledApiServices, disabledApiServices, err := c.getAPIServicesToManageFn()
		if err != nil {
			return err
		}
		return c.syncDisabledAPIServices(ctx, append(enabledApiServices, disabledApiServices...))
```

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: I don't think it is novel approach. For example, it is being used in:

- https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authenticator/token/bootstrap/bootstrap.go#L107

- https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/serviceaccount/legacy_serviceaccount_token_cleaner.go#L147

I can try to find more examples if you want.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: As of today `APIServicesAvailable` was reserved for checking availability which was defined as checked if:
- endpoints were assigned to services
- `apiregistrationv1.Available` condition was set on services
- discovery endpoint works for services

I'm just wondering if it makes sense to report `disabled` services on the same condition.

ingvagabund: I see what you mean now. A removed API service is no longer part of what is considered as available API services.

From https://github.com/openshift/api/blob/master/operator/v1/types.go#L154-L155:
```
// Degraded indicates that the operator (not the operand) is unable to fulfill the user intent
OperatorStatusTypeDegraded = "Degraded"
```
A new condition should probably be of type `Degraded`. I.e. disabled API services which are still available needs to be removed so the user intent is fulfilled. What about `APIServicesDegraded`?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: I was wondering if we have an example in our operators. Nevertheless, checking for the `DeletionTimestamp` is easy to implement. Will do.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: I'm not sure if we want to go `Degraded`. I think that:

```
Type:    "APIServicesDeleted",
Status:  operatorv1.ConditionFalse,
Reason:  "DisabledAPIServicesPresent",
Message: err.Error(),
```

should be fine.

p0lyn0mial: Also, We could split the `APIServiceController.sync` method into two flows for simplicity. I think I would do:
```
func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
	operatorConfigSpec, _, _, err := c.operatorClient.GetOperatorState()
	if err != nil {
		return err
	}

	switch operatorConfigSpec.ManagementState {
	case operatorsv1.Managed:
	case operatorsv1.Unmanaged:
		return nil
	case operatorsv1.Removed:
		enabledApiServices, disabledApiServices, err := c.getAPIServicesToManageFn()
		if err != nil {
			return err
		}
		return c.syncDisabledAPIServices(ctx, append(enabledApiServices, disabledApiServices...))
	default:
		syncCtx.Recorder().Warningf("ManagementStateUnknown", "Unrecognized operator management state %q", operatorConfigSpec.ManagementState)
		return nil
	}

       var errors []err
       // or if we rename syncDisabledAPIServices to something else then
       // we could call it syncDisabledAPIServices (which would also set the new status)
       err = c.reconcileDisabledServices()
       if err != nil { // append to errors }
       
       err = c.syncEnabledServices()
       if err != nil { // append to errors }

      return errors
}

Comment (ingvagabund) [reply pkg/operator/apiserver/controller/apiservice/apiservice_controller.go:100]: Would `syncEnabledServices` include calling the `preconditionForEnabledAPIServices` as well?

Comment (p0lyn0mial) [reply pkg/operator/apiserver/controller/apiservice/apiservice_controller.go:100]: yes, I think so :)

Comment (p0lyn0mial) [reply pkg/operator/apiserver/controller/apiservice/apiservice_controller.go:100]: as an optimisation step we could collect `conditions` from the individual `sync*` methods and update the config just once :)

File: pkg/operator/apiserver/controller/apiservice/apiservice_controller.go [hunk 11]
```diff
@@ -143,20 +154,40 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
...
 
-func (c *APIServiceController) syncAPIServices(ctx context.Context, apiServices []*apiregistrationv1.APIService, recorder events.Recorder) error {
+func (c *APIServiceController) syncDisabledAPIServices(ctx context.Context, apiServices []*apiregistrationv1.APIService) error {
 	errs := []error{}
-	var availableConditionMessages []string
 
 	for _, apiService := range apiServices {
+		if apiServiceObj, err := c.apiservicelister.Get(apiService.Name); err == nil {
+			if apiServiceObj.DeletionTimestamp != nil {
+				continue
```
p0lyn0mial, line 172: Technically objects can stuck in deletion (finalisers) for a long time. 
Do we want to handle this case somehow (log, set something on the condition) ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -73,53 +80,57 @@ func (c *APIServiceController) sync(ctx context.Context, syncCtx factory.SyncCon
...
+	ready, err := c.preconditionForEnabledAPIServices(enabledApiServices)
 	if err != nil {
 		if _, _, updateErr := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
 			Type:    "APIServicesAvailable",
 			Status:  operatorv1.ConditionFalse,
 			Reason:  "ErrorCheckingPrecondition",
 			Message: err.Error(),
-		})); updateErr != nil {
-			return errors.NewAggregate([]error{err, updateErr})
+		}), v1helpers.UpdateConditionFn(conditionAPIServicesDegraded)); updateErr != nil {
```
p0lyn0mial, line 116: I don't want to complicate the code but it looks like it might be easy to forget which conditions should be updated. I like the way it was done in the workload controller (https://github.com/openshift/library-go/blob/c021d427e60e6cbd278412a5a5e49784f2c2144b/pkg/operator/apiserver/controller/workload/workload.go#L182). Maybe we could do something similar here?

I think it boils down to:
1. define conditions with default values
2. register an update function for conditions in the defer statement
3. when you need to end processing, just set the condition and return  (it will be updated in the defer statement).

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: `syncEnabledAPIServices` returns an error when the corresponding api service Available condition is not ready. Updated the code to return error when a corresponding service is not deleted yet.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
ingvagabund: Done. I did not find any use for `updateGenerationFn` though.

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -62,6 +69,77 @@ func NewAPIServiceController(
...
+	conditionAPIServicesDegraded := operatorv1.OperatorCondition{
+		Type:   "APIServicesDegraded",
+		Status: operatorv1.ConditionFalse,
+	}
+	conditionAPIServicesAvailable := operatorv1.OperatorCondition{
+		Type:   "APIServicesAvailable",
+		Status: operatorv1.ConditionTrue,
+	}
+
+	if syncEnabledAPIServicesErr != nil {
```
p0lyn0mial, line 89: I think we should check if for any error or always for simplicity, wdyt?

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -62,6 +69,77 @@ func NewAPIServiceController(
...
+		// a closed context indicates that the process has been requested to shutdown
+		// in that case we might have failed to check availability of the downstream servers due to the context being closed
+		// in that case don't report the failure to avoid false positives and changing the condition of the operator
+		// the next process will perform the checks immediately after the startup
+		select {
+		case <-ctx.Done():
+			nerr := fmt.Errorf("the operator is shutting down, skipping updating availability of the aggreaged APIs, err = %v", syncEnabledAPIServicesErr)
```
p0lyn0mial, line 96: just adjust comment so that it is more generic

pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
p0lyn0mial: update: we decided to log instead of returning the error

p0lyn0mial:
just one more thing, it looks like we are going to always set the new condition, even if the `toRemoveServices` was empty.

p0lyn0mial:
> just one more thing, it looks like we are going to always set the new condition, even if the toRemoveServices was empty.

it seems that setting the new condition conditionally would make the code more complicated. Besides on clusters with this feature disabled it will always be set to `false`.

/hold cancel

PR #1538, opened 2023-06-08
Add cert status info if CSR is not approved

marioferh:
If CSR is not approved for any reason, csr.Status is printed in the logs to get some information.

Even if there is just an skip for next iteration, if there is a error in the approve func it will be miss in the logs and can be hard to debug.

### DIFF
--- a/pkg/operator/csr/cert_controller.go
+++ b/pkg/operator/csr/cert_controller.go
@@ -237,13 +237,13 @@ func (c *clientCertificateController) syncCSR(secret *corev1.Secret) (map[string
 		return nil, err
 	}
 
-	// skip if csr is not approved yet
 	if !isCSRApproved(csr) {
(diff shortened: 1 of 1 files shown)


deads2k:
We use V(2) when logging, so this won't appear.  Why is this better than retrieving the CSR when gathering debugging information?  That's the purpose behind `status.relatedObjects` handling in `oc adm inspect`

marioferh:
> We use V(2) when logging, so this won't appear. Why is this better than retrieving the CSR when gathering debugging information? That's the purpose behind `status.relatedObjects` handling in `oc adm inspect`

I have seen other V(4) logs as this one: https://github.com/openshift/library-go/blob/6e4855505a2da0caa107d56cf145bd01befa88b9/pkg/operator/csr/cert_controller.go#L193

I think this is better because if CSR is not approved you get the reason in the logs.

PR #1542, opened 2023-06-14
Fix flakiness of `TestGetName` and an issue in `GetName`

mpatlasov:
The PR consist of two commits. The first fixes two independent cases of flakiness in `TestGetName`. These issues are easy to reproduce by running `go test -count=1 ./pkg/build/naming/...` several times. Commit description explains them in full details.

The second commit fixes a real issue in `GetName` itself: if the `base` arg is not super-short (>= 10 symbols) and `suffix` arg is of 243 symbol exactly, the function returns value like this: `-somthing-hash`. This is violates https://kubernetes.io/docs/concepts/overview/working-with-objects/names/: the name must start with an alphanumeric character.


soltysh (approved):
/approve

elmiko (commented):
/lgtm

PR #1545, opened 2023-06-21
♻️ remove duplicated time provider behaviour

glynternet:
`now` and `clock` fields are both present in the `InstallerController` struct for the same purpose.

This PR migrates all `now` calls to `clock.Now`.


glynternet:
@dgrisonnet
@tkashem

Reckon we can get a look at this?

glynternet:
@dgrisonnet
@tkashem

Reckon we can get a look at this?

PR #1546, opened 2023-06-21
OCPBUGS-15256: installerpod: change config map based files mode to 0600

ingvagabund:
To be more compliant with CIS benchmarks

/hold
so the change can be verified in KSO and KCMO first.

### DIFF
--- a/pkg/operator/staticpod/installerpod/cmd.go
+++ b/pkg/operator/staticpod/installerpod/cmd.go
@@ -625,7 +625,7 @@ func (o *InstallOptions) writePod(rawPodBytes []byte, manifestFileName, resource
 func writeConfig(content []byte, fullFilename string) error {
 	klog.Infof("Writing config file %q ...", fullFilename)
 
-	filePerms := os.FileMode(0644)
+	filePerms := os.FileMode(0600)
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -625,7 +625,7 @@ func (o *InstallOptions) writePod(rawPodBytes []byte, manifestFileName, resource
 func writeConfig(content []byte, fullFilename string) error {
 	klog.Infof("Writing config file %q ...", fullFilename)
 
-	filePerms := os.FileMode(0644)
+	filePerms := os.FileMode(0600)
 	if strings.HasSuffix(fullFilename, ".sh") {
 		filePerms = 0755
```
rhmdnd, line 630: Do we also want to change this to 700? If so, should we just use `writeSecret` instead?

ingvagabund: @hasbro17 I found `cluster-backup.sh` and `cluster-restore.sh` scripts getting copied from a CM. It seems save to reduce the permissions to just 0700. Though, better to double check. Do you know if both scripts can work with the limited permissions?

soltysh (approved):
/approve

PR #1549, opened 2023-07-05
CFE-882: Route external certificate validation

thejasn:
Description
---
Updates route validations based on https://github.com/openshift/enhancements/pull/1307 which introduces a new field in the route API externalCertificate behind the TP feature gate.

### Changes
- `ValidateRoute()` has been updated to pass additional args (secrets lister and subjectaccessreviewer) to validate
`externalCertificate` and the rbac required. Now it also does additional subject access review authorization checks to verify if the router SA has the correct permissions to parse `externalCertificate`. It also adds a hard requirement that the secret referenced in `externalCertificate` be present prior to creating the route.
- `ValidateHostUpdate()` has also been updated since updating route host/subdomain is also affected by updating
certificates on the route. Now `certificateChangeRequiresAuth()` will always return `true` if the route has a `externalCertificate` set (check EP for additional info).
- ValidateHostExternalCertificate() has been introduced as part of the validations done during route updates, this function specifically checks if the user has the correct permissions on the `custom-host` sub-resource. Additional details can be found on the EP.


pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2027,162 @@ func TestValidateTLS(t *testing.T) {
...
+						Termination: routev1.TLSTerminationEdge,
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "tls-secret",
+						},
+					},
+				},
+			},
+			allow:          false,
+			validType:      true,
+			validNS:        true,
```
alebedev87, line 2114: Why not just `corev1.Secret` as a test input? That would be more intuitive. Because now the reader has to take a look at the test code to see how these inputs are transformed into the certificate secret.

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2027,162 @@ func TestValidateTLS(t *testing.T) {
...
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "tls-secret",
+						},
+					},
+				},
+			},
+			allow:          true,
+			validNS:        true,
+			validType:      false,
+			opts:           routecommon.RouteValidationOptions{AllowExternalCertificates: true},
```
alebedev87, line 2159: I don't see a non error test case for `validateTLS` where the certificate secret would be valid and the external certificate is allowed.

pkg/route/hostassignment/assignment.go
```diff
@@ -190,6 +215,10 @@ func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc S
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Certificate, older.Spec.TLS.Certificate, field.NewPath("spec", "tls", "certificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.DestinationCACertificate, older.Spec.TLS.DestinationCACertificate, field.NewPath("spec", "tls", "destinationCACertificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Key, older.Spec.TLS.Key, field.NewPath("spec", "tls", "key"))...)
+
+			if opts.AllowExternalCertificates && route.Spec.TLS.ExternalCertificate != nil && older.Spec.TLS.ExternalCertificate != nil {
+				errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "certificate"))...)
```
alebedev87, line 220:
```suggestion
				errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"))...)
```

pkg/route/hostassignment/assignment_test.go
```diff
@@ -52,6 +53,13 @@ func TestHostWithWildcardPolicies(t *testing.T) {
 		expected          string
 		expectedSubdomain string
 
+		// fields for externalCertificate
+		validType  bool
+		secretData map[string]string
+		validNS    bool
```
alebedev87, line 59: Same as in `validation` package. What do you think about providing `corev1.Secret` as a test case input?

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +223,62 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "certificate-changed-from-external-certificate",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
```
alebedev87, line 246:
```suggestion
			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
```
Just to keep the same ordering as other tests. Otherwise it's a little misleading as you read the test name and then don't see `ExternalCertificate` as current TLS.

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,36 @@
...
+
+	routev1 "github.com/openshift/api/route/v1"
+	routecommon "github.com/openshift/library-go/pkg/route"
+)
+
+// validateHostExternalCertificate checks if the user has permissions to create and update
+// custom-host subresource of routes. This check is required to be done prior to ValidateHostUpdate()
+// since updating hosts while using externalCertificate is contingent on the user having both these
+// permissions. The ValidateHostUpdate() cannot differentiate if the certificate has changed since
+// now the certificates will be present as a secret object, due to this it proceeds with the assumtion
```
alebedev87, line 16:
```suggestion
// now the certificates will be present as a secret object, due to this it proceeds with the assumption
```

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil
+	}
+
+	fldPath := field.NewPath("spec", "TLS", "externalCertificate")
+	var errs field.ErrorList
+	if (newTLS != nil && newTLS.ExternalCertificate != nil) || (oldTLS != nil && oldTLS.ExternalCertificate != nil) {
+		errs = append(errs, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarc)...)
+	}
+
+	return errs
```
alebedev87, line 35:
```suggestion
	if !opts.AllowExternalCertificates {
		// return nil since if the feature gate is off
		// ValidateHostUpdate() is sufficient to validate
		// permissions
		return nil
	}

	newTLS := new.Spec.TLS
	oldTLS := older.Spec.TLS
	if (newTLS != nil && newTLS.ExternalCertificate != nil) || (oldTLS != nil && oldTLS.ExternalCertificate != nil) {
		return routecommon.CheckRouteCustomHostSAR(ctx, field.NewPath("spec", "TLS", "externalCertificate"), sarc)
	}

	return nil
```

pkg/route/hostassignment/externalcertificate_test.go
```diff
@@ -0,0 +1,209 @@
...
+				new: &routev1.Route{
+					Spec: routev1.RouteSpec{},
+				},
+				older: &routev1.Route{
+					Spec: routev1.RouteSpec{},
+				},
+				allow: false,
+				opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			},
+			want: field.ErrorList{},
```
alebedev87, line 45: `field.ErrorList{}` is equivalent to `nil`? Hm, how the error check should look like on the caller side?
```go
if errs := ValidateHostExternalCertificate(); len(errs) != 0 {}
```
Also, `nil` can be returned too. I think just `nil` as "no error" would be the most idiomatic.

alebedev87 (commented):
First run: code oriented, not logic oriented.

pkg/route/validation/validation.go
```diff
@@ -342,6 +280,17 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
 		if len(tls.DestinationCACertificate) > 0 {
 			result = append(result, field.Invalid(fldPath.Child("destinationCACertificate"), "redacted destination ca certificate data", "edge termination does not support destination certificates"))
 		}
+
+		if opts.AllowExternalCertificates && tls.ExternalCertificate != nil {
+			if len(tls.Certificate) > 0 && len(tls.ExternalCertificate.Name) > 0 {
+				result = append(result, field.Invalid(fldPath.Child("externalCertificate"), tls.ExternalCertificate.Name, "cannot specify both tls.certificate and tls.externalCertificate"))
+				break
```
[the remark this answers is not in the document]
thejasn: The complementing else-if checks if external certificate  is present and attempts to validate external certificate. We can't have it just be an "else" since we need to check if external certificate is present. And without the break both if and else-if cases can be true. 

Attempting to validate external certificate when it's preconditions have failed isn't logically sound, hence the break.

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+			Verb:      "list",
+			Resource:  "secrets",
+			Name:      tls.ExternalCertificate.Name,
+		}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "router serviceaccount does not have permission to list this secret"))
+	}
+
+	secret, err := secrets.Secrets(route.Namespace).Get(ctx, tls.ExternalCertificate.Name, metav1.GetOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
+		errs = append(errs, field.NotFound(fldPath, err))
```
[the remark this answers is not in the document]
thejasn: The type conversion from k8s error to field error is so that the caller can append this error to a list of errors. Moreover, the "field not found" is the same as "value not found", 

```
// NotFound returns a *Error indicating "value not found".  This is
// used to report failure to find a requested value (e.g. looking up an ID).
func NotFound(field *Path, value interface{}) *Error {
	return &Error{ErrorTypeNotFound, field.String(), value, ""}
}
```

So the error message is going to look like "externalCertificate : not found : secret not found"

pkg/route/validation/validation_test.go
```diff
@@ -42,6 +52,63 @@ func createRouteSpecTo(name string, kind string) routev1.RouteTargetReference {
...
+	secrets   []*corev1.Secret
+	namespace string
+}
+
+func (t *testSecretGetter) Secrets(_ string) corev1client.SecretInterface {
+	existingObjects := []runtime.Object{}
+	for _, s := range t.secrets {
+		existingObjects = append(existingObjects, s)
+	}
+	return testclient.NewSimpleClientset(existingObjects...).CoreV1().Secrets(t.namespace)
```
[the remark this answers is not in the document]
thejasn: Need an interface that implements k8s runtime interface.

pkg/route/validation/validation_test.go
```diff
@@ -42,6 +52,63 @@ func createRouteSpecTo(name string, kind string) routev1.RouteTargetReference {
...
+}
+
+func (t *testSecretGetter) WithSecret(_ *routev1.Route, secret corev1.Secret) *testSecretGetter {
+	t.secrets = append(t.secrets, &secret)
```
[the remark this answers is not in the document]
thejasn: Addressed

pkg/route/common.go
```diff
@@ -0,0 +1,59 @@
...
+// routeValidationOptions used to tweak how/what fields are validated. These
+// options are propagated by the apiserver.
+type RouteValidationOptions struct {
+
+	// allowExternalCertificates option is set when RouteExternalCertificate
+	// feature gate is enabled on the apiserver.
+	AllowExternalCertificates bool
+}
+
+func CheckRouteCustomHostSAR(ctx context.Context, fldPath *field.Path, sarc SubjectAccessReviewCreator) field.ErrorList {
```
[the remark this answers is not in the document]
alebedev87: Resolved.

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+	errs = append(errs, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarc)...)
+
+	if err := authorizationutil.Authorize(sarc, &user.DefaultInfo{Name: "system:serviceaccount:openshift-ingress:router"},
+		&authorizationv1.ResourceAttributes{
+			Namespace: route.Namespace,
+			Verb:      "get",
+			Resource:  "secrets",
+			Name:      tls.ExternalCertificate.Name,
+		}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "router serviceaccount does not have permission to get this secret"))
```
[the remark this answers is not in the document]
alebedev87: Resolved.

PR #1551, opened 2023-07-11
s/toplogy/topology/

taotriad:
Fix multiple occurrences of topology being misspelled "toplogy".


taotriad:
/remove-lifecycle stale

It'd be nice if anyone could at least comment.

PR #1553, opened 2023-07-13
STOR-1376: Allow CredentialsRequestController to work in STS mode

jsafrane:
When using Amazon Web Services (AWS) Secure Token Service (STS), a OLM-based CSI driver operator must add `cloudTokenPath` and `providerSpec.stsIAMRoleARN` to its CredentialsRequest.

Add hook support to CredentialsRequestController to allow them so.

In addition, CredentialsRequestController must sync CredentialsRequest even when CCO is in Manual mode, because it can still create Secrets for STS CredentialsRequest.

Enhancement: https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/tokenized-auth-enablement-operators-on-cloud.md

@openshift/storage

### DIFF
--- a/pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
+++ b/pkg/operator/csi/credentialsrequestcontroller/credentials_request_controller.go
@@ -3,6 +3,7 @@ package credentialsrequestcontroller
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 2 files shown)


bertinatto (commented):
/lgtm

PR #1558, opened 2023-08-01
Fix techpreviewnouprade feature set initialization

thejasn:
Update `setFeatureGates()` to use `GetAll()` to fetch all known featuregates that were initialized instead of using `KnownFeatures()` since it returns the feature gates in the following format `RouteExternalCertificate=true|false (ALPHA - default=false)`. 

Follow up for: https://github.com/openshift/library-go/pull/1550

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -69,9 +69,9 @@ func setFeatureGates(featureGatesMap map[string]bool, featureGates featuregate.M
 	// ideally we filter these at the operator level, but that isn't trivial to do and this is.
 	// We don't allow users to set values, so hopefully we have e2e test that prevent invalid values.
 	allowedFeatureGates := map[string]bool{}
-	knownFeatures := sets.NewString(featureGates.KnownFeatures()...)
+	knownFeatures := featureGates.GetAll()
 	for featureGateName, val := range featureGatesMap {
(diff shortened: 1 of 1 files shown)


pkg/features/features.go
```diff
@@ -89,18 +89,22 @@ func setFeatureGates(featureGatesMap map[string]bool, featureGates featuregate.M
...
 	for _, enabled := range configv1.FeatureSets[configv1.Default].Enabled {
 		defaultFeatures.Insert(string(enabled.FeatureGateAttributes.Name))
-		enabledDefaultFeatures.Insert(string(enabled.FeatureGateAttributes.Name))
+		enabledFeatures.Insert(string(enabled.FeatureGateAttributes.Name))
 	}
 	for _, disabled := range configv1.FeatureSets[configv1.Default].Disabled {
 		defaultFeatures.Insert(string(disabled.FeatureGateAttributes.Name))
 	}
 
+	for _, enabled := range configv1.FeatureSets[configv1.TechPreviewNoUpgrade].Enabled {
```
deads2k, line 101: /hold

these are definitely not enabled.  this change is improper.

thejasn: Reverted and fixed the issue with `setFeatureGates()`.

PR #1566, opened 2023-08-22
Directly import sortorder from new url instead of replacing

RadekManak:
This PR removes the replace directive `vbom.ml/util => github.com/fvbommel/util`. Instead, it imports from the new URL directly.

The replace directive does not propagate to dependents. Forcing them to use replace too or get this error.

```
go: github.com/openshift/library-go@v0.0.0-20230809121909-d7e7beca5bae requires
        vbom.ml/util@v0.0.0-20180919145318-efcd4e0f9787: unrecognized import path "vbom.ml/util": https fetch: Get "https://vbom.ml/util?go-get=1": dial tcp: lookup vbom.ml: no such host
```


elmiko (commented):
seems like a good upgrade to me

/lgtm

soltysh (approved):
/approve

PR #1569, opened 2023-09-05
USHIFT-1593: Regenerate certs on hostname changes

pacevedom:
When generating certificates there is a check to see if the cert already exists. If it does, it compares the existing IPs and DNSs in the cert against the hostnames it should have. If there is one missing entry from the hostnames in the cert it is regenerated. If not, then the cert is left untouched.
There is no check to see if the entries in the cert are present in the hostname list, leaving room for outdated IP/DNS entries to remain.
In other words: If the certificate already exists and one or more entries are removed from the hostnames list, the cert will be left untouched and those wrong values will be part of the CN/SANs.

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -758,11 +758,10 @@ func GetServerCert(certFile, keyFile string, hostnames sets.String) (*TLSCertifi
 	}
 
 	cert := server.Certs[0]
-	stringCertIPs := make([]string, len(cert.IPAddresses))
-	for i, ip := range cert.IPAddresses {
-		stringCertIPs[i] = ip.String()
+	certNames := sets.NewString()
+	for _, ip := range cert.IPAddresses {
+		certNames.Insert(ip.String())
 	}
-	certNames := sets.NewString(stringCertIPs...)
 	certNames.Insert(cert.DNSNames...)
 	if hostnames.Equal(certNames) {
 		klog.V(4).Infof("Found existing server certificate in %s", certFile)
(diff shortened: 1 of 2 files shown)


pkg/crypto/crypto.go
```diff
@@ -758,15 +758,18 @@ func GetServerCert(certFile, keyFile string, hostnames sets.String) (*TLSCertifi
...
 	cert := server.Certs[0]
-	ips, dns := IPAddressesDNSNames(hostnames.List())
-	missingIps := ipsNotInSlice(ips, cert.IPAddresses)
-	missingDns := stringsNotInSlice(dns, cert.DNSNames)
-	if len(missingIps) == 0 && len(missingDns) == 0 {
+	stringCertIPs := make([]string, len(cert.IPAddresses))
+	for i, ip := range cert.IPAddresses {
+		stringCertIPs[i] = ip.String()
+	}
+	certNames := sets.NewString(stringCertIPs...)
```
stlaz, line 765: you can create the set and just insert to it to save the slice allocation

stlaz: also, I believe `sets.NewString` is deprecated, use `sets.New[string]()`

pacevedom: Good point on the first one, will do.
About the deprecation, it is indeed, but it implies changing the public API for the library (because of the comparison with the `hostnames` set. Unless I copy contents into a new `sets.New[string]()`)

pacevedom: First comment done.

stlaz: I see, you can leave the 2nd part as is for now, then.

stlaz (changes requested):
Please add a unit test proving that verifies the fix.

PR #1572, opened 2023-09-15
OCPBUGS-19056: apiserver: optimize IsLongRunningRequest

vrutkovs:
Avoid using expensive MatchString when request params are already set in requestInfo

Back in 3.x days this was necessary as buildconfigs and imagestreams was served under `/oapi`, but now they are proper CRs so its better to match them via requestInfo params

```
$ benchstat -delta-test=none old.txt new.txt               
name                                                                 old time/op  new time/op  delta
IsLongRunningRequest/path_/api-8                                     11.5ns ± 5%   9.9ns ± 8%  -14.20%
IsLongRunningRequest/path_/api/buildconfigs/foo/instantiatebinary-8   870ns ± 7%     6ns ± 9%  -99.32%
IsLongRunningRequest/path_/api/imagestreamimports-8                   391ns ± 3%    10ns ± 5%  -97.47%
```

Next this needs to be re-imported by openshift-apiserver

### DIFF
--- a/pkg/apiserver/apiserverconfig/longrunning.go
+++ b/pkg/apiserver/apiserverconfig/longrunning.go
@@ -2,25 +2,37 @@ package apiserverconfig
 
 import (
 	"net/http"
-	"regexp"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	genericfilters "k8s.io/apiserver/pkg/server/filters"
+
+	buildv1 "github.com/openshift/api/build/v1"
+	imagev1 "github.com/openshift/api/image/v1"
 )
 
-// request paths that match this regular expression will be treated as long running
-// and not subjected to the default server timeout.
-const originLongRunningEndpointsRE = "(/|^)(buildconfigs/.*/instantiatebinary|imagestreamimports)$"
-
 var (
-	originLongRunningRequestRE = regexp.MustCompile(originLongRunningEndpointsRE)
-	kubeLongRunningFunc        = genericfilters.BasicLongRunningRequestCheck(
-		sets.NewString("watch", "proxy"),
-		sets.NewString("attach", "exec", "proxy", "log", "portforward"),
-	)
+	longRunningVerbs        = sets.NewString("watch", "proxy")
+	longRunningSubresources = sets.NewString("attach", "exec", "proxy", "log", "portforward")
+	kubeLongRunningFunc     = genericfilters.BasicLongRunningRequestCheck(longRunningVerbs, longRunningSubresources)
 )
 
 func IsLongRunningRequest(r *http.Request, requestInfo *apirequest.RequestInfo) bool {
-	return originLongRunningRequestRE.MatchString(r.URL.Path) || kubeLongRunningFunc(r, requestInfo)
+	if requestInfo == nil {
+		return false
+	}
+
+	if requestInfo.APIGroup == buildv1.GroupName &&
+		requestInfo.Resource == "buildconfigs" &&
+		requestInfo.Subresource == "instantiatebinary" {
+		return true
+	}
+	if requestInfo.APIGroup == imagev1.GroupName &&
+		requestInfo.Resource == "imagestreamimports" {
+		return true
+	}
+	if kubeLongRunningFunc(r, requestInfo) {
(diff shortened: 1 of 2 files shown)


benluddy:
Is there E2E coverage for this if you were to open a proof PR?

vrutkovs:
We don't have e2e tests for this, the only thing it would affect is `apiserver_longrunning_requests` metrics iiuc

vrutkovs:
Tested https://github.com/openshift/kubernetes/pull/1789 and verified metrics in e2e-aws-ovn-cgroupsv2 test:
* `apiserver_request_duration_seconds_sum{resource="buildconfigs", subresource="instantiatebinary"}` shows multiple results (longest being 18 sec), not included in `apiserver_longrunning_requests{verb!="WATCH"}`
* `apiserver_request_duration_seconds_sum{resource="imagestreamimports"}` shows multiple results (longest being 80 sec), not included in `apiserver_longrunning_requests{verb!="WATCH"}`

So it doesn't introduce any changes to long request detection

pkg/apiserver/apiserverconfig/longrunning.go
```diff
@@ -2,25 +2,36 @@ package apiserverconfig
...
+	kubeLongRunningFunc = genericfilters.BasicLongRunningRequestCheck(
 		sets.NewString("watch", "proxy"),
 		sets.NewString("attach", "exec", "proxy", "log", "portforward"),
 	)
 )
 
 func IsLongRunningRequest(r *http.Request, requestInfo *apirequest.RequestInfo) bool {
-	return originLongRunningRequestRE.MatchString(r.URL.Path) || kubeLongRunningFunc(r, requestInfo)
+	if requestInfo.APIGroup == buildv1.GroupName &&
+		requestInfo.APIVersion == buildv1.GroupVersion.Version &&
```
p0lyn0mial, line 23: do we have to check the api version ?
i don't think it is required in this context. 
skipping the version check would make this code forward compatible.

pkg/apiserver/apiserverconfig/longrunning.go
```diff
@@ -2,25 +2,36 @@ package apiserverconfig
...
+		requestInfo.Resource == "buildconfigs" &&
+		requestInfo.Subresource == "instantiatebinary" {
+		return true
+	}
+	if requestInfo.APIGroup == imagev1.GroupName &&
+		requestInfo.APIVersion == imagev1.GroupVersion.Version &&
```
p0lyn0mial, line 29: we could also skip the version check for the img imports.

pkg/apiserver/apiserverconfig/longrunning.go
```diff
@@ -2,25 +2,36 @@ package apiserverconfig
...
-	kubeLongRunningFunc        = genericfilters.BasicLongRunningRequestCheck(
+	kubeLongRunningFunc = genericfilters.BasicLongRunningRequestCheck(
 		sets.NewString("watch", "proxy"),
 		sets.NewString("attach", "exec", "proxy", "log", "portforward"),
 	)
 )
 
 func IsLongRunningRequest(r *http.Request, requestInfo *apirequest.RequestInfo) bool {
-	return originLongRunningRequestRE.MatchString(r.URL.Path) || kubeLongRunningFunc(r, requestInfo)
+	if requestInfo.APIGroup == buildv1.GroupName &&
```
p0lyn0mial, line 22: can `requestInfo` param be `nil` ?

vrutkovs: Added a check to prevent it

pkg/apiserver/apiserverconfig/longrunning.go
vrutkovs: Good idea, all of these requests are longrunning, regardless of the version

p0lyn0mial:
@vrutkovs do you want to add some unit tests ? We could use `NewRequestInfo` (https://github.com/kubernetes/kubernetes/blob/bc6c7fa91201348d010b638fbadf32007c0ac546/staging/src/k8s.io/apiserver/pkg/endpoints/request/requestinfo.go#L117) to build the RequestInfo struct and pass it to the detection function.

vrutkovs:
Added unittests

soltysh (approved):
/approve

PR #1575, opened 2023-09-21
OCPBUGS-17113: rework `unstructured` types for monitoring 

rexagod:
This PR incorporates the following changes:

* refactoring the existing functions (and comments) a bit,
* cache-based apply for `unstructured` monitoring types, and,
* adding explicit `*KnownUnstructured` functions for `alertmanager` and `prometheus`.

***

[Slack discussion thread for the same.](https://redhat-internal.slack.com/archives/CB48XQ4KZ/p1694071423825099)

/cc @dgrisonnet


rexagod:
- [ ] Additionally, look into utilizing `observedGeneration` and `resourceVersion` fields whenever present to better cache based on object `spec` and `metadata` changes.

dgrisonnet:
I completely missed that PR, @rexagod is this still relevant?

rexagod:
Tests seem to be failing since earlier we only checked for `spec` changes when `Apply`ing a resource, unlike now, where we fail-fast if the `metadata` is the same. This functionality cannot be tested in a unit-test currently since we rely on `resourceVersion` and `generation` now, so I'll add an e2e test (which is a bit odd since I don't see those being run in the workflow, given there already exist some of those).

rexagod:
Ah, [makes sense](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/library-go/openshift-library-go-master-presubmits.yaml#L18).

rexagod:
I'd prefer a single `test-e2e` target in the aforementioned Prow definition, which would prevent us from having to add a new e2e target every-time it's introduced here, in `/release`.

rexagod:
(friendly ping @dgrisonnet)

rexagod:
(friendly ping @dgrisonnet)

rexagod:
cc @deads2k

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,156 @@ package resourceapply
...
 	}
 	if err != nil {
 		return nil, false, err
 	}
 
-	existingCopy := existing.DeepCopy()
+	// Skip if:
+	// * cache is non-nil, and,
+	// * the metadata hashes and resource version hashes match.
+	if cache.SafeToSkipApply(required, existing) {
```
deads2k, line 87: `hashOfResourceStruct` is unlikely to work on the map.  Either a unit test showing that it functions properly for `*unstructure.Unstructured` or update to have it hash the kuberentes-sigs yaml serialization.

rexagod: Added a unit-test.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,156 @@ package resourceapply
...
-	existingCopy := existing.DeepCopy()
+	// Check if the metadata objects differ.
+	modifiedPtr := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(modifiedPtr, &existingObjectMetaTyped, requiredObjectMetaTyped)
+	if !*modifiedPtr {
+		// Update cache even if certain fields are not modified, in order to maintain a consistent cache based on the
+		// resource hash. The resource hash depends on the entire metadata, not just the fields that were checked above,
+		cache.UpdateCachedResourceMetadata(required, existingCopy)
+		return existingCopy, false, nil
+	}
```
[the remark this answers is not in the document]
rexagod: ACK, I've moved this below the `spec` check.

rexagod:
(bump)

dusk125:
lgtm

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,159 @@ package resourceapply
...
+	existingCopy, modified, err := ensureGenericSpec(required, existingCopy, noDefaulting, equality.Semantic)
 	if err != nil {
-		recorder.Warningf("PrometheusRuleUpdateFailed", "Failed to update PrometheusRule.monitoring.coreos.com/v1: %v", err)
-		return nil, true, err
+		return nil, false, err
 	}
-
-	recorder.Eventf("PrometheusRuleUpdated", "Updated PrometheusRule.monitoring.coreos.com/v1 because it changed")
-	return newObj, true, err
+	if !modified {
```
deads2k, line 124: If spec did not change, but labels do change.  How does this code avoid returning early?

rexagod: Addressed in the latest commit, also added some tests.

pkg/operator/resource/resourceapply/monitoring_test.go
```diff
@@ -99,49 +153,80 @@ func TestApplyServiceMonitor(t *testing.T) {
...
+			if len(actions) != len(tc.expectActions) {
+				t.Fatalf("expected %d actions, got %d: %v", tc.expectActions.Len(), len(actions), actions)
+			}
+			actionSet := sets.New[string]()
+			for _, action := range actions {
+				actionSet.Insert(action.GetVerb())
+			}
+			if !tc.expectActions.Equal(actionSet) {
+				t.Fatalf("expected actions %v, got %v", tc.expectActions, actionSet)
+			}
```
rexagod, line 207: Casting actions' verbs to a set here to account for flakes such as the one below (owing to order guarantees).
```

```

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,150 @@ package resourceapply
...
+	existing, err := client.Resource(resourceGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		newObj, createErr := client.Resource(serviceMonitorGVR).Namespace(namespace).Create(ctx, required, metav1.CreateOptions{})
-		if createErr != nil {
-			recorder.Warningf("ServiceMonitorCreateFailed", "Failed to create ServiceMonitor.monitoring.coreos.com/v1: %v", createErr)
-			return nil, true, createErr
-		}
-		recorder.Eventf("ServiceMonitorCreated", "Created ServiceMonitor.monitoring.coreos.com/v1 because it was missing")
-		return newObj, true, nil
+		requiredCopy := required.DeepCopy()
```
machine424, line 74: why do we need a deepCopy here? I can't see any call that may alter it.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,150 @@ package resourceapply
...
 	}
 
-	existingCopy := existing.DeepCopy()
+	// Check if the metadata objects differ.
+	didMetadataModify := resourcemerge.BoolPtr(false)
+	resourcemerge.EnsureObjectMeta(didMetadataModify, &existingObjectMetaTyped, requiredObjectMetaTyped)
 
-	toUpdate, modified, err := ensureGenericSpec(required, existingCopy, noDefaulting, equality.Semantic)
+	// Deep-check the spec objects for equality, and update the cache in either case.
+	existingCopy, didSpecModify, err := ensureGenericSpec(required, existingCopy, noDefaulting, equality.Semantic)
```
machine424, line 124: IIUC, the default values set by the server (that are not overwritten) are ignored, thus in most cases, it will consider the objects as not equal.

Do we plan to allow overwriting `noDefaulting` and  `equality.Semantic` in the future?
In the meanwhile I think we should state it explicitly (as "improved" in `XXXImproved` may be misleading): "if you want to make sure UPDATES are only triggered when needed, you should enable the cache."

CMO doesn't use the cache for example (for now), as the no-cache `ApplyXXX` utils it's using now ensure limited UPDATES even without the cache.

machine424 (commented):
I suggest we shorten the feedback loop and try this on CMO before the merge/release.
And that could be considered as an extra test.
I can help with that.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: > Do we plan to allow overwriting noDefaulting and  equality.Semantic in the future?

I'm +1 on moving `noDefaulting` and `equality.Semantic` parameters to the function signature.

> In the meanwhile I think we should state it explicitly (as "improved" in XXXImproved may be misleading): "if you want to make sure UPDATES are only triggered when needed, you should enable the cache."

I'm not sure what is being referred to here? Are you suggesting a comment?

> CMO doesn't use the cache for example (for now), as the no-cache ApplyXXX utils it's using now ensure limited UPDATES even without the cache.

I believe we will need to initialize a cache for CMO to be able to benefit from the changes being made here, i.e., dropping `ApplyXXXImproved` in favor of `ApplyUnstructuredResourceImproved`.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Dropped.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Friendly ping. :)

dusk125 (changes requested):

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,164 @@ package resourceapply
...
 	}
 	if err != nil {
 		return nil, false, err
 	}
 
-	existingCopy := existing.DeepCopy()
+	// Skip if:
+	// * cache is non-nil, and,
+	// * the metadata hashes and resource version hashes match.
+	if cache.SafeToSkipApply(required, existing) {
```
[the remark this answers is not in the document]
rexagod: Ah! It seems I made the last change in a hurry, fixed.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Additionally, we might want to supply custom a `equalityChecker` or `defaultingFunc` that caters to the CRD to make things more robust.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,168 @@ package resourceapply
...
+	existingCopy := existing.DeepCopy()
+	existingObjectMeta, found, err := unstructured.NestedMap(existingCopy.Object, "metadata")
 	if err != nil {
 		return nil, false, err
 	}
-
-	if !modified {
-		return nil, false, nil
+	if !found {
+		return nil, false, errorsstdlib.New(fmt.Sprintf("metadata not found in %s", existingCopy.GetName()))
```
jan--f, line 110: Should these error returns maybe contain context about the object where the error occurred? Here its the existing object, further down the required?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -2,148 +2,168 @@ package resourceapply
...
-	newObj, err := client.Resource(prometheusRuleGVR).Namespace(namespace).Update(ctx, toUpdate, metav1.UpdateOptions{})
-	if err != nil {
-		recorder.Warningf("PrometheusRuleUpdateFailed", "Failed to update PrometheusRule.monitoring.coreos.com/v1: %v", err)
-		return nil, true, err
-	}
-
-	recorder.Eventf("PrometheusRuleUpdated", "Updated PrometheusRule.monitoring.coreos.com/v1 because it changed")
-	return newObj, true, err
+	// Perform update if resource exists but different from the required (desired) one.
+	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, required, metav1.UpdateOptions{})
```
jan--f, line 158: Should this function not check if `required` has a `ResourceVersion` set and if so if it corresponds to the existing object before updating? Wouldn't this otherwise risk overwriting prior updates?

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Hmm, so `found` here corresponds to the [`existingObject`](https://github.com/openshift/library-go/pull/1575/files#diff-55de3da9293d86816615a21ec35c9725ab821c04b938de22d815679554b2fe92R105) and the [`requiredObject`](https://github.com/openshift/library-go/pull/1575/files#diff-55de3da9293d86816615a21ec35c9725ab821c04b938de22d815679554b2fe92R112) below. Do you mean appending something more than just the object name in the error messages? LMK if I missed something.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: I believe due to the fact that:

* the resource version cache updates [here](https://github.com/openshift/library-go/blob/e6bb5506390c06b624c7192bc41ed4fc4b517618/pkg/operator/resource/resourceapply/resource_cache.go#L149) are always sourced from the existing state and never the required one, and,
* an explicit required resource version that is not equal to the existing one will cause the Kube API to throw an error,

should ensure a valid and latest copy of the resource version is maintained in the cache, and used for skipping updates. WDYT?

pkg/operator/resource/resourceapply/monitoring.go
jan--f: Sure I understand when the error is returned. But iiuc the error returned will always be something like `metadata not found in prometheus-k8s`, whether its the existing copy or the required object. Adding information which object generates an error might be a good idea?

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Right, the write happens [here](https://github.com/openshift/library-go/blob/e6bb5506390c06b624c7192bc41ed4fc4b517618/pkg/operator/resource/resourceapply/resource_cache.go#L117) where the cache key is dependent on the existing object's `ResourceVersion`. I've added a fail-fast check.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Oh, you mean the object key (as in `namespace/object`) for a better context? Done!

jan--f:
Thanks, this
/lgtm

pkg/operator/resource/resourceapply/monitoring.go
rexagod: @machine424 suggested we drop these parameters, since we will only be introducing this function for the Prometheus object for now. If there are any defaults that need to be set for it, they should be set in Prometheus operator, which makes sense.

However, when and if we decide to expand this in the future, we will need to add these to cater to other components' use cases that don't have such middlewares (and require setting defaults in CMO itself, as that's where their configuration is defined).

rexagod:
@jan--f Can you please `lgtm` again, I pushed a fixup commit that I dropped after a while.

jan--f:
/lgtm
:+1:

rexagod:
Rebased, needs `/lgtm`.

rexagod:
Friendly ping. 🙂 

cc @deads2k

pkg/operator/resource/resourceapply/monitoring.go
p0lyn0mial: Unfortunately, this doesn't seem to work because the server requires an RV to be set when performing an update. This manifests as (the operator is permanently broken):
```
{Failed to upgrade kube-apiserver, operator was degraded (KubeAPIServerStaticResources_SyncError): KubeAPIServerStaticResourcesDegraded: "assets/alerts/api-usage.yaml" (string): prometheusrules.monitoring.coreos.com "api-usage" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
KubeAPIServerStaticResourcesDegraded:   }
```

The issue is quite serious, as it will block upgrades to `4.18`. Please take a look at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1750, especially the `e2e-aws-ovn-upgrade` job.

@rexagod , are you willing to work on a fix? I think we would like to have a test to prevent future bugs.
If the code is part of 4.17 I think we should back port the fix and the test.

p0lyn0mial:
Unfortunately this PR doesn't work as expected, see https://github.com/openshift/library-go/pull/1575#discussion_r1791701946 for more info.

PR #1579, opened 2023-09-27
bump(openshift/client-go)

soltysh:
To pick up the work from https://github.com/openshift/client-go/pull/256


bertinatto (commented):
/lgtm

PR #1582, opened 2023-10-06
cert-inspection: add GatherCertsFromPlatformNamespaces, PKIRegistryInfo

vrutkovs:
This adds few functions used in e2e test (from https://github.com/openshift/origin/pull/28222)

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
@@ -1,6 +1,7 @@
 package certgraphanalysis
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
@@ -8,9 +9,9 @@ import (
 
 // TODO these should all be eliminated in favor of self-describing annotations.
 
-func guessLogicalNamesForCertKeyPairList(in *certgraphapi.CertKeyPairList) {
+func guessLogicalNamesForCertKeyPairList(in *certgraphapi.CertKeyPairList, nodes map[string]int) {
 	for i := range in.Items {
-		meaning := guessMeaningForCertKeyPair(in.Items[i])
+		meaning := guessMeaningForCertKeyPair(in.Items[i], nodes)
 		in.Items[i].LogicalName = meaning.name
 		in.Items[i].Description = meaning.description
 	}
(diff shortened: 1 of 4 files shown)


pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -0,0 +1,98 @@
+package certgraphapi
+
+import "fmt"
+
+type PKIRegistryInfo struct {
+	CertificateAuthorityBundles []PKIRegistryInClusterCABundle    `json:"certificateAuthorityBundles"`
```
deads2k, line 6: if you want these for in-cluster, name them in-cluster to let a different type tackle the on-disk ones.

pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -0,0 +1,98 @@
+package certgraphapi
+
+import "fmt"
+
+type PKIRegistryInfo struct {
```
deads2k, line 5: given a godoc comment indicating whether or not you anticipate this growing fields to represent information like, "supports offline-cert-regeneration" or whether you see that as distinct.

pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -0,0 +1,98 @@
...
+
+type PKIRegistryInClusterCertKeyPair struct {
+	SecretLocation InClusterSecretLocation `json:"secretLocation"`
+
+	CertKeyInfo PKIRegistryCertKeyPairInfo `json:"certKeyInfo"`
+}
+
+type PKIRegistryCertKeyPairInfo struct {
+	OwningJiraComponent string `json:"owningJiraComponent"`
+	Description         string `json:"description"`
```
deads2k, line 18: indicate how long you're expecting this to be?  One sentence or one paragraph?

pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type PKIRegistryCertificateAuthorityInfo struct {
+	OwningJiraComponent string `json:"owningJiraComponent"`
+	Description         string `json:"description"`
+
+	//CertificateData []PKIRegistryCertKeyMetadata
+}
+
+type RegistrySecretByNamespaceName []PKIRegistryInClusterCertKeyPair
```
deads2k, line 36: move functions to a helper package to better match "normal" APIs.  We keep them separate to handle dependency chains better.

pkg/certs/cert-inspection/certgraphapi/type_registry.go
deads2k, line 6: mock up an SSA tag here. I think it would be map

deads2k:
minor comments, lgtm otherwise.

pkg/certs/cert-inspection/certgraphapi/type_registry.go
vrutkovs: Lets rename `PKIRegistryInfo` into `ClusterPKIRegistryInfo` to differentiate it from on-disk certificates? We probably don't need to have different sets of on-disk registries for different topologies and platforms

pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
```diff
@@ -55,22 +56,59 @@ var secretLocationToLogicalName = map[certgraphapi.InClusterSecretLocation]logic
 	newSecretLocation("", ""): newMeaning("", ""),
 }
 
-func guessMeaningForCertKeyPair(in certgraphapi.CertKeyPair) logicalMeaning {
-	for _, loc := range in.Spec.SecretLocations {
-		if meaning, ok := secretLocationToLogicalName[loc]; ok {
-			return meaning
-		}
+func formatSecretLocation(loc certgraphapi.InClusterSecretLocation, nodes map[string]int) certgraphapi.InClusterSecretLocation {
```
deads2k, line 59: I'm ok with this approach as your starting point (won't block merge), but I suspect you'll actually want to scan everything and replace the nodename in all secret names, all namespaces, all dns names, all subjects, etc.

pkg/certs/cert-inspection/certgraphapi/type_registry.go
```diff
@@ -0,0 +1,45 @@
+package certgraphapi
+
+// PKIRegistryInfo holds information about TLS artifacts stored in etcd. This includes object location and metadata based on object annotations
+type PKIRegistryInfo struct {
+	// +mapType:=atomic
+	CertificateAuthorityBundles []PKIRegistryInClusterCABundle `json:"certificateAuthorityBundles"`
```
deads2k, line 6: I'll allow a followup, but I suspect this object grows into `inClusterCertificateAuthorityBundles` on `onDiskCertificateAuthorityBundles`

vrutkovs: I'd rather have `PKIRegistryInfo` include `OnDiskPKIRegistryInfo` and `ClusterPKIRegistryInfo` structs, but yeah, we could merge them in one structure later

PR #1584, opened 2023-10-10
track metadata from incluster secrets and configmaps

deads2k:
This will add enough information to the raw data for us to properly generate details with inverted indexes in origin.


vrutkovs (approved):
/lgtm

PR #1585, opened 2023-10-12
OCPBUGS-19527: wrap errors for clarity

simonpasquier:
OCPBUGS-19527 shows that when the code fails to retrieve the list of secrets from openshift-monitoring, the leaf error isn't logged (only  `"failed to list secrets in the openshift-monitoring namespace"`).

### DIFF
--- a/test/library/metrics/query.go
+++ b/test/library/metrics/query.go
@@ -19,6 +19,8 @@ import (
 	"k8s.io/client-go/transport"
 )
 
+const prometheusServiceAccountName = "prometheus-k8s"
+
 // NewPrometheusClient returns Prometheus API or error
 // Note: with thanos-querier you must pass an entire Alert as a query. Partial queries return an error, so have to pass the entire alert.
 // Example query for an Alert:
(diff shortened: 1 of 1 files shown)


simonpasquier:
@deads2k could you approve this please? Thanks.

test/library/metrics/query.go
```diff
@@ -28,29 +30,31 @@ import (
 func NewPrometheusClient(ctx context.Context, kclient kubernetes.Interface, rc routeclient.Interface) (prometheusv1.API, error) {
 	_, err := kclient.CoreV1().Services("openshift-monitoring").Get(ctx, "prometheus-k8s", metav1.GetOptions{})
```
slashpai, line 31: should we create a const for namespace as well since used in manu places?

slashpai, line 31:
```suggestion
	_, err := kclient.CoreV1().Services("openshift-monitoring").Get(ctx, prometheusServiceAccountName, metav1.GetOptions{})
```

simonpasquier:
@deads2k @soltysh could one of you two approve this?

test/library/metrics/query.go
simonpasquier: I'd defer to a follow-up PR since it's not directly related to the bug.

soltysh (approved):
/approve

PR #1586, opened 2023-10-13
OCPBUGS-19987: configure readiness probe for guard pods

tjungblu:


geliu2016 (approved):
/label cherry-pick-approved

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -5,8 +5,8 @@ import (
 	"crypto/sha1"
 	_ "embed"
 	"fmt"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
```
dgrisonnet, line 8: nit: this should be with the other library-go imports

pkg/operator/staticpod/controller/common/readyz_conf.go
```diff
@@ -0,0 +1,13 @@
+package common
```
dgrisonnet, line 1: this could be moved to https://github.com/openshift/library-go/tree/master/pkg/monitor I think

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -289,13 +286,14 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Host = operands[node.Name].Status.PodIP
-			// The readyz port as string type is expected to be convertible into int!!!
-			readyzPort, err := strconv.Atoi(c.readyzPort)
-			if err != nil {
-				panic(err)
-			}
-			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromInt(readyzPort)
-			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Path = c.readyzEndpoint
+
+			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Port = intstr.FromString(c.readyZConf.Port)
```
dgrisonnet, line 290: This change mean that we won't crash on config error anymore, but default to 0 and I don't think that's what we want since it hides the config error to the users.

pkg/operator/staticpod/controller/guard/guard_controller_test.go
```diff
@@ -2,6 +2,8 @@ package guard
 
 import (
 	"context"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
```
dgrisonnet, line 5: nit: this should be with the other library-go imports

pkg/operator/staticpod/controllers.go
```diff
@@ -2,6 +2,7 @@ package staticpod
 
 import (
 	"fmt"
+	"github.com/openshift/library-go/pkg/operator/staticpod/controller/common"
```
dgrisonnet, line 5: nit: this should be with the other library-go imports

pkg/operator/staticpod/controllers.go
```diff
@@ -105,6 +105,7 @@ type Builder interface {
 	WithCustomInstaller(command []string, installerPodMutationFunc installer.InstallerPodMutationFunc) Builder
 	WithPruning(command []string, staticPodPrefix string) Builder
 	WithPodDisruptionBudgetGuard(operatorNamespace, operatorName, readyzPort, readyzEndpoint string, createConditionalFunc func() (bool, bool, error)) Builder
+	WithPodDisruptionBudgetGuardReadyZ(operatorNamespace string, operatorName string, createConditionalFunc func() (bool, bool, error), readyZConf common.ReadyZConf) Builder
```
dgrisonnet, line 108: I don't think we should maintain both `WithPodDisruptionBudgetGuard` and `WithPodDisruptionBudgetGuardReadyZ`, but rather replace the original readyz args by ReadyZConf

pkg/operator/staticpod/controller/common/readyz_conf.go
tjungblu: done

pkg/operator/staticpod/controller/guard/guard_controller.go
tjungblu: ah good point. There's `IntOrString.Parse(string)` , which does that exact same thing.

pkg/operator/staticpod/controllers.go
tjungblu: mentioned on slack, that's for backward compatibility given we're currently updating all services with library-go for the http/2 fix. I've done the change in the latest commit anyway, because people can also pick the prior commit hashes when updating.

geliu2016 (approved):
/label cherry-pick-approved

dgrisonnet (approved):
/approve
/lgtm

Holding for someone else to have a second look, but it looks good to me :+1: 
/hold

pkg/operator/staticpod/controllers.go
```diff
@@ -104,7 +104,7 @@ type Builder interface {
 	// the installer pod is created for a revision.
 	WithCustomInstaller(command []string, installerPodMutationFunc installer.InstallerPodMutationFunc) Builder
 	WithPruning(command []string, staticPodPrefix string) Builder
-	WithPodDisruptionBudgetGuard(operatorNamespace, operatorName, readyzPort, readyzEndpoint string, createConditionalFunc func() (bool, bool, error)) Builder
+	WithPodDisruptionBudgetGuard(operatorNamespace string, operatorName string, readyZConf monitor.ReadyZConf, createConditionalFunc func() (bool, bool, error)) Builder
```
dgrisonnet, line 107: readyZConf should be passed by pointer to avoid unnecessary copy of the struct.

tjungblu: good point, will fix it along with anything else the 2nd reviewer needs

benluddy:
I see https://github.com/openshift/cluster-etcd-operator/pull/1134 adds a timeout to the readyz-to-etcd request. If I'm understanding the linked issue, the problem is the event spam generated by kubelet-side probe errors. Instead of increasing the configured probe timeout, could you set a much shorter timeout (a few seconds) for the request to etcd in the readyz handler? The kubelet would see a successful probe indicating not-ready and not a failed probe.

tjungblu:
> Instead of increasing the configured probe timeout

@benluddy I don't want to increase the timeout, I want to increase the InitialDelaySeconds. So no, just setting a different timeout on the server is not sufficient.

PR #1592, opened 2023-10-20
cert-inspection: remove hardcoded cert description and name guessing

vrutkovs:
This removes `logicalMeaning` and hardcoded table of cert description is removed (it should be stored in origin, imported from openshift/api)


pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
```diff
@@ -7,56 +7,7 @@ import (
...
-}
-
-func newSecretLocation(namespace, name string) certgraphapi.InClusterSecretLocation {
-	return certgraphapi.InClusterSecretLocation{
-		Namespace: namespace,
-		Name:      name,
-	}
-}
-
-var secretLocationToLogicalName = map[certgraphapi.InClusterSecretLocation]logicalMeaning{
```
deads2k, line 27: Are these logically per-location or per content?  Where are you moving them to?

vrutkovs: Per-location. Description is now in origin's `tls/rawTLSInfo`

pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
```diff
@@ -267,7 +201,7 @@ func formatKubeSchedulerVersionedCABundle(loc certgraphapi.InClusterConfigMapLoc
 		return loc, false
 	}
 	return certgraphapi.InClusterConfigMapLocation{
-		Name:      "serving-cert",
+		Name:      "serviceaccount-ca",
```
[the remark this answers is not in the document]
vrutkovs: I've filled in all revisioned / hashed secrets from test results in https://github.com/openshift/origin/pull/28305, but once it merges we'll probably add more.

I wonder if we should introduce "isHashed" and "isRevisioned" for certificates instead (can be deferred to phase 2)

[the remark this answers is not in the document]
deads2k: if you're ok with https://github.com/openshift/origin/pull/28362, we can probably defer.  though the benefit in this renaming is likely lost

pkg/certs/cert-inspection/certgraphanalysis/nameguessing.go
deads2k: > Per-location. Description is now in origin's `tls/rawTLSInfo`

We need some kind of historical to find and fix all the descriptions.  You want to create an issue or open PRs for each of these?

pkg/certs/cert-inspection/certgraphutils/util.go
```diff
@@ -50,9 +51,9 @@ func (n ConfigMapLocationByNamespaceName) Less(i, j int) bool {
 	return n[i].Name < n[j].Name
 }
 
-func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, certKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
+func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, certKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair, nodes map[string]int) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
 	for i, curr := range certKeyPairs {
-		if targetLocation == curr.SecretLocation {
+		if certgraphanalysis.FormatSecretLocation(targetLocation, nodes) == curr.SecretLocation {
```
deads2k, line 56: This change doesn't look correct to me.  The only thing that should return for "matches" should be an exact location match.  The failures we're seeing mean that we're not describing that location properly.

Revisions should be removed (that's https://github.com/openshift/origin/pull/28362) and the node name and IP should be replaced by <master-0> <IP-master-0-> and the like when we store the raw TLS info.  Probably in the serialization step, with a straight sed

vrutkovs: That would make raw TLS info in fact processed, not "raw", isn't it? My understanding was that rawTLSInfo produced in cluster would be actual state of the cluster - all revisions, hashes and IPs preserved. This would be useful for recert tool, which has work with actual cluster secrets directly.

However, ownership JSON would have preprocessed names - no revisions/hashes, synthetic names for certs containing IPs. The test would convert actual name using `FormatSecretLocation` and use this name to compare with ownership entries (I'll rename vars do use better names here)

pkg/certs/cert-inspection/certgraphutils/util.go
```diff
@@ -50,20 +51,22 @@ func (n ConfigMapLocationByNamespaceName) Less(i, j int) bool {
...
 }
 
-func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, certKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
-	for i, curr := range certKeyPairs {
-		if targetLocation == curr.SecretLocation {
-			return &certKeyPairs[i], nil
+func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, expectedCertKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair, nodes map[string]int) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
+	formattedSecretLocation := certgraphanalysis.FormatSecretLocation(targetLocation, nodes)
+	for i, expected := range expectedCertKeyPairs {
+		if formattedSecretLocation == expected.SecretLocation {
```
deads2k, line 57: This isn't acceptable. The match must be done based on the exact input.

pkg/certs/cert-inspection/certgraphutils/util.go
```diff
@@ -50,20 +50,20 @@ func (n ConfigMapLocationByNamespaceName) Less(i, j int) bool {
 	return n[i].Name < n[j].Name
 }
 
-func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, certKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
-	for i, curr := range certKeyPairs {
-		if targetLocation == curr.SecretLocation {
-			return &certKeyPairs[i], nil
+func LocateCertKeyPair(targetLocation certgraphapi.InClusterSecretLocation, expectedCertKeyPairs []certgraphapi.PKIRegistryInClusterCertKeyPair, nodes map[string]int) (*certgraphapi.PKIRegistryInClusterCertKeyPair, error) {
```
deads2k, line 53: nodes is no longer needed, right?

deads2k:
proof PR and I could see going down this route *if* you commit to fixing descriptions prior to EOY.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -159,14 +159,12 @@ func PKIListFromParts(ctx context.Context, inClusterResourceData *certgraphapi.P
 	for i := range certs {
 		certList.Items = append(certList.Items, *certs[i])
 	}
-	guessLogicalNamesForCertKeyPairList(certList, nodes)
```
deads2k, line 162: You will owe something functional for the cluster-debug-tools document generation in 4.15 once we make this change.  You ready?

deads2k:
/lgtm
/hold

holding for @vrutkovs to acknowledge https://github.com/openshift/library-go/pull/1592/files?diff=split&w=0#r1378898527 .  Feel free to release hold when you're ready.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
vrutkovs: Fairly certain all teams can complete descriptions by 4.15 GA

PR #1593, opened 2023-10-20
certrotation: add more options for consumers

stevekuznetsov:
certrotation: allow setting an owner reference on rotated content

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

certrotation: make status reporting optional

When we're rotating certificates in a context that does not have a
static pod on which OperatorConditions are exposed, we need to emit our
status reporting in another manner. Unfortunately, without some external
place to store the prior reporing state, we can't only report on state
transitions. Therefore, we choose to only report errors and hope that it
will not be spammy.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -28,6 +28,8 @@ type CABundleConfigMap struct {
 	Namespace string
 	// Name is the name of the ConfigMap to maintain.
 	Name string
+	// Owner is an optional reference to add to the secret that this rotator creates.
+	Owner *metav1.OwnerReference
 
 	// Plumbing:
 	Informer      corev1informers.ConfigMapInformer
@@ -47,6 +49,18 @@ func (c CABundleConfigMap) ensureConfigMapCABundle(ctx context.Context, signingC
 	if apierrors.IsNotFound(err) {
 		// create an empty one
 		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
+	}
+	if c.Owner != nil {
+		var found bool
+		for _, ref := range caBundleConfigMap.ObjectMeta.OwnerReferences {
+			if ref == *c.Owner {
+				found = true
+				break
+			}
+		}
+		if !found {
+			caBundleConfigMap.ObjectMeta.OwnerReferences = append(caBundleConfigMap.ObjectMeta.OwnerReferences, *c.Owner)
+		}
 	}
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
 	if err != nil {
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -51,16 +51,7 @@ func (c CABundleConfigMap) ensureConfigMapCABundle(ctx context.Context, signingC
 		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
 	}
 	if c.Owner != nil {
-		var found bool
-		for _, ref := range caBundleConfigMap.ObjectMeta.OwnerReferences {
-			if ref == *c.Owner {
-				found = true
-				break
-			}
-		}
-		if !found {
-			caBundleConfigMap.ObjectMeta.OwnerReferences = append(caBundleConfigMap.ObjectMeta.OwnerReferences, *c.Owner)
-		}
+		ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
 	}
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
 	if err != nil {
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -27,15 +27,41 @@ const (
 	RunOnceContextKey = "cert-rotation-controller.openshift.io/run-once"
 )
 
+// StatusReporter knows how to report the status of cert rotation
+type StatusReporter interface {
+	Report(ctx context.Context, syncErr error) (updated bool, updateErr error)
+}
+
+var _ StatusReporter = (*StaticPodConditionStatusReporter)(nil)
+
+type StaticPodConditionStatusReporter struct {
+	// conditionName is used in operator conditions to identify this controller, compare CertRotationDegradedConditionTypeFmt.
+	conditionName string
+
+	// Plumbing:
+	OperatorClient v1helpers.StaticPodOperatorClient
+}
+
+func (s *StaticPodConditionStatusReporter) Report(ctx context.Context, syncErr error) (bool, error) {
+	newCondition := operatorv1.OperatorCondition{
+		Type:   fmt.Sprintf(condition.CertRotationDegradedConditionTypeFmt, s.conditionName),
+		Status: operatorv1.ConditionFalse,
+	}
+	if syncErr != nil {
+		newCondition.Status = operatorv1.ConditionTrue
+		newCondition.Reason = "RotationError"
+		newCondition.Message = syncErr.Error()
+	}
+	_, updated, updateErr := v1helpers.UpdateStaticPodStatus(ctx, s.OperatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition))
+	return updated, updateErr
+}
+
 // CertRotationController does:
 //
 // 1) continuously create a self-signed signing CA (via RotatedSigningCASecret) and store it in a secret.
 // 2) maintain a CA bundle ConfigMap with all not yet expired CA certs.
 // 3) continuously create a target cert and key signed by the latest signing CA and store it in a secret.
 type CertRotationController struct {
-	// name is used in operator conditions to identify this controller, compare CertRotationDegradedConditionTypeFmt.
-	name string
-
 	// rotatedSigningCASecret rotates a self-signed signing CA stored in a secret.
 	rotatedSigningCASecret RotatedSigningCASecret
 	// CABundleConfigMap maintains a CA bundle config map, by adding new CA certs coming from rotatedSigningCASecret, and by removing expired old ones.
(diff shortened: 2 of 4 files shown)


pkg/operator/certrotation/cabundle.go
```diff
@@ -28,6 +28,8 @@ type CABundleConfigMap struct {
 	Namespace string
 	// Name is the name of the ConfigMap to maintain.
 	Name string
+	// Owner is an optional reference to add to the secret that this rotator creates.
+	Owner *metav1.OwnerReference
```
stlaz, line 32: Can you please elaborate on how this would be useful?

pkg/operator/certrotation/signer.go
```diff
@@ -37,6 +37,9 @@ type RotatedSigningCASecret struct {
 	// rotation on expiration only, but not interfere with the ordinary rotation controller.
 	RefreshOnlyWhenExpired bool
 
+	// Owner is an optional reference to add to the secret that this rotator creates.
+	Owner *metav1.OwnerReference
```
stlaz, line 41: Who is the typical owner, then? Why would you want the secret to be garbage collected if it disappears?

Please explain in the godoc or remove.

pkg/operator/certrotation/target.go
```diff
@@ -50,6 +50,9 @@ type RotatedSelfSignedCertKeySecret struct {
 	// rotation on expiration only, but not interfere with the ordinary rotation controller.
 	RefreshOnlyWhenExpired bool
 
+	// Owner is an optional reference to add to the secret that this rotator creates.
+	Owner *metav1.OwnerReference
```
stlaz, line 54: Who is the typical owner, then? Why would you want the secret to be garbage collected if it disappears?

Please explain in the godoc or remove.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -94,12 +94,22 @@ func (c CertRotationController) Sync(ctx context.Context, syncCtx factory.SyncCo
...
+		// if the user has provided a client, we can report status to the object
+		_, updated, updateErr := v1helpers.UpdateStaticPodStatus(ctx, c.OperatorClient, v1helpers.UpdateStaticPodConditionFn(newCondition))
+		if updateErr != nil {
+			return updateErr
+		}
+		if updated && syncErr != nil {
+			syncCtx.Recorder().Warningf("RotationError", newCondition.Message)
+		}
+	} else {
+		// otherwise, we're just going to log; since we don't have an object on the cluster
```
stlaz, line 107: Does hypershift not report its status anyhow?

stevekuznetsov: They do, but this code is very tightly coupled to the StaticPodCondition, etc - would you rather see the whole reporting bit factored out as an interface?

pkg/operator/certrotation/cabundle.go
stevekuznetsov: In the HyperShift case, secrets and configmaps have their ownerref used for an index, and event handlers for those GVRs use the owner ref to know what related HostedControlPlane object to reconcile when dependent resources change.

pkg/operator/certrotation/signer.go
stevekuznetsov: I'm not sure how to explain that - the semantic meaning / reason for removal is up to the user.

pkg/operator/certrotation/client_cert_rotation_controller.go
stlaz: If we can a reconciliation fail more visible for other operators, I think it would be worth it.

pkg/operator/certrotation/cabundle.go
```diff
@@ -48,6 +50,18 @@ func (c CABundleConfigMap) ensureConfigMapCABundle(ctx context.Context, signingC
 		// create an empty one
 		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
 	}
+	if c.Owner != nil {
+		var found bool
```
stlaz, line 54: `ensureOwnerReference()`

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -27,15 +27,41 @@ const (
...
+
+type StaticPodConditionStatusReporter struct {
+	// name is used in operator conditions to identify this controller, compare CertRotationDegradedConditionTypeFmt.
+	name string
+
+	// Plumbing:
+	OperatorClient v1helpers.StaticPodOperatorClient
+}
+
+func (s StaticPodConditionStatusReporter) Report(ctx context.Context, syncErr error) (bool, error) {
```
stlaz, line 45: pointer receiver please

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -27,15 +27,41 @@ const (
...
+}
+
+var _ StatusReporter = (*StaticPodConditionStatusReporter)(nil)
+
+type StaticPodConditionStatusReporter struct {
+	// name is used in operator conditions to identify this controller, compare CertRotationDegradedConditionTypeFmt.
+	name string
+
+	// Plumbing:
+	OperatorClient v1helpers.StaticPodOperatorClient
```
stlaz, line 42: move to v1helpers among the other status updating methods, add a constructor and make both the struct attributes private

stlaz: +replace `name` with `conditionName string` and pass the whole condition name in there

stevekuznetsov: > add a constructor and make both the struct attributes private

Why?

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -44,23 +70,21 @@ type CertRotationController struct {
 	RotatedSelfSignedCertKeySecret RotatedSelfSignedCertKeySecret
 
 	// Plumbing:
-	OperatorClient v1helpers.StaticPodOperatorClient
+	StatusReporter StatusReporter
 }
 
 func NewCertRotationController(
-	name string,
```
deads2k, line 51: Suggest keeping name for event reporting (probably should have been there.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -44,23 +70,21 @@ type CertRotationController struct {
 	RotatedSelfSignedCertKeySecret RotatedSelfSignedCertKeySecret
 
 	// Plumbing:
-	OperatorClient v1helpers.StaticPodOperatorClient
+	StatusReporter StatusReporter
 }
 
 func NewCertRotationController(
```
deads2k, line 76: No owner here?  Or optional With or something?

pkg/operator/certrotation/client_cert_rotation_controller.go
stevekuznetsov: The event recorder currently has `recorder.WithComponentSuffix("cert-rotation-controller")` - would you like to see the name as a further suffix on top?

pkg/operator/certrotation/client_cert_rotation_controller.go
stevekuznetsov: Sorry, I don't follow the comment - could you please clarify?

pkg/operator/certrotation/client_cert_rotation_controller.go
deads2k: > The event recorder currently has `recorder.WithComponentSuffix("cert-rotation-controller")` - would you like to see the name as a further suffix on top? Or replace the hard-coded suffix entirely?

Your call, but I think having a distinction there matters.

pkg/operator/certrotation/client_cert_rotation_controller.go
deads2k: When I create a cert rotation order, can I wire up an ownerref for everything created by the operator during construction?

stevekuznetsov: I'm not sure if the implementation in this PR is exactly what you're looking for, but, yes - the constituent parts of the rotator allow for an OwnerRef to be passed in.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -44,23 +70,22 @@ type CertRotationController struct {
...
 	recorder events.Recorder,
+	reporter StatusReporter,
 ) factory.Controller {
 	c := &CertRotationController{
-		name:                           name,
 		rotatedSigningCASecret:         rotatedSigningCASecret,
 		CABundleConfigMap:              caBundleConfigMap,
 		RotatedSelfSignedCertKeySecret: rotatedSelfSignedCertKeySecret,
-		OperatorClient:                 operatorClient,
+		StatusReporter:                 reporter,
```
dgrisonnet, line 88: @stevekuznetsov  why don't we set the conditionName of the reporter to name here? The current approach makes it pretty weird in consumers since we essentially have to duplicate the controllerName when instantiating a controller. For example here: https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/certrotationcontroller/certrotationcontroller.go#L131-L168, we have to do something like:

```diff
diff --git a/pkg/operator/certrotationcontroller/certrotationcontroller.go b/pkg/operator/certrotationcontroller/certrotationcontroller.go
index 676a26182..f6f18e4b2 100644
--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
+++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
@@ -163,8 +163,8 @@ func newCertRotationController(
                        Client:        kubeClient.CoreV1(),
                        EventRecorder: eventRecorder,
                },
-               operatorClient,
                eventRecorder,
+               certrotation.StatusReporter{componentName: "AggregatorProxyClientCert", OperatorClient: operatorClient},
        )
        ret.certRotators = append(ret.certRotators, certRotator)
```

stevekuznetsov: So we'd want to refactor `NewCertRotationController` to take a reporter factory or something and close over passing in a name for construction? Sounds reasonable - but we can't assume the reporter impl, is all.

atiratree: @stevekuznetsov @dgrisonnet  I have run into this when bumping the library-go as well. What about this approach? https://github.com/openshift/library-go/pull/1630

dgrisonnet: Your approach seems fine to me. The controller will always know which condition to report and I don't think we have a use-case where we would need one reporter per condition.

@stevekuznetsov would that approach perhaps also fit your use-case?

stevekuznetsov: Yep - either let's do that or just export the field, either one is good!

PR #1594, opened 2023-10-26
cert rotation: add optional jira component and description annotations to TLS artifacts

vrutkovs:
This updates RotatedSelfSignedCertKeySecret, RotatedSigningCASecret and CABundleConfigMap structs and functions so that managed secrets and configmaps are annotated to identify component and purpose of the secret.

See https://github.com/openshift/cluster-kube-apiserver-operator/pull/1568 for example of usage

### DIFF
--- a/pkg/operator/certrotation/annotations.go
+++ b/pkg/operator/certrotation/annotations.go
@@ -1,0 +1,34 @@
+package certrotation
+
+import (
+	"github.com/openshift/api/annotations"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewTLSArtifactObjectMeta(name, namespace, jiraComponent, description string) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Namespace: namespace,
+		Name:      name,
+		Annotations: map[string]string{
+			annotations.OpenShiftComponent:   jiraComponent,
+			annotations.OpenShiftDescription: description,
+		},
+	}
+}
+
+// EnsureTLSMetadataUpdate mutates objectMeta setting necessary annotations if unset
+func EnsureTLSMetadataUpdate(meta *metav1.ObjectMeta, jiraComponent, description string) bool {
+	modified := false
+	if meta.Annotations == nil {
+		meta.Annotations = make(map[string]string)
+	}
+	if len(jiraComponent) > 0 && meta.Annotations[annotations.OpenShiftComponent] != jiraComponent {
+		meta.Annotations[annotations.OpenShiftComponent] = jiraComponent
+		modified = true
+	}
+	if len(description) > 0 && meta.Annotations[annotations.OpenShiftDescription] != description {
+		meta.Annotations[annotations.OpenShiftDescription] = description
+		modified = true
+	}
+	return modified
+}
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -30,6 +30,10 @@ type CABundleConfigMap struct {
 	Name string
 	// Owner is an optional reference to add to the secret that this rotator creates.
 	Owner *metav1.OwnerReference
+	// JiraComponent annotates tls artifacts so that owner could be easily found
+	JiraComponent string
+	// Description is a human-readable one sentence description of certificate purpose
+	Description string
 
 	// Plumbing:
 	Informer      corev1informers.ConfigMapInformer
@@ -48,11 +52,28 @@ func (c CABundleConfigMap) ensureConfigMapCABundle(ctx context.Context, signingC
 	caBundleConfigMap := originalCABundleConfigMap.DeepCopy()
 	if apierrors.IsNotFound(err) {
 		// create an empty one
-		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
+		caBundleConfigMap = &corev1.ConfigMap{ObjectMeta: NewTLSArtifactObjectMeta(
+			c.Name,
+			c.Namespace,
+			c.JiraComponent,
+			c.Description,
+		)}
 	}
+
+	needsMetadataUpdate := false
 	if c.Owner != nil {
-		ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
+		needsMetadataUpdate = ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
 	}
+	if len(c.JiraComponent) > 0 || len(c.Description) > 0 {
+		needsMetadataUpdate = EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta, c.JiraComponent, c.Description) || needsMetadataUpdate
+	}
+	if needsMetadataUpdate && len(caBundleConfigMap.ResourceVersion) > 0 {
+		_, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
 	if err != nil {
 		return nil, err
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -44,6 +44,11 @@ type RotatedSigningCASecret struct {
 	// is used, early deletion will be catastrophic.
 	Owner *metav1.OwnerReference
 
+	// JiraComponent annotates tls artifacts so that owner could be easily found
+	JiraComponent string
+
+	// Description is a human-readable one sentence description of certificate purpose
+	Description string
 	// Plumbing:
 	Informer      corev1informers.SecretInformer
 	Lister        corev1listers.SecretLister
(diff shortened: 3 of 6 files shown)


pkg/operator/certrotation/signer.go
```diff
@@ -69,7 +75,7 @@ func (c RotatedSigningCASecret) ensureSigningCertKeyPair(ctx context.Context) (*
 
 	if needed, reason := needNewSigningCertKeyPair(signingCertKeyPairSecret.Annotations, c.Refresh, c.RefreshOnlyWhenExpired); needed {
 		c.EventRecorder.Eventf("SignerUpdateRequired", "%q in %q requires a new signing cert/key pair: %v", c.Name, c.Namespace, reason)
-		if err := setSigningCertKeyPairSecret(signingCertKeyPairSecret, c.Validity); err != nil {
+		if err := setSigningCertKeyPairSecret(signingCertKeyPairSecret, c.Validity, c.JiraComponent, c.Description); err != nil {
```
deads2k, line 78: how will this add the labels to secrets and configmaps in an existing cluster?

vrutkovs: atm these would be updated on rotation only. We'll need a followup to move annotation update elsewhere (possibly we should have a separate library function for that)

deads2k: > atm these would be updated on rotation only. We'll need a followup to move annotation update elsewhere (possibly we should have a separate library function for that)

We need to address this shortcoming as we merge the feature.  Upgraded clusters must be the same as freshly installed clusters.

pkg/operator/certrotation/signer.go
```diff
@@ -60,12 +67,19 @@ func (c RotatedSigningCASecret) ensureSigningCertKeyPair(ctx context.Context) (*
 	if apierrors.IsNotFound(err) {
 		// create an empty one
 		signingCertKeyPairSecret = &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: c.Namespace, Name: c.Name}}
```
deads2k, line 69: add description and component here?

pkg/operator/certrotation/signer.go
```diff
@@ -60,12 +67,19 @@ func (c RotatedSigningCASecret) ensureSigningCertKeyPair(ctx context.Context) (*
...
+		secretExists = false
 	}
 	signingCertKeyPairSecret.Type = corev1.SecretTypeTLS
 
 	if c.Owner != nil {
 		ensureOwnerReference(&signingCertKeyPairSecret.ObjectMeta, c.Owner)
 	}
+	if NeedsTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta, c.JiraComponent, c.Description) && secretExists {
```
deads2k, line 77: Is `NeedsTLSMetdataUpdate` mutating more like `ensureOwnerReference`?  If so, `ensureTLSMetadata`?

pkg/operator/certrotation/target.go
```diff
@@ -104,7 +111,7 @@ func (c RotatedSelfSignedCertKeySecret) ensureTargetCertKeyPair(ctx context.Cont
 
 	if reason := needNewTargetCertKeyPair(targetCertKeyPairSecret.Annotations, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired); len(reason) > 0 {
```
deads2k, line 112: Needs the `ensureTLSMetadata` block above this to handle the update case, right?

pkg/operator/csr/cert_controller.go
```diff
@@ -151,9 +157,16 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
 				Name:      c.SecretName,
```
deads2k, line 157: set known anntation values?

pkg/operator/certrotation/signer.go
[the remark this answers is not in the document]
vrutkovs: It mutates annotations, but it also needs "secretExists" input

pkg/operator/csr/cert_controller.go
vrutkovs: I'd prefer to contain all metadata changes once in `NeedsTLSMetadataUpdate`

pkg/operator/certrotation/signer.go
vrutkovs: Fixed by introducing `NeedsTLSMetadataUpdate`

pkg/operator/csr/cert_controller.go
```diff
@@ -151,9 +157,16 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
...
 			},
 		}
+		secretExists = false
 	case err != nil:
 		return fmt.Errorf("unable to get secret %q: %w", c.SecretNamespace+"/"+c.SecretName, err)
 	}
+	// ensure the secret has necessary metadata
+	if certrotation.NeedsTLSMetadataUpdate(&secret.ObjectMeta, c.JiraComponent, c.Description) && secretExists {
```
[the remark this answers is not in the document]
vrutkovs: Done, `apply` happens before content modification unless `get` returned `NotExists`

pkg/operator/certrotation/signer.go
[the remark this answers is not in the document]
deads2k: > It mutates annotations, but it also needs "secretExists" input

Mutation could happen regardless, you only need the check for writes, correct?

pkg/operator/certrotation/target.go
vrutkovs: done

pkg/operator/certrotation/annotations.go
```diff
@@ -0,0 +1,37 @@
...
+		Name:      name,
+		Annotations: map[string]string{
+			annotations.OpenShiftComponent:   jiraComponent,
+			annotations.OpenShiftDescription: description,
+		},
+	}
+}
+
+// NeedsTLSMetadataUpdate mutates objectMeta setting necessary annotations if unset
+func NeedsTLSMetadataUpdate(meta *metav1.ObjectMeta, jiraComponent, description string) bool {
```
deads2k, line 20: A boolean sounding method shouldn't mutate.  Notice that `ensure` isn't a bool-hint like `is`, `are`, `should`, `must`, `needs`.

pkg/operator/certrotation/cabundle.go
```diff
@@ -41,18 +45,29 @@ type CABundleConfigMap struct {
...
+		secretExists = false
 	}
 	if c.Owner != nil {
 		ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
 	}
+	if err := c.ensureTLSMetadata(ctx, caBundleConfigMap, secretExists); err != nil {
+		return nil, err
+	}
+
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
```
deads2k, line 71:
```suggestion
	needsMetadataUpdate := false
	if c.Owner != nil {
		ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
	}
	if len(c.JiraComponent) > 0 || len(c.Description) > 0{
	    needsMetadataupdate = ensureTLSMetadata(&caBundleConfigMap.ObjectMeta, c.JiraComponent, c.Descrption) || needsMetadataupdate
	}
	if needsMetadataUpdate && secretExists{
	        _, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, configMap)
		if err != nil {
			return err
		}
	}

	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
```

notice how it could (should?) be congruent to the ownerreference metadata change already possible.

pkg/operator/csr/cert_controller.go
```diff
@@ -265,6 +278,15 @@ func (c *clientCertificateController) syncCSR(secret *corev1.Secret) (map[string
 	return data, nil
 }
 
+func (c *clientCertificateController) ensureTLSMetadata(ctx context.Context, secret *corev1.Secret, exists bool) error {
+	if certrotation.NeedsTLSMetadataUpdate(&secret.ObjectMeta, c.JiraComponent, c.Description) && exists {
+		if err := c.saveSecret(secret); err != nil {
```
deads2k, line 283: this construction will cause the the saveSecret to get a conflict error, right?  You probably want this on an `else` for the c.csrName and you probably want the current impl to take into account both the syncCSR and the annotation state.

This particular controller can probably be split out from the others if you like.

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -58,10 +59,21 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
...
+			Namespace:   destinationConfigMap.Namespace,
+			Name:        destinationConfigMap.Name,
+			Annotations: make(map[string]string),
+		},
 		Data: map[string]string{
 			"ca-bundle.crt": string(caBytes),
 		},
-	}, nil
+	}
+	if len(jiraComponent) != 0 {
```
deads2k, line 72: followup, this is actually EnsureTLSMetadataUpdate now, right?

pkg/operator/certrotation/annotations.go
```diff
@@ -0,0 +1,37 @@
...
+		},
+	}
+}
+
+// EnsureTLSMetadataUpdate mutates objectMeta setting necessary annotations if unset
+func EnsureTLSMetadataUpdate(meta *metav1.ObjectMeta, jiraComponent, description string) bool {
+	modified := false
+	if meta.Annotations == nil {
+		meta.Annotations = map[string]string{
+			annotations.OpenShiftComponent:   "",
```
deads2k, line 24: followup, if you're lazy defining below, this leaves empty values distinct from unspecified.

vrutkovs:
/hold cancel

Works as expected in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1585

PR #1596, opened 2023-10-27
add indication of whether a secret is revisioned to raw data

deads2k:
Will link PR demonstrating usage.


deads2k:
proof will be in https://github.com/openshift/origin/pull/28362

PR #1601, opened 2023-11-01
cert-inspection: add RewriteNodeIP options

vrutkovs:
Skip secrets and configmaps hashed by openshift-monitoring, i.e. prometheus-adapter secret. Follow up for https://github.com/openshift/library-go/pull/1599

Testing this in https://github.com/openshift/origin/pull/28373

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
@@ -21,16 +21,10 @@ func (resourceFilteringOptions) approved() {}
 var (
 	SkipRevisioned = &resourceFilteringOptions{
 		rejectConfigMap: func(configMap *corev1.ConfigMap) bool {
-			if isRevisioned(configMap.OwnerReferences) {
-				return true
-			}
-			return false
+			return isRevisioned(configMap.OwnerReferences)
 		},
 		rejectSecret: func(secret *corev1.Secret) bool {
-			if isRevisioned(secret.OwnerReferences) {
-				return true
-			}
-			return false
+			return isRevisioned(secret.OwnerReferences)
 		},
 	}
 	SkipHashed = &resourceFilteringOptions{
--- a/pkg/certs/cert-inspection/certgraphanalysis/collector.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collector.go
@@ -2,7 +2,6 @@ package certgraphanalysis
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/openshift/api/annotations"
@@ -60,14 +59,6 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 	certs := []*certgraphapi.CertKeyPair{}
 	caBundles := []*certgraphapi.CertificateAuthorityBundle{}
 	errs := []error{}
-	nodes := map[string]int{}
-	nodeList, err := kubeClient.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to list cluster nodes: %v", err)
-	}
-	for i, node := range nodeList.Items {
-		nodes[node.Name] = i
-	}
 
 	// TODO here is the point where need to collect data like node names and IPs that need to be replaced
 	//  this will be something like options.Discovery(kubeClient, configClient).
@@ -78,13 +69,13 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 		errs = append(errs, err)
 	default:
 		for _, configMap := range configMapList.Items {
+			options.rewriteConfigMap(&configMap)
 			if !acceptConfigMap(&configMap) {
 				continue
 			}
 			if options.rejectConfigMap(&configMap) {
 				continue
 			}
-
 			details, err := InspectConfigMap(&configMap)
 			if err != nil {
 				errs = append(errs, err)
@@ -117,13 +108,13 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 		errs = append(errs, err)
 	default:
 		for _, secret := range secretList.Items {
+			options.rewriteSecret(&secret)
 			if !acceptSecret(&secret) {
 				continue
 			}
 			if options.rejectSecret(&secret) {
 				continue
 			}
-
 			details, err := InspectSecret(&secret)
 			if err != nil {
 				errs = append(errs, err)
@@ -149,18 +140,11 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 		}
 	}
 
-	for i := range caBundles {
-		options.rewriteCABundle(caBundles[i])
-	}
-	for i := range certs {
-		options.rewriteCertKeyPair(certs[i])
-	}
-
-	pkiList := PKIListFromParts(ctx, inClusterResourceData, certs, caBundles, nodes)
+	pkiList := PKIListFromParts(ctx, inClusterResourceData, certs, caBundles)
 	return pkiList, errors.NewAggregate(errs)
 }
 
-func PKIListFromParts(ctx context.Context, inClusterResourceData *certgraphapi.PerInClusterResourceData, certs []*certgraphapi.CertKeyPair, caBundles []*certgraphapi.CertificateAuthorityBundle, nodes map[string]int) *certgraphapi.PKIList {
+func PKIListFromParts(ctx context.Context, inClusterResourceData *certgraphapi.PerInClusterResourceData, certs []*certgraphapi.CertKeyPair, caBundles []*certgraphapi.CertificateAuthorityBundle) *certgraphapi.PKIList {
 	certs = deduplicateCertKeyPairs(certs)
 	certList := &certgraphapi.CertKeyPairList{}
 	for i := range certs {
--- a/pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
@@ -1,16 +1,23 @@
 package certgraphanalysis
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+	corev1 "k8s.io/api/core/v1"
 )
 
+type configMapRewriteFunc func(configMap *corev1.ConfigMap)
+type secretRewriteFunc func(secret *corev1.Secret)
 type caBundleRewriteFunc func(caBundle *certgraphapi.CertificateAuthorityBundle)
-
 type certKeyPairRewriteFunc func(certKeyPair *certgraphapi.CertKeyPair)
 
 type metadataOptions struct {
 	rewriteCABundle    caBundleRewriteFunc
 	rewriteCertKeyPair certKeyPairRewriteFunc
+	rewriteConfigMap   configMapRewriteFunc
+	rewriteSecret      secretRewriteFunc
 }
 
 func (metadataOptions) approved() {}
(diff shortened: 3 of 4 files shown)


pkg/certs/cert-inspection/certgraphanalysis/options.go
```diff
@@ -1,38 +1,80 @@
...
-			return false
+			return isRevisioned(configMap.OwnerReferences)
 		},
 		rejectSecret: func(secret *corev1.Secret) bool {
-			if isRevisioned(secret.OwnerReferences) {
-				return true
+			return isRevisioned(secret.OwnerReferences)
+		},
+	}
+	RewriteNames = &certGenerationOptions{
```
deads2k, line 29: This needs to be selective.  So multiple instances like `RewriteNodeName`, `RewriteNodeIP` so that it spans the logical name substitutions.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -41,6 +48,7 @@ func isPlatformNamespace(nsName string) bool {
 }
 
 type configMapFilterFunc func(configMap *corev1.ConfigMap) bool
+type configMapRewriteFunc func(configMap *corev1.ConfigMap, nodes map[string]int) (string, string)
```
deads2k, line 51: This should rewrite the actual configmap so other readers all see the same information.  It should indicate that it mutated by some kind of annotation.

deads2k, line 51: Accepting nodes here is a losing game.  We'll be adding more and more parameters. Instead when you create a rewriter do something like


```
RewriteNodeNames(nodes []*corev1.Node) *certGenerationOptions{
  return &certGenerationOptions{
     // body here using those nodes across all secret names
  }
}
```

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -87,28 +88,28 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
...
-							Namespace: configMap.Namespace,
-							Name:      configMap.Name,
-						},
-						CABundleInfo: certgraphapi.PKIRegistryCertificateAuthorityInfo{
-							OwningJiraComponent: configMap.Annotations[annotations.OpenShiftComponent],
-							Description:         configMap.Annotations[annotations.OpenShiftDescription],
-						},
-					})
-			}
+			name, namespace := options.rewriteConfigMap(&configMap)
```
deads2k, line 91: 1. mutation
2. rejection
3. !accept

pkg/certs/cert-inspection/certgraphanalysis/options.go
vrutkovs: Done, renamed into RewriteNodeIP

pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
```diff
@@ -44,3 +52,8 @@ func isRevisioned(ownerReferences []metav1.OwnerReference) bool {
 
 	return false
 }
+
+func hasMonitoringHashLabel(labels map[string]string) bool {
+	_, ok := labels["monitoring.openshift.io/hash"]
```
deads2k, line 57: you confirmed this hash isn't present on the original secret and configmap?

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -43,4 +52,41 @@ var (
...
+	RewriteNodeIPs = &metadataOptions{
+		rewriteCertKeyPair: func(certKeyPair *certgraphapi.CertKeyPair, o rewriteParams) {
+			rewriteIPInEtcdServingMetricsCertificateName(certKeyPair, o.nodes)
+			rewriteIPInEtcdServingCertificate(certKeyPair, o.nodes)
+			rewriteIPInEtcdPeerCertificate(certKeyPair, o.nodes)
+		},
+	}
 )
+
+func rewriteIPInEtcdServingCertificate(certKeyPair *certgraphapi.CertKeyPair, nodes map[string]int) {
```
deads2k, line 64: Let's talk at 10.  I think you've discovered two different mutators that need to happen at two different times.

1. secret itself requiring mutation so that the list of "these secret namespace,name tuples are present" is correct when stored.
2. the certkeypair and cabundle metadata so that names are consistent.

1 happens before filtering in the first loop
2 happens here.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -43,4 +52,41 @@ var (
 			}
 		},
 	}
+	RewriteNodeIPs = &metadataOptions{
+		rewriteCertKeyPair: func(certKeyPair *certgraphapi.CertKeyPair, o rewriteParams) {
```
deads2k, line 56: I think this construct is pointing to a need to make 

```go
type metadataOptions interface{
  rewriteCABundle(caBundle *certgraphapi.CertificateAuthorityBundle)
  rewriteCertKeyPair(certKeyPair *certgraphapi.CertKeyPair)
}
```

So that you can have a caller do

```go
certgraphanalysis.GatherCertsFromPlatformNamespaces(ctx, kubeClient, 
  certgraphanalysis.SkipRevisioned, 
  certgraphanalysis.ElideProxyCADetails,
  certgraphanalysis.RewriteNodeIPs(nodes),
)
```

by having

```go
func RewriteNodeIPs(nodes []*corev1.Nodes) metadataOptions{
  return &nodeIPRewriter{nodes: nodes}
}
```

pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
```diff
@@ -33,6 +33,14 @@ var (
 			return false
 		},
 	}
+	SkipHashed = &resourceFilteringOptions{
```
deads2k, line 36: a PR with just this could merge very quickly.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -84,28 +83,29 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
...
-							OwningJiraComponent: configMap.Annotations[annotations.OpenShiftComponent],
-							Description:         configMap.Annotations[annotations.OpenShiftDescription],
-						},
-					})
-			}
 			if err != nil {
 				errs = append(errs, err)
+				continue
+			}
+			if details == nil {
```
deads2k, line 91: +1 to this improvement

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
deads2k: writing the interface as above will allow us to separate decisions about "what sorts of things should we filter on" from "and now every option needs to change".  And rather than having missing information that is lightly enforced, we can use construction of various options to ensure data is correct and only those callers that want particular filters need to gather the data.

Choosing the nodes instead of a client makes unit testing easier and we have little performance benefit or temporal locality to gain by using a client.

pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
vrutkovs: Yeah, these are added only for the generated configmaps/secrets

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: In this case rewriting secret/configmap is sufficient

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -1,20 +1,81 @@
...
 import (
+	"fmt"
+	"strings"
+
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+	corev1 "k8s.io/api/core/v1"
 )
 
-type caBundleRewriteFunc func(caBundle *certgraphapi.CertificateAuthorityBundle)
+const lastRewrittenByAnnotation = "openshift.io/last-rewritten-by"
```
deads2k, line 11: followup: annotation to be something like `rewrite.cert-inspection.openshift.io/<rewriter's name>` so we track all ofthem.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -1,20 +1,81 @@
...
 
+type configMapRewriteFunc func(configMap *corev1.ConfigMap)
+type secretRewriteFunc func(secret *corev1.Secret)
+type caBundleRewriteFunc func(caBundle *certgraphapi.CertificateAuthorityBundle)
 type certKeyPairRewriteFunc func(certKeyPair *certgraphapi.CertKeyPair)
 
 type metadataOptions struct {
 	rewriteCABundle    caBundleRewriteFunc
 	rewriteCertKeyPair certKeyPairRewriteFunc
+	rewriteConfigMap   configMapRewriteFunc
```
deads2k, line 21: I think we'll want to split these into a different kind of option in the future, but not require now.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -1,20 +1,81 @@
...
+	rewriteSecret      secretRewriteFunc
 }
 
 func (metadataOptions) approved() {}
 
+func (l certGenerationOptionList) rewriteCABundle(caBundle *certgraphapi.CertificateAuthorityBundle) {
```
deads2k, line 27: why separate these from the definition of type in [s/filtering_options.go](https://github.com/openshift/library-go/pull/1601/files#diff-30028a369ddbe02bd5ade6c03b0f78f1bfab468370e390b854b24b5cb45255d8)

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -44,3 +105,48 @@ var (
...
+		rewriteSecret: func(secret *corev1.Secret) {
+			rewriteIPInEtcdServingMetricsSecret(secret, nodes)
+			rewriteIPInEtcdServingSecret(secret, nodes)
+			rewriteIPInEtcdPeerSecret(secret, nodes)
+		},
+	}
+}
+
+func rewriteIPInEtcdServingSecret(secret *corev1.Secret, nodes map[string]int) {
+	if secret.Namespace == "openshift-etcd" && strings.HasPrefix(secret.Name, "etcd-serving-") {
```
deads2k, line 124: why bound to etcd?  We want to do this for *all* secrets, right?  So the thing you're trying to replace is the master name.

deads2k: Same question throughout.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -44,3 +103,24 @@ var (
...
+
+func RewriteNodeIPs(nodeList *corev1.NodeList) *metadataOptions {
+	nodes := map[string]int{}
+	for i, node := range nodeList.Items {
+		nodes[node.Name] = i
+	}
+	return &metadataOptions{
+		rewriteSecret: func(secret *corev1.Secret) {
+			for nodeName, masterID := range nodes {
+				name := strings.ReplaceAll(secret.Name, nodeName, fmt.Sprintf("master-%d", masterID))
```
deads2k, line 115:
```suggestion
				name := strings.ReplaceAll(secret.Name, nodeName, fmt.Sprintf("<master-%d>", masterID))
```

This makes it clear that we have substitution happening when people do things like read generated docs

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -44,3 +103,24 @@ var (
 		},
 	}
 )
+
+func RewriteNodeIPs(nodeList *corev1.NodeList) *metadataOptions {
```
deads2k, line 107: Using `[]*corev1.Node` allows accurate mocking and makes it possible for consumers to wire in listers.  NodeList contains things like RV of the resource which isn't present in most controller use-cases.

deads2k:
will take future refinement as followups

/lgtm

deads2k:
/approve

wake up bot

PR #1602, opened 2023-11-01
guard: add workload partitioning annotation

dgrisonnet:
This is required to address https://issues.redhat.com/browse/OCPBUGS-22112.


dgrisonnet:
Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1576

PR #1603, opened 2023-11-02
Make client id configurable and only use hardcoded if it is not set

ardaguclu:
Requesting token functionality in library-go uses hardcoded client id. However,
this becomes an obstacle against the adoption of external oidc issuers. This PR
allows explicitly setting client id and if it is not set, hardcoded client id values will be used.


pkg/oauth/tokenrequest/request_token.go
```diff
@@ -87,8 +87,8 @@ func RequestTokenWithChallengeHandlers(clientCfg *restclient.Config, challengeHa
 // authzURLHandler is used to forward the user to the OAuth server's parametrized authorization URL to
 // retrieve the authorization code.
 // It starts a localhost server on port `callbackPort` (random port if unspecified) to exchange the authorization code for an access token.
-func RequestTokenWithLocalCallback(clientCfg *restclient.Config, authzURLHandler AuthorizationURLHandlerFunc, callbackPort int) (string, error) {
-	o, err := NewRequestTokenOptions(clientCfg, false).WithLocalCallback(authzURLHandler, callbackPort)
+func RequestTokenWithLocalCallback(clientCfg *restclient.Config, authzURLHandler AuthorizationURLHandlerFunc, callbackPort int, clientID string) (string, error) {
```
liouk, line 90: We could move the `clientID` argument right after `clientCfg` for the sake of consistency with `RequestTokenWithChallengeHandlers`.

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -133,7 +136,7 @@ func (o *RequestTokenOptions) WithChallengeHandlers(challengeHandlers ...challen
 // unstarted local callback server on the specified port.
 // If RequestTokenOptions.OsinConfig is nil, it will be defaulted using SetDefaultOsinConfig.
 // The caller is responsible for setting up the entire OsinConfig if the value is not nil.
-func (o *RequestTokenOptions) WithLocalCallback(handleAuthzURL AuthorizationURLHandlerFunc, localCallbackPort int) (*RequestTokenOptions, error) {
+func (o *RequestTokenOptions) WithLocalCallback(handleAuthzURL AuthorizationURLHandlerFunc, localCallbackPort int, clientID string) (*RequestTokenOptions, error) {
```
liouk, line 139: I'd also suggest here to be consistent with `WithChallengeHandlers`; we can move `clientID` to be the first argument.

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -121,7 +121,10 @@ func (o *RequestTokenOptions) WithChallengeHandlers(challengeHandlers ...challen
 	o.Handler = handler
 
 	if o.OsinConfig == nil {
-		if err := o.SetDefaultOsinConfig(openShiftCLIClientID, nil); err != nil {
+		if clientID == "" {
```
liouk, line 124: Replace with `len(clientID) == 0`.

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -143,7 +146,10 @@ func (o *RequestTokenOptions) WithLocalCallback(handleAuthzURL AuthorizationURLH
 
 	if o.OsinConfig == nil {
 		redirectUrl := fmt.Sprintf("http://%s/callback", o.LocalCallbackServer.ListenAddr())
-		if err := o.SetDefaultOsinConfig(openShiftCLIBrowserClientID, &redirectUrl); err != nil {
+		if clientID == "" {
```
liouk, line 149: Replace with `len(clientID) == 0`.

liouk (changes requested):

ardaguclu:
@liouk thanks for the review. Updated with all your suggestions.

liouk:
:+1: 

/lgtm

ardaguclu:
After completing my research about the implementation of supporting external OIDC providers in library-go, this PR is trivial compering to the other required changes and can be easily done inside of them. That's why, I'm closing this;

/close

PR #1604, opened 2023-11-02
cert-inspect: add functions to generate violation files

vrutkovs:
Add functions to generate cert metadata violations files


vrutkovs:
No need to have it in library-go, I'll make a PR for `origin` instead

PR #1611, opened 2023-11-06
STOR-1453: CSI: Set TLS ciphers based on observed configuration

bertinatto:
CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
+++ b/pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/configobserver"
+	libgoapiserver "github.com/openshift/library-go/pkg/operator/configobserver/apiserver"
 	"github.com/openshift/library-go/pkg/operator/configobserver/proxy"
 	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resourcesynccontroller"
@@ -22,9 +23,22 @@ func ProxyConfigPath() []string {
 	return []string{"targetcsiconfig", "proxy"}
 }
 
+// CipherSuitesPath returns the path for the observed TLS cipher suites. This
+// is a function to avoid exposing a slice that could potentially be appended.
+func CipherSuitesPath() []string {
+	return []string{"targetcsiconfig", "servingInfo", "cipherSuites"}
+}
+
+// MinTLSVersionPath the path for the observed minimum TLS version. This
+// is a function to avoid exposing a slice that could potentially be appended.
+func MinTLSVersionPath() []string {
+	return []string{"targetcsiconfig", "servingInfo", "minTLSVersion"}
+}
+
 // Listers implement the configobserver.Listers interface.
 type Listers struct {
-	ProxyLister_ configlistersv1.ProxyLister
+	ProxyLister_     configlistersv1.ProxyLister
+	APIServerLister_ configlistersv1.APIServerLister
 
 	ResourceSync       resourcesynccontroller.ResourceSyncer
 	PreRunCachesSynced []cache.InformerSynced
@@ -32,6 +46,10 @@ type Listers struct {
 
 func (l Listers) ProxyLister() configlistersv1.ProxyLister {
 	return l.ProxyLister_
+}
+
+func (l Listers) APIServerLister() configlistersv1.APIServerLister {
+	return l.APIServerLister_
 }
 
 func (l Listers) ResourceSyncer() resourcesynccontroller.ResourceSyncer {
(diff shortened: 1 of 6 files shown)


bertinatto:
/hold
for testing in https://github.com/openshift/gcp-pd-csi-driver-operator/pull/95

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -190,8 +190,43 @@ func WithPlaceholdersHook(configInformer configinformers.SharedInformerFactory)
...
+			return nil, fmt.Errorf("couldn't get the servingInfo.minTLSVersion config from observed config: %v", err)
+		}
+
+		pairs := []string{}
+		if cipherSuitesFound && len(cipherSuites) > 0 {
+			pairs = append(pairs, []string{"${TLS_CIPHER_SUITES}", strings.Join(cipherSuites, ",")}...)
+
+		}
+
+		if minTLSVersionFound && len(minTLSVersion) > 0 {
```
jsafrane, line 227: It would be good to mention that we expect `minTLSVersion` to be always present, because all TLS profiles specify it + `apiserver.spec.tlsSecurityProfile.custom.minTLSVersion` [cannot be empty](https://github.com/openshift/api/blob/c90b4c82852ead1fdb58866b685fe75b0b62163d/config/v1/0000_10_config-operator_01_apiserver-Default.crd.yaml#L144-L151).

Similarly for `cipherSuites` - validation prevents empty `spec.tlsSecurityProfile.custom.ciphers`

jsafrane, line 227: This hook will produce invalid Deployments (with literally `--tls-cipher-suites=${TLS_CIPHER_SUITES}`) until the observer finishes the initial `APIServer` CR observation. Is that what we want? Maybe `--tls-min-version=""  --tls-cipher-suites=""` would be a better default.

pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go
```diff
@@ -78,6 +78,7 @@ func NewCSIDriverControllerServiceController(
 	optionalInformers = append(optionalInformers, configInformer.Config().V1().Infrastructures().Informer())
 	var optionalManifestHooks []dc.ManifestHookFunc
 	optionalManifestHooks = append(optionalManifestHooks, WithPlaceholdersHook(configInformer))
+	optionalManifestHooks = append(optionalManifestHooks, WithServingInfo())
```
dobsonj, line 81: Can you also update the comment at lines 55-57 to mention the new `${TLS_MIN_VERSION}` placeholder?

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
dobsonj: Can the --tls-* arguments be omitted in that case? I'm not sure what behavior to expect if we set those arguments to empty strings, but if we omit those args, the deployment will be valid and kube-rbac-proxy will select its own defaults until the observer returns the value from APIServer.

jsafrane: > Can the --tls-* arguments be omitted in that case

It's actually hard, right now do a simple search for `${TLS_CIPHER_SUITES}` string and replace it. We would need to change that `${TLS_CIPHER_SUITES}` to contain also `--tls-cipher-suites` (which is easy), but then the kube-rbac-proxy needs to be able to handle empty string arguments.

bertinatto:
> /hold for testing in [openshift/gcp-pd-csi-driver-operator#95](https://github.com/openshift/gcp-pd-csi-driver-operator/pull/95)

I tested with AWS EBS and it worked well, including with observed proxy configuration.

/hold cancel

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
bertinatto: We shouldn't let any values other than those specified in the TLS profile to be set. For instance, if those args are set to empty strings, insecure ciphers would be specified instead (as per Go default ciphers). This is the worse possible outcome.

IMO the current behavior is fine, it's preferable to fail to start and it would only produce an invalid Deployment during cluster installation. I did add an additional check to return a different error message, though.

pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller_test.go
```diff
@@ -84,6 +100,12 @@ func makeFakeDriverInstance(modifiers ...driverModifier) *fakeDriverInstance {
 		},
 		Spec: opv1.OperatorSpec{
 			ManagementState: opv1.Managed,
+			ObservedConfig: runtime.RawExtension{
+				Raw: nil,
+				Object: &unstructured.Unstructured{
+					Object: map[string]any{"targetcsiconfig": defaultServingInfo},
```
jsafrane, line 106: Please add a test with empty `targetcsiconfig` and hook failing.

bertinatto: Added

bertinatto:
/assign @tkashem 
for approval

pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
```diff
@@ -66,16 +84,23 @@ func NewCSIConfigObserverController(
...
 			proxy.NewProxyObserveFunc(ProxyConfigPath()),
+			observeTLSSecurityProfile,
 		),
 	}
 
 	return c
 }
+
+func observeTLSSecurityProfile(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
+	return libgoapiserver.ObserveTLSSecurityProfileWithPaths(genericListers, recorder, existingConfig, MinTLSVersionPath(), CipherSuitesPath())
```
p0lyn0mial, line 105: An alternative would be to extract the expected configuration here and pass extracted/expected cfg to the existing function. Although this could be more error prone because you would have to remember to prune the cfg and handle some errors. WDYT ?

p0lyn0mial:
/approve
/hold

in case you would like to change https://github.com/openshift/library-go/pull/1611#discussion_r1399030628

pkg/operator/csi/csiconfigobservercontroller/csi_config_observer_controller.go
jsafrane: @p0lyn0mial Can you please elaborate? To what *existing function* should we pass the extracted configuration?

I thought it's point of an observer to *return* the extracted config, so the common observer can put it into `spec.observedConfig`.

jsafrane: Yeah, I'd like Fabio's version, IMO it's more explicit.
/hold cancel
(he's on PTO)

PR #1613, opened 2023-11-09
cert-inspect: collect certificates from disk

vrutkovs:
Tested in https://github.com/openshift/origin/pull/28444

TODO (can be follow ups):
* selinux labels, permissions
* replace IPs in paths
* keep just the last revision

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collector.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collector.go
@@ -8,7 +8,7 @@ import (
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/errors"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 )
@@ -19,6 +19,21 @@ func GatherCertsFromAllNamespaces(ctx context.Context, kubeClient kubernetes.Int
 
 func GatherCertsFromPlatformNamespaces(ctx context.Context, kubeClient kubernetes.Interface, options ...certGenerationOptions) (*certgraphapi.PKIList, error) {
 	return gatherFilteredCerts(ctx, kubeClient, platformConfigMaps, platformSecrets, options)
+}
+
+func GatherCertsFromDisk(ctx context.Context, kubeClient kubernetes.Interface, prefix, dir string, options ...certGenerationOptions) (*certgraphapi.PKIList, error) {
+	errs := []error{}
+
+	certs, err := gatherSecretsFromDisk(ctx, prefix, dir, options)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	caBundles, err := gatherCABundlesFromDisk(ctx, prefix, dir, options)
+	if err != nil {
+		errs = append(errs, err)
+	}
+	pkiList := PKIListFromParts(ctx, nil, certs, caBundles)
+	return pkiList, utilerrors.NewAggregate(errs)
 }
 
 var wellKnownPlatformNamespaces = sets.NewString(
@@ -145,7 +160,7 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 	}
 
 	pkiList := PKIListFromParts(ctx, inClusterResourceData, certs, caBundles)
-	return pkiList, errors.NewAggregate(errs)
+	return pkiList, utilerrors.NewAggregate(errs)
 }
 
 func recordedAnnotationsFrom(metadata metav1.ObjectMeta, annotationsToCollect []string) []certgraphapi.AnnotationValue {
@@ -186,4 +201,38 @@ func PKIListFromParts(ctx context.Context, inClusterResourceData *certgraphapi.P
 	}
 
 	return ret
+}
+
+func MergePKILists(ctx context.Context, first, second *certgraphapi.PKIList) *certgraphapi.PKIList {
+
+	if first == nil {
+		first = &certgraphapi.PKIList{}
+	}
+
+	if second == nil {
+		second = &certgraphapi.PKIList{}
+	}
+
+	certList := &certgraphapi.CertKeyPairList{
+		Items: append(first.CertKeyPairs.Items, second.CertKeyPairs.Items...),
+	}
+	certList = deduplicateCertKeyPairList(certList)
+
+	caBundlesList := &certgraphapi.CertificateAuthorityBundleList{
+		Items: append(first.CertificateAuthorityBundles.Items, second.CertificateAuthorityBundles.Items...),
+	}
+	caBundlesList = deduplicateCABundlesList(caBundlesList)
+
+	inClusterData := certgraphapi.PerInClusterResourceData{}
+	inClusterData.CertKeyPairs = append(inClusterData.CertKeyPairs, first.InClusterResourceData.CertKeyPairs...)
+	inClusterData.CertKeyPairs = append(inClusterData.CertKeyPairs, second.InClusterResourceData.CertKeyPairs...)
+	inClusterData.CertificateAuthorityBundles = append(inClusterData.CertificateAuthorityBundles, first.InClusterResourceData.CertificateAuthorityBundles...)
+	inClusterData.CertificateAuthorityBundles = append(inClusterData.CertificateAuthorityBundles, second.InClusterResourceData.CertificateAuthorityBundles...)
+	//TODO[vrutkovs] deduplicate inClusterData?
+
+	return &certgraphapi.PKIList{
+		CertificateAuthorityBundles: *caBundlesList,
+		CertKeyPairs:                *certList,
+		InClusterResourceData:       inClusterData,
+	}
 }
--- a/pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
@@ -1,31 +1,65 @@
 package certgraphanalysis
 
-import "github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+import (
+	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+)
 
 func deduplicateCertKeyPairs(in []*certgraphapi.CertKeyPair) []*certgraphapi.CertKeyPair {
 	ret := []*certgraphapi.CertKeyPair{}
 	for _, currIn := range in {
-		found := false
+		if currIn == nil {
+			panic("currIn is nil")
+		}
+
+		foundIdx := -1
 		for j, currOut := range ret {
-			if currIn == nil {
-				panic("one")
-			}
 			if currOut == nil {
-				panic("two")
+				panic("currOut is nil")
 			}
-			if currOut.Name == currIn.Name {
-				ret[j] = combineSecretLocations(ret[j], currIn.Spec.SecretLocations)
-				ret[j] = combineCertOnDiskLocations(ret[j], currIn.Spec.OnDiskLocations)
-				found = true
-				break
+			// currOut has no name set - skip and allow it to be found later
+			if len(currOut.Name) == 0 {
+				continue
+			}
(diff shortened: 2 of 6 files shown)


pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -69,90 +76,274 @@ func platformSecrets(obj *corev1.Secret) bool {
...
-						Description:         configMap.Annotations[annotations.OpenShiftDescription],
-					},
-				})
+	}
+
+	pkiList := PKIListFromParts(ctx, inClusterResourceData, certs, caBundles)
+	return pkiList, errors.NewAggregate(errs)
+}
+
+func GatherCertsFromDisk(ctx context.Context, kubeClient kubernetes.Interface, dir string, options ...*certGenerationOptions) (*certgraphapi.PKIList, error) {
```
deads2k, line 95: I'm guessing the plan is to create a command in `openshift-tests` or in `oc`.  If that's the case, create the `*Options` struct with the `Validate` and `Run` methods for standard embedding.

I'd expect to also see this library also contain the code (likely also in a different `*Options` struct) to 
1. select nodes
2. force schedule pods to each node
3. exfiltrate the result of the command

Running the collection as an init container, then a sleep as the main container will allow you to `oc copy` (embedded) the result out.

vrutkovs: I'd really prefer to avoid running new containers - its always not as reliable as "fetch data from already running container"

deads2k:
Can we get a function here that takes an image pull spec returns a PKIList that is merged from all masters?  We could then call that from openshift-tests and/or a command.

vrutkovs:
Sure, I could move `fetchOnDiskCertificates` to library-go, do we expect it to be reused anywhere else? That would require moving manifests there as well, not sure if its a good idea

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -145,7 +160,7 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 	}
 
 	pkiList := PKIListFromParts(ctx, inClusterResourceData, certs, caBundles)
-	return pkiList, errors.NewAggregate(errs)
+	return pkiList, apierrors.NewAggregate(errs)
```
deads2k, line 163: `utilerrors` by convention.  `apierrors` is for the one in apimachinery with `IsNotFoundError`

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -13,7 +15,14 @@ func deduplicateCertKeyPairs(in []*certgraphapi.CertKeyPair) []*certgraphapi.Cer
...
 			}
-			if currOut.Name == currIn.Name {
+			// Two certificates with matching names
+			if len(currOut.Name) != 0 && len(currIn.Name) != 0 && currOut.Name == currIn.Name {
+				ret[j] = combineSecretLocations(ret[j], currIn.Spec.SecretLocations)
+				ret[j] = combineCertOnDiskLocations(ret[j], currIn.Spec.OnDiskLocations)
+				found = true
+			}
+			// Key modulus and cert modulus match - add key to cert
+			if len(currOut.Name) != 0 && len(currIn.Name) == 0 && currOut.Spec.CertMetadata.CertIdentifier.PubkeyModulus == currIn.Spec.CertMetadata.CertIdentifier.PubkeyModulus {
```
deads2k, line 25: are these two mutually exclusive cases (switch) or do one and do the other (as written now)

deads2k, line 25: need `len(currOut.Name) == 0 && len(currIn.Name) != 0` too?  Seems like we would.

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -56,10 +81,25 @@ func combineCertOnDiskLocations(in *certgraphapi.CertKeyPair, rhs []certgraphapi
 			}
 		}
 		if !found {
+			// Store key to be merged into Cert
+			if curr.Cert.Path == "" {
```
deads2k, line 85:
```suggestion
			if len(curr.Cert.Path) == 0{
```

for kube standard.

Still reading to figure out why we assign an empty path to `keyLocation`.  Possible logic inversion?

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -56,10 +81,25 @@ func combineCertOnDiskLocations(in *certgraphapi.CertKeyPair, rhs []certgraphapi
...
 			out.Spec.OnDiskLocations = append(out.Spec.OnDiskLocations, curr)
 		}
 	}
 
+	// Fill in Key property if it was found earlier and unset
+	if len(keyLocation.Path) == 0 {
+		return out
+	}
+	for idx, loc := range out.Spec.OnDiskLocations {
+		if loc.Key.Path == "" {
```
deads2k, line 98:
```suggestion
		if len(loc.Key.Path) == 0 {
```

pkg/certs/cert-inspection/certgraphanalysis/disk_certs.go
```diff
@@ -0,0 +1,180 @@
...
+
+	pathWithoutPrefix := strings.Replace(path, prefix, "", 1)
+	return &certgraphapi.CertKeyPair{
+		Spec: certgraphapi.CertKeyPairSpec{
+			SecretLocations: nil,
+			OnDiskLocations: []certgraphapi.OnDiskCertKeyPairLocation{
+				{
+					Key: certgraphapi.OnDiskLocation{
+						Path: pathWithoutPrefix,
+						// TODO[vrutkovs]: fill in other settings
```
deads2k, line 77: permissions and se linux labels as likely next steps.  Followup fine.

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -56,10 +81,25 @@ func combineCertOnDiskLocations(in *certgraphapi.CertKeyPair, rhs []certgraphapi
...
+			out.Spec.OnDiskLocations[idx].Key = keyLocation
```
deads2k, line 99: I'm not clear on why we're doing this.  This appears to be saying, "every cert location that lacks a corresponding key location gets a key location from some other location pair".  But I'm not sure why we want to do that.

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
vrutkovs: This iterates over all found certs and keys in `rhs` and produces a new keypair. It would be a list of cert/key pairs, or a key or a cert separately. This is why keys (or certs) with empty path are not ignored - otherwise PEM certkey pairs would be missing (these are treated as certificates at the moment).

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
vrutkovs: we want to merge two locations if one has the cert and the other has the key (with matching `PubkeyModulus`). It also stores unmatching certs/key (needs improved test to handle PEM certkey pair)

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
vrutkovs: This would require updating certificate name from `currIn`. Instead we just iterate further until they swap and `currOut` has `Name` set

vrutkovs: Right, these are exclusive. Rewrote it to make use of `switch`

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -1,34 +1,68 @@
...
+			case 0:
+				// Certificate and key have matching modulus
+				if currOut.Spec.CertMetadata.CertIdentifier.PubkeyModulus == currIn.Spec.CertMetadata.CertIdentifier.PubkeyModulus {
+					foundIdx = j
+				}
+			default:
+				// Two certificates with matching names
+				if currOut.Name == currIn.Name {
+					foundIdx = j
+				}
```
deads2k, line 33:
```suggestion
			switch  {
			case len(currIn.Name) == 0 && currOut.Spec.CertMetadata.CertIdentifier.PubkeyModulus == currIn.Spec.CertMetadata.CertIdentifier.PubkeyModulus:
				// No name to check and certificate and key have matching modulus
				foundIdx = j
			case if currOut.Name == currIn.Name:
				// Two certificates with matching names
				foundIdx = j
```

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -1,34 +1,68 @@
...
-				found = true
-				break
+			// currOut has no name set - skip and allow it to be found later
+			if len(currOut.Name) == 0 {
+				continue
+			}
+			switch len(currIn.Name) {
+			case 0:
+				// Certificate and key have matching modulus
+				if currOut.Spec.CertMetadata.CertIdentifier.PubkeyModulus == currIn.Spec.CertMetadata.CertIdentifier.PubkeyModulus {
```
deads2k, line 26: should we actually just use PubkeyModulus?

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -1,34 +1,68 @@
...
-				panic("two")
+				panic("currOut is nil")
 			}
-			if currOut.Name == currIn.Name {
-				ret[j] = combineSecretLocations(ret[j], currIn.Spec.SecretLocations)
-				ret[j] = combineCertOnDiskLocations(ret[j], currIn.Spec.OnDiskLocations)
-				found = true
-				break
+			// currOut has no name set - skip and allow it to be found later
+			if len(currOut.Name) == 0 {
```
deads2k, line 20: so if we hit this code path where the name of currout is empty, we should still be able to match on the `PubkeyModulus` and then merge a "better" name in, right?

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -1,34 +1,68 @@
...
+			default:
+				// Two certificates with matching names
+				if currOut.Name == currIn.Name {
+					foundIdx = j
+				}
+			}
+			// Append currIn locations to found certkeypair
+			if foundIdx != -1 {
+				ret[foundIdx] = combineSecretLocations(ret[foundIdx], currIn.Spec.SecretLocations)
+				ret[foundIdx] = combineCertOnDiskLocations(ret[foundIdx], currIn.Spec.OnDiskLocations)
```
deads2k, line 38: I know this is pre-existing, but it looks like a mistake for our purposes where knowing what is reflected on both cluster and and ondisk only is valuable.

pkg/certs/cert-inspection/certgraphanalysis/disk_certs.go
deads2k: > permissions and se linux labels as likely next steps. Followup fine.

A peer to https://github.com/openshift/library-go/blob/master/pkg/certs/cert-inspection/certgraphapi/types.go#L15 would allow a spot to store things like the the file metadata in a spot peer to the inresource names and annotations to allow similar consuming code to be written.

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
vrutkovs: Yeah, correct. This would happen when key is found before cert, so the name cannot be extracted. Instead of backfilling the details its easier to skip the key so that it would be found on next iteration.

I'll add a comment clarifying this

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -56,28 +81,44 @@ func combineCertOnDiskLocations(in *certgraphapi.CertKeyPair, rhs []certgraphapi
 			}
 		}
 		if !found {
+			// Store key to be merged into Cert
+			if len(curr.Cert.Path) == 0 {
```
deads2k, line 85: I still don't understand why we want to skip this.  if it's not equal, just add it

deads2k:
The output ended correct, but I expect to see more action in here in the future

/lgtm

PR #1614, opened 2023-11-09
OCPBUGS-23796: add pdbUnhealthyPodEvictionPolicy option to a GuardController 

atiratree:
optionally `pdbUnhealthyPodEvictionPolicy` can be set to `AlwaysAllow` to allows
eviction of unhealthy (not ready) pods even if there are no disruptions allowed
on a PodDisruptionBudget. This can help to drain/maintain a node and recover
without a manual intervention when multiple instances of nodes or pods are
misbehaving. Use this with caution, as this option can disrupt perspective pods
that have not yet had a chance to become healthy (ready).

Solves the following issue:

1. upgrade the cluster 
2. 2 or more kube-apiserver pod do not become online. Network access could be lost due to misconfiguration or wrong rhel update. We can simulate this with:
    - ssh into a node 
    -  run `iptables -A INPUT -p tcp --destination-port 6443 -j DROP`
3. 2 or more kube-apiserver-guard pods lose readiness
4. kube-apiserver-guard-pdb PDB blocks the drain because `status.currentHealthy` is less than `status.desiredHealthy`
5. it is not possible to drain the node without overriding eviction requests (forcefully deleting the guard pods)`
```
    evicting pod openshift-kube-apiserver/kube-apiserver-guard-ip-10-0-19-181.eu-north-1.compute.internal
    error when evicting pods/"kube-apiserver-guard-ip-10-0-19-181.eu-north-1.compute.internal" -n "openshift-kube-apiserver" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

By loosening the unhealthyPodEvictionPolicy on the pdb, we (or MCO) can drain these nodes and do a maintenance (e.g. fixing the networking).

### DIFF
--- a/pkg/operator/staticpod/controllers.go
+++ b/pkg/operator/staticpod/controllers.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/policy/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/kubernetes"
@@ -68,11 +69,12 @@ type staticPodOperatorControllerBuilder struct {
 	staticPodPrefix string
 
 	// guard infomation
-	operatorName               string
-	operatorNamespace          string
-	readyzPort                 string
-	readyzEndpoint             string
-	guardCreateConditionalFunc func() (bool, bool, error)
+	operatorName                  string
+	operatorNamespace             string
+	readyzPort                    string
+	readyzEndpoint                string
+	pdbUnhealthyPodEvictionPolicy *v1.UnhealthyPodEvictionPolicyType
+	guardCreateConditionalFunc    func() (bool, bool, error)
 }
 
 func NewBuilder(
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controllers.go
```diff
@@ -105,6 +106,10 @@ type Builder interface {
 	WithCustomInstaller(command []string, installerPodMutationFunc installer.InstallerPodMutationFunc) Builder
 	WithPruning(command []string, staticPodPrefix string) Builder
 	WithPodDisruptionBudgetGuard(operatorNamespace, operatorName, readyzPort, readyzEndpoint string, createConditionalFunc func() (bool, bool, error)) Builder
+	// WithAlwaysAllowUnhealthyPodEvictionPolicy allows eviction of unhealthy (not ready) pods even if there are no disruptions allowed on a PodDisruptionBudget.
+	// This can help to drain/maintain a node and recover without a manual intervention when multiple instances of nodes or pods are misbehaving.
+	// Use this with caution, as this option can disrupt perspective pods that have not yet had a chance to become healthy.
+	WithAlwaysAllowUnhealthyPodEvictionPolicy() Builder
```
atiratree, line 112: - This is meant to be an opt-in, but it can be also added to `WithPodDisruptionBudgetGuard` options.
- I am not allowing passing `UnhealthyPodEvictionPolicyType` reference, because we cannot be sure if new values are added later to the API in the future that could disrupt high availability. The guarding  of the pods should be tested before allowing it.

atiratree:
the option was tested here: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1579

pkg/operator/staticpod/controllers.go
soltysh: @atiratree let's move it to `WithPodDisruptionBudgetGuard`, since that's the PDB you are modifying here, so that seems like the best API for this.

atiratree: done, and using a UnhealthyPodEvictionPolicyType type now. I am also testing this change again in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1579

soltysh (approved):
/lgtm
/approve

PR #1618, opened 2023-11-14
certgraphanalysis: add SkipHashed option

vrutkovs:
First part of https://github.com/openshift/library-go/pull/1601. Testing in https://github.com/openshift/origin/pull/28378

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collector.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collector.go
@@ -86,26 +86,28 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
 			}
 
 			details, err := InspectConfigMap(&configMap)
-			if details != nil {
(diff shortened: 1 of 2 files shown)


pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -123,26 +125,27 @@ func gatherFilteredCerts(ctx context.Context, kubeClient kubernetes.Interface, a
...
+						Namespace: secret.Namespace,
+						Name:      secret.Name,
+					},
+					CertKeyInfo: certgraphapi.PKIRegistryCertKeyPairInfo{
+						OwningJiraComponent: secret.Annotations[annotations.OpenShiftComponent],
+						Description:         secret.Annotations[annotations.OpenShiftDescription],
+					},
+				})
 		}
 	}
```
deads2k, line 151: the for loops below this are no longer needed, right?  I think it can be a followup, but should be done.

PR #1620, opened 2023-11-16
OCPBUGS-18697: guard controller: re-create pods with phase nor pending neither running

ingvagabund:
E.g. to re-create completed pods produced by kubelet preemption

/hold until the change proved to be working

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -321,6 +321,10 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 					klog.V(5).Infof("Guard Hostname changed, deleting %v so the guard can be re-created", pod.Name)
 					delete = true
 				}
+				if actual.Status.Phase != "" && actual.Status.Phase != corev1.PodPending && actual.Status.Phase != corev1.PodRunning {
+					klog.V(5).Infof("Pod phase is neither pending nor running, deleting %v so the guard can be re-created", pod.Name)
+					delete = true
+				}
 				if delete {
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -321,6 +321,10 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 					klog.V(5).Infof("Guard Hostname changed, deleting %v so the guard can be re-created", pod.Name)
 					delete = true
 				}
+				if actual.Status.Phase != "" && actual.Status.Phase != corev1.PodPending && actual.Status.Phase != corev1.PodRunning {
+					klog.V(5).Infof("Pod phase is neither pending nor running, deleting %v so the guard can be re-created", pod.Name)
```
tkashem, line 325: do we need any retry backoff here? (what if due to some underlying condition, the phase is set to `Failed`, we delete the guard Pod and the Pod phase is set to `Failed` again)

ingvagabund: The sync context in the factory controller builder relies on `k8s.io/client-go/util/workqueue.NewNamedRateLimitingQueue` which utilizes `k8s.io/client-go/util/workqueue.NewNamedRateLimitingQueue.DefaultControllerRateLimiter`. With `// 10 qps, 100 bucket size` [1]. Not the same as backing off for a specific pod. At least some protection. Though 10 removals of a failed pod per second is quite a lot. 

[1] https://github.com/openshift/library-go/blob/9991651fb2494e5aa03d1a82d782b7d398a7862c/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L42

PR #1624, opened 2023-11-21
OCPBUGS-18697: guard controller: delay in some cases to avoid quick reconciliation

ingvagabund:
Follow up for https://github.com/openshift/library-go/pull/1620/files#r1399213346.

E.g. to avoid:
```
E1114 01:36:14.472523       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.472549       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.472568       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.484316       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.484341       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.484363       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.506314       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.506347       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.506366       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.548493       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.548531       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.548550       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.630401       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.630423       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.630448       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.792342       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.792363       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.792381       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
```
and log only:
```
E1114 01:36:14.472523       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.472549       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.472549       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
E1114 01:36:14.972523       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-0
E1114 01:36:14.972549       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-1
E1114 01:36:14.972549       1 guard_controller.go:271] Missing operand on node libvirt-s390x-2-1-4fc-dwwjq-master-2
```

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/policy/v1"
@@ -53,6 +54,9 @@ type GuardController struct {
 	// installerPodImageFn returns the image name for the installer pod
 	installerPodImageFn   func() string
 	createConditionalFunc func() (bool, bool, error)
+
+	delayTimeout  time.Duration
+	delayNextSync bool
 }
 
 func NewGuardController(
@@ -101,6 +105,7 @@ func NewGuardController(
 		pdbLister:                     kubeInformersForTargetNamespace.Policy().V1().PodDisruptionBudgets().Lister(),
 		installerPodImageFn:           getInstallerPodImageFromEnv,
 		createConditionalFunc:         createConditionalFunc,
+		delayTimeout:                  30 * time.Second,
 	}
 
 	return factory.New().WithInformers(
@@ -165,6 +170,14 @@ var podTemplate []byte
 func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	klog.V(5).Info("Syncing guards")
 
+	if c.delayNextSync {
+		c.delayNextSync = false
+		klog.V(5).Info("Delaying sync for %v to avoid quick reconciliation ", c.delayTimeout)
+		// Delay from the previous iteration so the controller can update the operator status in case of errors.
+		// Delaying at the end of the iteration will delay the operator status update.
+		time.Sleep(c.delayTimeout)
+	}
+
 	if c.createConditionalFunc == nil {
 		return fmt.Errorf("createConditionalFunc not set")
 	}
@@ -180,6 +193,7 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 	}
 
 	errs := []error{}
+	var errWithDelay bool
 	if !shouldCreate {
 		pdb := resourceread.ReadPodDisruptionBudgetV1OrDie(pdbTemplate)
 		pdb.ObjectMeta.Name = getGuardPDBName(c.podResourcePrefix)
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -361,5 +371,11 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		}
 	}
 
+	// For cases which take longer or never to reconcile wait for some time
+	// to avoid quick reconciliation
+	if errWithDelay {
+		time.Sleep(c.delayTimeout)
+	}
```
dgrisonnet, line 378: Two comments about that delay.

1. reconciliation should be asynchronous, so I would expect the controller to continue to reconcile regardless of the state of the previous sync, but I might be wrong
2. adding delay here, will delay the report of the degraded status from https://github.com/openshift/library-go/pull/1624/files#diff-93e47de48c603f6e2ddfe947c96610f13116e244dd7127845f8d23ea3d244c0dR112

I wonder if we have some precedence in library-go of doing something like that, usually I would expect having a work queue with some backoff

tkashem: yes, I agree with @dgrisonnet , usually the solution here is to add the key to the queue asynchronously with a delay. (i think work queue has support for this)
But the issue is, our controller is not key based, it is periodic (like every 1s or something)

ingvagabund: The guard controller's event handlers registers changes from pods under openshift-kube-scheduler and from nodes. When saturating all master nodes with system-node-critical priority class pods the guard pods start to be periodically terminated and re-created. For 3 master nodes cluster and with the priority and fairness enabled it takes approximately 10 seconds to run the sync method. In these 10 seconds the event handler for pods is invoked multiple times which marks the "key" item in the queue as dirty. Meaning, even when the "key" item is popped from the queue, it never gets processed without getting repeatedly enqueued back. Thus, the "key" item never leaves the queue for more than a second or two.

ingvagabund:
Closing in favor of https://github.com/openshift/library-go/pull/1652

PR #1625, opened 2023-11-22
CFE-882: Route external certificate validation

chiragkyal:
## Description
Updates route validations based on openshift/enhancements#1307 which introduces a new field in the route API externalCertificate behind the TP feature gate.
### Changes
- `ValidateRoute()` has been updated to pass additional args (secrets lister and subjectaccessreviewer) to validate
`externalCertificate` and the rbac required. Now it also does additional subject access review authorization checks to verify if the router SA has the correct permissions to parse `externalCertificate`. It also adds a hard requirement that the secret referenced in `externalCertificate` be present prior to creating the route.
- `ValidateHostUpdate()` has also been updated since updating route host/subdomain is also affected by updating
certificates on the route. Now `certificateChangeRequiresAuth()` will always return `true` if the route has a `externalCertificate` set (check EP for additional info).
- `ValidateHostExternalCertificate()` has been introduced as part of the validations done during route updates, this function specifically checks if the user has the correct permissions on the `custom-host` sub-resource. Additional details can be found on the EP.

### DIFF
--- a/pkg/route/common.go
+++ b/pkg/route/common.go
@@ -1,0 +1,60 @@
+package route
+
+import (
+	"context"
+	"fmt"
+
+	authorizationv1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/endpoints/request"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/library-go/pkg/authorization/authorizationutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SubjectAccessReviewCreator is an interface for performing subject access reviews
+type SubjectAccessReviewCreator interface {
+	Create(ctx context.Context, sar *authorizationv1.SubjectAccessReview, opts metav1.CreateOptions) (*authorizationv1.SubjectAccessReview, error)
+}
+
+// RouteValidationOptions used to tweak how/what fields are validated. These
+// options are propagated by the apiserver.
+type RouteValidationOptions struct {
+
+	// AllowExternalCertificates option is set when the RouteExternalCertificate
+	// feature gate is enabled.
+	AllowExternalCertificates bool
+}
+
+// CheckRouteCustomHostSAR checks if user has permission to create and update routes/custom-host
+// sub-resource
+func CheckRouteCustomHostSAR(ctx context.Context, fldPath *field.Path, sarc SubjectAccessReviewCreator) field.ErrorList {
+	user, ok := request.UserFrom(ctx)
+	if !ok {
+		return field.ErrorList{field.InternalError(fldPath, fmt.Errorf("unable to verify host field can be set"))}
+	}
+
+	var errs field.ErrorList
+	if err := authorizationutil.Authorize(sarc, user, &authorizationv1.ResourceAttributes{
+		Namespace:   request.NamespaceValue(ctx),
+		Verb:        "create",
+		Group:       routev1.GroupName,
+		Resource:    "routes",
+		Subresource: "custom-host",
+	}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "user does not have create permission on custom-host"))
+	}
+
+	if err := authorizationutil.Authorize(sarc, user, &authorizationv1.ResourceAttributes{
+		Namespace:   request.NamespaceValue(ctx),
+		Verb:        "update",
+		Group:       routev1.GroupName,
+		Resource:    "routes",
+		Subresource: "custom-host",
+	}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "user does not have update permission on custom-host"))
+	}
+
+	return errs
+}
--- a/pkg/route/hostassignment/assignment.go
+++ b/pkg/route/hostassignment/assignment.go
@@ -12,15 +12,11 @@ import (
 
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/openshift/library-go/pkg/authorization/authorizationutil"
+	"github.com/openshift/library-go/pkg/route"
 )
 
 // HostGeneratedAnnotationKey is the key for an annotation set to "true" if the route's host was generated
 const HostGeneratedAnnotationKey = "openshift.io/host.generated"
-
-// Registry is an interface for performing subject access reviews
-type SubjectAccessReviewCreator interface {
-	Create(ctx context.Context, sar *authorizationv1.SubjectAccessReview, opts metav1.CreateOptions) (*authorizationv1.SubjectAccessReview, error)
-}
 
 type HostnameGenerator interface {
 	GenerateHostname(*routev1.Route) (string, error)
@@ -29,9 +25,18 @@ type HostnameGenerator interface {
 // AllocateHost allocates a host name ONLY if the route doesn't specify a subdomain wildcard policy and
 // the host name on the route is empty and an allocator is configured.
 // It must first allocate the shard and may return an error if shard allocation fails.
-func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessReviewCreator, routeAllocator HostnameGenerator) field.ErrorList {
+func AllocateHost(ctx context.Context, route *routev1.Route, sarc route.SubjectAccessReviewCreator, routeAllocator HostnameGenerator, opts route.RouteValidationOptions) field.ErrorList {
 	hostSet := len(route.Spec.Host) > 0
-	certSet := route.Spec.TLS != nil && (len(route.Spec.TLS.CACertificate) > 0 || len(route.Spec.TLS.Certificate) > 0 || len(route.Spec.TLS.DestinationCACertificate) > 0 || len(route.Spec.TLS.Key) > 0)
+	certSet := route.Spec.TLS != nil &&
+		(len(route.Spec.TLS.CACertificate) > 0 ||
+			len(route.Spec.TLS.Certificate) > 0 ||
+			len(route.Spec.TLS.DestinationCACertificate) > 0 ||
+			len(route.Spec.TLS.Key) > 0)
+
+	if opts.AllowExternalCertificates && route.Spec.TLS != nil && route.Spec.TLS.ExternalCertificate != nil {
+		certSet = certSet || len(route.Spec.TLS.ExternalCertificate.Name) > 0
+	}
+
 	if hostSet || certSet {
 		user, ok := request.UserFrom(ctx)
 		if !ok {
@@ -86,41 +91,65 @@ func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessR
 	return nil
 }
 
-func hasCertificateInfo(tls *routev1.TLSConfig) bool {
+func hasCertificateInfo(tls *routev1.TLSConfig, opts route.RouteValidationOptions) bool {
 	if tls == nil {
 		return false
 	}
-	return len(tls.Certificate) > 0 ||
+	hasInfo := len(tls.Certificate) > 0 ||
 		len(tls.Key) > 0 ||
 		len(tls.CACertificate) > 0 ||
 		len(tls.DestinationCACertificate) > 0
+
+	if opts.AllowExternalCertificates && tls.ExternalCertificate != nil {
+		hasInfo = hasInfo || len(tls.ExternalCertificate.Name) > 0
+	}
+	return hasInfo
 }
 
-func certificateChangeRequiresAuth(route, older *routev1.Route) bool {
+// certificateChangeRequiresAuth determines whether changes to the TLS certificate configuration require authentication.
+// Note: If either route uses externalCertificate, this function always returns true, as we cannot definitively verify if
+// the content of the referenced secret has been modified. Even if the secret name remains the same,
+// we must assume that the secret content is changed, necessitating authentication.
+func certificateChangeRequiresAuth(route, older *routev1.Route, opts route.RouteValidationOptions) bool {
 	switch {
 	case route.Spec.TLS != nil && older.Spec.TLS != nil:
 		a, b := route.Spec.TLS, older.Spec.TLS
-		if !hasCertificateInfo(a) {
+		if !hasCertificateInfo(a, opts) {
 			// removing certificate info is allowed
 			return false
 		}
-		return a.CACertificate != b.CACertificate ||
+
+		certChanged := a.CACertificate != b.CACertificate ||
 			a.Certificate != b.Certificate ||
 			a.DestinationCACertificate != b.DestinationCACertificate ||
 			a.Key != b.Key
+
+		if opts.AllowExternalCertificates {
+			if route.Spec.TLS.ExternalCertificate != nil || older.Spec.TLS.ExternalCertificate != nil {
+				certChanged = true
+			}
+		}
+
+		return certChanged
 	case route.Spec.TLS != nil:
 		// using any default certificate is allowed
-		return hasCertificateInfo(route.Spec.TLS)
+		return hasCertificateInfo(route.Spec.TLS, opts)
 	default:
 		// all other cases we are not adding additional certificate info
 		return false
 	}
 }
 
-func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc SubjectAccessReviewCreator) field.ErrorList {
+// ValidateHostUpdate checks if the user has the correct permissions based on the updates
+// done to the route object. If the route's host/subdomain has been updated it checks if
+// the user has "update" permission on custom-host subresource. If only the certificate
+// has changed, it checks if the user has "create" permission on the custom-host subresource.
+// Caveat here is that if the route uses externalCertificate, the certChanged condition will
+// always be true since we cannot verify state of external secret object.
+func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc route.SubjectAccessReviewCreator, opts route.RouteValidationOptions) field.ErrorList {
 	hostChanged := route.Spec.Host != older.Spec.Host
 	subdomainChanged := route.Spec.Subdomain != older.Spec.Subdomain
-	certChanged := certificateChangeRequiresAuth(route, older)
+	certChanged := certificateChangeRequiresAuth(route, older, opts)
 	if !hostChanged && !certChanged && !subdomainChanged {
 		return nil
 	}
@@ -190,6 +219,14 @@ func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc S
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Certificate, older.Spec.TLS.Certificate, field.NewPath("spec", "tls", "certificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.DestinationCACertificate, older.Spec.TLS.DestinationCACertificate, field.NewPath("spec", "tls", "destinationCACertificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Key, older.Spec.TLS.Key, field.NewPath("spec", "tls", "key"))...)
+
+			if opts.AllowExternalCertificates {
+				if route.Spec.TLS.ExternalCertificate == nil || older.Spec.TLS.ExternalCertificate == nil {
+					errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate, older.Spec.TLS.ExternalCertificate, field.NewPath("spec", "tls", "externalCertificate"))...)
+				} else {
+					errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"))...)
+				}
+			}
 			return errs
 		}
 	}
--- a/pkg/route/hostassignment/assignment_test.go
+++ b/pkg/route/hostassignment/assignment_test.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/request"
 
 	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/library-go/pkg/route"
 )
 
 type testAllocator struct {
@@ -51,6 +52,9 @@ func TestHostWithWildcardPolicies(t *testing.T) {
 
 		expected          string
 		expectedSubdomain string
+
+		// field for externalCertificate
+		opts route.RouteValidationOptions
 
 		errs  int
 		allow bool
@@ -216,6 +220,138 @@ func TestHostWithWildcardPolicies(t *testing.T) {
 			errs:           1,
 		},
 		{
+			name:           "no-certificate-changed-to-external-certificate-denied",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           1,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "no-certificate-changed-to-external-certificate-allowed",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          true,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "external-certificate-changed-to-certificate-denied",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           2,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "external-certificate-changed-to-certificate-allowed",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          true,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "certificate-changed-to-external-certificate-denied",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           2,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "certificate-changed-to-external-certificate-allowed",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          true,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:     "certificate-changed-to-external-certificate-allowed-but-featuregate-is-not-set",
+			host:     "host",
+			expected: "host",
+			oldHost:  "host",
+			// if the featuregate was disabled, and ExternalCertificate wasn't previously set, apiserver will strip ExternalCertificate field.
+			// https://github.com/openshift/openshift-apiserver/blob/1fac5e7e3a6153efae875185af2dba48fbad41ab/pkg/route/apiserver/registry/route/strategy.go#L73-L93
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: nil},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          true,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: false},
+		},
+		{
+			name:           "external-certificate-changed-denied",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "old-b"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           1,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "external-certificate-changed-allowed",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "old-b"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          true,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "external-certificate-secret-unchanged",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
 			name:           "ca-certificate-unchanged",
 			host:           "host",
 			expected:       "host",
@@ -368,9 +504,9 @@ func TestHostWithWildcardPolicies(t *testing.T) {
 						},
 					},
 				}
-				errs = ValidateHostUpdate(ctx, route, oldRoute, &testSAR{allow: tc.allow})
+				errs = ValidateHostUpdate(ctx, route, oldRoute, &testSAR{allow: tc.allow}, tc.opts)
 			} else {
-				errs = AllocateHost(ctx, route, &testSAR{allow: tc.allow}, testAllocator{})
+				errs = AllocateHost(ctx, route, &testSAR{allow: tc.allow}, testAllocator{}, tc.opts)
 			}
 
 			if route.Spec.Host != tc.expected {
--- a/pkg/route/hostassignment/externalcertificate.go
+++ b/pkg/route/hostassignment/externalcertificate.go
@@ -1,0 +1,35 @@
+package hostassignment
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	routev1 "github.com/openshift/api/route/v1"
+	routecommon "github.com/openshift/library-go/pkg/route"
+)
+
+// ValidateHostExternalCertificate checks if the user has permissions to create and update
+// custom-host subresource of routes. This check is required to be done prior to ValidateHostUpdate()
+// since updating hosts while using externalCertificate is contingent on the user having both these
+// permissions. The ValidateHostUpdate() cannot differentiate if the certificate has changed since
+// now the certificates will be present as a secret object, due to this it proceeds with the assumption
+// that the certificate has changed when the route has externalCertificate set.
+// TODO: Consider merging this function into ValidateHostUpdate.
+func ValidateHostExternalCertificate(ctx context.Context, new, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, opts routecommon.RouteValidationOptions) field.ErrorList {
+
+	if !opts.AllowExternalCertificates {
+		// Return nil since the feature gate is off.
+		// ValidateHostUpdate() is sufficient to validate
+		// permissions.
+		return nil
+	}
+
+	newTLS := new.Spec.TLS
+	oldTLS := older.Spec.TLS
+	if (newTLS != nil && newTLS.ExternalCertificate != nil) || (oldTLS != nil && oldTLS.ExternalCertificate != nil) {
+		return routecommon.CheckRouteCustomHostSAR(ctx, field.NewPath("spec", "tls", "externalCertificate"), sarc)
+	}
+
+	return nil
+}
--- a/pkg/route/hostassignment/externalcertificate_test.go
+++ b/pkg/route/hostassignment/externalcertificate_test.go
@@ -1,0 +1,340 @@
+package hostassignment
+
+import (
+	"context"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/endpoints/request"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	routev1 "github.com/openshift/api/route/v1"
+	routecommon "github.com/openshift/library-go/pkg/route"
+)
+
+func TestValidateHostExternalCertificate(t *testing.T) {
+	tests := []struct {
+		name  string
+		ctx   context.Context
+		new   *routev1.Route
+		old   *routev1.Route
+		allow bool
+		opts  routecommon.RouteValidationOptions
+		want  field.ErrorList
+	}{
+		{
+			name: "Updating new route and old route with nil TLS",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			allow: false,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want:  nil,
+		},
+		{
+			name: "Updating new route with nil TLS",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						Certificate: "old-cert",
+					},
+				},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			allow: false,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want:  nil,
+		},
+		{
+			name: "Updating new route with certificate from nil TLS",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						Certificate: "new-cert",
+					},
+				},
+			},
+			allow: false,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want:  nil,
+		},
+		{
+			name: "Updating new route with externalCertificate from nil TLS without permission",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "dummy-cert",
+						},
+					},
+				},
+			},
+			allow: false,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want: field.ErrorList{
+				field.Forbidden(nil, "somedetail"),
+				field.Forbidden(nil, "somedetail"),
+			},
+		},
+		{
+			name: "Updating new route with externalCertificate from nil TLS with permission",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "dummy-cert",
+						},
+					},
+				},
+			},
+			allow: true,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want:  nil,
+		},
+		{
+			name: "Updating new route with nil TLS from externalCertificate without permission",
+			ctx:  request.WithUser(context.Background(), &user.DefaultInfo{Name: "user1"}),
+			old: &routev1.Route{
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "dummy-cert",
+						},
+					},
+				},
+			},
+			new: &routev1.Route{
+				Spec: routev1.RouteSpec{},
+			},
+			allow: false,
+			opts:  routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			want: field.ErrorList{
+				field.Forbidden(nil, "somedetail"),
+				field.Forbidden(nil, "somedetail"),
+			},
+		},
+		{
+			name: "Updating new route with nil TLS from externalCertificate with permission",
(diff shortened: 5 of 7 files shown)


chiragkyal:
Continued from https://github.com/openshift/library-go/pull/1549

pkg/route/validation/validation.go
```diff
@@ -70,8 +73,8 @@ var (
 	permittedResponseHeaderValueRE = regexp.MustCompile(strings.Replace(permittedHeaderValueTemplate, "XYZ", "res", 1))
 )
 
-func ValidateRoute(route *routev1.Route) field.ErrorList {
-	return validateRoute(route, true)
+func ValidateRoute(ctx context.Context, route *routev1.Route, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
```
alebedev87, line 76: Passing `context.Context` as the first parameter is go idiomatic, I agree on this. But do we pass a meaningful context from the api server or just `context.Background()`/`context.TODO()`?

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
 	return result
 }
 
+// validateTLSExternalCertificate
```
alebedev87, line 306:
```suggestion
// validateTLSExternalCertificate <comment>
```

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+func validateTLSExternalCertificate(ctx context.Context, route *routev1.Route, fldPath *field.Path, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter) field.ErrorList {
+	tls := route.Spec.TLS
+	var errs field.ErrorList
+
+	errs = append(errs, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarCreator)...)
+
+	if err := authorizationutil.Authorize(sarCreator, &user.DefaultInfo{Name: "system:serviceaccount:openshift-ingress:router"},
```
alebedev87, line 313: Nit pic: the router service account can be a constant.

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
 	return result
 }
 
+// validateTLSExternalCertificate
+func validateTLSExternalCertificate(ctx context.Context, route *routev1.Route, fldPath *field.Path, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter) field.ErrorList {
+	tls := route.Spec.TLS
+	var errs field.ErrorList
+
+	errs = append(errs, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarCreator)...)
```
alebedev87, line 311:
```suggestion
	errs := append(errs, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarCreator)...)
```

pkg/route/validation/validation.go
```diff
@@ -354,6 +303,58 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+			Resource:  "secrets",
+			Name:      tls.ExternalCertificate.Name,
+		}); err != nil {
+		errs = append(errs, field.Forbidden(fldPath, "router serviceaccount does not have permission to list this secret"))
+	}
+
+	secret, err := secretsGetter.Secrets(route.Namespace).Get(ctx, tls.ExternalCertificate.Name, metav1.GetOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
+		errs = append(errs, field.NotFound(fldPath, err))
+		return errs
```
alebedev87, line 346:
```suggestion
		return append(errs, field.NotFound(fldPath, err))
```

pkg/route/validation/validation_test.go
```diff
@@ -1800,6 +1867,13 @@ func TestValidateTLS(t *testing.T) {
 		name           string
 		route          *routev1.Route
 		expectedErrors int
+
+		// fields for externalCertificate
+		allow     bool
+		validType bool
+		validNS   bool
```
alebedev87, line 1874: Why obfuscating the test case structure by adding these indirections which are used to build a `Secret`? Why not pass a `Secret` as a whole, similar to `Route`?

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2020,162 @@ func TestValidateTLS(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			allow:          true,
+			validNS:        true,
+			validType:      false,
+			opts:           routecommon.RouteValidationOptions{AllowExternalCertificates: true},
 			expectedErrors: 1,
 		},
```
alebedev87, line 2154: I see only error cases for the new code, we need some non error test cases (when `expectedErrors: 0`).

pkg/route/hostassignment/assignment_test.go
```diff
@@ -52,6 +53,13 @@ func TestHostWithWildcardPolicies(t *testing.T) {
 		expected          string
 		expectedSubdomain string
 
+		// fields for externalCertificate
+		validType  bool
+		secretData map[string]string
+		validNS    bool
```
alebedev87, line 59: Where are these fields used?

pkg/route/hostassignment/externalcertificate_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	routev1 "github.com/openshift/api/route/v1"
+	routecommon "github.com/openshift/library-go/pkg/route"
+)
+
+func TestValidateHostExternalCertificate(t *testing.T) {
+
+	type args struct {
```
alebedev87, line 20: I don't see a reason for a dedicated structure for the function arguments. The idiomatic go table tests pass args as fields of `testCase` struct.

pkg/route/hostassignment/externalcertificate_test.go
```diff
@@ -0,0 +1,209 @@
...
+		ctx   context.Context
+		new   *routev1.Route
+		older *routev1.Route
```
alebedev87, line 23:
```suggestion
		old *routev1.Route
```

alebedev87 (commented):
Focus of this review was on the public functions and their signatures as they should be concise and meaningful. Also I took a glance at the test cases t understand the new logic but the test cases need some structural changes. Will have to do another re-iteration with the focus on the logic.

pkg/route/hostassignment/assignment.go
```diff
@@ -86,41 +91,61 @@ func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessR
...
-		return hasCertificateInfo(route.Spec.TLS)
+		return hasCertificateInfo(route.Spec.TLS, opts)
 	default:
 		// all other cases we are not adding additional certificate info
 		return false
 	}
 }
 
-func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc SubjectAccessReviewCreator) field.ErrorList {
+// validateHostUpdate checks if the user has the correct permissions based on the updates
```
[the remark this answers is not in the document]
chiragkyal: Done

pkg/route/validation/validation.go
chiragkyal: Added comment, thanks!

pkg/route/validation/validation.go
chiragkyal: Yeah, agree! Done

pkg/route/validation/validation.go
chiragkyal: Done

pkg/route/validation/validation_test.go
chiragkyal: Thanks for the suggestion. I've updated the structure of the testCases

pkg/route/validation/validation_test.go
chiragkyal: Thanks for the catch, added new test (`Valid Edge termination with externalCertificate`)  where `expectedErrors: 0`

pkg/route/hostassignment/assignment_test.go
chiragkyal: lol :)  those were not used anywhere. Removed, thanks

pkg/route/hostassignment/externalcertificate_test.go
chiragkyal: I think the thought process could be to make the testCases look a bit more structured, but I agree that go table tests pass args as fields of testCase struct. I've updated them, thanks. 
I've also added a few more tests to cover other scenarios.

pkg/route/hostassignment/externalcertificate_test.go
chiragkyal: Done

pkg/route/validation/validation.go
chiragkyal: I think this will cause `undefined: errs` inside append.

pkg/route/validation/validation.go
chiragkyal: I think it would be a meaningful context, and not just `context.Background()`/`context.TODO()`. This will be the same context which was previously getting passed in `AllocateHost` and `ValidateHostUpdate`  from api server.

 https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apiserver/registry/route/strategy.go#L95-L121

pkg/route/common.go
```diff
@@ -0,0 +1,60 @@
...
+	"k8s.io/apiserver/pkg/endpoints/request"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"github.com/openshift/library-go/pkg/authorization/authorizationutil"
+)
+
+// SubjectAccessReviewCreator is an interface for performing subject access reviews
+type SubjectAccessReviewCreator interface {
+	Create(ctx context.Context, sar *authorizationv1.SubjectAccessReview, opts metav1.CreateOptions) (*authorizationv1.SubjectAccessReview, error)
+}
```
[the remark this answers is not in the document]
chiragkyal: I too, am honestly not sure about the earlier "reason of being" for keeping this same as [client-go's SubjectAccessReviewInterface](https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/authorization/v1#SubjectAccessReviewInterface). I'm fine removing this and directly using client-go. WDYT, should we?

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,34 @@
...
+// custom-host subresource of routes. This check is required to be done prior to ValidateHostUpdate()
+// since updating hosts while using externalCertificate is contingent on the user having both these
+// permissions. The ValidateHostUpdate() cannot differentiate if the certificate has changed since
+// now the certificates will be present as a secret object, due to this it proceeds with the assumption
+// that the certificate has changed when the route has externalCertificate set.
+func ValidateHostExternalCertificate(ctx context.Context, new, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, opts routecommon.RouteValidationOptions) field.ErrorList {
+	newTLS := new.Spec.TLS
+	oldTLS := older.Spec.TLS
+
+	if !opts.AllowExternalCertificates {
```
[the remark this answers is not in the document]
chiragkyal: I think `ValidateHostExternalCertificate` is the [only place](https://github.com/openshift/openshift-apiserver/pull/407/files#diff-3c95009cb726729b48ee6ff52d62e70f3352aa6af1a0dedd2047ecdc5a398ce9R125-R127)  where apiserver checks for featuregate, rest places it relies on `RouteValidationOptions` struct to decide inside library go. 

I see challenges in removing this:

- `AllocateHost` and `ValidateHostUpdate` also rely on this struct and take decisions accordingly, how are we going to handle logic for them?
- If we want to add new methods for route validation, then I think we will need to have 2 methods, for example `ValidateRouteExternalCertificate` and `ValidateRouteUpdateExternalCertificate` in library-go as well as in [api-server's validation.go](https://github.com/openshift/openshift-apiserver/blob/master/pkg/route/apis/route/validation/validation.go). I think it can cause more overhead in longer run, and prone to future refactoring. 

IMHO, we should rather remove the feature check in api-server and let the flow decide based upon `RouteValidationOptions` here in library-go to remove redundancy. (all the functions are already doing that).

pkg/route/common.go
```diff
@@ -0,0 +1,60 @@
...
+// options are propagated by the apiserver.
+type RouteValidationOptions struct {
+
+	// AllowExternalCertificates option is set when RouteExternalCertificate
+	AllowExternalCertificates bool
+}
+
+// CheckRouteCustomHostSAR checks if user has permission to create and update routes/custom-host
+// sub-resource
+func CheckRouteCustomHostSAR(ctx context.Context, fldPath *field.Path, sarc SubjectAccessReviewCreator) field.ErrorList {
```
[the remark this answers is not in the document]
chiragkyal: Please see https://github.com/openshift/library-go/pull/1625#discussion_r1452080004 for keeping `RouteValidationOptions`

pkg/route/validation/validation.go
alebedev87: Yeah, you are right, my bad. You can do something like:

```go
errs := append(field.ErrorList{}, routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarCreator)...)
```

or simply:
```go
errs := routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarCreator)
```

pkg/route/validation/validation_test.go
alebedev87: I still see `allow` `validType` and `validNS` fields.

pkg/route/validation/validation_test.go
alebedev87: Where is it? I cannot see it.

pkg/route/validation/validation.go
alebedev87: I have an impression that you may have forgotten to push some commits because I see resolved comments which are not actually addressed.

alebedev87 (commented):
I see resolved comments which are said to be addressed but they are actually not. Maybe you forgot to push the latest changes? If yes, can you please do so? Thanks!

chiragkyal:
@alebedev87 that's weird, I do have the latest changes pushed. Can you access https://github.com/openshift/library-go/pull/1625/commits/bb12c499ed386faa3527a37a6ca09a64313c7579 commit and let me know please, if it's visible.

pkg/route/validation/validation.go
chiragkyal: Can you please try accessing https://github.com/openshift/library-go/pull/1625/commits/bb12c499ed386faa3527a37a6ca09a64313c7579 and let me know?

alebedev87:
@chiragkyal : I can see things now, seems like a GitHub glitch as I refreshed the PR many times. Sorry about that, I continue the review!

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,58 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
```
alebedev87, line 273: Can you please add test cases with `allow: true`?

alebedev87 (commented):
Another iteration, not fully done with the review yet.

pkg/route/common.go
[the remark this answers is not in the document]
chiragkyal: Sure, removed this interface and using client go's one

pkg/route/validation/validation.go
chiragkyal: yeah, updated to
```go
errs := routecommon.CheckRouteCustomHostSAR(ctx, fldPath, sarc)
```

pkg/route/validation/validation_test.go
```diff
@@ -42,6 +52,63 @@ func createRouteSpecTo(name string, kind string) routev1.RouteTargetReference {
...
+	t.secrets = append(t.secrets, secret)
+	return t
+}
+
+func (t *testSecretGetter) WithNamespace(namespace string) *testSecretGetter {
+	t.namespace = namespace
+	return t
+}
+
+func (t *testSecretGetter) WithSecretData(route *routev1.Route, data map[string]string) *testSecretGetter {
```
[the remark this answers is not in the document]
chiragkyal: Agreed! Removed all `WithX()`

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2018,296 @@ func TestValidateTLS(t *testing.T) {
...
 
+	ctx := request.WithUser(context.Background(), &user.DefaultInfo{})
 	for _, tc := range tests {
-		errs := validateTLS(tc.route, nil)
+
+		secrets := &testSecretGetter{}
+		secrets.WithNamespace(tc.route.Namespace)
+		secrets.WithSecret(tc.secret)
+
+		errs := validateTLS(ctx, tc.route, nil, &testSARCreator{allow: tc.allow}, secrets, tc.opts)
```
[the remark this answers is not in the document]
chiragkyal: Sure, I've removed all `WithX()` functions and now using 
```go
errs := validateTLS(ctx, tc.route, nil, &testSARCreator{allow: tc.allow}, &testSecretGetter{namespace: tc.route.Namespace, secret: tc.secret}, tc.opts)
```

pkg/route/validation/validation_test.go
```diff
@@ -42,6 +52,63 @@ func createRouteSpecTo(name string, kind string) routev1.RouteTargetReference {
...
+	t.sar = subjectAccessReview
+	return &authorizationv1.SubjectAccessReview{
+		Status: authorizationv1.SubjectAccessReviewStatus{
+			Allowed: t.allow,
+		},
+	}, t.err
+}
+
+type testSecretGetter struct {
+	getter    corev1client.SecretsGetter
```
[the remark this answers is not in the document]
chiragkyal: No, it is not used anywhere, good catch, thanks! Removed it.

pkg/route/validation/validation_test.go
```diff
@@ -42,6 +52,63 @@ func createRouteSpecTo(name string, kind string) routev1.RouteTargetReference {
...
+	secrets   []*corev1.Secret
```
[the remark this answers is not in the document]
chiragkyal: Agreed! One secret should suffice. Updated to 
```go
type testSecretGetter struct {
	namespace string
	secret    *corev1.Secret
}
```

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2018,296 @@ func TestValidateTLS(t *testing.T) {
...
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "tls-secret",
+						},
+					},
+				},
+			},
+			opts:           routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			expectedErrors: 1,
+		},
+		{
```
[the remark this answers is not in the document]
chiragkyal: Sure, moved this test after that logical grouping

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2018,296 @@ func TestValidateTLS(t *testing.T) {
...
+			},
+			secret: &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "tls-secret",
+					Namespace: "sandbox",
+				},
+				Type: corev1.SecretTypeTLS,
+			},
+			allow:          true,
+			expectedErrors: 0,
```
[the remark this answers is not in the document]
chiragkyal: My bad! It should have been set. Updated it. Thanks!

pkg/route/validation/validation_test.go
```diff
@@ -1946,12 +2018,296 @@ func TestValidateTLS(t *testing.T) {
 					},
 				},
 			},
+
+			expectedErrors: 1,
+		},
+		{
+			name: "Invalid Reencrypt termination as externalCertificate and certificate set",
```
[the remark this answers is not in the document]
chiragkyal: Updated to `route` for all the test cases

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,58 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+		{
+			name:           "external-certificate-secret-changed",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
```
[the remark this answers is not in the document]
chiragkyal: The is to test `unchaged` scenario. I've updated the test name to `external-certificate-secret-unchanged`. Hope it looks better now.

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,58 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "old-b"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           1,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:           "external-certificate-secret-changed",
```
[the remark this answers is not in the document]
chiragkyal: Updated the test name to `external-certificate-secret-unchanged`. Hope it looks better now.

pkg/route/hostassignment/assignment_test.go
chiragkyal: Added a few tests with `allow: true` and renamed the other tests with `denied/allowed`

pkg/route/hostassignment/externalcertificate.go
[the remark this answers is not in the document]
chiragkyal: I didn't get it properly. We should keep `opts.AllowExternalCertificates` here in library-go and populate this from api-server.  
Api-server will not decide the flow; rather it will rely on this field to decide the flow in library-go. 

This will give more control and less code refactoring in the future.

alebedev87 (commented):
Last iteration, some questions and remarks. Overall I think I'm close to the end :)

pkg/route/hostassignment/assignment.go
```diff
@@ -86,41 +92,61 @@ func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessR
...
-		return a.CACertificate != b.CACertificate ||
+
+		certChanged := a.CACertificate != b.CACertificate ||
 			a.Certificate != b.Certificate ||
 			a.DestinationCACertificate != b.DestinationCACertificate ||
 			a.Key != b.Key
+
+		if opts.AllowExternalCertificates && route.Spec.TLS != nil && older.Spec.TLS != nil {
+			if route.Spec.TLS.ExternalCertificate != nil || older.Spec.TLS.ExternalCertificate != nil {
+				certChanged = true
```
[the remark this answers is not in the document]
chiragkyal: Because we cannot definitively verify if the content of the secret that is referenced has been modified.

The reason this also mentioned in the EP:

If the old route or the new route uses `.spec.tls.externalCertificate` this validation will always have the precondition [certificateChangeRequiresAuth()](https://github.com/openshift/library-go/blob/d8d3f3f8a9e4a82c110a89a13229ce1412a88e4a/pkg/route/hostassignment/assignment.go#L123C29-L123C29) return true since we cannot definitively verify if the content of the secret that is referenced has been modified. Since the previous validation func (ValidateHostExternalCertificate()) would have already validation user permissions, we can safely make this assumption.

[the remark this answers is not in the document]
alebedev87: I was wondering abou the secret name, not the secret contents.

pkg/route/validation/validation_test.go
```diff
@@ -1946,13 +1990,293 @@ func TestValidateTLS(t *testing.T) {
...
+						},
+						Certificate: "dummy",
+					},
+				},
+			},
+			opts:           routecommon.RouteValidationOptions{AllowExternalCertificates: true},
+			expectedErrors: 1,
+		},
+		{
+			name: "Invalid Passthrough route as externalCertificate and certificate set",
```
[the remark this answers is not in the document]
chiragkyal: >passthrough route cannot have any certificate.

Agreed with that, but I think this test is meant to test false positives, to ensure it error'd out. I saw we have an existing test `Passthrough termination, cert` to test similar false positives.

pkg/route/validation/validation.go
```diff
@@ -228,85 +234,9 @@ func ValidateRouteStatusUpdate(route *routev1.Route, older *routev1.Route) field
...
-	block = &pem.Block{
-		Type: "PUBLIC KEY",
-	}
-	if block.Bytes, err = x509.MarshalPKIXPublicKey(key); err != nil {
-		return nil, err
-	}
-	return block, nil
-}
-
-func certificateBlockVerifier(block *pem.Block) (*pem.Block, error) {
```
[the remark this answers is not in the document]
chiragkyal: I didn't find any usage of these codes. Are they used somewhere?  Maybe these are some previous leftover items. 
Do you foresee any reason to keep them here?

pkg/route/common.go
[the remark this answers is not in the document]
alebedev87: Fine to close this one.

pkg/route/common.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// CheckRouteCustomHostSAR checks if user has permission to create and update routes/custom-host
+// sub-resource
+func CheckRouteCustomHostSAR(ctx context.Context, fldPath *field.Path, sarc authorizationclient.SubjectAccessReviewInterface) field.ErrorList {
+
+	var errs field.ErrorList
+	user, ok := request.UserFrom(ctx)
+	if !ok {
+		return field.ErrorList{field.InternalError(fldPath, fmt.Errorf("unable to verify access"))}
```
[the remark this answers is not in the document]
alebedev87: > We can update the error message only.

Right, I meant only the error message.

pkg/route/hostassignment/assignment_test.go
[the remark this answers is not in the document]
alebedev87: Yes, thanks!

pkg/route/validation/validation.go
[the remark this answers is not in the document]
alebedev87: > Are they used somewhere?

I don't see where it is (or was) used. It's a good idea to remove unused code indeed.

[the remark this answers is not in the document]
alebedev87: Ideally would be to put this into a dedicated commit but I don't insist.

pkg/route/common.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	if err := authorizationutil.Authorize(sarc, user, &authorizationv1.ResourceAttributes{
```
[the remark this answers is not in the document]
chiragkyal: Thanks, updated

pkg/route/hostassignment/assignment.go
```diff
@@ -12,26 +12,32 @@ import (
...
-
 type HostnameGenerator interface {
 	GenerateHostname(*routev1.Route) (string, error)
 }
 
 // AllocateHost allocates a host name ONLY if the route doesn't specify a subdomain wildcard policy and
 // the host name on the route is empty and an allocator is configured.
 // It must first allocate the shard and may return an error if shard allocation fails.
-func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessReviewCreator, routeAllocator HostnameGenerator) field.ErrorList {
+func AllocateHost(ctx context.Context, route *routev1.Route, sarc authorizationclient.SubjectAccessReviewInterface, routeAllocator HostnameGenerator, opts route.RouteValidationOptions) field.ErrorList {
```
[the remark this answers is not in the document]
chiragkyal: No Problem! Got the idea. Added back.

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
alebedev87: The problem with this test case is that it's same as `Passthrough termination, cert` which existed before your PR. The certificate field is checked before the externalCertificate and there is no dedicated `if` condition for both certificates set at the same time.

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,97 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "a"}},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
```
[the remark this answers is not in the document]
chiragkyal: Added a test `"certificate-changed-from-external-certificate-allowed-but-featuregate-is-not-set",`

pkg/route/validation/validation.go
```diff
@@ -342,6 +287,17 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
 		if len(tls.DestinationCACertificate) > 0 {
 			result = append(result, field.Invalid(fldPath.Child("destinationCACertificate"), "redacted destination ca certificate data", "edge termination does not support destination certificates"))
 		}
+
+		if opts.AllowExternalCertificates && tls.ExternalCertificate != nil {
+			if len(tls.Certificate) > 0 && len(tls.ExternalCertificate.Name) > 0 {
+				result = append(result, field.Invalid(fldPath.Child("externalCertificate"), tls.ExternalCertificate.Name, "cannot specify both tls.certificate and tls.externalCertificate"))
+				break
```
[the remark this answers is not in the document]
chiragkyal: Good catch! this `break` is not doing any extra thing, we can remove it. 
Removed.

pkg/route/validation/validation_test.go
```diff
@@ -1946,13 +1990,293 @@ func TestValidateTLS(t *testing.T) {
...
+						},
+						Certificate: "dummy",
+					},
+				},
+			},
+			opts:           routecommon.RouteValidationOptions{AllowExternalCertificates: true},
 			expectedErrors: 1,
 		},
+		{
+			name: "Invalid Passthrough route as externalCertificate",
```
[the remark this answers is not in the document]
chiragkyal: Done

pkg/route/validation/validation.go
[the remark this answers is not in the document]
chiragkyal: Sure, I'll put this into a separate commit, when I'll squash everything. (after your lgtm :) )

pkg/route/validation/validation_test.go
```diff
@@ -1946,13 +1990,293 @@ func TestValidateTLS(t *testing.T) {
...
+					},
+				},
+			},
+			secret: &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "tls-secret",
+					Namespace: "sandbox",
+				},
+				Type: corev1.SecretTypeTLS,
+			},
```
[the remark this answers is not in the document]
chiragkyal: `secret` field retrival will happen after SAR verification, true,  but the function will not return after SAR verification, it will continue and will add another error in the list: `secret not found`. 
So, in total 6 errors will be returned, which is not we are expecting in this test case of unauthorized access.

pkg/route/validation/validation_test.go
```diff
@@ -1946,13 +1990,293 @@ func TestValidateTLS(t *testing.T) {
...
+				Spec: routev1.RouteSpec{
+					TLS: &routev1.TLSConfig{
+						Termination: routev1.TLSTerminationReencrypt,
+						ExternalCertificate: &routev1.LocalObjectReference{
+							Name: "tls-secret",
+						},
+					},
+				},
+			},
+			secret: &corev1.Secret{
```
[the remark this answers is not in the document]
chiragkyal: Same problem as https://github.com/openshift/library-go/pull/1625#discussion_r1483123795

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
chiragkyal: The code will go through all the `if` conditions, no matter  `certificate` field is present or not, it will check `externalCertificate` 
So, for this test case `expectedErrors: 2` unlike `expectedErrors: 1` for `Passthrough termination, cert` test. 

I'm fine removing it though, it's just extra bit of testing, still relevant IMHO. wdyt ?

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
alebedev87: Right, I forgot the validation goes on. Fine to close this one.

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
alebedev87: Fine to close this one too.

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
alebedev87: I see what you mean. In that case why not setting all the fields which are checked (`CACertificate`, `Key` etc.)? `Certificate` field is no different from any of them. That may be misleading as the test name suggests like there is special handling when `Certificate` and `ExternalCertificate` are set like in the edge and reencrypt cases.

I think that more reasonable would be to test the setting of `ExternalCertificate` alone without `Certificate`because this was the new code introduced. Adding a test for all the `if`s of the passthrough route can be a bonus test case.

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,110 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+		{
+			name:           "certificate-changed-from-external-certificate-allowed-but-featuregate-is-not-set",
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "a"},
```
alebedev87, line 280: Doesn't it make more sense to set `ExternalCertificate` for both old and current TLS configs? Because `Certificate` field is not used in the condition you added to `ValidateHostUpdate`:
```
if opts.AllowExternalCertificates && route.Spec.TLS.ExternalCertificate != nil && older.Spec.TLS.ExternalCertificate != nil {
  errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"))...)
}
```

Btw does it mean that the external certificate field is not immutable anymore if the featuregate is not set?

pkg/route/hostassignment/externalcertificate.go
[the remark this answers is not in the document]
alebedev87: I agree on letting library-go decide on the flow. What I'm wondering about is why there is a dedicated method for the external certificate, why this code is not integrated into `ValidateHostUpdate`.

pkg/route/hostassignment/assignment_test.go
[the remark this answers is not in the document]
alebedev87: Posted [a question](https://github.com/openshift/library-go/pull/1625#discussion_r1483712146) about this new test.

pkg/route/hostassignment/assignment.go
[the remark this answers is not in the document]
alebedev87: Can the external certificate secret name change?

pkg/route/hostassignment/externalcertificate.go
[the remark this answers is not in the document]
chiragkyal: I think it had been touched upon during EP review, and also documented [here](https://github.com/thejasn/enhancements/blob/d4c7701193646105e2c60b2b996047ddfed083a4/enhancements/ingress/route-secret-injection-for-external-certificate-management.md#implementation-detailsnotesconstraints-optional). Let me know if that has changed afterwards.

pkg/route/hostassignment/assignment_test.go
chiragkyal: We can keep `ExternalCertificate` for both old and current TLS configs, but in any case `hasCertificateInfo()` will return `false` if the featuregate is not set; which means `certChanged` will be `false` and `ValidateHostUpdate` will return with `nil`.

So, if featuregate is not set and new TLS is using `ExternalCertificate` , `ValidateHostUpdate` will  be nil

pkg/route/validation/validation_test.go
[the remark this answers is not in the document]
chiragkyal: Okay, that makes sense to make. I've removed this test since we are  checking `ExternalCertificate` alone without `Certificate`..And all  other `if`s are already checked individually.

pkg/route/hostassignment/assignment.go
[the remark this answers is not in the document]
chiragkyal: Yeah! The external certificate secret name can be changed, but the caveat here is even if the secret name remains the same, we need to assume that the secret content is changed, and hence, the certificate is changed.

pkg/route/hostassignment/assignment_test.go
chiragkyal: As per the discussion over slack, if the featuregate was disabled, and `ExternalCertificate` wasn't previously set, apiserver will strip this field and will set to `nil`
https://github.com/openshift/openshift-apiserver/blob/1fac5e7e3a6153efae875185af2dba48fbad41ab/pkg/route/apiserver/registry/route/strategy.go#L73-L93

So, for the new route, `ExternalCertificate`  will be `nil`. Updated the test.

pkg/route/hostassignment/assignment_test.go
```diff
@@ -215,6 +219,112 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			errs:           0,
+
+			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
+		},
+		{
+			name:     "certificate-changed-from-external-certificate-allowed-but-featuregate-is-not-set",
+			host:     "host",
+			expected: "host",
+			oldHost:  "host",
+			// if the featuregate was disabled, and ExternalCertificate wasn't previously set, apiserver will strip this field.
```
alebedev87, line 279:
```suggestion
			// if the featuregate was disabled, and ExternalCertificate wasn't previously set, apiserver will strip ExternalCertificate field.
```

chiragkyal: Done

chiragkyal:
I've just squashed the commits. would request your re-lgtm.

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,34 @@
...
+		// return nil since if the feature gate is off
+		// ValidateHostUpdate() is sufficient to validate
+		// permissions
+		return nil
+	}
+
+	if (newTLS != nil && newTLS.ExternalCertificate != nil) || (oldTLS != nil && oldTLS.ExternalCertificate != nil) {
+		return routecommon.CheckRouteCustomHostSAR(ctx, field.NewPath("spec", "TLS", "externalCertificate"), sarc)
```
chiragkyal, line 30: `TLS` should be `tls` . I'll update it along with future review comments (if any). 
```suggestion
		return routecommon.CheckRouteCustomHostSAR(ctx, field.NewPath("spec", "tls", "externalCertificate"), sarc)
```

pkg/route/hostassignment/assignment.go
```diff
@@ -190,6 +215,10 @@ func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc S
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Certificate, older.Spec.TLS.Certificate, field.NewPath("spec", "tls", "certificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.DestinationCACertificate, older.Spec.TLS.DestinationCACertificate, field.NewPath("spec", "tls", "destinationCACertificate"))...)
 			errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.Key, older.Spec.TLS.Key, field.NewPath("spec", "tls", "key"))...)
+
+			if opts.AllowExternalCertificates && route.Spec.TLS.ExternalCertificate != nil && older.Spec.TLS.ExternalCertificate != nil {
+				errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"))...)
+			}
```
Miciah, line 221: Is this test case correct?  
```go
		{
			name:           "no-certificate-changed-to-external-certificate-denied",
			host:           "host",
			expected:       "host",
			oldHost:        "host",
			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: &routev1.LocalObjectReference{Name: "b"}},
			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, ExternalCertificate: nil},
			wildcardPolicy: routev1.WildcardPolicyNone,
			allow:          false,
			errs:           1,

			opts: route.RouteValidationOptions{AllowExternalCertificates: true},
		},
```
The above test case fails if I add it to `TestHostWithWildcardPolicies`:
```text
go test ./pkg/route/hostassignment -run TestHostWithWildcardPolicies
# github.com/openshift/library-go/pkg/route/hostassignment.test
guile: warning: failed to install locale
--- FAIL: TestHostWithWildcardPolicies (0.00s)
    --- FAIL: TestHostWithWildcardPolicies/no-certificate-changed-to-external-certificate-denied (0.00s)
        assignment_test.go:506: expected 1 errors, got 0: []
```
It seems to me that the logic in `ValidateHostUpdate` should be as follows:
```suggestion
			if opts.AllowExternalCertificates {
				if route.Spec.TLS.ExternalCertificate == nil || older.Spec.TLS.ExternalCertificate == nil {
					errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate, older.Spec.TLS.ExternalCertificate, field.NewPath("spec", "tls", "externalCertificate"))...)
				} else {
					errs = append(errs, apimachineryvalidation.ValidateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"))...)
				}
			}
```
You'll need to modify the "external-certificate-changed-to-certificate-denied" and "certificate-changed-to-external-certificate-denied" test cases to expect an additional error.

Miciah:
It looks like this suggestion from the original PR got dropped: https://github.com/openshift/library-go/pull/1549#discussion_r1357389686

pkg/route/hostassignment/assignment.go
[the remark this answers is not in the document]
Miciah: I do see you added a comment for this caveat to `ValidateHostUpdate`, but perhaps it deserves a code comment in `certificateChangeRequiresAuth` as well.

pkg/route/validation/validation.go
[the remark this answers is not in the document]
Miciah: It seems that you squashed the PR into a single commit.  Did you mean to keep this change as a separate commit?  (It does make it easier to review the PR if this sort of cleanup is put in a separate commit from the main focus of the PR.)

pkg/route/hostassignment/assignment.go
```diff
@@ -86,41 +91,61 @@ func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessR
...
 			return false
 		}
-		return a.CACertificate != b.CACertificate ||
+
+		certChanged := a.CACertificate != b.CACertificate ||
 			a.Certificate != b.Certificate ||
 			a.DestinationCACertificate != b.DestinationCACertificate ||
 			a.Key != b.Key
+
+		if opts.AllowExternalCertificates && route.Spec.TLS != nil && older.Spec.TLS != nil {
```
Miciah, line 123: The nil checks are redundant as this is under `case route.Spec.TLS != nil && older.Spec.TLS != nil:`.
```suggestion
		if opts.AllowExternalCertificates {
```

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,34 @@
...
+	routev1 "github.com/openshift/api/route/v1"
+	routecommon "github.com/openshift/library-go/pkg/route"
+)
+
+// ValidateHostExternalCertificate checks if the user has permissions to create and update
+// custom-host subresource of routes. This check is required to be done prior to ValidateHostUpdate()
+// since updating hosts while using externalCertificate is contingent on the user having both these
+// permissions. The ValidateHostUpdate() cannot differentiate if the certificate has changed since
+// now the certificates will be present as a secret object, due to this it proceeds with the assumption
+// that the certificate has changed when the route has externalCertificate set.
```
Miciah, line 17: Ultimately, `ValidateUpdate` is going to call `ValidateHostExternalCertificate` and then call `ValidateHostUpdate`, right?  This comment is helpful, but it still isn't quite clear why the logic in `ValidateHostExternalCertificate` cannot simply be added to `ValidateHostUpdate`.

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -0,0 +1,34 @@
...
+func ValidateHostExternalCertificate(ctx context.Context, new, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, opts routecommon.RouteValidationOptions) field.ErrorList {
+	newTLS := new.Spec.TLS
+	oldTLS := older.Spec.TLS
+
+	if !opts.AllowExternalCertificates {
+		// return nil since if the feature gate is off
+		// ValidateHostUpdate() is sufficient to validate
+		// permissions
```
Miciah, line 25:
```suggestion
		// Return nil since if the feature gate is off.
		// ValidateHostUpdate() is sufficient to validate
		// permissions.
```

pkg/route/validation/validation.go
```diff
@@ -354,6 +308,62 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+		errs = append(errs, field.Forbidden(fldPath, "router serviceaccount does not have permission to list this secret"))
+	}
+
+	// The secret should be in the same namespace as that of the route.
+	secret, err := secretsGetter.Secrets(route.Namespace).Get(ctx, tls.ExternalCertificate.Name, metav1.GetOptions{})
+	if err != nil && apierrors.IsNotFound(err) {
+		return append(errs, field.NotFound(fldPath, err))
+	} else if err != nil {
+		return append(errs, field.InternalError(fldPath, err))
+	}
```
Miciah, line 357:
```suggestion
	if err != nil {
		if apierrors.IsNotFound(err) {
			return append(errs, field.NotFound(fldPath, err))
		}
		return append(errs, field.InternalError(fldPath, err))
	}
```

pkg/route/validation/validation.go
```diff
@@ -354,6 +308,62 @@ func validateTLS(route *routev1.Route, fldPath *field.Path) field.ErrorList {
...
+
+	// The secret should be of type kubernetes.io/tls
+	if secret.Type != corev1.SecretTypeTLS {
+		errs = append(errs, field.Invalid(fldPath, tls.ExternalCertificate.Name, "secret of type 'kubernetes/tls' required"))
```
Miciah, line 361: That should be "kubernetes.io/tls", or better, just build the error message using the named const:
```suggestion
		errs = append(errs, field.Invalid(fldPath, tls.ExternalCertificate.Name, fmt.Sprintf("secret of type %q required", corev1.SecretTypeTLS)))
```

pkg/route/common.go
```diff
@@ -0,0 +1,59 @@
...
+// SubjectAccessReviewCreator is an interface for performing subject access reviews
+type SubjectAccessReviewCreator interface {
+	Create(ctx context.Context, sar *authorizationv1.SubjectAccessReview, opts metav1.CreateOptions) (*authorizationv1.SubjectAccessReview, error)
+}
+
+// RouteValidationOptions used to tweak how/what fields are validated. These
+// options are propagated by the apiserver.
+type RouteValidationOptions struct {
+
+	// AllowExternalCertificates option is set when RouteExternalCertificate
```
Miciah, line 25:
```suggestion
	// AllowExternalCertificates option is set when the RouteExternalCertificate
	// feature gate is enabled.
```

pkg/route/hostassignment/externalcertificate.go
alebedev87: Reminded me of [my comment](https://github.com/openshift/library-go/pull/1625#discussion_r1472856895).

pkg/route/validation/validation.go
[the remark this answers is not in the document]
chiragkyal: My bad :(.. I forgot to split this into a separate commit earlier. I've now updated it.

pkg/route/hostassignment/externalcertificate.go
chiragkyal: Done

pkg/route/hostassignment/assignment.go
chiragkyal: Thanks for the suggestion and finding this edge case. I've updated the logic in `ValidateHostUpdate` as suggested, and also added the above test case along with "no-certificate-changed-to-external-certificate-allowed" test into `TestHostWithWildcardPolicies`

pkg/route/common.go
```diff
@@ -0,0 +1,59 @@
...
+	AllowExternalCertificates bool
+}
```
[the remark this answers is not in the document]
chiragkyal: That's correct. The `AllowExternalCertificates` flag (along with RouteValidationOptions struct) will be removed while graduating from TP -> GA.

pkg/route/hostassignment/assignment.go
chiragkyal: Thanks, addressed the suggestion.

pkg/route/hostassignment/externalcertificate.go
chiragkyal: Done.

pkg/route/validation/validation.go
chiragkyal: Addressed.

pkg/route/validation/validation.go
chiragkyal: Thanks, I've updated the error message.

pkg/route/common.go
chiragkyal: Addressed, thanks.

pkg/route/hostassignment/externalcertificate.go
chiragkyal: The EP did mention about invoking `ValidateHostExternalCertificate` prior to `ValidateHostUpdate`, but it's body lacks the actual reason why we decided to choose this approach.

As per our discussion over Slack and further investigation, I've added a `TODO` to consider merging `ValidateHostExternalCertificate` function into `ValidateHostUpdate` later.

pkg/route/hostassignment/assignment.go
```diff
@@ -86,41 +91,65 @@ func AllocateHost(ctx context.Context, route *routev1.Route, sarc SubjectAccessR
...
+		hasInfo = hasInfo || len(tls.ExternalCertificate.Name) > 0
+	}
+	return hasInfo
 }
 
-func certificateChangeRequiresAuth(route, older *routev1.Route) bool {
+// certificateChangeRequiresAuth determines whether changes to the TLS certificate configuration require authentication.
+// Note: If either route uses externalCertificate, this function always returns true, as we cannot definitively verify if
+// the content of the referenced secret has been modified. Even if the secret name remains the same,
+// we must assume that the secret content is changed, necessitating authentication.
```
Miciah, line 112:
```suggestion
// we must assume that the secret content is changed, necessitating authorization.
```

Miciah:
Excellent!  
/approve
/lgtm

I added two comments since your last update, but they are very minor things.  @alebedev87 might want to take another quick look after your most recent changes, so I'll add a hold just in case.  
/hold

pkg/route/hostassignment/assignment.go
```diff
@@ -12,26 +12,31 @@ import (
...
 	hostSet := len(route.Spec.Host) > 0
-	certSet := route.Spec.TLS != nil && (len(route.Spec.TLS.CACertificate) > 0 || len(route.Spec.TLS.Certificate) > 0 || len(route.Spec.TLS.DestinationCACertificate) > 0 || len(route.Spec.TLS.Key) > 0)
+	certSet := route.Spec.TLS != nil &&
+		(len(route.Spec.TLS.CACertificate) > 0 ||
+			len(route.Spec.TLS.Certificate) > 0 ||
+			len(route.Spec.TLS.DestinationCACertificate) > 0 ||
+			len(route.Spec.TLS.Key) > 0)
+
+	if opts.AllowExternalCertificates && route.Spec.TLS != nil && route.Spec.TLS.ExternalCertificate != nil {
+		certSet = certSet || len(route.Spec.TLS.ExternalCertificate.Name) > 0
```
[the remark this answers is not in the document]
chiragkyal: Covered in https://github.com/openshift/library-go/pull/1706

PR #1627, opened 2023-11-22
OCPNODE-1890: Bump k8s api to v0.29.0

sairameshv:
cc: @soltysh


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/library-go
 
-go 1.20
+go 1.21.3
```
soltysh, line 3: Same as with the others go 1.20

go.mod
```diff
@@ -1,6 +1,8 @@
...
+
+toolchain go1.21.4
```
soltysh, line 5: Drop it, we don't want to enable toolchain, yet.

soltysh (changes requested):

soltysh (approved):
/lgtm
/approve

PR #1630, opened 2023-11-24
make status reporting reuse name of CertRotationController

atiratree:
to fix the following issue: https://github.com/openshift/library-go/pull/1593#discussion_r1399767897

I have found out when bumping library-go https://github.com/openshift/cluster-kube-apiserver-operator/pull/1579 that it is not possible to set the private `conditionName` in StaticPodConditionStatusReporter. 

This solution would also help with the deduplication of the name.


dgrisonnet (approved):
/lgtm
/approve

PR #1632, opened 2023-11-28
controllercmd: allow configuring topology detection

stevekuznetsov:
In a HyperShift context, the toplogy that a component is expected to adhere to is not the management-cluster-wiide infrastructure, but the tenant-specific value. This commit allows consumers to detect topology in less rigid ways.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -95,9 +95,31 @@ type ControllerBuilder struct {
 	// We use this flag to determine at runtime if we can alter leader election for SNO configurations
 	userExplicitlySetLeaderElectionValues bool
 
+	// different deployment strategies will require sensing topologies in disjoint manners
+	topologyDetector TopologyDetector
+
 	// Allow enabling HTTP2
 	enableHTTP2 bool
 }
+
+type TopologyDetector interface {
+	DetectTopology(ctx context.Context, restClient *rest.Config) (configv1.TopologyMode, error)
+}
+
+type infrastructureStatusTopologyDetector struct{}
+
+func (i infrastructureStatusTopologyDetector) DetectTopology(ctx context.Context, restClient *rest.Config) (configv1.TopologyMode, error) {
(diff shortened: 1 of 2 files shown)


pkg/controller/controllercmd/builder.go
```diff
@@ -95,6 +95,9 @@ type ControllerBuilder struct {
 	// We use this flag to determine at runtime if we can alter leader election for SNO configurations
 	userExplicitlySetLeaderElectionValues bool
 
+	// different deployment strategies will require sensing topologies in disjoint manners
```
deads2k, line 98: make the interface broader to allow people to avoid using restClients which resist unit testing.

pkg/controller/controllercmd/builder.go
```diff
@@ -160,6 +173,11 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	return b
 }
 
+func (b *ControllerBuilder) WithLeaderElectionTopologyDetector(topologyDetector func(ctx context.Context, restClient *rest.Config) (configv1.TopologyMode, error)) *ControllerBuilder {
```
deads2k, line 176: Create an interface for the method we want.

```go
type TopologyDetector struct{
  GetTopology(ctx) (TopologyMode, error)
}
```

And default impls for your favorites, named for the proper deployment strategies taking the aforementioned clients (not rest.Config).

pkg/controller/controllercmd/builder.go
```diff
@@ -160,6 +183,11 @@ func (b *ControllerBuilder) WithLeaderElection(leaderElection configv1.LeaderEle
 	return b
 }
 
+func (b *ControllerBuilder) WithLeaderElectionTopologyDetector(topologyDetector TopologyDetector) *ControllerBuilder {
```
deads2k, line 186:
```suggestion
func (b *ControllerBuilder) WithTopologyDetector(topologyDetector TopologyDetector) *ControllerBuilder {
```

This would be logically for everyone, correct?

deads2k:
name fix needed.

/approve

PR #1634, opened 2023-11-29
AUTH-442: add ClusterFleetEvaluation type to status controller

ibihim:
# What

Add ClusterFleetEvaluationType to status controller.

# Why

It is a valid ClusterOperator status, but doesn't have a mechanism to be pulled from Operator's status into ClusterOperator's status.


pkg/operator/status/status_controller.go
```diff
@@ -199,10 +200,11 @@ func (c StatusSyncer) Sync(ctx context.Context, syncCtx factory.SyncContext) err
 		clusterOperatorObj.Status.RelatedObjects = c.relatedObjects
 	}
 
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition("Degraded", operatorv1.ConditionFalse, c.degradedInertia, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition("Progressing", operatorv1.ConditionFalse, nil, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition("Available", operatorv1.ConditionTrue, nil, currentDetailedStatus.Conditions...))
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition("Upgradeable", operatorv1.ConditionTrue, nil, currentDetailedStatus.Conditions...))
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(string(configv1.OperatorDegraded), operatorv1.ConditionFalse, c.degradedInertia, currentDetailedStatus.Conditions...))
```
dgrisonnet, line 203: Instead of casting everytime, we could change the signature of the UnionClusterCondition function and cast there. We should only change the signature of the cluster one because the cluster operator condition is typed (https://github.com/openshift/api/blob/a782d1c1541ca8ae0fd17fff5f1dc8a8990dc195/config/v1/types_cluster_operator.go#L122) which is the not the case of the operator condition: https://github.com/openshift/api/blob/master/operator/v1/types.go#L165

(no clue why that is the case :man_shrugging:)

dgrisonnet (commented):
One little nit, but looks good to me overall

stlaz:
/hold
What's the first commit from Steve?
Join the bump and vendor commits to a single one

dgrisonnet (approved):
/lgtm

soltysh (approved):
/approve

PR #1639, opened 2023-12-07
HOSTEDCP-1344: pkg/crypto: expose an unsafe self-signed cert generator

stevekuznetsov:
When testing, it's useful to change what the current time is.


deads2k:
use your power wisely.

/lgtm

PR #1641, opened 2023-12-09
OCPBUGS-21846: revision_controller: update last revision only when a revision is completely rendered

ingvagabund:
/hold

until tested through operators:
KS: https://github.com/openshift/cluster-kube-scheduler-operator/pull/514 worked
KCM: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/773#issuecomment-1848898230 worked

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -89,12 +89,14 @@ func (c RevisionController) createRevisionIfNeeded(ctx context.Context, recorder
 
 	// check to make sure that the latestRevision has the exact content we expect.  No mutation here, so we start creating the next Revision only when it is required
 	if isLatestRevisionCurrent {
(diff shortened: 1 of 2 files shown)


pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -89,12 +89,14 @@ func (c RevisionController) createRevisionIfNeeded(ctx context.Context, recorder
 
 	// check to make sure that the latestRevision has the exact content we expect.  No mutation here, so we start creating the next Revision only when it is required
 	if isLatestRevisionCurrent {
+		recorder.Eventf("LatestRevision", "returning early, %d triggered and up to date", latestAvailableRevision)
```
deads2k, line 92: I added this for debugging, but it's likely to be very noisy.  Perhaps just a log message at v(4) so it's not available by default?

deads2k:
some comments that needs resolving before merge.  lgtm otherwise.

/approve

PR #1644, opened 2023-12-12
AUTH-471: pkg/operator/status - add EvaluationConditionDetected

ibihim:
Cherry-pick of https://github.com/openshift/library-go/pull/1634

# What

Add ClusterFleetEvaluationType to status controller.

# Why

It is a valid ClusterOperator status, but doesn't have a mechanism to be pulled from Operator's status into ClusterOperator's status.


soltysh (approved):
/approve

PR #1652, opened 2024-01-04
OCPBUGS-18697: controller/factory: skip event handler registration for base controller when ResyncEvery and WithoutEventHandlers is set

ingvagabund:
When ResyncEvery is used update events can still invoke Sync more often than resyncInterval specified for ResyncEvery. Making ResyncEvery obsolete.

Expected usage:
```
return factory.New().WithInformers(
		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
		kubeInformersClusterScoped.Core().V1().Nodes().Informer(),
).ResyncEvery(10*time.Second).WithoutEventHandlers().WithSync(c.sync).WithSyncDegradedOnError(operatorClient).ToController("ControllerName", eventRecorder), nil
```

Follow up for https://github.com/openshift/library-go/pull/1620/files#r1399213346.

Testing in https://github.com/openshift/cluster-kube-scheduler-operator/pull/525.

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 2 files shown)


ingvagabund:
/hold
The solution here is not sufficient

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -106,7 +107,7 @@ func NewGuardController(
 	return factory.New().WithInformers(
 		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
 		kubeInformersClusterScoped.Core().V1().Nodes().Informer(),
-	).WithSync(c.sync).WithSyncDegradedOnError(operatorClient).ToController("GuardController", eventRecorder), nil
+	).ResyncEvery(20*time.Second).WithoutEventHandlers().WithSync(c.sync).WithSyncDegradedOnError(operatorClient).ToController("GuardController", eventRecorder), nil
```
bertinatto, line 110: Have you considered using `ResyncEvery()` without `WithInformers()` instead?

If you need `WithInformers()` just for the cache-waiting behavior, would a new `WithCachesSynced(informers)` be a more intuitive API rather than `WithoutEventHandlers()`?

ingvagabund:
Closing for not as I don't have capacity to pursue the solution. It's currently a low priority item.

PR #1653, opened 2024-01-04
expose more funcs for etcd cert_rotation 

tjungblu:
for CEO we would like to dynamically run this controller multiple times within the lifetime of another. We would need to expose some struct fields for that.


soltysh (approved):
/lgtm
/approve

PR #1656, opened 2024-01-05
add types to allow analysis of ondisk certificates

deads2k:


pkg/certs/cert-inspection/certgraphapi/types.go
```diff
@@ -28,6 +29,18 @@ type PerInClusterResourceData struct {
 	CertKeyPairs []PKIRegistryInClusterCertKeyPair `json:"certKeyPairs"`
 }
 
+// PerOnDiskResourceData tracks metadata that corresponds to specific files on disk.
+// This data should not duplicate the analysis of the certkeypair lists, but is pulled from files on disk.
+// It will be stitched together by a generator after the fact.
+type PerOnDiskResourceData struct {
```
vrutkovs, line 35: Perhaps it shouldn't make a distinction between CA/certs/keys? PEM files containing the cert and the key would have the same metadata, so
```
type PerOnDiskResourceData struct {
	TLSArtifact []OnDiskLocationWithMetadata `json:"tlsArtifact"`
}
```
would be sufficient

PR #1657, opened 2024-01-09
move diff before update to klog v2

deads2k:
if we're about to make a server call from an operator, it's worth the cost to log the reason for the change.  We won't add to the event (yet) due to the additional size, but it could be considered for a good enough reason.

Came up while checking https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.15-e2e-aws-sdn-cgroupsv2/1743684397107777536


stlaz (changes requested):
`klog` once again allows using `klog.V(x).Infof` again, please use that instead

stlaz:
/lgtm
the above were just perf optimizations

PR #1658, opened 2024-01-09
when updating status, do a live GET after conflict

deads2k:
The retry logic already existed, but relied on updating the cache.  In runs like https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.15-e2e-aws-sdn-cgroupsv2/1743684397107777536 (and others I've seen) this appears to be highly latent when observing the 

> base_controller.go:268] ConsoleOperator reconciliation failed: Operation cannot be fulfilled on consoles.operator.openshift.io "cluster": the object has been modified; please apply your changes to the latest version and try again

messages.  It's not likely to be connectivity since the updates are going through.  The lister is still preferred, but after the first failure (made this easy to tune), the code does a live GET of the current state.

/assign @stlaz 

eliminates the noise.

### DIFF
--- a/pkg/operator/v1helpers/helpers.go
+++ b/pkg/operator/v1helpers/helpers.go
@@ -159,8 +159,21 @@ type UpdateStatusFunc func(status *operatorv1.OperatorStatus) error
 func UpdateStatus(ctx context.Context, client OperatorClient, updateFuncs ...UpdateStatusFunc) (*operatorv1.OperatorStatus, bool, error) {
 	updated := false
 	var updatedOperatorStatus *operatorv1.OperatorStatus
+	numberOfAttempts := 0
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		_, oldStatus, resourceVersion, err := client.GetOperatorState()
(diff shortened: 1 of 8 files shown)


pkg/operator/v1helpers/helpers.go
```diff
@@ -159,8 +159,21 @@ type UpdateStatusFunc func(status *operatorv1.OperatorStatus) error
 func UpdateStatus(ctx context.Context, client OperatorClient, updateFuncs ...UpdateStatusFunc) (*operatorv1.OperatorStatus, bool, error) {
 	updated := false
 	var updatedOperatorStatus *operatorv1.OperatorStatus
+	numberOfFailedAttempts := 0
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		_, oldStatus, resourceVersion, err := client.GetOperatorState()
+		defer func() {
+			numberOfFailedAttempts++
```
stlaz, line 165: nit: to be accurate, this should be called `numberOfAttempts` as it's going to be increased even on a success

PR #1661, opened 2024-01-12
collect certs on disk: set owner, permissions and selinux label, skip revisioned folders

vrutkovs:
Tested in https://github.com/openshift/origin/pull/28523

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collector.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collector.go
@@ -21,18 +21,29 @@ func GatherCertsFromPlatformNamespaces(ctx context.Context, kubeClient kubernete
 	return gatherFilteredCerts(ctx, kubeClient, platformConfigMaps, platformSecrets, options)
 }
 
-func GatherCertsFromDisk(ctx context.Context, kubeClient kubernetes.Interface, prefix, dir string, options ...certGenerationOptions) (*certgraphapi.PKIList, error) {
+func GatherCertsFromDisk(ctx context.Context, kubeClient kubernetes.Interface, dir string, options ...certGenerationOptions) (*certgraphapi.PKIList, error) {
 	errs := []error{}
 
-	certs, err := gatherSecretsFromDisk(ctx, prefix, dir, options)
+	certs, certMetadata, err := gatherSecretsFromDisk(ctx, dir, options)
 	if err != nil {
 		errs = append(errs, err)
 	}
-	caBundles, err := gatherCABundlesFromDisk(ctx, prefix, dir, options)
+	caBundles, caBundlesMetadata, err := gatherCABundlesFromDisk(ctx, dir, options)
 	if err != nil {
 		errs = append(errs, err)
 	}
 	pkiList := PKIListFromParts(ctx, nil, certs, caBundles)
+
+	fileMetadata := []certgraphapi.OnDiskLocationWithMetadata{}
+	for _, metadata := range certMetadata {
+		fileMetadata = append(fileMetadata, *metadata)
+	}
+	for _, metadata := range caBundlesMetadata {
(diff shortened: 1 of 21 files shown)


deads2k:
/approve

If you want to update for types like https://github.com/openshift/library-go/pull/1656 (or something similar) now is probably the right time.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -216,23 +244,73 @@ func MergePKILists(ctx context.Context, first, second *certgraphapi.PKIList) *ce
...
-		CertificateAuthorityBundles: *caBundlesList,
-		CertKeyPairs:                *certList,
+		CertificateAuthorityBundles: deduplicateCABundlesList(caBundleList),
+		CertKeyPairs:                deduplicateCertKeyPairList(certList),
+		OnDiskResourceData:          deduplicateOnDiskMetadata(onDiskResourceData),
 		InClusterResourceData:       inClusterData,
 	}
 }
+
+func stripPrefixFromPaths(certList *certgraphapi.CertKeyPairList, caBundlesList *certgraphapi.CertificateAuthorityBundleList, onDiskResourceData *certgraphapi.PerOnDiskResourceData, param pathUpdateParams, pathFuncs ...pathUpdateFn) {
```
deads2k, line 280: so this isn't prefixes, it's `mutatePathFns`?

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -216,23 +244,73 @@ func MergePKILists(ctx context.Context, first, second *certgraphapi.PKIList) *ce
...
+
+	onDiskResourceData := &certgraphapi.PerOnDiskResourceData{
+		TLSArtifact: append(first.OnDiskResourceData.TLSArtifact, second.OnDiskResourceData.TLSArtifact...),
+	}
+
+	param := pathUpdateParams{
+		prefix:   prefix,
+		nodeList: nodeList,
+	}
+	stripPrefixFromPaths(certList, caBundleList, onDiskResourceData, param,
```
deads2k, line 266: Why did this move from when we collect data to when we merge it?  Doing it at collection adjusts how every layer in the stack sees the data. Versus doing it here where it means other locations in the code see the old names.

pkg/certs/cert-inspection/certgraphanalysis/deduplication.go
```diff
@@ -183,3 +183,20 @@ func CombineCABundleOnDiskLocations(in *certgraphapi.CertificateAuthorityBundle,
...
 	return out
 }
+
+func deduplicateOnDiskMetadata(in *certgraphapi.PerOnDiskResourceData) certgraphapi.PerOnDiskResourceData {
+	ret := certgraphapi.PerOnDiskResourceData{}
+	for _, curr := range in.TLSArtifact {
+		found := false
+		for _, existing := range ret.TLSArtifact {
+			if existing.Path == curr.Path {
+				found = true
```
deads2k, line 193: break after?

deads2k:
moving the strip to the merge bothers me: https://github.com/openshift/library-go/pull/1661#discussion_r1459602967

I'm not clear on why it should happen there instead of with our other mutating collection options.

Squash up to something more reasonable please.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
vrutkovs: That would require all collection layers to be aware of rootfsmount (doable) and require node list permissions. Its implementable, but that makes us dedup and filter twice - once on the node and once for incluster resources

deads2k: Let's push  it into the collection path.

PR #1665, opened 2024-01-17
resourceapply: update events to be more legible

stevekuznetsov:
- `s/cause by/caused by/` for better English
 - remove newlines in event logs, everyone exepcts single-line logs


deads2k:
> * emove newlines in event logs, everyone exepcts single-line logs

not true and not done in this PR

PR #1668, opened 2024-01-23
prefer lister when its resourceVersion is different than the last try

deads2k:
/hold

needs proof prior to merge

/assign @stlaz


deads2k:
/hold cancel

the proof worked.

PR #1670, opened 2024-01-24
NO-JIRA: Export OAuth metadata extraction function to be used in oc

ardaguclu:
As I promised to this change in https://github.com/openshift/oc/pull/1640#discussion_r1449928616, this PR
exports OAuth metadata extraction function to be used by `oc`.


ardaguclu:
@stlaz just friendly reminder to not slip through the cracks and still not urgent :).

ardaguclu:
It seems that one way or another, we will abandon url extraction and I don't think we need this refactoring.
/close

PR #1674, opened 2024-02-05
add cert rotation based on subject/authority keys

tjungblu:
This PR adds another rotation criterion for signer changes.
For etcd we need to rollout the bundles first, which requires
a more deeper check when the leaf certs need to be recreated.

This requires parsing of the certificate in question and compare
SubjectKey and AuthorityKey.

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -1,6 +1,7 @@
 package certrotation
 
 import (
+	"bytes"
 	"context"
 	"crypto/x509"
 	"fmt"
@@ -74,7 +75,7 @@ type TargetCertCreator interface {
 	// NewCertificate creates a new key-cert pair with the given signer.
 	NewCertificate(signer *crypto.CA, validity time.Duration) (*crypto.TLSCertificateConfig, error)
 	// NeedNewTargetCertKeyPair decides whether a new cert-key pair is needed. It returns a non-empty reason if it is the case.
-	NeedNewTargetCertKeyPair(currentSecretAnnotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string
+	NeedNewTargetCertKeyPair(currentCert *x509.Certificate, currentSecretAnnotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string
 	// SetAnnotations gives an option to override or set additional annotations
 	SetAnnotations(cert *crypto.TLSCertificateConfig, annotations map[string]string) map[string]string
 }
@@ -117,7 +118,18 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 		}
 	}
 
-	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret.Annotations, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired); len(reason) > 0 {
+	var currentCert *x509.Certificate
+	if crtBytes, ok := targetCertKeyPairSecret.Data["tls.crt"]; ok {
+		pemCerts, err := crypto.CertsFromPEM(crtBytes)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse pem x509 tls.crt in secret %s: %w", targetCertKeyPairSecret.Name, err)
+		}
+		if len(pemCerts) > 0 {
+			currentCert = pemCerts[len(pemCerts)-1]
+		}
+	}
+
+	if reason := c.CertCreator.NeedNewTargetCertKeyPair(currentCert, targetCertKeyPairSecret.Annotations, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired); len(reason) > 0 {
 		c.EventRecorder.Eventf("TargetUpdateRequired", "%q in %q requires a new target cert/key pair: %v", c.Name, c.Namespace, reason)
 		if err := setTargetCertKeyPairSecret(targetCertKeyPairSecret, c.Validity, signingCertKeyPair, c.CertCreator, c.AdditionalAnnotations); err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -117,7 +118,18 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 		}
 	}
 
-	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret.Annotations, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired); len(reason) > 0 {
+	var currentCert *x509.Certificate
+	if crtBytes, ok := targetCertKeyPairSecret.Data["tls.crt"]; ok {
+		pemCerts, err := crypto.CertsFromPEM(crtBytes)
+		if err != nil {
+			return nil, fmt.Errorf("could not parse pem x509 tls.crt in secret %s: %w", targetCertKeyPairSecret.Name, err)
```
stlaz, line 125: should we regen the cert in this case, then?

pkg/operator/certrotation/target.go
```diff
@@ -135,23 +147,56 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 	return targetCertKeyPairSecret, nil
 }
 
-func needNewTargetCertKeyPair(annotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string {
+func needNewTargetCertKeyPair(currentSecretCert *x509.Certificate, annotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string {
```
stlaz, line 150: nit: I think `currentCert` should do, I'm getting a bit confused by the `secret` part :slightly_smiling_face:

pkg/operator/certrotation/target.go
```diff
@@ -135,23 +147,56 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
...
+	}
+
+	if !bundleContainsSigner {
+		return fmt.Sprintf("issuer %q, not in ca bundle:\n%s", signerCommonName, certs.CertificateBundleToString(caBundleCerts))
+	}
+
+	// in some cases, e.g. with etcd, we need to bundle the signer CA before we can rotate a certificate
+	// hence we also check whether the signer itself has changed, denoted by its AKI/SKI
+	if len(currentSecretCert.AuthorityKeyId) > 0 && len(signer.Config.Certs) > 0 && len(signer.Config.Certs[0].SubjectKeyId) > 0 {
+		if !reflect.DeepEqual(currentSecretCert.AuthorityKeyId, signer.Config.Certs[0].SubjectKeyId) {
```
stlaz, line 180: `bytes.Equal`

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -23,6 +23,8 @@ const (
 	CertificateIssuer = "auth.openshift.io/certificate-issuer"
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
+	// CertificateIssuerSignature contains the signature of the certificate that signed another certificate.
+	CertificateIssuerSignature = "auth.openshift.io/certificate-signature"
```
stlaz, line 27: unused

pkg/operator/certrotation/target_test.go
```diff
@@ -433,7 +437,83 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+				ski := make([]byte, 32)
+				_, err := cryptorand.Read(ski)
+				if err != nil {
+					return nil, err
+				}
+
+				return newTestCACertificateWithAuthority(pkix.Name{CommonName: "signer-tests"}, int64(1), metav1.Duration{Duration: time.Hour * 24 * 60}, time.Now, ski)
+			},
+		},
+	}
```
stlaz, line 469: I'd expect 4 tests, splitting each of these in two:
1. signer matches the leaf in the secret
2. signer does not match

You may want to add Reactors to the mocked clients, checking for Updates

pkg/operator/certrotation/client_cert_rotation_controller.go
tjungblu: removed

pkg/operator/certrotation/target.go
tjungblu: fixed

pkg/operator/certrotation/target.go
tjungblu: good question, I'm wondering if that would cause issues in etcd. Because that would also (potentially) mean that you could pre-maturely sign a leaf with a new signer certificate that's not yet rolled out in a bundle.

I'll leave this for David / Haseeb, if they have some more opinion on this.

pkg/operator/certrotation/target.go
tjungblu: also an interesting implementation, TIL:

```
func Equal(a, b []byte) bool {
	// Neither cmd/compile nor gccgo allocates for these string conversions.
	return string(a) == string(b)
}
```

fixed.

pkg/operator/certrotation/target.go
stlaz: If you don't, you might forever crash on this and won't be able to roll out a new certificate. You'd also have to make sure that etcd would never pick the contents of the secret for all its pods because that would brick your cluster as none of the pods would be able to start.

hasbro17: `pemCerts, err := crypto.CertsFromPEM(crtBytes)`

What does an error in this case represent? A malformed or corrupted `tls.crt` file right.

If so, then we really have no option but to regenerate this now otherwise this will get picked up by the installer pod in the next revision rollout. I think that's what @stlaz is referring to. (Not quite sure if the first etcd pod to rollout with the malformed peer cert will then start crashing or just report unhealthy). Plus we can't block on this and expect the secret to get patched manually with a valid tls.crt so another reason to regenerate.

> Because that would also (potentially) mean that you could pre-maturely sign a leaf with a new signer certificate that's not yet rolled out in a bundle

@tjungblu On our end, aren't we always passing the old singer from `openshift-config` into `EnsureTargetCertKeyPair()`? i.e
https://github.com/openshift/cluster-etcd-operator/blob/5b9240f586d7fc789e5a78e2608f39b729d047e0/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go#L187

So even if we regenerate here we should be fine since we're using the old signer no?

And per https://github.com/openshift/cluster-etcd-operator/pull/1200 the signer in `openshift-config/etcd-singer` is only updated after we've already had a rollout with the new signer.
So the only way we prematurely regenerate here with a new signer (that isn't already bundled with each etcd member) is if we create a new signer `openshift-etcd/etcd-singer` and then immediately overwrite `openshift-config/etcd-singer` to trigger leaf-cert regen without waiting for a full rollout first.
But that would be a problem even without this case.

So I would reckon we can safely regenerate the leaf cert here. Unless I'm missing a scenario on how we end up prematurely signing the leaf cert.

hasbro17:
@stlaz Do the AKI/SKI checks look good to you? Or is the hold for something else?

pkg/operator/certrotation/target.go
tjungblu: the question is: how could this certificate not become parseable anymore? it must be a user error in the first place, so is it better to alert the admin via a degraded CEO directly that the cert they created is wrong or to silent regenerate it?

hasbro17: > how could this certificate not become parseable anymore? it must be a user error in the first place

That's a good point. I wasn't exactly sure on how we get to this error case. If we can agree what this state represents, (which may likely be a user overwrite)  then I can see the case for not silently recreating certs, and letting the degraded state be more obvious.

@stlaz Any thoughts on which way we want to handle this?

stlaz: Two examples pop out:
1. user malformed the secret
2. etcd memory corruption

tjungblu: in neither scenarios, just recreating a certificate is the appropriate action IMHO.

tjungblu:
tbc in #1691

PR #1675, opened 2024-02-06
CFE-866: feature: monitoring and handling "single" secret dynamically

chiragkyal:
Functionality for monitoring and handling "single" secret dynamically. 

### Changes in `single_item_monitor.go`:

- `ObjectKey`  representing the unique identifier for a resource.
- `singleItemMonitor`  responsible for monitoring a single resource using a SharedInformer.
- Defined methods for starting and stopping informers, adding and removing event handlers, and retrieving items from the informer's cache.

### Changes in `secret_monitor.go`:

- Implemented functionalities to create and manage informers for specific secrets.
- `SecretEventHandlerRegistration`  for registering and unregistering event handlers for secret monitoring.
- `SecretMonitor`  for monitoring and handling specific secrets using `singleItemMonitor`.
- Added methods for adding and removing secret event handlers, retrieving secret objects from the informer's cache.
- operations for counting the number of event handlers associated with a secret.

Related: [CFE-866](https://issues.redhat.com//browse/CFE-866)
Implements: https://github.com/openshift/enhancements/pull/1307

### DIFF
--- a/pkg/secret/OWNERS
+++ b/pkg/secret/OWNERS
@@ -1,0 +1,7 @@
+reviewers:
+  - deads2k
+  - soltysh
+  - chiragkyal
+approvers:
+  - deads2k
+  - soltysh
--- a/pkg/secret/OWNERS
+++ b/pkg/secret/OWNERS
@@ -1,7 +1,9 @@
 reviewers:
   - deads2k
   - soltysh
+  - Miciah
   - chiragkyal
 approvers:
   - deads2k
   - soltysh
+  - Miciah
--- a/pkg/secret/monitor.go
+++ b/pkg/secret/monitor.go
@@ -49,7 +49,7 @@ func (i *singleItemMonitor) HasSynced() bool {
 	return i.informer.HasSynced()
 }
 
-// StartInformer starts and runs the informer util the provided context is canceled,
+// StartInformer starts and runs the informer until the provided context is canceled,
 // or StopInformer() is called. It will block, so call via goroutine.
 func (i *singleItemMonitor) StartInformer(ctx context.Context) {
 	i.lock.Lock()
@@ -102,7 +102,7 @@ func (i *singleItemMonitor) AddEventHandler(handler cache.ResourceEventHandler) 
 	defer i.lock.Unlock()
 
 	if i.stopped {
-		return nil, fmt.Errorf("can not add handler %v to already stopped informer", handler)
+		return nil, fmt.Errorf("cannot add handler %v to already stopped informer", handler)
 	}
 
 	registration, err := i.informer.AddEventHandler(handler)
@@ -120,6 +120,10 @@ func (i *singleItemMonitor) AddEventHandler(handler cache.ResourceEventHandler) 
 func (i *singleItemMonitor) RemoveEventHandler(handle SecretEventHandlerRegistration) error {
 	i.lock.Lock()
 	defer i.lock.Unlock()
+
+	if handle == nil {
+		return fmt.Errorf("nil handler registration is provided")
+	}
 
 	if i.stopped {
 		return fmt.Errorf("can not remove handler %v from stopped informer", handle.GetHandler())
--- a/pkg/secret/monitor.go
+++ b/pkg/secret/monitor.go
@@ -50,13 +50,13 @@ func (i *singleItemMonitor) HasSynced() bool {
 }
 
 // StartInformer starts and runs the informer until the provided context is canceled,
-// or StopInformer() is called. It will block, so call via goroutine.
+// or StopInformer() is called.
 func (i *singleItemMonitor) StartInformer(ctx context.Context) {
 	i.lock.Lock()
+	defer i.lock.Unlock()
 
 	if !i.stopped {
 		klog.Warning("informer is already running")
-		i.lock.Unlock()
 		return
 	}
 
@@ -75,9 +75,8 @@ func (i *singleItemMonitor) StartInformer(ctx context.Context) {
 
 	klog.Info("starting informer")
 	i.stopped = false
-	i.lock.Unlock()
 
-	i.informer.Run(i.stopCh)
+	go i.informer.Run(i.stopCh)
 }
 
 // StopInformer stops the informer.
--- a/pkg/secret/monitor.go
+++ b/pkg/secret/monitor.go
@@ -1,0 +1,136 @@
+package secret
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+// ObjectKey represents the unique identifier for a resource, used to access it in the cache.
+type ObjectKey struct {
+	// Namespace is the namespace in which the resource is located.
+	Namespace string
+	// Name denotes metadata.name of a resource being monitored by informer
+	Name string
+}
+
+// singleItemMonitor monitors a single resource using a SharedInformer.
+type singleItemMonitor struct {
+	key      ObjectKey
+	informer cache.SharedInformer
+	lock     sync.Mutex
+	stopped  bool
+	stopCh   chan struct{}
+}
+
+// NewObjectKey creates a new ObjectKey for the given namespace and name.
+func NewObjectKey(namespace, name string) ObjectKey {
+	return ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}
+}
+
+// newSingleItemMonitor creates a new singleItemMonitor for the given key and informer.
+func newSingleItemMonitor(key ObjectKey, informer cache.SharedInformer) *singleItemMonitor {
+	return &singleItemMonitor{
+		key:      key,
+		informer: informer,
+		stopped:  true,
+		stopCh:   make(chan struct{}),
+	}
+}
+
+// HasSynced returns true if the informer's cache has been successfully synced.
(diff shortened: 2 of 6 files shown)


pkg/secret/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+	if !i.stopped {
+		klog.Warning("informer is already running")
+		i.lock.Unlock()
+		return
+	}
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			klog.Info("stopping informer due to context cancellation")
```
deads2k, line 66: I suspect you'll end up wanting these at higher verbosity.  I won't dictate for phase one since it's primarily your operator consuming.

pkg/secret/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+			if !i.StopInformer() {
+				klog.Error("failed to stop informer")
```
deads2k, line 68: Future-you can tell me if this was a good idea.  I'm not sure either way.

pkg/secret/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+		// this case is required to exit from the goroutine
+		// after normal StopInformer() call
+		case <-i.stopCh:
+			klog.Info("successfully stopped")
+		}
+	}()
+
+	klog.Info("starting informer")
+	i.stopped = false
+	i.lock.Unlock()
```
deads2k, line 79: move to `defer unlock` just under the lock.  Both your unlocks are at the tail.

pkg/secret/secret_monitor.go
```diff
@@ -0,0 +1,217 @@
...
+	RemoveSecretEventHandler(SecretEventHandlerRegistration) error
+
+	// GetSecret retrieves the secret object from the informer's cache using the provided SecretEventHandlerRegistration.
+	// This allows accessing the latest state of the secret without making an API call.
+	GetSecret(SecretEventHandlerRegistration) (*v1.Secret, error)
+}
+
+// secretEventHandlerRegistration is an implementation of the SecretEventHandlerRegistration.
+type secretEventHandlerRegistration struct {
+	cache.ResourceEventHandlerRegistration
```
deads2k, line 44: don't anonymously embed.  You need to carefully control each method exposed.

pkg/secret/monitor.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+// HasSynced returns true if the informer's cache has been successfully synced.
+func (i *singleItemMonitor) HasSynced() bool {
+	return i.informer.HasSynced()
+}
+
+// StartInformer starts and runs the informer util the provided context is canceled,
+// or StopInformer() is called. It will block, so call via goroutine.
```
deads2k, line 53: `Start*` usually doesn't block in kube.  `Run` usually does.  This looks easier to use if it doesn't block.  Is there a use-case for blocking?

pkg/secret/secret_monitor.go
```diff
@@ -0,0 +1,217 @@
...
+		klog.Info("secret informer started", " item key ", key)
+	}
+
+	// add the event handler
+	registration, err := m.itemMonitor.AddEventHandler(handler)
+	if err != nil {
+		return nil, err
+	}
+	m.numHandlers.Add(1)
+	klog.Info("secret handler added", " item key ", key)
```
deads2k, line 136: again, nearly certain to be too noisy, but I'll allow you start with it if you want.  I hope to a see PR quieting it down later this release.

pkg/secret/secret_monitor.go
```diff
@@ -0,0 +1,217 @@
...
+		// remove the key from map
+		delete(s.monitors, key)
+		klog.Info("secret informer stopped", " item key ", key)
+	}
+
+	return nil
+}
+
+// GetSecret retrieves the secret object from the informer's cache. Error if the secret is not found in the cache.
+func (s *secretMonitor) GetSecret(handlerRegistration SecretEventHandlerRegistration) (*v1.Secret, error) {
```
deads2k, line 181: pass context

pkg/secret/secret_monitor.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	m.numHandlers.Add(1)
+	klog.Info("secret handler added", " item key ", key)
+
+	return registration, nil
+}
+
+// RemoveSecretEventHandler removes a secret event handler and stops the informer if no handlers are left.
+// If the handler is not found or if there is an issue removing it, an error is returned.
+func (s *secretMonitor) RemoveSecretEventHandler(handlerRegistration SecretEventHandlerRegistration) error {
```
deads2k, line 143: take context

deads2k:
fairly minor comments. Please address and have a teammate check your test coverage.

chiragkyal:
/assign @alebedev87
FYI: this PR was split from https://github.com/openshift/library-go/pull/1626 to have only the secret monitoring piece.

pkg/secret/monitor.go
chiragkyal: This function is ultimately calling `informer.Run` which is a blocking call, and for that reason it's also blocking. 
If `StartInformer()` name is misleading, should we name it `RunInformer()`?

pkg/secret/monitor.go
chiragkyal: Since `informer.Run` is a blocking call, we need to unlock before invoking (starting any informer) it. I think moving unlock to `defer` will lock this function forever.(after first call)

pkg/secret/secret_monitor.go
chiragkyal: Here also, do we need this for future usage? Unlike `AddSecretEventHandler()` where we need a context to stop the informer, I'm wondering what the context would do during removal.

pkg/secret/monitor.go
chiragkyal: Updated to V(3)

pkg/secret/secret_monitor.go
chiragkyal: I've added a `TODO` to drop this later on. For now, this would be helpful for debugging.

pkg/secret/secret_monitor.go
chiragkyal: Passing context now

alebedev87:
Reviewed as a cherry picked commit in https://github.com/openshift/library-go/pull/1626.

/lgtm
/approve

pkg/secret/OWNERS
```diff
@@ -0,0 +1,7 @@
+reviewers:
+  - deads2k
+  - soltysh
+  - chiragkyal
+approvers:
+  - deads2k
+  - soltysh
```
deads2k, line 7: @soltysh your call

pkg/secret/monitor.go
```diff
@@ -0,0 +1,140 @@
...
+		// after normal StopInformer() call i.e when stopCh is closed.
+		case <-i.stopCh:
+		}
+	}()
+
+	klog.Info("starting informer")
+	i.stopped = false
+	i.lock.Unlock()
+
+	i.informer.Run(i.stopCh)
```
deads2k, line 80:
```suggestion
	go i.informer.Run(i.stopCh)
```

to avoid having your start block.

pkg/secret/monitor.go
deads2k: > Since `informer.Run` is a blocking call, we need to unlock before invoking (starting any informer) it. I think moving unlock to `defer` will lock this function forever.(after first call)

that line looks bugged.

pkg/secret/monitor.go
deads2k: > This function is ultimately calling `informer.Run` which is a blocking call, and for that reason it's also blocking.
> If `StartInformer()` name is misleading, should we name it `RunInformer()`?

How about making it non-blocking?

pkg/secret/secret_monitor.go
deads2k: > Do we need this for future usage? Unlike `AddSecretEventHandler()` where we need a context to stop the informer, I'm wondering what the context would do during removal.

Cancellation and timeouts.  I won't stick on it.

pkg/secret/monitor.go
alebedev87: It's blocking by design, the comment states this:

> It will block, so call via goroutine.

pkg/secret/monitor.go
chiragkyal: After the latest changes, the `StartInformer` is non-blocking. So I've moved it to `defer unlock`

pkg/secret/monitor.go
chiragkyal: Updated to make it non-blocking

pkg/secret/monitor.go
chiragkyal: Updated to make it non-blocking

pkg/secret/OWNERS
soltysh: sgtm

pkg/secret/OWNERS
```diff
@@ -0,0 +1,9 @@
+reviewers:
+  - deads2k
+  - soltysh
+  - Miciah
+  - chiragkyal
+approvers:
+  - deads2k
+  - soltysh
+  - Miciah
```
chiragkyal, line 9: @Miciah, as discussed, added you also

PR #1676, opened 2024-02-08
resourceapply: prevent ApplyConfigMapImproved from triggering unnecessary updates

machine424:
on ConfigMaps where service-ca-operator injects Service CA.

Needed for the https://github.com/openshift/cluster-monitoring-operator/pull/2226 effort where we're trying to migrate towards `resourceapply`'s intelligent utils (I haven't finished estimating the overhead of that intelligence on CMO's CPU and memory) to reduce UPDATE requests and avoid conflicts with the other controllers that we share the resources with.


machine424:
maybe
/jira refresh
will trigger it

machine424:
maybe
/lgtm
I don't know if the labels will be reset if I close and re-open

machine424:
Well let's give it a try

machine424:
The first one was missed, I think we'll need another lgtm here @deads2k
Thanks.

PR #1677, opened 2024-02-15
OCPBUGS-22332: route: Improve errors for missing custom-host permissions

Miciah:
Return "you do not have permission" error messages instead of "field is immutable" error messages when the user tries to update a route's `spec.host`, `spec.subdomain`, or `spec.tls` field without the required "custom-host" permission.

Note that this change does not affect the HTTP status code for these errors, which will continue to be HTTP 422, "Unprocessable Entity", as both [the openshift-apiserver admission](https://github.com/openshift/openshift-apiserver/blob/c5ba848e78463f9c73e1d2daec9e565daa4a6f4a/vendor/k8s.io/apiserver/pkg/registry/rest/update.go#L154-L156) and [the kube-apiserver admission](https://github.com/openshift/kubernetes/blob/5a4819c6048e42515c3c5538b723ba6272440eee/openshift-kube-apiserver/admission/route/hostassignment/admission.go#L135-L137) wrap errors from the route validation's `ValidateHostUpdate` function using [`"k8s.io/apimachinery/pkg/api/errors".NewInvalid`](https://github.com/openshift/kubernetes/blob/5a4819c6048e42515c3c5538b723ba6272440eee/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go#L281-L310).

* `pkg/route/hostassignment/assignment.go` (`routeHostPermissionErrMsg`, `routeSubdomainPermissionErrMsg`, `routeTLSPermissionErrMsg`): New consts.
(`AllocateHost`): Use `routeHostPermissionErrMsg` and `routeTLSPermissionErrMsg` instead of string literals.
(`validateImmutableField`): New helper function, based on `ValidateImmutableField` from apimachinery but using a custom error message.
(`ValidateHostUpdate`): Use the new consts and helper function instead of `ValidateImmutableField` from apimachinery.

### DIFF
--- a/pkg/route/hostassignment/assignment.go
+++ b/pkg/route/hostassignment/assignment.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
-	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/endpoints/request"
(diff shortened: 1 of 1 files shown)


Miciah:
Rebased for changes in #1625.

gcs278:
Reviewed. Usually I like to test out PRs that I review, but since this also needs vendoring elsewhere, I'll just provide LGTM based on my assessment that the code updates are only impacting error messages, which feels low-risk to me.
/lgtm

/hold
@alebedev87 did you want a quick review? Let's give 1 more week, but if you don't have time, let's send this through as it's low risk.

alebedev87:
LGTM
/unhold

PR #1681, opened 2024-03-01
OCPBUGS-30119: certrotation: update all secret types to `kubernetes.io/tls` preserving existing content

vrutkovs:
Pre-4.7 clusters had `SecretTypeTLS` type set, so on 4.15 upgrade these clusters had the secret delete and recreated. This commit will ensure these are being created without cert and key being regenerated.

Fixes two issues wrt handling secrets:
* existing secret type is converted to `kubernetes.io/tls`. If it cannot be done (i.e. `tls.crt` or `tls.key` is missing) it would be regenerated with `SignerUpdateRequired` event
* if the type can be converted existing data is preserved, so born-before-4.7 clusters look identical to 4.15 clusters in regards to secret type

### DIFF
--- a/pkg/operator/certrotation/metadata.go
+++ b/pkg/operator/certrotation/metadata.go
@@ -1,0 +1,36 @@
+package certrotation
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ensureMetadataUpdate(secret *corev1.Secret, owner *metav1.OwnerReference, additionalAnnotations AdditionalAnnotations) bool {
+	needsMetadataUpdate := false
+	// no ownerReference set
+	if owner != nil {
+		needsMetadataUpdate = ensureOwnerReference(&secret.ObjectMeta, owner)
+	}
+	// ownership annotations not set
+	return additionalAnnotations.EnsureTLSMetadataUpdate(&secret.ObjectMeta) || needsMetadataUpdate
+}
+
+func ensureSecretTLSTypeSet(secret *corev1.Secret) bool {
+	// Existing secret not found - no need to update metadata (will be done by needNewSigningCertKeyPair / NeedNewTargetCertKeyPair)
+	if len(secret.ResourceVersion) == 0 {
+		return false
+	}
+
+	// convert outdated secret type (created by pre 4.7 installer)
+	if secret.Type != corev1.SecretTypeTLS {
+		secret.Type = corev1.SecretTypeTLS
+		// wipe secret contents if tls.crt and tls.key are missing
+		_, certExists := secret.Data[corev1.TLSCertKey]
+		_, keyExists := secret.Data[corev1.TLSPrivateKeyKey]
+		if !certExists || !keyExists {
+			secret.Data = map[string][]byte{}
+		}
+		return true
+	}
+	return false
+}
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -72,21 +72,9 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 		}
 	}
 
-	needsMetadataUpdate := false
-	// no ownerReference set
-	if c.Owner != nil {
-		needsMetadataUpdate = ensureOwnerReference(&signingCertKeyPairSecret.ObjectMeta, c.Owner)
-	}
-	// ownership annotations not set
-	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta) || needsMetadataUpdate
-	// convert outdated secret type (set pre 4.7)
-	if signingCertKeyPairSecret.Type != corev1.SecretTypeTLS {
-		signingCertKeyPairSecret.Type = corev1.SecretTypeTLS
-		needsMetadataUpdate = true
-	}
-	// apply changes (possibly via delete+recreate) if secret exists and requires metadata update
+	// apply necessary metadata (possibly via delete+recreate) if secret exists
 	// this is done before content update to prevent unexpected rollouts
-	if needsMetadataUpdate && len(signingCertKeyPairSecret.ResourceVersion) > 0 {
+	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(signingCertKeyPairSecret) {
 		actualSigningCertKeyPairSecret, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
 		if err != nil {
 			return nil, err
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -62,24 +62,36 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 	signingCertKeyPairSecret := originalSigningCertKeyPairSecret.DeepCopy()
 	if apierrors.IsNotFound(err) {
 		// create an empty one
-		signingCertKeyPairSecret = &corev1.Secret{ObjectMeta: NewTLSArtifactObjectMeta(
-			c.Name,
-			c.Namespace,
-			c.AdditionalAnnotations,
-		)}
+		signingCertKeyPairSecret = &corev1.Secret{
+			ObjectMeta: NewTLSArtifactObjectMeta(
+				c.Name,
+				c.Namespace,
+				c.AdditionalAnnotations,
+			),
+			Type: corev1.SecretTypeTLS,
+		}
 	}
-	signingCertKeyPairSecret.Type = corev1.SecretTypeTLS
 
 	needsMetadataUpdate := false
+	// no ownerReference set
 	if c.Owner != nil {
 		needsMetadataUpdate = ensureOwnerReference(&signingCertKeyPairSecret.ObjectMeta, c.Owner)
 	}
+	// ownership annotations not set
 	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta) || needsMetadataUpdate
+	// convert outdated secret type (set pre 4.7)
+	if signingCertKeyPairSecret.Type != corev1.SecretTypeTLS {
+		signingCertKeyPairSecret.Type = corev1.SecretTypeTLS
+		needsMetadataUpdate = true
+	}
+	// apply changes (possibly via delete+recreate) if secret exists and requires metadata update
+	// this is done before content update to prevent unexpected rollouts
 	if needsMetadataUpdate && len(signingCertKeyPairSecret.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
+		actualSigningCertKeyPairSecret, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
 		if err != nil {
 			return nil, err
 		}
+		signingCertKeyPairSecret = actualSigningCertKeyPairSecret
 	}
 
 	if needed, reason := needNewSigningCertKeyPair(signingCertKeyPairSecret.Annotations, c.Refresh, c.RefreshOnlyWhenExpired); needed {
--- a/pkg/operator/certrotation/signer_test.go
+++ b/pkg/operator/certrotation/signer_test.go
@@ -15,6 +15,7 @@ import (
 	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
 
+	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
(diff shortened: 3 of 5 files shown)


pkg/operator/certrotation/metadata.go
```diff
@@ -0,0 +1,22 @@
...
+	// no ownerReference set
+	if owner != nil {
+		needsMetadataUpdate = ensureOwnerReference(&secret.ObjectMeta, owner)
+	}
+	// ownership annotations not set
+	needsMetadataUpdate = additionalAnnotations.EnsureTLSMetadataUpdate(&secret.ObjectMeta) || needsMetadataUpdate
+	// convert outdated secret type (created by pre 4.7 installer)
+	if secret.Type != corev1.SecretTypeTLS {
+		secret.Type = corev1.SecretTypeTLS
+		needsMetadataUpdate = true
```
benluddy, line 19: Would it be worthwhile to check that the secret we're converting has `tls.crt` and `tls.key` before updating the type, since updating the type makes validation stricter (https://github.com/kubernetes/kubernetes/blob/d51e2da869350b2b20a33d060e17bd1d2165e02d/pkg/apis/core/validation/validation.go#L6392-L6398), or will it be enough to see UpdateSecretFailed event spam?

benluddy (commented):
I'm still concerned about the recreate path of ApplySecretImproved. If control flow is interrupted between the delete and create (e.g. due to process crash), or the context being used for the client requests is canceled, what happens to the cert data? It seems to me like we would lose it in that case. Even if unlikely, the consequences could be painful. Could we temporarily stage the data in a second secret before deleting the target secret, or something similar?

vrutkovs:
> the context being used for the client requests is canceled, what happens to the cert data? It seems to me like we would lose it in that case

Right, `ApplySecretImproved` is not atomic - imo we should first create a copy, then do delete + recreate. I'd however solve this separately, as it may affect a lot more cases - and doesn't need to be merged urgently

benluddy:
> > the context being used for the client requests is canceled, what happens to the cert data? It seems to me like we would lose it in that case
> 
> Right, `ApplySecretImproved` is not atomic - imo we should first create a copy, then do delete + recreate. I'd however solve this separately, as it may affect a lot more cases - and doesn't need to be merged urgently

Is it safe to solve separately? IIUC, merging this PR to migrate secret types will present a lot of chances to force an unnecessary rotation on older clusters.

vrutkovs:
We already do "migration" in 4.15 with content replace anyway, so this PR helps in happy path. Corner case - cert sync is interrupted right during the migration - can be safely fixed outside of this PR

benluddy:
> "migration"

:joy: Oh, of course. Thanks!

pkg/operator/certrotation/metadata.go
stlaz: Recreated by whom? We don't react to events but the secret content, right?

pkg/operator/certrotation/signer.go
```diff
@@ -62,24 +62,24 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
...
-	if c.Owner != nil {
-		needsMetadataUpdate = ensureOwnerReference(&signingCertKeyPairSecret.ObjectMeta, c.Owner)
-	}
-	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta) || needsMetadataUpdate
-	if needsMetadataUpdate && len(signingCertKeyPairSecret.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
+	// apply necessary metadata (possibly via delete+recreate) if secret exists
+	// this is done before content update to prevent unexpected rollouts
+	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && len(signingCertKeyPairSecret.ResourceVersion) > 0 {
+		actualSigningCertKeyPairSecret, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
```
stlaz, line 78: nit: the following seems as a bit of a cleaner method to achieve the same
```go
var err error
signingCertKeyPairSecret, _, err = resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
```

pkg/operator/certrotation/metadata.go
vrutkovs: Recreated by `ApplySecret`, which attempts apply first, but if it fails it does delete+create. If create with `kubernetes.io/tls` and existing content fails, it would throw an error, delete the content and generate a new cert.

vrutkovs: Updated `ensureMetadataUpdate` to update secret type (removing data if `tls.crt`/`tls.key` are missing) in this case, added unit tests

pkg/operator/certrotation/signer.go
vrutkovs: Following the pattern as in `if needed, reason := needNewSigningCertKeyPair`

pkg/operator/certrotation/metadata.go
```diff
@@ -0,0 +1,37 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ensureMetadataUpdate(secret *corev1.Secret, owner *metav1.OwnerReference, additionalAnnotations AdditionalAnnotations) bool {
+
+	// Existing secret not found - no need to update metadata (will be done by needNewSigningCertKeyPair / NeedNewTargetCertKeyPair)
+	if len(secret.ResourceVersion) == 0 {
+		return false
+	}
```
stlaz, line 13: this probably belongs to the other function, otherwise nothing sets the ownerRef - or does it?

vrutkovs: Yeah, good point, this also shows different function flows

vrutkovs: Fixed

sdodson:
Adding approved label based on https://github.com/openshift/library-go/pull/1681#issuecomment-1976964820

pkg/operator/certrotation/metadata.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func ensureSecretTLSTypeSet(secret *corev1.Secret) bool {
+	// Existing secret not found - no need to update metadata (will be done by needNewSigningCertKeyPair / NeedNewTargetCertKeyPair)
+	if len(secret.ResourceVersion) == 0 {
+		return false
+	}
+
+	// convert outdated secret type (created by pre 4.7 installer)
+	if secret.Type != corev1.SecretTypeTLS {
```
benluddy, line 25: It should behave the same in practice, but since we're correcting for a specific mistake, how about:

```suggestion
	if secret.Type == "SecretTypeTLS" {
```

And as an aside, can we have an issue to track removing this after N releases?

pkg/operator/certrotation/signer_test.go
```diff
@@ -50,13 +51,24 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
 				if len(actual.Data["tls.crt"]) == 0 || len(actual.Data["tls.key"]) == 0 {
 					t.Error(actual.Data)
 				}
+				if len(actual.Annotations) == 0 {
+					t.Errorf("expected certificates to be annotated")
+				}
```
benluddy, line 56: This looks redundant, any value for annotations that would fail here is also going to fail below.

pkg/operator/certrotation/signer_test.go
```diff
@@ -50,13 +51,24 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
...
+				ownershipValue, found := actual.Annotations[annotations.OpenShiftComponent]
```
benluddy, line 57: I prefer to avoid sharing values like this between tests and implementation, since it creates the opportunity to accidentally make a breaking change without seeing a test failure.

Feel free to ignore if you disagree or are comfortable with it in this instance because it is declared in openshift/api.

pkg/operator/certrotation/signer_test.go
```diff
@@ -96,6 +121,90 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
...
+				}
+
+				if !actions[0].Matches("get", "secrets") {
+					t.Error(actions[0])
+				}
+				if !actions[1].Matches("delete", "secrets") {
+					t.Error(actions[1])
+				}
+				if !actions[2].Matches("create", "secrets") {
+					t.Error(actions[0])
```
benluddy, line 150:
```suggestion
					t.Error(actions[2])
```

pkg/operator/certrotation/signer_test.go
```diff
@@ -96,6 +121,90 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
...
```
benluddy, line 199:
```suggestion
					t.Error(actions[2])
```

pkg/operator/certrotation/target_test.go
```diff
@@ -190,6 +201,132 @@ func TestEnsureTargetCertKeyPair(t *testing.T) {
...
+				}
+
+				if !actions[0].Matches("get", "secrets") {
+					t.Error(actions[0])
+				}
+				if !actions[1].Matches("delete", "secrets") {
+					t.Error(actions[1])
+				}
+				if !actions[2].Matches("create", "secrets") {
+					t.Error(actions[0])
```
benluddy, line 249: There are few more of these:

```suggestion
					t.Error(actions[2])
```

pkg/operator/certrotation/signer.go
```diff
@@ -62,24 +62,24 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
...
-	needsMetadataUpdate := false
-	if c.Owner != nil {
-		needsMetadataUpdate = ensureOwnerReference(&signingCertKeyPairSecret.ObjectMeta, c.Owner)
-	}
-	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta) || needsMetadataUpdate
-	if needsMetadataUpdate && len(signingCertKeyPairSecret.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
+	// apply necessary metadata (possibly via delete+recreate) if secret exists
+	// this is done before content update to prevent unexpected rollouts
+	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(signingCertKeyPairSecret) {
```
benluddy, line 77: Why is this `&&` and not `||`?

pkg/operator/certrotation/signer_test.go
vrutkovs: Yes, but I'd prefer to a clear error message

pkg/operator/certrotation/signer_test.go
vrutkovs: >it creates the opportunity to accidentally make a breaking change without seeing a test failure

We explicitly set `AdditionalAnnotations` in https://github.com/openshift/library-go/pull/1681/files#diff-d7a4e30c6635930d8721bf060fe2803587326298e0eace9bada256f1b7377459R228-R230

pkg/operator/certrotation/signer.go
vrutkovs: We want to update labels before content changes only when:
* secret type won't change (`ensureSecretTLSTypeSet` returns true)
* metadata (annotations, ownership) needs updating (`ensureMetadataUpdate` returns true)

The function are also mutating `signingCertKeyPairSecret`, changing it type and clearing data, so the secret would be deleted + created during `if needed, reason := needNewSigningCertKeyPair` clause. It needs to be done once to avoid updating it twice for metadata and content ("update no annotations" and "update SecretTLSType secrets" unit test cases)

stlaz: You'll want to make sure your secret always passes both functions, right?

Either operator might short-circuit in a different scenario.

vrutkovs: >You'll want to make sure your secret always passes both functions, right?

correct, if either doesn't pass possible updates will be applied by ApplySecret in `needNewSigningCertKeyPair`

pkg/operator/certrotation/metadata.go
benluddy: Right, I understand, but my point is that we expect any secret managed by this controller will either have type `kubernetes.io/tls` or, if it was created pre-4.7, `SecretTypeTLS`. If we discover that there is some instance of a managed secret with a completely unexpected type, I would rather understand why our expectation was wrong than stomp it.

vrutkovs: Its perfectly possible for users to find a way to replace these certificates behind cert-syncer back and replace it with invalid contents. This code will ensure that we attempt to convert cases like pre-4.7 if possible, but also prioritize availability - invalid content would be stomped for cluster to continue functioning. Why the bad secret was stomped can be found out via audit logs (or etcd contents).
This change will also protects us from other similar cases like "old/misconfigured installer has generated certificates which we no longer expect"

pkg/operator/certrotation/signer_test.go
benluddy: That link didn't work for me. In this case I'm specifically referring to a change to the constant `OpenShiftComponent`, which would be a breaking change but would not cause this test to fail.

vrutkovs: Ah, hmm, `JiraComponent` and `OpenshiftComponent` are linked in https://github.com/openshift/library-go/blob/master/pkg/operator/certrotation/annotations.go#L27-L30, not sure how to make sure unit tests don't depend on it though

pkg/operator/certrotation/signer_test.go
vrutkovs: Fixed

pkg/operator/certrotation/signer.go
benluddy: > so the secret would be deleted + created during if needed, reason := needNewSigningCertKeyPair clause

Thanks, this is the part that I was missing.

pkg/operator/certrotation/signer_test.go
benluddy: Fine with me given that nothing out of the control of this test should be touching annotations.

PR #1687, opened 2024-03-06
certrotation: cover corner case when invalid secret type has necessary metadata

vrutkovs:
In this case cert-rotation won't convert secret type as `ensureSecretTLSTypeSet` would never run

### DIFF
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -87,7 +87,9 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 
 	// apply necessary metadata (possibly via delete+recreate) if secret exists
 	// this is done before content update to prevent unexpected rollouts
-	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(signingCertKeyPairSecret) {
+	needsMetadataUpdate := ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
+	needsTypeUpdate := ensureSecretTLSTypeSet(signingCertKeyPairSecret)
+	if needsMetadataUpdate || needsTypeUpdate {
 		actualSigningCertKeyPairSecret, _, err := applyFn(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
 		if err != nil {
 			return nil, false, err
--- a/pkg/operator/certrotation/signer_test.go
+++ b/pkg/operator/certrotation/signer_test.go
@@ -276,7 +276,7 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
 			verifyActions: func(t *testing.T, updateOnly bool, client *kubefake.Clientset, controllerUpdatedSecret bool) {
 				t.Helper()
 				actions := client.Actions()
-				if len(actions) != 2 {
+				if len(actions) != 4 {
 					t.Fatal(spew.Sdump(actions))
 				}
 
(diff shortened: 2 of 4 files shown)


vrutkovs:
/cc @stlaz @benluddy 

This covers a corner case Standa found, follow up for https://github.com/openshift/library-go/pull/1681#discussion_r1511899808

pkg/operator/certrotation/signer.go
```diff
@@ -87,7 +87,9 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 
 	// apply necessary metadata (possibly via delete+recreate) if secret exists
 	// this is done before content update to prevent unexpected rollouts
-	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(signingCertKeyPairSecret) {
+	needsMetadataUpdate := ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
+	needsTypeUpdate := ensureSecretTLSTypeSet(signingCertKeyPairSecret)
```
p0lyn0mial, line 91: Why does ensureSecretTLSTypeSet have to mess with the content of the secret?

xref: https://github.com/openshift/library-go/blob/0b298992f8cc44a4dc612800bd6766acdba97008/pkg/operator/certrotation/metadata.go#L18

pkg/operator/certrotation/signer_test.go
```diff
@@ -276,7 +276,7 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
 			verifyActions: func(t *testing.T, updateOnly bool, client *kubefake.Clientset, controllerUpdatedSecret bool) {
 				t.Helper()
 				actions := client.Actions()
-				if len(actions) != 2 {
+				if len(actions) != 4 {
```
p0lyn0mial, line 279: why does the create scenario sends additional `get` and `update` requests?

pkg/operator/certrotation/signer_test.go
```diff
@@ -319,9 +325,16 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
 		{
 			name: "update no annotations",
 			initialSecret: &corev1.Secret{
-				ObjectMeta: metav1.ObjectMeta{Namespace: "ns", Name: "signer", ResourceVersion: "10"},
-				Type:       corev1.SecretTypeTLS,
-				Data:       map[string][]byte{"tls.crt": {}, "tls.key": {}},
+				ObjectMeta: metav1.ObjectMeta{Namespace: "ns", Name: "signer",
+					ResourceVersion: "10",
+					Annotations: map[string]string{
```
p0lyn0mial, line 330: why the change, the scenario says `update no annotations`

pkg/operator/certrotation/signer.go
vrutkovs: `ensureSecretTLSTypeSet` may change secret type if its not `kubernetes.io/tls`

pkg/operator/certrotation/signer_test.go
vrutkovs: hmm, I should rename this to "no necessary signer annotations"

pkg/operator/certrotation/signer_test.go
vrutkovs: Previously only one `create` was invoked, because `ensureMetadataUpdate` returned true (no ownership annotations/refs) and `ensureSecretTLSTypeSet` returned `false` (if ResourceVersion is unset no type change necessary). That means the secret was created after `needNewSigningCertKeyPair` was returning `true`.

With this change `needsMetadataUpdate || needsTypeUpdate` and secret is created with empty data and required annotaitons/ownerrefs, later being updated with secret contents when `needNewSigningCertKeyPair` returns `true`.

This may seem like extra actions, but it protects us from unexpected changes when secret contents and metadata are fully compliant, but secret type is still invalid. Its not possible in the standard flow for born-in-4.7-clusters, but may be possible in later cycles

pkg/operator/certrotation/signer_test.go
vrutkovs: Done

vrutkovs:
Included in https://github.com/openshift/library-go/pull/1693

PR #1693, opened 2024-03-13
WIP certrotation: don't let multiple controllers rewrite metadata of the shared resource

vrutkovs:
This prevents http 409 errors when both labels and content need updating in one sync. See this [test failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1655/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1767865041786245120) when new library-go is being pulled in cluster-kube-apiserver-operator

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -58,11 +58,19 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		)}
 	}
 
+	isPrimaryOwner := false
 	needsMetadataUpdate := false
 	if c.Owner != nil {
 		needsMetadataUpdate = ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
+		if caBundleConfigMap.ObjectMeta.OwnerReferences[0] == *c.Owner {
+			isPrimaryOwner = true
+		}
 	}
-	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta) || needsMetadataUpdate
+	// Don't update metadata if its not primary CM owner
+	if isPrimaryOwner {
+		needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta) || needsMetadataUpdate
+	}
+
 	if needsMetadataUpdate && len(caBundleConfigMap.ResourceVersion) > 0 {
 		actualCABundleConfigMap, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
 		if err != nil {
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -64,10 +64,11 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta) || needsMetadataUpdate
 	if needsMetadataUpdate && len(caBundleConfigMap.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		actualCABundleConfigMap, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
 		if err != nil {
 			return nil, err
 		}
+		caBundleConfigMap = actualCABundleConfigMap
 	}
 
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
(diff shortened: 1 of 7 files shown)


vrutkovs:
/hold

Testing this in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1655

vrutkovs:
/hold cancel

Seems to work as expected

pkg/operator/certrotation/cabundle.go
```diff
@@ -64,10 +64,11 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta) || needsMetadataUpdate
 	if needsMetadataUpdate && len(caBundleConfigMap.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		actualCABundleConfigMap, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
```
p0lyn0mial, line 67: do you mind adding a unit test ?

p0lyn0mial:
In the future we could refactor the controller to only issue a single update when meta and content require updating.

p0lyn0mial:
@vrutkovs a wrong update ? I don't see a unit test.

pkg/operator/certrotation/cabundle.go
vrutkovs: Done, not sure if it really ensures hotlooping is not happening (imo fuzzing would confirm it)

p0lyn0mial: This PR doesn't solve the hotloop issue. The hotloop issue was resolved by https://github.com/openshift/cluster-kube-apiserver-operator/pull/1661.

There were `4` controllers fighting over `kube-control-plane-signer-ca`

p0lyn0mial:
> See this [test failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1655/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1767865041786245120) when new library-go is being pulled in cluster-kube-apiserver-operator

For the record, the above failure was due to https://github.com/openshift/cluster-kube-apiserver-operator/pull/1661 not due to the issue addressed by this PR.

pkg/operator/certrotation/cabundle_test.go
```diff
@@ -64,6 +63,66 @@ func TestEnsureConfigMapCABundle(t *testing.T) {
...
+				if !actions[0].Matches("get", "configmaps") {
+					t.Error(actions[0])
+				}
+				if !actions[1].Matches("update", "configmaps") {
+					t.Error(actions[1])
+				}
+				if !actions[2].Matches("get", "configmaps") {
+					t.Error(actions[2])
+				}
+				if !actions[3].Matches("update", "configmaps") {
```
p0lyn0mial, line 106: the second call is unnecessary, we should refactor the controller to issue only a single API call.

p0lyn0mial: in addition it is worth to also examine the payload of the request (updated configmap), please add it to your todo list.

p0lyn0mial:
/assign @stlaz 

for approval

pkg/operator/certrotation/cabundle.go
vrutkovs: Fixed this by allowing metadata change to the first controller only (whoever has set the first ownership ref).

I'll run this a few time to make sure this won't make TLS metadata jitter from run to run

pkg/operator/certrotation/cabundle.go
```diff
@@ -58,16 +58,25 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		)}
 	}
 
+	isPrimaryOwner := false
 	needsMetadataUpdate := false
 	if c.Owner != nil {
 		needsMetadataUpdate = ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
+		if caBundleConfigMap.ObjectMeta.OwnerReferences[0] == *c.Owner {
```
p0lyn0mial, line 65: Why can't we just have a single owner?

vrutkovs: CA bundle can contain multiple signers, controlled by different controllers. We could have a single ownerRef, but that wouldn't reflect reality. It also makes easier to find such CA bundles in the must-gathers

p0lyn0mial: > CA bundle can contain multiple signers, controlled by different controllers

This is wrong and should be fixed. Why can't we have a single signer ?
Could you point me to code that does this ?

p0lyn0mial: Still, there is a single signer and a CA bundle. 
Also as far as I can tell `kas-o` doesn't make use of the `Owner` field for both signer and CA resources.

p0lyn0mial: I don't understand why we have to have multiple controllers controlling a single resource. I would prefer to have a single controller for that.

It seems to me that we would like to mark which components are using the resource, which could be done with a single controller.

vrutkovs:
Running one controller for a resource would fix this and few more problems, closing in favor of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669

PR #1694, opened 2024-03-15
OCPBUGS-24195: the apiservice controller waits until bootstrap complete

p0lyn0mial:
the proof PR at https://github.com/openshift/cluster-authentication-operator/pull/662

### DIFF
--- a/pkg/operator/apiserver/controller/apiservice/apigroup.go
+++ b/pkg/operator/apiserver/controller/apiservice/apigroup.go
@@ -9,18 +9,21 @@ import (
 
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	kubeinformers "k8s.io/client-go/informers"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 
+	"github.com/openshift/library-go/pkg/operator/bootstrap"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
-func preconditionsForEnabledAPIServices(kubeInformers kubeinformers.SharedInformerFactory) func(apiServices []*apiregistrationv1.APIService) (bool, error) {
-	endpointsLister := kubeInformers.Core().V1().Endpoints().Lister()
+func preconditionsForEnabledAPIServices(endpointsListerForTargetNs corev1listers.EndpointsLister, configmapListerForKubeSystemNs corev1listers.ConfigMapLister) func(apiServices []*apiregistrationv1.APIService) (bool, error) {
 	return func(apiServices []*apiregistrationv1.APIService) (bool, error) {
-		return checkEndpointsPresence(endpointsLister, apiServices)
+		areEndpointsPresent, err := checkEndpointsPresence(endpointsListerForTargetNs, apiServices)
+		if !areEndpointsPresent || err != nil {
+			return false, err
+		}
+		return bootstrap.IsBootstrapComplete(configmapListerForKubeSystemNs)
 	}
 }
 
--- a/pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
+++ b/pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
@@ -9,12 +9,14 @@ import (
 	"strings"
 	"testing"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/fake"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	kubetesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
 	apiregistrationv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
@@ -29,13 +31,14 @@ import (
 
 func TestAvailableStatus(t *testing.T) {
 	testCases := []struct {
-		name                string
-		expectedStatus      operatorv1.ConditionStatus
-		expectedReasons     []string
-		expectedMessages    []string
-		existingAPIServices []runtime.Object
-		apiServiceReactor   kubetesting.ReactionFunc
-		daemonReactor       kubetesting.ReactionFunc
+		name                               string
+		expectedStatus                     operatorv1.ConditionStatus
+		expectedReasons                    []string
+		expectedMessages                   []string
+		existingAPIServices                []runtime.Object
+		apiServiceReactor                  kubetesting.ReactionFunc
+		daemonReactor                      kubetesting.ReactionFunc
+		preconditionsForEnabledAPIServices apiServicesPreconditionFuncType
 	}{
 		{
 			name:           "Default",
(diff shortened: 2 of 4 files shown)


p0lyn0mial:
only the last commit is relevant here https://github.com/openshift/library-go/pull/1694/commits/59904ac938942f9307e1c1f64bba310592523062, the previous 4 were taken from https://github.com/openshift/library-go/pull/1715

benluddy:
/lgtm cancel

Wrong tab...

benluddy:
:sweat_smile:
/hold cancel

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
```diff
@@ -147,6 +150,25 @@ func TestAvailableStatus(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "PreconditionsError",
+			preconditionsForEnabledAPIServices: func([]*apiregistrationv1.APIService) (bool, error) {
+				return false, fmt.Errorf("dummy error")
```
benluddy, line 156: What about coverage for `return true, fmt.Errorf("dummy error")`?

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
```diff
@@ -412,6 +437,89 @@ func TestDisabledAPIService(t *testing.T) {
...
+func TestPreconditionsForEnabledAPIServices(t *testing.T) {
+	var (
+		bootstrapComplete = &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: "bootstrap", Namespace: "kube-system"},
+			Data:       map[string]string{"status": "complete"},
+		}
+
+		bootstrapNoStatus = &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: "bootstrap", Namespace: "kube-system"},
+			Data:       map[string]string{"noStatus": "complete"},
```
benluddy, line 449: Is "noStatus" an arbitrary non-"status" key?

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
```diff
@@ -412,6 +437,89 @@ func TestDisabledAPIService(t *testing.T) {
...
+		existingEndpoints          []*corev1.Endpoints
+
+		expectedStatus bool
+	}{
+		{
+			name: "EmptyAPIServices, NoBootstrapConfigMap",
+		},
+		{
+			name:                       "EmptyAPIServices, IncompleteBootstrapConfigMap",
+			existingBootstrapConfigMap: bootstrapNoStatus,
```
benluddy, line 466: Your call, but I would prefer inlining *corev1.ConfigMap literals here unless they were enormous or reused many times. That would save me as a reader from doing the lookup and mental translation between the variable identifier and the contents of the configmap.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
```diff
@@ -412,6 +437,89 @@ func TestDisabledAPIService(t *testing.T) {
...
+					e := &corev1.Endpoints{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      s.Spec.Service.Name,
+							Namespace: s.Spec.Service.Namespace,
+						},
+					}
+					return e
+				}(),
+			},
+		},
```
benluddy, line 490: How about a case where the "status" key is present in the configmap but has a value other than "complete"?

pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -9,18 +9,21 @@ import (
...
 )
 
-func preconditionsForEnabledAPIServices(kubeInformers kubeinformers.SharedInformerFactory) func(apiServices []*apiregistrationv1.APIService) (bool, error) {
-	endpointsLister := kubeInformers.Core().V1().Endpoints().Lister()
+func preconditionsForEnabledAPIServices(endpointsListerForTargetNs corev1listers.EndpointsLister, configmapListerForKubeSystemNs corev1listers.ConfigMapLister) func(apiServices []*apiregistrationv1.APIService) (bool, error) {
 	return func(apiServices []*apiregistrationv1.APIService) (bool, error) {
-		return checkEndpointsPresence(endpointsLister, apiServices)
+		areEndpointsPresent, err := checkEndpointsPresence(endpointsListerForTargetNs, apiServices)
+		if !areEndpointsPresent || err != nil {
+			return areEndpointsPresent, err
```
benluddy, line 24: Nit: I prefer this for clarity as a reader.

```suggestion
			return false, err
```

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
p0lyn0mial: yes.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
p0lyn0mial: np, can change.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
p0lyn0mial: already covered in https://github.com/openshift/library-go/blob/master/pkg/operator/bootstrap/bootstrap_test.go#L20, do we want to duplicate that case here as well?

pkg/operator/apiserver/controller/apiservice/apigroup.go
p0lyn0mial: sure, np.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
p0lyn0mial: https://github.com/openshift/library-go/pull/1694/files#diff-03f35e8de3821373388cc9e73d45c2c7ab8883cc85c911df416e41cca81c7350R161 ?

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
benluddy: I think it would be a good idea to either duplicate it or inject `IsBootstrapComplete`. We're testing the behavior of `preconditionsForEnabledAPIServices` here, so we can't rely on the knowledge that its implementation calls `IsBootstrapComplete` which has test coverage for this case.

p0lyn0mial: ok, extended the test.

pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go
benluddy: Is there a test covering a case where `preconditionsForEnabledAPIServices` returns `true` _and_ a non-nil error?

p0lyn0mial: yes, all the other cases defined in this test, by default `preconditionsForEnabledAPIServices` is set to` func([]*apiregistrationv1.APIService) (bool, error) { return true, nil }` - https://github.com/openshift/library-go/blob/0cd17b55520a43287c67078fd984663e6d91e8a6/pkg/operator/apiserver/controller/apiservice/apiservice_controller_test.go#L200

benluddy: But the returned error value is nil there?

p0lyn0mial: ah, sorry, i misread your comment, okay, we can add one more case to this test.

p0lyn0mial: done.

benluddy:
/lgtm

Thanks!

PR #1699, opened 2024-03-27
certrotation: a unit test for the cert issue

p0lyn0mial:
`loadbalancer-serving-signer secret` in `openshift-kube-apiserver-operator` namespace
is reconciled by multiple controllers at the same time without any coordination.
it looks like we have `2` separate processes and `4` different controllers competing over the same resource.

the unit test created in this PR demonstrates that the secret crypto material
can be regenerated, which has serious consequences for the platform
as it can break external clients and the cluster itself.

The test, runs two controllers `A` and `B`, esentially
1.  `A` deletes via ApplySecret
2.  `B` observes NotFound, generates new signer
3.  `A` re-creates via ApplySecret with the old content.
4.  `B` calls ApplySecret, it observes a diff and calls Update, stomping original contents

The race condition [was introduced a long time](https://github.com/openshift/cluster-kube-apiserver-operator/pull/405/files#diff-ac3e17f1cb0499adf65a2c4b25d58416b9eb47d39f4ee00eb18773de7bbd2afeR244) ago but wasn't triggered because the secret wasn't updated. In 4.15, a new annotation to the secret was added, which caused the race condition to occur.

Since the code is racy other scenarios are also possible, for example:
 1.  `A` deletes via ApplySecret
2.  `B` observes NotFound and generates new signer
3.  `B` creates via ApplySecret, with the new content
4.  `A` fails on re-creating

controllers reconciling the secret:

- https://github.com/openshift/cluster-kube-apiserver-operator/blob/release-4.15/pkg/operator/certrotationcontroller/certrotationcontroller.go#L338
- https://github.com/openshift/cluster-kube-apiserver-operator/blob/release-4.15/pkg/operator/certrotationcontroller/certrotationcontroller.go#L392

processes reconciling the secret:
- https://github.com/openshift/cluster-kube-apiserver-operator/blob/63691a0ee938b1b7002777db8464d50791efb94a/pkg/operator/starter.go#L318
- https://github.com/openshift/cluster-kube-apiserver-operator/blob/052beea48fdeb9be7cdafe7a5ae2e7c4c74c3a0f/pkg/cmd/certregenerationcontroller/cmd.go#L105

In general, `NewCertRotationController` react to changes on secrets in
`openshift-kube-apiserver-operator` ns (https://github.com/openshift/cluster-kube-apiserver-operator/blob/ae0c9631284bca8cfcbc305674781e097f7238ae/vendor/github.com/openshift/library-go/pkg/operator/certrotation/client_cert_rotation_controller.go#L94)

### DIFF
--- a/pkg/operator/certrotation/signer_test.go
+++ b/pkg/operator/certrotation/signer_test.go
@@ -1,23 +1,208 @@
 package certrotation
 
 import (
+	"bytes"
 	"context"
+	"reflect"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
 
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	kubefake "k8s.io/client-go/kubernetes/fake"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openshift/api/annotations"
+	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
+
+// loadbalancer-serving-signer secret in openshift-kube-apiserver-operator namespace
+// is reconciled by multiple controllers at the same time without any coordination.
+// it looks like we have `2` separate processes and `4` different controllers competing over the same resource.
+//
+// the following unit test demonstrates that the secret crypto material
+// can be regenerated, which has serious consequences for the platform
+// as it can break external clients and the cluster itself.
+//
+// controllers reconciling the secret:
+//  - https://github.com/openshift/cluster-kube-apiserver-operator/blob/release-4.15/pkg/operator/certrotationcontroller/certrotationcontroller.go#L338
+//  - https://github.com/openshift/cluster-kube-apiserver-operator/blob/release-4.15/pkg/operator/certrotationcontroller/certrotationcontroller.go#L392
+
+// processes reconciling the secret:
+//   - https://github.com/openshift/cluster-kube-apiserver-operator/blob/63691a0ee938b1b7002777db8464d50791efb94a/pkg/operator/starter.go#L318
+//   - https://github.com/openshift/cluster-kube-apiserver-operator/blob/052beea48fdeb9be7cdafe7a5ae2e7c4c74c3a0f/pkg/cmd/certregenerationcontroller/cmd.go#L105
+//
+// In general, NewCertRotationController react to changes on secrets in
+// openshift-kube-apiserver-operator ns (https://github.com/openshift/cluster-kube-apiserver-operator/blob/ae0c9631284bca8cfcbc305674781e097f7238ae/vendor/github.com/openshift/library-go/pkg/operator/certrotation/client_cert_rotation_controller.go#L94)
+func TestEnsureSigningCertKeyPairRace(t *testing.T) {
+	secondControllerStartChan := make(chan struct{})
+	secondControllerEndChan := make(chan struct{})
+	firstControllerEndChan := make(chan struct{})
+
+	// represents a secret that was created before 4.7 and
+	// hasn't been updated until now (upgrade to 4.15)
+	existingSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "ns", Name: "signer", ResourceVersion: "10",
+			Annotations: map[string]string{
+				"auth.openshift.io/certificate-not-after":  "2108-09-08T22:47:31-07:00",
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
@tkashem can you please double check if the test make sense ?

p0lyn0mial:
/hold

this test shows the issue which must be resolved and then we should also update the test.

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,216 @@
...
+	certBytesFromControllerTwo, keyBytesFromControllerTwo, err := signingKeyPairFromControllerTwo.Config.GetPEMBytes()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// note that GetTLSCertificateConfigFromBytes (executed by the target method) check if
+	// the crypto material is not empty
+	//
+	// the following 2 checks show that the second controller
+	// created a new signing key pair!
+	if bytes.Equal(certBytesFromControllerOne, certBytesFromControllerTwo) {
```
p0lyn0mial, line 172: this should have been `if !bytes.Equal(...)`

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,216 @@
...
 )
 
+// loadbalancer-serving-signer secret in openshift-kube-apiserver-operator namespace
+// is reconciled by multiple controllers at the same time without any coordination.
+//
+// the following unit test demonstrates that the secret crypto material
+// can be regenerated, which has serious consequences for the platform
+// as it can break external clients and the cluster itself.
+//
+// controllers reconciling the secret:
```
p0lyn0mial, line 35: it looks like we have `2` separate processes and `4` different controllers competing over the same resource.

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,219 @@
...
+
+			// cannot use fakeKubeClient.ActionsA()
+			// because it deadlocks, fakeKubeClient.RLock already acquired
+			//
+			// the created secret from the actions
+			// is validated at the end of this test
+			// and must match this one
+			createdSecret := existingSecret.DeepCopy()
+			createdSecret.Type = corev1.SecretTypeTLS
+			createdSecret.Annotations["openshift.io/owning-component"] = "test"
```
benluddy, line 115: I don't understand this part. Won't the default behavior of the test clientset return the object you expect (i.e. the one created by the first controller) without this?

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,219 @@
...
+		t.Fatal(err)
+	}
+	if !bytes.Equal(existingSecret.Data["tls.crt"], certBytesFromControllerOne) {
+		t.Fatal("returned certificate data from controller one doesn't match existing secreted data")
+	}
+	if !bytes.Equal(existingSecret.Data["tls.key"], keyBytesFromControllerOne) {
+		t.Fatal("returned certificate key from controller one doesn't match existing secreted data")
+	}
+	close(firstControllerEndChan)
+	<-secondControllerEndChan
```
benluddy, line 164: In general, it might be a good idea to have a timeout on all of these channel receive statements. A change to the one of the functions covered by this test could make the test itself timeout or deadlock instead of failing in an obvious way.

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,219 @@
...
+		t.Fatal("cert key from both controllers are equal")
+	}
+
+	actions := fakeKubeClient.Actions()
+	updateActionValidated := false
+	createActionValidated := false
+	for _, action := range actions {
+		switch action.GetVerb() {
+		case "create":
+			if createAction, ok := action.(clienttesting.CreateAction); ok {
```
benluddy, line 188: For future-proofing against silently skipping these assertions, how about failing if !ok?

pkg/operator/certrotation/signer_test.go
p0lyn0mial: yes, I agree it might be confusing

pkg/operator/certrotation/signer_test.go
```diff
@@ -1,24 +1,219 @@
...
+//
+// In general, NewCertRotationController react to changes on secrets in
+// openshift-kube-apiserver-operator ns (https://github.com/openshift/cluster-kube-apiserver-operator/blob/ae0c9631284bca8cfcbc305674781e097f7238ae/vendor/github.com/openshift/library-go/pkg/operator/certrotation/client_cert_rotation_controller.go#L94)
+func TestEnsureSigningCertKeyPairRace(t *testing.T) {
+	secondControllerStartChan := make(chan struct{})
+	secondControllerEndChan := make(chan struct{})
+	firstControllerEndChan := make(chan struct{})
+
+	// represents a secret that was created before 4.7 and
+	// hasn't been updated until now (upgrade to 4.15)
```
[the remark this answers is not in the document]
p0lyn0mial: Yes, the secret must not have `kubernetes.io/tls` because the` ensureSecretTLSTypeSet` method will prevent the delete and re-create path. However, the method is quite new and was added a few weeks ago. So, not all versions have this patch. Besides, clusters created **before** version `4.7` have a different secret type set.

p0lyn0mial:
/close

pulled into https://github.com/openshift/library-go/pull/1705

PR #1701, opened 2024-03-28
Use syscalls conditionally only for linux

soltysh:
/assign @ardaguclu @deads2k


ardaguclu:
/lgtm
/hold
in case @deads2k would like to look at

PR #1703, opened 2024-03-29
[poc] prevent deletion

deads2k:
/hold 

I'm not going to take it further, this is just to illustrate the concept.

### DIFF
--- a/pkg/operator/resource/resourceapply/core.go
+++ b/pkg/operator/resource/resourceapply/core.go
@@ -430,31 +430,8 @@ func ApplySecretImproved(ctx context.Context, client coreclientv1.SecretsGetter,
 	}
 
 	var actual *corev1.Secret
-	/*
-	 * Kubernetes validation silently hides failures to update secret type.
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/core.go
```diff
@@ -430,31 +430,8 @@ func ApplySecretImproved(ctx context.Context, client coreclientv1.SecretsGetter,
 	}
 
 	var actual *corev1.Secret
-	/*
-	 * Kubernetes validation silently hides failures to update secret type.
-	 * https://github.com/kubernetes/kubernetes/blob/98e65951dccfd40d3b4f31949c2ab8df5912d93e/pkg/apis/core/validation/validation.go#L5048
-	 * We need to explicitly opt for delete+create in that case.
-	 */
-	if existingCopy.Type == existing.Type {
```
p0lyn0mial, line 438: Instead of removing this code I think that we should relax is so that our special secrets are updated even if the type is different (xref: https://github.com/openshift/kubernetes/pull/1923)

PR #1708, opened 2024-04-04
Add Communication Matrix Module

liornoy:
This commit adds a new module for generating a communication matrix. The communication matrix lists the open ports in the cluster and provides different formats to print the results.

Use the `/pkg/network/commatrix/cmd/Makefile` to generate the matrix. The OUTPUT variable is the desired output format, which can be set to YAML, CSV, or JSON.
The CLUSTER_ENV variable is the environment in which the cluster is deployed; currently, it can be set to either `baremetal` or `aws`.


liornoy:
closing this in favor of openshift-kni/commatrix repo

PR #1709, opened 2024-04-05
WIP NO-JIRA: experimental adaptation of signer rotation test as fuzz target

vrutkovs:


benluddy:
Thanks for running with this. There were a few things that I wanted to look into but did not have time:

1. The fuzz input is just a PRNG seed, so it doesn't really have an correlation with the outcome of the test. In its current state, I don't think we're benefiting from fuzzing vs. simply iterating over int64s sequentially. What might work is changing the fuzz input to represent the sequence of choices that the dispatcher makes. For example, if you have fewer than 256 instrumented actors, the fuzz input could be `[]byte` and the dispatcher would do `for _, n := range choices { dispatch(actors[n%len(actors)]) }`.
2. The dispatcher relies on knowing how many instrumented actors there are (and when they terminate) so that it can wait until every actor is blocked before it dispatches one of them. It's easy to use it wrong in a test and deadlock. I would like to see some kind of dispatcher builder that tests can use to register an actor and get back a lightweight handle for blocking and unregistering (and after unregistering, the handle only returns errors). Or any better design that makes it harder to misuse in tests.
3. The sleep at the beginning of the fuzz target badly limits the fuzzer throughput.
4. Instead of, or in addition to, having actors block on string action names, it could be nice to save the call stack.

vrutkovs:
Implemented a list of choices (which replaces rng and loops over number of workers). Also added fuzzing over number of workers. However its odd that although `UseSecretUpdateOnly` is fuzzed too I don't see failures :/

vrutkovs:
Addressed only 1 from https://github.com/openshift/library-go/pull/1709#issuecomment-2039960940 so far

PR #1713, opened 2024-04-07
[WIP] test framework to test concurrent controllers with linearization

tkashem:


pkg/operator/certrotation/signer.go
```diff
@@ -183,7 +183,7 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 
 // setSigningCertKeyPairSecret creates a new signing cert/key pair and sets them in the secret
 func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validity time.Duration) error {
-	signerName := fmt.Sprintf("%s_%s@%d", signingCertKeyPairSecret.Namespace, signingCertKeyPairSecret.Name, time.Now().Unix())
+	signerName := fmt.Sprintf("%s_%s@%d", signingCertKeyPairSecret.Namespace, signingCertKeyPairSecret.Name, time.Now().UnixNano())
```
vrutkovs, line 186: That seems to be required for tests only, right? I don't think its suitable for production though, as it may grow too large?

pkg/testing/linearizer/linearizer.go
```diff
@@ -0,0 +1,252 @@
...
+	"fmt"
+	"math/rand"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"testing"
+)
+
+/*
+   A controller has this sequence: Get, Create, and Update
```
vrutkovs, line 13: `Create` here assumes "Delete and Create" on type change or its purely "Create" when the item doesn't exist?

pkg/operator/certrotation/signer.go
tkashem: yes, for tests only, if we want to make sure the annotation changes as expected, otherwise we have to add `sleep` to the test

pkg/testing/linearizer/linearizer.go
tkashem: this is just an example to demonstrate how the framework generates the API call sequence, we could replace the Create/Delete with DoFoo/DoBar here

pkg/testing/linearizer/linearizer.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func label(actorID int) string { return fmt.Sprintf("actor-%d", actorID) }
+
+type actor struct {
+	id int
+	d  *dispatcher
+}
+
+func (a actor) Start() <-chan struct{} {
```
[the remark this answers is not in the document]
tkashem: it's so that we can linearize the startup logic as well, the test here doesn't instrument the lister logic and a controller can start with a get from the lister.
A lister reacts to the state of the storage, so as long as we linearize all the API invocations we are good, this also means that the startup logic must me linearized as well.

tkashem:
graphic visualization of an example  run, where two controller instances `{0,1}` run concurrently, each path from the `start` node to a leaf node represents a linearized run where the API calls from the two instances are interleaved. The path marked in `red` shows a failed run.

image

vrutkovs:
This may be useful in case we won't be able to avoid having a proper leader election for copies of the controller - or we find a resource managed by multiple controllers.

PR #1716, opened 2024-04-14
Add Communication Matrix Module

sabinaaledort:
This commit adds a new module to generate a communication matrix.
The communication matrix lists the open ports in the cluster and
provides different formats to print the results.

Use the `/pkg/network/commatrix/Makefile` to generate the matrix.
The OUTPUT variable is the desired output format, which can be set to YAML, CSV,
or JSON.
The CLUSTER_ENV variable is the environment in which the cluster is deployed;
currently, it can be set to either `baremetal` or `aws`.


sabinaaledort:
Stored in https://github.com/openshift-kni/commatrix

PR #1717, opened 2024-04-15
cert collection: skip secret if it contains bootstrap certs

vrutkovs:
etcd leaves bootstrap certificates around, which cannot be anonymized in RewriteNodeIPs. This function receives bootstrap IP, fetched from an annotation in `etcd-endpoints-*` configmap in `openshift-etcd` namespace.

Tested in https://github.com/openshift/origin/pull/28657

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collection_options.go
@@ -56,6 +56,19 @@ var (
 	}
 )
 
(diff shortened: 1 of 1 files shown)


deads2k:
>which cannot be anonymized in RewriteNodeIPs
Why can't we write that IP to `<bootstrap-host>`?

deads2k:
why can't something like https://github.com/openshift/library-go/pull/1723 rewrite the bootstrapIP?

PR #1718, opened 2024-04-15
WRKLDS-1167: tolerate node-role.kubernetes.io/control-plane:NoExecute 

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.

installer-pods: already tolerate every taint

pruner-pods: already tolerate every taint

guard-pods: toleration to relevant NoExecute taint added

monitor-pods: already tolerate every taint

### DIFF
--- a/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
+++ b/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
@@ -13,21 +13,7 @@ spec:
   priorityClassName: "system-cluster-critical"
   terminationGracePeriodSeconds: 3
   tolerations:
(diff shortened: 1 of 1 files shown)


knelasevero:
As per discussion on https://redhat-internal.slack.com/archives/GK58XC2G2/p1713794549929979?thread_ts=1713772212.661319&cid=GK58XC2G2

There is no real reaso to have specific tolerations in this generated pod, so we can just tolerate all taints here as we do with the other ones

soltysh (approved):
/lgtm
/approve

PR #1719, opened 2024-04-15
Avoid using sets when existing slices suffice

skitt:
Now that Go includes slice operations such as slices.Contains, some uses of sets are no longer useful. In particular, creating a set using a slice just to check whether the slice contains a specific item is more expensive than checking the slice directly.

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -96,6 +96,7 @@ func (c syncContext) enqueueKeys(keys ...string) {
 // (or its tombstone) is a namespace  and it matches a name of any namespaces
 // that we are interested in
 func namespaceChecker(interestingNamespaces []string) func(obj interface{}) bool {
+	// This is used for quick lookups in informers
 	interestingNamespacesSet := sets.New(interestingNamespaces...)
 
 	return func(obj interface{}) bool {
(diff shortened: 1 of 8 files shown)


pkg/controller/factory/controller_context.go
```diff
@@ -96,19 +96,17 @@ func (c syncContext) enqueueKeys(keys ...string) {
...
-			return interestingNamespacesSet.Has(ns.Name)
+			return slices.Contains(interestingNamespaces, ns.Name)
 		}
 
 		// the object might be getting deleted
 		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
 		if ok {
 			if ns, ok := tombstone.Obj.(*corev1.Namespace); ok {
-				return interestingNamespacesSet.Has(ns.Name)
+				return slices.Contains(interestingNamespaces, ns.Name)
```
bertinatto, line 109: I'm in favor of using `slices.Contains` instead of constructs like:

```
sets.New(HostnameMatchSpecCandidates(hostname)...).Has(matchSpec)
```

But I'm not sure about cases like this one, as we're going from O(1) to O(n) and I'm not sure how often these are called.

skitt: I’ve restored two uses of sets where performance might matter.

PR #1720, opened 2024-04-17
NO-JIRA: certrotation: improve error messages

vrutkovs:
* Explicitly log "secret/configmap doesn't exist" events
* when CA bundle is updated log which signer update caused it

Tested in this [job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-gcp-modern/1805164315519488000) with shortened rotation periods.
[This Loki query](https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%7B%22datasource%22%3A%22PCEB727DF2F34084E%22%2C%22queries%22%3A%5B%7B%22expr%22%3A%22%7Binvoker%3D%5C%22openshift-internal-ci%2Frelease-openshift-origin-installer-launch-gcp-modern%2F1805164315519488000%5C%22%2C+type%3D%5C%22pod%5C%22%7D+%7C+json+%7C+namespace%3D%7E%5C%22openshift-%28etcd-operator%7Ckube%29.*%5C%22+%7C+line_format+%5C%22%7B%7B._entry%7D%7D%5C%22+%7C%7E+%5C%22%28SignerUpdateRequired%7CTargetUpdateRequired%7CCABundleUpdateRequired%29%5C%22+%7C+pattern+%5C%22%3C_%3E+reason%3A+%27%3Creason%3E%27+%5C%5C%5C%22%3Cobj_name%3E%5C%5C%5C%22+in+%5C%5C%5C%22%3Cobj_namespace%3E%5C%5C%5C%22+%3Cmessage%3E%5C%22+%7C++line_format+%5C%22%7B%7B.reason%7D%7D+%7B%7B.obj_namespace%7D%7D+%7B%7B.obj_name%7D%7D+%7B%7B.message%7D%7D%5C%22%22%2C%22refId%22%3A%22A%22%2C%22editorMode%22%3A%22code%22%2C%22queryType%22%3A%22range%22%7D%5D%2C%22range%22%3A%7B%22from%22%3A%221719144542906%22%2C%22to%22%3A%221719230942906%22%7D%7D) shows new events:
* `SignerUpdateRequired openshift-kube-apiserver-operator node-system-admin-signer requires a new signing cert/key pair: secret doesn't exist` instead of "missing notBefore annotation"
* `SignerUpdateRequired openshift-kube-apiserver-operator node-system-admin-signer requires a new signing cert/key pair: past refresh time (80% of validity): 2024-06-24 11:46:13.8 +0000 UTC`
* `CABundleUpdateRequired openshift-kube-controller-manager-operator csr-controller-signer-ca requires a new cert: signer update openshift-kube-controller-manager-operator/csr-signer`

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -41,7 +41,7 @@ type CABundleConfigMap struct {
 	EventRecorder events.Recorder
 }
 
-func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA) ([]*x509.Certificate, error) {
+func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA, signingCertKeyPairLocation string) ([]*x509.Certificate, error) {
 	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
 	// doesn't have any expired certs
 	modified := false
@@ -73,7 +73,15 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		return nil, err
 	}
 	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
-		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert", c.Name, c.Namespace)
+		reason := ""
+		if originalCABundleConfigMap == nil {
+			reason = "configmap doesn't exist"
+		} else if originalCABundleConfigMap.Data == nil {
+			reason = "configmap is empty"
+		} else if !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
+			reason = fmt.Sprintf("signer update %s", signingCertKeyPairLocation)
+		}
+		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert: %s", c.Name, c.Namespace, reason)
 		LabelAsManagedConfigMap(caBundleConfigMap, CertificateTypeCABundle)
 
 		modified = true
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -68,6 +68,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, false, err
 	}
+	var signerExists = true
 	signingCertKeyPairSecret := originalSigningCertKeyPairSecret.DeepCopy()
 	if apierrors.IsNotFound(err) {
 		// create an empty one
@@ -79,7 +80,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 			),
 			Type: corev1.SecretTypeTLS,
 		}
-		modified = true
+		signerExists = false
 	}
 
 	applyFn := resourceapply.ApplySecret
@@ -94,7 +95,10 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 	modified = needsMetadataUpdate || needsTypeChange || modified
 
 	signerUpdated := false
-	if needed, reason := needNewSigningCertKeyPair(signingCertKeyPairSecret.Annotations, c.Refresh, c.RefreshOnlyWhenExpired); needed {
+	if needed, reason := needNewSigningCertKeyPair(signingCertKeyPairSecret, c.Refresh, c.RefreshOnlyWhenExpired); needed || !signerExists {
+		if !signerExists {
+			reason = "secret doesn't exist"
+		}
 		c.EventRecorder.Eventf("SignerUpdateRequired", "%q in %q requires a new signing cert/key pair: %v", c.Name, c.Namespace, reason)
 		if err := setSigningCertKeyPairSecret(signingCertKeyPairSecret, c.Validity); err != nil {
 			return nil, false, err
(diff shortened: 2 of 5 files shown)


pkg/operator/certrotation/signer.go
```diff
@@ -67,6 +67,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, false, err
 	}
+	var signerExists = true
```
vrutkovs, line 70: This is required to pass unit tests, which do delete+create (and don't set ResourceVersion afterwards). It would be reworked after `ApplySecretDoNotUse` is removed

pkg/operator/certrotation/cabundle.go
```diff
@@ -75,7 +75,15 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		return nil, err
 	}
 	if originalCABundleConfigMap == nil || originalCABundleConfigMap.Data == nil || !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
-		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert", c.Name, c.Namespace)
+		reason := ""
+		if originalCABundleConfigMap == nil {
+			reason = "configmap doesn't exist"
+		} else if originalCABundleConfigMap.Data == nil {
+			reason = "configmap is empty"
```
tkashem, line 82: a minor suggestions:
- can we show the name and namespace of the object that does not exist for reason "configmap doesn't exist" and "configmap is empty"

pkg/operator/certrotation/cabundle.go
```diff
@@ -75,7 +75,15 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
...
+		} else if !equality.Semantic.DeepEqual(originalCABundleConfigMap.Data, caBundleConfigMap.Data) {
+			reason = fmt.Sprintf("signer update %s", signingCertKeyPairLocation)
+		}
+		c.EventRecorder.Eventf("CABundleUpdateRequired", "%q in %q requires a new cert: %s", c.Name, c.Namespace, reason)
```
tkashem, line 86: question: what does the c.Name and c.Namespace mean in relation to the object in the reason, or are we saying that `the  CA bundle configmap located in `c.Namespace/c.Name` requires a new cert`?

pkg/operator/certrotation/signer.go
```diff
@@ -153,7 +159,7 @@ func needNewSigningCertKeyPair(annotations map[string]string, refresh time.Durat
 	validity := notAfter.Sub(notBefore)
 	at80Percent := notAfter.Add(-validity / 5)
 	if time.Now().After(at80Percent) {
-		return true, fmt.Sprintf("past its latest possible time %v", at80Percent)
+		return true, fmt.Sprintf("past refresh time (80%% of validity): %v", at80Percent)
```
tkashem, line 162: nit: do you think it will help if we also include the validity in the message?

pkg/operator/certrotation/target.go
```diff
@@ -146,7 +146,12 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 	return targetCertKeyPairSecret, nil
 }
 
-func needNewTargetCertKeyPair(annotations map[string]string, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string {
+func needNewTargetCertKeyPair(secret *corev1.Secret, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool) string {
+	if secret.ResourceVersion == "" {
+		return "secret doesn't exist"
```
tkashem, line 151: Relying on the "ResourceVersion" being empty to indicate that the object does not exist may lead to bugs in the future, sometimes we set the resource version of an existing object to empty string before we send it to the apiserver for update
The more reliable way for us to know whether an object exists is when we get a  `NotFoundError` from the apiserver,  and use the not found error to produce this error message seems more intuitive to me, thoughts?

tkashem (commented):
I would recommend refactoring the code in a way that makes it easier to prepare the error messages, but that would be a long term goal.
This PR is a nice stop gap, it looks good to me except for a few minor comments

pkg/operator/certrotation/cabundle.go
vrutkovs: This error would be appended to the message with namespace/name: `openshift-kube-apiserver-operator node-system-admin-signer requires a new signing cert/key pair: secret doesn't exist` in `SignerUpdateRequired` event

pkg/operator/certrotation/signer.go
vrutkovs: Yes, this would show happy path, where certs are updated at 80%, as the cert can be force refreshed in the future

pkg/operator/certrotation/target.go
vrutkovs: Good idea, implemented

vrutkovs:
/cc @tjungblu 

This changes signatures for `EnsureConfigMapCABundle` and `NeedNewTargetCertKeyPair` used in cluster-etcd-operator

tjungblu:
ACK

pkg/operator/certrotation/cabundle.go
```diff
@@ -41,7 +41,7 @@ type CABundleConfigMap struct {
 	EventRecorder events.Recorder
 }
 
-func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA) ([]*x509.Certificate, error) {
+func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA, signingCertKeyPairLocation string) ([]*x509.Certificate, error) {
```
tjungblu, line 44: silly question while going through the breaking change in CEO, isn't that simply `ca.Namespace/ca.Name` at this point?

vrutkovs: That would be a location of the signer used to build this CA, so its `secret.Namespace/secret.Name`

tkashem:
/lgtm

Thanks!

soltysh (approved):
/approve

PR #1722, opened 2024-04-19
API-1802: cert-rotation: allow specifying multiple target certs in CertRotationController

vrutkovs:
Instead of defining several controllers managing the same signer/CA bundle pair and different target certs the same controller can accept a list of target certs to create.

Tested with
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=300","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806403521822593024)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=600","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806454959277871104)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=900","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455065116938240)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=1200","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455176609927168)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=1500","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455276568580096)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=1800","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455378926374912)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=2100","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455474938187776)
* [`workflow-test openshift-e2e-cert-rotation-suspend-sno 4.17,https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669,https://github.com/openshift/cluster-etcd-operator/pull/1284 "CLUSTER_AGE_DAYS=2400","CLUSTER_AGE_STEP=300","PACKET_OS=rocky_9"`](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-assisted-modern/1806455571197464576)

### DIFF
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -65,9 +67,8 @@ type CertRotationController struct {
 	RotatedSigningCASecret RotatedSigningCASecret
 	// CABundleConfigMap maintains a CA bundle config map, by adding new CA certs coming from rotatedSigningCASecret, and by removing expired old ones.
 	CABundleConfigMap CABundleConfigMap
-	// RotatedSelfSignedCertKeySecret rotates a key and cert signed by a signing CA and stores it in a secret.
-	RotatedSelfSignedCertKeySecret RotatedSelfSignedCertKeySecret
-
+	// RotatedTargetSecrets contains a list of key and cert signed by a signing CA to rotate.
+	RotatedTargetSecrets []RotatedSelfSignedCertKeySecret
 	// Plumbing:
 	StatusReporter StatusReporter
 }
@@ -81,11 +82,11 @@ func NewCertRotationController(
 	reporter StatusReporter,
 ) factory.Controller {
 	c := &CertRotationController{
-		Name:                           name,
-		RotatedSigningCASecret:         rotatedSigningCASecret,
-		CABundleConfigMap:              caBundleConfigMap,
-		RotatedSelfSignedCertKeySecret: rotatedSelfSignedCertKeySecret,
-		StatusReporter:                 reporter,
+		Name:                   name,
+		RotatedSigningCASecret: rotatedSigningCASecret,
+		CABundleConfigMap:      caBundleConfigMap,
+		RotatedTargetSecrets:   []RotatedSelfSignedCertKeySecret{rotatedSelfSignedCertKeySecret},
+		StatusReporter:         reporter,
 	}
 	return factory.New().
 		ResyncEvery(time.Minute).
@@ -102,6 +103,42 @@ func NewCertRotationController(
 			"CertRotationController", // don't change what is passed here unless you also remove the old FooDegraded condition
 			recorder.WithComponentSuffix("cert-rotation-controller").WithComponentSuffix(name),
 		)
+}
+
+func NewCertRotationControllerMultipleTargets(
+	name string,
+	rotatedSigningCASecret RotatedSigningCASecret,
+	caBundleConfigMap CABundleConfigMap,
+	rotatedTargetSecrets []RotatedSelfSignedCertKeySecret,
+	recorder events.Recorder,
+	reporter StatusReporter,
+) factory.Controller {
+	informers := sets.New[factory.Informer](
+		rotatedSigningCASecret.Informer.Informer(),
+		caBundleConfigMap.Informer.Informer(),
+	)
+
+	for _, target := range rotatedTargetSecrets {
+		informers = informers.Insert(target.Informer.Informer())
+	}
+
+	c := &CertRotationController{
+		Name:                   name,
+		RotatedSigningCASecret: rotatedSigningCASecret,
+		CABundleConfigMap:      caBundleConfigMap,
+		RotatedTargetSecrets:   rotatedTargetSecrets,
+		StatusReporter:         reporter,
+	}
+	return factory.New().
+		ResyncEvery(time.Minute).
+		WithSync(c.Sync).
+		WithInformers(
+			informers.UnsortedList()...,
+		).
+		WithPostStartHooks(
+			c.targetCertRecheckerPostRunHook,
+		).
+		ToController("MultipleTargetCertRotationController", recorder.WithComponentSuffix("cert-rotation-controller").WithComponentSuffix(name))
 }
 
 func (c CertRotationController) Sync(ctx context.Context, syncCtx factory.SyncContext) error {
@@ -125,7 +162,15 @@ func (c CertRotationController) Sync(ctx context.Context, syncCtx factory.SyncCo
 }
 
 func (c CertRotationController) getSigningCertKeyPairLocation() string {
-	return fmt.Sprintf("%s/%s", c.RotatedSelfSignedCertKeySecret.Namespace, c.RotatedSelfSignedCertKeySecret.Name)
+	var firstSecret RotatedSelfSignedCertKeySecret
+	if c.RotatedTargetSecrets == nil {
+		return ""
+	}
+	if len(c.RotatedTargetSecrets) == 0 {
+		return ""
+	}
+	firstSecret = c.RotatedTargetSecrets[0]
+	return fmt.Sprintf("%s/%s", firstSecret.Namespace, firstSecret.Name)
 }
 
 func (c CertRotationController) SyncWorker(ctx context.Context) error {
@@ -139,8 +184,14 @@ func (c CertRotationController) SyncWorker(ctx context.Context) error {
 		return err
 	}
 
-	if _, err := c.RotatedSelfSignedCertKeySecret.EnsureTargetCertKeyPair(ctx, signingCertKeyPair, cabundleCerts); err != nil {
-		return err
+	var errs []error
+	for _, secret := range c.RotatedTargetSecrets {
+		if _, err := secret.EnsureTargetCertKeyPair(ctx, signingCertKeyPair, cabundleCerts); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) != 0 {
+		return errors.NewAggregate(errs)
 	}
 
 	return nil
@@ -148,21 +199,20 @@ func (c CertRotationController) SyncWorker(ctx context.Context) error {
 
 func (c CertRotationController) targetCertRecheckerPostRunHook(ctx context.Context, syncCtx factory.SyncContext) error {
 	// If we have a need to force rechecking the cert, use this channel to do it.
-	refresher, ok := c.RotatedSelfSignedCertKeySecret.CertCreator.(TargetCertRechecker)
-	if !ok {
-		return nil
-	}
-	targetRefresh := refresher.RecheckChannel()
-	go wait.Until(func() {
-		for {
-			select {
-			case <-targetRefresh:
-				syncCtx.Queue().Add(factory.DefaultQueueKey)
-			case <-ctx.Done():
-				return
-			}
+	for _, target := range c.RotatedTargetSecrets {
+		if refresher, ok := target.CertCreator.(TargetCertRechecker); ok {
+			go wait.Until(func() {
+				for {
+					select {
+					case <-refresher.RecheckChannel():
+						syncCtx.Queue().Add(factory.DefaultQueueKey)
+					case <-ctx.Done():
+						return
+					}
+				}
+			}, time.Minute, ctx.Done())
 		}
-	}, time.Minute, ctx.Done())
+	}
 
 	<-ctx.Done()
 	return nil
--- a/pkg/operator/certrotation/client_cert_rotation_controller.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -80,6 +81,8 @@ func NewCertRotationController(
 	rotatedSelfSignedCertKeySecret RotatedSelfSignedCertKeySecret,
 	recorder events.Recorder,
 	reporter StatusReporter,
+	controlledSecrets *[]metav1.ObjectMeta,
+	controlledConfigMaps *[]metav1.ObjectMeta,
 ) factory.Controller {
 	c := &CertRotationController{
 		Name:                   name,
@@ -88,6 +91,19 @@ func NewCertRotationController(
 		RotatedTargetSecrets:   []RotatedSelfSignedCertKeySecret{rotatedSelfSignedCertKeySecret},
 		StatusReporter:         reporter,
 	}
+	*controlledSecrets = append(*controlledSecrets, metav1.ObjectMeta{
+		Name:      rotatedSigningCASecret.Name,
+		Namespace: rotatedSigningCASecret.Namespace,
+	})
+	*controlledConfigMaps = append(*controlledConfigMaps, metav1.ObjectMeta{
+		Name:      caBundleConfigMap.Name,
+		Namespace: caBundleConfigMap.Namespace,
+	})
+	*controlledSecrets = append(*controlledSecrets, metav1.ObjectMeta{
+		Name:      rotatedSelfSignedCertKeySecret.Name,
+		Namespace: rotatedSelfSignedCertKeySecret.Namespace,
+	})
+
 	return factory.New().
 		ResyncEvery(time.Minute).
 		WithSync(c.Sync).
@@ -112,14 +128,28 @@ func NewCertRotationControllerMultipleTargets(
 	rotatedTargetSecrets []RotatedSelfSignedCertKeySecret,
 	recorder events.Recorder,
 	reporter StatusReporter,
+	controlledSecrets *[]metav1.ObjectMeta,
+	controlledConfigMaps *[]metav1.ObjectMeta,
 ) factory.Controller {
 	informers := sets.New[factory.Informer](
 		rotatedSigningCASecret.Informer.Informer(),
 		caBundleConfigMap.Informer.Informer(),
 	)
 
+	*controlledSecrets = append(*controlledSecrets, metav1.ObjectMeta{
+		Name:      rotatedSigningCASecret.Name,
+		Namespace: rotatedSigningCASecret.Namespace,
+	})
+	*controlledConfigMaps = append(*controlledConfigMaps, metav1.ObjectMeta{
+		Name:      caBundleConfigMap.Name,
+		Namespace: caBundleConfigMap.Namespace,
+	})
 	for _, target := range rotatedTargetSecrets {
 		informers = informers.Insert(target.Informer.Informer())
+		*controlledSecrets = append(*controlledSecrets, metav1.ObjectMeta{
+			Name:      target.Name,
+			Namespace: target.Namespace,
+		})
 	}
 
 	c := &CertRotationController{
--- a/pkg/operator/certrotation/client_cert_rotation_controller_test.go
+++ b/pkg/operator/certrotation/client_cert_rotation_controller_test.go
@@ -1,0 +1,419 @@
+package certrotation
+
+import (
+	"context"
+	"crypto/x509"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/events/eventstesting"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/informers"
+	kubefake "k8s.io/client-go/kubernetes/fake"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	clienttesting "k8s.io/client-go/testing"
+	"k8s.io/client-go/tools/cache"
+)
+
+type fakeStatusReporter struct{}
+
+func (f *fakeStatusReporter) Report(ctx context.Context, controllerName string, syncErr error) (updated bool, updateErr error) {
+	return false, nil
+}
+
+type fakeTargetCertCreator struct {
+	name             string
+	hostnamesChanged chan struct{}
+	channelClosed    bool
+	callCount        int
+}
+
+func (f *fakeTargetCertCreator) NewCertificate(signer *crypto.CA, validity time.Duration) (*crypto.TLSCertificateConfig, error) {
+	return nil, nil
+}
+
+func (f *fakeTargetCertCreator) NeedNewTargetCertKeyPair(currentCertSecret *corev1.Secret, signer *crypto.CA, caBundleCerts []*x509.Certificate, refresh time.Duration, refreshOnlyWhenExpired bool, _ bool) string {
+	f.callCount += 1
+	return ""
+}
+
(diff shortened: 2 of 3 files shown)


dinhxuanvu (commented):
/lgtm

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -132,24 +169,40 @@ func (c CertRotationController) SyncWorker(ctx context.Context) error {
...
+		if refresher, ok := target.CertCreator.(TargetCertRechecker); ok {
+			targetRefreshers = append(targetRefreshers, refresher.RecheckChannel())
+		}
 	}
-	targetRefresh := refresher.RecheckChannel()
+	aggregateTargetRefresher := make(chan struct{})
+	for _, ch := range targetRefreshers {
+		go func(c <-chan struct{}) {
+			for msg := range c {
+				aggregateTargetRefresher <- msg
```
tkashem, line 197: goroutine leaking: `<-ctx.Done()`, could be a problem for integration tests that check goroutine leakages?

tkashem (commented):
a) `RotatedSigningCASecret` creates the signer CA secret
b) `CABundleConfigMap` creates a configmap, and
c) `RotatedSelfSignedCertKeySecret` creates a secret

I like the idea of a controller doing one thing, can we explore the idea of the individual controllers?
a) `SignerCAController`: this controller manager the signer secret object. 
b) `CABundleController`: it watched the secret object from `a` and creates a single configmap and manages it.
c) `CertKeySecretController`: this can watch objects from `a` and `b` and creates a secret with cert/key and manages it.

With this, we can have N instances of `CertKeySecretController`, where each instance  derives it cert/key from a single instance of `a` and `b`

vrutkovs:
>can we explore the idea of the individual controllers?

That's possible, but the goal is to create target certs, signers and CA is a prerequisite to it. We could have separate signer/ca controllers, but we might end up with signer certs not producing any target certs or CA bundles without any new signers etc.

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -101,6 +102,42 @@ func NewCertRotationController(
...
+	return factory.New().
+		ResyncEvery(time.Minute).
+		WithSync(c.Sync).
+		WithInformers(
+			informers.UnsortedList()...,
+		).
+		WithPostStartHooks(
+			c.targetCertRecheckerPostRunHook,
+		).
+		ToController("CertRotationController", recorder.WithComponentSuffix("cert-rotation-controller").WithComponentSuffix(name))
```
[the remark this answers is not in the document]
vrutkovs: Good idea

pkg/operator/certrotation/client_cert_rotation_controller.go
vrutkovs: `ctx.Done` would close them iiuc, but yeah, worth adding a unit test which uses runhook and ensures no goroutines are leaking

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -132,24 +169,40 @@ func (c CertRotationController) SyncWorker(ctx context.Context) error {
...
-	targetRefresh := refresher.RecheckChannel()
+	aggregateTargetRefresher := make(chan struct{})
+	for _, ch := range targetRefreshers {
+		go func(c <-chan struct{}) {
+			for msg := range c {
+				aggregateTargetRefresher <- msg
+			}
+		}(ch)
+	}
+
```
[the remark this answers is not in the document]
vrutkovs: Reworked this to make it look like b). Added a test which verifies goroutines don't leak

I don't quite understand what c) is meant for - make controller accept a single channel reused across all `CertCreators`? Not sure what's the benefit of that

pkg/operator/certrotation/client_cert_rotation_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+	time.AfterFunc(1*time.Second, func() {
+		cancel()
+	})
+	firstTargetCertCreator.triggerSync()
+	secondTargetCertCreator.triggerSync()
+	c.Run(controllerCtx, 1)
+
+	// Ensure both target certs have been called exactly three times
+	// initial sync and two hook calls for target certs
+	// TODO[vrutkovs]: informers make unpredictable number of calls
```
vrutkovs, line 406: Not sure how to tackle that - or how to make sure two hook syncs were included in regular informer sync. `informerFactory` and `NewCertRotationControllerMultipleTargets` promise to sync every minute, but it happens much more often

p0lyn0mial:
An alternative design would be to have a controller per certificate – this would actually preserve the current behaviour. I prefer having a single controller per certificate as it is easier to debug, report status, retry on error, and reason about. Thoughts?

Our issue is that both the `RotatedSigningCASecret` and `CABundleConfigMap` are reconciled by multiple controllers (`NewCertRotationController`) without any coordination.

I think the least invasive change would be to make both `RotatedSigningCASecret` and `CABundleConfigMap` thread-safe. Is there an easy way to achieve this?

Another idea that comes to mind (I think Abu suggested the same) is to turn both `RotatedSigningCASecret` and `CABundleConfigMap` into controllers and slightly modify `NewCertRotationController` to read secrets instead. Thoughts ?

vrutkovs:
>An alternative design would be to have a controller per certificate – this would actually preserve the current behaviour. I prefer having a single controller per certificate as it is easier to debug, report status, retry on error, and reason about. Thoughts?

Similar to Abu's idea in https://github.com/openshift/library-go/pull/1722#pullrequestreview-2139025726? That would prevent races, but would make sequence of events to do a proper rollout complicated (three different controllers would need to be properly synced so that CA bundle would be updated before target cert etc.). Also may lead to "orphan" controllers managing CA bundle without a target cert.

>Our issue is that both the RotatedSigningCASecret and CABundleConfigMap are reconciled by multiple controllers (NewCertRotationController) without any coordination.

Yes, this issue still potentially remains. The PR focuses on solving a much more widespread issue of multiple target certs

>I think the least invasive change would be to make both RotatedSigningCASecret and CABundleConfigMap thread-safe. Is there an easy way to achieve this?

I don't know if its feasible, as its not just thread-safety but also process-safety we're concerned about

p0lyn0mial:
> would make sequence of events to do a proper rollout complicated (three different controllers would need to be properly synced so that CA bundle would be updated before target cert etc.)

Don't we do it all the time? 

The aggregator controller waits until a service is created before it wires an HTTP handler. 
The degraded webhook controller waits until a webhook is created before it can validate it. 

What I like about these and other controllers is that when you look inside, you will see that these controllers are reconciling a single resource. They are simply reading their prerequisites and reacting to any changes before reconciling.


In our case, it would boil down to three separate controllers that reconcile their resources, where the second and the third controller read the crypto material from the lister before reconciling. 

For example: NewCertRotationController would:
```
Read signerCA from the lister 
Parse the signerCA 
Stop on any error 

Read the caBundle from the lister 
Parse the caBundle 
Stop on any error 

Reconcile clientCertificate with signerCA and caBundle
```

When there are issues with the signerCA you go to `RotatedSigningCASecret` to debug it since it is responsible for generating the resource. Thoughts ?

vrutkovs:
>Don't we do it all the time?

Yes - this is what this controller is doing in the end. So adding three more controllers for each resource won't magically solve anything - we also need a way to group them to ensure CA bundle and target certs updated after signer reissue. We also need a way of regenerating certificates on request when hostname change. In the end existing controller won't go away, but we'll add three more controller types and complicate syncing and ordering of execution.

This won't qualify as a simple bugfix two weeks before feature freeze - this is a full blown rework tbf

p0lyn0mial:
> So adding three more controllers for each resource won't magically solve anything.

Having a single controller for reconciling a resource would solve the race condition we are facing.

>It also needs

> - get notified that signer was updated
> - hostnames have changed

> otherwise it would rely on "lets wait a minute until next sync to catch", which means signer is already updated by CA is not - so its unclear when other components are allowed to use the updated signer.

yes, usually controllers react to changes made to other resources and we should do the same.


> This won't qualify as a simple bugfix two weeks before feature freeze - this is a full blown rework tbf

yes, I agree it would require more code. The race we are facing is not new, it has been introduced a long time ago and it wasn't obvious. I think that having a controller per resource leads to simpler program in the end.

vrutkovs:
Discussed on the meeting - this looks good enough for 4.17, later need to be reworked into more granular controllers, `TargetCertRechecker` interface should be reworked to avoid using channels

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -132,30 +169,35 @@ func (c CertRotationController) SyncWorker(ctx context.Context) error {
...
-				syncCtx.Queue().Add(factory.DefaultQueueKey)
-			case <-ctx.Done():
-				return
-			}
+	for _, target := range c.RotatedTargetSecrets {
+		if refresher, ok := target.CertCreator.(TargetCertRechecker); ok {
+			go wait.Until(func() {
+				for {
+					select {
+					case <-refresher.RecheckChannel():
```
tkashem, line 192: so if we have `N` targets, that means `N` goroutines, each of which is waiting to receive from  a channel to be notified of an event.
The producer ( the entity that writes to a channel) will probably write once when the host name changes. it is also possible that a channel is shared among the `N` instances depending on how we instantiate the controller.

Let's say we have three targets, `{a, b, and c}`; so goroutine for `c` when it receives from its channel it will add a key to the queue, which will trigger the sync method, and the sync method applies to all targets. It's not a huge issue, but it seems to be an unnecessary complexity we can avoid?

a couple of questions with this mechanism:
- what should happen when the channel is closed, should we treat this an a `event occurred`? probably not, maybe we should check with `case _, ok := <-refresher.RecheckChannel()`?
- the controller syncs every `1m`, do we really need this channel based mechanism to trigger sync, are we losing anything other than 'few seconds of promptness' if we don't have this channel based trigger? 

If we really need this mechanism, i would suggest abstracting this mechanism:
```
// this abstracts the work queue used by the controller
type queueAdder interface {
    Add()
}

type channelBasedTrigger struct {
    ch chan time.Time
    queue queueAdder // or we can even use 'f func() {}'
}

func (t channelBasedTrigger) Event() { t.ch <- time.Now() } // used by the producer

func (t channelBasedTrigger) Run(ctx context.Context) {
       go wait.Until(func() {
				for {
					select {
					case t, ok := <-t.ch:
						if ok {
                                                      t.queue.Add() // or f()
                                                 }
					case <-ctx.Done():
						return
					}
				}
			}, time.Minute, ctx.Done())
}
``` 

This way we can keep the logic of channel based trigger mechanism co-located, more testable, and more generic too. Also, we can use one instance of `channelBasedTrigger` to inform `N` controllers.
There might be some constructs in the apimachinery/util/wait package that will allow us to achieve this.

vrutkovs: >what should happen when the channel is closed, should we treat this an a event occurred? 

iiuc both struct implementing `TargetCertRechecker` interface and the controllers are created in `NewCertRotationController`, so the channel is never closed when the controller is still running.

>If we really need this mechanism, i would suggest abstracting this mechanism:

So far we use it in just one `ServingRotation` struct, so probably there is an easier way out

p0lyn0mial:
@vrutkovs Do we know exactly where the race condition occurs? 
I know we run multiple controllers for the same types, but I would like to understand how the race condition manifests. 

I’ve opened https://github.com/openshift/library-go/pull/1753 as an alternative to this PR. 
To be honest, I still think the best approach would be to split it into multiple controllers.

vrutkovs:
>Do we know exactly where the race condition occurs?

Usually its various CA bundles being updated incorrectly, especially those which are used in several targets.

>I’ve opened https://github.com/openshift/library-go/pull/1753 as an alternative to this PR.

I don't think its alternative really, we can have both. 1753 alone won't solve it (you still need a process-safe way for it)

p0lyn0mial:
> Usually its various CA bundles being updated incorrectly, especially those which are used in several targets.

@vrutkovs could you be more specific ? Do we know what exactly is broken ?

I have checked [the code for updating](https://github.com/openshift/library-go/blob/842f91933ca5744f1f2f15825f178745e2ab1d56/pkg/operator/certrotation/cabundle.go#L116) the CA bundles and it looks good.

vrutkovs:
>Do we know what exactly is broken?

See [this job](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.16-e2e-metal-ovn-sno-cert-rotation-shutdown-2y-age-90d) history - it intermittently passes or fails. [This job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.16-e2e-metal-ovn-sno-cert-rotation-shutdown-2y-age-90d/1804666431120347136) is a good example. Here kube-apiserver won't start as `aggregator-client-ca` was updated in an unsafe manner.
After kube-control-plane-signer was rotated, the CA bundle was expected to have both previous and new values for these signers, but instead it has new signer added twice (note the notBefore dates):
```
tlsconfig.go:178] "Loaded client CA" index=5 certName="client-ca-bundle::/etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt,request-header::/etc/kubernetes/static-pod-certs/configmaps/aggregator-client-ca/ca-bundle.crt" certDetail="\"openshift-kube-apiserver-operator_kube-control-plane-signer@1770952982\" [] issuer=\"<self>\" (2026-02-13 03:23:01 +0000 UTC to 2026-04-14 03:23:02 +0000 UTC (now=2026-02-13 03:43:34.322043373 +0000 UTC))"
tlsconfig.go:178] "Loaded client CA" index=6 certName="client-ca-bundle::/etc/kubernetes/static-pod-certs/configmaps/client-ca/ca-bundle.crt,request-header::/etc/kubernetes/static-pod-certs/configmaps/aggregator-client-ca/ca-bundle.crt" certDetail="\"openshift-kube-apiserver-operator_kube-control-plane-signer@1770952982\" [] issuer=\"<self>\" (2026-02-13 03:23:01 +0000 UTC to 2026-04-14 03:23:02 +0000 UTC (now=2026-02-13 03:43:34.322057069 +0000 UTC))"
```

So `kube-apiserver-check-endpoints` sidecar can't start:
```
F0213 03:54:32.947195       1 cmd.go:170] error initializing delegating authentication: unable to load configmap based request-header-client-ca-file: Unauthorized
```
and kube-apiserver log is full of
```
2026-02-13T03:43:41.311586909Z E0213 03:43:41.311514      16 authentication.go:73] "Unable to authenticate the request" err="[x509: certificate signed by unknown authority, verifying certificate SN=5335778093844856917, SKID=, AKID=54:83:47:1C:75:87:ED:E4:01:8C:25:18:0D:D8:CF:05:5E:2C:C4:FE failed: x509: certificate signed by unknown authority (possibly because of \"crypto/rsa: verification error\" while trying to verify candidate authority certificate \"openshift-kube-apiserver-operator_kube-control-plane-signer@1770952982\")]"
```

>I have checked [the code for updating](https://github.com/openshift/library-go/blob/842f91933ca5744f1f2f15825f178745e2ab1d56/pkg/operator/certrotation/cabundle.go#L116) the CA bundles and it looks good.

The code is correct, but when its not thread-safe and also doesn't protect from several processes updating the same CA configmap. This PR would eliminate process races, #1753 would ensure `CABundleConfigMap` update is thread safe

p0lyn0mial:
> After kube-control-plane-signer was rotated, the CA bundle was expected to have both previous and new values for these signers, but instead it has new signer added twice (note the notBefore dates):

that is the part I would like to understand deeper, [the updating code](https://github.com/openshift/library-go/blob/842f91933ca5744f1f2f15825f178745e2ab1d56/pkg/operator/certrotation/cabundle.go#L116) seems to be removing duplicates and expired certificates.

p0lyn0mial:
> This PR would eliminate process races, https://github.com/openshift/library-go/pull/1753 would ensure CABundleConfigMap update is thread safe

Both PR seem to be removing the race within a single process. I still thing the best way forward would be to create separate controllers.

vrutkovs:
>this is the part I would like to understand deeper, [the updating code](https://github.com/openshift/library-go/blob/842f91933ca5744f1f2f15825f178745e2ab1d56/pkg/operator/certrotation/cabundle.go#L116) seems to be removing duplicates and expired certificates.

Correct, this code removes expired signers - but previous signer has not expired yet. Its being refreshed at 80% lifetime, so it still has 20% of a lifetime to be active (its 2.5 month of a year validity).

> I still thing the best way forward would be to create separate controllers.

Sure, this PR is just interim fix for us to make it until 4.17 feature freeze. Once we establish one way of process and thread safety - and have e2e tests passing - we can experiment with more substantial code rework

p0lyn0mial:
> I don't quite get why you need to multi-thread those with goroutines or even need multiple control loops

It is not about multi-threading for performance reasons. It is about having a single control loop per resource. I think this is already a well-established pattern upstream. I think that having a single controller that manages a resource is easy to understand and debug.

> Sure, this PR is just interim fix for us to make it until 4.17 feature freeze. Once we establish one way of process and thread safety - and have e2e tests passing - we can experiment with more substantial code rework

This is a crucial piece of code, and I wouldn't rush it. Besides, we still need to fix kubelet, client-go, and tons of other things before the platform will be able to recover itself from expired certificates. Thus, I don't see the point in developing temporary solutions. We are not dealing with an escalation that requires an immediate fix. I would rather implement a proper fix or not fix it at all.

tjungblu:
> This is a crucial piece of code, and I wouldn't rush it. 

so you propose to rewrite the entire codebase to fit some upstream pattern? :) sounds great :+1:

p0lyn0mial:
> so you propose to rewrite the entire codebase to fit some upstream pattern? :) sounds great 👍

I'm proposing wrapping `CABundleConfigMap` and `RotatedSigningCASecret` into separate controllers (the controllers would simply call the `Ensure...` methods) and writing a new controller for `CertRotation`.

The new `CertRotation` wouldn't differ much from the existing one. The only difference would be reading the signer and the CA from the cache/lister. We could also consider removing the side chan for the hostname for simplicity - just as you did for your controller.

Then, when we need to compose these controllers we would only have a singe instance of `CABundleConfigMap` and `RotatedSigningCASecret` and many instances of `CertRotation`.

Does it make sense to you as well ?

vrutkovs:
>I'm proposing wrapping

This is what we agreed to few weeks back and noone is debating that choice. The immediate question is why all of this is being discussed in an unrelated PR with a temporary fix for 4.17 - and why is it being stalled for several weeks already

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -65,9 +67,8 @@ type CertRotationController struct {
 	RotatedSigningCASecret RotatedSigningCASecret
 	// CABundleConfigMap maintains a CA bundle config map, by adding new CA certs coming from rotatedSigningCASecret, and by removing expired old ones.
 	CABundleConfigMap CABundleConfigMap
-	// RotatedSelfSignedCertKeySecret rotates a key and cert signed by a signing CA and stores it in a secret.
-	RotatedSelfSignedCertKeySecret RotatedSelfSignedCertKeySecret
-
+	// RotatedTargetSecrets contains a list of key and cert signed by a signing CA to rotate.
+	RotatedTargetSecrets []RotatedSelfSignedCertKeySecret
```
dusk125, line 71: This will change the external API and clients would need to update if they're accessing this field directly.

Since you're already adding a new "Create" function for the multi rotation controller, would it make more sense to have two cert controllers? One for single rotation and one for multi.

vrutkovs: >This will change the external API and clients would need to update if they're accessing this field directly

Correct, this would require users. So far we identified that only cluster-kube-apiserver-operator and cluster-kube-control-manager-operator use multitarget controllers (and they don't need to access `RotatedTargetSecrets` directly)

>would it make more sense to have two cert controllers? One for single rotation and one for multi.

It may be beneficial later to move common parts into functions and reuse them, however at this point only two different `New...` functions are used

deads2k:
Is the goal of this change to avoid races between multiple updating controllers?  If so, this solution appears to miss the critical change required to resolve the cross-binary active-active problem.  The core mistake is using the `.Apply` functions that smash on conflicts.  Changing this code back to the original `.Update` will cause conflicts and force retries.  The retry would fail until the cache updates and only one operator will win the update.

vrutkovs:
>Is the goal of this change to avoid races between multiple updating controllers?

This prevents races within one controller (thread-level). Component PRs like [this](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669) ensure that multiple controllers don't run simultaneously.

> The core mistake is using the .Apply functions that smash on conflicts

We don't use `Apply` in `applyConfigMap/applySecret` - counterintuitively these use `Update`: https://github.com/openshift/library-go/blob/e0aa70d556789b3ff5e9af85e838320b3063aa14/pkg/operator/resource/resourceapply/core.go#L351

deads2k:
> We don't use `Apply` in `applyConfigMap/applySecret` - counterintuitively these use `Update`:

By having it read the live secret and use its RV instead of the RV the controller used to make decisions, conflicts are avoided and produce a situation that looks like

1. client/1 reads rv/6 to make decision about update
2. client/2 reads rv/6 to make decision about update
3. client/1 calls applySecret first, 
    1. apply secret reads secret rv/6
    2. apply secret writes secret rv/7
4. client/2 calls applySecret second
    1. apply secret reads secret rv/7 *** this is not good.
    2. apply secret writes secret rv/8

Using .Update instead should eliminate the read in 4.1 and result in a conflict instead of the race.

vrutkovs:
We already use .Update there - also discussed it with Lukasz in https://github.com/openshift/library-go/pull/1763#discussion_r1686625674

Also, this PR would prevent client/1 and client/2 from ever happening. If `applySecret` needs to be updated to handle races it should be done in a separate PR (as the change would affect a lot more components)

p0lyn0mial:
> Am I missing something? We already use .Update there - also discussed it with Lukasz in https://github.com/openshift/library-go/pull/1763#discussion_r1686625674 and he wants me to change this to use .Update too.

@vrutkovs let's set up a pair programming session tomorrow/this-week where I can show you how we could use optimistic concurrency control built into kube to solve the race.

vrutkovs:
Ah, the problem here was:
* we do a first `Get` and modify the secret
* `applySecret` is called
* `applySecret` does another Get and then does Update

Without second `Get` we get conflicts as expected - tested this in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1719. So perhaps we should have a setting for applySecret to skip the first `Get` to avoid this?

However, its orthogonal to this PR - we can both remove racing controllers and make sure `applySecret` can be thread-safe.

PR #1723, opened 2024-04-23
rewrite bootstrap IPs

deads2k:


pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -199,6 +199,41 @@ func RewriteNodeIPs(nodeList []*corev1.Node) *metadataOptions {
 	}
 }
 
+func RewriteBootstrapIP(bootstrapIP string) *metadataOptions {
+	return rewriteString(bootstrapIP, "<bootstrap-ip>")
+}
+
+// TODO decide if this is a good idea to expose when we actually need it.  I bet we're better off naming them explicitly.
```
vrutkovs, line 206: etcd-operator will GC this secret eventually. Currently it runs every odd hour, so it won't affect most e2e jobs. However, if we always expect this cert to be present in the cluster it may not be the case.

deads2k: > etcd-operator will GC this secret eventually. Currently it runs every odd hour, so it won't affect most e2e jobs. However, if we always expect this cert to be present in the cluster it may not be the case.

The alternative Skip method in https://github.com/openshift/library-go/pull/1717 is vulnerable in exactly the same way, right?

deads2k: Or are you saying, "don't record because some CI runs don't contain the secret and thus end up looking like certificate information is missing?

vrutkovs: yeah, we shouldn't record it in the TLS artifacts as it may not exist and just generally useless. However, TLS registry may contain items which may not be present in the clusters (i.e. TechPreview featureset) and test won't fail.

If both of these solutions are equivalent its probably best to keep it for completeness

deads2k: > yeah, we shouldn't record it in the TLS artifacts as it may not exist and just generally useless. However, TLS registry may contain items which may not be present in the clusters (i.e. TechPreview featureset) and test won't fail.
> 
> If both of these solutions are equivalent its probably best to keep it for completeness

I was about to bring up the same.  I think having a record will improve our tracking ability.  If it's non-viable today I'm willing to merge the skip, but I think we restrict in the other direction (no unknown certs), but we can record extra ones.

vrutkovs:
/close 

Lets use https://github.com/openshift/library-go/pull/1725 instead

PR #1724, opened 2024-04-24
API-1802: certrotation: issue one Update request in cert rotation functions

vrutkovs:
Each function used in cert rotation controller should issue just one ApplySecret/ApplySecret per call. This call may  not be atomic, so running several replicas of this controller can cause unexpected races

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -44,6 +44,8 @@ type CABundleConfigMap struct {
 func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA) ([]*x509.Certificate, error) {
 	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
 	// doesn't have any expired certs
+	modified := false
+
 	originalCABundleConfigMap, err := c.Lister.ConfigMaps(c.Namespace).Get(c.Name)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
@@ -56,19 +58,15 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 			c.Namespace,
 			c.AdditionalAnnotations,
 		)}
+		modified = true
 	}
 
-	needsMetadataUpdate := false
+	needsOwnerUpdate := false
 	if c.Owner != nil {
-		needsMetadataUpdate = ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
+		needsOwnerUpdate = ensureOwnerReference(&caBundleConfigMap.ObjectMeta, c.Owner)
 	}
-	needsMetadataUpdate = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta) || needsMetadataUpdate
-	if needsMetadataUpdate && len(caBundleConfigMap.ResourceVersion) > 0 {
-		_, _, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
-		if err != nil {
-			return nil, err
-		}
-	}
+	needsMetadataUpdate := c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&caBundleConfigMap.ObjectMeta)
+	modified = needsOwnerUpdate || needsMetadataUpdate || modified
 
 	updatedCerts, err := manageCABundleConfigMap(caBundleConfigMap, signingCertKeyPair.Config.Certs[0])
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/operator/certrotation/signer.go
```diff
@@ -87,13 +89,9 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
...
-	if ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(signingCertKeyPairSecret) {
-		actualSigningCertKeyPairSecret, _, err := applyFn(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
-		if err != nil {
-			return nil, false, err
-		}
-		signingCertKeyPairSecret = actualSigningCertKeyPairSecret
-	}
+	needsMetadataUpdate := ensureMetadataUpdate(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
+	needsTypeChange := ensureSecretTLSTypeSet(signingCertKeyPairSecret)
+	modified = needsMetadataUpdate || needsTypeChange || modified
```
bertinatto, line 94: Previously, the code called `applyFn` based on the result of an `&&` condition. Now it uses `||`. Can you confirm if this change is intentional?

The new code looks correct to me, but I just wanted to confirm.

pkg/operator/certrotation/target.go
```diff
@@ -120,13 +123,9 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
...
-	if ensureMetadataUpdate(targetCertKeyPairSecret, c.Owner, c.AdditionalAnnotations) && ensureSecretTLSTypeSet(targetCertKeyPairSecret) {
-		actualTargetCertKeyPairSecret, _, err := applyFn(ctx, c.Client, c.EventRecorder, targetCertKeyPairSecret)
-		if err != nil {
-			return nil, err
-		}
-		targetCertKeyPairSecret = actualTargetCertKeyPairSecret
-	}
+	needsMetadataUpdate := ensureMetadataUpdate(targetCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
+	needsSecretTypeUpdate := ensureSecretTLSTypeSet(targetCertKeyPairSecret)
+	modified = needsMetadataUpdate || needsSecretTypeUpdate || modified
```
bertinatto, line 128: Same question WRT `||` and `&&`.

pkg/operator/certrotation/signer.go
```diff
@@ -104,13 +102,18 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
...
+		signerUpdated = true
+	}
+
+	if modified {
 		actualSigningCertKeyPairSecret, _, err := applyFn(ctx, c.Client, c.EventRecorder, signingCertKeyPairSecret)
 		if err != nil {
 			return nil, false, err
 		}
 		signingCertKeyPairSecret = actualSigningCertKeyPairSecret
-		signerUpdated = true
```
bertinatto, line 112: Could we remove `signerUpdated` now that we have `modified`?

pkg/operator/certrotation/signer.go
vrutkovs: Yes. Previously we used `&&` to decide if `Update` needs to be called when either type changed or metadata changed. However, `&&` didn't run `ensureSecretTLSTypeSet` if first part was true, and the type didn't change on first iteration, only the metadata.

By splitting this into separate calls we ensure both `ensureMetadataUpdate` and `ensureSecretTLSTypeSet` are called. Now we can use `||` because either of those will require a separate `Update` call - also, we want `|| modified` there too, as the secret may not exist at all.

Do you think this needs to be a comment to make sure it won't get broken later on?

pkg/operator/certrotation/signer.go
vrutkovs: We still need it in `return` - `modified` covers any change to the secret, but `signerUpdate` is a flag for signer contents change (which triggers CA bundle update in `syncWorker`)

pkg/operator/certrotation/target.go
vrutkovs: Same logic:
* `modified` may be set to true when secret doesn't exist
* both `ensureMetadataUpdate` and `ensureSecretTLSTypeSet` need to run
* if any change was done to the secret it needs to run `Update` to apply it

bertinatto (commented):
I'm not fully familiar with this controller, but I reviewed the changes and they appear to be a nice improvement. Tagging the PR, but I'll let the owners approve it.

/lgtm

pkg/operator/certrotation/cabundle.go
```diff
@@ -44,6 +44,8 @@ type CABundleConfigMap struct {
 func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingCertKeyPair *crypto.CA) ([]*x509.Certificate, error) {
 	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
 	// doesn't have any expired certs
+	needsApply := false
```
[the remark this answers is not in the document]
dinhxuanvu: Second this comment.

dinhxuanvu (commented):
Looking good. I think it is ready for a lgtm if you can update the variable name to match the others.

pkg/operator/certrotation/cabundle.go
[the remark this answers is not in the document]
vrutkovs: Fixed. `modified` was used in `ApplyConfigMap` call, so I renamed it to `updated`

dinhxuanvu (approved):
/lgtm

soltysh (approved):
/approve

PR #1732, opened 2024-05-06
OCPBUGS-23435: workloadctl: account for terminating pods

stlaz:
Don't set Progressing=False if some pods from the previous generation are still running.

/assign @openshift/openshift-team-auth 
/cc @deads2k

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -262,6 +262,21 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		desiredReplicas = *(workload.Spec.Replicas)
 	}
 
+	selector, err := metav1.LabelSelectorAsSelector(workload.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("failed to construct label selector: %v", err)
+	}
+	matchingPods, err := c.podsLister.List(selector)
+	if err != nil {
+		return err
+	}
+	// Terminatig pods don't account for any of the other status fields but
+	// still can exist in a state when they are accepting connections and would
+	// contribute to unexpected behavior if we report Progressing=False.
+	// The case of too many pods might occur for example if `TerminationGracePeriodSeconds`
+	// is set.
+	tooManyMatchingPods := int32(len(matchingPods)) > desiredReplicas
+
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
 	workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas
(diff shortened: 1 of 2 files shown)


stlaz:
/hold
I accidentally used a live client instead of a lister

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -274,6 +289,10 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		deploymentProgressingCondition.Status = operatorv1.ConditionTrue
 		deploymentProgressingCondition.Reason = "PodsUpdating"
 		deploymentProgressingCondition.Message = fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas)
+	} else if tooManyMatchingPods {
+		deploymentProgressingCondition.Status = operatorv1.ConditionTrue
+		deploymentProgressingCondition.Reason = "PreviousGenPodsPresent"
+		deploymentProgressingCondition.Message = fmt.Sprintf("deployment/%s.%s: %d pod(s) from the previous generation are still present", workload.Name, c.targetNamespace, len(matchingPods)-int(desiredReplicas))
```
atiratree, line 295: This may not always be correct. There may be extra pods for different reasons. E.g. if they are disrupted/evicted for some reason, the deployment controller will create extra pods to account for that.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -262,6 +262,21 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+	matchingPods, err := c.podsLister.List(selector)
+	if err != nil {
+		return err
+	}
+	// Terminatig pods don't account for any of the other status fields but
+	// still can exist in a state when they are accepting connections and would
+	// contribute to unexpected behavior if we report Progressing=False.
+	// The case of too many pods might occur for example if `TerminationGracePeriodSeconds`
+	// is set.
+	tooManyMatchingPods := int32(len(matchingPods)) > desiredReplicas
```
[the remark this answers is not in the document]
stlaz: is `pod-template-hash` set on 100% of our deployments?

pkg/operator/apiserver/controller/workload/workload.go
stlaz: How would you word it, then? Just too many pods?

atiratree: In k8s, this is considered a complete deployment, so it depends on what you want to convey. There can also be extra pods a rollout with maxSurge, but in this case the pods will have different revisions/hashes.

pkg/operator/apiserver/controller/workload/workload.go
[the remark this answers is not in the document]
atiratree: yes it is

https://github.com/kubernetes/kubernetes/blob/0590bb1ac495ae8af2a573f879408e48800da2c5/pkg/controller/deployment/sync.go#L191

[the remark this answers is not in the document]
stlaz: I think the pod spec can be the same, only the underlying config changed. Would that still work?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: Apart from the terminating pods which are a subject of this bug, no.

I guess, there might be some exotic cases where the pods can be owned by another controller, but that can be safely ignored here.

atiratree:
following up on this issue in https://github.com/openshift/library-go/pull/1855

PR #1739, opened 2024-05-30
NO-JIRA: add exponential backoff to static pod installer

tjungblu:
This overrides the default rate limiting for the controller to be less frequent.
Additionally, the base controller now rate limits informer invocations
in the same way.

We observed the installer controller to hot loop, especially when revisioned
resources are still missing to be copied while a revision is rolled out, which
causes CI failures with pathological events.

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -32,6 +32,14 @@ func NewSyncContext(name string, recorder events.Recorder) SyncContext {
 	}
 }
 
+// NewSyncContextQueue allows to customize the underlying work queue.
+func NewSyncContextQueue(name string, recorder events.Recorder, q workqueue.RateLimitingInterface) SyncContext {
+	return syncContext{
+		queue:         q,
+		eventRecorder: recorder.WithComponentSuffix(strings.ToLower(name)),
+	}
+}
+
 func (c syncContext) Queue() workqueue.RateLimitingInterface {
 	return c.queue
 }
@@ -88,11 +96,11 @@ func (c syncContext) eventHandler(queueKeysFunc ObjectQueueKeysFunc, filter Even
 
 func (c syncContext) enqueueKeys(keys ...string) {
 	for _, qKey := range keys {
-		c.queue.Add(qKey)
+		c.queue.AddRateLimited(qKey)
 	}
 }
 
-// namespaceChecker returns a function which returns true if an inpuut obj
+// namespaceChecker returns a function which returns true if an input obj
 // (or its tombstone) is a namespace  and it matches a name of any namespaces
 // that we are interested in
 func namespaceChecker(interestingNamespaces []string) func(obj interface{}) bool {
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"golang.org/x/time/rate"
+	"k8s.io/client-go/util/workqueue"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


tjungblu:
/hold

that didn't fix it

tjungblu:
payload run from https://github.com/openshift/cluster-etcd-operator/pull/1271 and https://pr-payload-tests.ci.openshift.org/runs/ci/37480d60-227d-11ef-895c-d46724932afb-0 shows no failure anymore.

pkg/controller/factory/controller_context.go
```diff
@@ -88,11 +96,11 @@ func (c syncContext) eventHandler(queueKeysFunc ObjectQueueKeysFunc, filter Even
 
 func (c syncContext) enqueueKeys(keys ...string) {
 	for _, qKey := range keys {
-		c.queue.Add(qKey)
+		c.queue.AddRateLimited(qKey)
```
p0lyn0mial, line 99: This will/can slow down processing of events in general.
Consider the following case:

1. We added 10 items in a rate limited way
2. The controller pops the key from the queue
3. The controller fails and adds the key back to the queue
4. The next time we will process the item will be significantly delayed.

If you want to stop producing too many events I think we could.
1. Rate limit event generation

2. Check why the events were generated in the first place.
How does the revision controller work ?
In particular why it created a new revision event if the required resources weren't created ?

I think we should better understand `2`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -189,13 +191,26 @@ func NewInstallerController(
 	}
 
 	c.ownerRefsFn = c.setOwnerRefs
-	c.factory = factory.New().WithInformers(operatorClient.Informer(), kubeInformersForTargetNamespace.Core().V1().Pods().Informer())
+	c.factory = factory.New().WithInformers(
+		operatorClient.Informer(),
+		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
+		kubeInformersForTargetNamespace.Core().V1().ConfigMaps().Informer(),
```
p0lyn0mial, line 197: This is OK. I don't like that the usage of these informers is implicit.
Could you add a comment somewhere [here](https://github.com/openshift/cluster-kube-apiserver-operator/blob/ae0c9631284bca8cfcbc305674781e097f7238ae/vendor/github.com/openshift/library-go/pkg/operator/staticpod/controllers.go#L212) saying that `configMapClient` and `secretClient` use listers and the controllers must implicitly wait unit the lister will be synced.

Please also add a comment here saying that listers come from these two clients that are passed as a dependency to this controller.

p0lyn0mial: Do you mind checking if the other controllers (accepting these two clients) properly wait for synchronisation ? (Can be a follow-up PR)

pkg/controller/factory/controller_context.go
tjungblu: then let's make this a configuration, I would like to set this behavior for all CEO controllers, I constantly run into issues like that. We can turn this config off by default.

tjungblu: (I'll send a different PR for that)

pkg/operator/staticpod/controller/installer/installer_controller.go
tjungblu: yep, let's split this PR out into two then.

pkg/controller/factory/controller_context.go
p0lyn0mial: > I would like to set this behavior for all CEO controllers

I think that customers will complain that CEO is very slow :/

tjungblu: I think it makes no sense to retry any CEO controller at 5 millis, especially when a static pod rollout takes minutes

p0lyn0mial: `enqueueKeys` is used for reacting to changes not retrying.

p0lyn0mial: Passing a custom rate limiter and adding keys in a rate limited way on errors sounds better :)

pkg/operator/staticpod/controller/installer/installer_controller.go
p0lyn0mial: ok, thanks.

pkg/controller/factory/controller_context.go
tjungblu: I know, but let's say we're blocked on progressing due to lost quorum like here: 
https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/etcdcertsigner/etcdcertsignercontroller.go#L172-L179

we had the issue recently that node status was updating several times a second (it was a bug), do you think this sync loop should be retried every couple of milliseconds on an informer event? The etcd condition is never going to change on a millisecond scale. The default settings are a total waste of CPU time.

tjungblu: Having said that, I'm pretty sure we could shave-off multiple cores in the control plane by switching to a tick-based approach of reacting to events once a second. With very little loss in reactivity to any watch event.

p0lyn0mial: Not sure, because adding 10 items with the same key doesn't mean you will get 10 items. You will get only 1 :)

tjungblu:
replaced by https://github.com/openshift/library-go/pull/1742

pkg/operator/staticpod/controller/installer/installer_controller.go
tjungblu: part two here: https://github.com/openshift/library-go/pull/1743

PR #1740, opened 2024-05-31
OCPBUGS-33958: retrieve registered secret name from secretManager

chiragkyal:
Update `IsRouteRegistered()` to `LookupRouteSecret()` which will also return the registered secret name from secretManager.

```
LookupRouteSecret returns the secret name associated with a route,
and true indicating if the route is registered with manager.
If the route is not registered, an empty string and false are returned.
```

Related to [OCPBUGS-33958](https://issues.redhat.com/browse/OCPBUGS-33958)

### DIFF
--- a/pkg/route/secretmanager/manager.go
+++ b/pkg/route/secretmanager/manager.go
@@ -17,7 +17,7 @@ type SecretManager interface {
 	RegisterRoute(ctx context.Context, namespace string, routeName string, secretName string, handler cache.ResourceEventHandlerFuncs) error
 	UnregisterRoute(namespace string, routeName string) error
 	GetSecret(ctx context.Context, namespace string, routeName string) (*v1.Secret, error)
-	IsRouteRegistered(namespace string, routeName string) bool
+	LookupRouteSecret(namespace string, routeName string) (string, bool)
 	Queue() workqueue.RateLimitingInterface
 }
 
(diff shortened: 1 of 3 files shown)


dusk125:
Can you update the PR to keep the old `IsRouteRegiestered` and add `LookupRouteSecret` as a new function please? There could be a use case for both and this would allow for existing code to stay the same, while new code and use the Lookup if needed.

chiragkyal:
>Can you update the PR to keep the old IsRouteRegiestered and add LookupRouteSecret as a new function please? There could be a use case for both and this would allow for existing code to stay the same, while new code and use the Lookup if needed.

I think `LookupRouteSecret` is an updated version of `IsRouteRegiestered`, which works similarly with an extra return type. AFAIK, only https://github.com/openshift/router is consuming this code, and I've a PR to update it https://github.com/openshift/router/pull/614. Do we have anything else which could be impacted by this change?

pkg/route/secretmanager/manager.go
```diff
@@ -29,7 +29,7 @@ type manager struct {
 	// Map of registered handlers for each route.
 	// Populated inside RegisterRoute() and used in UnregisterRoute(), GetSecret.
 	// generateKey() will create the map key.
-	registeredHandlers map[string]secret.SecretEventHandlerRegistration
+	registeredHandlers map[string]*referencedSecret
```
Miciah, line 32: Is there a reason to use a pointer type `map[string]*referencedSecret` instead of `map[string]referencedSecret`?

chiragkyal: Thanks for pointing that out! I've updated the `registeredHandlers` map to use `map[string]referencedSecret` instead of `map[string]*referencedSecret`. Since referencedSecret doesn't need to be modified once it's stored in the map, storing it by value is better.

Miciah:
Thanks!

/lgtm

PR #1741, opened 2024-05-31
STOR-1767: Delete resources if spec.ManagementState is equal to Removed for removable CSI drivers

mpatlasov:
We want to disable storage if system administrator sets `spec.ManagementState` of ClusterCSIDriver to `Removed`. This PR ensures that NodeService and Deployment controllers delete resources when `spec.ManagementState` is `Removed`.

Also, it patches `StaticResourceController` Sync() to call `DeleteAll()` if the ManagementState is Removed, and implements support of multiple conditional StaticResourceControllers.

/cc @openshift/storage @jsafrane

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -32,16 +32,16 @@ var defaultCacheSyncTimeout = 10 * time.Minute
 
 // CSIControllerSet contains a set of controllers that are usually used to deploy CSI Drivers.
 type CSIControllerSet struct {
-	logLevelController                   factory.Controller
-	managementStateController            factory.Controller
-	staticResourcesController            factory.Controller
-	conditionalStaticResourcesController factory.Controller
-	credentialsRequestController         factory.Controller
-	csiConfigObserverController          factory.Controller
-	csiDriverControllerServiceController factory.Controller
-	csiDriverNodeServiceController       factory.Controller
-	serviceMonitorController             factory.Controller
-	csiStorageclassController            factory.Controller
+	logLevelController                    factory.Controller
+	managementStateController             factory.Controller
+	staticResourcesController             factory.Controller
+	conditionalStaticResourcesControllers []factory.Controller
+	credentialsRequestController          factory.Controller
+	csiConfigObserverController           factory.Controller
+	csiDriverControllerServiceController  factory.Controller
+	csiDriverNodeServiceController        factory.Controller
+	serviceMonitorController              factory.Controller
+	csiStorageclassController             factory.Controller
 
 	operatorClient v1helpers.OperatorClientWithFinalizers
 	eventRecorder  events.Recorder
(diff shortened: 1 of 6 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -271,8 +270,10 @@ func (c *CSIControllerSet) WithStorageClassController(
 
 // New returns a basic *ControllerSet without any controller.
 func NewCSIControllerSet(operatorClient v1helpers.OperatorClientWithFinalizers, eventRecorder events.Recorder) *CSIControllerSet {
-	return &CSIControllerSet{
+	c := &CSIControllerSet{
```
dusk125, line 273: nit:
This could remain as anonymous structure pointer return
```go
return &CSIControllerSet{
  operatorClient: operatorClient,
  eventRecorder: eventRecorder,
  conditionalStaticResourcesControllers: []factory.Controller{},
}
```

pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -96,7 +95,7 @@ func (c *CSIControllerSet) WithConditionalStaticResourcesController(
 	files []string,
 	shouldCreateFnArg, shouldDeleteFnArg resourceapply.ConditionalFunction,
 ) *CSIControllerSet {
-	c.conditionalStaticResourcesController = staticresourcecontroller.NewStaticResourceController(
+	c.conditionalStaticResourcesControllers = append(c.conditionalStaticResourcesControllers, staticresourcecontroller.NewStaticResourceController(
```
dusk125, line 98: Can you explain why we need to move to a slice of controllers here please?

It seems like the current expectation is that a user would only specify a single Conditional Static Resources Controller (the function doc still refers to adding a singular controller). If this expectation has changed, can you provide some information that is driving this please?

mpatlasov: Thank you for review , @dusk125 ! I explained it generally in commit message:

> Different assets may require different Create/Delete conditions.

And more specifically, we already have one single controller in [driver_starter.go](https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/master/pkg/operator/vspherecontroller/driver_starter.go#L82):

```
	).WithConditionalStaticResourcesController(
		"VMwareVSphereDriverConditionalStaticResourcesController",
		c.kubeClient,
		c.apiClients.DynamicClient,
		c.apiClients.KubeInformers,
		assets.ReadFile,
		[]string{
			"volumesnapshotclass.yaml",
		},
```
Create/Delete functions there are specific for VolumeSnapshotClass. Now I want to convert its StaticResourcesController:
```
	).WithStaticResourcesController(
		"VMwareVSphereDriverStaticResourcesController",
		c.kubeClient,
		c.apiClients.DynamicClient,
		c.apiClients.KubeInformers,
		assets.ReadFile,
		[]string{
			"rbac/privileged_role.yaml",
```
to ConditionalStaticResourcesController and need to specify other Create/Delete functions for them -- they will check for ManagementState==Removed.

dusk125: Thanks @mpatlasov! That makes sense and I think it's fine to allow this function to be called multiple times in order to setup multiple ConditionalControllers.

Can you update the doc comment on the WithConditional function to include that it can be called multiple times to setup multiple controllers and that it only adds and does not replace any existing CondtionalControllers please?

mpatlasov: Done:
```
+// The func can be called multiple times to setup multiple controllers. The calls do not replace any existing controllers.
```

pkg/operator/csi/csicontrollerset/csi_controller_set.go
mpatlasov: Done

PR #1742, opened 2024-06-05
ETCD-627: configure static pod installer rate limits and backoffs

tjungblu:
This PR allows the sync context to be configured with rate limits and a flag to throttle informer events. It also allows the static pod installer to pass its sync context from the builder.

### DIFF
--- a/pkg/controller/factory/controller_context.go
+++ b/pkg/controller/factory/controller_context.go
@@ -17,9 +17,10 @@ import (
 // syncContext implements SyncContext and provide user access to queue and object that caused
 // the sync to be triggered.
 type syncContext struct {
-	eventRecorder events.Recorder
-	queue         workqueue.RateLimitingInterface
-	queueKey      string
+	eventRecorder          events.Recorder
+	queue                  workqueue.RateLimitingInterface
+	queueKey               string
+	rateLimitEventsEnabled bool
 }
 
 var _ SyncContext = syncContext{}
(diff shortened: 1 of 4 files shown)


pkg/controller/factory/controller_context.go
```diff
@@ -88,7 +102,11 @@ func (c syncContext) eventHandler(queueKeysFunc ObjectQueueKeysFunc, filter Even
 
 func (c syncContext) enqueueKeys(keys ...string) {
 	for _, qKey := range keys {
-		c.queue.Add(qKey)
+		if c.rateLimitEventsEnabled {
+			c.queue.AddRateLimited(qKey)
```
p0lyn0mial, line 106: I still don't understand how this could help in general.
Note that adding 10 items with the same key doesn't mean you will get 10 items. You will get only 1.

I though you wanted to pass a custom rate limiter by using the already existing method `WithSyncContext`.

In case you want to only resync every X seconds then instead of changing the factory, you could:

1. Apply a filter function for your informers via `WithFilteredEventsInformers`
2. Discards all updates in the the filters.
3. Set a custom resync of X seconds.

1. Register informers with `WithBareInformers` (double check it won't register any event handlers)
2. Set a custom resync of X seconds.

I think it should help you validate your idea (which IMO will slow down your operator).

p0lyn0mial: Try with `WithBareInformers` I think you won't get any events and the add a custom resync to your controller.

tjungblu: so I should fork the entire static pod installer_controller.go in CEO to just set a resync interval?

tjungblu: > 1. Rate limit event generation

Yes, I can do that - wouldn't it be convenient if there is a queue that I can configure with a rate limiter already in place for that?

> isn't that strange that only CEO is producing too many events ?

it certainly isn't: 
https://search.dptools.openshift.org/?search=RequiredInstallerResourcesMissing+.*kube.*&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

I see 
* openshift-kube-controller-manager-operator
* openshift-kube-apiserver-operator

tjungblu:
closing in favor of #1744

PR #1745, opened 2024-06-06
csi: Add Deployment, DaemonSet hooks for config map hash annotations

stephenfin:
We already have these for secrets. Add them for config maps.

This is needed for the OpenStack Cinder CSI Driver Operator. See https://github.com/openshift/openstack-cinder-csi-driver-operator/pull/168/ for more information.

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
@@ -102,6 +102,25 @@ func WithCABundleDeploymentHook(
 	}
 }
 
+// WithConfigMapHashAnnotationHook creates a deployment hook that annotates a Deployment with a config map's hash.
+func WithConfigMapHashAnnotationHook(
+	namespace string,
+	configMapName string,
+	configMapInformer corev1.ConfigMapInformer,
+) dc.DeploymentHookFunc {
+	return func(opSpec *opv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		inputHashes, err := resourcehash.MultipleObjectHashStringMapForObjectReferenceFromLister(
+			configMapInformer.Lister(),
+			nil,
+			resourcehash.NewObjectRef().ForConfigMap().InNamespace(namespace).Named(configMapName),
+		)
+		if err != nil {
+			return fmt.Errorf("invalid dependency reference: %w", err)
+		}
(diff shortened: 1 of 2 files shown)


stephenfin:
/retest-required

CI failures appear to be due to timeouts, not anything to do with this change.

dusk125:
non-binding lgtm

deads2k:
this looks ok, but would we better off by refactoring `addObjectHash` to take a `metav1.Object`  instead of a daemonset or deployment and then collapse the cases?

/approve
delegated to @dusk125

stephenfin:
> this looks ok, but would we better off by refactoring `addObjectHash` to take a `metav1.Object` instead of a daemonset or deployment and then collapse the cases?

Happy to do this here or in a follow-up PR, whatever works.

dusk125:
> Happy to do this here or in a follow-up PR, whatever works.

I think it's fine to do that here. Thanks!

stephenfin:
> this looks ok, but would we better off by refactoring `addObjectHash` to take a `metav1.Object` instead of a daemonset or deployment and then collapse the cases?

Actually, looking into this more, `metav1.Object` is an interface meaning I can't access e.g. `.Spec.Template.Annotations` without a cast...which kind of defeats the whole point of DRYing things up. We'd also need to move the function to a common location and make it public. I'm not sure if we want to do that either.

Have I missed something?

bertinatto:
> > this looks ok, but would we better off by refactoring `addObjectHash` to take a `metav1.Object` instead of a daemonset or deployment and then collapse the cases?
> 
> Actually, looking into this more, `metav1.Object` is an interface meaning I can't access e.g. `.Spec.Template.Annotations` without a cast...which kind of defeats the whole point of DRYing things up. We'd also need to move the function to a common location and make it public. I'm not sure if we want to do that either.
> 
> Have I missed something?
> 
> _Edit:_ We could add two call to the new `AddObjectHash`, one for `deployment` and one for `deployment.Spec.Template` (ditto for `daemonset`) but again, that's not all that DRY.

Right, we'd need to use type assertion, like we do in the static resource controller [here](https://github.com/bertinatto/library-go/blob/e20ca28ddfe7a929b914a626bd4f80c30b6bd7b7/pkg/operator/staticresourcecontroller/static_resource_controller.go#L222). Since the code duplication hasn't been introduced by this PR, I personally think  it's fine doing this refactoring in a separate PR.

/lgtm

PR #1746, opened 2024-06-10
OCPBUGS-35231: cert-inspection: parse secret/configmap/files as kubeconfigs

vrutkovs:
Extract CA and certs used in kubeconfigs as TLS artifacts too.

Tested in https://github.com/openshift/origin/pull/28864

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
@@ -6,6 +6,8 @@ import (
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
 	certificatesv1 "k8s.io/api/certificates/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/cert"
 )
 
@@ -13,6 +15,9 @@ func InspectSecret(obj *corev1.Secret) (*certgraphapi.CertKeyPair, error) {
 	resourceString := fmt.Sprintf("secrets/%s[%s]", obj.Name, obj.Namespace)
 	tlsCrt, isTLS := obj.Data["tls.crt"]
 	if !isTLS {
+		if detail, err := InspectSecretAsKubeConfig(obj); err == nil {
+			return detail, nil
+		}
 		return nil, nil
 	}
 	//fmt.Printf("%s - tls (%v)\n", resourceString, obj.CreationTimestamp.UTC())
@@ -62,6 +67,9 @@ func InspectCSR(obj *certificatesv1.CertificateSigningRequest) (*certgraphapi.Ce
 func InspectConfigMap(obj *corev1.ConfigMap) (*certgraphapi.CertificateAuthorityBundle, error) {
 	caBundle, ok := obj.Data["ca-bundle.crt"]
 	if !ok {
+		if detail, err := InspectConfigMapAsKubeConfig(obj); err == nil {
+			return detail, nil
+		}
 		return nil, nil
 	}
 	if len(caBundle) == 0 {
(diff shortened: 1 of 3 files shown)


dusk125:
non-binding lgtm

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
```diff
@@ -77,6 +85,92 @@ func InspectConfigMap(obj *corev1.ConfigMap) (*certgraphapi.CertificateAuthority
...
 	}
 	caBundleDetail = addConfigMapLocation(caBundleDetail, obj.Namespace, obj.Name)
+	return caBundleDetail, nil
+}
+
+func extractKubeConfigFromConfigMap(obj *corev1.ConfigMap) (*rest.Config, error) {
+	if obj == nil {
+		return nil, fmt.Errorf("empty object")
+	}
+	for _, v := range obj.Data {
```
deads2k, line 95: this appears to be parsing every key of potentially every configmap to see if it's a kubeconfig?  Instead of doing this, could we choose specific keys we find valid?

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
```diff
@@ -77,6 +85,92 @@ func InspectConfigMap(obj *corev1.ConfigMap) (*certgraphapi.CertificateAuthority
 		return nil, err
 	}
 	caBundleDetail = addConfigMapLocation(caBundleDetail, obj.Namespace, obj.Name)
+	return caBundleDetail, nil
+}
+
+func extractKubeConfigFromConfigMap(obj *corev1.ConfigMap) (*rest.Config, error) {
```
deads2k, line 91: might be more useful to return the actual kubeconfig type since it can contain multiple stanzas and rest.Config only holds one.

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
vrutkovs: Yeah, I can modify this to a hardcoded list of keys, but I'm worried that it might miss some new kubeconfigs.

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
vrutkovs: Good idea, I'll implement that

vrutkovs: This would require reworking entire `InspectSecret/InspectConfigMap` as they are expected to return just one secret. I'll implement this in a separate PR

vrutkovs: Followup - https://github.com/openshift/library-go/pull/1751

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
```diff
@@ -6,13 +6,18 @@ import (
...
 	"k8s.io/client-go/util/cert"
 )
 
 func InspectSecret(obj *corev1.Secret) (*certgraphapi.CertKeyPair, error) {
 	resourceString := fmt.Sprintf("secrets/%s[%s]", obj.Name, obj.Namespace)
 	tlsCrt, isTLS := obj.Data["tls.crt"]
 	if !isTLS {
+		if detail, err := InspectSecretAsKubeConfig(obj); err == nil {
+			return detail, nil
+		}
```
bertinatto, line 20: Would it be useful to somehow log that this has been attempted and an error occurred?

bertinatto (commented):
LGTM.
Given my limited knowledge here, I haven't found anything obviously wrong. Just added a question.

pkg/certs/cert-inspection/certgraphanalysis/analyzer.go
vrutkovs: Any error is treated as critical here, so "collect TLS artifacts" test would fail if errors found. Also, any valid certificate would be treated as invalid kubeconfig, so we should continue instead of returning error.

bertinatto:
> LGTM. Given my limited knowledge here, I haven't found anything obviously wrong. Just added a question.

/approve

I'd appreciate if someone more familiar with this code would give it a final review.

sanchezl (approved):
/lgtm

PR #1747, opened 2024-06-11
certgraphanalysis: include ondisk certs in ElideProxyCADetails

vrutkovs:
Include several ondisk certs in ElideProxyCADetails to mark them as trusted CA bundles

Tested in https://github.com/openshift/origin/pull/28868


dusk125:
non-binding lgtm

deads2k:
someday it would be nice to pay down debts, but this is better than where we're currently at

/approve
delegating to @dusk125

PR #1748, opened 2024-06-14
Add support for ValidatingAdmissionPolicy v1

dinhxuanvu:
For k8s 1.30, ValidatingAdmissionPolicy is GA and only v1 is enabled by default.

### DIFF
--- a/pkg/operator/resource/resourceapply/admissionregistration.go
+++ b/pkg/operator/resource/resourceapply/admissionregistration.go
@@ -226,6 +226,65 @@ func ApplyValidatingAdmissionPolicyV1beta1(ctx context.Context, client admission
 	return actual, true, nil
 }
 
+// ApplyValidatingAdmissionPolicyV1 ensures the form of the specified
+// validatingadmissionpolicyconfiguration is present in the API. If it does not exist,
+// it will be created. If it does exist, the metadata of the required
+// validatingadmissionpolicyconfiguration will be merged with the existing validatingadmissionpolicyconfiguration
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
This PR is a part of a series of works to enable ValidatingAdmissionPolicy v1 and remove v1beta1.
- https://github.com/openshift/machine-config-operator/pull/4407
- https://github.com/openshift/cluster-kube-apiserver-operator/pull/1687
FYI @soltysh

deads2k:
/hold

Doing this will make it impossible to have a single version of an operator built to service 4.16 techpreview and 4.17.  Instead of doing this, please add new methods for v1 and keep the old.

dinhxuanvu:
@deads2k Done. PTAL.

dusk125:
/lgtm
Leaving for David to final review and remove the hold

bertinatto (commented):
LGTM too. 
The new V1 functions look correct and consistent with the v1beta1 ones, which are in use today (in MCO at least).

soltysh (approved):
/approve
/hold cancel

djoshy:
Hello! We ran into an interesting hot looping apply bug on the MCO side with the policies: https://issues.redhat.com/browse/OCPBUGS-36654, I'm not sure if this is related to when the feature GAed or something else, so would appreciate a look. 

The MCO flow also need v1 versions of the [resourceread](https://github.com/openshift/library-go/blob/51af9160f7d8223c0f70eb2cdf167a7ef00859ad/pkg/operator/resource/resourceread/admission.go#L39-L55) functions(although if there is a way to just use apply directly, we're happy to adopt that too)

PR #1749, opened 2024-06-14
WIP: Added the callback shouldRevisionInstall to provide additional c…

jubittajohn:
…onditions for installation


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -442,6 +449,16 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 		return true, requeueAfter, nil
 	}
 
+	// Check if revision should be installed based on if the quorum is about to be violated
+	if c.shouldRevisionInstall != nil {
+		quorumSafe, err := c.shouldRevisionInstall()
+		if !quorumSafe {
+			return true, 30 * time.Second, err
```
lance5890, line 456: - when quorumSafe failed, the err is not nil, then  return `30 * time.Second`makes no sense, what about `return true, 0, err`
- please see https://github.com/openshift/library-go/blob/aed018c215a122871be1768155cf9f3e658278fc/pkg/operator/staticpod/controller/installer/installer_controller.go#L1070-L1074

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -442,6 +449,16 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 		return true, requeueAfter, nil
 	}
 
+	// Check if revision should be installed based on if the quorum is about to be violated
+	if c.shouldRevisionInstall != nil {
+		quorumSafe, err := c.shouldRevisionInstall()
```
tjungblu, line 454: I would avoid leaking our usage of "quorum" in the variable name, just call it `shouldInstall`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -442,6 +449,16 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 		return true, requeueAfter, nil
 	}
 
+	// Check if revision should be installed based on if the quorum is about to be violated
+	if c.shouldRevisionInstall != nil {
```
tjungblu, line 453: as far as I understand, this blocks the entire controller, which we don't want, we want the current revisions to continue to rollout. 

We can be more granular here and just block the `ensureInstallerPod` routine from spawning just the installer pod. That way it wouldn't block existing rollouts. Having said that, I think there are generally very little tests around this, so we need to run a couple of payload tests in cluster-etcd-operator (CEO) as well

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -852,6 +859,13 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
 
 // ensureInstallerPod creates the installer pod with the secrets required to if it does not exist already
 func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, ns *operatorv1.NodeStatus) error {
+	// checks if new revision should be rolled out
+	if c.shouldRevisionInstall != nil {
+		shouldInstall, err := c.shouldRevisionInstall()
+		if !shouldInstall {
```
tjungblu, line 865: shouldn't that be

```suggestion
		if err != nil {
    		   return err
		}
		if !shouldInstall {
		  return nil
		}
```

tjungblu: maybe also add an info logging statement here, so we know if it was skipped

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -2084,22 +2085,123 @@ func TestInstallerController_manageInstallationPods(t *testing.T) {
...
+		{
+			name: "if the revision shouldn't install and a new revision is to be rolled out, requeue the new installer pod",
+			fields: fields{
+				targetNamespace: "test-namespace",
+				staticPodName:   "test-pod",
+				command:         []string{},
+				kubeClient:      fake.NewSimpleClientset(),
+				eventRecorder:   eventstesting.NewTestingEventRecorder(t),
+				shouldRevisionInstall: func() (bool, error) {
+					return false, fmt.Errorf("revision shouldn't be installed")
```
tjungblu, line 2097: ah okay, now I get it. If you want to use an error for control flow, that's fine. But then we don't really need the boolean to denote whether to skip it or not. 

I (personally) would use the error for real errors though.

pkg/operator/staticpod/controllers.go
```diff
@@ -75,6 +75,9 @@ type staticPodOperatorControllerBuilder struct {
 	readyzEndpoint                string
 	pdbUnhealthyPodEvictionPolicy *v1.UnhealthyPodEvictionPolicyType
 	guardCreateConditionalFunc    func() (bool, bool, error)
+
+	// should revision install information
```
tjungblu, line 79: not sure what this comment is supposed to mean

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -2084,22 +2085,123 @@ func TestInstallerController_manageInstallationPods(t *testing.T) {
...
+					return true, nil
+				},
+			},
+			args: args{
+				operatorSpec: &operatorv1.StaticPodOperatorSpec{
+					OperatorSpec: operatorv1.OperatorSpec{},
+				},
+				originalOperatorStatus: &operatorv1.StaticPodOperatorStatus{
+					LatestAvailableRevision: 1,
+					NodeStatuses: []operatorv1.NodeStatus{
```
tjungblu, line 2170: I only very superficially looked through your tests, but I'm generally missing one with multiple NodeStatuses. 

Maybe more meaningful for etcd would be a test with three nodes at those revisions:

master-0 = 4
master-1 = 3 (would be next to get to 4)
master-2 = 3

Now our quorum guard says master-0 went away (eg, machine was turned off). What happens to the remaining revisions? 

I would expect this to block the installation of rev 4 on master-1 and master-2. 
Maybe also think of a couple of such tests, maybe with a more structured permutation approach. We can discuss this also next Tuesday.

pkg/operator/staticpod/controllers.go
jubittajohn: Here I was trying to provide a description for the newly added field `shouldRevisionInstall`. May be I should be using a more meaningful description.

pkg/operator/staticpod/controller/installer/installer_controller.go
jubittajohn: @tjungblu But only if there is error the requeue  will happen as per the current logic here. https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/installer_controller.go#L481.

tjungblu: then we need to rethink the ensureInstallerPod signature potentially. I don't think we should trigger an event for this for example:
```
c.eventRecorder.Warningf("InstallerPodFailed", "Failed to create installer pod for revision %d count %d on node %q: %v", 
       currNodeState.TargetRevision, currNodeState.LastFailedCount, currNodeState.NodeName, err) 
```

that's misleading, because there was no actual error that failed the installation. We just skipped it.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -852,6 +859,13 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
...
+			return err
```
tjungblu, line 866: @jubittajohn  there's an example below in this file on how to log: 
> klog.V(4).Infof("Backing off node %s installer retry %d until %v", currNodeState.NodeName, currNodeState.LastFailedCount+1, earliestRetry)

you can keep the same verbosity level, just think of a useful log message

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -478,14 +486,19 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
...
 					c.eventRecorder.Warningf("InstallerPodFailed", "Failed to create installer pod for revision %d count %d on node %q: %v",
 						currNodeState.TargetRevision, currNodeState.LastFailedCount, currNodeState.NodeName, err)
 					// if a newer revision is pending, continue, so we retry later with the latest available revision
 					if !(operatorStatus.LatestAvailableRevision > currNodeState.TargetRevision) {
 						return true, 0, err
 					}
 				}
+				if requeue {
+					klog.V(4).Infof("Requeuing the creation of installer pod for revision %d on node %q", currNodeState.TargetRevision, currNodeState.NodeName)
+					return true, requeueAfter, nil
```
lance5890, line 500: - when hit the requeue logic , the requeueAfter should not be greater than 0 
- check  https://github.com/jubittajohn/library-go/blob/963d6624ccdd64dbc9df2dd52f7a401b7e79f314/pkg/operator/staticpod/controller/installer/installer_controller.go#L445-L447

tjungblu:
/lgtm (non binding)

can you please also squash all commits into one and remove the WIP please?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -98,6 +98,9 @@ type InstallerController struct {
 	clock            clock.Clock
 	installerBackOff func(count int) time.Duration
 	fallbackBackOff  func(count int) time.Duration
+
+	// shouldRevisionInstall is a callback function that determines whether a new revision should be installed
+	shouldRevisionInstall func() (bool, error)
```
dgrisonnet, line 103: I think we should follow the same precondition pattern that we have in the static resource controller: https://github.com/openshift/library-go/blob/master/pkg/operator/staticresourcecontroller/static_resource_controller.go#L55-L59. Meaning that we would allow for any condition to be defined to gate the revision process.

Also, we should pass on a context in that method and propagate it all the way down in case we need to cancel it for some reasons. Especially in https://github.com/openshift/cluster-etcd-operator/blob/master/pkg/operator/ceohelpers/bootstrap.go#L139 and maybe some other functions that don't take it yet.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -851,7 +864,18 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
 }
 
 // ensureInstallerPod creates the installer pod with the secrets required to if it does not exist already
-func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, ns *operatorv1.NodeStatus) error {
+// returns whether or not requeue and if an error happened while creating installer pod
+func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, ns *operatorv1.NodeStatus) (bool, error) {
+	// checks if new revision should be rolled out
+	if c.shouldRevisionInstall != nil {
```
dgrisonnet, line 870: I think this check should be moved to `manageInstallationPods` because this function is supposed to be only about creating the installer pod.

dgrisonnet: we could maybe even move it inside the `sync` function before the call to `manageInstallationPods` since we want to skip the installation process entirely when the preconditions are not fulfilled.

tjungblu: yeah we thought so, but it's better to continue the current rollout. See https://github.com/openshift/library-go/pull/1749#discussion_r1642499291

pkg/operator/staticpod/controller/installer/installer_controller.go
tjungblu: good idea, @jubittajohn can we create it slightly different like this?

> type StaticPodInstallerPreconditionsFuncType func(ctx context.Context) (bool, error)

and then the struct value could look like this

>	installPrecondition StaticPodInstallerPreconditionsFuncType

that way is more consistent with the other packages.

dusk125:
lgtm, but I'll let Damien have the last word

pkg/operator/staticpod/controller/installer/installer_controller.go
dgrisonnet: good point. it made me notice that we might be subject to the following issue: https://github.com/openshift/library-go/pull/1749#discussion_r1662686144

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -851,7 +866,18 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
...
+// returns whether or not requeue and if an error happened while creating installer pod
+func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, ns *operatorv1.NodeStatus) (bool, error) {
+	// checks if new revision should be rolled out
+	if c.installPrecondition != nil {
+		shouldInstall, err := c.installPrecondition(ctx)
+		if err != nil {
+			return true, err
+		}
+		if !shouldInstall {
+			return true, nil
```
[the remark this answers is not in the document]
jubittajohn: @dgrisonnet As pointed out , `ensureInstallerPod` returns `true` at t2.  
@tjungblu The `if `condition mentioned would evaluate to `false` at t2 allowing the installer creation process to continue. The mentioned condition only evaluates to `true` when a later revision is pending compared to the one currently being rolled out. Therefore, the `ensureInstallerPod` is invoked.

To address this, I have added a check as suggested  to ensure that preconditions are only evaluated when the installer pod isn't already present.  Additionally, I  have added a unit test for this scenario to confirm this behavior.  

Could you please review the new changes ?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -851,7 +866,26 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
 }
 
 // ensureInstallerPod creates the installer pod with the secrets required to if it does not exist already
-func (c *InstallerController) ensureInstallerPod(ctx context.Context, operatorSpec *operatorv1.StaticPodOperatorSpec, ns *operatorv1.NodeStatus) error {
+// returns whether or not requeue and if an error happened while creating installer pod
```
tjungblu, line 869:
```suggestion
// returns whether or not to requeue and if an error happened while ensuring the installer pod
```

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -851,7 +866,26 @@ func getInstallerPodName(ns *operatorv1.NodeStatus) string {
...
+	installerPodName := getInstallerPodName(ns)
+	_, err := c.podsGetter.Pods(c.targetNamespace).Get(ctx, installerPodName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		if c.installPrecondition != nil {
+			shouldInstall, err := c.installPrecondition(ctx)
+			if err != nil {
+				return true, err
+			}
+			if !shouldInstall {
+				return true, nil
```
tjungblu, line 882: can we leave a log statement here, too, please?

tjungblu:
```suggestion
                                klog.Infof("Skipping the creation of installer pod [%s] because of precondition check", installerPodName)
				return true, nil
```

tjungblu: just as an example, feel free to find a more fitting log statement

pkg/operator/staticpod/controller/installer/installer_controller.go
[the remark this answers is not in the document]
tjungblu: looks good, thanks for the test case :)

tjungblu:
@dgrisonnet any progress here? I would like to add this before feature freeze so we have some time to soak this in CEO.

dgrisonnet:
I synced on slack with @jubittajohn, but there was some concerns from my team with adding new logic to the installer controller, so we were exploring some other options.

The first one was the revision controller, but this was not possible because rollouts can still occur even if a revision is gated.
The other option that we are currently investigating is the etcd quorum guard pod: https://github.com/openshift/cluster-etcd-operator/blob/3ee038dbe146adc8aa9ef9a6624af1f1f33c5936/docs/etcd-quorum-guard.md
The idea is that if we were able to migrate the quorum safety check from the operator to an etcd readyz check, we could leverage guard pods to gate rollouts.

dgrisonnet:
That said, if the investigation doesn't conclude, I am fine with going with this approach.

dusk125:
@dgrisonnet we had a quorum check implicitly a part of the readyz implementation that was removed to fix: https://issues.redhat.com//browse/OCPBUGS-19987 and PR https://github.com/openshift/cluster-etcd-operator/pull/1134 - the implicit check here is that a linearizable read needs quorum where a serizialable does not.

The gist is that a node can still be 'ready' while quorum is down - being able to still serve serializable requests (instead of linearizable requests).

IMO, this isn't adding that much complexity to the installer controller, just a "should we continue installing or not", the complexity of that decision lies with the actual user of the installer controller.

tjungblu:
Thanks @dgrisonnet 

> there was some concerns from my team with adding new logic to the installer controller,

@soltysh / @deads2k  can we assume that static pods in library-go are deprecated then? This is already the second change we could not make to bugfix etcd operator because of concerns from the team in the last four weeks.

dgrisonnet:
@dusk125 etcd seems to have readyz checks for both serializable and linearizable requests: https://github.com/etcd-io/etcd/blob/8b1b69b1e26621e1038f78a3fe84e0919b00034d/server/etcdserver/api/etcdhttp/health.go#L252-L254. As far as I understand it, if any of these would fail, the node would report not ready. So if quorum was down the node wouldn't be able to serve linearizable requests and would report not ready.

I looked at the links you shared, but I am failing to see when etcd-operator readyz comes into the picture. https://github.com/openshift/cluster-etcd-operator/blob/37ac49d0f8695be21ba8a75b0fab161c8a083121/pkg/operator/starter.go#L311-L324 should be pointing to etcd members readyz endpoints. So there might be something that I don't understand here, and I am also wondering if the change is still relevant with the new http probes in the picture.

I am not against the change in the installer controller, but from my perspective having a solution relying on guard pods which protects pods from rollouts is better than a solution that prevents the installation of new revisions. And looking at the new probes from a high level, it seems quite promising to me.

dusk125:
@dgrisonnet it's true that etcd *can* do both, that bug makes it so that we only use serializable reads. The operator controls the pod-spec for the etcd pods and thus, the readyz command for etcd comes from the operator: https://github.com/openshift/cluster-etcd-operator/blob/master/bindata/etcd/pod.yaml#L289

PR #1751, opened 2024-06-20
NO-JIRA: return multiple certificates when scanning secrets

vrutkovs:


vrutkovs:
Included in https://github.com/openshift/library-go/pull/1746

PR #1752, opened 2024-07-03
operator/certrotation: introduce an optional lock to CABundleConfigMap and RotatedSigningCASecret

p0lyn0mial:
To preserve the current behaviour and make the controllers thread safe
I've decided to introduce an optional mutex which can be used when the instance 
is shared across multiple controllers.
This approach requires the fewest changes and resolves the race condition.

I think even a simpler version of this approach would work: https://github.com/openshift/library-go/pull/1753

Perhaps the best approach would be to create separate controllers for `CABundleConfigMap` and `RotatedSigningCASecret`.

I've also opened a proof PR at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1709

Alternative for https://github.com/openshift/library-go/pull/1722

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
/hold

before merging we should run the test that fails due to the race condition.

p0lyn0mial:
/close

in favour of https://github.com/openshift/library-go/pull/1753

PR #1753, opened 2024-07-03
operator/certrotation: introduce an optional lock to CABundleConfigMap and RotatedSigningCASecret

p0lyn0mial:
To preserve the current behaviour and make the controllers thread safe
I've decided to introduce an optional mutex which can be used when the instance
is shared across multiple controllers.

This approach requires the fewest changes and resolves the race condition.

Perhaps the best approach would be to create separate controllers for `CABundleConfigMap` and `RotatedSigningCASecret.`

proof: https://gist.github.com/p0lyn0mial/176bece543907dbd9ac6b5306a170eda

a simpler version of https://github.com/openshift/library-go/pull/1752

Alternative for https://github.com/openshift/library-go/pull/1722

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"sync"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
/hold

before merging we should run the test that fails due to the race condition, see: https://github.com/openshift/library-go/pull/1722#issuecomment-2206471815

p0lyn0mial:
/close 

now, once we know where the issue is we can apply a different fix https://github.com/openshift/library-go/pull/1763

PR #1754, opened 2024-07-05
OCPBUGS-34373: routes/custom-host permission update for externalCertificate

chiragkyal:
- “create” is required to set `spec.tls.externalCertificate` on route creation

- either “create” or “update”  is required to update `spec.tls.externalCertificate`
- Note: since the state of the external secret cannot be verified, even keeping the secret name unchanged, requires permission check.

-  NO PERMISSION is required to remove `spec.tls.externalCertificate`

- EP changes https://github.com/openshift/enhancements/pull/1652

The decision was taken based on https://github.com/openshift/origin/pull/18177#issuecomment-360660024 discussion on custom-host permission and keep it identical with the existing `spec.tls.Certificate` field.


pkg/route/hostassignment/assignment_test.go
```diff
@@ -197,6 +216,29 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Key: "b"},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			// removing key info is allowed
+			errs: 0,
+		},
```
alebedev87, line 241: Do we need these tests for the change you did or you just wanted to enhance the unit tests for other cases?

pkg/route/hostassignment/assignment_test.go
```diff
@@ -219,14 +261,37 @@ func TestHostWithWildcardPolicies(t *testing.T) {
...
+			host:           "host",
+			expected:       "host",
+			oldHost:        "host",
+			tls:            &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge},
+			oldTLS:         &routev1.TLSConfig{Termination: routev1.TLSTerminationEdge, Certificate: "b"},
+			wildcardPolicy: routev1.WildcardPolicyNone,
+			allow:          false,
+			// removing certificate info is allowed
+			errs: 0,
+		},
```
alebedev87, line 286: Same question.

pkg/route/hostassignment/assignment.go
```diff
@@ -246,7 +255,9 @@ func ValidateHostUpdate(ctx context.Context, route, older *routev1.Route, sarc r
 				if route.Spec.TLS.ExternalCertificate == nil || older.Spec.TLS.ExternalCertificate == nil {
 					errs = append(errs, validateImmutableField(route.Spec.TLS.ExternalCertificate, older.Spec.TLS.ExternalCertificate, field.NewPath("spec", "tls", "externalCertificate"), routeTLSPermissionErrMsg)...)
 				} else {
-					errs = append(errs, validateImmutableField(route.Spec.TLS.ExternalCertificate.Name, older.Spec.TLS.ExternalCertificate.Name, field.NewPath("spec", "tls", "externalCertificate"), routeTLSPermissionErrMsg)...)
+					// since the state of the external secret cannot be verified, return error (even when secret name remains unchanged)
+					// without performing immutability checks, if externalCertificate is set.
+					errs = append(errs, field.Invalid(field.NewPath("spec", "tls", "externalCertificate"), route.Spec.TLS.ExternalCertificate, routeTLSPermissionErrMsg))
```
alebedev87, line 260: Which unit test covers this change?

pkg/route/hostassignment/assignment.go
```diff
@@ -137,7 +137,7 @@ func certificateChangeRequiresAuth(route, older *routev1.Route, opts route.Route
 			a.Key != b.Key
 
 		if opts.AllowExternalCertificates {
-			if route.Spec.TLS.ExternalCertificate != nil || older.Spec.TLS.ExternalCertificate != nil {
+			if route.Spec.TLS.ExternalCertificate != nil {
```
alebedev87, line 140: I suppose this change is responsible for the note from the PR description:

```
Note: since the state of the external secret cannot be verified, even keeping the secret name unchanged, requires permission check.
```

I think this code deserves a comment with similar description.

chiragkyal: The function godoc has a similar description. Let me know if we need to include anything else. 

```go
// Note: If (newer/updated) route uses externalCertificate, this function always returns true, as we cannot definitively verify if
// the content of the referenced secret has been modified. Even if the secret name remains the same,
// we must assume that the secret content is changed, necessitating authorization.
```

pkg/route/hostassignment/assignment_test.go
chiragkyal: These tests are unrelated to the changes I made related to externalCertificate; I just wanted to enhance different test scenarios.

chiragkyal:
/cc @Miciah 
for approval

pkg/route/hostassignment/externalcertificate.go
```diff
@@ -1,35 +0,0 @@
...
-	}
-
-	newTLS := new.Spec.TLS
-	oldTLS := older.Spec.TLS
-	if (newTLS != nil && newTLS.ExternalCertificate != nil) || (oldTLS != nil && oldTLS.ExternalCertificate != nil) {
-		return routecommon.CheckRouteCustomHostSAR(ctx, field.NewPath("spec", "tls", "externalCertificate"), sarc)
-	}
-
-	return nil
-}
```
Miciah, line 35: Since `ValidateHostExternalCertificate` is used in openshift-apiserver, removing its definition now could cause difficulties if someone else needs a change from library-go in openshift-apiserver and tries to bump it before you update openshift-apiserver not to use `ValidateHostExternalCertificate`.  That is:

1. You remove `ValidateHostExternalCertificate` from library-go.  
2. You don't have time to post the openshift-apiserver PR right away, or reviews get delayed.  
3. Meanwhile, someone else adds function xyz to library-go and bumps openshift-apiserver to get xyz.
4. Building openshift-apiserver fails because `ValidateHostExternalCertificate` is not defined.  

It would be safer to do the following:

1. Update logic in library-go, but keep `ValidateHostExternalCertificate` (maybe add a `// Deprecated` comment) so library-go can be bumped in openshift-apiserver without breaking anything.  
5. Update code in openshift-apiserver not to use `ValidateHostExternalCertificate`.
6. Remove `ValidateHostExternalCertificate` from library-go.

Miciah:
Need to resolve https://github.com/openshift/enhancements/pull/1652#discussion_r1770751465.

pkg/route/hostassignment/externalcertificate.go
chiragkyal: Thanks for the suggestion, and I agree that we should be doing the way you suggested above. 

However, even if we keep `ValidateHostExternalCertificate` definition, the [unit tests present in the openshift-apiserver](https://github.com/openshift/openshift-apiserver/blob/cb676917bae9003e14277684f1606d12b09bf484/pkg/route/apiserver/registry/route/strategy_test.go#L153) will complain due to the changes in `validateTLSExternalCertificate` function because we are dropping `custom-host` permission check from that function as well.  There are fair amount of unit tests overlap between `openshift-apiserver` and library-go, so updating logic in library-go requires immediate intervention (bump) in `openshift-apiserver` as well. 

If these weren't any unit tests in `openshift-apiserver`, then it could have been much easier.

PR #1756, opened 2024-07-07
Fix for issue 1775: in ldapquery, validation out of bounds query should be case insensitive as ldap is.

germanparente:
Any compare operation in ldap should be case insensitive. 

This is to fix a well known issue in oc groups sync command that is comparing base dn to a dn and showing out of bounds if the compare operation fails.


germanparente:
I will do. Thanks for reviewing

dusk125:
/lgtm
Passing to @deads2k for final review

PR #1760, opened 2024-07-09
STOR-1767: Support deletion of ValidatingWebhookConfiguration

mpatlasov:
The [PR](https://github.com/openshift/vmware-vsphere-csi-driver-operator/pull/233) needs to delete static resources conditionally, including [webhook/configuration.yaml](https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/master/assets/webhook/configuration.yaml) which is of `ValidatingWebhookConfiguration` type.

`ApplyDirectly()` from [generic.go](https://github.com/mpatlasov/library-go/blob/master/pkg/operator/resource/resourceapply/generic.go) already knows how to apply this type. This PR lets it delete it in `DeleteAll()`.

### DIFF
--- a/pkg/operator/resource/resourceapply/admissionregistration.go
+++ b/pkg/operator/resource/resourceapply/admissionregistration.go
@@ -152,6 +152,18 @@ func ApplyValidatingWebhookConfigurationImproved(ctx context.Context, client adm
 	return actual, true, nil
 }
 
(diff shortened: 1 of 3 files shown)


mpatlasov:
@soltysh , @sttts , Can you /approve please? This is a simple fix blocking me on "Provide API to disable storage on vSphere" epic. Thanks!

PR #1763, opened 2024-07-12
[POC] certs: simplified test scenarios for a race when multiple CertRotationControllers are used

p0lyn0mial:
This PR adds a series of unit tests that show/prove where the race condition is.

The description of the problem can be found at https://docs.google.com/document/d/1iOy-qYFj2mXnDxLPbkPic7ylNMf1QY6YctnUFjHOj14/edit


pkg/operator/certrotation/target_test.go
```diff
@@ -631,3 +636,217 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+		EventRecorder: events.NewInMemoryRecorder("test"),
+	}
+
+	firstSigner, signerUpdated, err := target.EnsureSigningCertKeyPair(context.TODO())
+	require.NoError(t, err)
+	require.True(t, signerUpdated)
+
+	secondSigner, signerUpdated, err := target.EnsureSigningCertKeyPair(context.TODO())
+	require.NoError(t, err)
+	require.True(t, signerUpdated)
```
vrutkovs, line 847: Why would we expect that signer would be updated here?

pkg/operator/certrotation/target_test.go
```diff
@@ -631,3 +636,217 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+		Validity:      24 * time.Hour,
+		Refresh:       12 * time.Hour,
+		Client:        fakeKubeClient.CoreV1(),
+		Lister:        corev1listers.NewSecretLister(coreV1Indexer),
+		EventRecorder: events.NewInMemoryRecorder("test"),
+	}
+
+	firstSigner, signerUpdated, err := target.EnsureSigningCertKeyPair(context.TODO())
+	require.NoError(t, err)
+	require.True(t, signerUpdated)
```
vrutkovs, line 843: We don't cache signer here. If a proper caching is added this will no longer generate different commonName signer:
```
	signerSecret := &corev1.Secret{
			Name:      "signer",
			Namespace: "ns",
		},
	}
	ensureMetadataUpdate(signerSecret, signer.Owner, signer.AdditionalAnnotations)
	err = setSigningCertKeyPairSecret(signerSecret, signer.Validity)
	require.NoError(t, err)
	coreV1Indexer.Add(signerSecret)
```

pkg/operator/certrotation/target_test.go
```diff
@@ -631,3 +636,217 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+	firstSignerSecret := &corev1.Secret{}
+	require.NoError(t, setSigningCertKeyPairSecret(firstSignerSecret, time.Hour))
+
+	// the secondSignerSecret would be generated by the secondCertController
+	secondSignerSecret := &corev1.Secret{}
+	require.NoError(t, setSigningCertKeyPairSecret(secondSignerSecret, time.Hour))
+
+	firstSecretCommonName := firstSignerSecret.Annotations[CertificateIssuer]
+	secondSecretCommonName := secondSignerSecret.Annotations[CertificateIssuer]
+	require.Equal(t, firstSecretCommonName, secondSecretCommonName)
```
vrutkovs, line 672: This relies on common name to include unix timestamp, so it would pass in the unit test, but not real run

pkg/operator/certrotation/target_test.go
```diff
@@ -631,3 +636,217 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+	}
+
+	expectedCertificates := []*x509.Certificate{}
+
+	res, err := target.EnsureConfigMapCABundle(context.TODO(), firstSigner)
+	require.NoError(t, err)
+	_, err = fakeKubeClient.CoreV1().ConfigMaps("ns").Get(context.Background(), "trust-bundle", metav1.GetOptions{})
+	require.NoError(t, err)
+	// by not adding the caBundleConfigMap
+	// to the lister, we simulate the race condition
```
vrutkovs, line 748: This isn't correct. Not adding it to a lister will make sure it would be created from scratch, which is different from "racing update" as it doesn't have necessary metadata

pkg/operator/certrotation/target_test.go
```diff
@@ -631,3 +636,217 @@ func TestEnsureTargetSignerCertKeyPair(t *testing.T) {
...
+	res, err := target.EnsureConfigMapCABundle(context.TODO(), firstSigner)
+	require.NoError(t, err)
+	require.Equal(t, expectedCertificates, res)
+	caBundleConfigMap, err := fakeKubeClient.CoreV1().ConfigMaps("ns").Get(context.Background(), "trust-bundle", metav1.GetOptions{})
+	require.NoError(t, err)
+	coreV1Indexer.Add(caBundleConfigMap)
+
+	res, err = target.EnsureConfigMapCABundle(context.TODO(), secondSigner)
+	require.NoError(t, err)
+	// by not adding the raBundleConfigMap
```
vrutkovs, line 807: I don't understand what would be the fix here. There is no way to fix `EnsureConfigMapCABundle` to make it pass as we intentionally ignore result of this function call

p0lyn0mial:
These tests are simplified. I don't intend to merge this PR. 
The tests here helped me understood where the race condition is as described in the linked docs.

pkg/operator/certrotation/target_test.go
vrutkovs: Right, what I'm saying is this test is not really indicative for the fix, the actual solution would prevent this situation from happening - so later on there is no way to fix the test

p0lyn0mial: I don't intend to merge this PR. The tests here helped me understood where the race condition is as described in the linked docs.

p0lyn0mial: > the actual solution would prevent this situation from happening - so later on there is no way to fix the test

this is not true, switching from `apply` to `update` will not change the function that generates the common name.

vrutkovs: Confusingly we don't do `apply` in `applySecret`, its using `Update` - https://github.com/openshift/library-go/blob/e0aa70d556789b3ff5e9af85e838320b3063aa14/pkg/operator/resource/resourceapply/core.go#L458

We can do update it to use `get with specific revision` but I'm struggling to write a unit test for it

p0lyn0mial: We don't need to pass an additional `RV`. The RV is read from the resource. The optimistic lock is built-in. Give it a try.
Also switch the CABundle to use update semantic instead of apply.

vrutkovs: >We don't need to pass an additional RV. The RV is read from the resource Give it a try.

In that case just using "optimistic lock" is insufficient. 

CABundle uses `applyConfigMap` which is also Update: https://github.com/vrutkovs/library-go/blob/f8b331e2b5bcb147092181769e1094494f663574/pkg/operator/resource/resourceapply/core.go#L348

p0lyn0mial: the apply function is more like overwrite, we need to change the controller/struct/type to issue an update for a resource it has currently reconciled.

vrutkovs: >we need to change the controller/struct/type to issue an update for a resource it has currently reconciled

This is what its doing now, isn't it? I don't quite understand what needs fixing and why the bug resurfaces for this controller only - it sounds like it could have happened across any library-go user but it didn't?

In any case, we have an alternative solution which has unittests and passing e2e tests - https://github.com/openshift/library-go/pull/1722

p0lyn0mial: > I don't quite understand what needs fixing and why the bug resurfaces for this controller only

Have a look at the doc linked in the description of this PR. It contains more details about the possible race condition.

> In any case, we have an alternative solution which has unittests and passing e2e tests - https://github.com/openshift/library-go/pull/1722

#1722 is not going to solve the issue because the controllers run across multiple processes. 

@vrutkovs why don't you sync with @deads2k or @soltysh on #1722 and ask them for their feedback ?

vrutkovs: #1722 fixes the thread level and https://github.com/openshift/cluster-kube-apiserver-operator/pull/1669 will fix process level races

PR #1764, opened 2024-07-22
continue when create err

lance5890:
address https://github.com/openshift/library-go/issues/1765


dusk125:
@lance5890 can you add a test case that shows the panic is avoided please?

PR #1766, opened 2024-07-22
OCPBUGS-37274: Replace FIPS_ENABLED string in assets for csi-driver

mpatlasov:
AWS-EFS CSi driver needs to know whether FIPS is enabled: https://github.com/openshift/aws-efs-csi-driver/pull/80 The operator sets FIPS_ENABLED in assets: https://github.com/openshift/aws-efs-csi-driver-operator/pull/144 , https://github.com/openshift/csi-operator/pull/245 This patch makes sure that the string `$(FIPS_ENABLED)` is substituted with true or false dependently on `/proc/sys/crypto/fips_enabled`. Other users of library-go must not notice this change because they do not use FIPS_ENABLED in asset yamls.

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
@@ -215,6 +215,14 @@ func WithPlaceholdersHook(configInformer configinformers.SharedInformerFactory) 
 		logLevel := loglevel.LogLevelToVerbosity(spec.LogLevel)
 		pairs = append(pairs, []string{"${LOG_LEVEL}", strconv.Itoa(logLevel)}...)
 
+		// FIPS
+		fipsEnabled := "\"false\""
+		contents, err := os.ReadFile("/proc/sys/crypto/fips_enabled")
+		if err == nil && string(contents) == "1\n" {
+			fipsEnabled = "\"true\""
+		}
+		pairs = append(pairs, []string{"${FIPS_ENABLED}", fipsEnabled}...)
+
 		replaced := strings.NewReplacer(pairs...).Replace(string(manifest))
 		return []byte(replaced), nil
 	}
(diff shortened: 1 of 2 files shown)


mpatlasov:
/hold

(Let's try to achieve the same goal without change in library go: https://github.com/openshift/csi-operator/pull/249)

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -215,6 +215,14 @@ func WithPlaceholdersHook(configInformer configinformers.SharedInformerFactory)
...
 		pairs = append(pairs, []string{"${LOG_LEVEL}", strconv.Itoa(logLevel)}...)
 
+		// FIPS
+		fipsEnabled := "\"false\""
+		contents, err := os.ReadFile("/proc/sys/crypto/fips_enabled")
+		if err == nil && string(contents) == "1\n" {
+			fipsEnabled = "\"true\""
+		}
+		pairs = append(pairs, []string{"${FIPS_ENABLED}", fipsEnabled}...)
+
```
bertinatto, line 225: Since this is specific to EFS (so far), have you considered having this in the operator repo instead?

Here's an EBS example [1] used as a DeploymentHook [2]. If we really want to use a ManifestHook [3], we can plumb them through here [4].


[1] https://github.com/bertinatto/aws-ebs-csi-driver-operator-1/blob/10637f68e0e7a3da1b37781ff8f5bc68022dbf5f/pkg/operator/starter.go#L523-L530
[2] https://github.com/bertinatto/aws-ebs-csi-driver-operator-1/blob/10637f68e0e7a3da1b37781ff8f5bc68022dbf5f/vendor/github.com/openshift/library-go/pkg/operator/deploymentcontroller/deployment_controller.go#L27
[3] https://github.com/bertinatto/aws-ebs-csi-driver-operator-1/blob/10637f68e0e7a3da1b37781ff8f5bc68022dbf5f/vendor/github.com/openshift/library-go/pkg/operator/deploymentcontroller/deployment_controller.go#L31
[4] https://github.com/bertinatto/aws-ebs-csi-driver-operator-1/blob/10637f68e0e7a3da1b37781ff8f5bc68022dbf5f/vendor/github.com/openshift/library-go/pkg/operator/csi/csicontrollerset/csi_controller_set.go#L183

mpatlasov:
https://github.com/openshift/csi-operator/pull/249 merged. Closing this PR...

PR #1705, opened 2024-03-29
use update only secret-apply for cert rotation logic

tkashem:

PR #1772, opened 2024-08-06
OCPBUGS-38335: certrotation: avoid using applySecret to catch racy cert updates

vrutkovs:
`applySecret`/`applyConfigMap` does `Get` initially, so instead of using the version we got from informer it refreshes it and may overwrite if other changes were done in parallel.

Instead of using applySecret/applyConfigMap we should be invoking Create/Update here directly. If the informer has outdated version its update would be rejected by etcd and sync function would be restarted


pkg/operator/certrotation/cabundle.go
```diff
@@ -88,12 +87,16 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 
 	if modified {
-
-		actualCABundleConfigMap, updated, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		var actualCABundleConfigMap *corev1.ConfigMap
+		if len(caBundleConfigMap.ResourceVersion) == 0 {
+			actualCABundleConfigMap, err = c.Client.ConfigMaps(c.Namespace).Create(ctx, caBundleConfigMap, metav1.CreateOptions{})
+		} else {
+			actualCABundleConfigMap, err = c.Client.ConfigMaps(c.Namespace).Update(ctx, caBundleConfigMap.DeepCopy(), metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
p0lyn0mial: Reporting an event on successful update is a good idea.

We don't need to call `DeepCopy()` here, please remove it.

pkg/operator/certrotation/cabundle.go
```diff
@@ -88,12 +87,16 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
...
+		}
 		if err != nil {
 			return nil, err
 		}
-		if updated {
+		if equality.Semantic.DeepEqual(actualCABundleConfigMap.Data, caBundleConfigMap.Data) {
```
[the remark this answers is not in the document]
p0lyn0mial: Why do we need this additional check anyway ? (line `89`)
We already know that the update was required. So, if the call succeeded then we can simply report success.

pkg/operator/certrotation/cabundle.go
```diff
@@ -88,12 +87,16 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 
 	if modified {
-
-		actualCABundleConfigMap, updated, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		var actualCABundleConfigMap *corev1.ConfigMap
+		if len(caBundleConfigMap.ResourceVersion) == 0 {
+			actualCABundleConfigMap, err = c.Client.ConfigMaps(c.Namespace).Create(ctx, caBundleConfigMap, metav1.CreateOptions{})
```
p0lyn0mial, line 92: Given that we would like to report successful creation as well why not change the code to be more explicit.
For example:
1. we would introduce a new `var creationRequired bool` or `var creationNeeded bool` variable.
2. we would set the new variable at[ line 60](https://github.com/openshift/library-go/blob/d3df6775ec935973df3fdc6c231147865e4bad6a/pkg/operator/certrotation/cabundle.go#L60) 
3. we would rename `modified` to `updateRequired` or `updateNeeded`
4. we would introduce a branch before line 89, for example:
```
if creationRequired {
 create
 report
 assign to `caBundleConfigMap`
 } else if updateRequired {
 update 
 report
 assign to `caBundleConfigMap`
 }
 ```

Comment (p0lyn0mial) [reply pkg/operator/certrotation/cabundle.go:92]: Alternatively we could move the creation to [line 53](https://github.com/openshift/library-go/blob/d3df6775ec935973df3fdc6c231147865e4bad6a/pkg/operator/certrotation/cabundle.go#L53).

The creation is a rare operation and immediately after creation the sync loop will be triggered and we can issue an update.

File: pkg/operator/certrotation/signer.go [hunk 4]
```diff
@@ -51,12 +51,6 @@ type RotatedSigningCASecret struct {
 	Lister        corev1listers.SecretLister
 	Client        corev1client.SecretsGetter
 	EventRecorder events.Recorder
-
-	// Deprecated: DO NOT enable, it is intended as a short term hack for a very specific use case,
-	// and it works in tandem with a particular carry patch applied to the openshift kube-apiserver.
-	// we will remove this when we migrate all of the affected secret
-	// objects to their intended type: https://issues.redhat.com/browse/API-1800
-	UseSecretUpdateOnly bool
```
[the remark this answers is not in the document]
p0lyn0mial: +1, we should cleanup and remove unused code like `ApplySecretDoNotUse`.

pkg/operator/certrotation/cabundle.go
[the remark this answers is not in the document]
vrutkovs: Done, removed

pkg/operator/certrotation/cabundle.go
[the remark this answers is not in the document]
vrutkovs: Good idea, we already have `modified`, so lets use it to emit the event

pkg/operator/certrotation/signer.go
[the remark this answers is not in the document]
vrutkovs: Removed

pkg/operator/certrotation/cabundle.go
vrutkovs: Good idea, now all functions should have `exists` and `modified`

pkg/operator/resource/resourceapply/core.go
```diff
@@ -88,14 +88,6 @@ func ApplySecret(ctx context.Context, client coreclientv1.SecretsGetter, recorde
 	return applySecretImproved(ctx, client, recorder, required, noCache, false)
 }
 
-// ApplySecretDoNotUse is depreated and will be removed
-// Deprecated: DO NOT USE, it is intended as a short term hack for a very specific use case,
-// and it works in tandem with a particular carry patch applied to the openshift kube-apiserver.
-// Use ApplySecret instead.
-func ApplySecretDoNotUse(ctx context.Context, client coreclientv1.SecretsGetter, recorder events.Recorder, required *corev1.Secret) (*corev1.Secret, bool, error) {
```
p0lyn0mial, line 95: The original [PR](https://github.com/openshift/library-go/pull/1705/files#diff-81c137f4d13fefcae5735d1cf0e4aaaf14c96b3e4931c1fd1a242ae1a25a5489) introduced more changes. Not all were remove. Please open a new PR that reverts https://github.com/openshift/library-go/pull/1705.

pkg/operator/certrotation/cabundle.go
p0lyn0mial: we don't want to report https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/event_helpers.go#L25 anymore ?

pkg/operator/certrotation/cabundle.go
```diff
@@ -88,14 +89,16 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 
 	if modified {
-
-		actualCABundleConfigMap, updated, err := resourceapply.ApplyConfigMap(ctx, c.Client, c.EventRecorder, caBundleConfigMap)
+		var actualCABundleConfigMap *corev1.ConfigMap
+		if !exists {
+			actualCABundleConfigMap, err = c.Client.ConfigMaps(c.Namespace).Create(ctx, caBundleConfigMap, metav1.CreateOptions{})
+		} else {
+			actualCABundleConfigMap, err = c.Client.ConfigMaps(c.Namespace).Update(ctx, caBundleConfigMap, metav1.UpdateOptions{})
```
p0lyn0mial, line 96: why don't want to report https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/event_helpers.go#L34 anymore ?

vrutkovs: Good point, we get an event that its about to be updated, but not the update result. Added this to the latest commit

pkg/operator/certrotation/cabundle.go
vrutkovs: Fixed

pkg/operator/certrotation/signer.go
```diff
@@ -100,7 +101,16 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
...
+		var actualSigningCertKeyPairSecret *corev1.Secret
+		var action string
+		if !exists {
+			actualSigningCertKeyPairSecret, err = c.Client.Secrets(c.Namespace).Create(ctx, signingCertKeyPairSecret, metav1.CreateOptions{})
+			action = "Create"
+		} else {
+			actualSigningCertKeyPairSecret, err = c.Client.Secrets(c.Namespace).Update(ctx, signingCertKeyPairSecret, metav1.UpdateOptions{})
+			action = "Update"
+		}
+		emitEvent(c.EventRecorder, "Secret", action, resourcehelper.FormatResourceForCLIWithNamespace(actualSigningCertKeyPairSecret), err)
```
bertinatto, line 113: If you do https://github.com/openshift/library-go/pull/1772/files#r1714300480, this would get shorter:

```go
var actualSigningCertKeyPairSecret *corev1.Secret
if !exists {
	actualSigningCertKeyPairSecret, err = c.Client.Secrets(c.Namespace).Create(ctx, signingCertKeyPairSecret, metav1.CreateOptions{})
	emitEvent(c.EventRecorder, "Create", actualSigningCertKeyPairSecret, err)
} else {
	actualSigningCertKeyPairSecret, err = c.Client.Secrets(c.Namespace).Update(ctx, signingCertKeyPairSecret, metav1.UpdateOptions{})
	emitEvent(c.EventRecorder, "Update", actualSigningCertKeyPairSecret, err)
}

```

pkg/operator/certrotation/target.go
```diff
@@ -134,7 +135,16 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 	}
 
 	if modified {
-		actualTargetCertKeyPairSecret, _, err := resourceapply.ApplySecret(ctx, c.Client, c.EventRecorder, targetCertKeyPairSecret)
+		var actualTargetCertKeyPairSecret *corev1.Secret
+		var action string
+		if len(targetCertKeyPairSecret.ResourceVersion) == 0 {
```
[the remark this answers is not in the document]
vrutkovs: Fixed

pkg/operator/certrotation/event.go
```diff
@@ -0,0 +1,19 @@
+package certrotation
+
+import (
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+func emitEvent(recorder events.Recorder, objType string, action string, obj string, err error) {
```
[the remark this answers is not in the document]
vrutkovs: Good idea. Also using `FormatResourceForCLI` instead of namespaced version as events are namespaced anyway

pkg/operator/certrotation/signer.go
vrutkovs: to minimize copying important params I introduced "action" which is the only variable there

bertinatto (commented):
While I'm not familiar with these controllers, I did my best to review this, and to the best of my knowledge, these changes appear to be correct.

/lgtm

pkg/operator/certrotation/event.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func emitEvent(recorder events.Recorder, action string, obj runtime.Object, err error) {
```
p0lyn0mial, line 11: Instead of this new function I would prefer to reuse [reportCreateEvent](https://github.com/openshift/library-go/blob/6ca0621840e45bed8686978644512dc3320fe5c7/pkg/operator/resource/resourceapply/event_helpers.go#L25C6-L25C23) and [reportUpdateEvent](https://github.com/openshift/library-go/blob/6ca0621840e45bed8686978644512dc3320fe5c7/pkg/operator/resource/resourceapply/event_helpers.go#L34C6-L34C23).

I think we could open a new PR in which we would:
1. crate a new `helpers` pkg in `library-go/pkg/operator/resource/helpers`
2. we would move `operator/resource/resourceapply/event_helpers.go` to the new pkg (with tests)
3. we would export all `report` methods.

Then we could remove this file and use the new exported methods directly. WDYT?

pkg/operator/certrotation/cabundle.go
```diff
@@ -88,14 +89,20 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	}
 
 	if modified {
```
p0lyn0mial, line 91: This code is a bit hard to read. I would like to make it more explicit. Given that we would export the event reporting methods. We could do:

```
	if creationRequired {
		actualCABundleConfigMap, err := c.Client.ConfigMaps(c.Namespace).Create(ctx, caBundleConfigMap, metav1.CreateOptions{})
		resourcehelpers.ReportCreateEvent(c.EventRecorder, actualCABundleConfigMap, err)
		if err != nil {
			return nil, err
		}
		klog.V(2).Infof("Created ca-bundle.crt configmap %s/%s with:\n%s", certs.CertificateBundleToString(updatedCerts), caBundleConfigMap.Namespace, caBundleConfigMap.Name)
		caBundleConfigMap = actualCABundleConfigMap
	} else if updateRequired {
		actualCABundleConfigMap, err := c.Client.ConfigMaps(c.Namespace).Update(ctx, caBundleConfigMap, metav1.UpdateOptions{})
		resourcehelpers.ReportUpdateEvent(c.EventRecorder, actualCABundleConfigMap, err)
		if err != nil {
			return nil, err
		}
		klog.V(2).Infof("Updated ca-bundle.crt configmap %s/%s with:\n%s", certs.CertificateBundleToString(updatedCerts), caBundleConfigMap.Namespace, caBundleConfigMap.Name)
		caBundleConfigMap = actualCABundleConfigMap
	}
```

WDYT ?

pkg/operator/certrotation/cabundle.go
```diff
@@ -45,6 +44,7 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
 	// doesn't have any expired certs
 	modified := false
+	exists := true
```
p0lyn0mial, line 47: please rename to `creationRequired`

pkg/operator/certrotation/cabundle.go
```diff
@@ -45,6 +44,7 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 	// by this point we have current signing cert/key pair.  We now need to make sure that the ca-bundle configmap has this cert and
 	// doesn't have any expired certs
 	modified := false
```
p0lyn0mial, line 46: please rename to `updateRequired`

pkg/operator/certrotation/cabundle.go
```diff
@@ -59,6 +59,7 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 			c.AdditionalAnnotations,
 		)}
 		modified = true
```
p0lyn0mial, line 61: give the code below we could remove it.

pkg/operator/certrotation/cabundle.go
```diff
@@ -59,6 +59,7 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
...
+		exists = false
```
p0lyn0mial, line 62: here, we would set `creationRequired` to `true`

pkg/operator/certrotation/cabundle.go
```diff
@@ -59,6 +59,7 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
...
 	}
 
 	needsOwnerUpdate := false
```
p0lyn0mial, line 65: and we could modify line `70` to: 

`updateRequired = needsOwnerUpdate || needsMetadataUpdate`

pkg/operator/certrotation/signer.go
```diff
@@ -95,15 +96,25 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 
 		LabelAsManagedSecret(signingCertKeyPairSecret, CertificateTypeSigner)
 
-		modified = true
+		updateRequired = true
 		signerUpdated = true
```
p0lyn0mial, line 100: do we really need this var ?

I think it could be replaced by `creationRequired || updateRequired`, right ?

pkg/operator/certrotation/target.go
```diff
@@ -108,8 +110,8 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 			),
 			Type: corev1.SecretTypeTLS,
 		}
-		modified = true
-		secretDoesntExist = true
+		updateRequired = true
```
p0lyn0mial, line 113: is this needed here ?

pkg/operator/certrotation/signer.go
vrutkovs: Yes. `updateRequired` means "any part of the secret has changed", but `signerUpdated` is specific to certificate part (we need to pass it on to trigger CA bundle / target cert updates)

p0lyn0mial: sure, I think that the following code:
```
	// at this point, the secret has the correct signer, so we should read that signer to be able to sign
	signingCertKeyPair, err := crypto.GetCAFromBytes(signingCertKeyPairSecret.Data["tls.crt"], signingCertKeyPairSecret.Data["tls.key"])
	if err != nil {
		return nil, signerUpdated, err
	}

	return signingCertKeyPair, signerUpdated, nil
```

could be replaced be:
```
	// at this point, the secret has the correct signer, so we should read that signer to be able to sign
	signingCertKeyPair, err := crypto.GetCAFromBytes(signingCertKeyPairSecret.Data["tls.crt"], signingCertKeyPairSecret.Data["tls.key"])
	if err != nil {
		return nil, creationRequired || updateRequired, err
	}

	return signingCertKeyPair, creationRequired || updateRequired, nil
```

and we could remove `signerUpdated`, am I right ?

p0lyn0mial:
@vrutkovs could you also open a proof PR in `kas-o` so that we get a CI signal?

vrutkovs:
Opened https://github.com/openshift/cluster-kube-apiserver-operator/pull/1721

pkg/operator/certrotation/signer.go
p0lyn0mial: ok, do you also want to move the refactor commit to a new PR ?

p0lyn0mial:
> Opened https://github.com/openshift/cluster-kube-apiserver-operator/pull/1721

the pr was created one week ago and e2e tests haven't been executed. Could you address https://github.com/openshift/library-go/pull/1772#discussion_r1716408279 and then pull this pr into 1721 and kick off the tests?

pkg/operator/certrotation/target.go
vrutkovs: Oh, right, no longer needed

pkg/operator/certrotation/signer.go
vrutkovs: Lets leave this version as is and experiment if `signerUpdated` can be removed in a separate PR.

I think we want more specific messages for signer/cabundle/target too, i.e. `Updated contents` and `Updated metadata` probably

p0lyn0mial:
/lgtm

let's wait for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1721 before merging.

pkg/operator/certrotation/target.go
```diff
@@ -108,44 +109,46 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
...
-			return nil, err
-		}
-		targetCertKeyPairSecret = actualTargetCertKeyPairSecret
-	}
+	needsMetadataUpdate := ensureMetadataUpdate(targetCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
+	needsTypeChange := ensureSecretTLSTypeSet(targetCertKeyPairSecret)
+	updateRequired = needsMetadataUpdate || needsTypeChange
 
-	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired, secretDoesntExist); len(reason) > 0 {
+	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired, creationRequired); len(reason) > 0 {
```
p0lyn0mial, line 119: does `c.CertCreator.NeedNewTargetCertKeyPair` persist any resource in etcd (does it use `Apply*` method?)

vrutkovs: No, we don't even pass `client` there. It decides on the contents of the secret whether it needs to be regenerated

p0lyn0mial: ok, thanks.

pkg/operator/certrotation/signer.go
```diff
@@ -74,18 +75,18 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 			),
 			Type: corev1.SecretTypeTLS,
 		}
-		signerExists = false
+		creationRequired = true
 	}
 
 	// apply necessary metadata (possibly via delete+recreate) if secret exists
```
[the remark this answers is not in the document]
p0lyn0mial: ping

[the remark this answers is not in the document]
vrutkovs: Fixed

p0lyn0mial:
/lgtm

let's wait for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1721 before merging.

soltysh (approved):
/approve

PR #1778, opened 2024-08-20
SPLAT-1772: drop vars removed in api#1780

rvanderp3:
Dropped variables which are no longer present in openshift/api.  These variables do not appear to be used elsewhere, but their presence in library-go causes api bumps in other repos which use library-go to fail.


rvanderp3:
/hold

hold while tests run

PR #1780, opened 2024-08-22
audit/manifests: log requests to Events

dgrisonnet:
Re-enable audit logs about Events to be able to identify misconfigured clients.

### DIFF
--- a/pkg/operator/apiserver/audit/manifests/base-policy.yaml
+++ b/pkg/operator/apiserver/audit/manifests/base-policy.yaml
@@ -6,11 +6,6 @@
     omitStages:
     - "RequestReceived"
     rules:
-    # Don't log requests for events
-    - level: None
-      resources:
-      - group: ""
-        resources: ["events"]
     # Don't log authenticated requests to certain non-resource URL paths.
     - level: None
       userGroups: ["system:authenticated", "system:unauthenticated"]
--- a/pkg/operator/apiserver/audit/testdata/allrequestbodies.yaml
+++ b/pkg/operator/apiserver/audit/testdata/allrequestbodies.yaml
@@ -8,11 +8,6 @@ omitManagedFields: true
 omitStages:
 - "RequestReceived"
 rules:
-# Don't log requests for events
-- level: None
-  resources:
-  - group: ""
-    resources: ["events"]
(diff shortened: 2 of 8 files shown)


pkg/operator/apiserver/audit/manifests/base-policy.yaml
```diff
@@ -6,8 +6,11 @@
     omitStages:
     - "RequestReceived"
     rules:
-    # Don't log requests for events
-    - level: None
+    # Log events at the metadata level. This is necessary to identify
+    # misconfigured clients. They could either spam mutating requests on Events
+    # or list all the Events in the cluster which are two operations that can
+    # lead to disruption.
+    - level: Metadata
```
p0lyn0mial, line 13: I think we could simply remove it. There is [a catch-all rule](https://github.com/openshift/library-go/blob/8052e81a61efbeabcd1ba0e660d9777a93fc6859/pkg/operator/apiserver/audit/manifests/default-rules.yaml#L10) which would be applied

pkg/operator/apiserver/audit/testdata/allrequestbodies.yaml
```diff
@@ -8,8 +8,10 @@ omitManagedFields: true
 omitStages:
 - "RequestReceived"
 rules:
-# Don't log requests for events
-- level: None
+# Log events at the metadata level. This is necessary to identify misconfigured
+# clients. They could either spam mutating requests on Events or list all the
+# Events in the cluster which are two operations that can lead to disruption.
+- level: Metadata
```
p0lyn0mial, line 14: I would also remove it. If a user requested all request bodies, let's give them what they want.

p0lyn0mial (commented):
It would be useful to know the impact on storage, CPU, and memory in large clusters after enabling event logging. I understand that we are confident the impact will be negligible.

Could you also open a proof PR for kas and oas and manually validate if the policy was changed ?

dinhxuanvu (commented):
/lgtm

p0lyn0mial:
/approve

/hold 
in case you would like to test something.

dgrisonnet:
Verified on the proof PR.

/unhold

PR #1783, opened 2024-08-30
share logic for getting the in-use MachineConfigs

tremes:
This is moving the logic from the https://github.com/openshift/oc/blob/master/pkg/cli/admin/prune/renderedmachineconfigs/renderedmachineconfigs.go#L397 so that it can be reused in the Insights Operator.


tremes:
@wking can you please review and help me with some reviewers and approvers suggestion for this new package?

pkg/machineconfig/in_use_machineconfigs.go
```diff
@@ -0,0 +1,58 @@
...
+
+	mcfgclientset "github.com/openshift/client-go/machineconfiguration/clientset/versioned"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// GetInUseMachineConfigs filters in-use MachineConfig resources and returns set of their names.
+func GetInUseMachineConfigs(ctx context.Context, clientConfig *rest.Config, poolFilter string) (sets.Set[string], error) {
```
tremes, line 15: Optionally I can pass the client instances (instead of the config). I think it would allow for some testing.

PR #1786, opened 2024-08-30
Add support for deletion of deployment and daemonset via DeleteAll fu…

gnufied:
…nction

### DIFF
--- a/pkg/operator/resource/resourceapply/apps_test.go
+++ b/pkg/operator/resource/resourceapply/apps_test.go
@@ -134,6 +134,77 @@ func TestApplyDeployment(t *testing.T) {
 	}
 }
 
+func TestDeleteDeployment(t *testing.T) {
+	tests := []struct {
+		name              string
+		desiredDeployment *appsv1.Deployment
+		expectError       bool
+	}{
+		{
+			name:              "when deployment exists",
+			desiredDeployment: workload(),
+			expectError:       false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			eventRecorder := events.NewInMemoryRecorder("")
+			fakeKubeClient := fake.NewSimpleClientset()
+			if tt.desiredDeployment != nil {
+				fakeKubeClient = fake.NewSimpleClientset(tt.desiredDeployment)
+			}
+			_, _, err := resourceapply.ApplyDeployment(context.TODO(), fakeKubeClient.AppsV1(), eventRecorder, tt.desiredDeployment, 0)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			_, deletedFlag, err := resourceapply.DeleteDeployment(context.TODO(), fakeKubeClient.AppsV1(), eventRecorder, tt.desiredDeployment)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			if !deletedFlag {
+				t.Fatalf("expected deployment to be deleted but failed")
+			}
+		})
+	}
+}
+
+func TestDeleteDaemonSet(t *testing.T) {
+	tests := []struct {
+		name             string
+		desiredDaemonSet *appsv1.DaemonSet
+		expectError      bool
+	}{
+		{
+			name:             "when daemonset exists",
+			desiredDaemonSet: daemonSet(),
+			expectError:      false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			eventRecorder := events.NewInMemoryRecorder("")
+			fakeKubeClient := fake.NewSimpleClientset()
+			if tt.desiredDaemonSet != nil {
+				fakeKubeClient = fake.NewSimpleClientset(tt.desiredDaemonSet)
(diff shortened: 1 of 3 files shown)


dusk125:
Can you add unit tests please?

gnufied:
@dusk125 done.

dusk125:
Thanks!
/lgtm

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -317,6 +320,20 @@ func DeleteAll(ctx context.Context, clients *ClientHolder, recorder events.Recor
...
+			} else {
+				_, result.Changed, result.Error = DeleteDeployment(ctx, client, recorder, t)
+			}
+		case *appsv1.DaemonSet:
+			client := clients.daemonSetsGetter()
+			if client == nil {
+				result.Error = fmt.Errorf("missing kubeClient")
+			} else {
+				_, result.Changed, result.Error = DeleteDaemonSet(ctx, client, recorder, t)
+			}
```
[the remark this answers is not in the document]
bertinatto: Same comment for `deploymentsGetter()`

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -388,3 +405,17 @@ func (c *ClientHolder) secretsGetter() corev1client.SecretsGetter {
...
+	}
+	return c.kubeClient.AppsV1()
+}
+
+func (c *ClientHolder) daemonSetsGetter() appv1client.DaemonSetsGetter {
+	if c.kubeClient == nil {
+		return nil
+	}
+	return c.kubeClient.AppsV1()
+}
```
bertinatto, line 421: You can likely delete these two methods. They seem to be modeled after `configMapsGetter()` and `secretsGetter()` above, but those functions are designed to return a cached client (preferably), whereas the new ones here always return a non-cached client.

pkg/operator/resource/resourceapply/apps_test.go
```diff
@@ -134,6 +134,77 @@ func TestApplyDeployment(t *testing.T) {
...
+			if tt.desiredDeployment != nil {
+				fakeKubeClient = fake.NewSimpleClientset(tt.desiredDeployment)
+			}
+			_, _, err := resourceapply.ApplyDeployment(context.TODO(), fakeKubeClient.AppsV1(), eventRecorder, tt.desiredDeployment, 0)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+
+			_, deletedFlag, err := resourceapply.DeleteDeployment(context.TODO(), fakeKubeClient.AppsV1(), eventRecorder, tt.desiredDeployment)
+			if err != nil {
```
bertinatto, line 162: Didn't you want to use `expectError` here?

pkg/operator/resource/resourceapply/apps_test.go
```diff
@@ -134,6 +134,77 @@ func TestApplyDeployment(t *testing.T) {
...
+	tests := []struct {
+		name              string
+		desiredDeployment *appsv1.Deployment
+		expectError       bool
+	}{
+		{
+			name:              "when deployment exists",
+			desiredDeployment: workload(),
+			expectError:       false,
+		},
```
bertinatto, line 147: Perhaps a new test case where the object doesn't exist or can't be deleted?

gnufied: I do not see much value in that test case. Quite frankly, I don't see much value in the test I already wrote but I was just checking the checkboxes. :(

pkg/operator/resource/resourceapply/apps_test.go
gnufied: fixed

pkg/operator/resource/resourceapply/generic.go
gnufied: fixed

pkg/operator/resource/resourceapply/apps_test.go
bertinatto: Could you please elaborate on why you don't see value in these tests?

I believe it was a reasonable request from @dusk125 because: 1) the functions are new, 2) the tests are straightforward to implement, and 3) they help prevent unintended changes from future refactoring.

We certainly don't have full coverage for everything we add, but IMO it's a good thing to aim for.

gnufied: Sure,  the current test exercises machinery of fake kubeclients. The code in question itself is mostly logic free and if the contract of - `clients.kubeClient.AppsV1()` holds true then it will work as expected. 

IMO - a unit test should test the code it is changing and shouldn't necessarily be testing api clients themselves. So if I add a unit test, which deletes non-existant object or an object that can't be deleted, we are in-truth testing the API client themselves. There is no special handling for that bit in my code.

gnufied: >It doesn't actually return NotFound errors (lines 251, 263), and that's why I wanted to see a test case for deleting a non-existent object. 

Yeah I know. I am dubious about usefulness of a unit test for that, but I am less sure now. :-)

I have added some unit tests for non-existant object deletion fwiw now.

PR #1789, opened 2024-09-04
WIP: CLID-160: upgrade distribution/distribution to the latest released version

aguidirh:
Currently in oc-mirror it is not possible to upgrade containers/image dependency because of the indirect dependency on library-go. The upgrade of containers/image is required to fix CVE-2024-3727

In order to upgrade the containers/image it is required to upgrade the distribution/distribution version in the library-go. Before this PR the library-go was depending on a non released version of distribution/distribution `v3.0.0-20230511163743-f7717b7855ca`. This PR upgrades the library-go to the [latest v3 release version](https://github.com/distribution/distribution/releases/tag/v3.0.0-beta.1).

Since code required by library-go was removed from the latest released version of distribution/distribution, the required code was copied as a new package (pkg/registry) on the library-go


dinhxuanvu (changes requested):
Please separate vendor changes from code changes for easier review.

dinhxuanvu (commented):
Looking good overall. The `registry` carry is fine with me but need someone else to take a look.
@bertinatto PTAL. Thanks
/lgtm

pkg/registry/client/auth/challenge/authchallenge.go
```diff
@@ -0,0 +1,237 @@
+package challenge
```
ardaguclu, line 1: Why are we moving all these changes inside library-go?. All these packages have been removed entirely from distribution package without moving to somewhere else that we can use?

bertinatto: Right, I think we need a better solution. This is a lot of code that likely won't be maintained and will quickly become outdated, not to mention the potential copyright issues we might be infringing on.

go.mod
```diff
@@ -1,20 +1,22 @@
...
-toolchain go1.22.1
+toolchain go1.23.0
 
 require (
 	github.com/RangelReale/osincli v0.0.0-20160924135400-fababb0555f2
 	github.com/blang/semver/v4 v4.0.0
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
-	github.com/distribution/distribution/v3 v3.0.0-20230511163743-f7717b7855ca
+	github.com/distribution/distribution/v3 v3.0.0-beta.1
+	github.com/distribution/reference v0.6.0
```
bertinatto, line 12: I'd prefer to update to the same version as kube 1.31: https://github.com/kubernetes/kubernetes/blob/release-1.31/go.mod#L26

go.mod
```diff
@@ -1,20 +1,22 @@
...
 
-toolchain go1.22.1
+toolchain go1.23.0
 
 require (
 	github.com/RangelReale/osincli v0.0.0-20160924135400-fababb0555f2
 	github.com/blang/semver/v4 v4.0.0
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
-	github.com/distribution/distribution/v3 v3.0.0-20230511163743-f7717b7855ca
+	github.com/distribution/distribution/v3 v3.0.0-beta.1
```
bertinatto, line 11: This seems to be a pre-release: https://github.com/distribution/distribution/releases/tag/v3.0.0-beta.1

pkg/registry/client/auth/challenge/authchallenge.go
bertinatto: Adding @flavianmissi, who [upgraded to v3](https://github.com/openshift/library-go/pull/1529#issuecomment-1549459055), to the loop.

Flavian, unfortunately distribution v3 broke the API (which is acceptable because  that wasn't an official release). Now we're not sure what to do with the code that uses it. Could you take a look?

aguidirh: @ardaguclu they removed some packages from the latest release (which is acceptable since it is a different major version). That's why I decided to include them here.

@bertinatto do you think we could face copyright issues even if the code is open source? I do not understand very well the licenses tough.

go.mod
aguidirh: @bertinatto are you talking only about `github.com/distribution/reference v0.5.0` ?

go.mod
aguidirh: Yes, but even being a beta, it is better than `v3.0.0-20230511163743-f7717b7855ca` which is not even alpha. It was depending of a random commit sha from the master branch instead of a released (git tag) one.

pkg/registry/client/auth/challenge/authchallenge.go
ardaguclu: > @ardaguclu they removed some packages from the latest release (which is acceptable since it is a different major version). That's why I decided to include them here.

There is no major version bump; we are moving from `github.com/distribution/distribution/v3 v3.0.0-20230511163743-f7717b7855ca` to `github.com/distribution/distribution/v3 v3.0.0-beta.1`. 

>  This is a lot of code that likely won't be maintained and will quickly become outdated
I fully agree with that and I don't think we ever want to import this code and start maintaining it.

ardaguclu: I've found the commit https://github.com/distribution/distribution/pull/4126/commits/cc23fdacffec257f28a040928f379227bad75602 that moves all these packages under internal that would clearly break us.

aguidirh: v3 is a major and that's why they can remove stuff they had in v2 previously. 

Also v3 is still beta so they can break the API as they want to and `v3.0.0-20230511163743-f7717b7855ca` was not even a release (it was a commit sha from master)

aguidirh: I just updated this PR keeping only the minimal required. I removed the following packages from library-go since we can still reference them:

pkg/registry/storage/cache
pkg/registry/storage/cache/memory
pkg/registry/dcontext

The client package was moved to the internal folder in distribution/distribution so we cannot reference it so that's why I kept it here in library-go.

oc-mirror needs this upgrade for several reasons. We know this is an inevitable change so why should we postpone it? The code is done so it is up to you to decide what to do.

cc: @lmzuccarelli @sherine-k

ardaguclu: > oc-mirror needs this upgrade for several reasons.

Could you please be specific about the reasons?. Oc does not require to bump and I'm wondering what additional functionality oc-mirror depends on library-go that requires bumping the distribution in here?

go.mod
bertinatto: @aguidirh correct.

pkg/registry/client/auth/challenge/authchallenge.go
ardaguclu: What about `github.com/containers/image/v5 v5.30.1`?. I tried with the instructions you gave only docker/docker failed which is relatively old version anyway https://github.com/openshift/oc-mirror/blob/41bb263ef632090ed69526593415af68b0fbcc0b/go.mod#L274.

aguidirh: @ardaguclu in order to upgrade docker/docker I will need to upgrade oc in oc-mirror as well. 

After upgrading oc (github.com/openshift/oc v0.0.0-alpha.0.0.20240906182345-ee7076133c38) I started to get the errors below:

```
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:21:2: cannot find module providing package github.com/distribution/distribution/v3/manifest/schema1
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:22:2: cannot find module providing package github.com/distribution/distribution/v3/reference
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:24:2: cannot find module providing package github.com/distribution/distribution/v3/registry/client
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:25:2: cannot find module providing package github.com/distribution/distribution/v3/registry/client/auth
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:26:2: cannot find module providing package github.com/distribution/distribution/v3/registry/client/auth/challenge
../../../../pkg/mod/github.com/openshift/library-go@v0.0.0-20240528110646-354b673304be/pkg/image/registryclient/client.go:27:2: cannot find module providing package github.com/distribution/distribution/v3/registry/client/transport
```

Now I will need to upgrade library-go as well.... do you understand where I am going? I already followed this path before and that's what brought me to open the PR in library-go and oc.

aguidirh:
Closing this PR as it looks like there is no interest in merging it.

PR #1790, opened 2024-09-04
Add jan as approver

gnufied:
/assign @jsafrane @deads2k


deads2k:
/approve
/hold

only @jsafrane can lgtm and remove the hold
/assign @jsafrane

gnufied:
@jsafrane updated.

I am also adding myself as appover for `pkg/operator/csi` if that is alright. cc @deads2k

PR #1792, opened 2024-09-10
update OpenShift API version

tremes:


bertinatto (commented):
@tremes we bumped openshift/api recently here: https://github.com/openshift/library-go/pull/1785/files. Can we close this one?

PR #1794, opened 2024-09-16
API-1835: Apply 02 more loops

deads2k:


pkg/operator/apiserver/controller/apiservice/apiservice_controller.go
```diff
@@ -77,7 +77,7 @@ func NewAPIServiceController(
 			kubeInformersForNamespaces.InformersFor(metav1.NamespaceSystem).Core().V1().ConfigMaps().Informer(),
 			apiregistrationInformers.Apiregistration().V1().APIServices().Informer(),
 		)...,
-	).ToController(c.name, eventRecorder.WithComponentSuffix("apiservice-"+name+"-controller"))
+	).ToController(c.controllerInstanceName, eventRecorder.WithComponentSuffix("apiservice-"+instanceName+"-controller"))
```
p0lyn0mial, line 80: and events are not "persisted" so changing the name for this controller is fine.

pkg/operator/apiserver/controller/auditpolicy/auditpolicy_controller.go
```diff
@@ -32,6 +35,7 @@ type auditPolicyController struct {
 // NewAuditPolicyController create a controller that watches the config.openshift.io/v1 APIServer object
 // and reconciles a ConfigMap in the target namespace with the audit.k8s.io/v1 policy.yaml file.
 func NewAuditPolicyController(
+	name string,
```
p0lyn0mial, line 38: nit: please rename to `instanceName`.

pkg/operator/apiserver/controller/nsfinalizer/finalizer_controller.go
```diff
@@ -53,7 +53,10 @@ func NewFinalizerController(
 	return factory.New().ResyncEvery(time.Minute).WithSync(c.sync).WithInformers(
 		kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
 		kubeInformersForTargetNamespace.Apps().V1().DaemonSets().Informer(),
-	).ToController(fullname, eventRecorder.WithComponentSuffix("finalizer-controller"))
+	).ToController(
```
p0lyn0mial, line 56: why? the controller doesn't use WithSyncDegradedOnError which is required for reporting a degraded condition on sync err, right ?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -110,7 +110,10 @@ func NewController(name, operatorNamespace, targetNamespace, targetOperandVersio
 
 	return c.WithSync(controllerRef.sync).
 		WithInformers(informers...).
-		ToController(fmt.Sprintf("%sWorkloadController", name), eventRecorder)
+		ToController(
+			fmt.Sprintf("%sWorkloadController", name), // don't change what is passed here unless you also remove the old FooDegraded condition
```
p0lyn0mial, line 114: why? the controller doesn't use WithSyncDegradedOnError which is required for reporting a degraded condition on sync err, right ?

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller.go
```diff
@@ -115,13 +119,13 @@ func NewCSIDriverNodeServiceController(
 	).WithSyncDegradedOnError(
 		operatorClient,
 	).ToController(
-		c.name,
-		recorder.WithComponentSuffix("csi-driver-node-service_"+strings.ToLower(name)),
+		c.instanceName, // don't change what is passed here unless you also remove the old FooDegraded condition
```
p0lyn0mial, line 122: this one actually uses `WithSyncDegradedOnError` so in this case we cannot change the name without a clean up routine.

deads2k:
will get the comments in a followup.

PR #1787, opened 2024-08-30
API-1835: Apply operator status

deads2k:
use apply in operator status instead of updates.  This will
1. eliminate unnecessary conflicts and repeated writes
2. eliminate retries in controller logic
3. make it easier to describe desired state from multiple different controllers
4. make it possible to locate conflicting writers in CI

Messing around with the core bits in this repo and will try revendoring into a couple.

PR #1795, opened 2024-09-17
API-1835: use SSA in CSI controllers

bertinatto:
This is a follow-up to https://github.com/openshift/library-go/pull/1787 to cover CSI controllers.

Proof: https://github.com/openshift/gcp-pd-csi-driver-operator/pull/128

Here's a plot comparing a job with these changes and a job without these changes:

verb_usage_comparison

CC @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller_test.go
+++ b/pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller_test.go
@@ -160,16 +160,6 @@ func withLogLevel(logLevel opv1.LogLevel) driverModifier {
 	}
 }
 
-func withGeneration(generations ...int64) driverModifier {
-	return func(i *fakeDriverInstance) *fakeDriverInstance {
-		i.Generation = generations[0]
-		if len(generations) > 1 {
-			i.Status.ObservedGeneration = generations[1]
-		}
-		return i
-	}
-}
-
 func withGenerations(daemonSet int64) driverModifier {
 	return func(i *fakeDriverInstance) *fakeDriverInstance {
 		i.Status.Generations = []opv1.GenerationStatus{
@@ -473,7 +463,6 @@ func TestSync(t *testing.T) {
 				driver: makeFakeDriverInstance(
 					// withStatus(replica1),
 					withGenerations(1),
-					withGeneration(1, 1),
 					withTrueConditions(conditionAvailable),
 					withFalseConditions(conditionProgressing)),
 			},
@@ -486,7 +475,6 @@ func TestSync(t *testing.T) {
 				driver: makeFakeDriverInstance(
 					// withStatus(replica0),
 					withGenerations(1),
-					withGeneration(1, 1),
 					withTrueConditions(conditionProgressing), // The operator is Progressing
 					withFalseConditions(conditionAvailable)), // The operator is not Available (node not running...)
 			},
@@ -505,7 +493,6 @@ func TestSync(t *testing.T) {
 				driver: makeFakeDriverInstance(
 					// withStatus(replica1),
 					withGenerations(1),
-					withGeneration(1, 1),
 					withTrueConditions(conditionAvailable),
 					withFalseConditions(conditionProgressing)),
 			},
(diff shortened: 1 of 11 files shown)


pkg/operator/resource/resourcemerge/apps_ssa.go
```diff
@@ -0,0 +1,80 @@
...
+	if generations == nil {
+		generations = &[]applyoperatorv1.GenerationStatusApplyConfiguration{}
+	}
+
+	existingGeneration := SSAGenerationFor(*generations, schema.GroupResource{Group: *newGeneration.Group, Resource: *newGeneration.Resource}, *newGeneration.Namespace, *newGeneration.Name)
+	if existingGeneration == nil {
+		*generations = append(*generations, newGeneration)
+		return
+	}
+
```
deads2k, line 35: Under what conditions will this value be present? Because we're starting from scratch with SSA, it seems reasonable to assume that we'll never encounter a case where this already has a value.  And if it did, the apiserver will reject the duplicate.

pkg/operator/resource/resourcemerge/apps_ssa.go
```diff
@@ -0,0 +1,80 @@
...
+	SSASetGeneration(generations, applyoperatorv1.GenerationStatusApplyConfiguration{
+		Group:          ptr.To("apps"),
+		Resource:       ptr.To("deployments"),
+		Namespace:      ptr.To(actual.Namespace),
+		Name:           ptr.To(actual.Name),
+		LastGeneration: ptr.To(actual.Generation),
+	})
+}
+
+func SSAExpectedDaemonSetGeneration(required *appsv1.DaemonSet, previousGenerations []applyoperatorv1.GenerationStatusApplyConfiguration) int64 {
```
deads2k, line 61: dead code?

pkg/operator/resource/resourcemerge/apps_ssa.go
```diff
@@ -0,0 +1,80 @@
...
+	if existingGeneration == nil {
+		*generations = append(*generations, newGeneration)
+		return
+	}
+
+	existingGeneration.LastGeneration = newGeneration.LastGeneration
+	existingGeneration.Hash = newGeneration.Hash
+}
+
+func SSAExpectedDeploymentGeneration(required *appsv1.Deployment, previousGenerations []applyoperatorv1.GenerationStatusApplyConfiguration) int64 {
```
deads2k, line 40: dead code?

pkg/operator/resource/resourcemerge/apps_ssa.go
bertinatto: That makes sense. This can be greatly simplified.

These are left overs from the previous iteration (when I was copying the generations from the OperatorStatus).

pkg/operator/resource/resourcemerge/apps_ssa.go
bertinatto: removed

pkg/operator/resource/resourcemerge/apps_ssa.go
bertinatto: removed

bertinatto:
CC @jsafrane 
Just so you aware we're working on this. It'd be nice if you could review once this is ready.
For context, this is a follow-up to https://github.com/openshift/library-go/pull/1787, where we changed the main controllers to use [SSA](https://kubernetes.io/docs/reference/using-api/server-side-apply/), and CSI was left out.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller_test.go
```diff
@@ -160,16 +160,6 @@ func withLogLevel(logLevel opv1.LogLevel) driverModifier {
 	}
 }
 
-func withGeneration(generations ...int64) driverModifier {
-	return func(i *fakeDriverInstance) *fakeDriverInstance {
-		i.Generation = generations[0]
-		if len(generations) > 1 {
-			i.Status.ObservedGeneration = generations[1]
```
deads2k, line 167: Yup, that's weird.

deads2k:
Once we understand why the observedgeneration is changing, this lgtm.

pkg/operator/csi/csidrivernodeservicecontroller/csi_driver_node_service_controller_test.go
bertinatto: It turns out nothing was really setting this field, we were just adding and confirming its presence in tests. :disappointed_relieved:

PR #1796, opened 2024-09-23
OCPBUGS-36233: Replicas hook consider Control plane topology

dfajmon:
If ControlPlaneTopology is set to `HighlyAvailable` return pod count 2 else 1.

/cc @openshift/storage

### DIFF
--- a/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
+++ b/pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
@@ -12,9 +12,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	corev1 "k8s.io/client-go/informers/core/v1"
-	corev1listers "k8s.io/client-go/listers/core/v1"
 
 	configv1 "github.com/openshift/api/config/v1"
 	opv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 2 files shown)


gnufied:
Wait, I just noticed - this is against 4.17 version. Why? Shouldn't we have a PR against master?

dusk125:
Good catch @gnufied thanks

dfajmon:
changed to master :+1:

pkg/operator/csi/csidrivercontrollerservicecontroller/helpers.go
```diff
@@ -141,22 +139,20 @@ func WithSecretHashAnnotationHook(
...
-// number of replicas will be two. The number of nodes is determined by the node
-// selector specified in the field deployment.Spec.Templates.NodeSelector.
-// When node ports or hostNetwork are used, maxSurge=0 should be set in the
+// WithReplicasHook sets the deployment.Spec.Replicas field according to the ControlPlaneTopology
+// mode. If the topology is set to 'HighlyAvailable' then the number of replicas will be two.
+// Else it will be one. When node ports or hostNetwork are used, maxSurge=0 should be set in the
 // Deployment RollingUpdate strategy to prevent the new pod from getting stuck
 // waiting for a node with free ports.
-func WithReplicasHook(nodeLister corev1listers.NodeLister) dc.DeploymentHookFunc {
+func WithReplicasHook(configInformer configinformers.SharedInformerFactory) dc.DeploymentHookFunc {
```
RomanBednar, line 147: With this change, should we follow up with changing all the callers(operators)? So far I could find the issue with upgrades mentioned for vsphere and csi-snapshot-controller, but leaving some operators unpatched would cause library-go bumps to break them.

Patching everything is a bit more complicated since some of the operators moved to csi-operator repo, but at different releases. Those that moved later than 4.16 need to be patched in their original repos (assuming this  PR really needs to be backported to 4.16).

AWS EBS (4.15) - patch and backport in csi-operator only
AWS EFS (4.17) - patch csi-operator and backport to 4.16 in the original repo
Azure Disk, Azure File (4.16) - patch and backport in csi-operator only
Cinder (4.18) - patch csi-operator and backport to 4.17+4.16 in the original repo
Manila (4.18) - patch csi-operator and backport to 4.17+4.16 in the original repo

That leaves the following callers, not migrated to csi-operator yet, that will be affected and need a patch/backport:

https://github.com/openshift/csi-operator/blob/eb61f0a46d4da914118c3ef60594df3919fa66d0/pkg/driver/common/operator/hooks.go#L54

https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/d2e34142d8cdbb2b956c4265317a3ff59f00dfb5/pkg/operator/vspherecontroller/driver_starter.go#L137

https://github.com/openshift/vmware-vsphere-csi-driver-operator/blob/d2e34142d8cdbb2b956c4265317a3ff59f00dfb5/pkg/operator/vspherecontroller/webhook_controller.go#L29

https://github.com/openshift/cluster-csi-snapshot-controller-operator/blob/8122add3d1529f0ea8fa918b45e3d7b35d5a0388/pkg/operator/starter.go#L283

https://github.com/openshift/gcp-filestore-csi-driver-operator/blob/32bc98ab7c4ce42da8f9b0305ea7a68c78eb000b/pkg/operator/starter.go#L130

https://github.com/openshift/csi-driver-shared-resource-operator/blob/c095a1f2bbd0c365b74e548498d13f8463c69eb2/pkg/deploymentcontroller/deployment.go#L49

https://github.com/openshift/ibm-powervs-block-csi-driver-operator/blob/62ff68d79f14271434c3c45ce950ff7197291ab3/pkg/operator/starter.go#L130

jsafrane: Are we going to backport this? IMO it's not that important bug

RomanBednar: After considering what the backport would require I was thinking the same, the bug does not block the upgrades correct?

gnufied: Just to close the loop on this, we decided not to backport this. We are only going to do this 4.18 and onwards.

PR #1801, opened 2024-09-26
API-1835: update the revision controller to use the type moved latestrevision

deads2k:
also needed to bring library-go in sync with openshift/api.

/assign @p0lyn0mial


pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -353,42 +353,59 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
 		return nil
 	}
 
-	if shouldCreateNewRevision, err := c.revisionPrecondition(ctx); err != nil || !shouldCreateNewRevision {
-		return err
-	}
-
 	// If the operator status's latest available revision is not the same as the observed latest revision, update the operator.
+	// This needs to be done even if the revision precondition is not required because it ensures our operator status is
```
p0lyn0mial, line 357: Okay I think i understand it now. the CM is created by the createNewRevision which is used by the `getLatestAvailableRevision` method below. 

which means that the revision was created thus we should set it, right ?

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -353,42 +353,59 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+	case syncErr != nil:
+		// this is updated in status inside of createRevisionIfNeeded
+		return syncErr
+
+	}
 	if requeue && syncErr == nil {
 		return factory.SyntheticRequeueError
 	}
-	err = syncErr
+	if wroteStatus {
```
p0lyn0mial, line 396: why is this needed?

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -353,42 +353,59 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+	wroteStatus, requeue, syncErr := c.createRevisionIfNeeded(ctx, syncCtx.Recorder(), operatorStatus.LatestAvailableRevision)
+	switch {
+	case requeue && syncErr == nil:
+		return factory.SyntheticRequeueError
+	case syncErr != nil:
+		// this is updated in status inside of createRevisionIfNeeded
+		return syncErr
+
+	}
 	if requeue && syncErr == nil {
```
p0lyn0mial, line 393: rm as you moved this logic to the `switch` case

pkg/operator/revisioncontroller/revision_controller.go
p0lyn0mial: also, should be part of the `switch` case.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -353,42 +353,59 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
...
-		Type:   condition.RevisionControllerDegradedConditionType,
-		Status: operatorv1.ConditionFalse,
-	}
-	if err != nil {
-		cond.Status = operatorv1.ConditionTrue
-		cond.Reason = "Error"
-		cond.Message = err.Error()
-	}
-	if _, _, updateError := v1helpers.UpdateStatus(ctx, c.operatorClient, v1helpers.UpdateConditionFn(cond)); updateError != nil {
+	status := applyoperatorv1.OperatorStatus().
```
p0lyn0mial, line 401: if the `wroteStatus == false` then we should add the `syncErr` msg.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -127,36 +119,43 @@ func (c RevisionController) createRevisionIfNeeded(ctx context.Context, recorder
...
+				WithMessage(err.Error()),
+			).
+			WithLatestAvailableRevision(currentLastAvailableRevision)
+		// different fieldmanager is used to conflict. We could also choose to *always* set the lastavailablerevision
+		// update this tomorrow, that's probably better.
+		updateError := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, status)
+		if updateError != nil {
 			recorder.Warningf("RevisionCreateFailed", "Failed to create revision %d: %v", nextRevision, err.Error())
-			return true, updateError
+			return true, true, updateError
```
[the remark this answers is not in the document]
deads2k: > shouldn't we return `false, true, updateError` ?

We made the attempt so the calling function would write status improperly.

pkg/operator/revisioncontroller/revision_controller.go
deads2k: > Okay I think i understand it now. the CM is created by the createNewRevision which is used by the `getLatestAvailableRevision` method below.
> 
> which means that the revision was created thus we should set it, right ?

What is CM in that sentence?

If a revision exists and is ready (adding that code now), then always set the value.  This catches cases were a previous update failed.

pkg/operator/revisioncontroller/revision_controller.go
deads2k: >if the wroteStatus == false then we should add the syncErr msg.

At this point `syncErr` is guaranteed to be nil, so there's nothing to include

PR #1802, opened 2024-09-26
API-1835: fix up minor comments from Lucasz

deads2k:
from https://github.com/openshift/library-go/pull/1794


p0lyn0mial:
great, but please check the other controllers as well. I think you need to check all of them. For example the comments also apply to https://github.com/deads2k/library-go/blob/a86ed6242ce1d886cd79fbb1785829208d4d4f35/pkg/operator/certrotation/client_cert_rotation_controller.go#L102 which you didn't change here.

PR #1807, opened 2024-09-27
OCPBUGS-42083: Include bootstrap IP in etcd endpoints list

vrutkovs:
During bootstrap kube-apiserver points to master etcd instances. These instances are being added to etcd-servers list one by one, to ensure that if any master can't reach some other etcds this revision
 won't block the entire rollout.
 However, at some point the masters may end up with just one etcd
 (localhost and its IP), which makes it non-HA and apiserver may
 lose etcd connection if this etcd is being reconfigured.

This change updates the list to include bootstrap etcd so that any master would always have an etcd instance to connect to.

Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1745

### DIFF
--- a/pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
+++ b/pkg/operator/configobserver/etcd/observe_etcd_endpoints.go
@@ -18,7 +18,8 @@ const (
 	etcdEndpointName      = "etcd-endpoints"
 
 	// EtcdEndpointName points to the old host-etcd-2 endpoint applicable for clusters prior to 4.5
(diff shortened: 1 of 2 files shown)


deads2k:
found the code that removes this after bootstrapping: https://github.com/openshift/cluster-etcd-operator/blob/72555433bbf7bdb9ce5f8d93c3f49375c2770f0a/pkg/operator/etcdendpointscontroller/etcdendpointscontroller.go#L100-L105

/approve

PR #1808, opened 2024-09-27
OCPBUGS-36233: Replicas hook consider Control plane topology

dfajmon:
If ControlPlaneTopology is set to `HighlyAvailable` return pod count 2 else 1.

backporting from #1796


dfajmon:
/hold
discussing if we really need backport for this

dfajmon:
/close
no backport needed

PR #1811, opened 2024-09-27
API-1835: add mutation tracking to the fake client

deads2k:
Adds the ability to track mutations to our manifest client with unit tests.  We can build on this for comparing predicted versus actual mutations.

### DIFF
--- a/pkg/manifestclient/mutation_tracker.go
+++ b/pkg/manifestclient/mutation_tracker.go
@@ -1,0 +1,168 @@
+package manifestclient
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sync"
+)
+
+type AllActionsTracker struct {
+	lock sync.RWMutex
+
+	ActionToTracker map[Action]*ActionTracker
+}
+
+type Action string
+
+const (
+	// this is really a subset of patch, but we treat it separately because it is useful to do so
+	ActionApply        Action = "server-side-apply"
+	ActionApplyStatus  Action = "server-side-apply-status"
+	ActionUpdate       Action = "update"
+	ActionUpdateStatus Action = "update-status"
+	ActionCreate       Action = "create"
+	ActionDelete       Action = "delete"
+)
+
+type ActionMetadata struct {
+	Action    Action
+	GVR       schema.GroupVersionResource
+	Namespace string
+	Name      string
+}
+
+type ActionTracker struct {
+	Action            Action
+	ResourceToTracker map[schema.GroupVersionResource]*ResourceTracker
+}
+
+type ResourceTracker struct {
+	GVR                schema.GroupVersionResource
+	NamespaceToTracker map[string]*NamespaceTracker
+}
+
+type NamespaceTracker struct {
+	Namespace     string
+	NameToTracker map[string]*NameTracker
+}
+
+type NameTracker struct {
+	Name               string
+	SerializedRequests []SerializedRequest
+}
+
+type SerializedRequest struct {
+	Options []byte
+	Body    []byte
+}
+
+func (a *AllActionsTracker) AddRequest(metadata ActionMetadata, request SerializedRequest) {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	if a.ActionToTracker == nil {
+		a.ActionToTracker = map[Action]*ActionTracker{}
+	}
+	if _, ok := a.ActionToTracker[metadata.Action]; !ok {
+		a.ActionToTracker[metadata.Action] = &ActionTracker{Action: metadata.Action}
+	}
+	a.ActionToTracker[metadata.Action].AddRequest(metadata, request)
+}
+
+func (a *AllActionsTracker) ListActions() []Action {
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	return sets.KeySet(a.ActionToTracker).UnsortedList()
+}
+
+func (a *AllActionsTracker) MutationsForAction(action Action) *ActionTracker {
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+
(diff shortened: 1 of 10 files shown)


pkg/manifestclient/write_roundtripper.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	bodyObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, bodyContent)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode body: %w", err)
+	}
+	bodyYAMLBytes, err := yaml.Marshal(bodyObj.(*unstructured.Unstructured).Object)
+	if err != nil {
+		return nil, fmt.Errorf("unable to encode body: %w", err)
+	}
+	bodyJSONBytes, err := json.Marshal(bodyObj.(*unstructured.Unstructured).Object)
```
p0lyn0mial, line 106: why ? cannot you just return `bodyObj` ?

pkg/manifestclient/write_roundtripper.go
```diff
@@ -0,0 +1,133 @@
...
+		bodyContent, err = io.ReadAll(req.Body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read body: %w", err)
+		}
+	}
+	bodyObj, err := runtime.Decode(unstructured.UnstructuredJSONScheme, bodyContent)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode body: %w", err)
+	}
+	bodyYAMLBytes, err := yaml.Marshal(bodyObj.(*unstructured.Unstructured).Object)
```
p0lyn0mial, line 102: i assume that yaml encoding is just for convenience.

pkg/manifestclient/readwrite_roundtripper.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"bytes"
+	"embed"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// Enter here and call `NewForConfigAndClient(&rest.Config{}, httpClient)`
```
p0lyn0mial, line 11: ?

pkg/manifestclient/readwrite_roundtripper.go
```diff
@@ -0,0 +1,94 @@
...
+
+type MutationTrackingClient interface {
+	GetHTTPClient() *http.Client
+	GetMutations() *AllActionsTracker
+}
+
+func (rt *readWriteRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	switch req.Method {
+	case "GET", "HEAD":
+		return rt.readDelegate.RoundTrip(req)
```
p0lyn0mial, line 80: have you pushed all your changes? i cannot find the readDelegate impl.

pkg/manifestclient/mutation_tracker.go
```diff
@@ -0,0 +1,170 @@
...
+	Name               string
+	SerializedRequests []SerializedRequest
+}
+
+type SerializedRequest struct {
+	Options []byte
+	Body    []byte
+}
+
+var count = 1
```
p0lyn0mial, line 59: unused?

pkg/manifestclient/write_roundtripper.go
```diff
@@ -0,0 +1,133 @@
...
+	requestInfo, err := mrt.requestInfoResolver.NewRequestInfo(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed reading requestInfo: %w", err)
+	}
+
+	if !requestInfo.IsResourceRequest {
+		return nil, fmt.Errorf("non-resource requests are not supported by this implementation")
+	}
+	if len(requestInfo.Subresource) != 0 && requestInfo.Subresource != "status" {
+		return nil, fmt.Errorf("subresource %v is not supported by this implementation", requestInfo.Subresource)
```
p0lyn0mial, line 70: why ? what about `scale` ?

pkg/manifestclient/write_roundtripper.go
```diff
@@ -0,0 +1,133 @@
...
+		Name:      requestInfo.Name,
+	}
+	if action == ActionCreate {
+		// in this case, the name isn't in the URL, it's in the body
+		metadata.Name = bodyObj.(*unstructured.Unstructured).GetName()
+	}
+
+	mrt.actionTracker.AddRequest(metadata, serializedRequest)
+
+	return bodyJSONBytes, nil
```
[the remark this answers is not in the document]
deads2k: > does it mean that the users don't make use the returned objects ?

Will document.  Correct, they cannot use the output.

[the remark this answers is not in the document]
deads2k: > Will document. Correct, they cannot use the output.

Could I make it more obvious with some different kind of return?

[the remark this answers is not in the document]
deads2k: I'm making it very obvious with the return value.

pkg/manifestclient/write_roundtripper.go
deads2k: > why ? cannot you just return `bodyObj` ?

roundtrippers return bytes.  if we desire a refactor at some point, maybe.  But I think this gives us flexibility where we want it.

pkg/manifestclient/readwrite_roundtripper.go
deads2k: > have you pushed all your changes? i cannot find the readDelegate impl.

It merged a couple months back.

pkg/manifestclient/write_roundtripper.go
deads2k: > why ? what about `scale` ?

If we need it, we'll add it.  For now getting this off the ground with enough to start using.

p0lyn0mial:
lgtm

let's give it a try.

PR #1816, opened 2024-10-03
API-1835: Update the static pod state controller to use Apply

p0lyn0mial:


p0lyn0mial:
/hold

holding off for a proof pr.

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
```diff
@@ -162,25 +165,25 @@ func (c *StaticPodStateController) sync(ctx context.Context, syncCtx factory.Syn
...
-	}
+	cond := applyoperatorv1.OperatorCondition().
+		WithType(condition.StaticPodsDegradedConditionType).
+		WithStatus(operatorv1.ConditionFalse)
 	// Failing errors
 	if failingErrorCount > 0 {
-		cond.Status = operatorv1.ConditionTrue
-		cond.Reason = "Error"
-		cond.Message = v1helpers.NewMultiLineAggregate(errs).Error()
+		cond.WithStatus(operatorv1.ConditionTrue).
```
p0lyn0mial, line 173: note that we don't have to reassign the updated cond since we are dealing with pointers.
we should fix the places that already have been changed for consistency.

deads2k:
Lgtm. But needs to pass tests

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
bertinatto: IMO reassigning the variable looks more natural with the builder pattern. I guess it boils down to personal preference (but I agree we should keep consistency).

p0lyn0mial: OK, given that in the other places we already reassign I will update this PR to do the same.

p0lyn0mial:
ok, the `StaticPodsDegraded` condition was properly updated i[n ](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1749/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-gcp-operator/1842153975281881088/artifacts/e2e-gcp-operator/gather-must-gather/artifacts/). this pr can be safely merged.

pkg/operator/staticpod/controller/staticpodstate/staticpodstate_controller.go
p0lyn0mial: Updated, PTAL.

PR #1818, opened 2024-10-04
refactor dynamic client creation and add testing ones

deads2k:
makes it more practical to start testing with different clients.

### DIFF
--- a/pkg/operator/genericoperatorclient/test_operator_client.go
+++ b/pkg/operator/genericoperatorclient/test_operator_client.go
@@ -1,0 +1,35 @@
+package genericoperatorclient
+
+import (
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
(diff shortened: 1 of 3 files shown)


pkg/operator/genericoperatorclient/test_operator_client.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/rest"
+	"net/http"
+)
+
+func NewOperatorClientWithClient(httpClient *http.Client, gvr schema.GroupVersionResource, gvk schema.GroupVersionKind, extractApplySpec OperatorSpecExtractorFunc, extractApplyStatus OperatorStatusExtractorFunc) (v1helpers.OperatorClientWithFinalizers, dynamicinformer.DynamicSharedInformerFactory, error) {
```
p0lyn0mial, line 12: The file name indicates that the method is intended to be used by test code. However, the package name and the file name itself will expose these methods to the production code. Is that intentional?

PR #1821, opened 2024-10-07
[WIP] apply node controller

p0lyn0mial:


pkg/apiserver/jsonpatch/jsonpatch.go
```diff
@@ -34,6 +34,10 @@ func (p *PatchSet) WithTest(path string, value interface{}) *PatchSet {
 	return p
 }
 
+func (p *PatchSet) IsEmpty() bool {
```
deads2k, line 37: outside the diff, but having addOperation reorder the patches makes it impossible to chain multiple mutation options together with preconditions that depend on prior mutations.  If we remove the reordering, then we keep the easy path easy and make the hard path possible.  If we keep the reordering, the hard path is impossible.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -72,9 +77,27 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+			}
+			newTargetNodeStates = append(newTargetNodeStates, newTargetNodeState)
 		} else {
 			syncCtx.Recorder().Warningf("MasterNodeRemoved", "Observed removal of master node %s", nodeState.NodeName)
+			jsonPatch.WithRemove(fmt.Sprintf("/status/nodeStatuses/%d", i), jsonpatch.NewTestCondition(fmt.Sprintf("/status/nodeStatuses/%d/nodeName", i), nodeState.NodeName))
+		}
+	}
+	if !jsonPatch.IsEmpty() {
+		if err = c.operatorClient.PatchOperatorStatus(ctx, jsonPatch); err != nil {
+			return err
```
deads2k, line 100: this will never show in degraded.  need to write status  in order to get the degraded condition.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -110,42 +134,30 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
-		return nil
-	}, v1helpers.UpdateStaticPodConditionFn(newCondition), func(status *operatorv1.StaticPodOperatorStatus) error {
-		status.NodeStatuses = newTargetNodeStates
-		return nil
-	})
-
-	if updateError != nil {
-		return updateError
-	}
+	oldNodeDegradedCondition := v1helpers.FindOperatorCondition(originalOperatorStatus.Conditions, condition.NodeControllerDegradedConditionType)
```
deads2k, line 150: move closer to usage so no one tries to build a new condition from old state.

pkg/operator/staticpod/controller/node/node_controller.go
deads2k: > this will never show in degraded. need to write status in order to get the degraded condition.

And when you do this, remember to write the node status we do know about in this state.

pkg/apiserver/jsonpatch/jsonpatch.go
p0lyn0mial: I just checked the following command on my local cluster and it works:
```
k patch nodestatuses.example.com sample-nodestatus --type='json' -p='[
  {
    "op": "test",
    "path": "/spec/nodeStatuses/0/nodeName",
    "value": "node1"
  },
  {
    "op": "remove",
    "path": "/spec/nodeStatuses/0"
  },
  {
    "op": "test",
    "path": "/spec/nodeStatuses/0/nodeName",
    "value": "node2"
  },
  {
    "op": "remove",
    "path": "/spec/nodeStatuses/0"
  }
]'
```

p0lyn0mial: so the `test` operations don't have to be at the beginning as we originally thought.

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: every `err` should be reported as `degraded` condition ?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -72,9 +77,26 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+				newTargetNodeState.WithLastFailedTime(*originalOperatorStatus.NodeStatuses[i].LastFailedTime)
+			}
+			newTargetNodeStates = append(newTargetNodeStates, newTargetNodeState)
 		} else {
 			syncCtx.Recorder().Warningf("MasterNodeRemoved", "Observed removal of master node %s", nodeState.NodeName)
+			removeAtIndex := i
+			if i > 0 {
+				removeAtIndex = removeAtIndex - 1
+			}
+			jsonPatch.WithRemove(fmt.Sprintf("/status/nodeStatuses/%d", removeAtIndex), jsonpatch.NewTestCondition(fmt.Sprintf("/status/nodeStatuses/%d/nodeName", removeAtIndex), nodeState.NodeName))
```
bertinatto, line 99: I'm not sure I understand this subtraction. Could you elaborate why you need to do this?

I'm confused because whether `i==0` or `i==1` we end up removing the same index?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -110,42 +149,29 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+	if err = c.operatorClient.ApplyStaticPodOperatorStatus(ctx, c.controllerInstanceName, status); err != nil {
+		return err
 	}
 
-	oldNodeDegradedCondition := v1helpers.FindOperatorCondition(oldStatus.Conditions, condition.NodeControllerDegradedConditionType)
-	if oldNodeDegradedCondition == nil || oldNodeDegradedCondition.Message != newCondition.Message {
-		syncCtx.Recorder().Eventf("MasterNodesReadyChanged", newCondition.Message)
+	oldNodeDegradedCondition := v1helpers.FindOperatorCondition(originalOperatorStatus.Conditions, condition.NodeControllerDegradedConditionType)
+	if oldNodeDegradedCondition == nil || oldNodeDegradedCondition.Message != *degradedCondition.Message {
+		syncCtx.Recorder().Eventf("MasterNodesReadyChanged", *degradedCondition.Message)
```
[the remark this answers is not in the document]
p0lyn0mial: it will be, added a new test case (`scenario 12`) for it, ptal

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: added a comment, ptal

bertinatto: Looking much better, thanks!

pkg/operator/staticpod/controller/node/node_controller.go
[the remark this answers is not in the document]
p0lyn0mial: That's okay, unless we want to record an event. I don't see any reason to record an event for `MasterNodeNotRemoved`.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -72,9 +78,30 @@ func (c NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) e
...
+				WithNodeName(originalOperatorStatus.NodeStatuses[i].NodeName).
+				WithCurrentRevision(originalOperatorStatus.NodeStatuses[i].CurrentRevision).
+				WithTargetRevision(originalOperatorStatus.NodeStatuses[i].TargetRevision).
+				WithLastFailedRevision(originalOperatorStatus.NodeStatuses[i].LastFailedRevision).
+				WithLastFailedReason(originalOperatorStatus.NodeStatuses[i].LastFailedReason).
+				WithLastFailedCount(originalOperatorStatus.NodeStatuses[i].LastFailedCount).
+				WithLastFallbackCount(originalOperatorStatus.NodeStatuses[i].LastFallbackCount).
+				WithLastFailedRevisionErrors(originalOperatorStatus.NodeStatuses[i].LastFailedRevisionErrors...)
+			if originalOperatorStatus.NodeStatuses[i].LastFailedTime != nil {
+				newTargetNodeState.WithLastFailedTime(*originalOperatorStatus.NodeStatuses[i].LastFailedTime)
```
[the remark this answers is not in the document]
p0lyn0mial: not sure if i understand what needs to be updated here.

i'm updating `WithLastFailedTime` only when `LastFailedTime != nil` to avoid panicking.

pkg/operator/staticpod/controller/node/node_controller.go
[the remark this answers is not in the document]
bertinatto: If you think we don't need to record the event, I'm fine with that. It seems like the original code intended to record an event for any changes in degraded condition.

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -48,122 +51,403 @@ func addNodeReadyCondition(node *corev1.Node, status corev1.ConditionStatus) *co
...
+					NodeName: "test-node-3",
+				},
+				{
+					NodeName: "test-node-4",
+				},
+			},
+			verifyJSONPatch: func(actualJSONPatch *jsonpatch.PatchSet) error {
+				expectedJSONPatch := jsonpatch.New().
+					WithRemove("/status/nodeStatuses/0", jsonpatch.NewTestCondition("/status/nodeStatuses/0/nodeName", "test-node-1")).
+					WithRemove("/status/nodeStatuses/1", jsonpatch.NewTestCondition("/status/nodeStatuses/1/nodeName", "test-node-3"))
```
bertinatto, line 171: hm, I was confused at first, but this is `1` because the previous `WithRemove` reshuffled the indexes

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -48,122 +51,403 @@ func addNodeReadyCondition(node *corev1.Node, status corev1.ConditionStatus) *co
...
+					NodeName: "test-node-2",
+				},
+				{
+					NodeName: "test-node-3",
+				},
+				{
+					NodeName: "test-node-4",
+				},
+			},
+			triggerStatusUpdateError: fmt.Errorf("nasty err"),
```
bertinatto, line 200: nit: this could be `triggerStatusApplyError`

PR #1823, opened 2024-10-08
regression: Set resource version for CRs

rexagod:
Set resource version for all `monitoring.go` GVRs. Earlier, this was not done by the machinery responsible for it, which caused manifest applications that had no resource version present to encounter the following error:
```
metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```
This patch addresses that regression, which was introduced originally in #1575.

### DIFF
--- a/pkg/operator/resource/resourceapply/monitoring.go
+++ b/pkg/operator/resource/resourceapply/monitoring.go
@@ -136,7 +136,7 @@ func ApplyUnstructuredResourceImproved(
 		return nil, false, err
 	}
 
-	// Check if the metadata objects differ.
+	// Check if the metadata objects differ. This only checks for selective fields (excluding the resource version, among others).
 	didMetadataModify := ptr.To(false)
 	resourcemerge.EnsureObjectMeta(didMetadataModify, &existingObjectMetaTyped, requiredObjectMetaTyped)
 
@@ -162,10 +162,14 @@ func ApplyUnstructuredResourceImproved(
 		klog.Infof("%s %q changes: %v", resourceGVR.String(), namespace+"/"+name, JSONPatchNoError(existing, existingCopy))
 	}
 
+	// CRs, unlike certain core objects (e.g., deployments), need to specify the resource version for updates.
+	requiredCopy := required.DeepCopy()
+	requiredCopy.SetResourceVersion(existingCopy.GetResourceVersion())
+
 	// Perform update if resource exists but different from the required (desired) one.
-	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, required, metav1.UpdateOptions{})
-	resourcehelper.ReportUpdateEvent(recorder, required, err)
-	cache.UpdateCachedResourceMetadata(required, actual)
+	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, requiredCopy, metav1.UpdateOptions{})
+	resourcehelper.ReportUpdateEvent(recorder, requiredCopy, err)
+	cache.UpdateCachedResourceMetadata(requiredCopy, actual)
 	return actual, true, err
 }
 
--- a/test/e2e-monitoring/OWNERS
+++ b/test/e2e-monitoring/OWNERS
@@ -1,0 +1,11 @@
+reviewers:
+  - dgrisonnet
+  - p0lyn0mial
+  - sttts
+  - tkashem
+  - rexagod
+approvers:
+  - dgrisonnet
+  - p0lyn0mial
+  - sttts
+  - tkashem
--- a/test/e2e-monitoring/monitoring_test.go
+++ b/test/e2e-monitoring/monitoring_test.go
@@ -1,0 +1,125 @@
+package e2e_monitoring
+
+import (
+	"context"
+	"testing"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/test/library"
+	monv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/dynamic"
+)
+
+func TestResourceVersionApplication(t *testing.T) {
+	config, err := library.NewClientConfigForTest()
+	require.NoError(t, err)
+
+	// Define the resource.
+	duration := monv1.Duration("5m")
+	resource := monv1.PrometheusRule{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "monitoring.coreos.com/v1",
+			Kind:       "PrometheusRule",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-prometheusrule",
+			Namespace: "default",
+		},
+		Spec: monv1.PrometheusRuleSpec{
+			Groups: []monv1.RuleGroup{
+				{
+					Name: "example",
+					Rules: []monv1.Rule{
+						{
+							Alert: "Foo",
+							Expr:  intstr.FromString("foo > 0"),
+							For:   &duration,
+							Labels: map[string]string{
+								"bar": "baz",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	gvr := schema.GroupVersionResource{
+		Group:    resource.GroupVersionKind().Group,
+		Version:  resource.GroupVersionKind().Version,
+		Resource: "prometheusrules",
+	}
+
+	// Initialize pre-requisites.
+	cache := resourceapply.NewResourceCache()
+	recorder := events.NewInMemoryRecorder("TestResourceVersionApplication")
+	dynamicClient, err := dynamic.NewForConfig(config)
+	require.NoError(t, err)
+
+	// Create the resource.
+	unstructuredResource := &unstructured.Unstructured{}
+	unstructuredResourceMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&resource)
+	require.NoError(t, err)
+	unstructuredResource.SetUnstructuredContent(unstructuredResourceMap)
+	_, didUpdate, err := resourceapply.ApplyUnstructuredResourceImproved(
+		context.TODO(),
+		dynamicClient,
+		recorder,
+		unstructuredResource,
+		cache,
+		gvr,
+		nil,
+		nil,
+	)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		t.Fatalf("Failed to create resource: %v", err)
+	}
+	require.True(t, didUpdate)
+
+	// Update the resource with a change, without specifying a resource version.
+	resource.Spec.Groups[0].Rules[0].Labels["bar"] = "qux"
+	unstructuredResourceMap, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&resource)
+	require.NoError(t, err)
+	unstructuredResource.SetUnstructuredContent(unstructuredResourceMap)
+	_, didUpdate, err = resourceapply.ApplyUnstructuredResourceImproved(
+		context.TODO(),
+		dynamicClient,
+		recorder,
+		unstructuredResource,
+		cache,
+		gvr,
+		nil,
+		nil,
+	)
+	if err != nil {
+		t.Fatalf("Failed to update resource: %v", err)
+	}
+	require.True(t, didUpdate)
+
+	// Update the resource without any changes, without specifying a resource version.
+	_, didUpdate, err = resourceapply.ApplyUnstructuredResourceImproved(
+		context.TODO(),
+		dynamicClient,
+		recorder,
+		unstructuredResource,
+		cache,
+		gvr,
+		nil,
+		nil,
+	)
+	if err != nil {
+		t.Fatalf("Failed to update resource: %v", err)
+	}
+	require.False(t, didUpdate)
+
+	// Delete the resource.
+	err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).Delete(context.TODO(), resource.GetName(), metav1.DeleteOptions{})
+	require.NoError(t, err)
+}


test/e2e-monitoring/monitoring_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/dynamic"
+)
+
+func TestResourceVersionApplication(t *testing.T) {
```
deads2k, line 21: when were and how does this run?

dgrisonnet: @rexagod I added the CI wiring in https://github.com/openshift/release/pull/57664, if you add something similar to the e2e-aws-encryption rule in the Makefile https://github.com/openshift/library-go/blob/master/Makefile#L26-L27, the test will be run in CI

dgrisonnet:
The code change looks good to me, pending the resolution of https://github.com/openshift/library-go/pull/1823#discussion_r1793445225

deads2k:
/approve 
/assign @p0lyn0mial 

for lgtm

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -162,10 +162,14 @@ func ApplyUnstructuredResourceImproved(
...
 
+	// CRs, unlike certain core objects (e.g., deployments), need to specify the resource version for updates.
+	requiredCopy := required.DeepCopy()
+	requiredCopy.SetResourceVersion(existingCopy.GetResourceVersion())
+
 	// Perform update if resource exists but different from the required (desired) one.
-	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, required, metav1.UpdateOptions{})
-	resourcehelper.ReportUpdateEvent(recorder, required, err)
-	cache.UpdateCachedResourceMetadata(required, actual)
+	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, requiredCopy, metav1.UpdateOptions{})
```
p0lyn0mial, line 170: given that `EnsureObjectMeta` and `ensureGenericSpec`modify `existingCopy` and not `requiredCopy` i think we should actually pass `existingCopy` here, am i right ?

rexagod: Right, better to use the `existingCopy` as that also consists of the merged labels, annotations, and owner references, instead of replacing all that with `requiredCopy`'s values. I'd send a patch.

p0lyn0mial: ok, please also cover this case in the unit tests (that is make sure the existingCopy is applied).

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -139,6 +139,10 @@ func ApplyUnstructuredResourceImproved(
 	// Check if the metadata objects differ. This only checks for selective fields (excluding the resource version, among others).
 	didMetadataModify := ptr.To(false)
 	resourcemerge.EnsureObjectMeta(didMetadataModify, &existingObjectMetaTyped, requiredObjectMetaTyped)
+	existingCopy.Object["metadata"], err = runtime.DefaultUnstructuredConverter.ToUnstructured(&existingObjectMetaTyped)
+	if err != nil {
+		return nil, false, err
+	}
```
rexagod, line 145: Note that this is a [breaking change](https://github.com/rexagod/cluster-monitoring-operator/blob/70fc3336873b4cc508f14e484b844ee601720f98/pkg/client/client.go#L671-L704), as earlier the method was only used to detect if the metadata fields differed, but not actually mutate them.

p0lyn0mial: Previously the metadata from the `required` obj were merged to the `existing` obj but  the `required` obj was stored in the database. I thin this behaviour was unexpected, am I right ?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -136,9 +136,13 @@ func ApplyUnstructuredResourceImproved(
 		return nil, false, err
 	}
 
-	// Check if the metadata objects differ.
+	// Check if the metadata objects differ. This only checks for selective fields (excluding the resource version, among others).
 	didMetadataModify := ptr.To(false)
 	resourcemerge.EnsureObjectMeta(didMetadataModify, &existingObjectMetaTyped, requiredObjectMetaTyped)
+	existingCopy.Object["metadata"], err = runtime.DefaultUnstructuredConverter.ToUnstructured(&existingObjectMetaTyped)
```
p0lyn0mial, line 142: shouldn't we update the metadata conditionally  ? (when didMetadataModify == true)

p0lyn0mial:
@rexagod In general, I have a couple of suggestions for you to consider:

1. It seems that `ApplyUnstructuredResourceImproved` method could be made unexported/private. Do you agree?

2. The calling functions of the `applyUnstructuredResourceImproved` pass nil for `defaultingFunc` and `equalityChecker`, so I would like to remove these arguments from applyUnstructuredResourceImproved.

3. I would like to remove the `Fail-fast if the resource versions differ` code, since we merge into the existing object, not the required one.

4. I would like to remove an extra copy from the `ensureGenericsSpec` method.

5. I would like to create a new method, `EnsureObjectMetaForUnstructured`, where `EnsureObjectMeta` is currently located.

6. I would like to move the code that merges metadata into this new function (code from line `105` to line `144`).

What do you think?

pkg/operator/resource/resourceapply/monitoring.go
rexagod: ACK, done.

rexagod:
> It seems that ApplyUnstructuredResourceImproved method could be made unexported/private. Do you agree?

The motivation to move to an `unstructured.Unstructured` approach was to accommodate for resources that do not have their own `Apply<Resource>` methods exposed. The ones in the library right now are only there for the sake of backwards-compatibility.

> The calling functions of the applyUnstructuredResourceImproved pass nil for defaultingFunc and equalityChecker, so I would like to remove these arguments from applyUnstructuredResourceImproved.

To refer to my same reasoning as above, those are `nil` for the exposed resource-dedicated methods as that was the original behavior. However, when using `ApplyUnstructuredResourceImproved` for either (a) unknown resources (ones that do not have a dedicated method here), or (b) more control over the caching mechanism, those are still needed to leverage what the machinery has to offer. For instance, the exposed methods do not use a cache as well, to preserve older behavior, but nonetheless, something the user may still need.

> I would like to remove the Fail-fast if the resource versions differ code, since we merge into the existing object, not the required one.

Done.

> I would like to remove an extra copy from the ensureGenericsSpec method.

Done.

> I would like to create a new method, EnsureObjectMetaForUnstructured, where EnsureObjectMeta is currently located. I would like to move the code that merges metadata into this new function (code from line 105 to line 144).

Done.

rexagod:
Thank you for the thorough review, @p0lyn0mial. I'd like to point out that I've refactored `monitoring.go` code to make it more consistent and readable (for instance, the `ensureGenericSpec` signature changes), in addition to the much-needed refactors you pointed out owing to the older code patterns that existed pre-#1575.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -88,7 +84,8 @@ func ApplyUnstructuredResourceImproved(
 	}
 	existing, err := client.Resource(resourceGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		want, err := client.Resource(resourceGVR).Namespace(namespace).Create(ctx, required, metav1.CreateOptions{})
+		var want *unstructured.Unstructured
```
p0lyn0mial, line 87: nit: since we are returning from this branch anyway I would remove this variable and just simply `want, err := client`

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -102,70 +99,42 @@ func ApplyUnstructuredResourceImproved(
...
 
-	// Cast the metadata to the correct type.
-	var existingObjectMetaTyped, requiredObjectMetaTyped metav1.ObjectMeta
-	err = runtime.DefaultUnstructuredConverter.FromUnstructured(existingObjectMeta, &existingObjectMetaTyped)
-	if err != nil {
-		return nil, false, err
-	}
-	err = runtime.DefaultUnstructuredConverter.FromUnstructured(requiredObjectMeta, &requiredObjectMetaTyped)
+	// Replace and/or merge certain metadata fields.
+	didMetadataModify := ptr.To(false)
```
p0lyn0mial, line 105: could you change to a non-pointer variable to avoid dereferencing bellow ? (`didMetadataModify := false`)

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -102,70 +99,42 @@ func ApplyUnstructuredResourceImproved(
...
-
 	// Deep-check the spec objects for equality, and update the cache in either case.
 	if defaultingFunc == nil {
 		defaultingFunc = noDefaulting
 	}
 	if equalityChecker == nil {
 		equalityChecker = equality.Semantic
 	}
-	existingCopy, didSpecModify, err := ensureGenericSpec(required, existingCopy, defaultingFunc, equalityChecker)
+	didSpecModify := ptr.To(false)
```
p0lyn0mial, line 118: same here.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -102,70 +99,42 @@ func ApplyUnstructuredResourceImproved(
...
+	// Perform update if resource exists but different from the required (desired) one.
 	if klog.V(4).Enabled() {
 		klog.Infof("%s %q changes: %v", resourceGVR.String(), namespace+"/"+name, JSONPatchNoError(existing, existingCopy))
 	}
-
-	// Perform update if resource exists but different from the required (desired) one.
-	actual, err := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, required, metav1.UpdateOptions{})
-	resourcehelper.ReportUpdateEvent(recorder, required, err)
-	cache.UpdateCachedResourceMetadata(required, actual)
+	var actual *unstructured.Unstructured
```
p0lyn0mial, line 134: I would also remove it and simply do `actual, err :=`

test/e2e-monitoring/OWNERS
```diff
@@ -0,0 +1,11 @@
+reviewers:
+  - dgrisonnet
+  - p0lyn0mial
+  - sttts
+  - tkashem
+  - rexagod
+approvers:
+  - dgrisonnet
+  - p0lyn0mial
+  - sttts
```
p0lyn0mial, line 10: unfortunately Stefan is no longer with us.

p0lyn0mial:
ApplyUnstructuredResourceImproved

Is `ApplyUnstructuredResourceImproved` method used outside of the library-go? If not, I would make it private and remove the nil parameters. I would only add them when they are truly needed.

Also, could you please remove `fixup! regression:` from the commit messages?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -84,11 +82,10 @@ func ApplyUnstructuredResourceImproved(
 	}
 	existing, err := client.Resource(resourceGVR).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
 	if errors.IsNotFound(err) {
-		var want *unstructured.Unstructured
-		want, err = client.Resource(resourceGVR).Namespace(namespace).Create(ctx, required, metav1.CreateOptions{})
-		resourcehelper.ReportCreateEvent(recorder, required, err)
+		want, errCreate := client.Resource(resourceGVR).Namespace(namespace).Create(ctx, required, metav1.CreateOptions{})
```
p0lyn0mial, line 85: why not to `want, err :=` ? do we care if the `err` will be shadowed ?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -131,11 +128,10 @@ func ApplyUnstructuredResourceImproved(
 	if klog.V(4).Enabled() {
 		klog.Infof("%s %q changes: %v", resourceGVR.String(), namespace+"/"+name, JSONPatchNoError(existing, existingCopy))
 	}
-	var actual *unstructured.Unstructured
-	actual, err = client.Resource(resourceGVR).Namespace(namespace).Update(ctx, existingCopy, metav1.UpdateOptions{})
-	resourcehelper.ReportUpdateEvent(recorder, existingCopy, err)
+	actual, errUpdate := client.Resource(resourceGVR).Namespace(namespace).Update(ctx, existingCopy, metav1.UpdateOptions{})
```
p0lyn0mial, line 131: why not to `actual, err := ?` do we care if the `err` will be shadowed ?

p0lyn0mial:
@rexagod  Thank you very much for your work - much appreciated!

would you be willing to open a PR similar to [this one](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1752) so we can be sure that the changes introduced in this PR are working as expected?

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Following the same pattern as [here](https://github.com/openshift/library-go/pull/1823#discussion_r1796709432), I'd prefer to keep this change, but let me know if you think otherwise.

rexagod:
Raised https://github.com/openshift/cluster-kube-apiserver-operator/pull/1753.

p0lyn0mial:
ok, the `ci/prow/e2e-aws-ovn-upgrade` on [the PR ](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1753/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn-upgrade/1844743049813430272)was green. Please rebase this PR and let's merge it.

p0lyn0mial:
@rexagod I think you need to add `GO_TEST_PACKAGES :=./test/e2e-monitoring/...` to the `test-e2e-monitoring` target.

p0lyn0mial:
mhm, I think that `ci/prow/e2e-monitoring` isn't running `TestResourceVersionApplication`.

Makefile
```diff
@@ -26,5 +26,6 @@ verify-podnetworkconnectivitychecks:
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 .PHONY: test-e2e-encryption
 
-test-e2e-monitoring:
 .PHONY: test-e2e-monitoring
+test-e2e-monitoring:
+	go test -v -timeout 5m ./test/e2e-monitoring/...
```
rexagod, line 31: I wasn't able to find any usages for `GO_TEST_PACKAGES` in the CI so dropped that.

rexagod:
> Is it when you run make test-e2e-monitoring on your local machine ?

@p0lyn0mial The target was a [placeholder](https://github.com/openshift/release/pull/57664#issuecomment-2410603189) to get the `release` PR in. I've amended that to run the tests now. However, `test-e2e-encryption` needs a similar refactor as that's still a placeholder IIUC.

Makefile
dgrisonnet: `GO_TEST_PACKAGES` is used here https://github.com/openshift/build-machinery-go/blob/5725581bdf8f467ee8fbd394f081a10df670c1ec/make/targets/golang/test-unit.mk#L12. I'd suggest using it so that we stick to the same `go test` invocation everywhere

dgrisonnet:
`test-e2e-encryption` is not a placeholder, the rule in invoking https://github.com/openshift/build-machinery-go/blob/5725581bdf8f467ee8fbd394f081a10df670c1ec/make/targets/golang/test-unit.mk#L12 on https://github.com/openshift/library-go/tree/master/test/e2e-encryption.

rexagod:
Ah, good to know why we specify `GO_TEST_PACKAGES`, however, I'm not sure how frequently this runs? `/test e2e-aws-encryption` also returns a `make: Nothing to be done for 'test-e2e-encryption'` at the moment, which led me to believe it to be a placeholder.

I ask this because I believe we would've caught the issue on `master` if it did anytime soon (PTAL below).

```
┌[rexagod@nebuchadnezzar] [/dev/ttys003] [master] 
└[~/repositories/work/library-go]> gd upstream/master
┌[rexagod@nebuchadnezzar] [/dev/ttys003] [master] 
└[~/repositories/work/library-go]> go vet ./...
# github.com/openshift/library-go/test/e2e-encryption
# [github.com/openshift/library-go/test/e2e-encryption]
vet: test/e2e-encryption/encryption_test.go:85:119: not enough arguments in call to genericoperatorclient.NewClusterScopedOperatorClient
        have (*rest.Config, schema.GroupVersionResource)
        want (clock.PassiveClock, *rest.Config, schema.GroupVersionResource, schema.GroupVersionKind, genericoperatorclient.OperatorSpecExtractorFunc, genericoperatorclient.OperatorStatusExtractorFunc)
```

p0lyn0mial:
let's try with https://github.com/openshift/library-go/pull/1836 - i have run it on my local machine:
```
❯ make test-e2e-monitoring
go test -mod=vendor -race ./test/e2e-monitoring/...
ok  	github.com/openshift/library-go/test/e2e-monitoring	(cached)
```

i'm waiting for the ci job to finish, if it works then we need something similar for the `e2e-aws-encryption`.

dgrisonnet:
> /test e2e-aws-encryption (and local invocation) returns a make: Nothing to be done for 'test-e2e-encryption' at the moment

Oh I see, I missed that. If you add the following line, it should fix it:
```diff
diff --git a/Makefile b/Makefile                                 
index 43276499c..b466e269a 100644
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ verify-podnetworkconnectivitychecks:
 	$(MAKE) -C pkg/operator/connectivitycheckcontroller verify
 
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
+test-e2e-encryption: test-unit
 .PHONY: test-e2e-encryption
 
 test-e2e-monitoring
```

It must have been missed in the past

dgrisonnet:
Hmm actually it reminded me of something and it looks like I had a PR for it that never merged: https://github.com/openshift/library-go/pull/1470

p0lyn0mial:
@rexagod https://github.com/openshift/library-go/pull/1836 worked, we have two options we can merge it or you can pull the changes to your pr.

rexagod:
> i'm waiting for the ci job to finish, if it works then we need something similar for the e2e-aws-encryption.

It seems that on a fresh cluster the test fails due to the absence of `encryptiontests.operator.openshift.io` CRD, which [I'm not sure which operator provides](https://github.com/search?q=%2Fencryptiontests.operator.openshift.io%2F+owner%3Aopenshift+&type=code)?

rexagod:
> i'm waiting for the ci job to finish, if it works then we need something similar for the e2e-aws-encryption.

Since my contextual `test-e2e-encryption`-knowledge is limited and the fact that we've identified at-least a couple of issues [[1]](https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/e2e-encryption/encryption_test.go#L85)[[2]](https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/e2e-encryption/encryption_test.go#L376) around the test, I'd reckon it'd be better if one of the previous authors with some spare cycles can take a look.

Otherwise, I'll revisit this as soon as I can find some.

p0lyn0mial:
> Since my contextual test-e2e-encryption-knowledge is limited and the fact that we've identified at-least a couple of issues [[1]](https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/e2e-encryption/encryption_test.go#L85)[[2]](https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/e2e-encryption/encryption_test.go#L376) around the test, I'd reckon it'd be better if one of the previous authors with some spare cycles can take a look.

@dgrisonnet has a pr for it - https://github.com/openshift/library-go/pull/1470

test/e2e-monitoring/monitoring_test.go
```diff
@@ -0,0 +1,125 @@
...
+		cache,
+		gvr,
+		nil,
+		nil,
+	)
+	if err != nil {
+		t.Fatalf("Failed to update resource: %v", err)
+	}
+	require.False(t, didUpdate)
+
```
p0lyn0mial, line 121: Would it make sense to GET the updated obj and compare it to an expected one?

p0lyn0mial (commented):
Just one last comment, otherwise LGTM.

Makefile
```diff
@@ -26,5 +26,6 @@ verify-podnetworkconnectivitychecks:
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 .PHONY: test-e2e-encryption
 
-test-e2e-monitoring:
+test-e2e-monitoring: GO_TEST_PACKAGES :=./test/e2e-monitoring/...
```
dgrisonnet, line 29: I think it would be useful to add verbosity in case the test starts failing in CI in the future.
You can do that by adding:
```
test-e2e-encryption: GO_TEST_FLAGS += -v
```

rexagod: I've added similar flags in addition to this, however, one that seemed quite off was the 4 hour timeout limit for the `encryption` test. Is that taking into account certain build overheads (which only happen in the CI)? I've added a 5 minute timeout for the `monitoring` test, but LMK if that should be increased to a similar amount.

test/e2e-monitoring/monitoring_test.go
rexagod: Added, PTAL.

Makefile
```diff
@@ -24,7 +24,15 @@ verify-podnetworkconnectivitychecks:
...
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
+test-e2e-encryption: GO_TEST_FLAGS += -v
+test-e2e-encryption: GO_TEST_FLAGS += -timeout 4h
+test-e2e-encryption: GO_TEST_FLAGS += -p 1
+test-e2e-encryption: test-unit
 .PHONY: test-e2e-encryption
 
-test-e2e-monitoring:
+test-e2e-monitoring: GO_TEST_PACKAGES :=./test/e2e-monitoring/...
+test-e2e-encryption: GO_TEST_FLAGS += -v
```
p0lyn0mial, line 34: rename to `test-e2e-monitoring`

Makefile
```diff
@@ -24,7 +24,15 @@ verify-podnetworkconnectivitychecks:
...
+test-e2e-encryption: GO_TEST_FLAGS += -timeout 5m
```
p0lyn0mial, line 35: please remove this entry

Makefile
```diff
@@ -24,7 +24,15 @@ verify-podnetworkconnectivitychecks:
...
+test-e2e-encryption: GO_TEST_FLAGS += -p 1
```
p0lyn0mial, line 36: please remove this entry

Makefile
```diff
@@ -24,7 +24,15 @@ verify-podnetworkconnectivitychecks:
 	$(MAKE) -C pkg/operator/connectivitycheckcontroller verify
 
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
+test-e2e-encryption: GO_TEST_FLAGS += -v
```
p0lyn0mial, line 27: let's not change flags for `test-e2e-encryption` in this PR.

test/e2e-monitoring/monitoring_test.go
```diff
@@ -0,0 +1,134 @@
...
+		nil,
+		nil,
+	)
+	if err != nil {
+		t.Fatalf("Failed to update resource: %v", err)
+	}
+	require.False(t, didUpdate)
+	existingResource, err = dynamicClient.Resource(gvr).Namespace(resource.GetNamespace()).Get(context.TODO(), resource.GetName(), metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Equal(t, gotUnstructured.UnstructuredContent(), existingResource.UnstructuredContent())
```
p0lyn0mial, line 129: does `require.Equal` know how to compare `maps` reliably ?

p0lyn0mial: instead of comparing to the obj we received from the apply function shouldn't we compare to the expected one ?

p0lyn0mial: expected object is the one we created at the beginning of the test (`resource`) which was later updated at line `90`, right ?

p0lyn0mial: by comparing against the expected object we know there no other (unexpected) modifications were made.

p0lyn0mial: > does require.Equal know how to compare maps reliably ?

ok, it looks like it uses `reflect.DeepEqual` under the hood.

Makefile
rexagod: Ah! 🤦🏼

test/e2e-monitoring/monitoring_test.go
rexagod: > we know there no other (unexpected) modifications were made.

There's a few server-side added fields we'll need to inject, but I wonder if [`12f2bce` (#1833)](https://github.com/openshift/library-go/pull/1833/commits/12f2bce9c9c28af97c00861845613ae2d3f0622d#diff-c44fd9ab87f08101aadd180c482e8f18c866a305b78c7289a95a518cacdf2203) suffices for this requirement?

rexagod: Pushed.

p0lyn0mial: I would prefer comparing against the expected object. Now, I think we should compare both the received object and the stored object against the expected one, since callers might use the received object as well, wdyt?

I don't expect many fields to be changed by the server. We could just overwrite or clear the ones we know will be provided by the server, such as RV.

rexagod: Sorry, can you clarify what the `expected`, `received`, and `stored` objects are in this context?

To me (and I'm probably wrong here), `unstructuredResource` seems the same as the `expected` and `stored` object (as it's what we originally create in the beginning, and what we store?), whereas `gotUnstructured` is the one we `receive`?

rexagod: Also, I believe we are comparing the received object with the expected one here: [`17a435d` (#1823)](https://github.com/openshift/library-go/pull/1823/commits/17a435de41df465c17974773c06c77fa9d7fddfe#diff-c44fd9ab87f08101aadd180c482e8f18c866a305b78c7289a95a518cacdf2203R114-R120)? Not sure what the stored object is though (do you mean the one in the cluster?).

rexagod: Pushed a patch that compares the expected (original) resource to the existing (cluster) and the received (returned by `ApplyUnstructuredResourceImproved`).

Makefile
p0lyn0mial: can you remove changes to `test-e2e-encryption`? this part will be updated in https://github.com/openshift/library-go/pull/1470

Makefile
```diff
@@ -26,5 +26,9 @@ verify-podnetworkconnectivitychecks:
 test-e2e-encryption: GO_TEST_PACKAGES :=./test/e2e-encryption/...
 .PHONY: test-e2e-encryption
 
-test-e2e-monitoring:
+test-e2e-monitoring: GO_TEST_PACKAGES :=./test/e2e-monitoring/...
+test-e2e-monitoring: GO_TEST_FLAGS += -v
+test-e2e-monitoring: GO_TEST_FLAGS += -timeout 5m
```
rexagod, line 31: https://github.com/openshift/library-go/pull/1823/files#r1804621250

p0lyn0mial: in a follow-up pr i would remove the timeout (let's use the default timeout) and the `-p` flag.

p0lyn0mial:
/lgtm

@rexagod many thanks for the PR.

PR #1824, opened 2024-10-08
API-1835: demonstrate how we will store desired mutations

deads2k:
1. adds ability to write recorded mutation requests to disk
2. adds ability to read recorded mutation requests from disk
3. adds ability to compare various mutation requests for equivalence
4. adds test that can record mutation requests to a testdata location.  this tests 1 and we need it for real operators
5. adds test that reads the tesdata location. this tests 2 and 3
6. adds a fake annotation to deletion request bodies so that we know which instance was being deleted
7. adds function to make filenames go.mod friendly

### DIFF
--- a/pkg/manifestclient/mutation_directory_reader.go
+++ b/pkg/manifestclient/mutation_directory_reader.go
@@ -16,26 +16,49 @@ import (
 )
 
 func ReadMutationDirectory(mutationDirectory string) (*AllActionsTracker[FileOriginatedSerializedRequest], error) {
+	return readMutationFS(os.DirFS(mutationDirectory), ".")
+}
+
+func ReadEmbeddedMutationDirectory(inFS fs.FS, mutationDirectory string) (*AllActionsTracker[FileOriginatedSerializedRequest], error) {
+	return readMutationFS(inFS, mutationDirectory)
+}
+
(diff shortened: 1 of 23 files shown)


bertinatto (commented):
/lgtm

pkg/manifestclient/serialized_request.go
```diff
@@ -0,0 +1,220 @@
...
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type SerializedRequestish interface {
```
bertinatto, line 12: SerializedRequestAcessor?

pkg/manifestclient/mutation_directory_reader.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil, nil
+}
+
+func readSerializedRequestsFromActionDirectory(action Action, mustGatherDir string) ([]FileOriginatedSerializedRequest, error) {
+	currResourceList := []FileOriginatedSerializedRequest{}
+	errs := []error{}
+	err := filepath.WalkDir(mustGatherDir, func(currLocation string, currFile fs.DirEntry, err error) error {
+		if err != nil {
+			errs = append(errs, err)
+		}
```
bertinatto, line 64: OK, if an error related to this path is reported, we continue walking rather than stopping...

PR #1826, opened 2024-10-09
intro json patch library

p0lyn0mial:
proof pr at https://github.com/openshift/origin/pull/29163

### DIFF
--- a/pkg/apiserver/jsonpatch/jsonpatch.go
+++ b/pkg/apiserver/jsonpatch/jsonpatch.go
@@ -1,0 +1,65 @@
+package jsonpatch
+
+import (
+	"encoding/json"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/apiserver/jsonpatch/jsonpatch.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"encoding/json"
+)
+
+type PatchOperation struct {
+	Op    string      `json:"op,omitempty"`
+	Path  string      `json:"path,omitempty"`
+	Value interface{} `json:"value,omitempty"`
+}
```
deads2k, line 11: I was going to complain about this, but it appears that jsonpatch doesn't actually provide something nice here.  Wild.

PR #1828, opened 2024-10-10
NO-JIRA: don't use unnecessary generics

deads2k:


deads2k:
found while making use in the MOM repo.

PR #1830, opened 2024-10-11
API-1835: Migrate WorkloadController to SSA

bertinatto:
Proof: https://github.com/openshift/cluster-authentication-operator/pull/715.

Here's a verb usage comparison between **e2e-aws-single-node** jobs of the PR above and https://github.com/openshift/cluster-authentication-operator/pull/705.

verb_usage_comparison_auth

/assign @deads2k @p0lyn0mial

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -165,20 +165,16 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	}
 
 	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)).
-		WithStatus(operatorv1.ConditionTrue)
+		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable))
 
 	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix)).
-		WithStatus(operatorv1.ConditionFalse)
+		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
 
 	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix)).
-		WithStatus(operatorv1.ConditionFalse)
+		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
 
 	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)).
-		WithStatus(operatorv1.ConditionFalse)
+		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
 
 	status := applyoperatorv1.OperatorStatus()
 	defer func() {
@@ -218,6 +214,11 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithStatus(operatorv1.ConditionFalse).
 			WithReason("PreconditionNotFulfilled")
 
+		workloadDegradedCondition = workloadDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("PreconditionNotFulfilled").
+			WithMessage(message)
+
 		return kerrors.NewAggregate(errs)
 	}
 
@@ -248,6 +249,11 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithMessage(message)
 
 		deploymentDegradedCondition = deploymentDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
+			WithMessage(message)
+
+		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("NoDeployment").
 			WithMessage(message)
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -241,6 +241,11 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("SyncError").
 			WithMessage(message)
+	} else if workload == nil {
+		workloadDegradedCondition = workloadDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
+			WithMessage(fmt.Sprintf("deployment/%s: could not be retrieved", c.targetNamespace))
 	} else {
 		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionFalse)
@@ -259,11 +264,6 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithMessage(message)
 
 		deploymentDegradedCondition = deploymentDegradedCondition.
-			WithStatus(operatorv1.ConditionTrue).
-			WithReason("NoDeployment").
-			WithMessage(message)
-
-		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("NoDeployment").
 			WithMessage(message)
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -17,7 +17,6 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
-	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	openshiftconfigclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -177,6 +176,17 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
 
 	status := applyoperatorv1.OperatorStatus()
+	if workload != nil {
+		// The Hash field is not required since the LastGeneration field is enough to uniquely identify a Deployment's desired state
+		status = status.WithGenerations(applyoperatorv1.GenerationStatus().
+			WithGroup("apps").
+			WithResource("deployments").
+			WithNamespace(workload.Namespace).
+			WithName(workload.Name).
+			WithLastGeneration(workload.Generation),
+		)
+	}
+
 	defer func() {
 		status = status.WithConditions(
 			deploymentAvailableCondition,
@@ -329,15 +339,6 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		}
 		c.versionRecorder.SetVersion(operandName, c.targetOperandVersion)
 	}
-
-	// set generations
-	status = status.WithGenerations(&applyoperatorv1.GenerationStatusApplyConfiguration{
-		Group:          ptr.To("apps"),
-		Resource:       ptr.To("deployments"),
-		Namespace:      ptr.To(workload.Namespace),
-		Name:           ptr.To(workload.Name),
-		LastGeneration: ptr.To(workload.Generation),
-	})
 
 	if len(errs) > 0 {
 		return kerrors.NewAggregate(errs)
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -17,13 +17,14 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	openshiftconfigclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	"github.com/openshift/library-go/pkg/apps/deployment"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
-	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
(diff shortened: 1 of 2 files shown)


bertinatto:
/hold
for results from proof: https://github.com/openshift/cluster-authentication-operator/pull/715

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -161,40 +164,33 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		errs = []error{}
 	}
 
-	deploymentAvailableCondition := operatorv1.OperatorCondition{
-		Type:   fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable),
-		Status: operatorv1.ConditionTrue,
-	}
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
+		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)).
+		WithStatus(operatorv1.ConditionTrue)
```
deads2k, line 169: Thought we agreed to initialize without status to ensure that every branch sets all conditions

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -313,11 +324,14 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		c.versionRecorder.SetVersion(operandName, c.targetOperandVersion)
 	}
 
-	// set updateGenerationFn so that it is invoked in defer
-	updateGenerationFn = func(newStatus *operatorv1.OperatorStatus) error {
-		resourcemerge.SetDeploymentGeneration(&newStatus.Generations, workload)
-		return nil
-	}
+	// set generations
+	status = status.WithGenerations(&applyoperatorv1.GenerationStatusApplyConfiguration{
```
deads2k, line 328: need to ensure *every* branch has generations set.  Probably add a check in that defer.

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: My bad - fixed.

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: Created the status with the generations filled (like we do in the deployment controller).

bertinatto:
> /hold for results from proof: [openshift/cluster-authentication-operator#715](https://github.com/openshift/cluster-authentication-operator/pull/715)

/hold cancel

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -71,8 +71,10 @@ func TestUpdateOperatorStatus(t *testing.T) {
 						Message: "deployment/: could not be retrieved",
 					},
 					{
-						Type:   fmt.Sprintf("%sWorkloadDegraded", defaultControllerName),
-						Status: operatorv1.ConditionFalse,
+						Type:    fmt.Sprintf("%sWorkloadDegraded", defaultControllerName),
+						Status:  operatorv1.ConditionTrue,
+						Reason:  "NoDeployment",
+						Message: "deployment/: could not be retrieved",
```
bertinatto, line 77: @deads2k @p0lyn0mial this is interesting. Forcing every branch to set the conditions revealed an underlying issue. IMO this condition was set incorrectly, but please take a look.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -105,8 +107,8 @@ func TestUpdateOperatorStatus(t *testing.T) {
 					{
 						Type:    fmt.Sprintf("%sWorkloadDegraded", defaultControllerName),
 						Status:  operatorv1.ConditionTrue,
-						Message: "nasty error\n",
-						Reason:  "SyncError",
+						Reason:  "NoDeployment",
```
p0lyn0mial, line 110: We shouldn't overwrite `SyncError` if occurred, these errors are important. Does that make sense?

pkg/operator/apiserver/controller/workload/workload_test.go
p0lyn0mial: It looks like you were right, previously this condition was not set when the workload was missing. Setting it doesn't break anything.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -227,37 +236,49 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+			WithMessage(message)
+
+		deploymentDegradedCondition = deploymentDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
+			WithMessage(message)
+
+		workloadDegradedCondition = workloadDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
```
p0lyn0mial, line 267: I think that we shouldn't overwrite `SyncError` if set.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -209,15 +210,23 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 
-		deploymentProgressingCondition.Status = operatorv1.ConditionFalse
-		deploymentProgressingCondition.Reason = "PreconditionNotFulfilled"
+		deploymentProgressingCondition = deploymentProgressingCondition.
+			WithStatus(operatorv1.ConditionFalse).
+			WithReason("PreconditionNotFulfilled")
+
+		workloadDegradedCondition = workloadDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("PreconditionNotFulfilled").
```
p0lyn0mial, line 228: we don't know if the workload is degraded but because now we always have to set all conditions we need to do something about it.

maybe we should set the status to `Unknown`?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -209,15 +210,23 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 
 		// we are degraded, not available and we are not progressing
 
-		deploymentDegradedCondition.Status = operatorv1.ConditionTrue
-		deploymentDegradedCondition.Reason = "PreconditionNotFulfilled"
-		deploymentDegradedCondition.Message = message
+		deploymentDegradedCondition = deploymentDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
```
p0lyn0mial, line 214: maybe we should set status for all conditions in this case (`PreconditionNotFulfilled`) to `Unknown` ?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -268,18 +289,21 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
 	workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas
-	workloadIsBeingUpdatedTooLong, err := isUpdatingTooLong(previousStatus, deploymentProgressingCondition.Type)
+	workloadIsBeingUpdatedTooLong, err := isUpdatingTooLong(previousStatus, *deploymentProgressingCondition.Type)
```
p0lyn0mial, line 292: I think that `*deploymentProgressingCondition.Type` will always be set.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -161,40 +163,39 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
-		Type:   fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing),
-		Status: operatorv1.ConditionFalse,
+	status := applyoperatorv1.OperatorStatus()
+	if workload != nil {
+		status = status.WithGenerations(applyoperatorv1.GenerationStatus().
+			WithGroup("apps").
+			WithResource("deployments").
+			WithNamespace(workload.Namespace).
+			WithName(workload.Name).
+			WithLastGeneration(workload.Generation),
```
p0lyn0mial, line 185: not setting the generation will set it to `0`, right ?

bertinatto: Yes, it should be set to the zero value

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: Replied in the other comment.

pkg/operator/apiserver/controller/workload/workload_test.go
bertinatto: Good point. I added a commit to change that, PTAL.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -111,7 +113,7 @@ func NewController(name, operatorNamespace, targetNamespace, targetOperandVersio
 	return c.WithSync(controllerRef.sync).
 		WithInformers(informers...).
 		ToController(
-			fmt.Sprintf("%sWorkloadController", name), // don't change what is passed here unless you also remove the old FooDegraded condition
+			fmt.Sprintf("%sWorkloadController", instanceName),
```
p0lyn0mial, line 116: could we use `c.controllerInstanceName` here ?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -161,40 +163,39 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
-			v1helpers.UpdateConditionFn(deploymentDegradedCondition),
-			v1helpers.UpdateConditionFn(deploymentProgressingCondition),
-			v1helpers.UpdateConditionFn(workloadDegradedCondition),
-		}
-		if updateGenerationFn != nil {
-			updates = append(updates, updateGenerationFn)
-		}
-		if _, _, updateError := v1helpers.UpdateStatus(ctx, c.operatorClient, updates...); updateError != nil {
-			err = updateError
+		status := status.WithConditions(
```
p0lyn0mial, line 190: nit this could be `status = status.`

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -161,40 +163,39 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().
+		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
 
-	deploymentProgressingCondition := operatorv1.OperatorCondition{
-		Type:   fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing),
-		Status: operatorv1.ConditionFalse,
+	status := applyoperatorv1.OperatorStatus()
+	if workload != nil {
+		status = status.WithGenerations(applyoperatorv1.GenerationStatus().
```
p0lyn0mial, line 180: The previous [function](https://github.com/openshift/library-go/blob/816295c705d36a0115f4dc4340c94ebbc227158d/pkg/operator/resource/resourcemerge/apps.go#L37) was also setting `Hash` field. Not needed anymore ?

pkg/operator/apiserver/controller/workload/workload.go
p0lyn0mial: don't have a strong opinion either. It's not a big deal because it can be easily changed. Changing the type is much harder, but we can adjust it in the future.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -227,37 +236,52 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+			WithReason("NoDeployment").
+			WithMessage(message)
+
+		deploymentDegradedCondition = deploymentDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
+			WithMessage(message)
+
+		// Don't overwrite workoadDegraded if it's currently true
+		if workloadDegradedCondition.Status == nil || *workloadDegradedCondition.Status == operatorv1.ConditionFalse {
```
p0lyn0mial, line 266: would it make sense to move this code to line 243 where the `workloadDegradedCondition` is set ? I think it would look like:

```
if len(errs) > 0 {
		message := ""
		for _, err := range errs {
			message = message + err.Error() + "\n"
		}
		workloadDegradedCondition = workloadDegradedCondition.
			WithReason("SyncError").
			WithMessage(message)
	} else if workload == nil {
		workloadDegradedCondition = workloadDegradedCondition.
			WithReason("NoDeployment").
			WithMessage(fmt.Sprintf("deployment/%s: could not be retrieved", c.targetNamespace))
	} else {
		workloadDegradedCondition = workloadDegradedCondition.
			WithStatus(operatorv1.ConditionFalse)
	}
```

and then we could remove setting `workloadDegradedCondition` in this branch.

pkg/operator/apiserver/controller/workload/workload.go
deads2k: > he previous [function](https://github.com/openshift/library-go/blob/816295c705d36a0115f4dc4340c94ebbc227158d/pkg/operator/resource/resourcemerge/apps.go#L37) was also setting `Hash` field. Not needed anymore ?

Never set, rigth? https://github.com/openshift/library-go/blob/816295c705d36a0115f4dc4340c94ebbc227158d/pkg/operator/resource/resourcemerge/apps.go#L48-L59

p0lyn0mial: set in the SetGeneration function (https://github.com/openshift/library-go/blob/816295c705d36a0115f4dc4340c94ebbc227158d/pkg/operator/resource/resourcemerge/apps.go#L37)

deads2k: So always empty since the newGeneration never had it set, rigth?

bertinatto: Previously, this controller set the generation using the `SetGenerationDeployment` helper, which doesn't set the hash field:

https://github.com/openshift/library-go/blob/master/pkg/operator/apiserver/controller/workload/workload.go#L316-L320

This helpers calls `SetGeneration`, which does copy the field over if it exists, but in practice the field is always empty because `SetGeneration` is called through `SetDeploymentGeneration`.

Example:

```
omc get authentications.operator.openshift.io/cluster -o json | jq .status.generations
[
  {
    "group": "apps",
    "hash": "",
    "lastGeneration": 3,
    "name": "apiserver",
    "namespace": "openshift-oauth-apiserver",
    "resource": "deployments"
  },
  {
    "group": "apps",
    "hash": "",
    "lastGeneration": 2,
    "name": "oauth-openshift",
    "namespace": "openshift-authentication",
    "resource": "deployments"
  }
]
```

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: Done

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: Works for me. Added to a new commit, PTAL

p0lyn0mial:
LGTM

just squash the last commit `[workloadscontroller: address suggestions by Lukasz]`

pkg/operator/apiserver/controller/workload/workload.go
bertinatto: Added a comment to explain why the field is not required for a deployment

PR #1831, opened 2024-10-11
add support for minimumKubeletVersion

haircommander:
carrying https://github.com/openshift/api/pull/2059


haircommander:
api changes merged, so this is ready

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func ValidateMinimumKubeletVersion(nodesGetter corev1client.NodesGetter, minimumKubeletVersion string) *field.Error {
+	// unset, no error
+	if minimumKubeletVersion == "" {
+		return nil
+	}
+
+	fieldPath := field.NewPath("spec", "minimumKubeletVersion")
+	nodes, err := nodesGetter.Nodes().List(context.TODO(), metav1.ListOptions{})
```
benluddy, line 23: Can this be made to accept the lister interface instead of the client interface? Listing all nodes as part of validating an API field is probably more expensive than we would like.

haircommander: updated!

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,55 @@
...
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+func ValidateMinimumKubeletVersion(nodeLister corev1listers.NodeLister, minimumKubeletVersion string) *field.Error {
+	// unset, no error
+	if minimumKubeletVersion == "" {
+		return nil
+	}
+
+	fieldPath := field.NewPath("spec", "minimumKubeletVersion")
```
benluddy, line 21: xref https://github.com/openshift/kubernetes/pull/2104#discussion_r1828102862

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,55 @@
...
+		_, errStr := IsKubeletVersionTooOld(node, &version)
+		if errStr != "" {
+			return field.Invalid(fieldPath, minimumKubeletVersion, errStr)
+		}
+	}
+	return nil
+}
+
+func IsKubeletVersionTooOld(node *corev1.Node, minVersion *semver.Version) (bool, string) {
+	version, err := semver.Parse(strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v"))
```
benluddy, line 42: Note for other readers like me who did not know that node status could be set on create: https://github.com/kubernetes/kubernetes/blob/b4d91d1b8ae336ad85b59de0ce15b8b83fa6d710/pkg/registry/core/node/strategy.go#L94

pkg/config/node/minimum_kubelet_version.go
haircommander: good point

haircommander:
I dropped the field references and also I thought about it more and being opinonated about the of the nodes (node lister vs nodesGetter) began to feel weird so I now just have it taking a slice of nodes and the various callers can decide where to get them. left in a separate commit in case you liked it the old way @benluddy

benluddy:
> I dropped the field references and also I thought about it more and being opinonated about the of the nodes (node lister vs nodesGetter) began to feel weird so I now just have it taking a slice of nodes and the various callers can decide where to get them. left in a separate commit in case you liked it the old way @benluddy

That's OK, but note that the items field in `NodeList` has type `[]Node`, not `[]*Node`, so the caller would have to transform one to the other anyway.

haircommander:
any other thoughts @benluddy ??

haircommander:
@benluddy sorry for the churn, but I've updated the Is*TooOld functions to only return an error (and a specific one if the node is too old) so callers can only check one value (most were anyway, and https://github.com/openshift/kubernetes/pull/2104#discussion_r1847510351 lead me in that direction)

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
...
+
+// An error to be returned when kubelet version is lower than specified minimum kubelet version.
+// Used to differentiate between a parsing failure, and a failure because the kubelet is out of date.
+var ErrKubeletOutdated = fmt.Errorf("kubelet version is outdated")
+
+// ValidateMinimumKubeletVersion takes a list of nodes and a currently set min version.
+// It parses the min version and iterates through the nodes, comparing the version of the kubelets
+// to the min version.
+// It will error if any nodes are older than the min version.
+func ValidateMinimumKubeletVersion(nodes []*corev1.Node, minimumKubeletVersion string) error {
```
p0lyn0mial, line 20: I think this function should be part of an admission plugin defined in the server. If so, could you explain why it was added to the library-go?

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return nil
+}
+
+// IsNodeTooOld answers that very question. It takes a node object and a minVersion,
+// parses each into a semver version, and then determines whether the version of the kubelet on the
+// node is older than min version.
+// When the node is too old, it returns the error ErrKubeletOutdated. If a different error occurs, an error is returned.
+// If the node is new enough and no error happens, nil is returned.
+func IsNodeTooOld(node *corev1.Node, minVersion *semver.Version) error {
```
p0lyn0mial, line 44: Do the other functions need to be exported?

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
...
+func IsKubeletVersionTooOld(kubeletVersion string, minVersion *semver.Version) error {
+	if minVersion == nil {
+		return fmt.Errorf("given minimum version is nil")
+	}
+	version, err := ParseKubeletVersion(kubeletVersion)
+	if err != nil {
+		return fmt.Errorf("failed to parse node version %s: %v", kubeletVersion, err)
+	}
+	if minVersion.GT(*version) {
+		return fmt.Errorf("%w: kubelet version is %v, which is lower than minimumKubeletVersion of %v", ErrKubeletOutdated, *version, *minVersion)
```
p0lyn0mial, line 64: I think that there is no test case for this branch. Could you add it ?

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+
+	"github.com/blang/semver/v4"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// An error to be returned when kubelet version is lower than specified minimum kubelet version.
+// Used to differentiate between a parsing failure, and a failure because the kubelet is out of date.
+var ErrKubeletOutdated = fmt.Errorf("kubelet version is outdated")
```
p0lyn0mial, line 14: In the missing test case, we could ensure that the error can be unwrapped to the sentinel error.

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
...
+	if minVersion.GT(*version) {
+		return fmt.Errorf("%w: kubelet version is %v, which is lower than minimumKubeletVersion of %v", ErrKubeletOutdated, *version, *minVersion)
+	}
+	return nil
+}
+
+// ParseKubeletVersion parses it into a semver.Version object, stripping
+// any information in the version that isn't "major.minor.patch".
+func ParseKubeletVersion(kubeletVersion string) (*semver.Version, error) {
+	version, err := semver.Parse(strings.TrimPrefix(kubeletVersion, "v"))
```
p0lyn0mial, line 72: I think there is no test case that prefixes the version with `v`.

bertinatto (commented):
Added a few comments, but overall LGTM. I'll let the final tag for Lukasz.

/approve

pkg/operator/configobserver/node/observe_minimum_kubelet_version.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver"
+	"github.com/openshift/library-go/pkg/operator/events"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var minimumKubeletVersionConfigPath = "minimumKubeletVersion"
+
+// ObserveKubeletMinimumVersion watches the node configuration and generates the minimumKubeletVersion
+func ObserveMinimumKubeletVersion(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (ret map[string]interface{}, errs []error) {
```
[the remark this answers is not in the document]
haircommander: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1754 does so, this is a common pattern from what I can see

pkg/config/node/minimum_kubelet_version.go
haircommander: this library is to share code between [KAS](https://github.com/openshift/kubernetes/pull/2104) and [hypershift](https://github.com/openshift/hypershift/pull/4980) it seemed cleanest to me to put these library functions here, even though not all of them are needed by both

pkg/config/node/minimum_kubelet_version.go
haircommander: we'll use them elsewhere in [KAS](https://github.com/openshift/kubernetes/pull/2104) and [hypershift](https://github.com/openshift/hypershift/pull/4980)

pkg/config/node/minimum_kubelet_version.go
haircommander: https://github.com/openshift/library-go/pull/1831/files#diff-46d78ef74ff7e0efcd85f615801ad28063fd5732d099050c47db754940e16f91R78

pkg/config/node/minimum_kubelet_version.go
haircommander: https://github.com/openshift/library-go/pull/1831/files#diff-46d78ef74ff7e0efcd85f615801ad28063fd5732d099050c47db754940e16f91R193-R195 does with https://github.com/openshift/library-go/pull/1831/files#diff-46d78ef74ff7e0efcd85f615801ad28063fd5732d099050c47db754940e16f91R84

pkg/config/node/minimum_kubelet_version.go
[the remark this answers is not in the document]
haircommander: I thouht it was cleaner than needing to make a fake node object in https://github.com/openshift/hypershift/pull/4980 (where we just pass around the lowest kubelet version in the hosted cluster)

haircommander:
thanks @bertinatto and @p0lyn0mial ! I have updated / addressed your comments, PTAL again :)

pkg/config/node/minimum_kubelet_version.go
```diff
@@ -0,0 +1,80 @@
+package node
+
```
p0lyn0mial, line 2: It's not a big deal, but I think `pkg/config` is meant for client configuration. How about moving this code to `pkg/apiserver`?

pkg/operator/configobserver/node/observe_minimum_kubelet_version_test.go
```diff
@@ -0,0 +1,80 @@
...
+			name:                   "empty minimumKubeletVersion",
+			expectedObservedConfig: map[string]interface{}{},
+			minimumKubeletVersion:  "",
+		},
+		{
+			name: "set minimumKubeletVersion",
+			expectedObservedConfig: map[string]interface{}{
+				"minimumKubeletVersion": string("1.30.0"),
+			},
+			minimumKubeletVersion: "1.30.0",
```
p0lyn0mial, line 34: should this value also be prefixed with `v`?

p0lyn0mial:
I left a few suggestions, but overall, this PR looks great, thanks!

pkg/operator/configobserver/node/observe_minimum_kubelet_version_test.go
haircommander: actually the min version API doesn't accept a v. only on the kubelet's side do we strip it

pkg/config/node/minimum_kubelet_version.go
haircommander: done

pkg/operator/configobserver/node/observe_minimum_kubelet_version.go
```diff
@@ -0,0 +1,45 @@
...
+	ret = map[string]interface{}{}
+	listers := genericListers.(NodeLister)
+	configNode, err := listers.NodeLister().Get("cluster")
+	// we got an error so without the node object we are not able to determine minimumKubeletVersion
+	if err != nil {
+		// if config/v1/node/cluster object is not found, that can be treated as a non-error case
+		if !apierrors.IsNotFound(err) {
+			errs = append(errs, err)
+		} else { // but raise a warning
+			recorder.Warningf("ObserveMinimumKubeletVersion", "nodes.%s/cluster not found", configv1.GroupName)
```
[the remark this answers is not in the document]
haircommander: removed!

pkg/apiserver/node/minimum_kubelet_version_test.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, testCase := range testCases {
+		shouldStr := "should not be"
+		if testCase.shouldReject {
+			shouldStr = "should be"
+		}
+		t.Run(testCase.name, func(t *testing.T) {
+			err := ValidateMinimumKubeletVersion(testCase.nodes, testCase.version)
+			assert.Equal(t, testCase.shouldReject, err != nil, "minimum kubelet version %q %s rejected", testCase.version, shouldStr)
+
+			if testCase.shouldReject {
```
p0lyn0mial, line 190: I would remove `testCase.shouldReject` as well and update the code to:
```
if err == nil && scenario.expectedError != nil {
 t.Fatal("expected to get an error")
}

if err != nil && scenario.expectedError == nil {
 t.Fatal(err) // unexpected error
}

if err != nil && scenario.expectedError != nil { 
  if err.Error() != scenario.expectedError.Error() {
   t.Fatalf("unexpected error = %v, expected = %v", err, scenario.expectedError)
  }
  
  if strings.Contains(err.Error(), ErrKubeletOutdated.Error()) {
     assert.True(t, errors.Is(err, ErrKubeletOutdated), "error message should be ErrKubeletOutdated"			 
  }
}

```

p0lyn0mial (commented):
Left a few more suggestions that can be addressed in a new PR.

/lgtm

PR #1833, opened 2024-10-14
enhancement: drop the redundant `ApplyUnstructuredResourceImproved` return value

rexagod:
Drop the boolean return value that signified if an update happened as it
is not necessary and pollutes the signature.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

***

_Blocked by (and rebased over) https://github.com/openshift/library-go/pull/1823. The relevant commit is https://github.com/openshift/library-go/pull/1833/commits/12f2bce9c9c28af97c00861845613ae2d3f0622d._

### DIFF
--- a/pkg/operator/resource/resourceapply/monitoring.go
+++ b/pkg/operator/resource/resourceapply/monitoring.go
@@ -22,7 +22,7 @@ var serviceMonitorGVR = schema.GroupVersionResource{Group: "monitoring.coreos.co
 
 // ApplyAlertmanager applies the Alertmanager.
 func ApplyAlertmanager(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, alertmanagerGVR, nil, nil)
+	return ApplyUnstructuredResourceImprovedDeprecated(ctx, client, recorder, required, noCache, alertmanagerGVR, nil, nil)
 }
 
 // DeleteAlertmanager deletes the Alertmanager.
@@ -32,7 +32,7 @@ func DeleteAlertmanager(ctx context.Context, client dynamic.Interface, recorder 
 
 // ApplyPrometheus applies the Prometheus.
 func ApplyPrometheus(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, prometheusGVR, nil, nil)
+	return ApplyUnstructuredResourceImprovedDeprecated(ctx, client, recorder, required, noCache, prometheusGVR, nil, nil)
 }
 
 // DeletePrometheus deletes the Prometheus.
@@ -42,7 +42,7 @@ func DeletePrometheus(ctx context.Context, client dynamic.Interface, recorder ev
 
 // ApplyPrometheusRule applies the PrometheusRule.
 func ApplyPrometheusRule(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, prometheusRuleGVR, nil, nil)
+	return ApplyUnstructuredResourceImprovedDeprecated(ctx, client, recorder, required, noCache, prometheusRuleGVR, nil, nil)
 }
 
 // DeletePrometheusRule deletes the PrometheusRule.
@@ -52,7 +52,7 @@ func DeletePrometheusRule(ctx context.Context, client dynamic.Interface, recorde
 
 // ApplyServiceMonitor applies the ServiceMonitor.
 func ApplyServiceMonitor(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, serviceMonitorGVR, nil, nil)
+	return ApplyUnstructuredResourceImprovedDeprecated(ctx, client, recorder, required, noCache, serviceMonitorGVR, nil, nil)
 }
 
 // DeleteServiceMonitor deletes the ServiceMonitor.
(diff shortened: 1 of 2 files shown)


rexagod:
(cc @jan--f)

test/e2e-monitoring/monitoring_test.go
```diff
@@ -81,15 +117,19 @@ func TestResourceVersionApplication(t *testing.T) {
...
-	unstructuredResource.SetCreationTimestamp(gotUnstructured.GetCreationTimestamp())
-	unstructuredResource.SetUID(gotUnstructured.GetUID())
-	unstructuredResource.SetManagedFields(gotUnstructured.GetManagedFields())
-	require.Equal(t, unstructuredResource.UnstructuredContent(), gotUnstructured.UnstructuredContent())
+	unstructuredResource.SetResourceVersion(gotUnstructuredResource.GetResourceVersion())
+	unstructuredResource.SetGeneration(gotUnstructuredResource.GetGeneration())
+	unstructuredResource.SetCreationTimestamp(gotUnstructuredResource.GetCreationTimestamp())
+	unstructuredResource.SetUID(gotUnstructuredResource.GetUID())
+	unstructuredResource.SetManagedFields(gotUnstructuredResource.GetManagedFields())
+	require.Equal(t, unstructuredResource.UnstructuredContent(), gotUnstructuredResource.UnstructuredContent())
```
rexagod, line 132: Since we are already doing this now, LMK if the `didUpdate` function above feels redundant now. I kept it to visualize the workaround for the `didUpdate` return parameter in the test.

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -72,6 +72,26 @@ func ApplyUnstructuredResourceImproved(
 	resourceGVR schema.GroupVersionResource,
 	defaultingFunc mimicDefaultingFunc,
 	equalityChecker equalityChecker,
+) (*unstructured.Unstructured, error) {
```
rexagod, line 75: @jan--f I believe this satisfies the proposed refactor we talked about earlier?

jan--f:
Thanks, I like the slimmer interface.

/lgtm

rexagod:
> Once this PR has been reviewed and has the lgtm label, please assign [p0lyn0mial](https://github.com/p0lyn0mial) for approval.

/cc @p0lyn0mial

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -32,7 +32,7 @@ func DeleteAlertmanager(ctx context.Context, client dynamic.Interface, recorder
 
 // ApplyPrometheus applies the Prometheus.
 func ApplyPrometheus(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, prometheusGVR, nil, nil)
+	return ApplyUnstructuredResourceImprovedDeprecated(ctx, client, recorder, required, noCache, prometheusGVR, nil, nil)
```
p0lyn0mial, line 35: Okay, what is the plan for switching to `ApplyUnstructuredResourceImproved`?

rexagod:
Friendly ping, @p0lyn0mial for another look here please. I think the deprecation cycle should continue for a couple of releases to allow enough soak time.

pkg/operator/resource/resourceapply/monitoring.go
p0lyn0mial: OK, so the only difference between these two functions is the number of returned values. Am I right?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -32,7 +32,7 @@ func DeleteAlertmanager(ctx context.Context, client dynamic.Interface, recorder
 
 // ApplyPrometheus applies the Prometheus.
 func ApplyPrometheus(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
```
p0lyn0mial, line 34: So your plan is to change the methods, like ApplyAlertmanager, to use ApplyUnstructuredResourceImproved, and at the same time, you would also break compatibility?

pkg/operator/resource/resourceapply/monitoring.go
```diff
@@ -42,7 +42,7 @@ func DeletePrometheus(ctx context.Context, client dynamic.Interface, recorder ev
 
 // ApplyPrometheusRule applies the PrometheusRule.
 func ApplyPrometheusRule(ctx context.Context, client dynamic.Interface, recorder events.Recorder, required *unstructured.Unstructured) (*unstructured.Unstructured, bool, error) {
-	return ApplyUnstructuredResourceImproved(ctx, client, recorder, required, noCache, prometheusRuleGVR, nil, nil)
```
p0lyn0mial, line 45: Also, what is the migration path? Do you know how the redundant parameter is used?

And do you know how many users of ApplyUnstructuredResourceImproved we have? Based on. https://github.com/search?q=org%3Aopenshift%20ApplyUnstructuredResourceImproved&type=code, not that many :) 

Is there a value in introducing ApplyUnstructuredResourceImprovedDeprecated ?
Wouldn't be better to change the methods like ApplyAlertmanager ?

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Unfortunately so, since the `bool` value is [redundant](https://github.com/openshift/library-go/pull/1833/files#diff-c44fd9ab87f08101aadd180c482e8f18c866a305b78c7289a95a518cacdf2203R30).

pkg/operator/resource/resourceapply/monitoring.go
rexagod: Exactly, nothing else.

pkg/operator/resource/resourceapply/monitoring.go
rexagod: > Also, what is the migration path? Do you know how the redundant parameter is used?

The parameter is used to check if the function actually updated the object IIRC. However, this is redundant since that could be done [based on the returned object](https://github.com/openshift/library-go/pull/1833/files#diff-c44fd9ab87f08101aadd180c482e8f18c866a305b78c7289a95a518cacdf2203R30), and hence that parameter does not need to be baked into the existing machinery.

> And do you know how many users of ApplyUnstructuredResourceImproved we have? Based on. [github.com/search?q=org%3Aopenshift%20ApplyUnstructuredResourceImproved&type=code](https://github.com/search?q=org%3Aopenshift+ApplyUnstructuredResourceImproved&type=code&rgh-link-date=2025-01-24T09%3A58%3A57Z), not that many :)

I see. This is in-line with the fact that we ourselves use the wrapper `ApplyX` methods instead downstream. However, the `ApplyUnstructuredResourceImproved` change here breaks these functions too, as you mentioned above, namely in: https://github.com/search?q=org%3Aopenshift+ApplyUnstructuredResourceImproved+OR+ApplyAlertmanager+OR+ApplyPrometheus+OR+ApplyPrometheusRule+OR+ApplyServiceMonitor+owner%3Aopenshift&type=code&rgh-link-date=2025-01-24T09%3A58%3A57Z.

> Is there a value in introducing ApplyUnstructuredResourceImprovedDeprecated ?
> Wouldn't be better to change the methods like ApplyAlertmanager ?

In the future, we want to move the focus away from `ApplyX` functions, as that implies we'd need to keep `library-go` in sync with every arbitrary `X` monitoring CRD that is being used downstream. This essentially creates a loop where an `ApplyX` function isn't available for `X`, so downstream uses `ApplyUnstructuredResourceImproved` for the time being, before switching to `ApplyX` when it gets merged here, which itself will always be nothing more than a caller for it, and it would rid us of the unnecessary maintenance that comes as a side-effect (users will effectively need to `nil` the last two fields to achieve the same effect, as above).

p0lyn0mial: > However, this is redundant since that could be done [based on the returned object](https://github.com/openshift/library-go/pull/1833/files#diff-c44fd9ab87f08101aadd180c482e8f18c866a305b78c7289a95a518cacdf2203R30), and hence that parameter does not need to be baked into the existing machinery.

I think that returning a parameter indicating whether the object was updated is slightly more convenient than checking the object itself. It is not obvious how to do it. In practice, I think it would involve creating a common function to be used by the callers. 

Additionally, the signature of the Apply functions aligns with the signature exposed by the Apply functions for other resources. All return the additional parameter. Here are a few examples:

- [ApplyDeployments](https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L92)
- [ApplyNamespaces](https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/core.go#L61)
- [ApplyClusterRole](https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/rbac.go#L19)


Does it make sense ?

rexagod: It does, and seeing that this has been the convention, it wouldn't be nice to break user expectations. Closing, thank you for your review here!

PR #1837, opened 2024-10-16
add p0lyn0mial to OWNERS

p0lyn0mial:
/assign @bertinatto @deads2k

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,8 @@
 reviewers:
   - bertinatto
   - deads2k
+  - p0lyn0mial
 approvers:
   - bertinatto
   - deads2k
+  - p0lyn0mial


bertinatto (commented):
We @deads2k, @p0lyn0mial and I discussed offline and agreed on this:

/lgtm

deads2k:
Dropping a comment here so we link back to it.

remember that

1. you're the reviewer of last resort, so you may have to learn something pretty quickly
2. it's important to say "no" to things that don't have generic purpose that will force people to use the "one" copy.  Many helpers seem useful, but in practice never gain adoption
3. keep the dep tree small because diamond dependencies from this repo are awful.  it's especially important to avoid vendoring APIs because they depend on alternative levels of [k8s.io/apimachinery](http://k8s.io/apimachinery) that can conflict in ways that prevent kube updates from landing.

PR #1840, opened 2024-10-16
API-1835: add ability to get and list namespaces from must-gather

deads2k:
Needed to feed operators that get, list, watch namespaces.


bertinatto (commented):
/lgtm

PR #1841, opened 2024-10-17
API-1835: use discovery for the discovery endpoint

deads2k:
Updates the manifestclient to use aggregated discovery from https://github.com/openshift/oc/pull/1901 to provide discovery and provide ListKind for list requests.  This lets us have informers sync when we don't have any data for a type.

There's also a compromise in here to have namespaced lists where the namespace isn't in must-gather return empty list. While not true to the kube-apiserver, this allows all informers to sync.


pkg/manifestclient/group_resource_discovery.go
```diff
@@ -4,14 +4,58 @@ import (
...
+	default:
+		// TODO can probably do better
+		return mrt.getLegacyGroupResourceDiscovery(requestInfo)
+	}
+}
+
+func (mrt *manifestRoundTripper) getAggregatedDiscoveryForURL(filename, url string) ([]byte, error) {
+	discoveryBytes, err := fs.ReadFile(mrt.sourceFS, filename)
+	if os.IsNotExist(err) {
+		discoveryBytes, err = fs.ReadFile(defaultDiscovery, filepath.Join("default-discovery", filename))
```
p0lyn0mial, line 40: won't this get us into trouble? The `applycfg` cmd will provide the input directory, which could contain data from the current cluster discovery.

pkg/manifestclient/read_roundtripper.go
```diff
@@ -128,10 +140,94 @@ func isServerGroupResourceDiscovery(path string) bool {
...
 	if len(parts) != 4 {
 		return false
 	}
 	return parts[0] == "" && parts[1] == "apis"
 }
+
+//go:embed default-discovery
+var defaultDiscovery embed.FS
+
+func (mrt *manifestRoundTripper) getKindForResource(gvr schema.GroupVersionResource) (kindData, error) {
```
p0lyn0mial, line 157: `kindData` leaks the `err` filed which is not used outside of this function, can we get that fixed?

pkg/manifestclient/read_roundtripper.go
```diff
@@ -128,10 +140,94 @@ func isServerGroupResourceDiscovery(path string) bool {
...
+				if len(resourceInfo.ResponseKind.Group) > 0 {
+					kindForGVR.kind.Group = resourceInfo.ResponseKind.Group
+				}
+				if len(resourceInfo.ResponseKind.Version) > 0 {
+					kindForGVR.kind.Version = resourceInfo.ResponseKind.Version
+				}
+				kindForGVR.listKind = schema.GroupVersionKind{
+					Group:   kindForGVR.kind.Group,
+					Version: kindForGVR.kind.Version,
+					Kind:    resourceInfo.ResponseKind.Kind + "List",
```
p0lyn0mial, line 222: I think this won't work for CRDs.

pkg/manifestclient/group_resource_discovery.go
deads2k: > won't this get us into trouble? The `applycfg` cmd will provide the input directory, which could contain data from the current cluster discovery.

The logic (I think) is, "prefer the must-gather discovery, if must-gather discovery doesn't exist, use a default".  So it's discovery of last resort.  It let's us work with older must-gathers and incomplete ones.

pkg/manifestclient/read_roundtripper.go
deads2k: > `kindData` leaks the `err` filed which is not used outside of this function, can we get that fixed?

How would you cache the error with out storing it in kindData?  We hit this path thousands of times.

pkg/manifestclient/read_roundtripper.go
deads2k: > I think this won't work for CRDs.

Technically not correct.  Practically the best option we have.  Turns out we don't include ListKind in discovery.

deads2k:
negotiated via slack. We think any misses will be caught during integration testing.  And it would have been nice if new-discovery had included ListKind :(

PR #1845, opened 2024-10-18
API-1835: migrate the installer controller to SSA

p0lyn0mial:
proof pr at https://github.com/openshift/cluster-kube-apiserver-operator/pull/1756

PR #1849, opened 2024-10-22
NO-JIRA: take into account existing operator's status when applying the configuration

p0lyn0mial:
xref: https://github.com/openshift/library-go/pull/1845


deads2k:
merged in https://github.com/openshift/library-go/pull/1854

PR #1851, opened 2024-10-22
API-1835: add generateName for mutation: came up on CSRs

deads2k:
Found in cluster-authentication-operator


bertinatto (commented):
/lgtm

PR #1855, opened 2024-10-22
OCPBUGS-23435: Fix Progressing and Degraded condition during workload rollouts.

atiratree:
Deployment rollout should depend on the Deployment Progressing condition
to asses the state of the deployment as there can be unavailable pods
during the deployment. This is capped by a 15 minute deadline. As a result
- Degraded condition should be False during the deployment rollout.
- Progressing condition should be True  during the deployment rollout.

followup for https://github.com/openshift/library-go/pull/1732 

- fixes https://issues.redhat.com//browse/OCPBUGS-23435
- partially fixes https://issues.redhat.com/browse/OCPBUGS-38675

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -289,7 +289,9 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
-	workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas
+	// Update is done when all pods have been updated to the latest revision
+	// and the deployment controller has reported NewReplicaSetAvailable
+	workloadIsBeingUpdated := !workloadAtHighestGeneration || !hasDeploymentProgressed(workload.Status)
 	workloadIsBeingUpdatedTooLong, err := isUpdatingTooLong(previousStatus, *deploymentProgressingCondition.Type)
 	if !workloadAtHighestGeneration {
 		deploymentProgressingCondition = deploymentProgressingCondition.
@@ -300,8 +302,15 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		deploymentProgressingCondition = deploymentProgressingCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("PodsUpdating").
-			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas))
+			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
 	} else {
+		// Terminating pods don't account for any of the other status fields but
+		// still can exist in a state when they are accepting connections and would
+		// contribute to unexpected behavior when we report Progressing=False.
+		// The case of too many pods might occur for example if `TerminationGracePeriodSeconds` is set
+		//
+		// The workload should ensure this does not happen by using for example EnsureAtMostOnePodPerNode
+		// so that the old pods terminate before the new ones are started.
 		deploymentProgressingCondition = deploymentProgressingCondition.
 			WithStatus(operatorv1.ConditionFalse).
 			WithReason("AsExpected")
@@ -351,6 +360,17 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 func isUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingConditionType string) (bool, error) {
 	progressing := v1helpers.FindOperatorCondition(operatorStatus.Conditions, progressingConditionType)
 	return progressing != nil && progressing.Status == operatorv1.ConditionTrue && time.Now().After(progressing.LastTransitionTime.Add(15*time.Minute)), nil
+}
+
+// hasDeploymentProgressed returns true if the deployment reports NewReplicaSetAvailable
+// via the DeploymentProgressing condition
+func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
+	for _, cond := range status.Conditions {
+		if cond.Type == appsv1.DeploymentProgressing {
+			return cond.Status == corev1.ConditionTrue && cond.Reason == "NewReplicaSetAvailable"
+		}
+	}
+	return false
 }
 
 // EnsureAtMostOnePodPerNode updates the deployment spec to prevent more than
--- a/pkg/operator/apiserver/controller/workload/workload_test.go
+++ b/pkg/operator/apiserver/controller/workload/workload_test.go
@@ -140,6 +140,9 @@ func TestUpdateOperatorStatus(t *testing.T) {
 				},
 				Status: appsv1.DeploymentStatus{
 					AvailableReplicas: 0,
+					Conditions: []appsv1.DeploymentCondition{
+						{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionFalse, LastUpdateTime: metav1.NewTime(time.Now().Add(-6 * time.Minute)), LastTransitionTime: metav1.NewTime(time.Now().Add(-6 * time.Minute)), Reason: "ProgressDeadlineExceeded", Message: "timed out"},
+					},
 				},
 			},
 			pods: []*corev1.Pod{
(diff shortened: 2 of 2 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -353,6 +375,33 @@ func isUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
 	return progressing != nil && progressing.Status == operatorv1.ConditionTrue && time.Now().After(progressing.LastTransitionTime.Add(15*time.Minute)), nil
 }
 
+// hasDeploymentProgressed returns true if the deployment reports NewReplicaSetAvailable
+// via the DeploymentProgressing condition
+func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
```
ardaguclu, line 380: nit:
```suggestion
func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
	for _, cond := range status.Conditions {
		if cond.Type == appsv1.DeploymentProgressing {
			return cond.Status == corev1.ConditionTrue && cond.Reason == "NewReplicaSetAvailable"
		}
	}
	return false
}
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -353,6 +375,33 @@ func isUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
...
+	for i := range status.Conditions {
+		c := status.Conditions[i]
+		if c.Type == appsv1.DeploymentProgressing {
```
ardaguclu, line 383: Would it make sense taking `workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas` into account in addition to this? or `NewReplicaSetAvailable` is already covering?

ardaguclu (commented):
Looks good to me. Dropped 2 unblocker comments.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: yup, that looks better - I have mostly copied the utility functions from k/k :D

pkg/operator/apiserver/controller/workload/workload.go
atiratree: that is already  covered by the condition
- https://github.com/kubernetes/kubernetes/blob/5ffb0528dd0ca2faa9c8c4c360216268926ee0ee/pkg/controller/deployment/progress.go#L60
- https://github.com/kubernetes/kubernetes/blob/5ffb0528dd0ca2faa9c8c4c360216268926ee0ee/pkg/controller/deployment/util/deployment_util.go#L708

There can be a short amount of time when this condition is stale, before the deployment controller observes the new revision - so I made the `workloadIsBeingUpdated` variable depend also on the `workloadAtHighestGeneration` and added a test case for that.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -353,6 +375,32 @@ func isUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
...
+			numberOfTerminatingPods += 1
+		}
+	}
+	return int32(numberOfTerminatingPods)
+}
+
+func isPodTerminating(p *corev1.Pod) bool {
+	return p.Status.Phase != corev1.PodFailed && p.Status.Phase != corev1.PodSucceeded &&
+		p.DeletionTimestamp != nil
+}
```
bertinatto, line 402: Just to make it clear, the pod is considered terminating when it has been marked for deletion but has not yet fully exited? If you happen to push again, could you add a comment for future reference?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -287,9 +287,26 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 	}
 
+	selector, err := metav1.LabelSelectorAsSelector(workload.Spec.Selector)
+	if err != nil {
+		return fmt.Errorf("failed to construct label selector: %v", err)
+	}
+	matchingPods, err := c.podsLister.Pods(c.targetNamespace).List(selector)
+	if err != nil {
+		return err
+	}
```
bertinatto, line 297: Is it guaranteed that workload will always have a selector set?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -287,9 +287,26 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+	// The case of too many pods might occur for example if `TerminationGracePeriodSeconds` is set.
+	terminatingPods := countTerminatingPods(matchingPods)
+
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
-	workloadIsBeingUpdated := workload.Status.UpdatedReplicas < desiredReplicas
+	// update is done when all pods have been updated to the latest revision
+	// and the deployment controller has reported NewReplicaSetAvailable
+	workloadIsBeingUpdated := !workloadAtHighestGeneration || !hasDeploymentProgressed(workload.Status)
+	workloadHasOldTerminatingPods := workload.Status.AvailableReplicas == desiredReplicas && terminatingPods > 0
```
bertinatto, line 309: In theory this math could indicate a terminating pod that's not actually "old", right?

Like, `desiredReplicas == 3` and I have `AvailableReplicas == 3`, but one of the available replicas is an old pod, and one of the terminating pods is not "old".

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -300,7 +317,12 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		deploymentProgressingCondition = deploymentProgressingCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("PodsUpdating").
-			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas))
+			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
+	} else if workloadHasOldTerminatingPods {
+		deploymentProgressingCondition = deploymentProgressingCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("PreviousGenerationPodsPresent").
+			WithMessage(fmt.Sprintf("deployment/%s.%s: %d pod(s) from the previous generation are still present", workload.Name, c.targetNamespace, terminatingPods))
```
bertinatto, line 325: I don't see any changes to the Degraded condition, could you explain how this patch fixes it? Is it somehow affected by this hunk?

bertinatto (commented):
Can you create a proof PR for an operator that makes heavy use of this controller?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: It mostly works (`workload.Status.AvailableReplicas` does not include terminating replica), but yeah there is one edge case. If you delete a new pod and it terminates for a longer time then the new one takes to become available. Then this status would be incorrect.

I originally wanted to bring in the deployment controller logic in, but in the end decided against it. It would require a lot of code for matching pods/hashes and revisions/replicasets. And I would prefer not to maintain this logic.

I have noticed that all of the uses of the WorkloadController use `EnsureAtMostOnePodPerNode` function in the following repos:
- https://github.com/openshift/cluster-authentication-operator
- https://github.com/openshift/cluster-openshift-apiserver-operator

When we depend on the `DeploymentProgressing` condition from now on and `EnsureAtMostOnePodPerNode` function, we first terminate the pod on each node before we create a new one.

So I have removed the terminating pods logic and added a note that this is a user of this controller responsibility to ensure if old terminating pods can be a part of the complete deployment or not. Which is done by using `EnsureAtMostOnePodPerNode` together with this PR. Or it can be also done by a `DeploymentPodReplacementPolicy` feature once it is merged upstream.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: The workload is a deployment and it does not allow an empty (select all) selector. Nevertheless I have removed this logic.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: I have decided against the dependency on the terminating pods.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: The condition resolution is changed by the `workloadIsBeingUpdated` variable which depends on the Deployment's `Progressing` condition now

atiratree:
@bertinatto @ardaguclu  updated the PR and opened the proof PRs:
- https://github.com/openshift/cluster-authentication-operator/pull/739
- https://github.com/openshift/cluster-openshift-apiserver-operator/pull/600

bertinatto (commented):
/lgtm

PR #1857, opened 2024-10-23
API-1835: tremendously increase QPS for manifest client to avoid delay

deads2k:
found while testing authentication operator.

### DIFF
--- a/pkg/manifestclient/readwrite_roundtripper.go
+++ b/pkg/manifestclient/readwrite_roundtripper.go
@@ -5,11 +5,20 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"k8s.io/client-go/rest"
 	"net/http"
 	"os"
 )
 
-// Enter here and call `NewForConfigAndClient(&rest.Config{}, httpClient)`
+// RecommendedRESTConfig is meant to be paired with the HTTPClients below
+func RecommendedRESTConfig() *rest.Config {
+	return &rest.Config{
+		QPS:   1000,
+		Burst: 10000,
+	}
+}
+
+// Enter here and call `NewForConfigAndClient(manifestclient.RecommendedRESTConfig(), httpClient)`
 func NewHTTPClient(mustGatherDir string) MutationTrackingClient {
 	mutationTrackingRoundTripper := newReadWriteRoundTripper(os.DirFS(mustGatherDir))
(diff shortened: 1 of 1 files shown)


pkg/manifestclient/readwrite_roundtripper.go
```diff
@@ -5,11 +5,20 @@ import (
...
 	"io/fs"
+	"k8s.io/client-go/rest"
 	"net/http"
 	"os"
 )
 
-// Enter here and call `NewForConfigAndClient(&rest.Config{}, httpClient)`
+// RecommendedRESTConfig is meant to be paired with the HTTPClients below
+func RecommendedRESTConfig() *rest.Config {
+	return &rest.Config{
```
p0lyn0mial, line 15: Is this configuration required for the fake client, the live client, or both? 
Why can't we use the configuration that is already used by the operator? (IIRC provided by the library-go)

deads2k: > Is this configuration required for the fake client, the live client, or both?
> Why can't we use the configuration that is already used by the operator? (IIRC provided by the library-go)

when wiring to actual operators, it turns out we don't take the input that way.  This is a very natural package and direct peer to the http client that's wired in usage.

p0lyn0mial: I've noticed that the live client [will use ](https://github.com/openshift/cluster-authentication-operator/pull/718/files#diff-a58d662aec11fcd8b98a7666c97bc4de17040abbfed4535d56055b40b78fc2e5R19)the same settings. 
Currently, the QPS settings are configured by the [controllercmd](https://github.com/openshift/library-go/blob/2ba740733b461ee8fe7e95a7d95f06b30ebcaa1d/pkg/controller/controllercmd/builder.go#L428). 

Is there a reason we can't preserve the current settings only for the live client?

deads2k: like cloud providers, external OIDC providers.  We'll probably (definitely) discover that direct kube-apiserver checks actually fail in the general case

deads2k:
We can re-open the discussion on monday, but I'm trying to resolve the all the vendored deps for our first integration test in auth-operator and this makes us go past our timeout.

Basically, because this rest.Config has no connection information in it, you cannot use it to connect somewhere useful.

PR #1858, opened 2024-10-23
API-1835: allow library-go to read must-gathers without any content

deads2k:
found during testing.

/assign @bertinatto


pkg/manifestclient/list.go
```diff
@@ -122,9 +128,12 @@ func (mrt *manifestRoundTripper) listAll(requestInfo *apirequest.RequestInfo) ([
 
 func (mrt *manifestRoundTripper) listAllNamespaces() ([]byte, error) {
 	possibleNamespaceFiles, err := allPossibleNamespaceFiles(mrt.sourceFS)
-	if err != nil {
+	switch {
+	case errors.Is(err, fs.ErrNotExist):
+	case err != nil:
```
bertinatto, line 133: Note to self: previously it'd return the error if it didn't exist, now it does nothing

bertinatto (commented):
/lgtm

PR #1861, opened 2024-10-25
API-1835: manifestclient/write_roundtripper: set the controller name in the annotation if specified in the ctx

p0lyn0mial:
requires https://github.com/openshift/library-go/pull/1860


p0lyn0mial:
/hold

for https://github.com/openshift/library-go/pull/1860

PR #1863, opened 2024-10-28
API-1835: make manifestclient list work when no namespaces are present

deads2k:


bertinatto (commented):
/lgtm

Looks like we forgot to do this in https://github.com/openshift/library-go/pull/1858

PR #1866, opened 2024-10-30
API-1825: revisioncontroller: add controller-instance-name label to secrets and configMaps

bertinatto:
Proof: https://github.com/openshift/cluster-authentication-operator/pull/733

/hold

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -253,6 +253,9 @@ func (c RevisionController) createNewRevision(ctx context.Context, recorder even
 			Annotations: map[string]string{
 				"operator.openshift.io/revision-ready": "false",
 			},
+			Labels: map[string]string{
+				"operator.openshift.io/controller-instance-name": c.controllerInstanceName,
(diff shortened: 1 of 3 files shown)


pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -285,7 +288,17 @@ func (c RevisionController) createNewRevision(ctx context.Context, recorder even
...
+		obj, _, err := resourceapply.SyncConfigMapWithLabels(
+			ctx,
+			c.configMapGetter,
+			recorder,
+			c.targetNamespace,
+			cm.Name,
+			c.targetNamespace,
+			nameFor(cm.Name, revision),
+			ownerRefs,
+			map[string]string{"operator.openshift.io/controller-instance-name": c.controllerInstanceName},
```
deads2k, line 300: nice.  I think this may be unique enough.  We'll need a thing that labels existing revisions to handle an upgrade case too.

deads2k:
worked in https://github.com/openshift/cluster-authentication-operator/pull/733/files!

/lgtm

PR #1869, opened 2024-11-01
API-1835: manifestclient: list items by labelSelector

bertinatto:


pkg/manifestclient/list.go
```diff
@@ -109,6 +117,13 @@ func (mrt *manifestRoundTripper) listAll(requestInfo *apirequest.RequestInfo) ([
 		if retList.Items[0].GroupVersionKind() != kind.kind {
 			return nil, fmt.Errorf("inconsistent item kind: got %v, expected %v", retList.Items[0].GroupVersionKind(), kind.kind)
 		}
+		if requestInfo.LabelSelector != "" {
+			filteredList, err := filterByLabelSelector(retList, requestInfo.LabelSelector)
```
p0lyn0mial, line 121: would also work ?
```
retList, err = filterByLabelSelector(retList, requestInfo.LabelSelector)
```

pkg/manifestclient/read_roundtripper.go
```diff
@@ -6,21 +6,30 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/sets"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/features"
 	"k8s.io/apiserver/pkg/server"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 )
 
+func init() {
+	// This feature gate is needed to set requestInfo.LabelSelector
+	utilruntime.Must(utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=true", features.AuthorizeWithSelectors)))
```
p0lyn0mial, line 30: maybe we should add a new unit test in a new package that would simply use the manifest client with a label selector ? 

for example: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/features/testing/features_init_test.go

pkg/manifestclient/list.go
```diff
@@ -74,6 +75,13 @@ func (mrt *manifestRoundTripper) listAll(requestInfo *apirequest.RequestInfo) ([
 		if retList.GroupVersionKind() != kind.listKind {
 			return nil, fmt.Errorf("inconsistent list kind: got %v, expected %v", retList.GroupVersionKind(), kind.listKind)
 		}
+		if requestInfo.LabelSelector != "" {
```
p0lyn0mial, line 78: nit: by checking the value of the selector inside the `filterByLabelSelector` we would get:
```
retList, err = filterByLabelSelector(retList, requestInfo.LabelSelector)
if err != nil {
 return err
}
```

pkg/manifestclient/list.go
```diff
@@ -74,6 +75,13 @@ func (mrt *manifestRoundTripper) listAll(requestInfo *apirequest.RequestInfo) ([
 		if retList.GroupVersionKind() != kind.listKind {
```
p0lyn0mial, line 75: I know that a label selector can also be specified for watch requests. 
Would it be worth adding a simple test for this case to ensure it won't blow up?

The current implementation doesn't support watching changes on resources.

pkg/manifestclient/list.go
```diff
@@ -270,3 +285,22 @@ func allPossibleNamespaceFiles(sourceFS fs.FS) ([]string, error) {
 
 	return allPossibleListFileLocations, nil
 }
+
+func filterByLabelSelector(list *unstructured.UnstructuredList, labelSelector string) (*unstructured.UnstructuredList, error) {
+	parsedSelector, err := labels.Parse(labelSelector)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse label selector: %v", err)
```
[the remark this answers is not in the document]
bertinatto: Fixed

pkg/manifestclient/list.go
bertinatto: I'm not sure I'm following, but can this be done when we support Watch?

pkg/manifestclient/list.go
bertinatto: Done

pkg/manifestclient/list.go
bertinatto: Done

pkg/manifestclient/list.go
deads2k: We will never support watch.  Watch will always just hold a connection open so that we never encourage live mutation and thus allow interdependence between loops.

pkg/manifestclient/list.go
```diff
@@ -74,6 +75,10 @@ func (mrt *manifestRoundTripper) listAll(requestInfo *apirequest.RequestInfo) ([
 		if retList.GroupVersionKind() != kind.listKind {
 			return nil, fmt.Errorf("inconsistent list kind: got %v, expected %v", retList.GroupVersionKind(), kind.listKind)
 		}
+		retList, err := filterByLabelSelector(retList, requestInfo.LabelSelector)
+		if err != nil {
+			return nil, fmt.Errorf("failed to filter by labelSelector %s: %w", requestInfo.LabelSelector, err)
```
p0lyn0mial, line 80: nit: the error msg already contains the labelSelector.

maybe we should just return the received errr:
```
return nil, err
```

pkg/manifestclient/list.go
bertinatto: We discussed on Slack and agreed that a test for List should be enough, since we don't support Watch. @deads2k let me know if you disagree.

PR #1870, opened 2024-11-04
API-1835: migrate startup monitor conditon to ssa

p0lyn0mial:
/hold

for proof

### DIFF
--- a/pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
+++ b/pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
@@ -13,11 +14,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/utils/ptr"
 )
 
 // startupMonitorPodConditionController monitors the startup monitor pod and reports unwanted state/condition/phase as a degraded condition
 type startupMonitorPodConditionController struct {
-	operatorClient operatorv1helpers.StaticPodOperatorClient
+	controllerInstanceName string
+	operatorClient         operatorv1helpers.StaticPodOperatorClient
 
 	targetName string
 	podLister  corev1listers.PodNamespaceLister
@@ -26,13 +29,14 @@ type startupMonitorPodConditionController struct {
 }
 
 // New returns a controller for monitoring the lifecycle of the startup monitor pod
-func New(targetNamespace string,
-	targetName string,
+func New(
+	instanceName, targetNamespace, targetName string,
 	operatorClient operatorv1helpers.StaticPodOperatorClient,
 	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
 	startupMonitorEnabledFn func() (bool, error),
 	eventRecorder events.Recorder) factory.Controller {
 	fd := &startupMonitorPodConditionController{
+		controllerInstanceName:  factory.ControllerInstanceName(instanceName, "StartupMonitorPodCondition"),
 		operatorClient:          operatorClient,
 		targetName:              targetName,
 		podLister:               kubeInformersForNamespaces.InformersFor(targetNamespace).Core().V1().Pods().Lister().Pods(targetNamespace),
@@ -46,21 +50,21 @@ func New(targetNamespace string,
 			operatorClient.Informer(),
 		).
 		ToController(
-			"StartupMonitorPodCondition", // don't change what is passed here unless you also remove the old FooDegraded condition
+			fd.controllerInstanceName,
 			eventRecorder,
 		)
 }
 
 func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ factory.SyncContext) (err error) {
-	startupPodDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodDegraded", Status: operatorv1.ConditionFalse}
-	startupPodContainerExcessiveRestartsDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerExcessiveRestartsDegraded", Status: operatorv1.ConditionFalse}
+	startupPodDegraded := applyoperatorv1.OperatorCondition().WithType("StartupMonitorPodDegraded")
+	startupPodContainerExcessiveRestartsDegraded := applyoperatorv1.OperatorCondition().WithType("StartupMonitorPodContainerExcessiveRestartsDegraded")
+	status := applyoperatorv1.OperatorStatus()
 
 	defer func() {
 		if err == nil {
-			if _, _, updateError := operatorv1helpers.UpdateStatus(ctx, fd.operatorClient,
-				operatorv1helpers.UpdateConditionFn(*startupPodDegraded),
-				operatorv1helpers.UpdateConditionFn(*startupPodContainerExcessiveRestartsDegraded)); updateError != nil {
-				err = updateError
+			status = status.WithConditions(startupPodDegraded, startupPodContainerExcessiveRestartsDegraded)
+			if applyError := fd.operatorClient.ApplyOperatorStatus(ctx, fd.controllerInstanceName, status); applyError != nil {
+				err = applyError
 			}
 		}
 	}()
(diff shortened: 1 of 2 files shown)


pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -46,21 +49,24 @@ func New(targetNamespace string,
...
+	status := applyoperatorv1.OperatorStatus()
 
 	defer func() {
 		if err == nil {
-			if _, _, updateError := operatorv1helpers.UpdateStatus(ctx, fd.operatorClient,
-				operatorv1helpers.UpdateConditionFn(*startupPodDegraded),
-				operatorv1helpers.UpdateConditionFn(*startupPodContainerExcessiveRestartsDegraded)); updateError != nil {
-				err = updateError
+			status = status.WithConditions(
+				conditionToApplyOperatorCondition(startupPodDegraded),
```
deads2k, line 65: No.  Switch the conditions themselves to applyconfiguration types so we avoid accidents due to conversion.

p0lyn0mial: Good point, reworked the PR.

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -111,71 +115,70 @@ func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ fact
...
 	if len(monitorPod.Status.Reason) > 0 {
 		pendingPodReason = monitorPod.Status.Reason
 	}
 	if len(monitorPod.Status.Message) > 0 {
 		pendingPodMessage = monitorPod.Status.Message
 	}
-	startupPodDegraded.Reason = pendingPodReason
-	startupPodDegraded.Message = fmt.Sprintf("the pod %v has been in %v phase for more than max tolerated time (%v) due to %v", monitorPod.Name, corev1.PodPending, maxPendingDuration, pendingPodMessage)
+	startupPodDegraded = startupPodDegraded.WithReason(pendingPodReason).
+		WithMessage(fmt.Sprintf("the pod %v has been in %v phase for more than max tolerated time (%v) due to %v", monitorPod.Name, corev1.PodPending, maxPendingDuration, pendingPodMessage))
```
bertinatto, line 132: What about moving WithStatus here and set everything at once? Like:

```
startupPodDegraded = startupPodDegraded.
    WithReason(pendingPodReason).
    WithMessage(...))
```

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -111,71 +115,70 @@ func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ fact
 	return nil
 }
 
-func withDegradedWhenInPendingPhaseFor(monitorPod *corev1.Pod, maxPendingDuration time.Duration, startupPodDegraded *operatorv1.OperatorCondition) {
+func withDegradedWhenInPendingPhaseFor(monitorPod *corev1.Pod, maxPendingDuration time.Duration, startupPodDegraded *applyoperatorv1.OperatorConditionApplyConfiguration) {
 	if monitorPod.Status.Phase != corev1.PodPending || monitorPod.Status.StartTime == nil || time.Now().Sub(monitorPod.Status.StartTime.Time) < maxPendingDuration {
 		return
 	}
```
bertinatto, line 121: We'll need to set the status for the `startupPodDegraded` condition here

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -111,71 +115,70 @@ func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ fact
...
 		}
 		return
 	}
 }
 
-func withDegradedWhenHasFailedContainer(monitorPod *corev1.Pod, startupPodDegraded *operatorv1.OperatorCondition) {
+func withDegradedWhenHasFailedContainer(monitorPod *corev1.Pod, startupPodDegraded *applyoperatorv1.OperatorConditionApplyConfiguration) {
 	if monitorPod.Status.Phase != corev1.PodFailed {
 		return
 	}
```
bertinatto, line 156: We'll need to set the status here too

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -111,71 +115,70 @@ func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ fact
...
+func withDegradedWhenHasExcessiveRestarts(monitorPod *corev1.Pod, maxRestartCount int, startupPodContainerExcessiveRestartsDegraded *applyoperatorv1.OperatorConditionApplyConfiguration) {
 	for _, containerStatus := range monitorPod.Status.ContainerStatuses {
 		if int(containerStatus.RestartCount) >= maxRestartCount {
-			startupPodContainerExcessiveRestartsDegraded.Status = operatorv1.ConditionTrue
-			startupPodContainerExcessiveRestartsDegraded.Reason = "ExcessiveRestarts"
-			startupPodContainerExcessiveRestartsDegraded.Message = fmt.Sprintf("at least one container %s in %s pod has restarted %d times, max allowed is %d", containerStatus.Name, monitorPod.Name, containerStatus.RestartCount, maxRestartCount)
+			startupPodContainerExcessiveRestartsDegraded = startupPodContainerExcessiveRestartsDegraded.WithStatus(operatorv1.ConditionTrue).
+				WithReason("ExcessiveRestarts").
+				WithMessage(fmt.Sprintf("at least one container %s in %s pod has restarted %d times, max allowed is %d", containerStatus.Name, monitorPod.Name, containerStatus.RestartCount, maxRestartCount))
 			return
```
bertinatto, line 182: We'll need to re-work this function to make sure `startupPodContainerExcessiveRestartsDegraded` status is set to false when needed

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -46,21 +50,21 @@ func New(targetNamespace string,
...
 			eventRecorder,
 		)
 }
 
 func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ factory.SyncContext) (err error) {
-	startupPodDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodDegraded", Status: operatorv1.ConditionFalse}
-	startupPodContainerExcessiveRestartsDegraded := &operatorv1.OperatorCondition{Type: "StartupMonitorPodContainerExcessiveRestartsDegraded", Status: operatorv1.ConditionFalse}
+	startupPodDegraded := applyoperatorv1.OperatorCondition().WithType("StartupMonitorPodDegraded").WithStatus(operatorv1.ConditionFalse)
+	startupPodContainerExcessiveRestartsDegraded := applyoperatorv1.OperatorCondition().WithType("StartupMonitorPodContainerExcessiveRestartsDegraded").WithStatus(operatorv1.ConditionFalse)
+	status := applyoperatorv1.OperatorStatus()
```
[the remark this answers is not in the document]
bertinatto: Suggestion: if you remove WithStatus here and run unit tests, they will catch a few cases where false was expected, but the field was empty.

bertinatto:
I compared the verb usage around the `kubeapiservers` resource from `e2e-aws-ovn-single-node` job from the proof PR with the same job of https://github.com/openshift/cluster-kube-apiserver-operator/pull/176, which merged a fre days ago. Here's the result:

verb_usage_comparison_apiserver

pkg/operator/staticpod/controller/startupmonitorcondition/startup_monitor_condition.go
```diff
@@ -89,16 +95,24 @@ func (fd *startupMonitorPodConditionController) sync(ctx context.Context, _ fact
...
+		startupPodContainerExcessiveRestartsDegraded = startupPodContainerExcessiveRestartsDegraded.WithStatus(operatorv1.ConditionFalse)
 		return nil
 	} else if err != nil {
 		return err // unknown error
 	}
 
+	// by default, both conditions are in a non-degraded state
+	startupPodDegraded = startupPodDegraded.WithStatus(operatorv1.ConditionFalse)
+	startupPodContainerExcessiveRestartsDegraded = startupPodContainerExcessiveRestartsDegraded.WithStatus(operatorv1.ConditionFalse)
+
```
bertinatto, line 115: nit: Personally, I’d restructure this controller so that every code path sets status explicitly instead of relying on a default value. I think it would make things simpler to have a consistent approach in this controller. What do you think?

bertinatto: To be clear: I don't think this is a big deal and I'm fine with merging as is if you disagree

p0lyn0mial: I would prefer to set the default values here because `startupPodDegraded` can be set to `true` in `withDegradedWhenInPendingPhaseFor` and then we would have to add checks not to revert it to `false` in `withDegradedWhenHasFailedContainer`. Does it make sense ?

bertinatto (commented):
/lgtm

PR #1873, opened 2024-11-08
API-1835: track patches in mutation client

deads2k:
This required several changes

1. metadata file to track the patch type
2. this metadata file is a natural place to store our general metadata to avoid mutating user input
3. the body content isn't necessarily yaml anymore, but most of it is so I didn't change the file format
4. keeping the patch content "natural" allows us to easily test using kubectl
5. pulled controllerInstanceName and fieldManager into metadata to avoid that mutation above

/assign @bertinatto @p0lyn0mial


p0lyn0mial:
CI is not happy. I’m taking over the PR.

PR #1877, opened 2024-11-14
API-1835: operator/events/recorder: allows passing a clock to the event recorder

p0lyn0mial:


deads2k:
/hold

This will allow inconsistent usage of clock wiring, but explicitly requiring the value isn't a huge change, fails at compilation to chase, and providing the value on the `ControllerContext` makes it easy for us to continue proper wiring throughout the stack.

deads2k:
See https://github.com/openshift/library-go/pull/1880 for the required alternative.

PR #1878, opened 2024-11-14
manifestclient: fix discovery of resources in namespaces directory

bertinatto:
/assign @deads2k @p0lyn0mial


deads2k:
How hard is the test to write?

PR #1880, opened 2024-11-14
API-1835: require passive clock for event recorder to avoid accidents

deads2k:
The event recorder requires a clock to make a timestamp.  This **must** be specified and needs to be an affirmative choice so we can ensure that proper clocks are being wired.

In addition, providing a consistent clock on the controllerContext allows all operators to have easy access to a consistent clock.

/assign @p0lyn0mial


bertinatto (commented):
/lgtm

PR #1881, opened 2024-11-15
NO-JIRA: certgraphanalysis: save not-before and not-after, don't export to JSON

vrutkovs:
This is useful for cluster-debug-tools - https://github.com/vrutkovs/cluster-debug-tools/commit/ac958fff43542cdc5c02bdc04933714cc24465fa specifically. Alongside the time secret was created we also log timestamps when certificate is valid for 

These fields are not exported to JSON to avoid leaking these values to TLS registry exports. Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1785

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/certkeypair.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/certkeypair.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
 	"k8s.io/apimachinery/pkg/util/duration"
@@ -108,6 +109,8 @@ func toCertKeyMetadata(certificate *x509.Certificate) certgraphapi.CertKeyMetada
 		},
 		SignatureAlgorithm: certificate.SignatureAlgorithm.String(),
 		PublicKeyAlgorithm: certificate.PublicKeyAlgorithm.String(),
+		NotBefore:          certificate.NotBefore.Format(time.RFC3339),
+		NotAfter:           certificate.NotAfter.Format(time.RFC3339),
 		ValidityDuration:   duration.HumanDuration(certificate.NotAfter.Sub(certificate.NotBefore)),
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/certs/cert-inspection/certgraphanalysis/certkeypair.go
```diff
@@ -108,6 +108,8 @@ func toCertKeyMetadata(certificate *x509.Certificate) certgraphapi.CertKeyMetada
 		},
 		SignatureAlgorithm: certificate.SignatureAlgorithm.String(),
 		PublicKeyAlgorithm: certificate.PublicKeyAlgorithm.String(),
+		NotBefore:          certificate.NotBefore.String(),
```
p0lyn0mial, line 111: aren't we using `time.RFC3339` for formatting ?

pkg/certs/cert-inspection/certgraphapi/types.go
```diff
@@ -157,6 +157,8 @@ type CertKeyMetadata struct {
 	SignatureAlgorithm string
 	PublicKeyAlgorithm string
 	PublicKeyBitSize   string
+	NotBefore          string `json:"-"`
```
p0lyn0mial, line 160: mhm, `CertKeyMetadata` is defined in the library that may have many consumers with different expectations. Wouldn't it be better to allow consumers to control which fields they want to serialise instead ?

pkg/certs/cert-inspection/certgraphanalysis/certkeypair.go
vrutkovs: Good point, updated

pkg/certs/cert-inspection/certgraphanalysis/certkeypair.go
```diff
@@ -108,6 +109,8 @@ func toCertKeyMetadata(certificate *x509.Certificate) certgraphapi.CertKeyMetada
 		},
 		SignatureAlgorithm: certificate.SignatureAlgorithm.String(),
 		PublicKeyAlgorithm: certificate.PublicKeyAlgorithm.String(),
+		NotBefore:          certificate.NotBefore.Format(time.RFC3339),
+		NotAfter:           certificate.NotBefore.Format(time.RFC3339),
```
deads2k, line 113:
```suggestion
		NotAfter:           certificate.NotAfter.Format(time.RFC3339),
```

deads2k:
glad to see cluster-debug-tools still improving.  Probably best not to change serialization and grow unusual dependencies

/lgtm
/approve

PR #1882, opened 2024-11-18
OCPEDGE-1309: feat: add arbiter node check

eggfoobar:
Per EP: https://github.com/openshift/enhancements/pull/1674

This PR adds a check for arbiter node role to guard and node controller for static pods, this is exposed via a `WithArbiter`  builder function to the static pod builder. This should allow folks to decide if their component is designed for the arbiter or not.

### DIFF
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -25,6 +25,7 @@ type NodeController struct {
 	controllerInstanceName string
 	operatorClient         v1helpers.StaticPodOperatorClient
 	nodeLister             corelisterv1.NodeLister
+	extraNodeSelector      labels.Selector
 }
 
 // NewNodeController creates a new node controller.
@@ -33,12 +34,15 @@ func NewNodeController(
 	operatorClient v1helpers.StaticPodOperatorClient,
 	kubeInformersClusterScoped informers.SharedInformerFactory,
 	eventRecorder events.Recorder,
+	extraNodeSelector labels.Selector,
 ) factory.Controller {
 	c := &NodeController{
 		controllerInstanceName: factory.ControllerInstanceName(instanceName, "Node"),
 		operatorClient:         operatorClient,
 		nodeLister:             kubeInformersClusterScoped.Core().V1().Nodes().Lister(),
+		extraNodeSelector:      extraNodeSelector,
 	}
+
 	return factory.New().
 		WithInformers(
 			operatorClient.Informer(),
@@ -65,6 +69,17 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) 
 	nodes, err := c.nodeLister.List(labels.NewSelector().Add(*selector))
 	if err != nil {
 		return err
+	}
+
+	// Due to a design choice on ORing keys in label selectors, we run this query again to allow for additional
+	// selectors as well as selectors that want to OR with master nodes.
+	// see: https://github.com/kubernetes/kubernetes/issues/90549#issuecomment-620625847
+	if c.extraNodeSelector != nil {
+		extraNodes, err := c.nodeLister.List(c.extraNodeSelector)
+		if err != nil {
+			return err
+		}
+		nodes = append(nodes, extraNodes...)
 	}
 
 	jsonPatch := jsonpatch.New()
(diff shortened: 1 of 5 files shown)


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -67,6 +79,14 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		return err
 	}
 
+	if c.arbiterNodeSelector != nil {
+		arbiterNodes, err := c.nodeLister.List(c.arbiterNodeSelector)
```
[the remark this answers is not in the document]
eggfoobar: Ah, that's a good point, unfortunately, this won't work in this situation labels selectors by design do not support an `or` operation on keys ([see](https://github.com/kubernetes/kubernetes/issues/90549#issuecomment-620625847)). So this must be done as separately, otherwise if we were going of values we could build an `contains` query.

pkg/operator/staticpod/controllers.go
```diff
@@ -108,6 +112,9 @@ type Builder interface {
 	WithMinReadyDuration(minReadyDuration time.Duration) Builder
 	WithStartupMonitor(enabledStartupMonitor func() (bool, error)) Builder
 
+	// Informs controllers to handle extra nodes as well as master nodes.
```
p0lyn0mial, line 115: I think that the convention is that the comment should begin with the name of the method.

pkg/operator/staticpod/controllers.go
```diff
@@ -79,6 +80,9 @@ type staticPodOperatorControllerBuilder struct {
 	guardCreateConditionalFunc    func() (bool, bool, error)
 
 	revisionControllerPrecondition revisioncontroller.PreconditionFunc
+
+	// node information
```
p0lyn0mial, line 84: same comment applies here

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -67,6 +71,14 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		return err
 	}
 
+	if c.extraNodeSelector != nil {
```
p0lyn0mial, line 74: could you add a comment explaining that the OR operation is not supported on keys and include a link to the issue you mentioned in the other comment?

pkg/operator/staticpod/controllers.go
```diff
@@ -124,6 +131,11 @@ type Builder interface {
 	ToControllers() (manager.ControllerManager, error)
 }
 
+func (b *staticPodOperatorControllerBuilder) WithExtraNodeSelector(extraNodeSelector labels.Selector) Builder {
```
p0lyn0mial, line 134: What about other controllers potentially listing the master nodes? As I understand, we will only run the etcd member on the arbiter node. Is that correct?

p0lyn0mial:
While reviewing this PR I found that the master nodes selector is being parsed on each sync for some controllers and decided to open https://github.com/openshift/library-go/pull/1910, PTAL.

pkg/operator/staticpod/controller/node/node_controller.go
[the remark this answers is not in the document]
p0lyn0mial: ok, could you add a comment explaining that the OR operation is not supported on keys and include a link to the issue you mentioned in the other comment?

pkg/operator/staticpod/controllers.go
eggfoobar: Yup, you're right, updating

pkg/operator/staticpod/controllers.go
eggfoobar: Oh, this was just following the convention in the other groupings in this struct, but since it's also one un-exported variable, I'll just remove the comment

p0lyn0mial:
also, what is our e2e strategy for this feature? Is there a test in the origin that will add the arbiter node?

pkg/operator/staticpod/controllers.go
eggfoobar: Right, exactly, so unless explicitly required the other controllers should run only on masters, but if they come to the conclusion they need to also run on the arbiter, we can have a path for them to do it

eggfoobar:
> also, what is our e2e strategy for this feature? Is there a test in the origin that will add the arbiter node?

Yup! We will have tests added into origin for OLA validation, as well as CI lanes that do the install. This feature will be TechPreview in 4.19 and GA in 4.20 if all the CI signaling goes to plan.

p0lyn0mial:
thanks!

/lgtm

PR #1884, opened 2024-11-18
API-1835: make mutation filenames more stable

deads2k:
Using request number means that adding one request will cause hundreds of files to shift names.  By using hashes, only the impacted requests changes.

/assign @bertinatto


bertinatto (commented):
/lgtm

PR #1887, opened 2024-11-19
API1835: expose WithControllerInstanceName for controllers under pkg/operator

p0lyn0mial:


pkg/operator/apiserver/controller/auditpolicy/auditpolicy_controller.go
```diff
@@ -54,7 +54,7 @@ func NewAuditPolicyController(
 		targetConfigMapName:    targetConfigMapName,
 	}
 
-	return factory.New().WithSync(c.sync).ResyncEvery(10*time.Second).WithInformers(
+	return factory.New().WithSync(c.sync).WithControllerInstanceName(c.controllerInstanceName).ResyncEvery(10*time.Second).WithInformers(
```
deads2k, line 57: every 10s.  holy cow!

let's get a separate PR to fix this

PR #1888, opened 2024-11-21
API-1835: manifestclient: allow listing only cluster-scoped-resources

p0lyn0mial:
a best effort to allowlist only cluster-scoped resources.


pkg/manifestclient/list.go
```diff
@@ -241,7 +241,7 @@ func allPossibleListFileLocations(sourceFS fs.FS, requestInfo *apirequest.Reques
 
 		namespaces, err := allNamespacesWithData(sourceFS)
 		if err != nil {
```
deads2k, line 243: is this a not found error in your case?  If so, how about we ignore only that error.

p0lyn0mial: eh, it turns out that we already have a test that lists `cluster-scoped-resources` without the namespace directory.

The test works because `cluster-scoped-resources` are expected to be nested in a directory.

For example: `cluster-scoped-resources/operator.openshift.io/authentications/cluster.yaml`
NOT: `cluster-scoped-resources/operator.openshift.io/authentications.yaml`

I copy-pasted the directory structure from MOM, which apparently requires a fix. I will issue a separate PR for that.

p0lyn0mial: heh, okay, actually both are valid and can be found in a must-gather dir.

pkg/manifestclient/list.go
```diff
@@ -240,7 +240,10 @@ func allPossibleListFileLocations(sourceFS fs.FS, requestInfo *apirequest.Reques
 		allPossibleListFileLocations = append(allPossibleListFileLocations, filepath.Join(clusterParts...))
 
 		namespaces, err := allNamespacesWithData(sourceFS)
-		if err != nil {
+		switch {
+		case errors.Is(err, fs.ErrNotExist):
+			return allPossibleListFileLocations, nil
+		case err != nil:
```
bertinatto, line 246: Makes sense to me. So IIRC this branch takes all possible file locations for all resources, namespaced or not.
Previously, we scraped `cluster-scoped-resources` and then tried to scrape all existing namespace directories. If the namespace directories didn’t exist, it would throw an error and stop. Now, it just returns all the file locations it knows about so far without failing.

bertinatto (commented):
/lgtm

PR #1889, opened 2024-11-21
OCPBUGS-44842: certrotation: set not-before/not-after annotations

vrutkovs:
This ensures every secret managed by this controller has valid
annotation set. Instead of analyzing potentially sensitive secret
on customer cluster we should be able to tell if the certificate
rotation didn't happen and certificate expired by looking into
annotations


ibihim:
Technically, the wiring is fine, it makes sense to add `NotBefore` and `NotAfter` to `EnsureTLSMetadataUpdate`.

But the consequence is,
- that we need to invoke it twice and
- usually set the `AdditionalAnnotations` hand it down and execute it, while we could easily just set it in the given Secret.

To leverage `EnsureTLSMetadataUpdate`, we would need to split cert creation from injecting it into the Secret or create a more general `EnsureSecretUpdate` function that expects a TLS cert and then updates the metadata at once.

ibihim (commented):
Technically, the wiring is fine, it makes sense to add `NotBefore` and `NotAfter` to `EnsureTLSMetadataUpdate`.

But the consequence is,
- that we need to invoke it twice and
- usually set the `AdditionalAnnotations` hand it down and execute it, while we could easily just set it in the given Secret.

To leverage `EnsureTLSMetadataUpdate`, we would need to split cert creation from injecting it into the Secret or create a more general `EnsureSecretUpdate` function that expects a TLS cert and then updates the metadata at once.

vrutkovs:
>that we need to invoke it twice

Yes, its unfortunate but see below

>we could easily just set it in the given Secret.

It would require us to copy paste the code which does:
* make sure that annotations map is initialized (for newly created secrets)
* track annotation name (it may change in the future)
* this code can be reused in other components (e.g. ckoa PR uses AdditionalAnnotations to set not-before/not-after [here](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1768/files#diff-0195fcf95e1cbbc734d0c4159ae1851144006bd0022d3a04d0818e90b5e7831aR174-R177) as `ApplySecret` later already runs `EnsureTLSMetadata`)

pkg/operator/csr/cert_controller.go
```diff
@@ -179,6 +181,12 @@ func (c *clientCertificateController) sync(ctx context.Context, syncCtx factory.
 			newSecretConfig[k] = v
 		}
 		secret.Data = newSecretConfig
+
+		// Update not-before/not-after annotations
+		c.AdditionalAnnotations.NotBefore = leaf.NotBefore.Format(time.RFC3339)
+		c.AdditionalAnnotations.NotAfter = leaf.NotAfter.Format(time.RFC3339)
+		_ = c.AdditionalAnnotations.EnsureTLSMetadataUpdate(&secret.ObjectMeta)
```
jsafrane, line 188: Should there be an unit test for this?

vrutkovs: Good idea, added

PR #1891, opened 2024-11-21
API-1835: manifestclient: remove unused discovery code

bertinatto:
It seems like we can't reach this code path.

/assign @deads2k


deads2k:
good move for future us

/lgtm
/approve

PR #1892, opened 2024-11-25
API-1835: manifestclient: set GVK on patch response

p0lyn0mial:
xref: https://github.com/openshift/multi-operator-manager/pull/49#discussion_r1851718069

### DIFF
--- a/pkg/manifestclient/write_roundtripper.go
+++ b/pkg/manifestclient/write_roundtripper.go
@@ -28,18 +28,21 @@ type writeTrackingRoundTripper struct {
 	// requestInfoResolver is the same type constructed the same way as the kube-apiserver
 	requestInfoResolver *apirequest.RequestInfoFactory
 
+	getKindForResourceFunc func(gvr schema.GroupVersionResource) (kindData, error)
+
 	lock              sync.RWMutex
 	nextRequestNumber int
 	actionTracker     *AllActionsTracker[TrackedSerializedRequest]
 }
 
-func newWriteRoundTripper() *writeTrackingRoundTripper {
+func newWriteRoundTripper(getKindForResourceFunc func(gvr schema.GroupVersionResource) (kindData, error)) *writeTrackingRoundTripper {
 	return &writeTrackingRoundTripper{
 		nextRequestNumber: 1,
 		actionTracker:     &AllActionsTracker[TrackedSerializedRequest]{},
 		requestInfoResolver: server.NewRequestInfoResolver(&server.Config{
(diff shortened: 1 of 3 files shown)


pkg/manifestclient/write_roundtripper.go
```diff
@@ -28,18 +28,21 @@ type writeTrackingRoundTripper struct {
...
 
+	getKindForResourceFunc func(gvr schema.GroupVersionResource) (kindData, error)
+
 	lock              sync.RWMutex
 	nextRequestNumber int
 	actionTracker     *AllActionsTracker[TrackedSerializedRequest]
 }
 
-func newWriteRoundTripper() *writeTrackingRoundTripper {
+func newWriteRoundTripper(getKindForResourceFunc func(gvr schema.GroupVersionResource) (kindData, error)) *writeTrackingRoundTripper {
```
deads2k, line 38: don't pass a function here.  just hardcode the right thing.

pkg/manifestclient/write_roundtripper.go
```diff
@@ -213,8 +216,14 @@ func (mrt *writeTrackingRoundTripper) roundTrip(req *http.Request) ([]byte, erro
 	ret := &unstructured.Unstructured{Object: map[string]interface{}{}}
 	ret.SetName(serializedRequest.ActionMetadata.Name)
 	ret.SetNamespace(serializedRequest.ActionMetadata.Namespace)
-	if actionHasRuntimeObjectBody { // TODO might be able to do something generally based on discovery if absolutely necessary
+	if actionHasRuntimeObjectBody {
 		ret.SetGroupVersionKind(bodyObj.GetObjectKind().GroupVersionKind())
+	} else {
+		kindForResource, err := mrt.getKindForResourceFunc(gvr)
```
deads2k, line 222: right here, we should be able to use discovery to determine the GVK, same as we do for reading the list kind.

pkg/manifestclient/write_roundtripper.go
```diff
@@ -28,18 +28,21 @@ type writeTrackingRoundTripper struct {
 	// requestInfoResolver is the same type constructed the same way as the kube-apiserver
 	requestInfoResolver *apirequest.RequestInfoFactory
 
+	getKindForResourceFunc func(gvr schema.GroupVersionResource) (kindData, error)
```
deads2k, line 31: it would be prettier to tak ea discovery client here.

pkg/manifestclient/write_roundtripper.go
p0lyn0mial: done, ptal.

p0lyn0mial:
/remove-tide/merge-method-squash

PR #1901, opened 2024-12-03
OCPBUGS-45321:enabling file protocol in git by default to fix command failures

prabhapa:
git commands to create submodules and clones are failing because file protocol is not allowed in git by default. File protocol is required to be enabled in some unit test cases . Hence through this change its allowed by default . Changing this source package is a best approach in comparison to other complicated changes/ways of fixing it. OCPBUGS-44243 is the original bug raised in 4.18


adambkaplan:
/close

We decided to disable the builder unit tests that relied on the `file:` protocol

PR #1902, opened 2024-12-03
AUTH-543: Add optional operand deletion condition

liouk:
There might be cases (as demonstrated in [OCPBUGS-44937](https://issues.redhat.com/browse/OCPBUGS-44937)) where we might want to gracefully delete the operand workload of the workload controller, and keep the operator status available (instead of unavailable or degraded).

This PR adds an optional way of specifying a deletion condition which will trigger the deletion of the operand gracefully, keeping the operator's status as `Available=True`. To avoid breaking changes, I've added new setup functions to be used for wiring this particular case.

This is needed in the scope of https://github.com/openshift/cluster-authentication-operator/pull/740.

Proof PRs:
- https://github.com/openshift/cluster-authentication-operator/pull/772 (incl. usage demo for external OIDC usecase)
- https://github.com/openshift/cluster-openshift-apiserver-operator/pull/618

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -14,11 +16,11 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
-	"strings"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	openshiftconfigclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+	"github.com/openshift/library-go/pkg/apiserver/jsonpatch"
 	"github.com/openshift/library-go/pkg/apps/deployment"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -33,7 +35,25 @@ const (
 // Delegate captures a set of methods that hold a custom logic
 type Delegate interface {
 	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
-	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, []error)
+	//
+	// It returns a reference to the workload, a bool indicating whether the operator config is at the highest
+	// generation, a bool indicating whether the workload references must be removed from the operator status
+	// (e.g. in case the workload has been removed), two strings indicating the name & namespace
+	// of the workload that must be cleaned up from the operator status (respective to the conditions to be
+	// removed) and a list of errors.
+	//
+	// When a workload is removed (as indicated by the removeWorkload boolean), not only its respective
+	// status conditions will removed, but also its generations, and its version string will be
+	// set to "". To make sure that the StatusSyncer removes any versions that are equal to the empty
+	// string completely, use StatusSyncer.WithEmptyVersionRemoval().
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (
+		delegateWorkload *appsv1.Deployment,
+		operatorConfigAtHighestGeneration bool,
+		removeWorkload bool,
+		removedWorkloadName string,
+		removedWorkloadNamespace string,
+		errs []error,
+	)
 
 	// PreconditionFulfilled a method that indicates whether all prerequisites are met and we can Sync.
 	//
@@ -81,7 +101,7 @@ func NewController(instanceName, operatorNamespace, targetNamespace, targetOpera
 	kubeClient kubernetes.Interface,
 	podLister corev1listers.PodLister,
 	informers []factory.Informer,
-	tagetNamespaceInformers []factory.Informer,
+	targetNamespaceInformers []factory.Informer,
 	delegate Delegate,
 	openshiftClusterConfigClient openshiftconfigclientv1.ClusterOperatorInterface,
 	eventRecorder events.Recorder,
@@ -100,11 +120,11 @@ func NewController(instanceName, operatorNamespace, targetNamespace, targetOpera
 		delegate:                     delegate,
 		openshiftClusterConfigClient: openshiftClusterConfigClient,
 		versionRecorder:              versionRecorder,
-		queue:                        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), instanceName),
+		queue:                        workqueue.NewNamedRateLimitingQueue(workqueue.DefaultTypedControllerRateLimiter[any](), instanceName),
 	}
 
 	c := factory.New()
-	for _, nsi := range tagetNamespaceInformers {
+	for _, nsi := range targetNamespaceInformers {
 		c.WithNamespaceInformer(nsi, targetNamespace)
 	}
 
@@ -128,14 +148,14 @@ func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncCon
 	}
 
 	if fulfilled, err := c.delegate.PreconditionFulfilled(ctx); err != nil {
-		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, []error{err})
+		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, false, "", "", []error{err})
 	} else if !fulfilled {
-		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, nil)
+		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, false, "", "", nil)
 	}
 
-	workload, operatorConfigAtHighestGeneration, errs := c.delegate.Sync(ctx, controllerContext)
+	workload, operatorConfigAtHighestGeneration, removeWorkload, removedWorkloadName, removedWorkloadNamespace, errs := c.delegate.Sync(ctx, controllerContext)
 
-	return c.updateOperatorStatus(ctx, operatorStatus, workload, operatorConfigAtHighestGeneration, true, errs)
+	return c.updateOperatorStatus(ctx, operatorStatus, workload, operatorConfigAtHighestGeneration, true, removeWorkload, removedWorkloadName, removedWorkloadNamespace, errs)
 }
 
 // shouldSync checks ManagementState to determine if we can run this operator, probably set by a cluster administrator.
@@ -156,48 +176,78 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
 	}
 }
 
-// updateOperatorStatus updates the status based on the actual workload and errors that might have occurred during synchronization.
-func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *operatorv1.OperatorStatus, workload *appsv1.Deployment, operatorConfigAtHighestGeneration bool, preconditionsReady bool, errs []error) (err error) {
+func (c *Controller) updateOperatorStatus(ctx context.Context,
+	previousStatus *operatorv1.OperatorStatus,
+	workload *appsv1.Deployment,
+	operatorConfigAtHighestGeneration, preconditionsReady, removeWorkload bool,
+	removedWorkloadName, removedWorkloadNamespace string,
+	errs []error,
+) (err error) {
+
 	if errs == nil {
 		errs = []error{}
 	}
 
-	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable))
+	typeAvailable := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	if removeWorkload {
+		if len(removedWorkloadName) == 0 || len(removedWorkloadNamespace) == 0 {
+			return kerrors.NewAggregate(append(errs, fmt.Errorf("workload marked as removed but no name and/or namespace provided (name=%s, ns=%s)", removedWorkloadName, removedWorkloadNamespace)))
+		}
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
-
-	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
-
-	status := applyoperatorv1.OperatorStatus()
-	if workload != nil {
-		// The Hash field is not required since the LastGeneration field is enough to uniquely identify a Deployment's desired state
-		status = status.WithGenerations(applyoperatorv1.GenerationStatus().
-			WithGroup("apps").
-			WithResource("deployments").
-			WithNamespace(workload.Namespace).
-			WithName(workload.Name).
-			WithLastGeneration(workload.Generation),
+		// the workload has been removed; remove conditions, generations and version
+		patch := jsonpatch.Merge(
+			v1helpers.RemoveConditionsJSONPatch(previousStatus, []string{typeAvailable, typeDegraded, typeProgressing, typeWorkloadDegraded}),
+			v1helpers.RemoveWorkloadGenerationsJSONPatch(previousStatus, removedWorkloadName, removedWorkloadNamespace),
 		)
+
+		if !patch.IsEmpty() {
+			c.setVersion(removedWorkloadName, "")
+			patchErr := c.operatorClient.PatchOperatorStatus(ctx, patch)
+			if patchErr != nil {
+				errs = append(errs, patchErr)
+			}
+		}
+
+		return kerrors.NewAggregate(errs)
 	}
 
-	defer func() {
-		status = status.WithConditions(
-			deploymentAvailableCondition,
-			deploymentDegradedCondition,
-			deploymentProgressingCondition,
-			workloadDegradedCondition,
-		)
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
 
-		if applyError := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, status); applyError != nil {
-			err = applyError
-		}
-	}()
+	// update workload conditions, generations and version
+	statusApplyConfig := applyoperatorv1.OperatorStatus()
+	statusApplyConfig = c.updateOperatorStatusConditions(previousStatus, statusApplyConfig, workload, preconditionsReady, deploymentAvailableCondition, deploymentDegradedCondition, deploymentProgressingCondition, workloadDegradedCondition, errs)
+	statusApplyConfig = c.updateOperatorStatusGenerationsVersion(statusApplyConfig, workload, operatorConfigAtHighestGeneration, preconditionsReady)
+
+	applyErr := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, statusApplyConfig)
+	if applyErr != nil {
+		errs = append(errs, applyErr)
+	}
+
+	return kerrors.NewAggregate(errs)
+}
+
+func (c *Controller) updateOperatorStatusConditions(
+	previousStatus *operatorv1.OperatorStatus,
+	statusApplyConfig *applyoperatorv1.OperatorStatusApplyConfiguration,
+	workload *appsv1.Deployment,
+	preconditionsReady bool,
+	deploymentAvailableCondition, deploymentDegradedCondition, deploymentProgressingCondition, workloadDegradedCondition *applyoperatorv1.OperatorConditionApplyConfiguration,
+	errs []error,
+) *applyoperatorv1.OperatorStatusApplyConfiguration {
+
+	defer statusApplyConfig.WithConditions(
+		deploymentAvailableCondition,
+		deploymentDegradedCondition,
+		deploymentProgressingCondition,
+		workloadDegradedCondition,
+	)
 
 	if !preconditionsReady {
 		var message string
@@ -228,7 +278,7 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("PreconditionNotFulfilled").
 			WithMessage(message)
 
-		return kerrors.NewAggregate(errs)
+		return statusApplyConfig
 	}
 
 	if len(errs) > 0 {
@@ -267,7 +317,7 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("NoDeployment").
 			WithMessage(message)
 
-		return kerrors.NewAggregate(errs)
+		return statusApplyConfig
 	}
 
 	if workload.Status.AvailableReplicas == 0 {
@@ -336,22 +386,55 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("AsExpected")
 	}
 
+	return statusApplyConfig
+}
+
+func (c *Controller) updateOperatorStatusGenerationsVersion(
+	statusApplyConfig *applyoperatorv1.OperatorStatusApplyConfiguration,
+	workload *appsv1.Deployment,
+	operatorConfigAtHighestGeneration, preconditionsReady bool,
+) *applyoperatorv1.OperatorStatusApplyConfiguration {
+
+	if workload == nil {
+		return statusApplyConfig
+	}
+
+	statusApplyConfig = statusApplyConfig.WithGenerations(applyoperatorv1.GenerationStatus().
+		WithGroup("apps").
+		WithResource("deployments").
+		WithNamespace(workload.Namespace).
+		WithName(workload.Name).
+		WithLastGeneration(workload.Generation),
+	)
+
+	if !preconditionsReady {
+		return statusApplyConfig
+	}
+
+	desiredReplicas := int32(1)
+	if workload.Spec.Replicas != nil {
+		desiredReplicas = *(workload.Spec.Replicas)
+	}
+
+	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
+	workloadHasAllPodsAvailable := workload.Status.AvailableReplicas >= desiredReplicas
+
 	// if the deployment is all available and at the expected generation, then update the version to the latest
 	// when we update, the image pull spec should immediately be different, which should immediately cause a deployment rollout
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
 	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
 	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
-		operandName := workload.Name
-		if len(c.operandNamePrefix) > 0 {
-			operandName = fmt.Sprintf("%s-%s", c.operandNamePrefix, workload.Name)
-		}
-		c.versionRecorder.SetVersion(operandName, c.targetOperandVersion)
+		c.setVersion(workload.Name, c.targetOperandVersion)
 	}
 
-	if len(errs) > 0 {
-		return kerrors.NewAggregate(errs)
+	return statusApplyConfig
+}
+
+func (c *Controller) setVersion(operandName, version string) {
+	if len(c.operandNamePrefix) > 0 {
+		operandName = fmt.Sprintf("%s-%s", c.operandNamePrefix, operandName)
 	}
-	return nil
+	c.versionRecorder.SetVersion(operandName, version)
 }
 
 // hasDeploymentProgressed returns true if the deployment reports NewReplicaSetAvailable
--- a/pkg/operator/apiserver/controller/workload/workload_test.go
+++ b/pkg/operator/apiserver/controller/workload/workload_test.go
@@ -3,23 +3,25 @@ package workload
 import (
 	"context"
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"testing"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/diff"
 	corev1listers "k8s.io/client-go/listers/core/v1"
+	clocktesting "k8s.io/utils/clock/testing"
 	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/apiserver/jsonpatch"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
-	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/diff"
 )
 
 const (
@@ -36,6 +38,9 @@ type testDelegate struct {
 	// for Sync
 	syncWorkload            *appsv1.Deployment
 	syncIsAtHighestRevision bool
+	operandWorkloadDeleted  bool
+	deletedOperandName      string
+	deletedOperandNamespace string
 	syncErrrors             []error
 }
 
@@ -43,8 +48,8 @@ func (d *testDelegate) PreconditionFulfilled(_ context.Context) (bool, error) {
 	return d.preconditionReady, d.preconditionErr
 }
 
-func (d *testDelegate) Sync(_ context.Context, _ factory.SyncContext) (*appsv1.Deployment, bool, []error) {
-	return d.syncWorkload, d.syncIsAtHighestRevision, d.syncErrrors
+func (d *testDelegate) Sync(_ context.Context, _ factory.SyncContext) (*appsv1.Deployment, bool, bool, string, string, []error) {
+	return d.syncWorkload, d.syncIsAtHighestRevision, d.operandWorkloadDeleted, d.deletedOperandName, d.deletedOperandNamespace, d.syncErrrors
 }
 
 func TestUpdateOperatorStatus(t *testing.T) {
@@ -55,11 +60,18 @@ func TestUpdateOperatorStatus(t *testing.T) {
 		pods                            []*corev1.Pod
 		operatorConfigAtHighestRevision bool
 		operatorPreconditionsNotReady   bool
-		preconditionError               error
-		errors                          []error
-		previousConditions              []operatorv1.OperatorCondition
 
-		validateOperatorStatus func(*operatorv1.OperatorStatus) error
+		operatorOperandWorkloadDeleted bool
+		deletedOperandName             string
+		deletedOperandNamespace        string
+
+		preconditionError   error
+		errors              []error
+		previousConditions  []operatorv1.OperatorCondition
+		previousGenerations []operatorv1.GenerationStatus
+
+		validateOperatorStatus          func(*operatorv1.OperatorStatus) error
+		validateOperatorStatusJSONPatch func(*jsonpatch.PatchSet) error
 	}{
 		{
 			name: "scenario: no workload, no errors thus we are degraded and we are progressing",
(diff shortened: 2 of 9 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -356,6 +424,51 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+
+	status := applyoperatorv1.OperatorStatus()
+	defer func() {
+		status = status.WithConditions(deploymentAvailableCondition, workloadDegradedCondition)
+		if applyError := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, status); applyError != nil {
+			err = applyError
+		}
+	}()
+
+	if _, err := c.deploymentLister.Deployments(c.targetNamespace).Get(workloadName); err != nil && !apierrors.IsNotFound(err) {
```
everettraven, line 442: Is the `Get()` call necessary before attempting the `Delete()` call? What does the extra call buy us?

pkg/operator/apiserver/controller/workload/workload.go
everettraven, line 1: Would it be beneficial to add some unit tests for the new deletion behavior?

pkg/operator/apiserver/controller/workload/workload.go
liouk: The `Get()` happens on the lister (cache), which means that if the deployment has already been deleted, it'll save us an extra API call to `Delete()`.

pkg/operator/apiserver/controller/workload/workload.go
liouk: The unit tests would end up using mocks for most of the stuff that the deletion does; but on second thought it might be beneficial to test the operator status, so I'll add some :+1:

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -356,6 +424,55 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	return nil
 }
 
+func (c *Controller) deleteWorkload(ctx context.Context, workloadName string) (err error) {
```
atiratree, line 427: Do we need to split it from the `updateOperatorStatus` reconciliation logic? It would be nicer to keep the concern in the same place and consider all 4 conditions.

I do not have a problem with calling a util function in `updateOperatorStatus` if needed though.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -356,6 +424,55 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+		status = status.WithConditions(deploymentAvailableCondition, workloadDegradedCondition)
+		if applyError := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, status); applyError != nil {
+			err = applyError
+		}
+	}()
+
+	if _, getErr := c.deploymentLister.Deployments(c.targetNamespace).Get(workloadName); getErr != nil && !apierrors.IsNotFound(getErr) {
+		deploymentAvailableCondition = deploymentAvailableCondition.
+			WithStatus(operatorv1.ConditionFalse).
+			WithReason("DeletionError")
```
atiratree, line 445: nit: this isn't a deletion error

everettraven, line 445: Should we include a message in all conditions? I'm not familiar with how these conditions have been constructed in the past, but I have seen logs in the past that say not setting the message in a condition will eventually be fatal:
```
W1203 04:13:07.671272       1 dynamic_operator_client.go:355] .status.conditions["APIServerDeploymentAvailable"].message is missing; this will eventually be fatal
```

This was pulled from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/periodic-ci-openshift-multiarch-master-nightly-4.18-ocp-e2e-ovn-remote-libvirt-s390x/1863791759423705088/artifacts/ocp-e2e-ovn-remote-libvirt-s390x/gather-extra/artifacts/pods/openshift-apiserver-operator_openshift-apiserver-operator-dfc988b89-rsh2n_openshift-apiserver-operator.log

Even though it would be a repetitive message in this case, with ^ in mind maybe it is worth it to future-proof this implementation as much as we can for when that does get flipped to fatal?

pkg/operator/apiserver/controller/workload/workload.go
everettraven: Ah, I totally missed that the `Get()` would be on the cache - that makes sense :). Thanks!

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -356,6 +424,55 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+				WithMessage(deleteErr.Error())
+			return deleteErr
+		}
+	}
+
+	deploymentAvailableCondition = deploymentAvailableCondition.
+		WithStatus(operatorv1.ConditionTrue).
+		WithReason("AsExpected")
+	workloadDegradedCondition = workloadDegradedCondition.
+		WithStatus(operatorv1.ConditionFalse)
```
atiratree, line 471: What is the value of preserving these conditions (`Available=True`, `Degraded=False`) if no workload exists for them?

Or are we interested in measuring the availability of OIDC provider? Is it possible? And even then this probably isn't the right place to indicate it, right?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: It would be good to import and use this change in one of the consumers (e.g. cluster-authentication-operator) and show that the CI is passing there.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +227,29 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+	if removeConditions {
+		jsonPatch := v1helpers.RemoveConditionsJSONPatch(previousStatus, []string{typeAvailable, typeDegraded, typeProgressing, typeWorkloadDegraded})
```
atiratree, line 246: Seems we cannot use SSA for removing conditions, but I am not sure if patch is better than `v1helpers.UpdateStatus` here. Would be good to add an additional opinion on this.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +227,29 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+	if removeConditions {
```
atiratree, line 245: we should check if `workload == nil` to make sure the delegate has deleted (and not recreated) the workload

atiratree, line 245: even the workload should be gone and we should not get any errs, we should at least log them if they passed on by the delegate

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +227,29 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
```
atiratree, line 244: It might make more sense to still consider preconditions even when the workload will be deleted later. Thoughts?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -129,6 +189,14 @@ func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncCon
 		return err
 	}
 
+	if c.deletionConditionFn != nil {
+		if conditionMet, workload, err := c.deletionConditionFn(); err != nil {
+			return err
+		} else if conditionMet {
+			return c.deleteWorkload(ctx, workload)
```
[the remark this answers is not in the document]
atiratree: This still applies, why do we have to manage the deletion in here?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -83,8 +89,51 @@ func NewController(instanceName, operatorNamespace, targetNamespace, targetOpera
...
+// the "instanceName" param will be used to set conditions in the status field. It will be suffixed with "WorkloadController",
+// so it can end up in the condition in the form of "OAuthAPIWorkloadControllerDeploymentAvailable"
+//
+// the "operatorNamespace" is used to set "version-mapping" in the correct namespace
+//
+// the "targetNamespace" represent the namespace for the managed resource (DaemonSet)
+//
+// the "deletionConditionFn" will be used to check whether the workload specified by the
+// returned name which is part of targetNamespace must be deleted
+func NewControllerWithDeletion(instanceName, operatorNamespace, targetNamespace, targetOperandVersion, operandNamePrefix, conditionsPrefix string,
```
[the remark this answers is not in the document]
atiratree: Still applies, I think we do not need any change in the constructor. Let's drive such logic through the delegate.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: still applies, if we need to manage the status/conditions during scale down of the workload, it would be better to do it in a single place (updateOperatorStatus)

pkg/operator/apiserver/controller/workload/workload.go
liouk: Given that the delete would happen during the delegate's sync, we shouldn't normally reach the point of removing the conditions if preconditions are failing. But you are right, we should safe-guard against this -- I'll add a check before removing conditions.

pkg/operator/apiserver/controller/workload/workload.go
liouk: The advantage of the patch here in my opinion is that we're only adding to the patch the specific conditions that we want to remove, so it's more concise -- we won't have to manage the whole status object to perform the update, so maybe it's less prone to mistakes.

What would you think @bertinatto?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -35,7 +35,7 @@ const (
 // Delegate captures a set of methods that hold a custom logic
 type Delegate interface {
 	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
-	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, []error)
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, bool, []error)
```
atiratree, line 38: can we name or document the variables in the interface?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +159,30 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable))
+	typeAvailable := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
```
atiratree, line 172: nit: can we have these in the same order as the type vars?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +159,30 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+	if preconditionsReady && removeConditions && workload == nil {
```
atiratree, line 177: what do we want to happen when `removeConditions == true && workload != nil`?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +159,30 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+		}
 
-	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
+		err = c.operatorClient.PatchOperatorStatus(ctx, jsonPatch)
+		return kerrors.NewAggregate(append(errs, err))
+	}
 
 	status := applyoperatorv1.OperatorStatus()
 	if workload != nil {
```
atiratree, line 188: it would look nicer if the precondition handling was first, but it would cause a much bigger diff, so I think we can keep the code blocks as is

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -159,22 +159,30 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+	if preconditionsReady && removeConditions && workload == nil {
+		jsonPatch := v1helpers.RemoveConditionsJSONPatch(previousStatus, []string{typeAvailable, typeDegraded, typeProgressing, typeWorkloadDegraded})
+		if jsonPatch.IsEmpty() {
+			return kerrors.NewAggregate(errs)
+		}
```
atiratree, line 182: we could do without this newline

liouk:
Contains WIP commit; holding.

/hold

pkg/operator/v1helpers/helpers.go
```diff
@@ -565,3 +566,63 @@ func IsUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
...
+	var removedCount int
+	for i, cond := range operatorStatus.Conditions {
+		for _, conditionTypeToRemove := range conditionTypesToRemove {
+			if cond.Type != conditionTypeToRemove {
+				continue
+			}
+			removeAtIndex := i
+			if !jsonPatch.IsEmpty() {
+				removeAtIndex = removeAtIndex - removedCount
+			}
```
atiratree: Is this necessary? Can't we do just
```suggestion
			removeAtIndex := i - removedCount
```

pkg/operator/v1helpers/helpers.go
```diff
@@ -565,3 +566,63 @@ func IsUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
...
+
+	var removedCount int
+	for i, gen := range operatorStatus.Generations {
+		if gen.Name != name || gen.Namespace != namespace || gen.Group != "apps" || gen.Resource != "deployments" {
+			continue
+		}
+		removeAtIndex := i
+		if !jsonPatch.IsEmpty() {
+			removeAtIndex = removeAtIndex - removedCount
+		}
```
atiratree: same question here

pkg/operator/v1helpers/helpers.go
```diff
@@ -565,3 +566,63 @@ func IsUpdatingTooLong(operatorStatus *operatorv1.OperatorStatus, progressingCon
 	progressing := FindOperatorCondition(operatorStatus.Conditions, progressingConditionType)
 	return progressing != nil && progressing.Status == operatorv1.ConditionTrue && time.Now().After(progressing.LastTransitionTime.Add(progressingConditionTimeout))
 }
+
+func RemoveConditionsJSONPatch(operatorStatus *operatorv1.OperatorStatus, conditionTypesToRemove []string, jsonPatch *jsonpatch.PatchSet) *jsonpatch.PatchSet {
+	if operatorStatus == nil {
```
atiratree:
```suggestion
	if operatorStatus == nil || len(conditionTypesToRemove) == 0 {
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,13 @@ const (
 // Delegate captures a set of methods that hold a custom logic
 type Delegate interface {
 	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
-	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, []error)
+	//
+	// It returns a reference to the workload, a bool indicating whether the operator config is at the highest
+	// generation, a bool indicating whether the operator status conditions must be removed (e.g. in case the
+	// delegate intentionally deletes the workload), two strings indicating the name & namespace of the workload
+	// that must be cleaned up from the operator status (respective to the conditions to be removed) and a list
+	// errors.
```
atiratree:
```suggestion
	// of errors.
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,13 @@ const (
...
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, bool, string, string, []error)
```
atiratree: it could be also useful to name the return values since we have that many of them

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +164,84 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
+	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeDegraded)
+	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().WithType(typeProgressing)
+	workloadDegradedCondition := applyoperatorv1.OperatorCondition().WithType(typeWorkloadDegraded)
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+	status := applyoperatorv1.OperatorStatus()
```
atiratree: status is a bit confusing

```suggestion
	statusApplyConfig := applyoperatorv1.OperatorStatus()
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +164,84 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	mustApplyStatus := false
+	statusCleanupPatch := jsonpatch.New()
 
-	deploymentProgressingCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing))
+	// remove operator status conditions if requested by the delegate; otherwise update them
+	// if necessary
+	if removeConditions {
+		statusCleanupPatch = v1helpers.RemoveConditionsJSONPatch(previousStatus, []string{typeAvailable, typeDegraded, typeProgressing, typeWorkloadDegraded}, statusCleanupPatch)
```
atiratree, line 214: superfluous line

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +164,84 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	defer func() {
-		status = status.WithConditions(
-			deploymentAvailableCondition,
-			deploymentDegradedCondition,
-			deploymentProgressingCondition,
-			workloadDegradedCondition,
-		)
+	// remove operator status generations and version if requested by the delegate; otherwise update them
+	// if necessary
+	if len(removedWorkloadName) > 0 && len(removedWorkloadNamespace) > 0 {
```
atiratree: Do we have a use case where conditions should be removed but generations kept? Or vice-versa?

if not, can we use the same variable for toggling this and removeConditions? To unify the approach.

pkg/operator/status/status_controller.go
```diff
@@ -259,6 +268,17 @@ func (c *StatusSyncer) syncStatusVersions(clusterOperatorObj *configv1.ClusterOp
 		}
 	}
 
+	if c.removeEmptyVersions {
```
atiratree, line 272: Let's document the reason behind this part here.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: Maybe we should document that it is preferable to activate WithEmptyVersionRemoval when removeConditions/Generations is used.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -672,6 +683,46 @@ func TestUpdateOperatorStatus(t *testing.T) {
 				return areCondidtionsEqual(expectedConditions, actualStatus.Conditions)
 			},
 		},
+		{
+			name:                           "the delegate controller deletes its operand workload and we remove the conditions",
```
atiratree: looks good, but maybe we could also test the generations while we are at it

pkg/operator/status/status_controller_test.go
```diff
@@ -3,13 +3,14 @@ package status
...
 	"context"
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"reflect"
 	"regexp"
 	"strings"
 	"testing"
 	"time"
 
+	clocktesting "k8s.io/utils/clock/testing"
```
atiratree, line 12: nit: k8s.io group is also below

pkg/operator/v1helpers/helper_test.go
```diff
@@ -281,9 +282,195 @@ func TestRemoveCondition(t *testing.T) {
...
+				Group:     "apps",
+				Resource:  "deployments",
+				Name:      "test",
+				Namespace: "test",
+			},
+			{
+				Group:     "apps2",
+				Resource:  "deployments2",
+				Name:      "test2",
+				Namespace: "test2",
```
atiratree: Can we distinguish the name x namespace to test they do not get mixed up in the invocation?

pkg/operator/v1helpers/helpers.go
liouk: Good catch -- that code pre-existed and I just refactored it into a func.

pkg/operator/apiserver/controller/workload/workload.go
liouk: That's true, I can't think of such a use case. I'll refactor this func to group what's happening in a more unified approach.

pkg/operator/apiserver/controller/workload/workload.go
liouk: Great idea, that's not very obvious.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,24 @@ const (
...
+	// that must be cleaned up from the operator status (respective to the conditions to be deleted) and a list
+	// of errors.
+	//
+	// When a workload gets deleted, not only its respective status conditions will removed, but also its
+	// generations, and its version string will be set to "". To make sure that the StatusSyncer removes
+	// any versions that are equal to the empty string completely, use StatusSyncer.WithEmptyVersionRemoval().
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (
+		delegateWorkload *appsv1.Deployment,
+		operatorConfigAtHighestGeneration bool,
+		removeConditions bool,
```
atiratree: maybe we can then generalize that since the delegate (and this controller) doesn't have to mention the lower level operations done by this controller

```suggestion
		removeWorkload bool,
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +175,74 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable))
+	typeAvailable := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	if len(deletedWorkloadName) > 0 && len(deletedWorkloadNamespace) > 0 && removeConditions {
```
atiratree:
```suggestion
	if removeWorkload && len(deletedWorkloadName) > 0 && len(deletedWorkloadNamespace) > 0  {
```

reads better IMO :)

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,24 @@ const (
...
 	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
-	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, []error)
+	//
+	// It returns a reference to the workload, a bool indicating whether the operator config is at the highest
+	// generation, a bool indicating whether the operator status conditions must be removed (e.g. in case the
+	// delegate intentionally deletes the workload), two strings indicating the name & namespace of the workload
+	// that must be cleaned up from the operator status (respective to the conditions to be deleted) and a list
+	// of errors.
+	//
+	// When a workload gets deleted, not only its respective status conditions will removed, but also its
```
atiratree:
```suggestion
	// When a workload gets deleted, not only its respective status conditions will be removed, but also its
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +175,74 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	if workload != nil {
-		// The Hash field is not required since the LastGeneration field is enough to uniquely identify a Deployment's desired state
-		status = status.WithGenerations(applyoperatorv1.GenerationStatus().
-			WithGroup("apps").
-			WithResource("deployments").
-			WithNamespace(workload.Namespace).
-			WithName(workload.Name).
-			WithLastGeneration(workload.Generation),
-		)
+	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().WithType(typeAvailable)
```
atiratree, line 218: +1, easier to read in two separate block

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,24 @@ const (
 // Delegate captures a set of methods that hold a custom logic
 type Delegate interface {
 	// Sync a method that will be used for delegation. It should bring the desired workload into operation.
-	Sync(ctx context.Context, controllerContext factory.SyncContext) (*appsv1.Deployment, bool, []error)
+	//
+	// It returns a reference to the workload, a bool indicating whether the operator config is at the highest
+	// generation, a bool indicating whether the operator status conditions must be removed (e.g. in case the
```
atiratree:
```suggestion
	// generation, a bool indicating whether the workload reference (e.g. conditions, generations) should be removed from operator status (e.g. in case the
```

atiratree: and maybe update the rest a bit to match that?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: Ok, then maybe we can update the variable to be more generic: https://github.com/openshift/library-go/pull/1902#discussion_r2157297496

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,25 @@ const (
...
+	// When a workload gets deleted (as indicated by the removeWorkload boolean), not only its respective
+	// status conditions will removed, but also its generations, and its version string will be
+	// set to "". To make sure that the StatusSyncer removes any versions that are equal to the empty
+	// string completely, use StatusSyncer.WithEmptyVersionRemoval().
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (
+		delegateWorkload *appsv1.Deployment,
+		operatorConfigAtHighestGeneration bool,
+		removeWorkload bool,
+		deletedWorkloadName string,
+		deletedWorkloadNamespace string,
```
atiratree:
```suggestion
		removedWorkloadName string,
		removedWorkloadNamespace string,
```

for consistency with `removeWorkload`

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +176,74 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
 	}
 }
 
-// updateOperatorStatus updates the status based on the actual workload and errors that might have occurred during synchronization.
-func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *operatorv1.OperatorStatus, workload *appsv1.Deployment, operatorConfigAtHighestGeneration bool, preconditionsReady bool, errs []error) (err error) {
+func (c *Controller) updateOperatorStatus(ctx context.Context,
+	previousStatus *operatorv1.OperatorStatus,
+	workload *appsv1.Deployment,
+	operatorConfigAtHighestGeneration, preconditionsReady, removeWorkload bool,
+	deletedWorkloadName, deletedWorkloadNamespace string,
```
atiratree: \+ elsewhere

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +176,74 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
-	deploymentAvailableCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable))
+	typeAvailable := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	if removeWorkload && len(deletedWorkloadName) > 0 && len(deletedWorkloadNamespace) > 0 {
```
atiratree: Hm maybe it would be safer to output an error if removeWorkload is true, but name/ns are not present?

pkg/operator/apiserver/controller/workload/workload.go
liouk: This naming convention was intentional -- I chose `deleted` for workload name & namespace as it is the right verb for the resource action, while I chose `removeWorkload` for the operator status references because we'll be removing elements from lists/fields instead of deleting any resources.

If you still think this is more confusing than helpful, I'm happy to change all to use "remove".

atiratree: > I chose deleted for workload name & namespace as it is the right verb for the resource action,

I feel like this is more of an implementation detail for the delegate to work out. The workload controller only takes care if the workload is/was present. And if it never was, the delete might have never been called.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +176,78 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	typeAvailable := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeAvailable)
+	typeDegraded := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
+	typeProgressing := fmt.Sprintf("%sDeployment%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeProgressing)
+	typeWorkloadDegraded := fmt.Sprintf("%sWorkload%s", c.conditionsPrefix, operatorv1.OperatorStatusTypeDegraded)
 
-	workloadDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sWorkloadDegraded", c.conditionsPrefix))
+	if removeWorkload {
+		if len(removedWorkloadName) == 0 || len(removedWorkloadNamespace) == 0 {
+			return kerrors.NewAggregate(append(errs, fmt.Errorf("workload marked as removed but no name and/or namespace provided (name=%s/ns=%s)", removedWorkloadName, removedWorkloadNamespace)))
```
atiratree:
```suggestion
			return kerrors.NewAggregate(append(errs, fmt.Errorf("workload marked as removed but no name and/or namespace provided (name=%s, ns=%s)", removedWorkloadName, removedWorkloadNamespace)))
```
This might be confusing when comparing to conventional `ns/name` keys.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -33,7 +35,25 @@ const (
...
+	// generation, a bool indicating whether the workload references must be removed from the operator status
+	// (e.g. in case the workload has been removed), two strings indicating the name & namespace
+	// of the workload that must be cleaned up from the operator status (respective to the conditions to be
+	// removed) and a list of errors.
+	//
+	// When a workload is removed (as indicated by the removeWorkload boolean), not only its respective
+	// status conditions will removed, but also its generations, and its version string will be
+	// set to "". To make sure that the StatusSyncer removes any versions that are equal to the empty
+	// string completely, use StatusSyncer.WithEmptyVersionRemoval().
+	Sync(ctx context.Context, controllerContext factory.SyncContext) (
```
p0lyn0mial, line 49: Assuming the configuration is dynamic, I would prefer to extend the interface with a new method that the controller could use to break the existing flow. I think this would make the code easier to read and maintain. WDYT ?


We could add:
```
IsExternallyManaged(ctx context.Context) (bool, error)
```

or 
```
SkipSyncOnExternal
```

Then we could change the sync method to:
```
func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncContext) error {
...
  if run, err := c.shouldSync(ctx, operatorSpec, controllerContext.Recorder()); !run {
      return err
  }
  
  if skip, err := c.delegate.IsExternallyManaged(ctx); err != nil {
      return c.updateOperatorStatus(...)
  } else if skip {
      // TODO: impl cleanup
  }
...
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +176,78 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+	if removeWorkload {
+		if len(removedWorkloadName) == 0 || len(removedWorkloadNamespace) == 0 {
+			return kerrors.NewAggregate(append(errs, fmt.Errorf("workload marked as removed but no name and/or namespace provided (name=%s, ns=%s)", removedWorkloadName, removedWorkloadNamespace)))
+		}
 
-	deploymentDegradedCondition := applyoperatorv1.OperatorCondition().
-		WithType(fmt.Sprintf("%sDeploymentDegraded", c.conditionsPrefix))
+		// the workload has been removed; remove conditions, generations and version
+		patch := jsonpatch.Merge(
+			v1helpers.RemoveConditionsJSONPatch(previousStatus, []string{typeAvailable, typeDegraded, typeProgressing, typeWorkloadDegraded}),
```
p0lyn0mial, line 203: If this controller is the sole owner of these conditions, then I think simply not providing them will remove them.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -156,48 +176,78 @@ func (c *Controller) shouldSync(ctx context.Context, operatorSpec *operatorv1.Op
...
+			v1helpers.RemoveWorkloadGenerationsJSONPatch(previousStatus, removedWorkloadName, removedWorkloadNamespace),
```
p0lyn0mial, line 204: mhm, I would have to test it, but I would like to know what happens when this controller also doesn't provide the generation.

liouk:
> Is the external auth provider configuration something that needs to be set during installation, or can it be configured afterwards - or even both?

For the moment, in standalone OCP, this is a configuration that needs to be done after installation.

pkg/operator/apiserver/controller/workload/workload.go
liouk: I opted for the current approach because it allows the delegate controller (which is doing the actual workload sync work) to be responsible of managing deletion of its operand during its own sync, so that it can check any conditions necessary etc. This is also in line with the fact that the delegate controller is also responsible for creating the workload in the first place.

Moreover, the intention is to not simply skip sync, but instead have the workload controller clean up any conditions, versions and generations from the operator status.

pkg/operator/apiserver/controller/workload/workload.go
liouk: If I remember correctly there's no other component that explicitly removes conditions, and since we're using SSA, ignoring them would leave them as is.

I'll test this again and report back :+1:

pkg/operator/apiserver/controller/workload/workload.go
liouk: Given that the respective deployment is already deleted (and hence we're removing the generation), do you see any other effect that removing this from the operator status might have?

liouk:
As discussed with @p0lyn0mial, closing this PR in favor of #1977.

PR #1903, opened 2024-12-05
CNTRLPLANE-7: update to k8s 1.32

dusk125:
- [x] remove mod replaces


bertinatto (commented):
Could you keep the changes in different logical commits? Example:

* Updates Kubernetes dependencies to 1.32.1
* go mod vendor
* Fix code because...

PR #1904, opened 2024-12-09
HOSTEDCP-2207: Add filewatcher to library-go

bryan-cox:
Filewatcher was initially created during development of HyperShift's managed Azure service. Due to the Azure Cloud API authentication type used by this service, client certificate, whenever the certificate rotates, the pod needs to reauthenticate with Azure since Azure SDK for Go currently does not support re-authenticating with Azure with the new certificate.
 
This functionality was individually added to several OpenShift repos - CNCC, CIO, and CIRO - through the PRs mentioned below:
- https://github.com/openshift/cloud-network-config-controller/pull/156
- https://github.com/openshift/cluster-ingress-operator/pull/1151 
- https://github.com/openshift/cluster-image-registry-operator/pull/1155 

These repos should use the same codebase for this functionality rather than having 3 different instances of filewatcher.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/distribution/distribution/v3 v3.0.0-20230511163743-f7717b7855ca
 	github.com/evanphx/json-patch v4.12.0+incompatible
+	github.com/fsnotify/fsnotify v1.7.0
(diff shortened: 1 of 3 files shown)


bryan-cox:
cc: @bennerv

bryan-cox:
/close

We won't need this with the new authentication method being introduced in [CNTRLPLANE-103](https://issues.redhat.com/browse/CNTRLPLANE-103)

PR #1905, opened 2024-12-09
feat(pkg/operator/csr): add NewServiceAccountMultiSubjectsApprover approver

machine424:
To avoid the necessity of having multiple CSRApproverController instances like in https://github.com/openshift/cluster-policy-controller/commit/1f053433338879160c5a7f471bf5907f3315f08a

And instead be able to do https://github.com/openshift/cluster-policy-controller/pull/148/files instead.

~not really what I need actually, this would require more changes~

### DIFF
--- a/pkg/operator/csr/csr_approver.go
+++ b/pkg/operator/csr/csr_approver.go
@@ -175,9 +175,9 @@ func getCertApprovalCondition(status *certapiv1.CertificateSigningRequestStatus)
 }
 
 type ServiceAccountApprover struct {
-	saGroups        sets.Set[string] // saGroups is the set of groups for the SA expected to have created the CSR
-	saName          string
-	expectedSubject string
+	saGroups         sets.Set[string] // saGroups is the set of groups for the SA expected to have created the CSR
+	saName           string
+	expectedSubjects sets.Set[string]
 }
 
 // ServiceAccountApprover approves CSRs with a given subject issued by the provided service account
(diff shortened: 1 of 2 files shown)


pkg/operator/csr/csr_approver.go
```diff
@@ -175,19 +176,30 @@ func getCertApprovalCondition(status *certapiv1.CertificateSigningRequestStatus)
 }
 
 type ServiceAccountApprover struct {
-	saGroups        sets.Set[string] // saGroups is the set of groups for the SA expected to have created the CSR
-	saName          string
-	expectedSubject string
+	saGroups         sets.Set[string] // saGroups is the set of groups for the SA expected to have created the CSR
+	saName           string
+	expectedSubjects []string
```
everettraven, line 181: Instead of a string slice, should this be a `sets.Set[string]` to:
- Prevent duplicates (should duplicates be allowed?)
- Have faster lookup than `slices.Contains`
- Not require a new import (even though it is a standard library import)

machine424: For the intended use case, a slice would be sufficient. but sure let's got with a set to keep consistency.

everettraven (commented):
Changes seem good to me. I'm not an approver/reviewer in this repo so I'll leave approval/lgtm to those with the appropriate permissions.

machine424:
> Changes seem good to me. I'm not an approver/reviewer in this repo so I'll leave approval/lgtm to those with the appropriate permissions.

Thanks, but know that /lgtm are always welcome ;)

PR #1906, opened 2024-12-10
API-1835: manifestclient: format jsonpatch payloads

p0lyn0mial:


pkg/manifestclienttest/testdata/mutation-tests/PATCH-crd-in-dataset/Patch/cluster-scoped-resources/config.openshift.io/featuregates/943a-body-instance-name.yaml
```diff
@@ -0,0 +1,5 @@
+- op: test
```
deads2k, line 1: huh, why didn't we have a body before?

p0lyn0mial: we had, the diff also contains removed files (https://github.com/openshift/library-go/pull/1906/files#diff-0352883726ee8c93d2a3774e5149e400545aafca56b47cd0f03ff4c87fe90ad5)

PR #1910, opened 2024-12-13
NO-JIRA: operator/node-controller, guard-controller: parse the master node selector during construction

p0lyn0mial:
while reviewing https://github.com/openshift/library-go/pull/1882 I found that the master nodes selector is being parsed on each `sync` for some controllers.


p0lyn0mial:
/hold

for a proof pr

PR #1911, opened 2024-12-20
[WIP] OCPBUGS-45924: for PodSucceeded static pod should always be pending

tkashem:
the following conditions:
- a) pod.Status.Phase == PodSucceeded
- b) pod.Status.Condition[type==corev1.PodReady]: status == corev1.ConditionTrue

if kubelet does not update the Pod status (a and b) atomically then there is a gap between when 1) pod.Status.Phase is set to PodSucceeded and 2) the PodReady condition is updated to False

if a and b are true then we may signal that "static pod is ready" on the designated node, which may cause the installer controller to create a new installer pod on a different node
if a is true, this alone can be used to say that "static pod is pending" on the designated node.


tkashem:
proof PR: https://github.com/openshift/cluster-etcd-operator/pull/1385

PR #1900, opened 2024-11-28
OCPBUGS-44374: CSI controller conditions should provide reason and message

RomanBednar:
This change warns about `reason` and `message` fields in conditions being mandatory in the future: https://github.com/openshift/library-go/pull/1804/files#diff-49fde404f27181b69aea013be73a1439f91fcef3e1c11701f49fe01a5cef6268R351-R356

Causing overhaul of warning messages for controllers that don't set those:

```
W1125 12:50:39.932013   42393 dynamic_operator_client.go:352] .status.conditions["AWSEFSDriverControllerServiceControllerAvailable"].reason is missing; this will eventually be fatal
W1125 12:50:39.932042   42393 dynamic_operator_client.go:355] .status.conditions["AWSEFSDriverControllerServiceControllerAvailable"].message is missing; this will eventually be fatal
W1125 12:50:39.932047   42393 dynamic_operator_client.go:352] .status.conditions["AWSEFSDriverControllerServiceControllerProgressing"].reason is missing; this will eventually be fatal
W1125 12:50:39.932050   42393 dynamic_operator_client.go:355] .status.conditions["AWSEFSDriverControllerServiceControllerProgressing"].message is missing; this will eventually be fatal
```

```
W1125 14:48:15.962871   57793 dynamic_operator_client.go:352] .status.conditions["AWSEFSDriverNodeServiceControllerAvailable"].reason is missing; this will eventually be fatal
W1125 14:48:15.962883   57793 dynamic_operator_client.go:355] .status.conditions["AWSEFSDriverNodeServiceControllerAvailable"].message is missing; this will eventually be fatal
W1125 14:48:15.962887   57793 dynamic_operator_client.go:352] .status.conditions["AWSEFSDriverNodeServiceControllerProgressing"].reason is missing; this will eventually be fatal
W1125 14:48:15.962891   57793 dynamic_operator_client.go:355] .status.conditions["AWSEFSDriverNodeServiceControllerProgressing"].message is missing; this will eventually be fatal
```

With this patch the messages will always be set, although they're not very helpful:
```
  - lastTransitionTime: "2024-11-25T11:49:08Z"
    message: Deployment is available
    reason: AsExpected
    status: "True"
    type: AWSEFSDriverControllerServiceControllerAvailable
  - lastTransitionTime: "2024-11-25T11:50:06Z"
    message: Deployment is not progressing
    reason: AsExpected
    status: "False"
    type: AWSEFSDriverControllerServiceControllerProgressing
[...]

PR #1912, opened 2025-01-02
remove warning messages about missing reason and message in conditions

RomanBednar:
This is a followup for https://github.com/openshift/library-go/pull/1900 which covers only certain controllers. However operators can set conditions themselves which would make this warning show up when reason or message is omitted.

The warning is not relevant at this moment and we want to remove it in 4.18.


RomanBednar:
/assign @deads2k 

For review, was already discussed on Slack. We will followup with library-go bumps in storage operators.

RomanBednar:
ping @deads2k - it would be great to get it in before we start updating libraries in 4.19

PR #1913, opened 2025-01-08
OCPBUGS-45924: Prevent the installer controller from stomping statuses it just applied

dgrisonnet:
With the move to SSA, the state machine that used to prevent two installer pods from being scheduled concurrently was broken. Because we don't always have up-to-date information from the lister anymore, we need to make sure that the static pod operator data the controller is getting is up-to-date with the last changes it performed.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -102,6 +102,12 @@ type InstallerController struct {
 	clock            clock.Clock
 	installerBackOff func(count int) time.Duration
 	fallbackBackOff  func(count int) time.Duration
+
+	// track StaticPodOperatorStatus apply requests to perform live read during
+	// the next controller sync
+	podOperatorStatusApplied bool
+	// resource version of the last StaticPodOperatorStatus applied
+	lastPodOperatorAppliedRV uint64
 }
 
 // InstallerPodMutationFunc is a function that has a chance at changing the installer pod before it is created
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1146,6 +1159,20 @@ func (c InstallerController) Sync(ctx context.Context, syncCtx factory.SyncConte
 	} else if updatedNodeReportOnSuccessfulUpdateFn != nil {
 		updatedNodeReportOnSuccessfulUpdateFn()
 	}
+
+	// Perform a live get to obtain the RV of the object we just applied as for
```
benluddy, line 1163: If we get a non-nil error from ApplyStaticPodOperatorStatus, isn't it possible that the apply succeeded but our client observed an error (like an I/O error) instead of reading a successful HTTP response?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1106,12 +1109,22 @@ func (c InstallerController) ensureRequiredResourcesExist(ctx context.Context, r
...
 	}
 	operatorStatus := originalOperatorStatus.DeepCopy()
 
+	operatorRV, err := strconv.ParseUint(operatorResourceVersion, 10, 64)
+	if err != nil {
+		return err
+	}
+
+	if operatorRV < c.lastAppliedRV {
+		klog.V(4).Info("Skipping installer controller sync, StaticPodOperator lister is out of date")
```
benluddy, line 1124: Nit: Out of date might be OK, but to be precise this should say that it may not have observed the effect of its most recent write.

pkg/operator/staticpod/controller/installer/installer_controller.go
dgrisonnet: It is not possible since the Apply request is done last: 
https://github.com/openshift/library-go/blob/master/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L273-L280
https://github.com/openshift/library-go/blob/master/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L371-L379

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1106,12 +1110,39 @@ func (c InstallerController) ensureRequiredResourcesExist(ctx context.Context, r
...
 	if err != nil {
 		return err
 	}
 	operatorStatus := originalOperatorStatus.DeepCopy()
 
+	// Perform a live get to obtain the RV of the object the controller applied
+	// because for the MOM effort it is not allowed to get the RV from
+	// Apply response.
+	if c.podOperatorStatusApplied {
+		_, _, resourceVersion, err := c.operatorClient.GetOperatorStateWithQuorum(ctx)
```
benluddy, line 1123: Would it be more robust to save the RV from the first live get?

As written, this controller will only proceed when its cached RV is exactly the same as the RV from a live get. This requires at least 3 network hops: apiserver sends watch event, operator sends live get request, server sends live get response. Imagine another client is continuously spamming writes to the object more often than the time it takes for those 3 hops to complete.

dgrisonnet: oh yes definitely, I just moved code around with the new change, but it would definitely be better that way

benluddy:
https://github.com/openshift/cluster-etcd-operator/pull/1386

PR #1917, opened 2025-01-10
WIP: NO-JIRA: Remove empty message set for operator conditions

ardaguclu:
According to this comment https://github.com/openshift/library-go/blob/5e8bc563202897158370a76957da92155cf3cb2e/pkg/operator/genericoperatorclient/dynamic_operator_client.go#L355, setting empty string to `Message` field should not happen.


ardaguclu:
After discussed with @vrutkovs, closing this
/close

PR #1919, opened 2025-01-13
wire custom clock into clusteroperator status controller

deads2k:


eggfoobar:
/lgtm

Validated locally with auth-operator

PR #1922, opened 2025-01-16
[NO-JIRA]: add readApiServerOrDieV1 for kueue-operator work

kannon92:
I am tasked with building a KueueOperator for installing upstream kueue into Openshift.

I want to follow an established pattern of having the operator install data in binData. This seems to require reading resources and applying those resources. I noticed that APIServer is not yet a function in resourceread.

Proposing this PR to add this as a function so I can apply APIService from upstream Kueue.


kannon92:
/cc @jsafrane

Could I get a review on this?

PR #1923, opened 2025-01-17
Switch to canonical json-patch v4 import

skitt:
This matches k/k upstream, see
https://github.com/kubernetes/kubernetes/pull/123339 for details.


skitt:
@bertinatto what would the process be to get this and #1924 reviewed? Should I open JIRA issues?

bertinatto (commented):
Makes sense to me, this matches 4.19:

https://github.com/openshift/kubernetes/blob/release-4.19/go.mod#L94

Also, I don't see any code changes in the vendor directory, just an import in one of our packages (which is good).

We had problems with this package being inadvertently updated and importing breaking changes, so fixing the import makes sense.

/lgtm

PR #1924, opened 2025-01-17
Switch to openshift/osincli

skitt:
This is identical to RangelReale/osincli; the latter has been archived.

## Summary by CodeRabbit

* **Maintenance**
  * Updated the OAuth client dependency to a maintained compatible fork.
  * No user-facing behavior or functionality changes.


bertinatto (commented):
Since our fork doesn’t include any changes from the original package, and go mod already pins the original revision from ~10 years ago, I’m wondering if there are any tangible benefits to making this change.

Is there something I might be overlooking?

Would migrating the code to https://pkg.go.dev/golang.org/x/oauth2 be a more effective approach instead?

skitt:
> Since our fork doesn’t include any changes from the original package, and go mod already pins the original revision from ~10 years ago, I’m wondering if there are any tangible benefits to making this change.
> 
> Is there something I might be overlooking?

I think it’s beneficial to clarify the dependencies — in future, if a change needs to be made in osincli, it’s more obvious that it’s something we have direct influence over if it’s in the openshift org (which seems to have been the initial intention anyway).

> Would migrating the code to https://pkg.go.dev/golang.org/x/oauth2 be a more effective approach instead?

Probably, yes, but it’s a bit more effort ;-).

ardaguclu:
@skitt I would be happy to merge this PR, if you have any cycles fixing the merge conflicts.

ardaguclu:
/lgtm
@bertinatto for approval

skitt:
Thanks @ardaguclu!

bertinatto (commented):
/approve

bertinatto:
> > Since our fork doesn’t include any changes from the original package, and go mod already pins the original revision from ~10 years ago, I’m wondering if there are any tangible benefits to making this change.
> > Is there something I might be overlooking?
> 
> I think it’s beneficial to clarify the dependencies — in future, if a change needs to be made in osincli, it’s more obvious that it’s something we have direct influence over if it’s in the openshift org (which seems to have been the initial intention anyway).
> 
> > Would migrating the code to https://pkg.go.dev/golang.org/x/oauth2 be a more effective approach instead?
> 
> Probably, yes, but it’s a bit more effort ;-).

Approved the PR, thanks for the patience.

On a side note, now that we have coding agents it might be easier for us to migrate to `golang.org/x/oauth2`, so I created https://github.com/openshift/library-go/issues/2450 for us to track this.

PR #1929, opened 2025-01-29
OCPBUGS-45924: Fix "live get after write" in static pod installer controller.

benluddy:
The static pod installer controller builds an apply configuration for the status of the nodes it manages based on potentially-stale state from an informer cache. The controller is written to assume that it has observed the effect of its own previous writes. If the cache is stale, this assumption can be violated which results in unpredictable installation decisions.

A mechanism was recently introduced requiring the installer controller to wait for its lister to catch up to the latest version after performing a write. This mechanism did not work because it depended on keeping state across calls to the Sync method, and because Sync has a value receiver, field writes were not visible on subsequent calls.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -53,6 +53,14 @@ const (
 //go:embed manifests/installer-pod.yaml
 var podTemplate []byte
 
+type OperatorClient interface {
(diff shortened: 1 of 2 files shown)


benluddy:
/hold

Getting some E2E signal while working on the tests...

dgrisonnet (approved):
/lgtm

benluddy:
/hold cancel

I see this bug occur in practically every 4.19 E2E job, and I don't see it in the presubmits of https://github.com/openshift/cluster-etcd-operator/pull/1392.

PR #1931, opened 2025-01-29
CNTRLPLANE-190: Add support for modyfing asset object in StaticResourceController

atiratree:
- to allow setting e.g. dynamic namespace or owner references

This is needed for our new operators:
- https://github.com/openshift/jobset-operator/pull/1
- https://github.com/openshift/lws-operator/pull/9

To automatically set owner references to the managed objects, and have an option to deploy the operator in any namespace. Please see the usage in: https://github.com/atiratree/jobset-operator/blob/70a1985ed9814e0fd51c6326dbb176ab11628a15/pkg/operator/starter.go#L71-L103

### DIFF
--- a/pkg/operator/csi/csicontrollerset/csi_controller_set.go
+++ b/pkg/operator/csi/csicontrollerset/csi_controller_set.go
@@ -95,11 +95,13 @@ func (c *CSIControllerSet) WithConditionalStaticResourcesController(
 	manifests resourceapply.AssetFunc,
 	files []string,
 	shouldCreateFnArg, shouldDeleteFnArg resourceapply.ConditionalFunction,
+	manifestObjModifier resourceapply.ObjectModifierFunc,
 ) *CSIControllerSet {
 	c.conditionalStaticResourcesControllers = append(c.conditionalStaticResourcesControllers, staticresourcecontroller.NewStaticResourceController(
 		name,
 		manifests,
-		[]string{},
+		nil,
+		nil,
 		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithDynamicClient(dynamicClient),
 		c.operatorClient,
 		c.eventRecorder,
@@ -108,6 +110,7 @@ func (c *CSIControllerSet) WithConditionalStaticResourcesController(
 		files,
 		shouldCreateFnArg,
 		shouldDeleteFnArg,
+		manifestObjModifier,
 	).AddKubeInformers(kubeInformersForNamespace))
 	return c
 }
@@ -120,11 +123,13 @@ func (c *CSIControllerSet) WithStaticResourcesController(
 	kubeInformersForNamespace v1helpers.KubeInformersForNamespaces,
 	manifests resourceapply.AssetFunc,
 	files []string,
+	manifestObjModifier resourceapply.ObjectModifierFunc,
 ) *CSIControllerSet {
 	c.staticResourcesController = staticresourcecontroller.NewStaticResourceController(
 		name,
 		manifests,
 		files,
+		manifestObjModifier,
 		(&resourceapply.ClientHolder{}).WithKubernetes(kubeClient).WithDynamicClient(dynamicClient),
 		c.operatorClient,
 		c.eventRecorder,
@@ -232,6 +237,7 @@ func (c *CSIControllerSet) WithServiceMonitorController(
 	dynamicClient dynamic.Interface,
 	assetFunc resourceapply.AssetFunc,
 	file string,
+	manifestObjModifier resourceapply.ObjectModifierFunc,
 ) *CSIControllerSet {
 	// Use StaticResourceController to apply ServiceMonitors.
 	// Ensure that NotFound errors are ignored, e.g. when ServiceMonitor CRD missing.
@@ -239,6 +245,7 @@ func (c *CSIControllerSet) WithServiceMonitorController(
 		name,
 		assetFunc,
 		[]string{file},
+		manifestObjModifier,
 		(&resourceapply.ClientHolder{}).WithDynamicClient(dynamicClient),
 		c.operatorClient,
 		c.eventRecorder,
--- a/pkg/operator/resource/resourceapply/generic.go
+++ b/pkg/operator/resource/resourceapply/generic.go
@@ -41,6 +41,8 @@ type ApplyResult struct {
 // a resource. This conditional function can also be used to delete the resource when not needed
 type ConditionalFunction func() bool
 
+type ObjectModifierFunc func(runtime.Object) (runtime.Object, error)
+
 type ClientHolder struct {
 	kubeClient          kubernetes.Interface
 	apiExtensionsClient apiextensionsclient.Interface
@@ -83,7 +85,8 @@ func (c *ClientHolder) WithMigrationClient(client migrationclient.Interface) *Cl
 }
 
 // ApplyDirectly applies the given manifest files to API server.
-func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, files ...string) []ApplyResult {
+func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache,
+	manifests AssetFunc, manifestObjModifier ObjectModifierFunc, files ...string) []ApplyResult {
 	ret := []ApplyResult{}
 
 	for _, file := range files {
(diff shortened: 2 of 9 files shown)


pkg/operator/csi/csicontrollerset/csi_controller_set.go
```diff
@@ -95,11 +96,13 @@ func (c *CSIControllerSet) WithConditionalStaticResourcesController(
 	manifests resourceapply.AssetFunc,
 	files []string,
 	shouldCreateFnArg, shouldDeleteFnArg resourceapply.ConditionalFunction,
+	manifestObjModifier func(runtime.Object) (runtime.Object, error),
 ) *CSIControllerSet {
 	c.conditionalStaticResourcesControllers = append(c.conditionalStaticResourcesControllers, staticresourcecontroller.NewStaticResourceController(
 		name,
 		manifests,
-		[]string{},
```
ardaguclu, line 102: Why do we change this to nil?. Accidental change?.

pkg/operator/resource/resourceapply/generic.go
```diff
@@ -83,7 +83,8 @@ func (c *ClientHolder) WithMigrationClient(client migrationclient.Interface) *Cl
 }
 
 // ApplyDirectly applies the given manifest files to API server.
-func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache, manifests AssetFunc, files ...string) []ApplyResult {
+func ApplyDirectly(ctx context.Context, clients *ClientHolder, recorder events.Recorder, cache ResourceCache,
+	manifests AssetFunc, manifestObjModifier func(runtime.Object) (runtime.Object, error), files ...string) []ApplyResult {
```
ardaguclu, line 87: nit: Maybe;

```go
type ModifierFunc func(runtime.Object) (runtime.Object, error)
```
like `AssetFunc`.

pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -88,6 +88,10 @@ type conditionalManifests struct {
 
 	manifests resourceapply.AssetFunc
 
+	// manifestObjModifier is called on manifest resources. Can be used, e.g. for setting a namespace or an owner reference.
+	// The function calls should not result in a different name/namespace on successive calls for the same object.
```
ardaguclu, line 92: I'm not sure I fully understand this. We are going to call this to modify namespace, so function calls will result in a different name/namespace?

ardaguclu (commented):
This will increase the adoption of statisresourcecontroller in my opinion.

pkg/operator/csi/csicontrollerset/csi_controller_set.go
atiratree: since I was changing the line below, I also changed this line to improve readability

pkg/operator/resource/resourceapply/generic.go
atiratree: Good idea, I will extract the type.

pkg/operator/staticresourcecontroller/static_resource_controller.go
atiratree: I mean, you can change the stored asset to your namespace, but if you change it to other namespace during the runtime of the operator, it will not achieve the resource migration from one ns to another.

pkg/operator/resource/resourceapply/generic.go
atiratree: done

pkg/operator/staticresourcecontroller/static_resource_controller.go
atiratree: added to the comments

p0lyn0mial (commented):
I've been looking at the `StaticResourceController` so far, and I'm wondering what an alternative design might look like. Is there a way that wouldn't require modifying the static controller ? (and the others possibly)

I think the operator namespace is known at runtime. Therefore, in theory, before starting the static controller in your operator, we could "compile" the static resources and inject the namespace into them. These resources could then be passed to the static controller.

For adding owner references, we could introduce a separate controller. Of course, that would mean that the static resources are initially created without owner references. Would that be acceptable? Also, what is the purpose of adding the owner references for your use case ?

Could this approach also be applied to the other controllers you had to modify? Would it simplify the design?

atiratree:
> I've been looking at the StaticResourceController so far, and I'm wondering what an alternative design might look like. Is there a way that wouldn't require modifying the static controller ? (and the others possibly)

Alternative designs would be:
- allow the manifestObjModifier only in the `WithConditionalResources` and not have it in the constructor so it would be  more optional
- Allow only namespace and owner references options in `WithConditionalResources`

> I think the operator namespace is known at runtime. Therefore, in theory, before starting the static controller in your operator, we could "compile" the static resources and inject the namespace into them. These resources could then be passed to the static controller.

I also looked into modifying the `AssetFunc` to do that, but that is unfortunately quite cumbersome to use.

> For adding owner references, we could introduce a separate controller. Of course, that would mean that the static resources are initially created without owner references. Would that be acceptable? Also, what is the purpose of adding the owner references for your use case ?

It is hard to assume what namespaces will be used by the user of the `StaticResourceController`. There can be at least operator and operand namespace, but I can imagine more. As mentioned it is possible to resolve that with `WithConditionalResources`.

We would like to have all the operand resources be owned by the operator CR, to ensure all resources are cleaned up properly upon operator deletion.

Separate controller would be hard to synchronize, since StaticResourceController would try to overwrite the changes.

> Could this approach also be applied to the other controllers you had to modify? Would it simplify the design?

It can be implemented in the operator, but having this in StaticResourceController would help us simplify the codebase and not repeat the reconciliation code for most of the resources.

p0lyn0mial:
> I also looked into modifying the AssetFunc to do that, but that is unfortunately quite cumbersome to use.

I think it should be simple, I think something like https://go.dev/play/p/ShkNKPdbpWW# should work, no ?


The difficult part is getting the owner reference data. I think that in this case, we have two options: issue a live GET request to the server or wait until the informer has synced and read it from the cache.


By modifying the static data before passing it to the controller, we perform the modification only once. We don’t change the controller, and it can continue handling static resources. By adding a modification function that is called in different places, the controller acts more on dynamic data.

atiratree:
> I think it should be simple, I think something like https://go.dev/play/p/ShkNKPdbpWW# should work, no ?

We try to reuse the upstream manifests and generate them with each new release. So, we would like to preserve them as they are without additional changes (e.g. owner references).

> The difficult part is getting the owner reference data. I think that in this case, we have two options: issue a live GET request to the server or wait until the informer has synced and read it from the cache.

We do use a lister for that.

What is your opinion on changing just the `WithConditionalResources` function to keep the API more optional?

p0lyn0mial:
> What is your opinion on changing just the WithConditionalResources function to keep the API more optional?

Updating WithConditionalResources would also make the controller more dynamic.

> We try to reuse the upstream manifests (operand) and generate them with each new release. So, we would like to preserve them as they are without additional changes (e.g. owner references).

I think you could read the resources, cast them to unstructured objects, and then simply modify the metadata. I think you could write generic code that would work with any resource. To share the code between operators, we could add helpers directly to the library. What do you think?

I think it boils down to:

```
// TODO: read yaml/json and unmarshal to unstructured
u := obj.(*unstructured.Unstructured)
u.SetNamespace()
u.SetOwnerReferences()
// TODO: marshal back to yaml/json
```

For setting ns and owner ref we could have helpers.

atiratree:
> Updating WithConditionalResources would also make the controller more dynamic.

Having an `AssetFunction` do that instead will make it dynamic as well. Since I am using the `manifestObjModifier` in the same places the `AssetFunction` is called.

> I think you could read the resources, cast them to unstructured objects, and then simply modify the metadata. I think you could write generic code that would work with any resource. To share the code between operators, we could add helpers directly to the library. What do you think?

It seems unnecessary to do the unmarshalling twice. This also brings a bit of an inconvenience IMO. Having to do marshalling/unmarshalling for the consumer.

p0lyn0mial:
> Having an AssetFunction do that instead will make it dynamic as well

Exactly. The controller already allows for mutations, so why add yet another mechanism? Notice that we could implement AssetFunction as static, we could modify the resources only once at startup.

> It seems unnecessary to do the unmarshalling twice

I don’t know if unmarshalling is necessary. It depends on what can be passed to the apply function used by the controller. But since we would do it only once, it doesn’t seem like a big deal.

atiratree:
I have changed my code to use the AssetFunction: https://github.com/openshift/jobset-operator/pull/1/commits/b3c5890e70817ec89fb40760ad0a6edc6d6bf236

That is a lot of extra code to change for example just a namespace. I do not see other people using it in this way. Unless there is a simple API in the StaticResourceController, it is just easier to add it directly to your reconciliation loop.

And the double marshaling/unmarshaling is not a great data transformation either.

PR #1935, opened 2025-02-17
SPLAT-2052: [vsphere] enable multi-nic feature gate by default

rvanderp3:


rvanderp3:
/close

not necessary

PR #1936, opened 2025-02-19
CombineCABundleConfigMaps: use optimistic create/update

vrutkovs:
Instead of re-creating configmap from scratch every time this function should attempt to use existing configmap and replace the contents only. This would prevent extra configmap updates when metadata changes.

Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1812


pkg/operator/resourcesynccontroller/core.go
```diff
@@ -55,18 +62,12 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
...
-		),
-		Data: map[string]string{
-			"ca-bundle.crt": string(caBytes),
-		},
+	modified := additionalAnnotations.EnsureTLSMetadataUpdate(&cm.ObjectMeta)
+	cm.Data = map[string]string{
+		"ca-bundle.crt": string(caBytes),
 	}
-	return cm, nil
+	return cm, modified, nil
```
benluddy, line 72: How is the bool return value supposed to be interpreted? It's unclear to me what it means when EnsureTLSMetadataUpdate returns false (and so the returned bool value is false) but Data has been mutated.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	"fmt"
-	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	"net/http"
 	"sort"
 	"strings"
 	"sync"
 	"time"
 
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+
```
benluddy, line 14: Could this make backporting more painful?

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -14,15 +14,22 @@ import (
 	"github.com/openshift/library-go/pkg/operator/certrotation"
 )
 
-func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister corev1listers.ConfigMapLister, additionalAnnotations certrotation.AdditionalAnnotations, inputConfigMaps ...ResourceLocation) (*corev1.ConfigMap, error) {
+func CombineCABundleConfigMaps(destinationConfigMap *corev1.ConfigMap, lister corev1listers.ConfigMapLister, additionalAnnotations certrotation.AdditionalAnnotations, inputConfigMaps ...ResourceLocation) (*corev1.ConfigMap, bool, error) {
```
benluddy, line 17: This must not have any unit test coverage... could you add some to cover the behavior change being made please?

pkg/operator/resourcesynccontroller/core.go
vrutkovs: `modified` here returns if [update is necessary](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1812/files#diff-0af15f182aea9898e784fc589175ef4c2411c9391bb5a1f41327ba7c3e2172f5R367):
```
requiredConfigMap, updateRequired, err := resourcesynccontroller.CombineCABundleConfigMaps(
```

It means either metadata or contents has changed. I think I missed "content has changed, but not the metadata", so I'll add some unittests to verify that

pkg/operator/resourcesynccontroller/resourcesync_controller.go
vrutkovs: This may complicate it, yes, but I'd prefer to avoid packports to 4.18 and earlier as we don't have stable rotation tests for them anyway

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -14,15 +14,21 @@ import (
...
```
benluddy, line 17: Is there a reasonable way to introduce this as a function with a different name, and deprecate the existing one, to avoid lockstep updates to hypershift, kas-o, kcm-o, and scheduler-o?

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -55,18 +61,17 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
...
+	modified := additionalAnnotations.EnsureTLSMetadataUpdate(&cm.ObjectMeta)
+	newCMData := map[string]string{
+		"ca-bundle.crt": string(caBytes),
+	}
+	if !reflect.DeepEqual(cm.Data, newCMData) {
+		modified = true
 	}
-	return cm, nil
+
+	cm.Data = newCMData
```
benluddy, line 75: Nit: I expected to see the assignment to cm.Data inside the "then" block with the assignment to modified. It looks correct to me either way.

pkg/operator/resourcesynccontroller/core_test.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+}
+
+type mockConfigMapNamespaceLister struct {
+	namespace  string
+	configMaps map[string]map[string]*corev1.ConfigMap
+}
+
+func (m *mockConfigMapNamespaceLister) List(selector labels.Selector) ([]*corev1.ConfigMap, error) {
+	return nil, nil
```
benluddy, line 43: Nit: Panic in unimplemented stubs to prevent someone later using them by mistake.

pkg/operator/resourcesynccontroller/core_test.go
```diff
@@ -0,0 +1,285 @@
...
+			if !test.expectError && err != nil {
+				t.Errorf("Unexpected error: %v", err)
+			}
+			if test.expectModified != modified {
+				t.Errorf("Expected modified=%v but got %v", test.expectModified, modified)
+			}
+			if err == nil && result == nil {
+				t.Errorf("Expected result to not be nil when no error occurred")
+			}
+
```
benluddy, line 276: Any reason not to compare the full returned CM against an expected CM?

pkg/operator/resourcesynccontroller/core.go
vrutkovs: Yeah, I think it would be a good idea

pkg/operator/resourcesynccontroller/core_test.go
vrutkovs: Right, it would verify removal of expired certs too

everettraven (approved):
LGTM - had one minor, non-blocking, nit.

/lgtm

pkg/operator/resourcesynccontroller/core.go
```diff
@@ -70,3 +70,64 @@ func CombineCABundleConfigMaps(destinationConfigMap ResourceLocation, lister cor
...
+		certificates = append(certificates, inputCerts...)
+	}
+
+	certificates = crypto.FilterExpiredCerts(certificates...)
+	finalCertificates := []*x509.Certificate{}
+	// now check for duplicates. n^2, but super simple
+	for i := range certificates {
+		found := false
+		for j := range finalCertificates {
+			if reflect.DeepEqual(certificates[i].Raw, finalCertificates[j].Raw) {
```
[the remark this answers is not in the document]
vrutkovs: It also does `[]byte` -> `string` conversion, so it will use even more memory (and use more allocations). This is not a performance critical code, so I'd rather optimize for memory usage and leave it as is. We may want to do a more comprehensive review of certificate comparison later however

benluddy:
> @benluddy I saw your comments, is there anything else that needs to be addressed?

I don't think so. Looks good assuming it has been demonstrated to work in an operator.

PR #1939, opened 2025-02-28
OCPBUGS-35102: certrotation: ignore optimistic lock errors

vrutkovs:
Signer/CABundle/Target certcontrollers use informers to fetch updates and may work on outdated versions of the secret/configmap. If update has returned "please apply your changes to the latest version and try again" error the controller should silently skip it, otherwise it will produce multiple duplicate events and operator status updates.

### DIFF
--- a/pkg/operator/certrotation/cabundle.go
+++ b/pkg/operator/certrotation/cabundle.go
@@ -98,6 +98,10 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		caBundleConfigMap = actualCABundleConfigMap
 	} else if updateRequired {
 		actualCABundleConfigMap, err := c.Client.ConfigMaps(c.Namespace).Update(ctx, caBundleConfigMap, metav1.UpdateOptions{})
+		if apierrors.IsConflict(err) {
+			// ignore error if its attempting to update outdated version of the configmap
+			return nil, nil
+		}
 		resourcehelper.ReportUpdateEvent(c.EventRecorder, actualCABundleConfigMap, err)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


pkg/operator/certrotation/cabundle.go
```diff
@@ -98,6 +100,10 @@ func (c CABundleConfigMap) EnsureConfigMapCABundle(ctx context.Context, signingC
 		caBundleConfigMap = actualCABundleConfigMap
 	} else if updateRequired {
 		actualCABundleConfigMap, err := c.Client.ConfigMaps(c.Namespace).Update(ctx, caBundleConfigMap, metav1.UpdateOptions{})
+		if err != nil && strings.Contains(err.Error(), genericregistry.OptimisticLockErrorMsg) {
```
benluddy, line 103: I'm wary of doing this based on an error message substring because of the common practice of blindly concatenating error messages. Are you worried about existing or potential future cases where https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsConflict returns true but the error is not a resourceVersion mismatch?

benluddy (commented):
Is this worth doing if https://github.com/openshift/library-go/pull/1938 prevents pathological spam?

pkg/operator/certrotation/cabundle.go
vrutkovs: Since we only use SSA for these secrets I think any conflict would be resourceVersion mismatch iiuc

vrutkovs: Testing `apierrors.IsConflict` in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1819

pkg/operator/certrotation/client_cert_rotation_controller.go
```diff
@@ -130,7 +130,7 @@ func (c CertRotationController) getSigningCertKeyPairLocation() string {
 
 func (c CertRotationController) SyncWorker(ctx context.Context) error {
 	signingCertKeyPair, _, err := c.RotatedSigningCASecret.EnsureSigningCertKeyPair(ctx)
-	if err != nil {
+	if err != nil || signingCertKeyPair == nil {
 		return err
 	}
```
bertinatto, line 136: It looks like you could check for the conflict before the other class of errors, like:

```go
if apierrors.IsConflict(err) {
    return nil
}
if err != nil {
    return err
}
```

We use this pattern for checking NotFound errors.

PR #1941, opened 2025-03-04
OCPBUGS-52466: Exclude machineconfig and controllerconfig objects from audit logs

liouk:
`MachineConfig` and `ControllerConfig` APIs from the `machineconfiguration.openshift.io` group can contain sensitive data such as pull secrets (in general they contain everything that needs to be written to disk).

This PR excludes these objects from being logged into the audit logs for the `WriteRequestBodies` and `AllRequestBodies` policies.

### DIFF
--- a/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
+++ b/pkg/operator/apiserver/audit/manifests/allrequestbodies-rules.yaml
@@ -8,5 +8,7 @@
     resources: ["tokenreviews", "tokenrequests"]
   - group: "oauth.openshift.io"
     resources: ["oauthclients", "tokenreviews"]
+  - group: "machineconfiguration.openshift.io"
+    resource: ["machineconfig", "controllerconfig"]
 # catch-all rule to log all other requests with request and response payloads
 - level: RequestResponse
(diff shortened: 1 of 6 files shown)


liouk:
At the moment this PR is for demonstration purposes only to facilitate the discussion around the respective bug.

/hold

liouk:
Proof PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1821

yuqi-zhang (approved):
This makes sense from the MCO perspective, thanks for setting this up!

To clarify, does this remove any audit of those objects altogether? Or just ensures that their contents (which contain certificates/secrets in this case) do not get collected in the log

liouk:
> This makes sense from the MCO perspective, thanks for setting this up!
> 
> To clarify, does this remove any audit of those objects altogether? Or just ensures that their contents (which contain certificates/secrets in this case) do not get collected in the log

This change ensures that audit events for the specified resources (`MachineConfig`, `ControllerConfig`) will only get logged at Metadata level (which excludes request/response). So you'll see metadata for those requests (e.g. timestamp, verb, resource), but you won't see request/response bodies. See here for more info: https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#audit-policy

p0lyn0mial:
ok, it looks like at least `machineconfig.spec.config` and `controllerconfig.spec.internalRegistryPullSecret` can contain sensitive data. 

we also need to bump `kas` and the aggregated api servers to pick up this change. 

/approve

PR #1943, opened 2025-03-12
CNTRLPLANE-317: crypto: rework functions to accept time.Duration instead of days

vrutkovs:
This would help us build certificates with shorter lifetime. Instead of accepting days the functions are updated to accept arbitrary time.Duration. This is necessary for ShortCertRotation featuregate implementation (https://github.com/openshift/service-ca-operator/pull/245)

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -393,7 +393,7 @@ func GetTLSCertificateConfig(certFile, keyFile string) (*TLSCertificateConfig, e
 	}
 	certs, err := cert.ParseCertsPEM(certPEMBlock)
 	if err != nil {
-		return nil, fmt.Errorf("Error reading %s: %s", certFile, err)
+		return nil, fmt.Errorf("error reading %s: %s", certFile, err)
 	}
 
 	keyPEMBlock, err := os.ReadFile(keyFile)
@@ -419,7 +419,7 @@ func GetTLSCertificateConfigFromBytes(certBytes, keyBytes []byte) (*TLSCertifica
 
 	certs, err := cert.ParseCertsPEM(certBytes)
 	if err != nil {
-		return nil, fmt.Errorf("Error reading cert: %s", err)
+		return nil, fmt.Errorf("error reading cert: %s", err)
 	}
 
 	keyPairCert, err := tls.X509KeyPair(certBytes, keyBytes)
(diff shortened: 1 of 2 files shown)


pkg/crypto/crypto.go
```diff
@@ -628,23 +628,20 @@ func MakeSelfSignedCA(certFile, keyFile, serialFile, name string, expireDays int
...
+	return MakeSelfSignedCAConfigForSubject(subject, lifetime)
 }
 
-func MakeSelfSignedCAConfigForSubject(subject pkix.Name, expireDays int) (*TLSCertificateConfig, error) {
-	var caLifetimeInDays = DefaultCACertificateLifetimeInDays
-	if expireDays > 0 {
-		caLifetimeInDays = expireDays
+func MakeSelfSignedCAConfigForSubject(subject pkix.Name, lifetime time.Duration) (*TLSCertificateConfig, error) {
+	if lifetime <= 0 {
+		lifetime = DefaultCACertificateLifetimeDuration
```
bertinatto, line 638: nit: I wonder if it'd be a good idea to log this overwrite

pkg/crypto/crypto.go
```diff
@@ -773,13 +770,13 @@ func GetServerCert(certFile, keyFile string, hostnames sets.Set[string]) (*TLSCe
 		return server, nil
 	}
 
-	return nil, fmt.Errorf("Existing server certificate in %s does not match required hostnames.", certFile)
+	return nil, fmt.Errorf("existing server certificate in %s does not match required hostnames", certFile)
```
bertinatto, line 773: nit: I'd keep these in a separate commit

pkg/crypto/crypto.go
```diff
@@ -1145,8 +1136,8 @@ func NewClientCertificateTemplateForDuration(subject pkix.Name, lifetime time.Du
 	}
 }
 
-func warnAboutCertificateLifeTime(name string, defaultLifetimeInDays int) {
-	defaultLifetimeInYears := defaultLifetimeInDays / 365
+func warnAboutCertificateLifeTime(name string, defaultLifetimeDuration time.Duration) {
+	defaultLifetimeInYears := defaultLifetimeDuration / 365 / 24
```
bertinatto, line 1140: should we change this to log a more granular value rather than years?

pkg/crypto/crypto_test.go
```diff
@@ -265,30 +262,32 @@ func TestValidityPeriodOfClientCertificate(t *testing.T) {
 	}
 
 	tests := []struct {
-		passedExpireDays int
-		realExpireDays   int
+		passedDuration time.Duration
```
bertinatto, line 265: nit: maybe passedExpireDuration

pkg/crypto/crypto.go
vrutkovs: So far certificate issued for month is standard practice, but later we'll want to shorten those. Years is okay now

PR #1944, opened 2025-03-13
feat: add support for protobuf encoded requests

eggfoobar:
in k8s 0.32 the upstream resource clients default to encoding requests as protobufs, added logic to handle those in the manifest write roundtripper


eggfoobar:
Closing in favor of https://github.com/openshift/library-go/pull/1947

PR #1947, opened 2025-03-13
OCPEDGE-1693: feat: update client configs for manifestclient

eggfoobar:
added test to validate using the round tripper with the k8s/client-go clientsets

### DIFF
--- a/pkg/manifestclient/readwrite_roundtripper.go
+++ b/pkg/manifestclient/readwrite_roundtripper.go
@@ -5,9 +5,12 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"k8s.io/client-go/rest"
 	"net/http"
 	"os"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
 // RecommendedRESTConfig is meant to be paired with the HTTPClients below
@@ -15,10 +18,19 @@ func RecommendedRESTConfig() *rest.Config {
 	return &rest.Config{
 		QPS:   1000,
 		Burst: 10000,
+		ContentConfig: rest.ContentConfig{
+			ContentType: runtime.ContentTypeJSON,
+		},
 	}
 }
 
+// RecommendedKubernetesWithClient kubernetes client to be used with the HTTPClients below
+func RecommendedKubernetesWithClient(httpClient *http.Client) (*kubernetes.Clientset, error) {
+	return kubernetes.NewForConfigAndClient(RecommendedRESTConfig(), httpClient)
+}
+
 // Enter here and call `NewForConfigAndClient(manifestclient.RecommendedRESTConfig(), httpClient)`
+// For Kubernetes built in clients, use `manifestclient.RecommendedKubernetesWithClient(httpClient)`
 func NewHTTPClient(mustGatherDir string) MutationTrackingClient {
 	mutationTrackingRoundTripper := newReadWriteRoundTripper(os.DirFS(mustGatherDir))
 	return &mutationTrackingClient{
@@ -30,6 +42,7 @@ func NewHTTPClient(mustGatherDir string) MutationTrackingClient {
 }
 
 // Enter here and call `NewForConfigAndClient(manifestclient.RecommendedRESTConfig(), httpClient)`
+// For Kubernetes built in clients, use `manifestclient.RecommendedKubernetesWithClient(httpClient)`
 func NewTestingHTTPClient(embedFS fs.FS) MutationTrackingClient {
 	mutationTrackingRoundTripper := newReadWriteRoundTripper(embedFS)
 	return &mutationTrackingClient{
--- a/pkg/manifestclient/testing/create_test.go
+++ b/pkg/manifestclient/testing/create_test.go
@@ -1,0 +1,136 @@
+package testing
+
+import (
+	"context"
+	"net/http"
+	"path/filepath"
+	"regexp"
+	"testing"
+
+	"github.com/openshift/library-go/pkg/manifestclient"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	applyconfigurationscorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func TestUpdateConfigMapLabel(t *testing.T) {
+	client := setupKubernetesClient(t)
+	configMap, err := client.CoreV1().ConfigMaps("openshift-authentication").Get(context.TODO(), "audit", metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("error updating configmap: %v", err)
+	}
+	updated := configMap.DeepCopy()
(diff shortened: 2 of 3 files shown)


pkg/manifestclient/write_roundtripper.go
```diff
@@ -144,7 +144,7 @@ func (mrt *writeTrackingRoundTripper) roundTrip(req *http.Request) ([]byte, erro
 	if actionHasRuntimeObjectBody {
 		bodyObj, err = runtime.Decode(unstructured.UnstructuredJSONScheme, bodyContent)
 		if err != nil {
-			return nil, fmt.Errorf("unable to decode body: %w", err)
+			return nil, fmt.Errorf("unable to decode body (content-type is: %s, needed: application/json) : %w", req.Header.Get("Content-Type"), err)
```
benluddy, line 147: I would check and fail based on the content-type header before ever trying to decode the body.

pkg/manifestclient/readwrite_roundtripper.go
```diff
@@ -5,16 +5,20 @@ import (
...
+	"k8s.io/client-go/rest"
 )
 
 // RecommendedRESTConfig is meant to be paired with the HTTPClients below
 func RecommendedRESTConfig() *rest.Config {
 	return &rest.Config{
 		QPS:   1000,
 		Burst: 10000,
+		ContentConfig: rest.ContentConfig{
+			ContentType: "application/json",
```
benluddy, line 20: Is there a Godoc or similar that people are likely to see when they go to instantiate and use one of the roundtrippers from this package without using the "recommended" config?

pkg/manifestclient/write_roundtripper.go
eggfoobar: Updated

pkg/manifestclient/readwrite_roundtripper.go
eggfoobar: Yeah, there is, added an extra line to help push folks to using it correctly.

```
// For Kubernetes built in clients, use `manifestclient.RecommendedKubernetesWithClient(httpClient)`
```

pkg/manifestclient/write_roundtripper.go
```diff
@@ -17,10 +18,19 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/server"
-	"net/http"
 	"sigs.k8s.io/yaml"
 )
 
+const applicationJsonContentType = "application/json"
```
benluddy, line 24: Nit: There's a const in apimachinery for this one (https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#pkg-constants).

pkg/manifestclient/write_roundtripper.go
```diff
@@ -142,6 +152,9 @@ func (mrt *writeTrackingRoundTripper) roundTrip(req *http.Request) ([]byte, erro
 	var bodyObj runtime.Object
 	actionHasRuntimeObjectBody := action != ActionPatch && action != ActionPatchStatus
 	if actionHasRuntimeObjectBody {
+		if contentType := req.Header.Get("Content-Type"); !validHeaders.Has(contentType) {
```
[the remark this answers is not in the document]
eggfoobar: Yeah, the reason I put it here is simply because I didn't know if there were going to be any surprise headers that would trip this check but wouldn't need to be decoded, didn't want to needlessly fail if the other checks don't need it. I'll move it out if you think this isn't really a concern

pkg/manifestclient/write_roundtripper.go
eggfoobar: Thank you!

pkg/manifestclient/write_roundtripper.go
```diff
@@ -17,10 +18,19 @@ import (
...
 )
 
+const applicationJsonContentType = "application/json"
+
+var validHeaders = sets.Set[string]{}.Insert(
+	applicationJsonContentType,
+	string(types.ApplyPatchType),
+	string(types.JSONPatchType),
+	string(types.MergePatchType),
+	string(types.StrategicMergePatchType),
```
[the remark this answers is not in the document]
eggfoobar: Added, thanks!

pkg/manifestclient/write_roundtripper.go
```diff
@@ -17,10 +18,19 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/server"
-	"net/http"
 	"sigs.k8s.io/yaml"
 )
 
+const applicationJsonContentType = "application/json"
+
+var validHeaders = sets.Set[string]{}.Insert(
```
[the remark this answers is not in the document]
eggfoobar: Way cleaner, thank you!

pkg/manifestclient/write_roundtripper.go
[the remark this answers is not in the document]
benluddy: If we can't come up with a realistic example request for that case, then I'd rather discover it quickly through an error than learn far down the line that we shipped with support for some unintended behavior.

pkg/manifestclient/testing/create_test.go
```diff
@@ -0,0 +1,146 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	applyconfigurationscorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func TestUpdateConfigMapLabel(t *testing.T) {
+	client := setupKubernetesClient(t)
+	configMapList, err := client.CoreV1().ConfigMaps("openshift-oauth-apiserver").List(context.TODO(), metav1.ListOptions{})
```
bertinatto, line 22: You need to change the namespace to "openshift-authentication", there are no configMaps in this namespace

pkg/manifestclient/testing/create_test.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		t.Fatalf("error updating configmap: %v", err)
+	}
+
+	var errList []error
+	for _, configMap := range configMapList.Items {
+		updated := configMap.DeepCopy()
+		updated.Labels = map[string]string{"mynew": "label"}
+
+		_, err := client.CoreV1().ConfigMaps("").Update(context.TODO(), updated, metav1.UpdateOptions{})
```
bertinatto, line 32: The empty namespace here won't work

pkg/manifestclient/testing/create_test.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			errList = append(errList, fmt.Errorf("failed to update %s/%s :%w", updated.Namespace, updated.Name, err))
+		}
+	}
+	if errList != nil {
+		t.Fatalf("failed to update some configmaps: %v", errors.NewAggregate(errList))
+	}
+}
```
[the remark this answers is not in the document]
eggfoobar: Yup, you're right, updated.

bertinatto (commented):
/lgtm

PR #1955, opened 2025-04-10
Don't fail on CipherSuites lookup for TLSv1.3 ciphers

tiraboschi:
https://github.com/openshift/library-go/pull/1850
enabled TLS 1.3 ciphers.
But CipherSuite lookup functions are
still explicitly failing for TLSv1.3 ciphers.
Let's fix it to avoid the need to filter out
the list on caller side before calling
CipherSuitesOrDie


tiraboschi:
opened twice by mistake

PR #1957, opened 2025-04-22
OCPBUGS-54385: Avoid adding the same informer more than once

mpatlasov:
Adding the same informer several times is useless and has a negative side-effect: a change of object on API-server leads to scheduling reconcile loop several times. Calling reconcile more often than necessary is not great.

/cc @openshift/storage

### DIFF
--- a/pkg/controller/factory/factory.go
+++ b/pkg/controller/factory/factory.go
@@ -3,11 +3,13 @@ package factory
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/robfig/cron"
 	"k8s.io/apimachinery/pkg/runtime"
 	errorutil "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openshift/library-go/pkg/operator/events"
@@ -245,6 +247,11 @@ func (f *Factory) WithControllerInstanceName(controllerInstanceName string) *Fac
 	return f
 }
 
+type informerHandleTuple struct {
+	informer Informer
+	filter   uintptr
+}
+
 // Controller produce a runnable controller.
 func (f *Factory) ToController(name string, eventRecorder events.Recorder) Controller {
 	if f.sync == nil {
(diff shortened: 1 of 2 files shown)


pkg/operator/staticresourcecontroller/static_resource_controller.go
```diff
@@ -261,8 +265,11 @@ func (c *StaticResourceController) AddKubeInformers(kubeInformersByNamespace v1h
 	return ret
 }
 
-func (c *StaticResourceController) AddInformer(informer cache.SharedIndexInformer) *StaticResourceController {
-	c.factory.WithInformers(informer)
+func (c *StaticResourceController) AddInformer(set sets.Set[cache.SharedIndexInformer], informer cache.SharedIndexInformer) *StaticResourceController {
+	if !set.Has(informer) {
```
jsafrane, line 269: I think it would make sense to fix all controllers in the same way, not just `StaticResourceController`. There is a place where controller `Factory` adds event handlers: https://github.com/openshift/library-go/blob/ac3ba9eb16a23890629f719f3fe11428ad395796/pkg/controller/factory/factory.go#L298-L304

IMO it could skip `informer` + `f.informers[i].filter` tuple that was already added.

mpatlasov:
@jsafrane , I moved logic from static_resource_controller.go to factory.go [as you suggested](https://github.com/openshift/library-go/pull/1957#discussion_r2055517429). Can you please have a look again.

pkg/controller/factory/factory.go
```diff
@@ -245,6 +247,11 @@ func (f *Factory) WithControllerInstanceName(controllerInstanceName string) *Fac
 	return f
 }
 
+type informerHandleTuple struct {
+	informer Informer
+	filter   uintptr
+}
```
bertinatto, line 253: Do we really need `filter` in the key? If we have the same informer twice, but with different filter functions, then we'd still add it twice, right?

bertinatto (commented):
If possible, please add a PR that updates the library-go from this PR in an operator where the problem was observed.

pkg/controller/factory/factory.go
mpatlasov: @bertinatto , I think we do need `filter` in the key. Otherwise, if we have the same informer twice (but with different filter functions), then we'd add it only once, and afterwards any notification for the second filter would be missed (because we'd check it only against the first filter).

bertinatto: It seems like you're right, the filter is embedded in the resource handler in `controller_context.go:48`, so if we have the same informer twice (but with different filter functions), it'd be added only once, and we'd potentially miss events that would match the filter that was left out.

However, if I have the same informer twice, and only one of them has a filter, and then we get an event that would match the filtered handler, then both handlers would actually be called. It seems like this fix addresses exact duplicates, but doesn't handle all cases of redundant event processing. Does that make sense? Should we care?

bertinatto: Also, could you create a proof PR (e.g., replace CSO's library-go with the one from your PR)?

mpatlasov: > However, if I have the same informer twice, and only one of them has a filter, and then we get an event that would match the filtered handler, then both handlers would actually be called. It seems like this fix addresses exact duplicates, but doesn't handle all cases of redundant event processing. Does that make sense? Should we care?

Correct, the fix is only for exact duplicates. Redundant event processing is tolerable -- it only adds extra call to reconciliation loop (in my observations).

mpatlasov: > Also, could you create a proof PR (e.g., replace CSO's library-go with the one from your PR)?

https://github.com/openshift/cluster-storage-operator/pull/585

bertinatto: /approve

PR #1958, opened 2025-04-28
OCPBUGS-47773: Add API validation to reject # or whitespace in spec.path

Thealisyed:
Api validation was added to library-go in order to reject any router that is created with a '#' or whitespace in the route spec.path # as this causes HaProxy error and the ingress to become degraded

### DIFF
--- a/pkg/route/validation/validation.go
+++ b/pkg/route/validation/validation.go
@@ -77,7 +77,11 @@ var (
 )
 
 func ValidateRoute(ctx context.Context, route *routev1.Route, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
-	return validateRoute(ctx, route, true, sarCreator, secretsGetter, opts)
+	allErrs := field.ErrorList{}
+	pathFieldPath := field.NewPath("spec", "path")
+	allErrs = append(allErrs, validatePath(route.Spec.Path, pathFieldPath)...)
+	allErrs = append(allErrs, validateRoute(ctx, route, true, sarCreator, secretsGetter, opts)...)
+	return allErrs
 }
 
 // validLabels - used in the ValidateRouteUpdate function to check if "older" routes conform to DNS1123Labels or not
@@ -215,11 +219,29 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
 	return result
 }
 
+// validatePath tests the spec.path field for invalid syntax when the
+// rewrite-target annotation is set. This addresses OCPBUGS-47773 by rejecting invalid spec.path
+// values, while preserving compatibility for users who may have depended on the unintended
+// behavior caused by the bug.
+func validatePath(pathValue string, fldPath *field.Path) field.ErrorList {
+	result := field.ErrorList{}
+
+	if strings.ContainsAny(pathValue, "# ") {
+		result = append(result, field.Invalid(fldPath, pathValue, "cannot contain # or spaces"))
+	}
+
+	return result
+}
+
 func ValidateRouteUpdate(ctx context.Context, route *routev1.Route, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, secrets corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
 	allErrs := validateObjectMetaUpdate(&route.ObjectMeta, &older.ObjectMeta, field.NewPath("metadata"))
 	allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(route.Spec.WildcardPolicy, older.Spec.WildcardPolicy, field.NewPath("spec", "wildcardPolicy"))...)
 	hostnameUpdated := route.Spec.Host != older.Spec.Host
 	allErrs = append(allErrs, validateRoute(ctx, route, hostnameUpdated && validLabels(older.Spec.Host), sarc, secrets, opts)...)
+	if route.Spec.Path != older.Spec.Path {
+		pathFieldPath := field.NewPath("spec", "path")
+		allErrs = append(allErrs, validatePath(route.Spec.Path, pathFieldPath)...)
+	}
 	return allErrs
 }
 
--- a/pkg/route/validation/validation_test.go
+++ b/pkg/route/validation/validation_test.go
@@ -112,6 +112,18 @@ func TestValidateRoute(t *testing.T) {
 	var headerNameXSS string = "X-XSS-Protection"
 	invalidNumRequests := make([]routev1.RouteHTTPHeader, maxRequestHeaderList+1)
 	invalidNumResponses := make([]routev1.RouteHTTPHeader, maxResponseHeaderList+1)
+	routeWithPath := func(path string) *routev1.Route {
+		return &routev1.Route{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "wildcardpolicy",
+				Namespace: "foo",
+			},
+			Spec: routev1.RouteSpec{
+				To:   createRouteSpecTo("serviceName", "Service"),
+				Path: path,
+			},
+		}
+	}
 	tests := []struct {
 		name           string
 		route          *routev1.Route
@@ -1007,13 +1019,34 @@ func TestValidateRoute(t *testing.T) {
 			},
 			expectedErrors: 1,
 		},
+		{
+			name:           "spec.path should not contain hash symbol in the spec.path",
+			route:          routeWithPath("/new-path#"),
+			expectedErrors: 1,
+		},
+		{
+			name:           "spec.path should not contain whitespace in the spec.path",
+			route:          routeWithPath("/new path"),
+			expectedErrors: 1,
+		},
+		{
+			name:           "spec.path should not contain whitespace or # in the spec.path",
+			route:          routeWithPath("/invalid-path # "),
+			expectedErrors: 1,
+		},
+		{
+			name:           "spec.path is valid", //accepts a valid spec.path
+			route:          routeWithPath("/valid-path"),
+			expectedErrors: 0,
+		},
 	}
-
 	for _, tc := range tests {
-		errs := ValidateRoute(context.Background(), tc.route, &testSARCreator{allow: false}, &testSecretGetter{}, routecommon.RouteValidationOptions{AllowExternalCertificates: false})
-		if len(errs) != tc.expectedErrors {
-			t.Errorf("Test case %s expected %d error(s), got %d. %v", tc.name, tc.expectedErrors, len(errs), errs)
-		}
+		t.Run(tc.name, func(t *testing.T) {
+			errs := ValidateRoute(context.Background(), tc.route, &testSARCreator{allow: false}, &testSecretGetter{}, routecommon.RouteValidationOptions{AllowExternalCertificates: false})
+			if len(errs) != tc.expectedErrors {
+				t.Errorf("Test case %s expected %d error(s), got %d. %v", tc.name, tc.expectedErrors, len(errs), errs)
+			}
+		})
 	}
 }
 
@@ -2446,6 +2479,111 @@ func TestValidateRouteUpdate(t *testing.T) {
 			change: func(route *routev1.Route) {
 				route.Spec.Host = "name-namespace-1234567890-1234567890-1234567890-1234567890-12345.test.com"
 			}, // new route is invalid - do labels check
+			expectedErrors: 1,
+		},
+		{
+			name: "Reject update to spec.path containing #",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "valid-route",
+					Namespace:       "foo",
+					ResourceVersion: "1",
+				},
+				Spec: routev1.RouteSpec{
+					Path: "/valid-path", // Initial valid path
+					To: routev1.RouteTargetReference{
+						Name: "serviceName",
+						Kind: "Service",
+					},
+				},
+			},
+			change: func(route *routev1.Route) {
+				route.Spec.Path = "/path#invalid" // Introduce #
+			},
+			expectedErrors: 1,
+		},
+		{
+			name: "Reject update to spec.path containing whitespace",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "valid-route",
+					Namespace:       "foo",
+					ResourceVersion: "1",
+				},
+				Spec: routev1.RouteSpec{
+					Path: "/valid-path",
+					To: routev1.RouteTargetReference{
+						Name: "serviceName",
+						Kind: "Service",
+					},
+				},
+			},
+			change: func(route *routev1.Route) {
+				route.Spec.Path = "/invalid path" // Introduce whitespace
+			},
+			expectedErrors: 1,
+		},
+		{
+			name: "Allow update without changing spec.path",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "valid-route",
+					Namespace:       "foo",
+					ResourceVersion: "1",
+				},
+				Spec: routev1.RouteSpec{
+					Path: "/valid-path",
+					To: routev1.RouteTargetReference{
+						Name: "serviceName",
+						Kind: "Service",
+					},
+				},
+			},
+			change: func(route *routev1.Route) {
+				route.Spec.Host = "new.host.example.com" // Change unrelated field
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "Allow updates without modifying invalid spec.path",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "valid-route",
+					Namespace:       "foo",
+					ResourceVersion: "1",
+				},
+				Spec: routev1.RouteSpec{
+					Path: "/path#invalid", // Pre-existing invalid path
+					To: routev1.RouteTargetReference{
+						Name: "serviceName",
+						Kind: "Service",
+					},
+				},
+			},
+			change: func(route *routev1.Route) {
+				route.Spec.Host = "new.host.example.com"
+			},
+			expectedErrors: 0,
+		},
+		{
+			name: "Reject update that modifies invalid spec.path to another invalid value",
+			route: &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "invalid-route",
+					Namespace:       "foo",
+					ResourceVersion: "1",
+				},
+				Spec: routev1.RouteSpec{
+					Path: "/path with space", // Pre-existing invalid path
+					To: routev1.RouteTargetReference{
+						Name: "serviceName",
+						Kind: "Service",
+					},
+				},
+			},
+			change: func(route *routev1.Route) {
+				route.Spec.Path = "/path#invalid" // Change to another invalid path
+			},
 			expectedErrors: 1,
 		},
 	}


pkg/route/validation/validation.go
```diff
@@ -212,6 +212,27 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
 		result = append(result, errs...)
 	}
 
+	if errs := ValidatePathWithSpecRoute(route, specPath.Child("path")); len(errs) != 0 {
```
candita, line 215: See https://github.com/openshift/library-go/blob/master/pkg/route/validation/validation.go#L221-L222 for an idea on how to implement ratcheting, e.g. making sure that this applies only to new routes being created, and not to old routes being updated.

candita, line 215: See https://github.com/openshift/library-go/blob/master/pkg/route/validation/validation.go#L221-L222 for an idea on how to implement ratcheting, e.g. making sure that this applies only to new routes being created, and not to old routes being updated.

candita, line 215: See https://github.com/openshift/library-go/blob/master/pkg/route/validation/validation.go#L221-L222 for an idea on how to implement ratcheting, e.g. making sure that this applies only to new routes being created, and not to old routes being updated.

pkg/route/validation/validation.go
```diff
@@ -212,6 +212,27 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+		result = append(result, errs...)
+	}
+
+	return result
+}
+
+// ValidatePathWithSpecRoute tests the spec.path field for invalid syntax when the
+// rewrite-target annotation is set. This addresses OCPBUGS-47773 by rejecting invalid spec.path
+// values, while preserving compatibility for users who may have depended on the unintended
+// behavior caused by the bug.
```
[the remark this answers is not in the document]
candita: We do ratcheting because other code uses route validation.

pkg/route/validation/validation.go
Thealisyed: Implemented ratcheting

pkg/route/validation/validation.go
```diff
@@ -76,8 +76,8 @@ var (
 	permittedResponseHeaderValueRE = regexp.MustCompile(strings.Replace(permittedHeaderValueTemplate, "XYZ", "res", 1))
 )
 
-func ValidateRoute(ctx context.Context, route *routev1.Route, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
-	return validateRoute(ctx, route, true, sarCreator, secretsGetter, opts)
+func ValidateRoute(ctx context.Context, route *routev1.Route, older *routev1.Route, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
+	return validateRoute(ctx, route, older, true, sarCreator, secretsGetter, opts)
```
Miciah, line 80: We have `ValidateRoute` for validating on create and `ValidateRouteUpdate` for validating on update.  Overloading `ValidateRoute` to handle both the create and the update cases goes against conventions for API validation logic and makes it difficult to follow the logic.  

`validateRoute` has the common logic.  I'd keep that the same.  Then `ValidateRoute` can have the create-specific check:
```go
if strings.Contains(...) {
	result = append(result, field.Invalid(...))
}
```
And `ValidateRouteUpdate` can have the update-specific logic:
```go
if route.Spec.Path != older.Spec.Path {
	if strings.Contains(...) { 
		result = append(result, field.Invalid(...))
	}
}
```
`ValidatePathWithSpecRoute` can be simplified:
```go
func ValidatePath(pathValue string, fldPath *field.Path) field.ErrorList {
	result := field.ErrorList{}

	if strings.Contains(pathValue, "#") {
		result = append(result, field.Invalid(fldPath, pathValue, "cannot contain #"))
	}
	if strings.Contains(pathValue, " ") {
		result = append(result, field.Invalid(fldPath, pathValue, "cannot contain spaces"))
	}

	return result
}
```
Then `ValidateRoute` and `ValidateRouteUpdate` can both use `ValidatePath`.

pkg/route/validation/validation.go
```diff
@@ -212,14 +212,48 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+		return result
+	}
+
+	if route.Spec.Path != older.Spec.Path {
+		if strings.Contains(route.Spec.Path, "#") {
+			result = append(result, field.Invalid(fldPath, route.Spec.Path, "cannot contain # in the spec.path"))
+		}
+		if strings.Contains(route.Spec.Path, " ") {
+			result = append(result, field.Invalid(fldPath, route.Spec.Path, "cannot contain whitespace in the spec.path"))
+		}
```
Miciah, line 246: These two `if` blocks have the same logic.  Please refactor to avoid the redundancy.

pkg/route/validation/validation.go
```diff
@@ -212,14 +212,48 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+	result := field.ErrorList{}
+
+	// If older is nil, this is a new route
+	if older == nil {
+		if strings.Contains(route.Spec.Path, "#") {
+			result = append(result, field.Invalid(fldPath, route.Spec.Path, "cannot contain # in the spec.path"))
+		}
+		if strings.Contains(route.Spec.Path, " ") {
+			result = append(result, field.Invalid(fldPath, route.Spec.Path, "cannot contain whitespace in the spec.path"))
+		}
```
Miciah, line 236: Consider using [`strings.ContainsAny`](https://pkg.go.dev/strings#ContainsAny).  The error message would need to be more generic, but that's not necessarily a bad thing, and it makes it easy to add additional characters should the need arise.
```suggestion
		if strings.ContainsAny(route.Spec.Path, "# ") {
			result = append(result, field.Invalid(fldPath, route.Spec.Path, "cannot contain # or spaces"))
		}
```

Miciah:
Please add test cases in `TestValidateRouteUpdate`.

pkg/route/validation/validation.go
Thealisyed: Thanks!

Thealisyed:
I've addressed all the review comments and added the corresponding test cases in `TestValidateRouteUpdate`. Please have a look and let me know if everything looks good or if any adjustments are needed. 
Thanks @Miciah

ShudiLi:
tested it with 4.19.0-0-2025-05-07-031421-test-ci-ln-0yn4242-latest
```
1. created 4 routes with path having space, # characters
% oc -n test get route
NAME          HOST/PORT                                                                     PATH                             SERVICES      PORT          TERMINATION   WILDCARD
edge1         edge1-test.apps.ci-ln-0yn4242-72292.origin-ci-int-gce.dev.rhcloud.com         /route-admission-test!#          unsec-apach   unsec-apach   edge          None
reencrypt1    reencrypt1-test.apps.ci-ln-0yn4242-72292.origin-ci-int-gce.dev.rhcloud.com    /route-admission-test1 / test1   sec-apach2    sec-apach2    reencrypt     None
reencrypt2    reencrypt2-test.apps.ci-ln-0yn4242-72292.origin-ci-int-gce.dev.rhcloud.com    /route-admission-test!#          sec-apach2    sec-apach2    reencrypt     None
unsec-apach   unsec-apach-test.apps.ci-ln-0yn4242-72292.origin-ci-int-gce.dev.rhcloud.com   /route-admission-test1 / test1   unsec-apach   unsec-apach                 None

2.
% oc -n test get route -oyaml | grep haproxy.router.openshift.io/rewrite-target
      haproxy.router.openshift.io/rewrite-target: /
      haproxy.router.openshift.io/rewrite-target: /
      haproxy.router.openshift.io/rewrite-target: /
      haproxy.router.openshift.io/rewrite-target: /

3.
% oc get co ingress
NAME      VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
ingress   4.19.0-0-2025-05-07-031421-test-ci-ln-0yn4242-latest   True        False         False      118m    

4.
% oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.19.0-0-2025-05-07-031421-test-ci-ln-0yn4242-latest   True        False         109m    Cluster version is 4.19.0-0-2025-05-07-031421-test-ci-ln-0yn4242-latest
```

ShudiLi:
/label qe-approved
thanks

candita:
@Thealisyed this didn't go through the PR scrub process, so has no assignees and people aren't necessarily getting pinged.  

/assign @Miciah 
/assign

pkg/route/validation/validation.go
```diff
@@ -215,10 +219,28 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
 	return result
 }
 
+// ValidatePath tests the spec.path field for invalid syntax when the
+// rewrite-target annotation is set. This addresses OCPBUGS-47773 by rejecting invalid spec.path
+// values, while preserving compatibility for users who may have depended on the unintended
+// behavior caused by the bug.
+func ValidatePath(pathValue string, fldPath *field.Path) field.ErrorList {
```
candita, line 226: Should be a private function:
```suggestion
func validatePath(pathValue string, fldPath *field.Path) field.ErrorList {
```

pkg/route/validation/validation.go
```diff
@@ -215,10 +219,28 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+
 func ValidateRouteUpdate(ctx context.Context, route *routev1.Route, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, secrets corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
 	allErrs := validateObjectMetaUpdate(&route.ObjectMeta, &older.ObjectMeta, field.NewPath("metadata"))
 	allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(route.Spec.WildcardPolicy, older.Spec.WildcardPolicy, field.NewPath("spec", "wildcardPolicy"))...)
 	hostnameUpdated := route.Spec.Host != older.Spec.Host
+	if route.Spec.Path != older.Spec.Path {
+		pathFieldPath := field.NewPath("spec", "path")
+		allErrs = append(allErrs, ValidatePath(route.Spec.Path, pathFieldPath.Child("path"))...)
+	}
 	allErrs = append(allErrs, validateRoute(ctx, route, hostnameUpdated && validLabels(older.Spec.Host), sarc, secrets, opts)...)
```
candita, line 244: nit: add the new code at the end of the checks, so as to not separate hostnameUpdated and the use of that value.
```suggestion
	allErrs = append(allErrs, validateRoute(ctx, route, hostnameUpdated && validLabels(older.Spec.Host), sarc, secrets, opts)...)
	if route.Spec.Path != older.Spec.Path {
		pathFieldPath := field.NewPath("spec", "path")
		allErrs = append(allErrs, ValidatePath(route.Spec.Path, pathFieldPath.Child("path"))...)
	}
	
```

pkg/route/validation/validation_test.go
```diff
@@ -112,9 +112,22 @@ func TestValidateRoute(t *testing.T) {
...
+			Spec: routev1.RouteSpec{
+				To:   createRouteSpecTo("serviceName", "Service"),
+				Path: path,
+			},
+		}
+	}
 	tests := []struct {
 		name           string
 		route          *routev1.Route
+		older          *routev1.Route
```
candita, line 130: I don't see where the `older` value is being used in the tests.  I don't think this is implemented yet.

pkg/route/validation/validation_test.go
```diff
@@ -1007,13 +1020,50 @@ func TestValidateRoute(t *testing.T) {
...
+			route:          routeWithPath("/valid-path"),
+			older:          nil,
+			expectedErrors: 0,
+		},
+		{
+			name:           "spec.path is valid (updated route)", //accepts a valid spec.path for an updated route
+			route:          routeWithPath("/new-valid-path"),
+			older:          routeWithPath("/old-valid-path"),
+			expectedErrors: 0,
+		},
```
candita, line 1058: It needs one more test to show it accepts an invalid spec.path for an updated route.

pkg/route/validation/validation_test.go
```diff
@@ -1007,13 +1020,50 @@ func TestValidateRoute(t *testing.T) {
...
+			name:           "spec.path should not contain whitespace in the spec.path (new route)",
+			route:          routeWithPath("/foo bar"),
+			older:          nil,
+			expectedErrors: 1,
+		},
+		{
+			name:           "spec.path should not contain hash symbol in the spec.path (updated route)",
+			route:          routeWithPath("/new-path#"),
+			older:          routeWithPath("/old-path"), // Older route with a valid path
+			expectedErrors: 1,
```
candita, line 1039: ValidateRoute doesn't test updates, so this can be removed.

pkg/route/validation/validation_test.go
candita, line 1058: ValidateRoute doesn't test updates, so this can be removed.

pkg/route/validation/validation.go
Thealisyed: Thanks

pkg/route/validation/validation.go
```diff
@@ -215,11 +219,29 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+	return result
+}
+
 func ValidateRouteUpdate(ctx context.Context, route *routev1.Route, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, secrets corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
 	allErrs := validateObjectMetaUpdate(&route.ObjectMeta, &older.ObjectMeta, field.NewPath("metadata"))
 	allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(route.Spec.WildcardPolicy, older.Spec.WildcardPolicy, field.NewPath("spec", "wildcardPolicy"))...)
 	hostnameUpdated := route.Spec.Host != older.Spec.Host
 	allErrs = append(allErrs, validateRoute(ctx, route, hostnameUpdated && validLabels(older.Spec.Host), sarc, secrets, opts)...)
+	if route.Spec.Path != older.Spec.Path {
+		pathFieldPath := field.NewPath("spec", "path")
```
candita, line 242: nit: Ideally this `pathFieldPath` would be specified within `validatePath` instead of passed as a parameter, since it would always be the same value.  I can live with it unless you need to make other changes.

pkg/route/validation/validation.go
```diff
@@ -77,7 +77,11 @@ var (
 )
 
 func ValidateRoute(ctx context.Context, route *routev1.Route, sarCreator routecommon.SubjectAccessReviewCreator, secretsGetter corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
-	return validateRoute(ctx, route, true, sarCreator, secretsGetter, opts)
+	allErrs := field.ErrorList{}
+	pathFieldPath := field.NewPath("spec", "path")
```
candita, line 81: nit: see https://github.com/openshift/library-go/pull/1958/files#r2093506086

pkg/route/validation/validation_test.go
```diff
@@ -1008,12 +1008,13 @@ func TestValidateRoute(t *testing.T) {
 			expectedErrors: 1,
 		},
 	}
-
 	for _, tc := range tests {
-		errs := ValidateRoute(context.Background(), tc.route, &testSARCreator{allow: false}, &testSecretGetter{}, routecommon.RouteValidationOptions{AllowExternalCertificates: false})
-		if len(errs) != tc.expectedErrors {
-			t.Errorf("Test case %s expected %d error(s), got %d. %v", tc.name, tc.expectedErrors, len(errs), errs)
-		}
+		t.Run(tc.name, func(t *testing.T) {
```
candita, line 1012: I think you still need at least one test case that shows # and whitespace are rejected on a new route.

pkg/route/validation/validation_test.go
candita: @Thealisyed here I think you removed ALL the tests, but I meant just remove the `older` variable from the tests.

candita (changes requested):
Almost there.  Looks like you removed too much in the last change.   Also, please update your commit message to include the fact that we allow old routes that had the invalid value to be updated without an error.

pkg/route/validation/validation.go
Miciah: The convention is that if `validateFoo` is validating field `foo` and uses `validateBar` to validate sub-field `foo.bar`, then `validateFoo` passes the path `foo.bar` to `validateBar`.  This keeps the validation functions well scoped and reusable.  For example, if you also have a `validateBaz` that needs to validate `baz.bar`, it can re-use `validateBar` and just pass the path `baz.bar` to `validateBar`.

pkg/route/validation/validation.go
```diff
@@ -215,11 +219,29 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
+}
+
 func ValidateRouteUpdate(ctx context.Context, route *routev1.Route, older *routev1.Route, sarc routecommon.SubjectAccessReviewCreator, secrets corev1client.SecretsGetter, opts routecommon.RouteValidationOptions) field.ErrorList {
 	allErrs := validateObjectMetaUpdate(&route.ObjectMeta, &older.ObjectMeta, field.NewPath("metadata"))
 	allErrs = append(allErrs, apimachineryvalidation.ValidateImmutableField(route.Spec.WildcardPolicy, older.Spec.WildcardPolicy, field.NewPath("spec", "wildcardPolicy"))...)
 	hostnameUpdated := route.Spec.Host != older.Spec.Host
 	allErrs = append(allErrs, validateRoute(ctx, route, hostnameUpdated && validLabels(older.Spec.Host), sarc, secrets, opts)...)
+	if route.Spec.Path != older.Spec.Path {
+		pathFieldPath := field.NewPath("spec", "path")
+		allErrs = append(allErrs, validatePath(route.Spec.Path, pathFieldPath.Child("path"))...)
```
Miciah, line 243: You have an extra `.path`, don't you?

pkg/route/validation/validation_test.go
Thealisyed: Added the test cases

pkg/route/validation/validation.go
Thealisyed: Fixed, thanks!

Thealisyed:
Updated the commit message to be more detailed

wangke19:
> Need this [ART](https://github.com/openshift/openshift-apiserver/pull/521) bump to be merged as there is currently golang and k8s version differences

For main/master branch, we have one pr https://github.com/openshift/openshift-apiserver/pull/486, https://github.com/openshift/openshift-apiserver/pull/521 is for 4.19 branch.

pkg/route/validation/validation.go
```diff
@@ -215,11 +219,29 @@ func validateRoute(ctx context.Context, route *routev1.Route, checkHostname bool
...
 }
 
+// validatePath tests the spec.path field for invalid syntax when the
+// rewrite-target annotation is set. This addresses OCPBUGS-47773 by rejecting invalid spec.path
+// values, while preserving compatibility for users who may have depended on the unintended
+// behavior caused by the bug.
+func validatePath(pathValue string, fldPath *field.Path) field.ErrorList {
+	result := field.ErrorList{}
+
+	if strings.ContainsAny(pathValue, "# ") {
```
candita, line 229: I wonder now if we need to be concerned with tabs, line feeds, etc.  You would have to use a RegEx to find those.  Wdyt @Miciah ?

pkg/route/validation/validation_test.go
```diff
@@ -1007,13 +1019,34 @@ func TestValidateRoute(t *testing.T) {
...
+		{
+			name:           "spec.path should not contain hash symbol in the spec.path",
+			route:          routeWithPath("/new-path#"),
+			expectedErrors: 1,
+		},
+		{
+			name:           "spec.path should not contain whitespace in the spec.path",
+			route:          routeWithPath("/new path"),
+			expectedErrors: 1,
+		},
```
candita, line 1031: Sorry I didn't notice this before, but can you also test input with tabs?

PR #1961, opened 2025-05-06
OPRUN-3883: Add NetworkPolicies to the list of resources

tmshort:
Includes informers, appliers and readers.

### DIFF
--- a/pkg/operator/resource/resourceapply/networking.go
+++ b/pkg/operator/resource/resourceapply/networking.go
@@ -1,0 +1,59 @@
+package resourceapply
+
+import (
+	"context"
+
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	networkingclientv1 "k8s.io/client-go/kubernetes/typed/networking/v1"
+	"k8s.io/klog/v2"
+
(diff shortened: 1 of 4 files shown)


tmshort:
This is needed for OLM support of NetworkPolicies.

tmshort:
Ping @hexfusion @jsafrane

tmshort:
I was able to verify that this code works when pulled in manually to cluster-olm-operator.

pkg/operator/resource/resourceapply/networking.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, false, err
+	}
+
+	modified := false
+	existingCopy := existing.DeepCopy()
+
+	resourcemerge.EnsureObjectMeta(&modified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	if equality.Semantic.DeepEqual(existingCopy.Spec, required.Spec) && !modified {
+		return existingCopy, false, nil
+	}
```
lunarwhite, line 38: Hey @tmshort, is it the intended behavior to not copy `required.Spec` to `existingCopy.Spec` when the specs differ before updating? It keeps attempting to update with the old spec instead. I feel like this would prevent the managed NPs from being reconciled if manual edits are made to the spec.

tmshort: This is likely due to the code originating from other code in resourceapply. Not sure which one since this was back in May. But it's basically checking to see if the Metadata is the same (i.e. `modified` is false), and if the Specs are the same (i.e. `DeepEqual()` is true), then nothing needs to happen.

PR #1965, opened 2025-05-27
fix: ApplyDirectly add missing support for admissionregistrationv1 ValidatingAdmissionPolicy, ValidatingAdmissionPolicyBinding

damdo:
The support for applying admissionregistrationv1 ValidatingAdmissionPolicy, ValidatingAdmissionPolicyBinding was added in https://github.com/openshift/library-go/pull/1748 but that wasn't explicitly added also to ApplyDirectly, which results in an error when ApplyDirectly tries to apply such manifests:

```
"admissionregistration.k8s.io/v1/ValidatingAdmissionPolicy " at position 9: unhandled type *v1.ValidatingAdmissionPolicy
```


damdo:
@p0lyn0mial can we get an approver review on this one? TY

PR #1971, opened 2025-06-10
OCPBUGS-57049: Update cert annotations

vrutkovs:
Move testcase name out of auto-regenerate-after-offline-expiry, add
refresh-period.

Follow-up for https://github.com/openshift/origin/pull/29327
Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1768 and https://github.com/openshift/cluster-authentication-operator/pull/742

### DIFF
--- a/pkg/operator/certrotation/annotations.go
+++ b/pkg/operator/certrotation/annotations.go
@@ -16,9 +16,13 @@ const (
 	CertificateIssuer = "auth.openshift.io/certificate-issuer"
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
-	// AutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR and an e2e test name which verifies
+	// CertificateTestNameAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
+	CertificateTestNameAnnotation string = "certificates.openshift.io/test-name"
+	// CertificateAutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR adding this annotation which verifies
 	// that TLS artifact is correctly regenerated after it has expired
-	AutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	CertificateAutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	// CertificateRefreshPeriodAnnotation is the interval at which the certificate should be refreshed.
+	CertificateRefreshPeriodAnnotation string = "certificates.openshift.io/refresh-period"
 )
 
 type AdditionalAnnotations struct {
@@ -26,13 +30,16 @@ type AdditionalAnnotations struct {
 	JiraComponent string
 	// Description is a human-readable one sentence description of certificate purpose
 	Description string
-	// AutoRegenerateAfterOfflineExpiry contains a link to PR and an e2e test name which verifies
-	// that TLS artifact is correctly regenerated after it has expired
+	// TestName is an e2e test name which verifies that TLS artifact is created and used correctly
+	TestName string
+	// AutoRegenerateAfterOfflineExpiry contains a link to PR which adds this annotation on the TLS artifact
 	AutoRegenerateAfterOfflineExpiry string
 	// NotBefore contains certificate the certificate creation date in RFC3339 format.
 	NotBefore string
 	// NotAfter contains certificate the certificate validity date in RFC3339 format.
 	NotAfter string
+	// RefreshPeriod contains the interval at which the certificate should be refreshed.
+	RefreshPeriod string
 }
 
 func (a AdditionalAnnotations) EnsureTLSMetadataUpdate(meta *metav1.ObjectMeta) bool {
@@ -52,18 +59,34 @@ func (a AdditionalAnnotations) EnsureTLSMetadataUpdate(meta *metav1.ObjectMeta) 
 		meta.Annotations[annotations.OpenShiftDescription] = a.Description
 		modified = true
 	}
-	if len(a.AutoRegenerateAfterOfflineExpiry) > 0 && meta.Annotations[AutoRegenerateAfterOfflineExpiryAnnotation] != a.AutoRegenerateAfterOfflineExpiry {
-		diff := cmp.Diff(meta.Annotations[AutoRegenerateAfterOfflineExpiryAnnotation], a.AutoRegenerateAfterOfflineExpiry)
-		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", AutoRegenerateAfterOfflineExpiryAnnotation, meta.Namespace, meta.Name, diff)
-		meta.Annotations[AutoRegenerateAfterOfflineExpiryAnnotation] = a.AutoRegenerateAfterOfflineExpiry
+	if len(a.TestName) > 0 && meta.Annotations[CertificateTestNameAnnotation] != a.TestName {
+		diff := cmp.Diff(meta.Annotations[CertificateTestNameAnnotation], a.TestName)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateTestNameAnnotation, meta.Name, meta.Namespace, diff)
+		meta.Annotations[CertificateTestNameAnnotation] = a.TestName
+		modified = true
+	}
+	if len(a.AutoRegenerateAfterOfflineExpiry) > 0 && meta.Annotations[CertificateAutoRegenerateAfterOfflineExpiryAnnotation] != a.AutoRegenerateAfterOfflineExpiry {
+		diff := cmp.Diff(meta.Annotations[CertificateAutoRegenerateAfterOfflineExpiryAnnotation], a.AutoRegenerateAfterOfflineExpiry)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateAutoRegenerateAfterOfflineExpiryAnnotation, meta.Namespace, meta.Name, diff)
+		meta.Annotations[CertificateAutoRegenerateAfterOfflineExpiryAnnotation] = a.AutoRegenerateAfterOfflineExpiry
 		modified = true
 	}
 	if len(a.NotBefore) > 0 && meta.Annotations[CertificateNotBeforeAnnotation] != a.NotBefore {
+		diff := cmp.Diff(meta.Annotations[CertificateNotBeforeAnnotation], a.NotBefore)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateNotBeforeAnnotation, meta.Name, meta.Namespace, diff)
 		meta.Annotations[CertificateNotBeforeAnnotation] = a.NotBefore
 		modified = true
 	}
 	if len(a.NotAfter) > 0 && meta.Annotations[CertificateNotAfterAnnotation] != a.NotAfter {
+		diff := cmp.Diff(meta.Annotations[CertificateNotAfterAnnotation], a.NotAfter)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateNotAfterAnnotation, meta.Name, meta.Namespace, diff)
 		meta.Annotations[CertificateNotAfterAnnotation] = a.NotAfter
+		modified = true
+	}
+	if len(a.RefreshPeriod) > 0 && meta.Annotations[CertificateRefreshPeriodAnnotation] != a.RefreshPeriod {
+		diff := cmp.Diff(meta.Annotations[CertificateRefreshPeriodAnnotation], a.RefreshPeriod)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateRefreshPeriodAnnotation, meta.Name, meta.Namespace, diff)
+		meta.Annotations[CertificateRefreshPeriodAnnotation] = a.RefreshPeriod
 		modified = true
 	}
 	return modified
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -92,7 +92,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 			reason = "secret doesn't exist"
 		}
 		c.EventRecorder.Eventf("SignerUpdateRequired", "%q in %q requires a new signing cert/key pair: %v", c.Name, c.Namespace, reason)
-		if err = setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret, c.Validity, c.AdditionalAnnotations); err != nil {
+		if err = setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret, c.Validity, c.Refresh, c.AdditionalAnnotations); err != nil {
 			return nil, false, err
 		}
 
@@ -201,13 +201,13 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 
 // setSigningCertKeyPairSecretAndTLSAnnotations generates a new signing certificate and key pair,
 // stores them in the specified secret, and adds predefined TLS annotations to that secret.
-func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *corev1.Secret, validity time.Duration, tlsAnnotations AdditionalAnnotations) error {
+func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *corev1.Secret, validity, refresh time.Duration, tlsAnnotations AdditionalAnnotations) error {
 	ca, err := setSigningCertKeyPairSecret(signingCertKeyPairSecret, validity)
 	if err != nil {
 		return err
 	}
 
-	setTLSAnnotationsOnSigningCertKeyPairSecret(signingCertKeyPairSecret, ca, tlsAnnotations)
+	setTLSAnnotationsOnSigningCertKeyPairSecret(signingCertKeyPairSecret, ca, refresh, tlsAnnotations)
 	return nil
 }
 
@@ -243,10 +243,11 @@ func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validi
 //
 // These assumptions are safe because this function is only called after the secret
 // has been initialized in setSigningCertKeyPairSecret.
-func setTLSAnnotationsOnSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, ca *crypto.TLSCertificateConfig, tlsAnnotations AdditionalAnnotations) {
+func setTLSAnnotationsOnSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, ca *crypto.TLSCertificateConfig, refresh time.Duration, tlsAnnotations AdditionalAnnotations) {
 	signingCertKeyPairSecret.Annotations[CertificateIssuer] = ca.Certs[0].Issuer.CommonName
 
 	tlsAnnotations.NotBefore = ca.Certs[0].NotBefore.Format(time.RFC3339)
 	tlsAnnotations.NotAfter = ca.Certs[0].NotAfter.Format(time.RFC3339)
+	tlsAnnotations.RefreshPeriod = refresh.String()
 	_ = tlsAnnotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta)
 }
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -121,7 +121,7 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 
 	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired, creationRequired); len(reason) > 0 {
 		c.EventRecorder.Eventf("TargetUpdateRequired", "%q in %q requires a new target cert/key pair: %v", c.Name, c.Namespace, reason)
-		if err = setTargetCertKeyPairSecretAndTLSAnnotations(targetCertKeyPairSecret, c.Validity, signingCertKeyPair, c.CertCreator, c.AdditionalAnnotations); err != nil {
+		if err = setTargetCertKeyPairSecretAndTLSAnnotations(targetCertKeyPairSecret, c.Validity, c.Refresh, signingCertKeyPair, c.CertCreator, c.AdditionalAnnotations); err != nil {
 			return nil, err
 		}
 
@@ -234,13 +234,13 @@ func needNewTargetCertKeyPairForTime(annotations map[string]string, signer *cryp
 
 // setTargetCertKeyPairSecretAndTLSAnnotations generates a new cert/key pair,
 // stores them in the specified secret, and adds predefined TLS annotations to that secret.
-func setTargetCertKeyPairSecretAndTLSAnnotations(targetCertKeyPairSecret *corev1.Secret, validity time.Duration, signer *crypto.CA, certCreator TargetCertCreator, tlsAnnotations AdditionalAnnotations) error {
+func setTargetCertKeyPairSecretAndTLSAnnotations(targetCertKeyPairSecret *corev1.Secret, validity, refresh time.Duration, signer *crypto.CA, certCreator TargetCertCreator, tlsAnnotations AdditionalAnnotations) error {
 	certKeyPair, err := setTargetCertKeyPairSecret(targetCertKeyPairSecret, validity, signer, certCreator)
 	if err != nil {
 		return err
 	}
 
-	setTLSAnnotationsOnTargetCertKeyPairSecret(targetCertKeyPairSecret, certKeyPair, certCreator, tlsAnnotations)
+	setTLSAnnotationsOnTargetCertKeyPairSecret(targetCertKeyPairSecret, certKeyPair, certCreator, refresh, tlsAnnotations)
 	return nil
 }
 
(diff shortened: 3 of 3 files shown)


vrutkovs:
/hold cancel

Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1768

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
...
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
-	// AutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR and an e2e test name which verifies
+	// CertificateTestNameAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
+	CertificateTestNameAnnotation string = "certificates.openshift.io/test-name"
+	// CertificateAutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR and an e2e test name which verifies
 	// that TLS artifact is correctly regenerated after it has expired
-	AutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	CertificateAutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	// CertificateRefreshPeriodAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
```
sjenning: This comment is not correct.  Additionally, since this annotation does not use a standard duration representation, the details of what comes out of `durationRound()` should be detailed in this comment.

pkg/operator/certrotation/utils.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"strconv"
+	"time"
+)
+
+// durationRound formats a duration into a human-readable string.
+func durationRound(d time.Duration) string {
+	u := uint64(d)
+	neg := d < 0
+	if neg {
```
sjenning: Why not do away with `neg` and just do `if d < 0`?

Also, why do we just sign flip it if negative?

Also, why not just use `String()` of `time.Duration`?

vrutkovs: This is a copy of [sprig's implementation](https://github.com/gomodules/sprig/blob/master/date.go#L97-L139), which can accept negative durations (this is why the flip is required). We don't need `neg` anymore so yes, I think its worth being removed.

The issue with `time.Duration.String()` is that it returns string like `72h00m` instead of `3d` .

Added these notes as comments in the code too.

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
...
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
-	// AutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR and an e2e test name which verifies
+	// CertificateTestNameAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
+	CertificateTestNameAnnotation string = "certificates.openshift.io/test-name"
+	// CertificateAutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR adding this annotation which verifies
 	// that TLS artifact is correctly regenerated after it has expired
-	AutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	CertificateAutoRegenerateAfterOfflineExpiryAnnotation string = "certificates.openshift.io/auto-regenerate-after-offline-expiry"
+	// CertificateRefreshPeriodAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
```
p0lyn0mial: This comment is not accurate. Please update it.

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
...
 type AdditionalAnnotations struct {
 	// JiraComponent annotates tls artifacts so that owner could be easily found
 	JiraComponent string
 	// Description is a human-readable one sentence description of certificate purpose
 	Description string
-	// AutoRegenerateAfterOfflineExpiry contains a link to PR and an e2e test name which verifies
-	// that TLS artifact is correctly regenerated after it has expired
+	// TestName is an e2e test name which verifies that TLS artifact is created and used correctly
+	TestName string
+	// AutoRegenerateAfterOfflineExpiry contains a link to PR
```
p0lyn0mial: Please add more info. A link to which PR is required ?

pkg/operator/certrotation/target.go
```diff
@@ -258,8 +258,10 @@ func setTargetCertKeyPairSecret(targetCertKeyPairSecret *corev1.Secret, validity
 	if err != nil {
 		return err
 	}
+
 	annotations.NotBefore = certKeyPair.Certs[0].NotBefore.Format(time.RFC3339)
 	annotations.NotAfter = certKeyPair.Certs[0].NotAfter.Format(time.RFC3339)
+	annotations.RefreshPeriod = durationRound(refresh)
```
p0lyn0mial: Why is it okay to store an inaccurate duration? Won’t this value be used for debugging purposes?

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
 	CertificateIssuer = "auth.openshift.io/certificate-issuer"
 	// CertificateHostnames contains the hostnames used by a signer.
 	CertificateHostnames = "auth.openshift.io/certificate-hostnames"
-	// AutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR and an e2e test name which verifies
+	// CertificateTestNameAnnotation is an e2e test name which verifies that TLS artifact is created and used correctly
+	CertificateTestNameAnnotation string = "certificates.openshift.io/test-name"
```
p0lyn0mial, line 20: Is it a name of an e2e test in origin repo?

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
...
+	// CertificateAutoRegenerateAfterOfflineExpiryAnnotation contains a link to PR adding this annotation which verifies
```
p0lyn0mial, line 21: I have just checked the TLS registry and found:
```
 {
              "key": "certificates.openshift.io/auto-regenerate-after-offline-expiry",
              "value": "https://github.com/openshift/cluster-kube-apiserver-operator/pull/1631,'[sig-cli] Kubectl logs logs should be able to retrieve and filter logs  [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]'"
            },
 ```
 
 does it mean we also store the name of the test that verifies expiry ?

File: pkg/operator/certrotation/annotations.go (hunk 6 above)
Comment (p0lyn0mial) [reply pkg/operator/certrotation/annotations.go:20]: How will this annotation be used?

Comment (vrutkovs) [reply pkg/operator/certrotation/annotations.go:20]: This is the name of e2e test, yes. This test will be running during suspend/shutdown test in [this task](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/e2e/cert-rotation/test/openshift-e2e-cert-rotation-test-commands.sh#L13-L24)

File: pkg/operator/certrotation/annotations.go (hunk 7 above)
Comment (vrutkovs) [reply pkg/operator/certrotation/annotations.go:21]: We want to break this down in two parts:
* PR, which adds this annotation
* test name which verifies that the cert is updated and used correctly

The reason for this split is that a single test is sufficient to verify that the cert is fine, but we want to keep adding more metadata (i.e. refresh period) and so there is no need to copy test name in every annotation

File: pkg/operator/certrotation/annotations.go (hunk 3 above)
Comment (vrutkovs) [reply pkg/operator/certrotation/annotations.go]: Fixed

File: pkg/operator/certrotation/annotations.go (hunk 4 above)
Comment (vrutkovs) [reply pkg/operator/certrotation/annotations.go]: Updated

File: pkg/operator/certrotation/annotations.go (hunk 7 above)
Comment (p0lyn0mial) [reply pkg/operator/certrotation/annotations.go:21]: will the test name be stored under `certificates.openshift.io/test-name` annotation in the future?

File: pkg/operator/certrotation/annotations.go [hunk 8]
```diff
@@ -16,23 +16,30 @@ const (
...
+	// TestName is an e2e test name which verifies that TLS artifact is created and used correctly
+	TestName string
+	// AutoRegenerateAfterOfflineExpiry contains a link to PR which adds this annotation on the TLS artifact
 	AutoRegenerateAfterOfflineExpiry string
 	// NotBefore contains certificate the certificate creation date in RFC3339 format.
 	NotBefore string
 	// NotAfter contains certificate the certificate validity date in RFC3339 format.
 	NotAfter string
+	// RefreshPeriod contains the interval at which the certificate should be refreshed.
+	RefreshPeriod string
```
p0lyn0mial, line 42: In what format will the date be stored? I think this might be useful for tools that process the raw TLS info.

pkg/operator/certrotation/annotations.go
```diff
@@ -16,23 +16,30 @@ const (
...
 	// JiraComponent annotates tls artifacts so that owner could be easily found
 	JiraComponent string
 	// Description is a human-readable one sentence description of certificate purpose
 	Description string
-	// AutoRegenerateAfterOfflineExpiry contains a link to PR and an e2e test name which verifies
-	// that TLS artifact is correctly regenerated after it has expired
+	// TestName is an e2e test name which verifies that TLS artifact is created and used correctly
+	TestName string
+	// AutoRegenerateAfterOfflineExpiry contains a link to PR which adds this annotation on the TLS artifact
 	AutoRegenerateAfterOfflineExpiry string
```
p0lyn0mial, line 36: Also, should we rename this field to `AutoRegenerateAfterOfflineExpiryURL` ? 
Also, should we rename the annotation to `certificates.openshift.io/auto-regenerate-after-offline-expiry-url` ?

Maybe not now but as part of future refactoring.

pkg/operator/certrotation/annotations.go
p0lyn0mial, line 42: Also atm we store an inaccurate value in that field. Is that OK?

pkg/operator/certrotation/annotations.go
p0lyn0mial: awesome!

pkg/operator/certrotation/annotations.go
vrutkovs: Yes, correct - one test case is usually sufficient

pkg/operator/certrotation/annotations.go
vrutkovs: Its using `time.Duration` format: `8h` or `30d` or `10y`

p0lyn0mial: I think that the duration will be changed by the `durationRound` function which `formats a duration into a human-readable string` so machines won't be able to process it (if needed).

pkg/operator/certrotation/annotations.go
vrutkovs: The annotation is meant to indicate that this cert supports auto regeneration, so its presence in any format means this cert can be rotated. So far this annotation contains a PR where is has been set, but later on we might want to reuse it for a more structured info.

I'd prefer not to add `URL` in the name to avoid renaming it later

pkg/operator/certrotation/annotations.go
p0lyn0mial: where will this field be used ?

pkg/operator/certrotation/annotations.go
p0lyn0mial: OK

pkg/operator/certrotation/annotations.go
vrutkovs: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1768 would set it for cluster-kube-apiserver-operator (see TLS artifacts in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1768):
```
            {
              "key": "certificates.openshift.io/refresh-period",
              "value": "8y"
            },
```
It also shows 6h refresh in 4.20 now, as we're using a faster refresh rotation in dev branch

pkg/operator/certrotation/annotations.go
```diff
@@ -52,20 +59,36 @@ func (a AdditionalAnnotations) EnsureTLSMetadataUpdate(meta *metav1.ObjectMeta)
...
 	if len(a.NotAfter) > 0 && meta.Annotations[CertificateNotAfterAnnotation] != a.NotAfter {
+		diff := cmp.Diff(meta.Annotations[CertificateNotAfterAnnotation], a.NotAfter)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateNotAfterAnnotation, meta.Name, meta.Namespace, diff)
 		meta.Annotations[CertificateNotAfterAnnotation] = a.NotAfter
 		modified = true
 	}
+	if len(a.RefreshPeriod) > 0 && meta.Annotations[CertificateRefreshPeriodAnnotation] != a.RefreshPeriod {
+		diff := cmp.Diff(meta.Annotations[CertificateRefreshPeriodAnnotation], a.RefreshPeriod)
+		klog.V(2).Infof("Updating %q annotation for %s/%s, diff: %s", CertificateRefreshPeriodAnnotation, meta.Name, meta.Namespace, diff)
+		meta.Annotations[CertificateRefreshPeriodAnnotation] = a.RefreshPeriod
```
p0lyn0mial, line 89: why in the other places we use `durationRound` but not in this one?

pkg/operator/certrotation/signer.go
```diff
@@ -223,6 +223,7 @@ func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validi
 	signingCertKeyPairSecret.Data["tls.key"] = keyBytes.Bytes()
 	annotations.NotBefore = ca.Certs[0].NotBefore.Format(time.RFC3339)
 	annotations.NotAfter = ca.Certs[0].NotAfter.Format(time.RFC3339)
+	annotations.RefreshPeriod = durationRound(refresh)
```
p0lyn0mial: won't this value be overwritten by `EnsureTLSMetadataUpdate` which is called 2 lines below?

pkg/operator/certrotation/annotations.go
vrutkovs: Its informational only. We don't limit or verify this that it indeed happens at specified time (however we do ensure that this certificate is being rotated at all with ShortCertRotation featuregate)

pkg/operator/certrotation/signer.go
p0lyn0mial: it looks like it won't, `AdditionalAnnotations` is authoritative.

pkg/operator/certrotation/annotations.go
p0lyn0mial: Won’t this info be stored in a must-gather? If so, a developer might be misled by the inaccurate value while debugging a certificate-related issue, right?

p0lyn0mial: For example I could use `kubectl-dev_tool certs inspect-certs` for inspection. In that case I would expect the other tools to do the reformatting: `refreshPeriod: 72h3m0s (3d)`

pkg/operator/certrotation/signer.go
vrutkovs: Yes, `EnsureTLSMetadatUpdate` would update it only if it has changed (i.e. during OCP upgrade)

pkg/operator/certrotation/annotations.go
vrutkovs: In annotations `RefreshPeriod` is a string, it may be directly in the manifest. In the controller, we need to ensure  `RefreshPeriod` corresponds to actual controller refresh duration.

p0lyn0mial: How to we ensure that all controllers pass a value processed by `durationRound` ?

vrutkovs: After this is applied, [refresh period report in TLS registry](https://github.com/openshift/origin/blob/main/tls/refresh-period/refresh-period.md) will be updated, and we'll make sure all necessary controllers are setting the time correctly

p0lyn0mial: but does it mean that the callers will how use `durationRound` ?

say, i am going to add a new annotation to `https://github.com/p0lyn0mial/cluster-kube-apiserver-operator/blob/af13d7a097b5f42fde81964f42fe834e981de3b6/pkg/operator/certrotationcontroller/certrotationcontroller.go#L267`

pkg/operator/certrotation/annotations.go
vrutkovs: Hmm, initially I didn't think it would be a significant issue - most of our certificates are multiples of days - but a mix of days + hours may happen as a result of ShortCertRotation changes (or set by other controller).

I updated the implementation to output "5d4h25m" instead of rounding the duration and added unit tests

p0lyn0mial: Why can't we preserve the original value, state what format it is in, and adjust the other tools like the script that generates the registry to format the value differently?

pkg/operator/certrotation/annotations.go
vrutkovs: Not necessarily, in most cases controllers from library-go don't require any changes.
Implementations are allowed to set this manually - i.e. hardcode the string in the YAML manifest.

In case of  `localhost-serving-signer` secret, the controller already has `Refresh` set to `8 * 365 * 24 * time.Hour`. So when the secret would require an update, the annotation would be set automatically when `setSigningCertKeyPairSecret` is called. (todo for me is to ensure that this annotation is backfilled for existing secrets later)

pkg/operator/certrotation/annotations.go
vrutkovs: This isn't meant to be part of any kind of contract or requirement. It's an informational field only, we have no way to verify that it's correct - so it should just me human-readable.
Are we concerned about different formats of time.Duration?

p0lyn0mial: I think it would be better if all dates were in the same format for consistency (NotBefore, NotAfter). 

This way, the processing tools (the TLS script or cluster-debug-tools) would know what to expect and could reformat them into a more human-readable form if needed during output.

How hard would it be to move `durationToString` to the TLS registry script?

pkg/operator/certrotation/signer.go
```diff
@@ -200,7 +200,7 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 }
 
 // setSigningCertKeyPairSecret creates a new signing cert/key pair and sets them in the secret
-func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validity time.Duration, annotations AdditionalAnnotations) error {
+func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validity, refresh time.Duration, annotations AdditionalAnnotations) error {
```
p0lyn0mial: Also, I'd like to avoid passing params that are used only for annotations and decided to create https://github.com/openshift/library-go/pull/1980 

PTAL and let me know if this approach also make sense to you.

pkg/operator/certrotation/annotations.go
vrutkovs: This isn't a date, it's a duration. Its not the time when next refresh occurs, it would be notAfter+refreshTime.

We can store them as milliseconds, but it won't be human-friendly - and if there is a need for a tool which reads this annotation we might as well convert "5s" into milliseconds too. The format is no longer lossy anyway

p0lyn0mial: or just store `duration.String()`, at least it would be accepted by https://pkg.go.dev/time#ParseDuration

it would be then possible to convert to a human-readable form on the receiving end, before displaying to end users.

p0lyn0mial: whereas `8y` is not accepted by the `time.PraseDuration` - xref: https://go.dev/play/p/6qztJoDUlJJ

vrutkovs: ya, that would store `72h00m00s` instead of `3d` in annotations, this is why we need a custom `Durations.ToBetterString()`

vrutkovs:
TODO:
* Store `time.Duration.String()` value. 
   It may not be as easily approachable by users, but at least it can be easily parsed by `time.ParseDuration`
* Rewrite the value we store in TLS artifacts with a custom rewrite functions.
   Markdowns we generate are meant to be human-readable, and we don't have to be backwards-compatible on those.

p0lyn0mial:
> See follow-up in https://github.com/openshift/library-go/pull/1981

LGTM

let's rebase this pr and merge it, and then let's merge the follow-up pr

pkg/operator/certrotation/target.go
```diff
@@ -232,16 +233,22 @@ func needNewTargetCertKeyPairForTime(annotations map[string]string, signer *cryp
 	return ""
 }
 
-// setTargetCertKeyPairSecret creates a new cert/key pair and sets them in the secret.  Only one of client, serving, or signer rotation may be specified.
+// setTargetCertKeyPairSecretAndTLSAnnotations creates a new cert/key pair and sets them in the secret.  Only one of client, serving, or signer rotation may be specified.
```
p0lyn0mial: let's open a new pr for refactoring the target. i can do that. just let me know.

vrutkovs: Sure, please do. I'll focus on e2e tests

p0lyn0mial: ptal https://github.com/openshift/library-go/pull/1982

pkg/operator/certrotation/signer.go
```diff
@@ -92,6 +92,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 			reason = "secret doesn't exist"
 		}
 		c.EventRecorder.Eventf("SignerUpdateRequired", "%q in %q requires a new signing cert/key pair: %v", c.Name, c.Namespace, reason)
+		c.AdditionalAnnotations.RefreshPeriod = c.Refresh.String()
 		if err = setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret, c.Validity, c.AdditionalAnnotations); err != nil {
```
p0lyn0mial: let's pass `c.Refresh` to `setSigningCertKeyPairSecretAndTLSAnnotations` and then to `setTLSAnnotationsOnSigningCertKeyPairSecret` so that annotation assignments are in one place for the signer.

pkg/operator/certrotation/target.go
```diff
@@ -121,6 +121,7 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 
 	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired, creationRequired); len(reason) > 0 {
 		c.EventRecorder.Eventf("TargetUpdateRequired", "%q in %q requires a new target cert/key pair: %v", c.Name, c.Namespace, reason)
+		c.AdditionalAnnotations.RefreshPeriod = c.Refresh.String()
 		if err = setTargetCertKeyPairSecretAndTLSAnnotations(targetCertKeyPairSecret, c.Validity, signingCertKeyPair, c.CertCreator, c.AdditionalAnnotations); err != nil {
```
p0lyn0mial: i would also pass `c.Refresh` to `setTargetCertKeyPairSecretAndTLSAnnotations` and then to `setTLSAnnotationsOnTargetCertKeyPairSecret` so that annotation assignments are in one place for the target.

pkg/operator/certrotation/annotations.go
p0lyn0mial, line 89: do we need a unit test that checks assignment to the new annotation ?

pkg/operator/certrotation/signer.go
```diff
@@ -92,6 +92,7 @@ func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*
 			reason = "secret doesn't exist"
 		}
 		c.EventRecorder.Eventf("SignerUpdateRequired", "%q in %q requires a new signing cert/key pair: %v", c.Name, c.Namespace, reason)
+		c.AdditionalAnnotations.RefreshPeriod = c.Refresh.String()
```
p0lyn0mial: should this annotation be also set in line `83` ?

```
if !c.RefreshOnlyWhenExpired {
		needsMetadataUpdate := ensureOwnerRefAndTLSAnnotations(signingCertKeyPairSecret, c.Owner, c.AdditionalAnnotations)
		needsTypeChange := ensureSecretTLSTypeSet(signingCertKeyPairSecret)
		updateRequired = needsMetadataUpdate || needsTypeChange
	}
```

pkg/operator/certrotation/signer.go
p0lyn0mial: actually, see https://github.com/openshift/library-go/pull/1971#discussion_r2228157172

pkg/operator/certrotation/target.go
p0lyn0mial: actually, see https://github.com/openshift/library-go/pull/1971#discussion_r2228157172

pkg/operator/certrotation/annotations.go
vrutkovs: I don't think we do - some certs are not being refreshed and if it's incorrectly set by the controller we'll see this in TLS registry violations

pkg/operator/certrotation/signer.go
vrutkovs: So far, I'd like to set refresh period only when contents are being refreshed. Later on we'll need a new PR which backfills this on existing certificates

p0lyn0mial: What is the reason we don’t want to set the refresh annotation for existing certificates? Wouldn’t it be easier to set it now?

p0lyn0mial: I think we have at least three options: 

1. Update the callers to set the new annotation. 
2. Update the annotation at the beginning of the EnsureSigningCertKeyPair method. 
3. Modify the controller that calls EnsureSigningCertKeyPair to update the annotation.

p0lyn0mial: I think option 2 is the easiest. 
Setting the annotation at the beginning of the function ensures it is applied to both new and existing secrets.

pkg/operator/certrotation/annotations.go
p0lyn0mial: Ok.

pkg/operator/certrotation/signer.go
vrutkovs: I'd prefer a separate PR because this was not tested on existing PRs. If TRT comes and reverts because it breaks upgrades I'd rather have it already applied for new clusters at least.

I agree that 2. seems like the easiest option so far, but I didn't dig deep into this yet

p0lyn0mial: ok

PR #1972, opened 2025-06-10
WRKLDS-1550: Add pkg/image/registryclient/v2

tchap:
This is a separate module that is compatible with
pkg/image/registryclient, but uses distribution/v3@v3.0.0.

Since distribution/v3@v3.0.0 has removed the client package being used
by library-go currently, or rather moved it into internal, this patch
includes copying the packages from distribution/v3/internal/client into
registryclient/v2/internal. Some of the logic is exported as required
to make oc and openshift-apiserver work as that is where this patch has been tested.
See the linked PRs below.

So the approach taken was to copy registryclient into registryclient/v2,
then copy distribution/v3/internal/client into
registryclient/v2/internal, then keep exporting functionality so that oc
can be ported to distributions/v3@v3.0.0 and registryclient/v2.

Generally when a project importing library-go wants to migrate to
the current distribution/v3, they can just import this module instead of
using pkg/image/registryclient. The interface is compatible with
pkg/image/registryclient and all that is needed is changing the import
path.

### Note for the Reviewer

Although there is a lot of code, there is paradoxically not much to review. All code is actually copied from other sources and imports paths are amended so that the module works as a whole. Tests are also imported from the respective sources.

See the following PRs across our repos for exemplar usage:
- https://github.com/openshift/oc/pull/2036
- https://github.com/openshift/openshift-apiserver/pull/532


tchap:
/hold

until oc tests are fixed in the linked PR.

tchap:
@ardaguclu 

> Why do we copy/paste under /internal directory and add a new middle layer libraries?

I think that my approach was to copy the `distributions` internal packages into `internal` here and then wrap then as needed. Most of our projects actually use `registryclient` so the change is done in such a way that the users can simply change the import path to `registryclient/v2` and not really care about the rest. Otherwise it all becomes a bit too overwhelming.

> I'm surprised seeing that there isn't any changes in go.mod, will there be followup PRs to complete the transition?

These changes are added as a standalone module, so there is `go.mod` in `repositoryclient/v2` directory. This must be the case, because we cannot migrate the whole library-go to `distributions/v3`, that would break everything.

ardaguclu:
Last question: There are repositories (e.g. oc) that directly import distribution/distribution/v3/registry/client. What is the expected import for them?.

PR #1976, opened 2025-07-10
[release-4.18] OCPBUGS-58412: Fix revision readiness annotation for upgrade compatibility

dgrisonnet:
### Problem

When upgrading OpenShift to a version that includes revision readiness checks, the revision controller could become blocked if existing revision-status configmaps lack the `operator.openshift.io/revision-ready` annotation. The `getLatestAvailableRevision()` function only considers revisions with `revision-ready` set to `"true"` as valid, meaning older revision-status configmaps created before this feature would be ignored.

This could cause issues during version upgrades where:
1. An existing revision-status configmap exists (e.g., `revision-status-1`) without the readiness annotation
2. The new controller version checks for readiness but finds no "ready" revisions
3. The controller cannot properly determine the latest available revision

### Solution

This PR ensures backward compatibility by automatically setting the `operator.openshift.io/revision-ready` annotation to `"true"` on the current latest available revision's configmap during the sync process.

**Changes made:**

1. **Added revision readiness backfill logic** in `sync()` method:
   - When a `LatestAvailableRevision` exists, check if its corresponding configmap has the readiness annotation
   - If the annotation is missing or not set to `"true"`, update it accordingly
   - This prevents the controller from being blocked when upgrading to versions that require revision readiness

2. **Updated tests** to reflect the new behavior:
   - Added proper `revision-ready` annotations to test fixtures
   - Added a new test case `latest-revision-ready-annotation-set` to verify the annotation is correctly applied
   - Updated existing test objects to include the expected annotation

### Testing

- All existing tests pass with the new behavior
- New test case validates that the readiness annotation is properly set during sync
- Manual testing confirmed that upgrades from versions without revision readiness work correctly

This change is backward compatible and prevents upgrade blocking scenarios while maintaining the existing revision management functionality.

---

**Related:** OCPBUGS-58412

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -366,6 +366,28 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
 		return nil
 	}
 
+	// If there is a revision available, make sure that revision-ready is set to true on the
+	// revision-status{status.latestAvailableRevision} configmap. This prevents
+	// blocking the controller when upgrading to a version that checks revision
+	// readiness.
+	if operatorStatus.LatestAvailableRevision != 0 {
+		configMap, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(ctx, nameFor("revision-status", operatorStatus.LatestAvailableRevision), metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if configMap.Annotations == nil {
+			configMap.Annotations = make(map[string]string)
+		}
+
+		if configMap.Annotations["operator.openshift.io/revision-ready"] != "true" {
+			configMap.Annotations["operator.openshift.io/revision-ready"] = "true"
+			if _, err := c.configMapGetter.ConfigMaps(configMap.Namespace).Update(ctx, configMap, metav1.UpdateOptions{}); err != nil {
+				return err
+			}
+		}
+	}
+
 	// If the operator status's latest available revision is not the same as the observed latest revision, update the operator.
 	// This needs to be done even if the revision precondition is not required because it ensures our operator status is
 	// correct for all consumers.
--- a/pkg/operator/revisioncontroller/revision_controller_test.go
+++ b/pkg/operator/revisioncontroller/revision_controller_test.go
@@ -107,8 +107,11 @@ func TestRevisionController(t *testing.T) {
 				&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "test-config", Namespace: targetNamespace}},
 				&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "revision-status", Namespace: targetNamespace}},
 				&v1.ConfigMap{
-					ObjectMeta: metav1.ObjectMeta{Name: "revision-status-1", Namespace: targetNamespace},
-					Data:       map[string]string{"revision": "1"},
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "revision-status-1", Namespace: targetNamespace,
+						Annotations: map[string]string{"operator.openshift.io/revision-ready": "true"},
+					},
+					Data: map[string]string{"revision": "1"},
 				},
 			},
 			validateActions: func(t *testing.T, actions []clienttesting.Action, kclient *fake.Clientset) {
@@ -218,6 +221,16 @@ func TestRevisionController(t *testing.T) {
 				nil,
 				nil,
 			),
+			startingObjects: []runtime.Object{
+				&v1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "revision-status", Namespace: targetNamespace}},
+				&v1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "revision-status-1", Namespace: targetNamespace,
+						Annotations: map[string]string{"operator.openshift.io/revision-ready": "true"},
+					},
+					Data: map[string]string{"revision": "1"},
+				},
+			},
 			testConfigs:     []RevisionResource{{Name: "test-config"}},
 			testSecrets:     []RevisionResource{{Name: "test-secret"}},
 			expectSyncError: `configmaps "test-config" not found`,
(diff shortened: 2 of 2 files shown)


pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -356,6 +356,39 @@ func (c RevisionController) getLatestAvailableRevision(ctx context.Context) (int
...
+//  2. The revision-status was deleted.
+func (c RevisionController) fixOperatorLatestAvailableRevision(ctx context.Context, recorder events.Recorder, revision int32) error {
+	if revision == 0 {
+		return nil
+	}
+
+	configMap, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(ctx, fmt.Sprintf("revision-status-%d", revision), metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// Re-create the configmap and complete the revision.
+		_, err := c.createNewRevision(ctx, recorder, revision, "revision-status configmap missing")
```
dgrisonnet: This check was free so I added it if someone ever deletes the configmap inadvertently

benluddy: Is it safe?

Imagine the source of one of the revisioned resources changes, then the revision copy and revision status are deleted. The controller would re-create the revisioned resource, and now the content of revision-N depends on the time it is installed -- before or after the deletions.

I wouldn't include it in this patch unless it's necessary... but it sounds like a case we should consider for 4.20+ changes.

benluddy: My concern is basically that any revision >= latestAvailableRevision might have already rolled out, so we can't mutate the revisioned resources. If required revisioned resources are missing, then we would have to roll forward with a new revision.

I think we have to limit this to checking that all of the required revisioned resources exist and, if yes, flipping ready to true. Combined with the "ready" backport, is that enough to fix the bug?

benluddy: > I think we have to limit this to checking that all of the required revisioned resources exist and, if yes, flipping ready to true. Combined with the "ready" backport, is that enough to fix the bug?

Can we fix the bug with a single 4.18 patch that flips revision-ready from false to true on "revision-status-{status.latestAvailableRevision}"?

dgrisonnet: > Can we fix the bug with a single 4.18 patch that flips revision-ready from false to true on "revision-status-{status.latestAvailableRevision}"?

Yes we can do that, I'll submit a patch with that change.
My logic for calling `createNewRevision` instead of just flipping the annotation was in case some of the resource didn't get created for whatever reason.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -366,6 +366,21 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+	// Make sure that revision-ready is set to true on the
+	// revision-status{status.latestAvailableRevision} configmap. This prevents
+	// blocking the controller when upgrading to a version that checks revision
+	// readiness.
+	configMap, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(ctx, nameFor("revision-status", operatorStatus.LatestAvailableRevision), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if configMap.Annotations["operator.openshift.io/revision-ready"] != "true" {
+		configMap.Annotations["operator.openshift.io/revision-ready"] = "true"
```
benluddy: Shouldn't happen because this controller should be the only writer and should be creating these with at least one annotation, but it would panic here if configMap.Annotations is somehow nil.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -366,6 +366,21 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
 		return nil
 	}
 
+	// Make sure that revision-ready is set to true on the
+	// revision-status{status.latestAvailableRevision} configmap. This prevents
+	// blocking the controller when upgrading to a version that checks revision
+	// readiness.
```
benluddy: I think this is correct. The other controllers in the revision system don't care about the revision-status configmaps, so if latestAvailableRevision has already been updated then the revision has already "shipped" and might be installed.

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -366,6 +366,28 @@ func (c RevisionController) sync(ctx context.Context, syncCtx factory.SyncContex
...
+	if operatorStatus.LatestAvailableRevision != 0 {
+		configMap, err := c.configMapGetter.ConfigMaps(c.targetNamespace).Get(ctx, nameFor("revision-status", operatorStatus.LatestAvailableRevision), metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if configMap.Annotations == nil {
+			configMap.Annotations = make(map[string]string)
```
benluddy, line 380: Is this mutating an object in the informer cache?

dgrisonnet: Potentially, looking at the code, there is a comment saying that the object should be treated as read-only.

I'll make a copy.

dgrisonnet: oh wait nevermind, we have helpers in library-go making a copy of the object before returning it, so it should be safe to modify it in place: https://github.com/openshift/library-go/blob/ab97ebb73f0f3930a2506f4cbba64c19e20b7c4a/pkg/operator/v1helpers/core_getters.go#L55

p0lyn0mial:
> When upgrading OpenShift to a version that includes revision readiness checks, the revision controller could become blocked if existing revision-status configmaps lack the operator.openshift.io/revision-ready annotation

Is this PR fixing an issue that occurs during an upgrade from 4.17 to 4.18? I’m asking because the revision readiness checks were introduced in [4.15](https://github.com/openshift/library-go/commit/6770d7166292c6c582f9b0dfb3102637950c3e90).

p0lyn0mial:
I think this solves the issue described [in](https://github.com/openshift/library-go/pull/1976#pullrequestreview-3378484995), unless I’m misunderstanding what you mean by the `revision readiness check` in the description :) 

This PR makes sense. Thanks for fixing the issue.

/approve

pkg/operator/revisioncontroller/revision_controller.go
p0lyn0mial: not obvious :(

PR #1977, opened 2025-07-15
Add workload deletion condition method and clean-up

liouk:
There might be cases (as demonstrated in [OCPBUGS-44937](https://issues.redhat.com/browse/OCPBUGS-44937)) where we might want to gracefully delete the operand workload of the workload controller, and keep the operator status available (instead of unavailable or degraded).

This PR adds an optional way of specifying a deletion condition which will trigger the deletion of the operand gracefully, keeping the operator's status as Available=True.

This is needed in the scope of https://github.com/openshift/cluster-authentication-operator/pull/740.

This PR replaces #1902.


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -127,6 +134,21 @@ func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncCon
...
 	}
 
+	if deleted, operandName, err := c.delegate.WorkloadDeleted(ctx); err != nil {
+		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, []error{err})
+	} else if deleted {
+		// Server-Side-Apply with an empty operator status for the specific field manager will effectively
+		// remove any conditions and generations owned by it, because the respective API fields have 'map'
+		// as the list type where field managers can be list element-specific
+		if err := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, applyoperatorv1.OperatorStatus()); err != nil {
+			return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, []error{err})
```
p0lyn0mial: I would just simply return an err here, this is also what we do in updateOperatorStatus

pkg/operator/status/status_controller.go
```diff
@@ -130,6 +131,14 @@ func (c *StatusSyncer) WithVersionRemoval() *StatusSyncer {
 	return &output
 }
 
+// WithEmptyVersionRemoval returns a copy of the StatusSyncer that will
+// remove versions that are an empty string in VersionGetter from the status.
+func (c *StatusSyncer) WithEmptyVersionRemoval() *StatusSyncer {
```
p0lyn0mial: could we use `WithVersionRemoval` instead ?

liouk: There's already a `WithVersionRemoval()` method that does something slightly different: it removes versions that are missing in VersionGetter from the status. Also, there's no way to delete a version from the version getter, as it only allows to Get/Set one, hence the empty string need.

p0lyn0mial: mhm, maybe we should allow for version removal then ? 

I think that would involve exposing a new method on https://github.com/openshift/cluster-openshift-apiserver-operator/blob/476eac3b927b2a935f01f8ac4ec74f9a0cf54bc4/vendor/github.com/openshift/library-go/pkg/operator/status/version.go#L33 


If we had such method then we could pair it with WithVersionRemoval, right ?

pkg/operator/status/version.go
```diff
@@ -35,14 +35,15 @@ func (v *versionGetter) SetVersion(operandName, version string) {
...
-		// don't let a slow consumer block the rest
-		go func() {
-			ch <- struct{}{}
-		}()
-	}
+func (v *versionGetter) UnsetVersion(operandName string) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+
+	delete(v.versions, operandName)
```
p0lyn0mial, line 49: should we check if the version wasn't removed ?

pkg/operator/status/version.go
```diff
@@ -65,6 +66,16 @@ func (v *versionGetter) VersionChangedChannel() <-chan struct{} {
 	return channel
 }
 
+func (v *versionGetter) notifyChannels() {
```
p0lyn0mial: please change the name of the method to `notifyChannelsLocked` so that it is clear it must be called under lock.

pkg/operator/status/version.go
p0lyn0mial: and if it was maybe we don't have to notify (?)

liouk: `delete()` is a no-op if the map is nil/empty or the key doesn't exist, so this is a safe call.

pkg/operator/status/version.go
liouk: Great point

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -127,6 +134,20 @@ func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncCon
 		return err
 	}
 
+	if deleted, operandName, err := c.delegate.WorkloadDeleted(ctx); err != nil {
```
p0lyn0mial, line 143: I think we need to move it down. In theory it could be possible to define preconditions for the new method as well.

So I think the order should be: Precondiditons, WorkladDeleted, Sync

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -127,6 +134,20 @@ func (c *Controller) sync(ctx context.Context, controllerContext factory.SyncCon
...
+		return c.updateOperatorStatus(ctx, operatorStatus, nil, false, false, []error{err})
+	} else if deleted {
+		// Server-Side-Apply with an empty operator status for the specific field manager will effectively
+		// remove any conditions and generations owned by it, because the respective API fields have 'map'
+		// as the list type where field managers can be list element-specific
+		if err := c.operatorClient.ApplyOperatorStatus(ctx, c.controllerInstanceName, applyoperatorv1.OperatorStatus()); err != nil {
+			return err
+		}
+
+		c.versionRecorder.UnsetVersion(operandName)
```
p0lyn0mial: The corresponding Set method can be prefixed with `c.operandNamePrefix`. I think we need to do the same, no?

pkg/operator/status/status_controller.go
```diff
@@ -29,6 +29,8 @@ import (
 type VersionGetter interface {
 	// SetVersion is a way to set the version for an operand.  It must be thread-safe
 	SetVersion(operandName, version string)
+	// UnsetVersion removes a version for an operand.  It must be thread-safe
```
p0lyn0mial: please expand the comment by explaining what happens when the given operandName was already deleted.

pkg/operator/status/version.go
p0lyn0mial: but we will call notifyChannelsLocked event if nothing was removed.
let's only call notifyChannelsLocked when we removed an entry from the map.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -354,6 +371,14 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	return nil
 }
 
+func (c *Controller) addPrefix(name string) string {
```
p0lyn0mial: nit: please rename to `constructOperandNameFor(name string) string` or `constructOperandName(name string) string`

pkg/operator/status/version.go
p0lyn0mial: ping

p0lyn0mial (commented):
just to more comments, other than that LGTM

PR #1978, opened 2025-07-17
CNTRLPLANE-894: Add JobController implementation and related resources

slintes:
This PR introduces a new JobController that manages Kubernetes jobs, including functionality for job creation, deletion, and status reporting. It also includes the necessary resource application and merging logic for jobs, along with test cases to validate the controller's behavior.

The controller was implemented and is used in CEO for setting up two node fencing: https://github.com/openshift/cluster-etcd-operator/tree/main/pkg/tnf/pkg/jobs

### DIFF
--- a/pkg/operator/jobcontroller/job_controller.go
+++ b/pkg/operator/jobcontroller/job_controller.go
@@ -1,0 +1,356 @@
+package jobcontroller
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	opv1 "github.com/openshift/api/operator/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/management"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	batchinformersv1 "k8s.io/client-go/informers/batch/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+)
+
+// JobHookFunc is a hook function to modify the Job.
+type JobHookFunc func(*opv1.OperatorSpec, *batchv1.Job) error
+
+// JobController is a generic controller that manages a job.
+//
+// This controller optionally produces the following conditions:
+// <name>Available: indicates that the Job was successfully deployed and completed.
+// <name>Progressing: indicates that the Job is active.
+// <name>Degraded: produced when the Job failed.
+type JobController struct {
+	// instanceName is the name to identify what instance this belongs too: FooDriver for instance
+	instanceName string
+	// controllerInstanceName is the name to identify this instance of this particular control loop: FooDriver-CSIDriverNodeService for instance.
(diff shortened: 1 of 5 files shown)


pkg/operator/jobcontroller/job_controller.go
```diff
@@ -0,0 +1,356 @@
...
+		status,
+	)
+	if err != nil {
+		return err
+	}
+
+	// return an error for reporting degraded status!
+	// setting a condition manually, similar to available and progressing, doesn't work
+	if isFailed(job) {
+		return fmt.Errorf("job failed")
```
qJkee, line 304: If you return error there, controller will try requeue it

jsafrane:
> The controller was implemented and is used in CEO for setting up two node fencing: https://github.com/openshift/cluster-etcd-operator/tree/main/pkg/tnf/pkg/jobs

What is the reason to move it into library-go? From library-go README.md:
> Anything introduced here must have concrete use-cases in at least two separate openshift repos and be of some reasonable complexity. The bar here is high.

jsafrane:
> What do you think about this addition

What would help:

* A POC PR in a consumer repo (cluster-etcd-operator?) vendoring this via hacked `go.mod` and passing CI, esp. a two-node CI job.
* LGTM from someone with some domain knowledge (probably an etcd-operator person again).
* Having OWNERS file for the new controller.
* Most importantly, a comment what other repo will vendor this. And a POC PR + CI would help there too, if it makes sense at this stage. If it's going to be used only by cluster-etcd-operator, then keep the controller there.

pkg/operator/resource/resourceapply/batch.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcehelper"
+	"github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
+	batchv1 "k8s.io/api/batch/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	batchclientv1 "k8s.io/client-go/kubernetes/typed/batch/v1"
+)
+
+// ApplyJob ensures the form of the specified job is present in the API. If it
```
clobrano, line 16: I would like to stress a bit more that this `ApplyJob` might be destructive operation if the Job needs updates. Maybe a WARNING in the function header, and also a mention to the error returned when a job is deleted due to a modification.

pkg/operator/jobcontroller/job_controller.go
```diff
@@ -0,0 +1,356 @@
...
+		WithGenerations(&applyoperatorv1.GenerationStatusApplyConfiguration{
+			Group:          ptr.To("batch"),
+			Resource:       ptr.To("jobs"),
+			Namespace:      ptr.To(job.Namespace),
+			Name:           ptr.To(job.Name),
+			LastGeneration: ptr.To(job.Generation),
+		})
+
+	// Set Available condition
+	if slices.Contains(c.conditions, opv1.OperatorStatusTypeAvailable) {
```
clobrano, line 243: The logic for setting Available and Progressing condition seems quite the same. Is there any reason preventing to make a helper to reduce the duplication?

jaypoulz:
> Anything introduced here must have concrete use-cases in at least two separate OpenShift repos and be of some reasonable complexity. The bar here is high.

@jsafrane - So I understand that we want to prevent library-go from becoming a collection of single-use patterns that we think would be useful but are only used once.

I guess my question is, how do we balance that against discoverability? If this is a useful pattern (and we suspect it is) AND we have evidence that other repos are implementing the same kind of logic in their workflows (e.g. [insights](https://github.com/openshift/insights-operator/blob/7f8b40d9ccab7d83f183c5942044892c263d9c14/pkg/controller/periodic/job.go#L24), [local-static-provisioner](https://github.com/openshift/sig-storage-local-static-provisioner/blob/2529b85a745705a1125a16198ae133efb234c20f/pkg/deleter/jobcontroller.go#L68), [external-storage](https://github.com/openshift/external-storage/blob/34efc4c2dc0f7032db94b760a9c563ef914cd285/local-volume/provisioner/pkg/deleter/jobcontroller.go#L68), etc.), how do we advertise to teams that this is useful? From my perspective, our team isn't going to go and rewrite other team's code. And if I were a team that already wrote my own job controller, I probably wouldn't refactor to use lib-go job controller if mine were already working. The most likely candidates for people adopting this are people writing *new* features or repos that also need something like this. If that's the case, the first place they'll look is upstream, and then here.

So how do you recommend we proceed? Should we be holding off on contributing shared code until we ourselves run into the need to have it in multiple repos? How have other teams contributed features like this before?

PR #1979, opened 2025-07-18
OCPBUGS-59526: certgraphanalysis: remove hash from primary cert bundle secret

vrutkovs:
ROSA Hypershift tests create this secret with a certificate, but its name always starts with a random hash. This option would allow us to replace the variable part so that this certificate could be tracked in TLS registry.

See flaking test ` all registered tls artifacts must have no metadata violation regressions` in [this](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.20-e2e-rosa-sts-hypershift-ovn/1943821536276254720) ROSA test:
```
requirement/ownership: --namespace=openshift-ingress secret/2jvlkpa3d9m62op6t1789pd503rsl9g8-primary-cert-bundle-secret regressed and does not have value set
```

Tested in https://github.com/openshift/origin/pull/29995

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
@@ -149,6 +149,15 @@ var (
 			return timestampReg.ReplaceAllString(path, "<timestamp>.pem")
 		},
 	}
+	RewritePrimaryCertBundleSecret = &metadataOptions{
+		rewriteSecretFn: func(secret *corev1.Secret) {
+			if secret.Namespace != "openshift-ingress" || !strings.HasSuffix(secret.Name, "-primary-cert-bundle-secret") {
+				return
+			}
+			hash := strings.TrimSuffix(secret.Name, "-primary-cert-bundle-secret")
+			secret.Name = strings.ReplaceAll(secret.Name, hash, "<hash>")
+		},
+	}
 )
 
 // skipRevisionedInOnDiskLocation returns true if location is for revisioned certificate and needs to be skipped


pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -149,6 +149,15 @@ var (
 			return timestampReg.ReplaceAllString(path, "<timestamp>.pem")
 		},
 	}
+	RewritePrimaryCertBundleSecret = &metadataOptions{
+		rewriteSecretFn: func(secret *corev1.Secret) {
+			if secret.Namespace != "openshift-ingress" || !strings.HasSuffix(secret.Name, "-primary-cert-bundle-secret") {
+				return
+			}
+			hash := strings.TrimSuffix(secret.Name, "-primary-cert-bundle-secret")
+			secret.Name = strings.ReplaceAll(secret.Name, hash, "<hash>")
```
p0lyn0mial, line 158: this ends up: `<hash>-primary-cert-bundle-secret` how does it help the TLS registry to track this secret?

p0lyn0mial (commented):
lgtm, just one question.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
p0lyn0mial: ah, we expect a stable name, don't we ?

vrutkovs: Yes, a stable name across different runs is a requirement (similar to etcd serving certs we store as `etcd-<master-1>-serving` and so on)

p0lyn0mial: ok, so, since the name is not stable why the test is flaky and not failing permanently?

vrutkovs: The test is intentionally made flaky until 99% of certs are included in TLS registry as known violations (otherwise it would block a lot of merges and we're not ready for that yet)

p0lyn0mial: ok, sgtm.

PR #1980, opened 2025-07-23
operator/certrotation/signer: decouple setting annotation from setSigningCertKeyPairSecret into a separate fn

p0lyn0mial:


pkg/operator/certrotation/signer.go
```diff
@@ -221,11 +225,28 @@ func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validi
...
+		signingCertKeyPairSecret.Annotations = make(map[string]string)
+	}
+
 	annotations.NotBefore = ca.Certs[0].NotBefore.Format(time.RFC3339)
 	annotations.NotAfter = ca.Certs[0].NotAfter.Format(time.RFC3339)
-	signingCertKeyPairSecret.Annotations[CertificateIssuer] = ca.Certs[0].Issuer.CommonName
-
 	_ = annotations.EnsureTLSMetadataUpdate(&signingCertKeyPairSecret.ObjectMeta)
 
+	signingCertKeyPairSecret.Annotations[CertificateIssuer] = ca.Certs[0].Issuer.CommonName
```
vrutkovs, line 247: Why are we setting this after `EnsureTLSMetadataUpdate`?

p0lyn0mial: I have checked and `CertificateIssuer` is not set by the `EnsureTLSMetadataUpdate`.

It looks like not all annotations are set by `EnsureTLSMetadataUpdate`.

But I can move it up by one line.

vrutkovs: ah, okay, I thought I;m missing some important functionality part here :)

p0lyn0mial: no, this is just a no-op refactoring that renames some functions.

PR #1981, opened 2025-07-24
OCPBUGS-57049: cert inspection: add an option to humanize refresh period annotation

vrutkovs:
This option converts `time.Parse Duration`-compatible string of duration (i.e. `77h30m00s`) into
human-readable string (`3d5h30m`) so that TLS registry markdowns have improved readability


pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -149,6 +152,35 @@ var (
 			return timestampReg.ReplaceAllString(path, "<timestamp>.pem")
 		},
 	}
+	HumanizeRewritePeriod = &metadataOptions{
```
p0lyn0mial: ok, and how is this var being used ? by external tools that pull this library ?

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -149,6 +152,35 @@ var (
...
+		rewriteSecretFn: func(secret *corev1.Secret) {
+			if secret.Annotations == nil {
+				return
+			}
+			if period, ok := secret.Annotations[certrotation.CertificateRefreshPeriodAnnotation]; ok {
+				if d, err := time.ParseDuration(period); err == nil {
```
p0lyn0mial: nit: i would change to: (reads a bit easier)
```
d, err := time.ParseDuration(period)
if err != nil {
  fmt.Fprintf(os.Stderr, "Failed to parse rewrite period %s: %v\n", period, err)
  return
}
humanReadableDate := durationToHumanReadableString(d)
secret.Annotations[certrotation.CertificateRefreshPeriodAnnotation] = humanReadableDate
secret.Annotations[rewritePrefix+"HumanizeRewritePeriod"] = period
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -149,6 +152,35 @@ var (
...
+					fmt.Fprintf(os.Stderr, "Failed to parse rewrite period %s: %v\n", period, err)
+				}
+			}
+		},
+		rewriteConfigMapFn: func(configMap *corev1.ConfigMap) {
+			if configMap.Annotations == nil {
+				return
+			}
+			if period, ok := configMap.Annotations[certrotation.CertificateRefreshPeriodAnnotation]; ok {
+				if d, err := time.ParseDuration(period); err == nil {
```
p0lyn0mial: same here

p0lyn0mial (commented):
lgtm

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -149,6 +152,35 @@ var (
...
+	HumanizeRewritePeriod = &metadataOptions{
+		rewriteSecretFn: func(secret *corev1.Secret) {
+			if secret.Annotations == nil {
+				return
+			}
+			if period, ok := secret.Annotations[certrotation.CertificateRefreshPeriodAnnotation]; ok {
+				if d, err := time.ParseDuration(period); err == nil {
+					humanReadableDate := durationToHumanReadableString(d)
+					secret.Annotations[certrotation.CertificateRefreshPeriodAnnotation] = humanReadableDate
+					secret.Annotations[rewritePrefix+"HumanizeRewritePeriod"] = period
```
p0lyn0mial: shouldn't this be `rewritePrefix`+`certrotation.CertificateRefreshPeriodAnnotation` ?

vrutkovs: This annotation stores original value we may need for debugging. It should include the name of the function which did the rewrite (it's expected that the function would rewrite just one annotation, so it's more beneficial to save the function name instead of the annotation name, so that the source of the error could be found easier)

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: [origin](https://github.com/openshift/origin/pull/30016) would run this function to update secrets/configmap annotation when raw TLS data is collected

p0lyn0mial: nit: how about changing the name to `RewriteRefreshPeriod` or `ReformatRefreshPeriod` or `TransformRefreshPeriod` ?

vrutkovs: Yeah, lets go with `Rewrite` - its more similar to what other functions are named

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -158,8 +161,39 @@ var (
...
+	for k, v := range annotations {
+		result[k] = v
+	}
+	period, ok := annotations[certrotation.CertificateRefreshPeriodAnnotation]
+	if !ok {
+		return result
+	}
+	d, err := time.ParseDuration(period)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to parse rewrite period %s: %v\n", period, err)
```
wangke19:
```suggestion
		fmt.Fprintf(os.Stderr, "Failed to parse certificate refresh period %q: %v\n", period, err)
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -158,8 +161,39 @@ var (
...
+		return result
+	}
+	humanReadableDate := durationToHumanReadableString(d)
+	result[certrotation.CertificateRefreshPeriodAnnotation] = humanReadableDate
+	result[rewritePrefix+"HumanizeRewritePeriod"] = period
```
wangke19: The key name `HumanizeRewritePeriod` is misleading —it is saving the raw string here, not the humanized one.
```suggestion
        result[rewritePrefix+"HumanizedRefreshPeriod"] = humanReadableDate
	result[rewritePrefix+"RawRefreshPeriod"] = period
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -235,3 +269,74 @@ func StripRootFSMountPoint(rootfsMount string) *metadataOptions {
...
+// Unlike Go's built-in `time.Duration.String()`, which returns a string like "72h0m0s", this function returns a more concise format like "3d" or "5d4h25m".
+func durationToHumanReadableString(d time.Duration) string {
+	u := uint64(d)
+	if d < 0 {
+		u = -u
+	}
+
+	if u == 0 {
+		return "0s"
+	}
```
wangke19:
```suggestion
	if d == 0 {
		return "0s"
	}

	// Handle negative durations by taking absolute value
	u := d
	if d < 0 {
		u = -d
	}
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -235,3 +269,74 @@ func StripRootFSMountPoint(rootfsMount string) *metadataOptions {
...
+
+	// Unit values in uint64 nanoseconds
+	year := uint64(time.Hour) * 24 * 365
+	month := uint64(time.Hour) * 24 * 30
+	day := uint64(time.Hour) * 24
+	hour := uint64(time.Hour)
+	minute := uint64(time.Minute)
+	second := uint64(time.Second)
```
wangke19:
```suggestion
	totalSeconds := uint64(u.Seconds())

	// Define time units in seconds
	const (
		year   = 365 * 24 * 3600
		month  = 30 * 24 * 3600
		day    = 24 * 3600
		hour   = 3600
		minute = 60
		second = 1
	)
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -235,3 +269,74 @@ func StripRootFSMountPoint(rootfsMount string) *metadataOptions {
...
+	seconds := u / second
+	u %= second
+	if seconds > 0 {
+		b.WriteString(strconv.FormatUint(seconds, 10))
+		b.WriteString("s")
+	}
+
+	if b.Len() == 0 {
+		return "0s"
+	}
```
wangke19:
```suggestion
	years := u / year
		writeUnit := func(value uint64, suffix string) {
		if value > 0 {
			b.WriteString(strconv.FormatUint(value, 10))
			b.WriteString(suffix)
		}
	}

	years := totalSeconds / year
	totalSeconds %= year
	writeUnit(years, "y")

	months := totalSeconds / month
	totalSeconds %= month
	writeUnit(months, "mo")

	days := totalSeconds / day
	totalSeconds %= day
	writeUnit(days, "d")

	hours := totalSeconds / hour
	totalSeconds %= hour
	writeUnit(hours, "h")

	minutes := totalSeconds / minute
	totalSeconds %= minute
	writeUnit(minutes, "m")

	seconds := totalSeconds
	writeUnit(seconds, "s")
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_test.go
```diff
@@ -0,0 +1,30 @@
...
+		expected string
+	}{
+		{time.Hour * 24 * 365 * 9, "9y"},
+		{time.Hour * 24 * 365, "1y"},
+		{time.Hour * 24 * 30 * 5, "5mo"},
+		{time.Hour * 24 * 30, "1mo"},
+		{time.Hour * 24 * 2, "2d"},
+		{time.Hour * 2, "2h"},
+		{time.Hour*24*3 + time.Hour*2, "3d2h"},
+		{time.Hour*24*5 + time.Hour*4 + time.Minute*25, "5d4h25m"},
```
wangke19: covers various cases including zero, positive, negative, and composite durations:
```suggestion
		{0, "0s"},
		{time.Second, "1s"},
		{2 * time.Second, "2s"},
		{time.Minute, "1m"},
		{time.Minute + 30*time.Second, "1m30s"},
		{time.Hour, "1h"},
		{25 * time.Hour, "1d1h"},
		{30 * 24 * time.Hour, "1mo"},
		{365 * 24 * time.Hour, "1y"},
		{400 * 24 * time.Hour, "1y1mo5d"},
		{-time.Minute, "1m"},                // negative duration
		{-400 * 24 * time.Hour, "1y1mo5d"}, // negative composite
		{3*time.Minute + 4*time.Second, "3m4s"},
```

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
p0lyn0mial: I think that this should be `RewriteRefreshPeriod` now.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -158,8 +161,39 @@ var (
...
+		rewriteSecretFn: func(secret *corev1.Secret) {
+			secret.Annotations = humanizeRefreshPeriodFromMetadata(secret.Annotations)
+		},
+		rewriteConfigMapFn: func(configMap *corev1.ConfigMap) {
+			configMap.Annotations = humanizeRefreshPeriodFromMetadata(configMap.Annotations)
+		},
+	}
 )
 
+func humanizeRefreshPeriodFromMetadata(annotations map[string]string) map[string]string {
```
p0lyn0mial: in go maps can be changed in-place, no need to return a new map.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
p0lyn0mial: Failed to parse refresh period

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -235,3 +269,74 @@ func StripRootFSMountPoint(rootfsMount string) *metadataOptions {
 		},
 	}
 }
+
+// durationToHumanReadableString formats a duration into a human-readable string.
+// Unlike Go's built-in `time.Duration.String()`, which returns a string like "72h0m0s", this function returns a more concise format like "3d" or "5d4h25m".
+func durationToHumanReadableString(d time.Duration) string {
```
p0lyn0mial, line 271: could we include a link to to original implementation and state that we slightly changed the code ?

p0lyn0mial (commented):
a few more comments/questions

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: I like the idea of checking for `d == 0` first, but we also still want to do `u := uint64(d)` first to make sure its converted from the start

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: Fixed

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: Added a comment

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: Actually, we don't need to store values as `uint64` - it's best to be rounded to seconds

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -4,15 +4,27 @@ import (
...
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
 	corev1 "k8s.io/api/core/v1"
 )
 
-const rewritePrefix = "rewritten.cert-info.openshift.io/"
+const (
+	rewritePrefix = "rewritten.cert-info.openshift.io/"
+	// Unit values in seconds
+	year   = 60 * 60 * 24 * 365
```
p0lyn0mial: please move this vars inside `durationToHumanReadableString` so that they are private to that method.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
```diff
@@ -158,8 +170,35 @@ var (
...
+		},
+		rewriteConfigMapFn: func(configMap *corev1.ConfigMap) {
+			humanizeRefreshPeriodFromMetadata(configMap.Annotations)
+		},
+	}
 )
 
+func humanizeRefreshPeriodFromMetadata(annotations map[string]string) {
+	if annotations == nil {
+		annotations = map[string]string{}
```
p0lyn0mial: let's return here, in go a `nil` map is equivalent to an empty map from which reading is possible.

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: Wouldn't we reinitialize them every time we call a function?

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
vrutkovs: ah, in that case we don't need this check at all - `annotations[certrotation.CertificateRefreshPeriodAnnotation]` will guard us. Added a unit test

pkg/certs/cert-inspection/certgraphanalysis/metadata_options.go
p0lyn0mial: good point, we would, but these var will only be used by that function and it would be nice to keep them inside that function for clarity. i don't think this code must be lightning fast. an alternative would be to create an internal pkg for that function.

vrutkovs: ah,  okay, sure

wangke19:
> @wangke19 it looks like we are about to merge this PR. Is there anything else we should take into consideration ?
lgtm

PR #1982, opened 2025-07-24
operator/certrotation/target: decouple setting annotation from setTargetCertKeyPairSecret into a seperate fn

p0lyn0mial:
this is just a no-op refactoring that renames some functions.

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -121,7 +121,7 @@ func (c RotatedSelfSignedCertKeySecret) EnsureTargetCertKeyPair(ctx context.Cont
 
 	if reason := c.CertCreator.NeedNewTargetCertKeyPair(targetCertKeyPairSecret, signingCertKeyPair, caBundleCerts, c.Refresh, c.RefreshOnlyWhenExpired, creationRequired); len(reason) > 0 {
 		c.EventRecorder.Eventf("TargetUpdateRequired", "%q in %q requires a new target cert/key pair: %v", c.Name, c.Namespace, reason)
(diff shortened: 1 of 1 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -251,21 +263,28 @@ func setTargetCertKeyPairSecret(targetCertKeyPairSecret *corev1.Secret, validity
...
 	targetCertKeyPairSecret.Annotations[CertificateIssuer] = certKeyPair.Certs[0].Issuer.CommonName
 
-	_ = annotations.EnsureTLSMetadataUpdate(&targetCertKeyPairSecret.ObjectMeta)
-	certCreator.SetAnnotations(certKeyPair, targetCertKeyPairSecret.Annotations)
+	tlsAnnotations.NotBefore = certKeyPair.Certs[0].NotBefore.Format(time.RFC3339)
+	tlsAnnotations.NotAfter = certKeyPair.Certs[0].NotAfter.Format(time.RFC3339)
+	_ = tlsAnnotations.EnsureTLSMetadataUpdate(&targetCertKeyPairSecret.ObjectMeta)
 
-	return nil
+	certCreator.SetAnnotations(certKeyPair, targetCertKeyPairSecret.Annotations)
```
p0lyn0mial, line 287: @vrutkovs this is preexisting, i assume it is needed, right ?

vrutkovs: Yes, this is required for target certs (they should have an annotation w/ hostnames and such).

I wonder if we can run this after `setTargetCertKeyPairSecret` call so that `setTLSAnnotationsOnTargetCertKeyPairSecret` could be reused for signers too (I'm not insisting though)?

p0lyn0mial: I would keep a separate function, it is easier to read. in the future we might consider further refactoring.

vrutkovs: okay, sounds good

PR #1983, opened 2025-07-25
NO-JIRA: controllercmd: Allow setting kube config overrides

tchap:
Right now `ControllerBuilder` is always initialized with nil kube config overrides when starting a controller. This patch adds `ControllerCommandConfig.WithKubeConfigOverrides`, which sets the overrides object then passed into `ControllerBuilder`.


pkg/controller/controllercmd/cmd.go
```diff
@@ -114,6 +116,12 @@ func (c *ControllerCommandConfig) WithEventRecorderOptions(eventRecorderOptions
 	return c
 }
 
+// WithKubeConfigOverrides sets overrides that are applied to the loaded the kube config before the controller is started.
```
ingvagabund, line 119: `the loaded the` -> `the loaded`

PR #1985, opened 2025-07-29
OCPBUGS-59626: staticpodcontroller: require a separate informer for cluster-level events

vrutkovs:
An informer for an empty namespace will list resources from all namespaces and cluster level. This may be undesirable, as this make the controller react to events from different namespaces, while we're only interested in cluster-level events. Splitting this into a separate set of informers will prevent that.

Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1874/commits


vrutkovs:
>add a condition in KubeInformersForNamespaces that would panic when an empty namespace is specified to catch bad actors

I found a few places where empty namespaces are set and probably we want at least a warning (in some cases it might be perfectly legal if we don't watch for secrets/configmaps).

PR #1988, opened 2025-08-06
certrotationcontroller: ignore events from other configmaps/secrets

vrutkovs:
No need to call sync whenever any secret/configmap updates - we should call sync either on timer or when signer/target secret or CA configmap are updated.

Tested as https://github.com/openshift/cluster-kube-apiserver-operator/pull/1890


sanchezl (approved):
/lgtm

PR #1993, opened 2025-08-13
WIP: Fix panic in required labels

gnufied:
fixes https://issues.redhat.com/browse/OCPBUGS-60464#

needs tests


gnufied:
cc @dobsonj

pkg/operator/resource/resourceapply/storage.go
```diff
@@ -161,14 +161,20 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
...
 	for _, exemptLabel := range exemptCSIDriverLabels {
 		if existingValue, ok := existing.Labels[exemptLabel]; ok {
-			required.Labels[exemptLabel] = existingValue
+			if requiredLabels == nil {
+				requiredLabels = map[string]string{}
+			}
+			requiredLabels[exemptLabel] = existingValue
 		}
 	}
+	required.Labels = requiredLabels
```
dobsonj: @gnufied instead of doing this within the loop, can you add a simple 3-line change for `required.Labels` where we already do the same thing for `required.Annotations`?
https://github.com/openshift/library-go/blob/378de074fe7b73a4bc593ff807693c466889d3d5/pkg/operator/resource/resourceapply/storage.go#L142-L144

dobsonj: The call to `validateRequiredCSIDriverLabels(required)` on line 148 also accesses `required.Labels`, we should guarantee that it's non-nil before that too.

gnufied: This is fixed now.

PR #1994, opened 2025-08-18
OCPBUGS-57627: cert-inspection: create stubs for in-memory certificates

vrutkovs:
We have no method to detect those, we have to create stub entries for every instance of apiserver.

Tested in https://github.com/openshift/origin/pull/30005

### DIFF
--- a/pkg/certs/cert-inspection/certgraphanalysis/collector.go
+++ b/pkg/certs/cert-inspection/certgraphanalysis/collector.go
@@ -2,12 +2,14 @@ package certgraphanalysis
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/certs/cert-inspection/certgraphapi"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -299,3 +306,81 @@ func GetBootstrapIPAndHostname(ctx context.Context, kubeClient kubernetes.Interf
...
+	})
+
+	if err != nil {
+		return err
+	}
+
+	for i, pod := range podList.Items {
+		certKeyPair := certgraphapi.CertKeyPair{
+			Name:        fmt.Sprintf("%s-%d::1", detail.NamePrefix, i),
+			Description: "apiserver loopback certificate",
```
wangke19: The description is hardcoded as "apiserver loopback certificate". This limits the function's reusability. It would be better to add a Description field to the InMemoryCertDetail
     struct to allow callers to provide context-specific descriptions.

pkg/certs/cert-inspection/certgraphanalysis/collector.go
```diff
@@ -299,3 +306,81 @@ func GetBootstrapIPAndHostname(ctx context.Context, kubeClient kubernetes.Interf
...
+	return result, utilerrors.NewAggregate(errs)
+
+}
+
+func addInMemoryCertificateStub(ctx context.Context, list *certgraphapi.PKIList, kubeClient kubernetes.Interface, detail InMemoryCertDetail) error {
+	if list == nil {
+		list = &certgraphapi.PKIList{}
+	}
+
+	if list.InMemoryResourceData.CertKeyPairs == nil {
```
wangke19, line 339: We checking for nil on list.InMemoryResourceData.CertKeyPairs but then initializing list.CertKeyPairs.Items. This might likely be:
```
if list.InMemoryResourceData.CertKeyPairs == nil {
     list.InMemoryResourceData.CertKeyPairs = []certgraphapi.PKIRegistryInMemoryCertKeyPair{}
 }
 // and separately
 if list.CertKeyPairs.Items == nil {
      list.CertKeyPairs.Items = []certgraphapi.CertKeyPair{}
 }
```

vrutkovs: Great find

pkg/certs/cert-inspection/certgraphanalysis/collector.go
vrutkovs: Good idea

PR #2003, opened 2025-08-25
NO-JIRA: needNewTargetCertKeyPair: handle RefreshOnlyWhenExpired

vrutkovs:
Ensure that we exit early when no refresh is necessary in RefreshOnlyWhenExpired=true mode. This prevents from unnecessary check of CA bundle contents.

Tested in https://github.com/openshift/cluster-kube-apiserver-operator/pull/1905

### DIFF
--- a/pkg/operator/certrotation/target.go
+++ b/pkg/operator/certrotation/target.go
@@ -164,6 +164,11 @@ func needNewTargetCertKeyPair(secret *corev1.Secret, signer *crypto.CA, caBundle
 		return reason
 	}
 
+	// Exit early if we're only refreshing when expired and the certificate does not need an update
+	if refreshOnlyWhenExpired {
+		return ""
+	}
+
 	// check the signer common name against all the common names in our ca bundle so we don't refresh early
 	signerCommonName := annotations[CertificateIssuer]
 	if len(signerCommonName) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/operator/certrotation/target.go
```diff
@@ -164,6 +164,11 @@ func needNewTargetCertKeyPair(secret *corev1.Secret, signer *crypto.CA, caBundle
 		return reason
 	}
 
+	// Exit early if we're only refreshing when expired and the certificate does not need an update
+	if refreshOnlyWhenExpired {
+		return ""
```
wangke19, line 169: This will cause the function to always return "" (no refresh needed) when refreshOnlyWhenExpired is true, regardless of whether the cert is actually expired or not.
how about 
```
return reason // which is "" if not expired, it means no time-based refresh is needed.
```

vrutkovs: We check that cert didn't expire earlier, when running `needNewTargetCertKeyPairForTime`. We're avoiding an extra check whether the signer is included CA bundle. 

In refreshOnlyWhenExpired mode we can get a conflict and an empty signer would be passed here, failing this check. This issue will mostly be mitigated in https://github.com/openshift/library-go/pull/2004 (worker will exit sooner), but imo its useful to have another failsafe here too

PR #2006, opened 2025-08-29
OCPBUGS-54385: Add new helper to informers.go: NewKubeInformersForNamespacesWithResyncPeriod()

mpatlasov:
https://issues.redhat.com/browse/OCPBUGS-54385

New helper `NewKubeInformersForNamespacesWithResyncPeriod()` is useful for operators who want to opt out of automatic per-Informer resync. They will call this helper with zero `resyncInterval` arg.

### DIFF
--- a/pkg/operator/v1helpers/informers.go
+++ b/pkg/operator/v1helpers/informers.go
@@ -32,17 +32,21 @@ type KubeInformersForNamespaces interface {
 
 var _ KubeInformersForNamespaces = kubeInformersForNamespaces{}
 
-func NewKubeInformersForNamespaces(kubeClient kubernetes.Interface, namespaces ...string) KubeInformersForNamespaces {
+func NewKubeInformersForNamespacesWithResyncPeriod(kubeClient kubernetes.Interface, resyncInterval time.Duration, namespaces ...string) KubeInformersForNamespaces {
 	ret := kubeInformersForNamespaces{}
 	for _, namespace := range namespaces {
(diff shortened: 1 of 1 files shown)


pkg/controller/factory/factory.go
```diff
@@ -252,8 +252,19 @@ type informerHandleTuple struct {
...
+			ResyncPeriod: &resyncInterval[0],
+		})
+	} else {
+		informer.AddEventHandler(controller.syncContext.(syncContext).eventHandler(queueKeysFunc, filter))
+	}
+}
+
 // Controller produce a runnable controller.
-func (f *Factory) ToController(name string, eventRecorder events.Recorder) Controller {
+func (f *Factory) ToController(name string, eventRecorder events.Recorder, resyncInterval ...time.Duration) Controller {
```
jsafrane: I think this solution is on a wrong level. Someone has created an informer with 10 minute resync. It's not base controller factory's job to override that. The informer should be created with longer (or no) resync in the first place.

I think a new function `NewKubeInformersForNamespacesWithResyncPeriod` [here](https://github.com/openshift/library-go/blob/3ee684aaa12993b8893c43386129998d621a0e05/pkg/operator/v1helpers/informers.go#L35C1-L36C1) would be better

mpatlasov: OK, reworked to implement `NewKubeInformersForNamespacesWithResyncPeriod()`

PR #2009, opened 2025-09-09
WIP: certsyncpod: Swap secret/cm directories atomically

tchap:
Currently it can happen that cert-syncer replaces some of the secret/configmap files successfully and then fails. This can lead to problems when these are e.g. TLS cert/key files and the directory gets inconsistent. This may seem transient, but when cert-syncer is terminated in the middle, it can later fail to start as the whole kube-apiserver gets into a crash loop.

This introduces a new staticpod.SwapDirectoriesAtomic, which uses unix.Renameat2 with RENAME_EXCHANGE flag set.


p0lyn0mial:
@tchap is there a must-gather from an incident i could take a look at ?

tchap:
@p0lyn0mial I think that I consulted this one: https://access.redhat.com/support/cases/#/case/03849958/discussion?attachmentId=a096R00003JpGgMQAV

p0lyn0mial:
@vrutkovs do you have time to take a look at this issue ? 

I think that the issue might be real. I think the issue is when a two file cert is replaced. It can happen that the server picks up the update and notices the public/private key mismatch and crashes. Is there a way to repo this issue ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +237,80 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	srcDir, err := fs.MkdirTemp(filepath.Dir(contentDir), filepath.Base(contentDir)+"-*")
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to create source %s directory for %s/%s: %v", kind, objectNamespace, objectName, err)
+		errors = append(errors, err)
+		return errors
+	}
+	defer fs.RemoveAll(srcDir)
+
+	// Populate the tmp directory with files.
```
vrutkovs: This may cause TOCTOU (time of check, time of use) errors: scrDir at the time of preparation may be different from the time when its being atomically replaced - so some new files may be lost. Additionally, since we do a file copy instead of hardlinks, we may accidentally overwrite file contents too

vrutkovs:
Perhaps a better option would be syncing TLS certkeypair as a single item? Currently, all of keypairs are placed in a single directory, but it may not be the case later. These are easy to identify as all certkeypairs we have are stored in `kubernetes.io/tls` type secrets and named `tls.crt`/`tls.key`.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: Doesn't `WriteFileAtomic` that is being used currently have the same issue then?

tchap:
> Perhaps a better option would be syncing TLS certkeypair as a single item? Currently, all of keypairs are placed in a single directory, but it may not be the case later. These are easy to identify as all certkeypairs we have are stored in `kubernetes.io/tls` type secrets and named `tls.crt`/`tls.key`.

Just to be sure I understand, do you mean storing both in a single common file?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
vrutkovs: Yes, but at a very small scale - `writeFileAtomic` would be exposed only between `writeTemporaryFile` and `os.Rename`, while in case of a full directory copy this time may be increased.

Perhaps we could mitigate this by checking that src dir didn't change before making a rename?

tchap: I added the check in https://github.com/openshift/library-go/pull/2009/commits/9f285dbd570a6b3dc17692634ea29e509155b939

tchap: I am actually unsure what kind of check you had on your mind. Isn't reading the files in the check I implemented creating another window anyway?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +239,103 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	MkdirTemp:             os.MkdirTemp,
+	RemoveAll:             os.RemoveAll,
+	WriteFile:             os.WriteFile,
+	ReadFile:              os.ReadFile,
+	SwapDirectoriesAtomic: staticpod.SwapDirectoriesAtomic,
+}
+
+func writeFiles[C string | []byte](
+	fs *fileSystem, eventRecorder events.Recorder,
+	objectNamespace, objectName, kind string,
```
vrutkovs: Lets pass `metav1.TypeMeta` and `metav1.ObjectMeta` for type / location

pkg/operator/staticpod/certsyncpod/certsync_controller.go
vrutkovs: Check that
* directory contents didn't change (list of files is the same)
* file's `ModTime` and `Size` didn't change

This should guarantee that no files were updated before the rename

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +239,103 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
-		for filename, content := range secret.Data {
-			// TODO fix permissions
-			fullFilename := filepath.Join(contentDir, filename)
-			// if the existing is the same, do nothing
-			if reflect.DeepEqual(data[fullFilename], content) {
-				continue
-			}
+	}
+	if len(errors) > 0 {
+		return errors
```
vrutkovs: Remove tmp dir here

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +239,103 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+		if !bytes.Equal(content, []byte(files[filename])) {
+			eventRecorder.Warningf("CertificateUpdateFailed",
+				"Content mismatch detected for temporary %s file for %v/%v", kind, objectNamespace, objectName)
+			errors = append(errors, fmt.Errorf("checksum mismatch detected for %q", fullFilename))
+			continue
 		}
-		c.eventRecorder.Eventf("CertificateUpdated", "Wrote updated secret: %s/%s", secret.Namespace, secret.Name)
+	}
+	if len(errors) > 0 {
+		return errors
```
vrutkovs: Remove tmp dir too

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +239,103 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
 	}
 
-	return utilerrors.NewAggregate(errors)
+	// Swap directories atomically.
+	if err := fs.SwapDirectoriesAtomic(contentDir, srcDir); err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to enable new %s directory for %s/%s: %v", kind, objectNamespace, objectName, err)
+		errors = append(errors, err)
```
vrutkovs: Restore previous dir contents here

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,218 @@
...
+				"api.crt": []byte("rotated TLS cert"),
+				"api.key": []byte("rotated TLS key"),
+			},
+			checkState: checkOnError,
+		}
+	}
+
+	testCases := []testCase{
+		{
+			name:          "create target directory",
```
vrutkovs:
```suggestion
			name:          "empty target directory",
```

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,218 @@
...
+		errorTestCase("directory unchanged on failed to create temporary directory", func() *fileSystem {
+			fs := newRealFS()
+			fs.MkdirTemp = func(dir, pattern string) (string, error) {
+				return "", errors.New("nuked")
+			}
+			return fs
+		}),
+		errorTestCase("directory unchanged on failed to write file", func() *fileSystem {
+			fs := newRealFS()
+			fs.WriteFile = func(path string, data []byte, perm os.FileMode) error {
```
vrutkovs: This should be expanded to handle error on first and second file writes

pkg/operator/staticpod/file_utils_test.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+
+	testCases := []struct {
+		name        string
+		setup       func(t *testing.T, tmpDir string) (dirA, dirB string)
+		checkResult func(t *testing.T, dirA, dirB string, err error)
+	}{
+		{
+			name: "both directories exist",
```
vrutkovs:
```suggestion
			name: "merge directories content",
```

pkg/operator/staticpod/file_utils_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+				fileB, err := os.Open(filepath.Join(dirB, "1.txt"))
+				if err != nil {
+					t.Errorf("Expected directory %q to contain file 1.txt: %v", dirB, err)
+				}
+				defer fileB.Close()
+			},
+		},
+		{
+			name: "directory A does not exist",
```
vrutkovs:
```suggestion
			name: "source directory does not exist",
```

pkg/operator/staticpod/file_utils_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+				if err := os.Mkdir(dirA, 0755); err != nil {
+					t.Fatalf("Failed to create directory %q: %v", dirA, err)
+				}
+
+				return dirA, dirB
+			},
+			checkResult: func(t *testing.T, dirA, dirB string, err error) {
+				if !os.IsNotExist(err) {
+					t.Errorf("Expected a directory not exists error, got %v", err)
```
vrutkovs: Lets create a destination dir instead of throwing an error?

pkg/operator/staticpod/file_utils_test.go
```diff
@@ -0,0 +1,132 @@
...
+				if err != nil {
+					t.Errorf("Expected directory %q to contain file 2.txt: %v", dirA, err)
+				}
+				defer fileA.Close()
+
+				fileB, err := os.Open(filepath.Join(dirB, "1.txt"))
+				if err != nil {
+					t.Errorf("Expected directory %q to contain file 1.txt: %v", dirB, err)
+				}
+				defer fileB.Close()
```
vrutkovs: Check file contents here too

vrutkovs (changes requested):

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: It's already ensured by `defer fs.RemoveAll(srcDir)` right after it's created.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: Again, it's already being removed using `defer`.

pkg/operator/staticpod/file_utils_test.go
tchap: This utility function throws an error, but the sync function actually ensures the directory exists.

pkg/operator/staticpod/file_utils_test.go
tchap: There is no merging involved, I am not sure what the context is here.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: There is nothing to restore in the current logic, because we write the whole secrets/configmap directory anew and then we swap it atomically. The current state is not modified until it's finally swapped.

vrutkovs: Ah, so if the swapping fails the source dir is unaffected? In this case there is no need to do the restore, sure

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
tchap: Added in c01d4b2095e05c5552f02c744d675c7a55e8d0ac

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: Perhaps it's bad naming. The current implementation writes the whole secret directory separately anew, but only when there is actually any change when comparing the current state to the secret data. And when the whole secret directory is prepared, it's just atomically switched with the current secret directory using `renameat2`

tchap: So naturally when there is an error preparing that new copy, which is called `srcDir` here, we can just delete it and the current state is unaffected.

pkg/operator/staticpod/file_utils_test.go
tchap: Improved in 63b2d72d8

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: It's not clear to me what to check `ModTime` against.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,104 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
-		klog.Infof("Creating directory %q ...", contentDir)
-		if err := os.MkdirAll(contentDir, 0755); err != nil && !os.IsExist(err) {
-			c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed creating directory for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
+		errors = append(errors, writeFiles(&realFS, c.eventRecorder, secret.TypeMeta, secret.ObjectMeta, contentDir, data, 0600)...)
+	}
+
+	return utilerrors.NewAggregate(errors)
+}
+
+type fileSystem struct {
```
p0lyn0mial: what is the purpose of this struct if unit tests exercise a real FS calls ?

pkg/operator/staticpod/file_utils.go
```diff
@@ -31,3 +33,16 @@ func writeTemporaryFile(content []byte, filePerms os.FileMode, fullFilename stri
...
+//
+// This function requires absolute paths and will return an error if that's not the case.
+func SwapDirectoriesAtomic(dirA, dirB string) error {
+	if !filepath.IsAbs(dirA) {
+		return fmt.Errorf("not an absolute path: %q", dirA)
+	}
+	if !filepath.IsAbs(dirB) {
+		return fmt.Errorf("not an absolute path: %q", dirB)
+	}
+	return unix.Renameat2(0, dirA, 0, dirB, unix.RENAME_EXCHANGE)
```
p0lyn0mial: this is unix specific syscall, os specific syscalls should be guarded by a build tag.

for example: https://github.com/openshift/library-go/blob/master/pkg/network/dialer_linux.go

p0lyn0mial: is there os agnostic way of swapping two dirs ? does os.Rename provide atomicity ?

tchap: I don't think there is atomic directory swap function on Windows. `os.Rename` provides atomicity on Unix only on top of that, and it doesn't allow for switching directories.

tchap: Aligned in 1709074b9

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: To make particular FS calls fail and ensure the current state is unaffected.

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
tchap: Well, actually not such a great idea, then it's not atomic...

pkg/operator/staticpod/certsyncpod/certsync_controller.go
p0lyn0mial: mind pasting links to code that mocks certain calls ?

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// This function requires absolute paths and will return an error if that's not the case.
+func SwapDirectoriesAtomic(dirA, dirB string) error {
+	if !filepath.IsAbs(dirA) {
+		return fmt.Errorf("not an absolute path: %q", dirA)
+	}
+	if !filepath.IsAbs(dirB) {
+		return fmt.Errorf("not an absolute path: %q", dirB)
+	}
+	return unix.Renameat2(0, dirA, 0, dirB, unix.RENAME_EXCHANGE)
```
p0lyn0mial: could you please explain the flags (why 0, and RENAME_EXCHANGE ) and usage (or even how it works) in a comment ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: See https://github.com/openshift/library-go/pull/2009/files#diff-0a7dba73755f1e65e4b795ceec6a81cefa686f0da3b40eafaf0adced01dc6922R150-R194 . All the cases using `errorTestCase` actually.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
tchap: Added in 1547df3d7

pkg/operator/staticpod/file_utils_test.go
tchap: The utility function swaps directories, there is no notion of a source directory really.

pkg/operator/staticpod/file_utils_linux.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+// SwapDirectoriesAtomic can be used to swap two directories atomically.
+//
+// This function requires absolute paths and will return an error if that's not the case.
+func SwapDirectoriesAtomic(dirA, dirB string) error {
```
p0lyn0mial: I think that it would be valuable to have a unit test that shows the cert controller from kas will be notified after changing the dirs wdyt ?

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,33 @@
...
+		return fmt.Errorf("not an absolute path: %q", dirB)
+	}
+
+	// Renameat2 can be used to exchange two directories atomically when RENAME_EXCHANGE flag is specified.
+	// The paths to be exchanged can be specified in multiple ways:
+	//
+	//   * You can specify a file descriptor and a relative path to that descriptor.
+	//   * You can specify an absolute path, in which case the file descriptor is ignored.
+	//
+	// We make sure the path is absolute, hence we pass 0 as the file descriptor as it is ignored.
```
p0lyn0mial: Is there any benefit of using os.Symlink to change the content of two dirs ?

I think os.Symlink is what kubelet uses, xref: https://github.com/kubernetes/kubernetes/blob/v1.26.0/pkg/volume/util/atomic_writer.go#L128

could we reuse what kubelet uses ?

tchap: Honestly this is the most typical way how you would implement this and we should have simply used this.

Now if we were to migrate to this approach, we would need to migrate the current directory approach to a symlink approach. And to do that migration, we would still need to use renameat2 to swap the directory for a symlink, because removing a directory and creating a symlink is not atomic. Which means that if it were to fail, we would get a broken cluster, possibly.

We could certainly use `AtomicWriter`, though, to create a tmp symlink and the associated data directory, then swapping it for the secrets directory, if it exists. If the directory is just a symlink, we can use regular rename. 
 
Not sure this really saves any extra work and testing, though, compared to the PR...

pkg/operator/staticpod/file_utils_linux.go
tchap: I added a test in 7c32c07b93d099869d1afc27239a942054404707

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,33 @@
...
+	//
+	// For more details, see `man renameat2` as that is the associated C library function.
+	return unix.Renameat2(0, dirA, 0, dirB, unix.RENAME_EXCHANGE)
```
p0lyn0mial: is this syscall a mature feature ? how old it is ? just trying to gauge the availability of that feature in kernels.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,33 @@
+//go:build linux
```
p0lyn0mial: I think we should provide an implementation for operating systems other than linux, especially for people who run unit tests on their local machines, wdyt?

tchap: Well, the syscall doesn't exist on other systems. We would need to mock it and then somebody could accidentally use it without knowing there is no promise of atomicity.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
tchap: Linux 3.15, glibc 2.28.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+
+	// Create a tmp source directory to be swapped.
+	srcDir, err := fs.MkdirTemp(filepath.Dir(contentDir), filepath.Base(contentDir)+"-*")
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to create source %s directory for %s/%s: %v", kind, o.Namespace, o.Name, err)
+		errors = append(errors, err)
+		return errors
+	}
+	defer fs.RemoveAll(srcDir)
```
p0lyn0mial: should we handle err returned by fs.RemoveAll?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
-		for filename, content := range secret.Data {
-			// TODO fix permissions
-			fullFilename := filepath.Join(contentDir, filename)
-			// if the existing is the same, do nothing
-			if reflect.DeepEqual(data[fullFilename], content) {
-				continue
-			}
+	}
+	if len(errors) > 0 {
+		return errors
```
p0lyn0mial: previously we would continue populating the dir. why not to do the same ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	}
 
-			klog.Infof("Writing secret manifest %q ...", fullFilename)
-			if err := staticpod.WriteFileAtomic(content, 0600, fullFilename); err != nil {
-				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed writing file for secret: %s/%s: %v", secret.Namespace, secret.Name, err)
-				errors = append(errors, err)
-				continue
-			}
+	// Make sure the tmp directory is correct regarding the target state.
```
p0lyn0mial: what is the point of this check ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	//
+	// The same goal of atomic swap could be implemented using symlinks much like AtomicWriter does in
+	// https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go#L58
+	// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped,
+	// in other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+	// we would anyway need to atomically turn the current data directory to a symlink.
+	// This would all just increase complexity and require atomic swap on the OS level anyway.
+
+	var errors []error
+	kind := strings.ToLower(t.Kind)
```
p0lyn0mial: iirc there are some codecs that remove gvk, like https://github.com/kubernetes/kubernetes/blob/dee76a460ec80f15dc199c93e506586687d42291/staging/src/k8s.io/apimachinery/pkg/runtime/helper.go#L250


maybe just pass "configmap" and "secret" at the calling site. (just don't call the arg kind)

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+
+	// In case the target directory does not exist, create it so that the directory not existing is not a special case.
+	klog.Infof("Ensuring directory %q exists ...", contentDir)
+	if err := fs.MkdirAll(contentDir, 0755); err != nil && !os.IsExist(err) {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed creating content directory for %s: %s/%s: %v", kind, o.Namespace, o.Name, err)
+		errors = append(errors, err)
+		return errors
```
p0lyn0mial: nit: maybe just return []error{err} ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	}
+
+	// Create a tmp source directory to be swapped.
+	srcDir, err := fs.MkdirTemp(filepath.Dir(contentDir), filepath.Base(contentDir)+"-*")
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to create source %s directory for %s/%s: %v", kind, o.Namespace, o.Name, err)
+		errors = append(errors, err)
+		return errors
```
p0lyn0mial: return []error{err} ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	// This is currently implemented as really atomically exchanging directories.
+	//
+	// The same goal of atomic swap could be implemented using symlinks much like AtomicWriter does in
+	// https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go#L58
+	// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped,
+	// in other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+	// we would anyway need to atomically turn the current data directory to a symlink.
+	// This would all just increase complexity and require atomic swap on the OS level anyway.
+
+	var errors []error
```
p0lyn0mial: nit: this could be moved down to line 305

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	kind := strings.ToLower(t.Kind)
+
+	// In case the target directory does not exist, create it so that the directory not existing is not a special case.
+	klog.Infof("Ensuring directory %q exists ...", contentDir)
```
p0lyn0mial: how about `klog.Infof("Ensuring target directory %q exists ...", contentDir)` ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	if err := fs.MkdirAll(contentDir, 0755); err != nil && !os.IsExist(err) {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed creating content directory for %s: %s/%s: %v", kind, o.Namespace, o.Name, err)
+		errors = append(errors, err)
+		return errors
+	}
+
+	// Create a tmp source directory to be swapped.
```
p0lyn0mial: add `klog.Infof("Creating source directory %q exists ...", contentDir)`

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	srcDir, err := fs.MkdirTemp(filepath.Dir(contentDir), filepath.Base(contentDir)+"-*")
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to create source %s directory for %s/%s: %v", kind, o.Namespace, o.Name, err)
```
p0lyn0mial: `"Failed to create source directory for %s: %s/%s: %v", kind, o.Namespace, o.Name, err)`

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
 		}
-		c.eventRecorder.Eventf("CertificateUpdated", "Wrote updated secret: %s/%s", secret.Namespace, secret.Name)
+	}
+	if len(errors) > 0 {
+		return errors
 	}
 
-	return utilerrors.NewAggregate(errors)
+	// Swap directories atomically.
+	klog.Infof("Replacing %s directory %q ...", kind, contentDir)
```
p0lyn0mial: `klog.Infof("Atomically swapping target directory %q with source directory %q for %s ..."...)` ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	if err := fs.SwapDirectoriesAtomic(contentDir, srcDir); err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to enable new %s directory for %s/%s: %v", kind, o.Namespace, o.Name, err)
```
p0lyn0mial: `Failed to swap target directory %q with source directory %q for %s: %s/%s: %v` ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	}
+	defer fs.RemoveAll(srcDir)
+
+	// Populate the tmp directory with files.
+	for filename, content := range files {
+		fullFilename := filepath.Join(srcDir, filename)
+		klog.Infof("Writing %s manifest %q ...", kind, fullFilename)
+
+		if err := fs.WriteFile(fullFilename, []byte(content), filePerm); err != nil {
+			eventRecorder.Warningf("CertificateUpdateFailed",
```
p0lyn0mial: nit: i think that in library-go we usually put everything in a single line. for the sake of consistency i would change it to a single line ( applies to the other places)

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: You swap the whole new directory for the old one. If you fail to write a single file, you have to abort the whole swap, that's the point.

p0lyn0mial: so, previously the file could be potentially corrupted but present. 

stopping on errs ensure that the files won't be removed. ok, i am fine with it.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+				"Failed writing file for %s: %s/%s: %v", kind, o.Namespace, o.Name, err)
 			errors = append(errors, err)
 			continue
```
p0lyn0mial: since we are stopping on errs line 317 what is the point of continuing ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +240,115 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed",
+			"Failed to create source %s directory for %s/%s: %v", kind, o.Namespace, o.Name, err)
+		errors = append(errors, err)
+		return errors
+	}
+	defer fs.RemoveAll(srcDir)
+
+	// Populate the tmp directory with files.
+	for filename, content := range files {
```
p0lyn0mial: how did we handle deletions previously ? (key/file no longer present in the cm/secret).

p0lyn0mial: because in the new model non existing files will be "removed" during the dir swapping.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: I am thinking to make writeFiles simply return err and fail on the first err since it any error anyway causes the whole process to abort. So I guess writing more files after an error is pretty useless.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: Yeah, it's pretty useless, I added it based on a previous discussion, but I am going to revert it.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: It seems currently files removed from the secret/configmap are being left behind and never removed. The directory is only removed as a whole when the secret/configmap no longer exists altogether.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +237,82 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	// In case the target directory does not exist, create it so that the directory not existing is not a special case.
+	klog.Infof("Ensuring target directory %q exists ...", contentDir)
+	if err := fs.MkdirAll(contentDir, 0755); err != nil && !os.IsExist(err) {
+		eventRecorder.Warningf("CertificateUpdateFailed", "Failed creating content directory for %s: %s/%s: %v", typeName, o.Namespace, o.Name, err)
+		return err
+	}
+
+	// Create a tmp source directory to be swapped.
+	klog.Infof("Ensuring source directory %q exists ...", contentDir)
+	srcDir, err := fs.MkdirTemp(filepath.Dir(contentDir), filepath.Base(contentDir)+"-*")
```
vrutkovs: `contentDir` is source and `srcDir` is temporary destination? Names are very confusing

pkg/operator/staticpod/certsyncpod/certsync_controller.go
vrutkovs: My idea was to make sure that `contentDir`'s `ModTime` not newer than the time when we copied it to `srcDir` (to make sure new files were not created). Same for the size.

However my experiments show that `Size` doesn't change and `ModTime` doesn't react on file content update. Most likely the best function for this would be:
* find all files recursively (via `file.Walk`)
* collect file `ModTime` and append into an array
* make a hash out of it (md5 probably would be sufficient)

If this function produces equivalent hashes on src and dest dirs we're good to proceed with the swapping.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +237,82 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+	// Populate the tmp directory with files.
+	for filename, content := range files {
+		fullFilename := filepath.Join(srcDir, filename)
+		klog.Infof("Writing %s manifest %q ...", typeName, fullFilename)
+
+		if err := fs.WriteFile(fullFilename, []byte(content), filePerm); err != nil {
+			eventRecorder.Warningf("CertificateUpdateFailed", "Failed writing file for %s: %s/%s: %v", typeName, o.Namespace, o.Name, err)
+			return err
 		}
-		c.eventRecorder.Eventf("CertificateUpdated", "Wrote updated secret: %s/%s", secret.Namespace, secret.Name)
```
vrutkovs, line 281: Lets log every file write in `srcDir` too, it may be useful to see configmap file list right before the swap

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,391 @@
...
+				"tls.key": []byte("TLS key"),
+			},
+			objectData: map[string][]byte{
+				"tls.crt": []byte("rotated TLS cert"),
+				"tls.key": []byte("rotated TLS key"),
+			},
+			checkState: checkOnSuccess,
+		},
+		{
+			name:  "update target directory change filenames",
```
vrutkovs: lets convert this into "change filenames preserving the content"

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,391 @@
...
+		cert, key := dc.CurrentCertKeyContent()
+		return bytes.Equal(cert, []byte(tlsCertUpdated)) && bytes.Equal(key, []byte(tlsKeyUpdated)), nil
+	})
+	if err != nil {
+		t.Fatalf("Failed to wait for dynamic certificate: %v", err)
+	}
+}
+
+const (
+	tlsCert = `-----BEGIN CERTIFICATE-----
```
vrutkovs: Lets generate those one the fly?

pkg/operator/staticpod/file_utils_linux_test.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		{
+			name: "directory A not an absolute path",
+			setup: func(t *testing.T, tmpDir string) (string, string) {
+				dirA := "a"
+				dirB := filepath.Join(tmpDir, "b")
+				return dirA, dirB
+			},
+			checkResult: func(t *testing.T, dirA, dirB string, err error) {
+				if err == nil || err.Error() != `not an absolute path: "a"` {
```
vrutkovs:
```suggestion
				# This should probably be defined on the test level
				func expectedNotAbsolutePathError(dir string) error {
					 return errors.New(fmt.Sprintf(`not an absolute path: "%s"`, dir))
				}
				if err == nil || !errors.Is(err, expectedNotAbsolutePathError("a")) {
```

pkg/operator/staticpod/file_utils_linux_test.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		{
+			name: "directory B not an absolute path",
+			setup: func(t *testing.T, tmpDir string) (string, string) {
+				dirA := filepath.Join(tmpDir, "a")
+				dirB := "b"
+				return dirA, dirB
+			},
+			checkResult: func(t *testing.T, dirA, dirB string, err error) {
+				if err == nil || err.Error() != `not an absolute path: "b"` {
```
vrutkovs:
```suggestion
				if err == nil || !errors.Is(err, expectedNotAbsolutePathError("b")) {
```

pkg/operator/staticpod/file_utils_linux_test.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		// Check file content.
+		content, err := os.ReadFile(filepath.Join(path, entry.Name()))
+		if err != nil {
+			t.Errorf("Failed to read file %q: %v", entry.Name(), err)
+			continue
+		}
+		if !bytes.Equal(state.Content, content) {
+			t.Errorf("Unexpected content on file %q:\n\texpected %v\n\tgot %v", entry.Name(), string(state.Content), string(content))
```
vrutkovs:
```suggestion
			t.Errorf("Unexpected content in file %q:\n%v", cmp.Diff(string(state.Content), string(content)))
```

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (dir directoryState) CheckDirectoryMatches(t *testing.T, path string) {
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		t.Fatalf("Failed to read directory %q: %v", path, err)
+	}
+
+	for _, entry := range entries {
+		// Get the expected state.
+		state, ok := dir[entry.Name()]
```
vrutkovs: We need to make sure that all items in `dir` are present

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
p0lyn0mial: We need to add an implementation. It can be based on os.Rename.

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,391 @@
...
+		i := c
+		c++
+		if i == n {
+			return errors.New("nuked")
+		}
+		return writeFile(path, data, perm)
+	}
+}
+
+func TestDynamicCertificates(t *testing.T) {
```
p0lyn0mial: Please add a comment explaining why we run this test. (this is where the bug was)

p0lyn0mial: This test must be run only on linux otherwise it will be flaky.

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
```diff
@@ -0,0 +1,391 @@
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		dc.Run(ctx, 1)
+	}()
+	defer wg.Wait()
+	defer cancel()
+
+	// Poll until update detected.
```
p0lyn0mial: before updating the content let's make sure the initial certificate was loaded properly.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
p0lyn0mial: ok, so, this is different from the previous code, tough i don't think we ever removed a certificate.
 
@vrutkovs wdyt?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: You mean as an event? Because a regular log entry is there.

vrutkovs: I think a log line would be better, no need to emit too many events

tchap: It's already there then, I think: `klog.Infof("Writing %s manifest %q ...", typeName, fullFilename)`

pkg/operator/staticpod/file_utils_linux_test.go
tchap: You cannot use `Is` like that, it uses equality of the error, not of the error message, so you typically do stuff like `errors.Is(err, io.EOF)`, but you cannot do this. But I will add a helper function.

vrutkovs: oh, okay, comparing error message is fine then

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
tchap: Do you mean it doesn't have to be atomic? Because it can't be 🙂

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
tchap: The target directory is not actually empty, it doesn't exist.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: But the dest dir contains former secret/configmap files and the src dir contains the new updated version. There is nothing to compare for equality really.

pkg/operator/staticpod/certsyncpod/certsync_controller_test.go
vrutkovs: Hmm, maybe "target dir doesn't exist"?

I usually describe testcase based on the initial situation, not what this test is meant to do. Its a preference, so I'm not insisting.

pkg/operator/staticpod/file_utils_test.go
vrutkovs: if both directories exist they probably have different contents? This would be an interesting test case to have

pkg/operator/staticpod/certsyncpod/certsync_controller.go
vrutkovs: With certificates you can't have "remove just the cert and not the key", its either both are present or neither.

This option is possible for configmap keys though

pkg/operator/staticpod/certsyncpod/certsync_controller.go
vrutkovs: >dest dir contains former secret/configmap files

We need to make sure that dest dirt was not updated between the time we prepared src dir and the swap

pkg/operator/staticpod/file_utils_test.go
tchap: The test case covers writing different files into both directories, swapping and ensuring the directory state matches, including filenames and file contents, so this is already being tested.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: That hashing is now implemented, I am going to improve other tests now as discussed privately.

vrutkovs:
>we should have SwapDirectoriesAtomic implemented for all platforms. This is simply not possible because Linux is the only OS that supports such operation

Iiuc we want an implementation for unittests, which means it doesn't have to be truly atomic, just good enough to pass unittests on other platforms (mac primarily, let's hope no one uses Windows/BSD to run unit tests)

tchap:
I added the mock implementation and tested it on my Mac.

tchap:
/hold

until we are sure the testing KASO PR works just fine.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +239,104 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
 
-	return utilerrors.NewAggregate(errors)
+	// Make sure the target directory hasn't changed in the meantime.
+	klog.Infof("Hashing current content directory %q again and ensuring it's unchanged ...", targetDir)
+	targetDirHashAfter, err := fs.HashDirectory(targetDir)
+	if err != nil {
+		eventRecorder.Warningf("CertificateUpdateFailed", "Failed to hash current content directory for %s: %s/%s: %v", typeName, o.Namespace, o.Name, err)
+		return err
+	}
+	if !bytes.Equal(targetDirHashBefore, targetDirHashAfter) {
```
p0lyn0mial: If two processes are writing to the same directory, hashing won’t guarantee atomicity, because the second process could start modifying the directory after the check. 

A possible solution would be to make the second process also swap the directory atomically, perhaps using the same technique. 

I don’t think the order matters, so atomically swapping the contents should be enough. Thoughts?

vrutkovs: It needs to happen _twice_ - when both `targetDirHashBefore` and `targetDirHashAfter` are calculated _and_ needs to end up in the consistent state (if the second write is a bit slower/faster, the hashes will mismatch). IMO it's too rare to guard against

p0lyn0mial: The number of times we calculate and compare the hashes doesn’t matter, because the second process can start writing or changing the target directory after the last comparison or calculation. 

In this case, the second process is the installer pod. 

We should change the installer pod to perform an atomic swap rather than rely on the hashes.

vrutkovs: In my understanding, the desired procedure is:

1.  make a hash of source dir
1.  prepare a new dir from source dir
1. hash source dir again to make sure it didn't change during previous step
1. we swap atomically if hashes match

So the only chance of invalid swap here is if it initiates after step 1 and between step 3 and 4 - that would be a very rare event, wouldn't it?

p0lyn0mial: > step 3 and 4 - that would be a very rare event, wouldn't it?

This PR addresses a rare bug where kas can observe an old prv key together with a new pub key, which causes a crash :( 

I think we should take this potential issue into account and fix it.

tchap: I aligned `installerpod` to use the same atomic swap. I also refactored the PR a bit because I had to pull shared code into a separate package. The swapping function is now called `SyncDirectory` since it replaces the content actually.

pkg/operator/staticpod/internal/atomicfiles/sync_directory.go
```diff
@@ -0,0 +1,89 @@
+package atomicfiles
```
p0lyn0mial: how about `atomicdir` since we are dealing with directories not files/

pkg/operator/staticpod/internal/atomicfiles/sync_directory.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/klog/v2"
+)
+
+// SyncDirectory can be used to atomically synchronize target directory with the given file content map.
+// This is done by populating a temporary directory, then atomically swapping it with the target directory.
+// This effectively means that any extra files in the target directory are pruned.
+//
+// SyncDirectory is supposed to be used with secrets/configmaps, so typeName is expected to be "configmap" or "secret".
+// This does not affect the logic, but it's included in error messages.
+func SyncDirectory[C string | []byte](
```
p0lyn0mial: How about changing this method to: `Write(targetDir string, files map[string][]byte, filePerm os.FileMode`

After all we are writing files to the target directory in an atomic way. 
Callers have to transform files to []byte which is easy since strings in go are []bytes.

pkg/operator/staticpod/internal/atomicfiles/sync_directory.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Create a tmp source directory to be swapped.
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
+	if err != nil {
+		return fmt.Errorf("failed creating temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+	defer func() {
+		if err := fs.RemoveAll(tmpDir); err != nil {
+			klog.Errorf("Failed to remove temporary directory %q during cleanup: %v", tmpDir, err)
```
p0lyn0mial: We should return the error the the callers not only log it.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"path/filepath"
+
+	"golang.org/x/sys/unix"
+)
+
+// SwapDirectories can be used to swap two directories atomically.
+//
+// This function requires absolute paths and will return an error if that's not the case.
+func SwapDirectories(dirA, dirB string) error {
```
p0lyn0mial: could this function be private, since it is the same pkg ?

p0lyn0mial: assuming we are going to rename the pkg to `atomicdir` we could rename this function to `Swap(source, target string) error` or `Swap(old, new string) error` or maybe just `Rename(old, new string) error`

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
```diff
@@ -0,0 +1,33 @@
...
+	// Renameat2 can be used to exchange two directories atomically when RENAME_EXCHANGE flag is specified.
+	// The paths to be exchanged can be specified in multiple ways:
+	//
+	//   * You can specify a file descriptor and a relative path to that descriptor.
+	//   * You can specify an absolute path, in which case the file descriptor is ignored.
+	//
+	// We make sure the path is absolute, hence we pass 0 as the file descriptor as it is ignored.
+	//
+	// For more details, see `man renameat2` as that is the associated C library function.
+	return unix.Renameat2(0, dirA, 0, dirB, unix.RENAME_EXCHANGE)
```
p0lyn0mial: is there a const we could use instead of 0 (maybe `unix.AT_FDCWD`) ?

pkg/operator/staticpod/internal/atomicfiles/swap_directories_other.go
```diff
@@ -0,0 +1,41 @@
...
+	"os"
+	"path/filepath"
+)
+
+// SwapDirectories swaps two directories, but NOT atomically in this case.
+// Atomic implementation is only available for Linux.
+// This function is essentially a mock for tests, and it simply uses os.Rename.
+// In case there is any error, the swapping process is left in an inconsistent state.
+func SwapDirectories(dirA, dirB string) error {
+	// Still retain the constraints as in the Linux implementation.
```
p0lyn0mial: Add a warning that atomic swap is not possible on unsupported platform.

p0lyn0mial (commented):
@tchap for easier review please open a new pr with the SwapDirectories function + tests.

and then a new PR for the SyncDirectory function.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_other.go
tchap: Isn't the docstring a warning? Or you mean to add a comment to the actual Linux implementation as well...

pkg/operator/staticpod/internal/atomicfiles/swap_directories_linux.go
tchap: I guess we can make this work with relative paths, I overlooked the constant. So this does not have to fail with relative paths actually. Thanks.

pkg/operator/staticpod/internal/atomicfiles/swap_directories_other.go
p0lyn0mial: I was thinking about `klog.Warning` or maybe we could just return ` fmt.Errorf("not supported on this platform")` ?

tchap: But then the unit tests will fail... Not sure it's a big deal. They actually don't work already for some other packages 🙃

p0lyn0mial: could we move the unit tests also to a pkg that would compile/run only on the supported platform ?

tchap: I guess we could, but it will eventually mean all certsync controller tests need to be gated for Linux only.

p0lyn0mial: This should be okay since we support only linux right ? 

The reason I would like to have an impl for the other platforms is to be able to compile code locally to see if there are not errors. The binary is wrapped into a container anyway.

tchap: Right, I get it now.

pkg/operator/staticpod/internal/atomicfiles/sync_directory.go
tchap: I used Sync instead of Write because Write feels like if doesn't affect existing files. But I don't have a strong opinion really.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -60,15 +64,32 @@ func NewCertSyncController(targetDir, targetNamespace string, configmaps, secret
...
 func getSecretDir(targetDir, secretName string) string {
 	return filepath.Join(targetDir, "secrets", secretName)
 }
 
+func getSecretStagingDir(targetDir, secretName string) string {
+	return filepath.Join(getStagingDir(targetDir), "secrets", secretName)
+}
+
 func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	if err := dirutils.RemoveContent(getStagingDir(c.destinationDir)); err != nil {
```
p0lyn0mial: mhm, maybe this could be done in the `Sync` method, after creating the staging dir. wdyt?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -258,34 +266,32 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
...
-		}
-		for filename, content := range secret.Data {
-			if err := writeSecret(content, path.Join(contentDir, filename)); err != nil {
-				return err
-			}
+
+		contentDir := getSecretDir(resourceDir, secretBaseName)
+		stagingDir := getSecretStagingDir(resourceDir, secretBaseName)
+
+		if err := atomicdir.Sync(contentDir, 0700, stagingDir, secret.Data, 0600); err != nil {
```
p0lyn0mial: the sync method expects that the filenames don't hold a path, right ?

tchap: Yeah, it's actually being checked.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: It would make sense, but here we just call it once for the whole staging area while `Sync` works per object/directory. So actually it makes sense to call it here to prune old staging directories, not just the directory for the object being staged.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -152,27 +174,12 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
-				c.eventRecorder.Warningf("CertificateUpdateFailed", "Failed writing file for configmap: %s/%s: %v", configMap.Namespace, configMap.Name, err)
-				errors = append(errors, err)
-				continue
-			}
+		files := make(map[string][]byte, len(configMap.Data))
+		for k, v := range configMap.Data {
+			files[k] = []byte(v)
 		}
-		c.eventRecorder.Eventf("CertificateUpdated", "Wrote updated configmap: %s/%s", configMap.Namespace, configMap.Name)
+		// XXX: Are these permissions correct?
```
tchap: Not sure about this actually...

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
```diff
@@ -185,67 +182,3 @@ func TestSwap(t *testing.T) {
 		})
 	}
 }
-
```
tchap: I just moved this into a separate package to be able to import it.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
p0lyn0mial: I think that we should clean per resource cm/secrets otherwise the old data might be left from the previous runs, right ?

I think that we should do that inside the `Sync` func, wdyt ?


we could add cleaning somewhere [here](https://github.com/openshift/library-go/blob/0a7a5744dc627b54c2b28e633a8c8f1f590a1ab6/pkg/operator/staticpod/internal/atomicdir/sync.go#L55)  and fail if the cleaning fn returns an err, wdyt?

tchap: `atomicdir.Sync` works with a particular object, right? So you would tell it to sync `cm` into `configmaps/cm` and use `staging/cert-sync/configmaps/cm` for staging. So what should it prune exactly? When there is a leftover in `staging/cert-sync`, `Sync` cannot really remove it as it work with a subdir of that path. So my idea was to prune `staging/cert-sync` at the beginning of `sync` so that we are clean and that's it.

Having said that, we can also extend `Sync` to remove everything from `staging/cert-sync/configmaps/cm` just to be sure there are no leftovers and it's encapsulated, but I wanted to ensure that on a higher level with a single rm call.

pkg/operator/staticpod/installerpod/cmd.go
p0lyn0mial: previously perm was 0755, no ?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -258,34 +266,32 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
...
-		klog.Infof("Creating directory %q ...", contentDir)
-		if err := os.MkdirAll(contentDir, 0755); err != nil {
-			return err
-		}
-		for filename, content := range secret.Data {
-			if err := writeSecret(content, path.Join(contentDir, filename)); err != nil {
-				return err
-			}
+
+		contentDir := getSecretDir(resourceDir, secretBaseName)
```
p0lyn0mial: nit: `getSecretTargetDir`

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -258,34 +266,32 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
...
+		files := make(map[string][]byte, len(configmap.Data))
+		for k, v := range configmap.Data {
+			files[k] = []byte(v)
 		}
-		for filename, content := range configmap.Data {
-			if err := writeConfig([]byte(content), path.Join(contentDir, filename)); err != nil {
-				return err
-			}
+
+		if err := atomicdir.Sync(contentDir, 0700, stagingDir, files, 0600); err != nil {
```
p0lyn0mial: previously perm was 0755, no ?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -224,6 +226,12 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
 		return err
 	}
 
+	stagingDirBase := getStagingDir(resourceDir)
+	klog.Infof("Pruning staging directory %q ...", stagingDirBase)
+	if err := dirutils.RemoveContent(stagingDirBase); err != nil {
```
p0lyn0mial: os.RemoveAll won't work ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
p0lyn0mial: prev it was  0755 and 0644 right ?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -626,22 +632,22 @@ func (o *InstallOptions) writePod(rawPodBytes []byte, manifestFileName, resource
...
-func writeConfig(content []byte, fullFilename string) error {
-	klog.Infof("Writing config file %q ...", fullFilename)
+func getStagingDir(targetDir string) string {
+	return filepath.Join(targetDir, "staging", stagingDirUID)
+}
 
-	filePerms := os.FileMode(0600)
-	if strings.HasSuffix(fullFilename, ".sh") {
-		filePerms = 0755
-	}
```
tchap, line 635: Hmm, didn't notice this check for custom permission setting...

pkg/operator/staticpod/certsyncpod/certsync_controller_sync_linux_test.go
```diff
@@ -0,0 +1,674 @@
...
+			extraPaths := sets.NewString()
+			filepath.Walk(controller.destinationDir, func(path string, info os.FileInfo, err error) error {
+				if !info.IsDir() ||
+					path == controller.destinationDir ||
+					strings.HasSuffix(path, "/staging") ||
+					strings.HasSuffix(path, "/staging/cert-sync") ||
+					strings.HasSuffix(path, "/staging/cert-sync/secrets") ||
+					strings.HasSuffix(path, "/staging/cert-sync/configmaps") ||
+					path == filepath.Join(controller.destinationDir, "configmaps") ||
+					path == filepath.Join(controller.destinationDir, "secrets") {
```
tchap: This is not too pretty, but meh.

pkg/operator/staticpod/certsyncpod/certsync_controller.go
```diff
@@ -257,29 +245,57 @@ func (c *CertSyncController) sync(ctx context.Context, syncCtx factory.SyncConte
...
+func getSecretTargetDir(targetDir, secretName string) string {
+	return filepath.Join(targetDir, "secrets", secretName)
+}
+
+func getSecretStagingDir(targetDir, secretName string) string {
+	return filepath.Join(getStagingDir(targetDir), "secrets", secretName)
+}
+
+func getFilePermissions(filename string, defaults os.FileMode) os.FileMode {
+	if strings.HasSuffix(filename, ".sh") {
```
p0lyn0mial: did we have special permission for .sh files for the cert sync ?

pkg/operator/staticpod/certsyncpod/certsync_controller_linux_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
+
+	"github.com/openshift/library-go/pkg/operator/staticpod/internal/atomicdir"
+	"github.com/openshift/library-go/pkg/operator/staticpod/internal/atomicdir/types"
+)
+
+// TestDynamicCertificates makes sure the receiving side of certificate synchronization works as expected.
+// It reads and watches the certificates being synchronized in the same way as e.g. kube-apiserver,
+// the very same libraries are being used.
+func TestDynamicCertificates(t *testing.T) {
```
p0lyn0mial, line 656: could we move this test to `certsync_controller_sync_linux_test` file ? (so that theres a single file for tests)

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -626,22 +643,29 @@ func (o *InstallOptions) writePod(rawPodBytes []byte, manifestFileName, resource
...
+func getSecretStagingDir(targetDir, secretName string) string {
+	return filepath.Join(getStagingDir(targetDir), "secrets", secretName)
+}
 
-	filePerms := os.FileMode(0600)
-	if strings.HasSuffix(fullFilename, ".sh") {
-		filePerms = 0700
+func getFilePermissions(filename string, defaults os.FileMode) os.FileMode {
+	if strings.HasSuffix(filename, ".sh") {
+		defaults |= 0111
```
p0lyn0mial: could we just `return 0700` here ?

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -224,6 +226,12 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
 		return err
 	}
 
+	stagingDirBase := getStagingDir(resourceDir)
```
p0lyn0mial, line 225: in line 224 the installer will create the target dir. i think that for the certsync we delegated that to the atomicdir pkg. should we do the same here ?

pkg/operator/staticpod/certsyncpod/certsync_controller.go
tchap: Yeah, probably not, I will remove it then.

pkg/operator/staticpod/certsyncpod/certsync_controller_linux_test.go
tchap: I guess I will keep `certsync_controller_linux_test.go` file then.

pkg/operator/staticpod/installerpod/cmd.go
tchap: Not really, because the original code returns 700 for secrets and 755 for configmaps.

tchap: I will try to add some tests for the installer, though.

pkg/operator/staticpod/installerpod/cmd.go
```diff
@@ -258,34 +266,43 @@ func (o *InstallOptions) copySecretsAndConfigMaps(ctx context.Context, resourceD
...
-				return err
+
+		contentDir := getConfigMapTargetDir(resourceDir, configMapBaseName)
+		stagingDir := getConfigMapStagingDir(resourceDir, configMapBaseName)
+
+		files := make(map[string]types.File, len(configmap.Data))
+		for name, content := range configmap.Data {
+			files[name] = types.File{
+				Content: []byte(content),
+				Perm:    getFilePermissions(name, 0644),
```
p0lyn0mial: prev the permission for cm was `0600`, no ?

tchap: Chmm, ok, seems to be different from certsync. I got confused.

PR #2011, opened 2025-09-11
NO-JIRA: vendor: Update github.com/ldap/go-ldap/v3

tchap:
The updated version contains Client.Unbind method that is needed.

To make the update work, had to align a fake LDAP client, or rather extend it with new methods.

Required by https://github.com/openshift/oc/pull/2094


pkg/security/ldaptestclient/testclient.go
```diff
@@ -47,8 +48,18 @@ func (c *Fake) StartTLS(config *tls.Config) error {
 }
 
 // Close closes an LDAP connection
-func (c *Fake) Close() {
-	return
+func (c *Fake) Close() error {
```
p0lyn0mial, line 51: changes in this file should be moved to a separate commit

tchap: Done.

PR #2023, opened 2025-09-25
resourcesynccontroller/resourcesync_controller: intro NewResourceSyncControllerWithArtificialDelay

p0lyn0mial:
```
// NewResourceSyncControllerWithArtificialDelay creates a ResourceSyncController // which intentionally introduces a sleep to slow down resyncs. 
// Context:
//   - The controller watches multiple namespaces.
//   - Within those namespaces it observes both Secrets and ConfigMaps.
//   - This leads to a very high number of updates.
//   - When informers lag behind, resyncs are triggered too frequently,
//     causing update bursts and event spam.
//
// Analysis showed this is not a constant fight (not a permanent loop).
//
// TODO:
//   - Filter events to process only those relevant to this controller,
//     instead of all updates in watched namespaces.
//   - Consider introducing a work queue with backoff strategy
//     to better handle event bursts.
//     Before adding backoff, measure the actual frequency
//     of Sync method invocations to make informed adjustments.
```

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -47,10 +47,52 @@ type ResourceSyncController struct {
 
 	runFn   func(ctx context.Context, workers int)
 	syncCtx factory.SyncContext
+
+	// see NewResourceSyncControllerWithArtificialDelay for more info
+	artificialDelay bool
 }
 
 var _ ResourceSyncer = &ResourceSyncController{}
 var _ factory.Controller = &ResourceSyncController{}
+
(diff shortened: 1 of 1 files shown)


pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -188,6 +230,9 @@ func errorWithProvider(provider string, err error) error {
 }
 
 func (c *ResourceSyncController) Sync(ctx context.Context, syncCtx factory.SyncContext) error {
+	if c.artificialDelay {
+		time.Sleep(12 * time.Second)
```
p0lyn0mial, line 234: we could add some jitter but a simple sleep should be enough.

PR #2024, opened 2025-09-25
[release-4.30] resourcesynccontroller/resourcesync_controller: intro NewResourceSyncControllerWithArtificialDelay

p0lyn0mial:
```
// NewResourceSyncControllerWithArtificialDelay creates a ResourceSyncController // which intentionally introduces a sleep to slow down resyncs. 
// Context:
//   - The controller watches multiple namespaces.
//   - Within those namespaces it observes both Secrets and ConfigMaps.
//   - This leads to a very high number of updates.
//   - When informers lag behind, resyncs are triggered too frequently,
//     causing update bursts and event spam.
//
// Analysis showed this is not a constant fight (not a permanent loop).
//
// TODO:
//   - Filter events to process only those relevant to this controller,
//     instead of all updates in watched namespaces.
//   - Consider introducing a work queue with backoff strategy
//     to better handle event bursts.
//     Before adding backoff, measure the actual frequency
//     of Sync method invocations to make informed adjustments.
```

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -47,10 +47,52 @@ type ResourceSyncController struct {
 
 	runFn   func(ctx context.Context, workers int)
 	syncCtx factory.SyncContext
+
+	// see NewResourceSyncControllerWithArtificialDelay for more info
+	artificialDelay bool
 }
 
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
/close 

in favour of https://github.com/openshift/cluster-kube-controller-manager-operator/pull/881

PR #2026, opened 2025-10-01
OCPBUGS-33013: Add atomicdir package

tchap:
`pkg/operator/staticpod/internal/atomicdir` contains internal helpers for performing atomic operations on directories.

This patch contains only a standalone `swap` function, which will be subsequenty used for synchronizing directory with the given state.

This splits some logic from https://github.com/openshift/library-go/pull/2009

### DIFF
--- a/pkg/operator/staticpod/internal/atomicdir/swap_linux.go
+++ b/pkg/operator/staticpod/internal/atomicdir/swap_linux.go
@@ -1,0 +1,22 @@
+//go:build linux
+
+package atomicdir
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// swap can be used to exchange two directories atomically.
+func swap(firstDir, secondDir string) error {
+	// Renameat2 can be used to exchange two directories atomically when RENAME_EXCHANGE flag is specified.
+	// The paths to be exchanged can be specified in multiple ways:
+	//
+	//   * You can specify a file descriptor and a relative path to that descriptor.
+	//   * You can specify an absolute path, in which case the file descriptor is ignored.
+	//
+	// We use AT_FDCWD special file descriptor so that when any of the paths is relative,
+	// it's relative to the current working directory.
+	//
+	// For more details, see `man renameat2` as that is the associated C library function.
+	return unix.Renameat2(unix.AT_FDCWD, firstDir, unix.AT_FDCWD, secondDir, unix.RENAME_EXCHANGE)
+}
--- a/pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
+++ b/pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
@@ -1,0 +1,251 @@
+//go:build linux
+
+package atomicdir
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func TestSwap(t *testing.T) {
+	stateFirst := directoryState{
+		"1.txt": {
+			Content: []byte("hello 1 world"),
+			Perm:    0600,
+		},
+		"2.txt": {
+			Content: []byte("hello 2 world"),
+			Perm:    0400,
+		},
+	}
+	stateSecond := directoryState{
+		"a.txt": {
+			Content: []byte("hello a world"),
+			Perm:    0600,
+		},
+	}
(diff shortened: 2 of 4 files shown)


pkg/operator/staticpod/internal/atomicdir/swap_linux.go
```diff
@@ -0,0 +1,21 @@
...
+
+// swap can be used to exchange two directories atomically.
+func swap(dirA, dirB string) error {
+	// Renameat2 can be used to exchange two directories atomically when RENAME_EXCHANGE flag is specified.
+	// The paths to be exchanged can be specified in multiple ways:
+	//
+	//   * You can specify a file descriptor and a relative path to that descriptor.
+	//   * You can specify an absolute path, in which case the file descriptor is ignored.
+	//
+	// We make sure the path is absolute, hence we pass 0 as the file descriptor as it is ignored.
```
p0lyn0mial: is this still accurate ?

pkg/operator/staticpod/internal/atomicdir/swap_linux.go
```diff
@@ -0,0 +1,21 @@
+//go:build linux
+
+package atomicdir
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// swap can be used to exchange two directories atomically.
+func swap(dirA, dirB string) error {
```
p0lyn0mial: nit: how about: `firstDir` `secondDir` ?

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
```diff
@@ -0,0 +1,222 @@
...
+				if !os.IsNotExist(err) {
+					t.Errorf("Expected a directory not exists error, got %v", err)
+				}
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			pathA, pathB := tc.setup(t, t.TempDir())
```
p0lyn0mial: will the TempDir be automatically removed after the tests?

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	expectedFiles := sets.KeySet(dir)
+	for _, entry := range entries {
+		// Mark the file as visited.
+		expectedFiles.Delete(entry.Name())
+
+		// Get the expected state.
+		state, ok := dir[entry.Name()]
+		if !ok {
+			t.Errorf("Directory %q contains unexpected file %q", path, entry.Name())
```
p0lyn0mial, line 223: ok, so this will find extra files in path but not in dir

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
```diff
@@ -0,0 +1,222 @@
...
+				}
+
+				return dirA, dirB
+			},
+			checkResult: func(t *testing.T, dirA, dirB string, err error) {
+				if !os.IsNotExist(err) {
+					t.Errorf("Expected a directory not exists error, got %v", err)
+				}
+			},
+		},
```
p0lyn0mial, line 178: should we add a case where:
- one dir is empty
- or even both dirs are empty 

?

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
```diff
@@ -0,0 +1,222 @@
...
+		{
+			name: "success with path A relative",
+			setup: func(t *testing.T, tmpDir string) (string, string) {
+				dirA := filepath.Join(tmpDir, "a")
+				dirB := filepath.Join(tmpDir, "b")
+
+				dirStateA.Write(t, dirA)
+				dirStateB.Write(t, dirB)
+
+				cwd, err := os.Getwd()
```
p0lyn0mial, line 78: could we use expectNoError(err) ?

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
tchap: Yes, it is always automatically removed.

pkg/operator/staticpod/internal/atomicdir/swap_linux_test.go
tchap: Not sure I understand. Missing files that are in dir but not in path are checked at the end of this function.

PR #2027, opened 2025-10-01
OCPBUGS-33013: Add atomicdir.Sync function

tchap:
The function can be used to atomically sync a directory with the desired state.
This uses `atomicdir.swap` implemented earlier.

Split from https://github.com/openshift/library-go/pull/2009


pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/klog/v2"
+)
+
+// Sync can be used to atomically synchronize target directory with the given file content map.
+// This is done by populating a temporary directory, then atomically swapping it with the target directory.
+// This effectively means that any extra files in the target directory are pruned.
+//
+// Sync is supposed to be used with secrets/configmaps, so typeName is expected to be "configmap" or "secret"
+// or whatever should be used for potential error messages.
+func Sync[C string | []byte](
```
p0lyn0mial: let's change the signature of this method to `Sync(targetDir string, files map[string][]byte, filePerm os.FileMode) error`

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	RemoveAll:       os.RemoveAll,
+	WriteFile:       os.WriteFile,
+	SwapDirectories: swap,
+}
+
+func sync[C string | []byte](
+	fs *fileSystem, typeName string, o metav1.ObjectMeta,
+	targetDir string, files map[string]C, filePerm os.FileMode,
+) (err error) {
+	// We are going to prepare a tmp directory and write all files into that directory.
```
p0lyn0mial: let's add this text as a doc of the sync function.

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	// This is currently implemented as really atomically exchanging directories.
+	//
+	// The same goal of atomic swap could be implemented using symlinks much like AtomicWriter does in
+	// https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go#L58
+	// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped.
+	// In other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+	// we would anyway need to atomically turn the current data directory into a symlink.
+	// This would all just increase complexity and require atomic swap on the OS level anyway.
+
+	// In case the target directory does not exist, create it so that the directory not existing is not a special case.
```
p0lyn0mial: rm this line, there is a comment below that explain what it does.

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	klog.Infof("Ensuring content directory %q exists ...", targetDir)
+	if err := fs.MkdirAll(targetDir, 0755); err != nil && !os.IsExist(err) {
```
p0lyn0mial: `if err = ...` (err is already defined)

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("failed creating content directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+
+	// Create a tmp source directory to be swapped.
```
p0lyn0mial: rm this line

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return fmt.Errorf("failed creating temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+	defer func() {
+		if ex := fs.RemoveAll(tmpDir); ex != nil {
+			err = fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+		}
+	}()
+
+	// Populate the tmp directory with files.
```
p0lyn0mial: rm this line

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+	for filename, content := range files {
+		fullFilename := filepath.Join(tmpDir, filename)
+		klog.Infof("Writing %s manifest %q ...", typeName, fullFilename)
+
+		if err := fs.WriteFile(fullFilename, []byte(content), filePerm); err != nil {
+			return fmt.Errorf("failed writing file for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+		}
+	}
+
+	// Swap directories atomically.
```
p0lyn0mial: rm this line

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Create a tmp source directory to be swapped.
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
+	if err != nil {
+		return fmt.Errorf("failed creating temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+	defer func() {
+		if ex := fs.RemoveAll(tmpDir); ex != nil {
+			err = fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
```
p0lyn0mial: we missed `ex`

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+		return fmt.Errorf("failed creating content directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+
+	// Create a tmp source directory to be swapped.
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
+	if err != nil {
+		return fmt.Errorf("failed creating temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+	}
+	defer func() {
```
p0lyn0mial, line 56: we also need to take into account that existing err != nil.
maybe something like:
```
var errs []error
		errs = append(errs, err)
		if ex := fs.RemoveAll(tmpDir); ex != nil {
			nErr := fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, ex)
			errs = append(errs, nErr)
		}
		err =  errors.Join(errs)
```
?

we need to check how errors.Join handles nil

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+			err = fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err)
+		}
+	}()
+
+	// Populate the tmp directory with files.
+	for filename, content := range files {
+		fullFilename := filepath.Join(tmpDir, filename)
+		klog.Infof("Writing %s manifest %q ...", typeName, fullFilename)
+
+		if err := fs.WriteFile(fullFilename, []byte(content), filePerm); err != nil {
```
p0lyn0mial: err already exists

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,89 @@
...
+//
+// Sync is supposed to be used with secrets/configmaps, so typeName is expected to be "configmap" or "secret"
+// or whatever should be used for potential error messages.
+func Sync[C string | []byte](
+	typeName string, o metav1.ObjectMeta, targetDir string, files map[string]C, filePerm os.FileMode,
+) error {
+	return sync(&realFS, typeName, o, targetDir, files, filePerm)
+}
+
+type fileSystem struct {
```
p0lyn0mial, line 22: instead of a new struct let's define vars that will be change by the tests:
```
var mkDirAll = os.MkdirAll
...
```

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
```diff
@@ -0,0 +1,329 @@
...
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSync(t *testing.T) {
```
p0lyn0mial, line 15: since this test uses a realFS it needs to be run on linux

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
```diff
@@ -0,0 +1,329 @@
...
+			},
+			objectData: map[string][]byte{
+				"api.crt": []byte("rotated TLS cert"),
+				"api.key": []byte("rotated TLS key"),
+			},
+			checkState: checkOnError,
+		}
+	}
+
+	testCases := []testCase{
```
p0lyn0mial, line 58: could we change the test to have:
```
...
existingDirectoryData ...
filesToSync ...
expectedTargetDirectoryData ...
```
?

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: I am not super keen on this, because it means you have to reset it for every test since it's global state. I mean, it works, but it's not especially pretty...

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,106 @@
...
+// Sync can be used to atomically synchronize target directory with the given file content map.
+// This is done by populating a temporary directory, then atomically swapping it with the target directory.
+// This effectively means that any extra files in the target directory are pruned.
+//
+// Sync is supposed to be used with secrets/configmaps, so typeName is expected to be "configmap" or "secret"
+// or whatever should be used for potential error messages.
+//
+// The first return value indicates whether the state has been synchronized.
+// This can be the case even though an error is returned since that can be related to cleaning up.
+func Sync(typeName string, o metav1.ObjectMeta, targetDir string, files map[string][]byte, filePerm os.FileMode) (bool, error) {
```
p0lyn0mial: why not to drop `typeName string, o metav1.ObjectMeta` too ? why do we need this args ?

p0lyn0mial: why do we need to return additional `bool` ?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,106 @@
...
+// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped.
+// In other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+// we would anyway need to atomically turn the current data directory into a symlink.
+// This would all just increase complexity and require atomic swap on the OS level anyway.
+func sync(
+	fs *fileSystem, typeName string, o metav1.ObjectMeta,
+	targetDir string, files map[string][]byte, filePerm os.FileMode,
+) (synchronized bool, retErr error) {
+	var errs []error
+	defer func() {
```
p0lyn0mial: I preferred the previous version. 

How about changing the defer function down below to:
```
defer func() {
		if ex := fs.RemoveAll(tmpDir); ex != nil {
			if err != nil { /*prev error*/
				err = fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w, prevErr: %w", typeName, o.Namespace, o.Name, ex, err)
				return
			}
			err = fmt.Errorf("failed removing temporary directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, ex)
		}
	}()
```

?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	klog.Infof("Ensuring content directory %q exists ...", targetDir)
+	if err := fs.MkdirAll(targetDir, 0755); err != nil && !os.IsExist(err) {
+		addErr(fmt.Errorf("failed creating content directory for %s: %s/%s: %w", typeName, o.Namespace, o.Name, err))
+		return
+	}
+
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
```
p0lyn0mial: for the targetDir - `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client`
will the tempDir be `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/aggregator-client-{random}'` ?

pkg/operator/staticpod/internal/atomicdir/sync.go
p0lyn0mial: ok, let's leave it for now.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Yeah, I guess that can be removed. It's just for error messages, but the caller should handler that.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Yeah, exactly.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: I thought there would be value in signalling the state is synchronized, even though there is an error. For logging purposes, for example. Digging that from the error seems like unnecessary work.

p0lyn0mial: why not to use `err` for signalling the state ?

tchap: Strictly speaking you surely can, I was just trying to be nice to myself for the subsequent PR so that I don't have to dig into https://pkg.go.dev/io/fs#PathError and guess the state was actually written and this is an error from `os.RemoveAll` called on the tmp directory 🥲

p0lyn0mial: for our use case, the callers will collect the err (not interpret it) and continue synchronising the remaining items.
no ? 

and since the pkg is internal we control the callers.

tchap: Yeah, sure. But I was not really planning to fail the reconciliation call when we fail to remove the tmp directory as the last step. The desired state is there. But strictly speaking you can just retry and we are fine.

I am actually thinking whether we should use a certain temporary directory prefix so that we can actually GC tmp directories that were not removed during previous reconciliation loop iterations...

p0lyn0mial: > I am actually thinking whether we should use a certain temporary directory prefix so that we can actually GC tmp directories that were not removed during previous reconciliation loop iterations...

wdym? is there a way to auto remove old tmp dirs ?

tchap: Well, we can do it in the reconciliation loop to make sure they don't accumulate.

p0lyn0mial: might not be that easy since we have more than one process creating tmpDirs ...

p0lyn0mial: > Yeah, sure. But I was not really planning to fail the reconciliation call when we fail to remove the tmp directory as the last step. The desired state is there. But strictly speaking you can just retry and we are fine.

that is a good point, on the other hand if being unable to remove the tmpDir is very common the controller will eventually go degraded and the error won't go unnoticed

tchap: Yeah, that's true, I tend to forget this. I guess you can detect tmp dirs by prefix, but then check the time they were created? Like, if the directory is older than 24 hours, you are surely safe to delete it?

tchap: > that is a good point, on the other hand if being unable to remove the tmpDir is very common the controller will eventually go degraded and the error won't go unnoticed

Yeah, I can't think of being unable to remove these dirs unless something is rather wrong, but I just had this tendency to be defensive. Like, even if it happens occasionally, you really don't want to leave the directories there. It will just go unnoticed for a long time, if it happens occasionally. But yeah, the danger is probably realistically pretty low. I do somehow feel like I will be cursing this comment in a year or so 😅

tchap: I think that just using a common tmpdir prefix like `_atomicdir_tmp` doesn't really matter at all anyway, I will probably do that, I just wonder whether we want to skip the synchronized flag or not then. It's no extra work to have it there TBH. Maybe a little bit more unit testing...

p0lyn0mial: for sure worth considering, on the other hand the previous code (https://github.com/openshift/library-go/blob/43089c7fce438050c299fbf47a4013664e6d67bf/pkg/operator/staticpod/file_utils.go#L9)  also didn't clean up the tmp files so i don't think that issue is very common.

p0lyn0mial: I would keep it simple, I would drop the synchronised flag and i would always return an error and i would not clean up the old (unused) tmpDirs.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: I am changing this to `/etc/kubernetes/static-pod-resources/kube-apiserver-certs/secrets/_atomicdir_tmp_aggregator-client-{random}`

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Very well then.

pkg/operator/staticpod/internal/atomicdir/sync.go
p0lyn0mial: why ?

tchap: I think that it's still useful to have the tmp directories easy to recognize. There is no cost to this. Or you feel strongly about not having the common prefix?

p0lyn0mial: yeah, I like `aggregator-client-{random}` which already is easy to recognise.  I would use it unless there is something wrong with that.

tchap: Ok, I will roll that back.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Ok, this seems legit. Didn't know `fmt.Errorf` will create an error with `Unwrap() []error` when there are multiple `%w` in the formatting string, but it makes sense.

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
tchap: I will try to make it more declarative.

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// The same goal of atomic swap could be implemented using symlinks much like AtomicWriter does in
+// https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go#L58
+// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped.
+// In other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+// we would anyway need to atomically turn the current data directory into a symlink.
+// This would all just increase complexity and require atomic swap on the OS level anyway.
+func sync(fs *fileSystem, targetDir string, files map[string][]byte, filePerm os.FileMode) (retErr error) {
+	klog.Infof("Ensuring content directory %q exists ...", targetDir)
+	if err := fs.MkdirAll(targetDir, 0755); err != nil && !os.IsExist(err) {
```
p0lyn0mial: do we need `os.IsExist(err) ` ?

```
// If path is already a directory, MkdirAll does nothing
// and returns nil.
```

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,81 @@
...
+	"path/filepath"
+
+	"k8s.io/klog/v2"
+)
+
+// Sync can be used to atomically synchronize target directory with the given file content map.
+// This is done by populating a temporary directory, then atomically swapping it with the target directory.
+// This effectively means that any extra files in the target directory are pruned.
+//
+// The first return value indicates whether the state has been synchronized.
```
p0lyn0mial: outdated ?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,81 @@
...
+// This is currently implemented as really atomically swapping directories.
+//
+// The same goal of atomic swap could be implemented using symlinks much like AtomicWriter does in
+// https://github.com/kubernetes/kubernetes/blob/v1.34.0/pkg/volume/util/atomic_writer.go#L58
+// The reason we don't do that is that we already have a directory populated and watched that needs to we swapped.
+// In other words, it's for compatibility reasons. And if we were to migrate to the symlink approach,
+// we would anyway need to atomically turn the current data directory into a symlink.
+// This would all just increase complexity and require atomic swap on the OS level anyway.
+func sync(fs *fileSystem, targetDir string, files map[string][]byte, filePerm os.FileMode) (retErr error) {
+	klog.Infof("Ensuring content directory %q exists ...", targetDir)
```
p0lyn0mial: `Ensuring target directory...` ?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := fs.MkdirAll(targetDir, 0755); err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed creating content directory: %w", err)
```
p0lyn0mial: `target directory` ?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,81 @@
...
+
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
+	if err != nil {
+		return fmt.Errorf("failed creating temporary directory: %w", err)
+	}
+	defer func() {
+		if err := fs.RemoveAll(tmpDir); err != nil {
+			if retErr != nil {
+				retErr = fmt.Errorf("failed removing temporary directory %q: %w; previous error: %w", tmpDir, err, retErr)
```
p0lyn0mial: we need `return` statement, otherwise the retErr will always be replaced by the line below.

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
```diff
@@ -0,0 +1,349 @@
...
+			name:  "delete a single file",
+			newFS: newRealFS,
+			existingFiles: map[string][]byte{
+				"1.txt": []byte("1"),
+				"2.txt": []byte("2"),
+			},
+			filesToSync: map[string][]byte{
+				"1.txt": []byte("1"),
+			},
+			expectDirectorySynchronized: true,
```
p0lyn0mial: It would be nice to see the expected files in the dir after the sync (`expectedTargetDirectoryFiles`).

I think that in this case the `2.txt` file will be removed.

Now, I think that is different from the code that is currently in use. The question is if this is an issue.
So technically once we switch to this code some files might be removed from the target directory.

On the other hand after removing an entry from a CM/Secret we would want to remove the file, right ?

Please double check if the above is correct in https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/certsyncpod/certsync_controller.go#L71

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: This is what happens when you want to write if else and your brain switches to if and return in the middle, but forgets to write it down 😅

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
tchap: Correct. We already discussed this in #2009 that this will now be different. We will be deleting files now.

tchap: Rewrote the test cases to contain `expectedFiles`.

p0lyn0mial: I’m not sure if removing orphaned files is the best strategy. Just thinking out loud. 

The previous code has been in use for years, and as we know, behaviour eventually becomes the API. Changing this behaviour might break clusters. Removing a file is an irreversible operation. Previously, deletion was only possible after removing the entire CM/Secret, not individual keys. 

Maybe we should consider preserving the previous behaviour? One option would be to introduce a `targetDirPreservePolicy` that callers can set. If the policy is `Keep`, then we could simply copy orphaned files from the target directory to the temp directory. If there’s a real need to delete missing keys, that could be introduced per CM/Secret and passed into this method. 

WDYT?

/cc @benluddy

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,79 @@
...
+				return
+			}
+			retErr = fmt.Errorf("failed removing temporary directory %q: %w", tmpDir, err)
+		}
+	}()
+
+	for filename, content := range files {
+		fullFilename := filepath.Join(tmpDir, filename)
```
benluddy: I'd like to ensure that `fullFilename` is a descendant of `tmpDir` here, even if it's unlikely that a future caller will pass something like `map[string][]byte{"../foo": []byte{'b', 'a', 'r'}}`.

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,79 @@
...
+// we would anyway need to atomically turn the current data directory into a symlink.
+// This would all just increase complexity and require atomic swap on the OS level anyway.
+func sync(fs *fileSystem, targetDir string, files map[string][]byte, filePerm os.FileMode) (retErr error) {
+	klog.Infof("Ensuring target directory %q exists ...", targetDir)
+	if err := fs.MkdirAll(targetDir, 0755); err != nil {
+		return fmt.Errorf("failed creating target directory: %w", err)
+	}
+
+	klog.Infof("Creating temporary directory to swap for %q ...", targetDir)
+	tmpDir, err := fs.MkdirTemp(filepath.Dir(targetDir), filepath.Base(targetDir)+"-*")
```
benluddy: Could we handle a scenario where sync repeatedly fails between populating the staging directory and cleaning up?

What if the caller provided its own (static) directory path to use for staging instead of generating a random path internally? We'd have an upper limit of 1 abandoned staging directory per unique caller.

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
benluddy: What would the failure mode look like if we had a dependency on preserving orphaned files? It's hard for me to imagine a specific scenario where an orphaned file is necessary on an upgraded cluster but not on a freshly-installed cluster.

If we _don't_ have a dependency on orphaned files, is preserving them an ongoing risk? If we were building this from scratch today, we would definitely not preserve them (I think?).

If we can't prove that pruning orphaned files is safe, but we think that it probably is, is there some process that would make us comfortable with pruning? For example, we could preserve orphaned files with telemetry to detect when it happens in practice, or back up any orphaned files and keep them around for disaster recovery purposes.

I lean toward pruning them, but it also seems unlikely to me that we'd miss an issue like this during upgrade testing.

tchap: I can't say I've been with OpenShift for too long, but it seems to me very unlikely as well that such an issue would go unnoticed. The orphaned files are simply not present on any new cluster you decide to install, right? So from my perspective, I am fine with pruning, but yeah, I can't prove anything.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Yeah, we could do that. My idea was actually to delete any leftovers in the sync loop where this function will be used. But it's true that actually passing the staging directory makes this easier as the caller then knows where it is and can do garbage collection easily.

tchap: Added the explicit staging dir in a separate commit.

pkg/operator/staticpod/internal/atomicdir/sync.go
tchap: Amended.

pkg/operator/staticpod/internal/atomicdir/sync_linux_test.go
p0lyn0mial: Yeah, I also think that not removing files created based on the content of a CM/Secret would be considered a bug. OK, if you both lean toward pruning, then I’m fine with pruning as well.

pkg/operator/staticpod/internal/atomicdir/sync.go
p0lyn0mial: I’m fine with providing a static dir, though in practice the immediate callers will just create temporary dirs. 

Note that this package is internal, and we fully control the callers. For our own convenience, it’s simply more practical to have this function create tmp dir. 

If the callers are responsible for creating the static dir, then I also think they should be responsible for removing it.

tchap: The staging dir is created and deleted by this package, but specified by the caller.

p0lyn0mial: ok, I misunderstood the proposal here.

how will the callers prepare the staging dir path? Will it be stable?

tchap: Depends on the caller 🙂 It can be stable, like `secrets/tls-cert` is the target dir, `staging/secrets/tls-cert` is staging. Or the caller can actually create a random tmp directory and pass it here, MkdirAll will then be a noop.

tchap: We actually wanted to incorporate the caller, I guess, so it can be `staging/certsync/secrets/tls-cert`, perhaps.

p0lyn0mial: remember, there’s more than one process using this function. We need to make sure the staging dirs are unique, otherwise these two processes will step on each other’s toes, right?

tchap: Yep, that's why I mentioned the caller in the next comment. You can actually just delete `staging/certsync` every time `sync` function is called, because you have full control.

p0lyn0mial: ok, so does that mean we’ll add a unique caller "id" to the staging dir for each process?

tchap: I would certainly do that.

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,86 @@
...
+				retErr = fmt.Errorf("failed removing staging directory %q: %w; previous error: %w", stagingDir, err, retErr)
+				return
+			}
+			retErr = fmt.Errorf("failed removing staging directory %q: %w", stagingDir, err)
+		}
+	}()
+
+	for filename, content := range files {
+		// Make sure filename is a plain filename, not a path.
+		// This also ensures the temporary directory cannot be escaped.
```
p0lyn0mial: nit: the staging directory ... ?

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,86 @@
...
+
+		fullFilename := filepath.Join(stagingDir, filename)
+		klog.Infof("Writing file %q ...", fullFilename)
+
+		if err := fs.WriteFile(fullFilename, content, filePerm); err != nil {
+			return fmt.Errorf("failed writing %q: %w", fullFilename, err)
+		}
+	}
+
+	klog.Infof("Atomically swapping target directory %q with temporary directory %q ...", targetDir, stagingDir)
```
p0lyn0mial: `with staging ...`

pkg/operator/staticpod/internal/atomicdir/sync.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := fs.SwapDirectories(targetDir, stagingDir); err != nil {
+		return fmt.Errorf("failed swapping target directory %q with temporary directory %q: %w", targetDir, stagingDir, err)
```
p0lyn0mial: `with staging...`

tchap: All issues fixed and commits squashed.

PR #2029, opened 2025-10-02
operator/resourcesynccontroller/resourcesync_controller: add WithConditionPrefix

p0lyn0mial:
```
// WithConditionPrefix sets a prefix for the controller's condition types.
// This is useful when multiple operators reuse the same controller but would like to have
// distinct conditions.
```

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -45,6 +45,8 @@ type ResourceSyncController struct {
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces
 	operatorConfigClient       v1helpers.OperatorClient
 
+	degradedConditionType string
+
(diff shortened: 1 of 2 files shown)


p0lyn0mial:
/hold

for a proof pr in the operators.

pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -109,6 +112,14 @@ func (c *ResourceSyncController) Name() string {
 	return c.controllerInstanceName
 }
 
+// WithConditionPrefix sets a prefix for the controller's condition types.
+// This is useful when multiple operators reuse the same controller but would like to have
+// distinct conditions.
+func (c *ResourceSyncController) WithConditionPrefix(prefix string) *ResourceSyncController {
```
p0lyn0mial, line 118: We probably don’t want to change the instance name, so that there is only one manager. 

Alternatively, we could add a new constructor that allows assigning a new prefix.

PR #2030, opened 2025-10-03
CORS-4180: Update openshift/api version to get the latest featuregates

sadasu:
Featuregates for DualStack support for AWS and Azure were recently added. Would like to bring that in with the openshift/api version update.


sadasu:
@p0lyn0mial could you PTAL too? Thanks!

PR #2031, opened 2025-10-07
CNTRLPLANE-1544: staticpod: guard: Use user namespace

tchap:
#### Test for Downstream Components

- [x] kube-controller-manager: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/887
- [x] kube-apiserver: https://github.com/openshift/cluster-kube-apiserver-operator/pull/1936
- [x] etcd: https://github.com/openshift/cluster-etcd-operator/pull/1498
- [x] kube-scheduler: https://github.com/openshift/cluster-kube-scheduler-operator/pull/582

### DIFF
--- a/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
+++ b/pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
@@ -14,6 +14,7 @@ spec:
   terminationGracePeriodSeconds: 3
   tolerations:
     - operator: Exists
+  hostUsers: false
   containers:
     - name: guard
       image: # Value set by operator


pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
```diff
@@ -14,6 +15,12 @@ spec:
   terminationGracePeriodSeconds: 3
   tolerations:
     - operator: Exists
+  hostUsers: false
```
p0lyn0mial, line 17: are we sure it is a supported feature ?

for example, [the docs for 4.19](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/nodes/working-with-pods#nodes-pods-user-namespaces) says it is a TP feature only.

tchap: The linked issue mentions that this is included by default in OCP 4.20.
I am not sure the feature request would be there if this was not supported, right?

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: # Value set by operator
   annotations:
     target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+    openshift.io/required-scc: restricted-v3
```
p0lyn0mial: ok, so this sets 'RequirePodLevel'

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
p0lyn0mial: ok, also it will work during upgrades because for odd versions kubelet can be behind only 1 version, right (https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/architecture/control-plane#control-plane-and-node-host-compatibility)

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
```diff
@@ -14,6 +15,12 @@ spec:
   terminationGracePeriodSeconds: 3
   tolerations:
     - operator: Exists
+  hostUsers: false
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
```
p0lyn0mial: why do we want to set a concrete uid here ? i think that an uid will be assigned if not specified, right ? is there any benefit of setting a uid in this context ?

tchap: You are right that some of the IDs will get assigned automatically, let me test it on a pod.

In any case, it will get set to 1000 by default.

p0lyn0mial:
tests are green in the components repo

/hold cancel

PR #2032, opened 2025-10-08
NO-JIRA: Rely on os/user for user lookup instead of runc 

ardaguclu:
This PR is revival of https://github.com/openshift/library-go/pull/1949. 

User lookup functionalities are almost identical between [opencontainers/runc](https://github.com/moby/sys/blob/9dc3a90b82699c54554635f76a6a49757e02c523/user/lookup_unix.go#L37-L58) and [os/user](https://github.com/golang/go/blob/78b43037dc20b9f5d624260b50e15bfa8956e4d5/src/os/user/lookup_unix.go#L178-L189) (standard Go library). So we don't need to import an external package just to extra a name of the current user/group.

This PR moves away from opencontainers/runc package to os/user to drop this external dependency.


ardaguclu:
If the concern is about cgo usages; I think, this will be used https://github.com/golang/go/blob/78b43037dc20b9f5d624260b50e15bfa8956e4d5/src/os/user/lookup_unix.go#L5, if cgo is not enabled.

pkg/certs/cert-inspection/certgraphanalysis/locationmetadata.go
```diff
@@ -24,10 +25,10 @@ func getOnDiskLocationMetadata(path string) *certgraphapi.OnDiskLocationWithMeta
 	if info, err := os.Stat(path); err == nil {
 		ret.Permissions = info.Mode().Perm().String()
 		if statt, ok := info.Sys().(*syscall.Stat_t); ok {
-			if u, err := user.LookupUid(int(statt.Uid)); err == nil {
+			if u, err := user.LookupId(fmt.Sprintf("%d", statt.Uid)); err == nil {
 				ret.User = u.Name
 			}
-			if g, err := user.LookupGid(int(statt.Gid)); err == nil {
+			if g, err := user.LookupGroupId(fmt.Sprintf("%d", statt.Gid)); err == nil {
```
bertinatto: The last time I researched about this I found out that the opencontainers implementation was created because `os/user` doesn't not work well within containers.

That implementation is now deprecated and migrated to https://github.com/moby/sys/blob/main/user/lookup_unix.go.

I'm +1 to replace the deprecated implementation with something supported, but I'd like to understand better if it's safe to use `os/user` over `moby/sys`. Do you think you could look into that?

Also, can you create a proof PR that exercises this change?

ardaguclu: Thanks for review. I'm fine to move to `moby/sys` as well. Because this attempt also removes the vulnerable library from here.  Let me update the PR with this. I agree that this would be the safest option.

ardaguclu: @bertinatto I've updated the code to directly use it from `moby/sys/user`. The code is exactly the same. Do you still want me to open a proof PR?.

PR #2036, opened 2025-10-15
OCPBUGS-33013: atomicdir: Set permission for each file separately

tchap:


pkg/operator/staticpod/internal/atomicdir/directory_state.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/google/go-cmp/cmp"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type File struct {
+	Content []byte
+	Perm    os.FileMode
+}
```
tchap: Exported these types because they are gonna be used from multiple packages in the future.

pkg/operator/staticpod/internal/atomicdir/directory_state.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for filename, state := range dir {
+		fullFilename := filepath.Join(dirPath, filename)
+		if err := os.WriteFile(fullFilename, state.Content, state.Perm); err != nil {
+			t.Fatalf("Failed to write file %q: %v", fullFilename, err)
+		}
+	}
+}
+
+func (dir DirectoryState) CheckDirectoryMatches(t *testing.T, dirPath string, dirPerm os.FileMode) {
```
tchap: There was duplicate logic for checking directory content, actually in the end I selected the one that loads all directory content into a map and uses `cmp.Diff` to find any differences. Basically the `checkDirectoryContents` function that is removed now.

`DirectoryState` is otherwise just existing  code moved to a separate file.

pkg/operator/staticpod/internal/atomicdir/directory_state.go
p0lyn0mial: ok but this file contain test related code, right ? I would like to keep production code separate from the test code.

pkg/operator/staticpod/internal/atomicdir/directory_state.go
p0lyn0mial: where are we going to use this function ?

p0lyn0mial: ok in #2009 then I think we need a testing pkg

p0lyn0mial: here is an example https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/features/testing/features.go#L48

pkg/operator/staticpod/internal/atomicdir/directory_state.go
tchap: Well, yeah, but it needs to be exported. I can put it into a separate package, if you want.

tchap: `File` is actually used in prod and tests as well, but that's the only case.

p0lyn0mial: could we add it to https://github.com/tchap/library-go/blob/4e120fb5521576fc02fc58c6dbc53785d842767c/pkg/operator/staticpod/internal/atomicdir/sync.go ?

tchap: Done.

pkg/operator/staticpod/internal/atomicdir/directory_state.go
tchap: I pulled it into a separate package `dirutils`. Not such a great name, but it's internal, meh.

PR #2037, opened 2025-10-15
OCPSTRAT-2371: Update to Kubernetes v1.34.1

bertinatto:


pkg/operator/resource/resourceapply/apps_test.go
```diff
@@ -69,7 +69,7 @@ func TestApplyDeployment(t *testing.T) {
 			actualDeployment: workloadWithDefaultSpecHash(),
 			expectedDeployment: func() *appsv1.Deployment {
 				w := workload()
-				w.Annotations["operator.openshift.io/spec-hash"] = "5322a9feed3671ec5e7bc72c86c9b7e2f628b00e9c7c8c4c93a48ee63e8db47a"
+				w.Annotations["operator.openshift.io/spec-hash"] = "0c4e1cc4d475df8bd73d71d0c56fcf95774c9f4d8bcc676eec369aef3b13a570"
```
benluddy, line 72: What changed?

pkg/template/templateprocessing/template_test.go
```diff
@@ -249,7 +249,7 @@ func TestEvaluateLabels(t *testing.T) {
 				]
 			}`,
 			Output: `{
-				"kind":"Template","apiVersion":"template.openshift.io/v1","metadata":{"creationTimestamp":null},
```
benluddy, line 252: Yes, this is expected because the omitzero field tag was added to metav1.ObjectMeta's CreationTimestamp.

pkg/operator/resource/resourceapply/apps_test.go
benluddy: I'm guessing it's also due to the omitzero on the creation timestamp, because the PodTemplateSpec inside DeploymentSpec embeds ObjectMeta?

bertinatto: `creationTimestamp`

PR #2040, opened 2025-10-23
OCPBUGS-39241: Rate-limit RequiredInstallerResourcesMissing events in InstallerController to prevent SNO upgrade test failures 

ShazaAldawamneh:
Rate-limit RequiredInstallerResourcesMissing events in InstallerController

During SNO upgrades, the installer would emit a flood of
RequiredInstallerResourcesMissing events for transient missing secrets
and configmaps, causing the [bz-etcd] pathological test to fail.

This patch adds a 30-second rate-limit per unique set of missing resources:

Events for the same missing set are only emitted once per 30 seconds.
Aggregated errors are still returned to trigger retries.
Uses InstallerController.now() for testable timestamps.
lastMissingEvent map tracks last emission times per missing resource set.
This prevents event spam while preserving retry logic and ensures
transient missing resources are still detected correctly.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


everettraven (commented):
Is the event creation frequency the problem here, or is it a symptom of another issue?

The bug says we see this on upgrades from 4.17 -> 4.18. Are we seeing this on any other releases/upgrades?

I'm skeptical of rate limiting event creation as the right solution here.

PR #2041, opened 2025-10-25
NO-JIRA: Introduce KMS encryption mode in encryption library

ardaguclu:
This PR introduces KMS encryption mode into encryption library in library-go. Since there is no keys in KMS, it uses hashsum of KMS Configuration as well as KeyID returned by KMS plugin (by calling an grpc call to `Status` endpoint). 

This PR relies on the building blocks of encryption library that works for other encryption types.

https://github.com/openshift/library-go/pull/1876 was a good attempt. This PR adopts some bits from there.

### DIFF
--- a/pkg/operator/encryption/kms/plugin.go
+++ b/pkg/operator/encryption/kms/plugin.go
@@ -2,6 +2,7 @@ package kms
 
 import (
 	"fmt"
+	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
 
@@ -87,7 +88,7 @@ func (c *ContainerConfig) setDefaults() {
 		if err != nil {
 			panic(err)
 		}
-		c.SocketPath = socket
+		c.SocketPath = strings.Replace(socket, "unix://", "", 1)
 	}
 	if c.HealthPort == 0 {
 		c.HealthPort = defaultHealthPort


pkg/operator/encryption/crypto/keys.go
```diff
@@ -12,6 +12,7 @@ var (
 		state.AESGCM:    NewAES256Key,
 		state.SecretBox: NewAES256Key, // secretbox requires a 32 byte key so we can reuse the same function here
 		state.Identity:  NewIdentityKey,
+		state.KMS:       NewIdentityKey, // this is not used in KMS
```
flavianmissi: Why are you skipping this for KMS? `ModeToNewKeyFunc` is used by `generateKeySecret`, which is used by `checkAndCreateKeys`.
Instead of directly checking the kms mode and config in `checkAndCreateKeys`, have you considered adding a new mode function to this map for KMS?
This is the intended mechanism for creating new encryption keys, and I think we should use it, if we don't exactly create an encryption key for the KMS mode.
What do you think?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +107,25 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
...
+				ks = state.KeyState{
+					Key: apiserverconfigv1.Key{
+						Name: keyName,
+						// We set this unused secret just to align with what we set initially.
+						// This is unused.
+						Secret: base64.StdEncoding.EncodeToString(crypto.ModeToNewKeyFunc[state.KMS]()),
+					},
+					Mode:          state.KMS,
+					KMSConfigHash: configHash,
+					KMSKeyIDHash:  keyIDHash,
```
flavianmissi: I _think_ that if we use the `Key.Secret` part of `state.KeyState` for these two values we might not need to change the provider agnostic `state.KeyState` to contain KMS specific parameters.
What do you think?

pkg/operator/encryption/crypto/keys.go
ardaguclu: That sounds like a reasonable approach. We need to _somehow_ merge kmsconfig and key_id into one secret value. Do you have any idea how we can merge them?.

```go
struct KMSKeySecret {
KMSConfig kmsConfig
KeyID string
}
```

marshal -> hashsum into Secret. Does this make sense?. unix socket name and plugin provider name can contain the same hashsum, as long as we use same mechanism.

Also if we merge them, we'll lose the discrimination between kms-config-changes and key-id-rotated. But I don't think we'll need such discrimination.

ardaguclu: I think, we can use this struct marshalling mechanism and feeding into the Secret field of Key. I'll move that way to see the implementation.

ardaguclu: hmm... we need to generate unix socket path with a predefined hash value to obtain `key_id`. Without key_id, we can't generate unique secret. In any case, we'll need separate kmsconfighash for unix socket path.

I don't want to immediately abandon your suggestion, as it is indeed simpler. Let me try to think more about this.

flavianmissi: Can't we initially use the APIServer KMS Encryption config key id (i.e. KeyArn in AWS case)? 
If we detect a key_id update through the plugin Status gRPC then we can regenerate the unix socket name.
WDYT?

ardaguclu: Plugin controller (or whatever starts the KMS plugin) will serve in a predefined socket path (`unix://...`). This must be deterministic. Otherwise, we can't send request Status endpoint to get key_id.

ardaguclu: @flavianmissi I've updated to use `key.Secret` instead of another custom fields. Please let me know what you think. 

note: No tests are done up to now.

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -324,6 +326,243 @@ func TestKeyController(t *testing.T) {
...
+				{Group: "", Resource: "secrets"},
+			},
+			kmsConfigHash: "new-config-hash-xyz", // Different hash for new key ARN
+			kmsKeyIDHash:  "key-id-hash-abcd1234",
+			initialObjects: []runtime.Object{
+				encryptiontesting.CreateDummyKubeAPIPod("kube-apiserver-1", "kms", "node-1"),
+				func() *corev1.Secret {
+					s := encryptiontesting.CreateEncryptionKeySecretNoDataWithMode("kms", []schema.GroupResource{{Group: "", Resource: "secrets"}}, 1, "kms")
+					s.Annotations["encryption.apiserver.operator.openshift.io/kms-config-hash"] = "old-config-hash-1234"
+					s.Annotations["encryption.apiserver.operator.openshift.io/kms-key-id-hash"] = "key-id-hash-abcd1234"
```
ardaguclu: redundant

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -324,6 +326,243 @@ func TestKeyController(t *testing.T) {
...
+				{Group: "", Resource: "secrets"},
+			},
+			kmsConfigHash: "config-hash-12345678",
+			kmsKeyIDHash:  "new-key-id-hash-xyz",
+			initialObjects: []runtime.Object{
+				encryptiontesting.CreateDummyKubeAPIPod("kube-apiserver-1", "kms", "node-1"),
+				func() *corev1.Secret {
+					s := encryptiontesting.CreateEncryptionKeySecretNoDataWithMode("kms", []schema.GroupResource{{Group: "", Resource: "secrets"}}, 1, "kms")
+					s.Annotations["encryption.apiserver.operator.openshift.io/kms-config-hash"] = "config-hash-12345678"
+					s.Annotations["encryption.apiserver.operator.openshift.io/kms-key-id-hash"] = "old-key-id-hash-abc"
```
ardaguclu: redundant

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -265,40 +282,70 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+// defaultGetKMSHashes is the default implementation of getting KMS hashes
+// It calls the real KMS client to get the status and compute hashes
+func defaultGetKMSHashes(ctx context.Context, kmsConfig *configv1.KMSConfig) (string, []byte, error) {
+	// Generate unix socket path from KMS config and get the hash
+	socketPath, configHash, err := kms.GenerateUnixSocketPath(kmsConfig)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to generate KMS unix socket path: %w", err)
+	}
+
+	kmsClient, err := kms.NewKMSClient(socketPath)
```
flavianmissi: I think depending on the kms plugin to generate the unix socket path creates a fundamental chicken-egg problem where the unix socket path is generated based on a gRPC call to the plugin's Status endpoint, but the plugin requires a unix socket path before it can run. Unless I'm missing something crucial, this simply cannot work... right?

My suggestion is that, for dev preview, we just use the KeyArn from the APIServer config as the key_id. This won't be an issue with the aws kms plugin because the key_id is always matches the KeyArn. This is okay for the aws kms plugin because it does not change the key_id when the key is rotated, so this approach is only good for dev preview. We can refactor it later to something more robust.

What do y'all think?

ardaguclu: > I think depending on the kms plugin to generate the unix socket path creates a fundamental chicken-egg problem where the unix socket path is generated based on a gRPC call to the plugin's Status endpoint, but the plugin requires a unix socket path before it can run. Unless I'm missing something crucial, this simply cannot work... right?

Unix path is generated based on the hashsum of kmsconfig. It does not include key_id retrieved from Status endpoint. Otherwise, it would indeed be a chicken-egg problem.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: I believe that latest version of this PR incorporated these suggested changes.

PR #2044, opened 2025-11-03
STOR-2627: Re-apply CSIDriver after disabled fields were cleared

jsafrane:
When the API server clears disabled CSIDriver alpha/beta fields, `ApplyCSIDriver` should retry to apply those on the next call. We assume that the caller has evaluated `MutableCSINodeAllocatableCount` feature gate and sets the cleared field only when the feature gate is enabled (see [here](https://github.com/openshift/csi-operator/pull/435/)). The feature gate has not yet reached the kube-apiserver yet, so the API server still clears the field. But the API server will recognize it eventually and it will save the full CSIDriver.

It works well with StaticResourceController, which has forced 1 minute re-sync. So the operator will retry applying CSIDriver every 1 minute, until the API server stops clearing the field.

### DIFF
--- a/pkg/operator/resource/resourceapply/storage.go
+++ b/pkg/operator/resource/resourceapply/storage.go
@@ -135,9 +135,10 @@ func storageClassNeedsRecreate(oldSC, newSC *storagev1.StorageClass) bool {
 	return false
 }
 
-// ApplyCSIDriver merges objectmeta, does not worry about anything else
+// ApplyCSIDriver merges objectmeta and tries to update spec if any of the required fields were cleared by the API server.
+// It assumes they were cleared due to a feature gate not enabled in the API server and it will be enabled soon.
+// When used by StaticResourceController, it will retry periodically and eventually save the spec with the field.
 func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter, recorder events.Recorder, requiredOriginal *storagev1.CSIDriver) (*storagev1.CSIDriver, bool, error) {
-
 	required := requiredOriginal.DeepCopy()
 	if required.Annotations == nil {
 		required.Annotations = map[string]string{}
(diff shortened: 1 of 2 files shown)


pkg/operator/resource/resourceapply/storage.go
```diff
@@ -173,31 +174,57 @@ func ApplyCSIDriver(ctx context.Context, client storageclientv1.CSIDriversGetter
...
+	// Most CSIDriver fields are immutable. Any change to them should trigger Delete() + Create() calls.
+	needsRecreate := false
+
 	existingCopy := existing.DeepCopy()
-	resourcemerge.EnsureObjectMeta(&metadataModified, &existingCopy.ObjectMeta, required.ObjectMeta)
+	// Metadata change should need just Update() call.
+	resourcemerge.EnsureObjectMeta(&needsUpdate, &existingCopy.ObjectMeta, required.ObjectMeta)
 
 	requiredSpecHash := required.Annotations[specHashAnnotation]
 	existingSpecHash := existing.Annotations[specHashAnnotation]
```
gnufied, line 186: So here if we recomputed hash using `existing` object, it won't match the `required` object's hash. Should we do that instead? It is not like computing hash of `CSIDriver` spec is expensive or anything?

And we could do that all the time IMO? Changing the value IIRC doesn't require kubelet or external-attacher restart iirc

jsafrane: `existingHash` is `requiredHash` of a previous `ApplyCSIDriver` call. It is a hash of spec before it was sent to the API server. It will be different if the previous and the current `ApplyCSIDriver` got a different `required`. It does not say anything about the spec that is currently in the API server.

We can't really compare `existing.Spec` from the API server (or its hash) with `required.Spec`, because of defaulting. E.g `spec.requiresRepublish` in our operators is usually `nil`, but the API server defaults it to `false` and such comparison would detect a difference.

And from some reason CSIDriver does not have `metadata.generation` that would detect that user has changed the object `.spec` and it was not just defaulting.

gnufied: > because of defaulting. E.g spec.requiresRepublish in our operators is usually nil, but the API server defaults it to false and such comparison would detect a difference.

That seems fixable, but okay.  Lets consider this this comment as resolved. Do you want to add some tests?

jsafrane: Good catch, we need a test. I'll check if it's possible and how ugly it would be.

jsafrane: Unit test added

PR #2045, opened 2025-11-05
NO-JIRA: Introduce KMS encryption mode into encryption controllers

ardaguclu:
This PR adds KMS as a new mode in encryption controllers to be functioning along side with the other modes. Functionality has been proved to be working https://github.com/openshift/library-go/pull/2041. 

This PR takes the minimal bits from https://github.com/openshift/library-go/pull/2041 (because PoC PR also contains kms plugin communication to obtain the key_id. However, we decided that initially we don't need to track key_id).

Basically, idea is to track the hash of kmsconfig to detect any configuration changes. So that key_controller can create new secret to initiate migration. KMS key secret will not contain any confidential data. It will simply store the hash value of key_id. But for now, we don't access KMS plugin in operator controllers. That means key_id hash value will stay static.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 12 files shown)


ardaguclu:
Fake bumps to test CI:
* https://github.com/openshift/cluster-kube-apiserver-operator/pull/1957
* https://github.com/openshift/cluster-openshift-apiserver-operator/pull/629 (hasn't been bumped to 1.34 yet)
* https://github.com/openshift/cluster-authentication-operator/pull/805 (hasn't been bumped to 1.34 yet)

ardaguclu:
/hold
Since we discussed with @flavianmissi about the changes, this PR is not ready for review anymore. We first need EP.

PR #2050, opened 2025-11-13
CNTRLPLANE-1724: Add OTE helper functions in test/library/ote

gangwgr:
Added  new OTE helper functions for common test scenarios:
 `WaitForClusterOperatorStatus` - Core functionality for testing                                                                                                                                                              
  - `WaitForClusterOperatorHealthy` - Convenience wrapper                                                                                                                                                                        
  - `GetClusterOperatorConditionStatus` - Status retrieval helper                                                                                                                                                                
  - Supporting helper functions for condition details and logging

### DIFF
--- a/test/ote/util.go
+++ b/test/ote/util.go
@@ -1,0 +1,392 @@
+package ote
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"time"
+
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	"github.com/openshift/library-go/test/library"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const (
+	// DefaultStableDelay is the time to wait after reaching healthy state before confirming stability
+	DefaultStableDelay = 100 * time.Second
+	// MaxConsecutiveErrors is the maximum number of consecutive errors allowed when polling ClusterOperator status
+	MaxConsecutiveErrors = 5
+	// StatusLogInterval determines how often to log status details (every Nth attempt)
+	StatusLogInterval = 3
+	// RolloutLogInterval determines how often to log rollout progress (every Nth attempt)
+	RolloutLogInterval = 4
+)
+
+var (
+	// HealthyConditions defines the standard healthy state for a ClusterOperator
+	HealthyConditions = map[string]string{
+		"Available":   "True",
+		"Progressing": "False",
+		"Degraded":    "False",
+	}
+
+	// StandardConditionTypes is the ordered list of standard ClusterOperator condition types
+	StandardConditionTypes = []string{"Available", "Progressing", "Degraded"}
+)
+
+// WaitForClusterOperatorStatus waits for a ClusterOperator to reach the expected status conditions.
+// Example usage:
+//
+//	err := WaitForClusterOperatorStatus(ctx, t, coClient, "kube-apiserver",
+//	    map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"},
+//	    10*time.Minute, 1.0)
+func WaitForClusterOperatorStatus(ctx context.Context, t library.LoggingT, coClient configv1client.ClusterOperatorInterface, coName string, expectedStatus map[string]string, timeout time.Duration, waitMultiplier float64) error {
+	stableDelay := DefaultStableDelay
+
+	// Apply timeout multiplier
+	if waitMultiplier != 1.0 {
+		timeout = time.Duration(float64(timeout) * waitMultiplier)
+		t.Logf("Adjusted timeout for cluster type: %v (multiplier: %.2f)", timeout, waitMultiplier)
+	}
+
+	t.Logf("Waiting for ClusterOperator %s to reach status %v (timeout: %v)", coName, expectedStatus, timeout)
+
+	attempt := 0
+	consecutiveErrors := 0
+	maxConsecutiveErrors := MaxConsecutiveErrors
+	var lastStatus map[string]string
+	var lastConditionDetails map[string]conditionDetail
+
+	errCo := wait.PollUntilContextTimeout(ctx, 20*time.Second, timeout, false, func(pollCtx context.Context) (bool, error) {
+		attempt++
+		gottenStatus, conditionDetails, err := getClusterOperatorConditionStatusWithDetails(pollCtx, coClient, coName, expectedStatus)
+		if err != nil {
+			consecutiveErrors++
+			t.Logf("[Attempt %d] Error getting ClusterOperator status: %v (consecutive errors: %d)", attempt, err, consecutiveErrors)
+			// Fail fast if we hit too many consecutive errors
+			if consecutiveErrors >= maxConsecutiveErrors {
+				return false, fmt.Errorf("too many consecutive errors (%d) getting ClusterOperator status: %w", consecutiveErrors, err)
+			}
+			return false, nil
+		}
+		consecutiveErrors = 0
+
+		// Log detailed status changes
+		if !reflect.DeepEqual(lastStatus, gottenStatus) && lastStatus != nil {
+			logConditionChanges(t, attempt, coName, lastConditionDetails, conditionDetails)
+		}
+		lastStatus = gottenStatus
+		lastConditionDetails = conditionDetails
+
+		if reflect.DeepEqual(expectedStatus, gottenStatus) {
+			// Check if this is the stable healthy state
+			isHealthyState := reflect.DeepEqual(expectedStatus, HealthyConditions)
+			if isHealthyState {
+				// For True/False/False, wait some additional time and double check to ensure it is stably healthy
+				t.Logf("[Attempt %d] ClusterOperator %s reached healthy state, waiting %v for stability check", attempt, coName, stableDelay)
+				time.Sleep(stableDelay)
+
+				gottenStatus, conditionDetails, err := getClusterOperatorConditionStatusWithDetails(pollCtx, coClient, coName, expectedStatus)
+				if err != nil {
+					t.Logf("Error during stability check: %v", err)
+					return false, nil
+				}
+
+				if reflect.DeepEqual(expectedStatus, gottenStatus) {
+					t.Logf("ClusterOperator %s is stably available/non-progressing/non-degraded", coName)
+					return true, nil
+				}
+				t.Logf("ClusterOperator %s became unstable during stability check", coName)
+				logConditionDetails(t, conditionDetails)
+				return false, nil
+			} else {
+				t.Logf("[Attempt %d] ClusterOperator %s reached expected status: %v", attempt, coName, gottenStatus)
+				return true, nil
+			}
+		}
+
+		// Log progress every StatusLogInterval attempts with detailed condition info
+		if attempt%StatusLogInterval == 0 {
+			t.Logf("[Attempt %d] ClusterOperator %s status check:", attempt, coName)
+			logConditionDetails(t, conditionDetails)
+			t.Logf("  Expected: %v", expectedStatus)
+		}
+		return false, nil
+	})
+
+	if errCo != nil {
+		t.Logf("Failed waiting for ClusterOperator %s to reach expected status", coName)
+		if lastConditionDetails != nil {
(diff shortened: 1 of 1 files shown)


test/ote/util.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+// WaitForClusterOperatorStatus waits for a ClusterOperator to reach the expected status conditions.
+// Example usage:
+//
+//	err := WaitForClusterOperatorStatus(t, coClient, "kube-apiserver",
+//	    map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"},
+//	    10*time.Minute, 1.0)
+func WaitForClusterOperatorStatus(t library.LoggingT, coClient configv1client.ClusterOperatorInterface, coName string, expectedStatus map[string]string, timeout time.Duration, waitMultiplier float64) error {
+	stableDelay := 100 * time.Second
```
wangke19: Several hard-coded values should be constants.
 ```
stableDelay := 100 * time.Second  // Line 23
  maxConsecutiveErrors := 5          // Line 36
  if attempt%3 == 0 {               // Line 91
  if attempt%4 == 0 {               // Line 357
```

Suggestion:
```
const (
      DefaultStableDelay = 100 * time.Second
      MaxConsecutiveErrors = 5
      StatusLogInterval = 3  // log every 3rd attempt
      RolloutLogInterval = 4 // log every 4th attempt
  )
```

File: test/ote/util.go [hunk 2]
```diff
@@ -0,0 +1,374 @@
...
+	}
+}
+
+// GetClusterOperatorConditionStatus retrieves the current status values for specified conditionsof a ClusterOperator.
+// Example usage:
+//
+//	status, err := GetClusterOperatorConditionStatus(ctx, coClient, "kube-apiserver",
+//	map[string]string{"Available": "", "Progressing": "", "Degraded": ""})
+//	Returns: map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"}
+func GetClusterOperatorConditionStatus(ctx context.Context, coClient configv1client.ClusterOperatorInterface, coName string, statusToCompare map[string]string) (map[string]string, error) {
```
wangke19: The statusToCompare parameter name is misleading - it's actually used as a template for which conditions to query, not for comparison. Consider renaming:

```suggestion
func GetClusterOperatorConditionStatus(ctx context.Context, coClient configv1client.ClusterOperatorInterface, 
      coName string, conditionsToQuery map[string]string) (map[string]string, error)
```

test/ote/util.go
```diff
@@ -0,0 +1,374 @@
...
+//   - timeout: Maximum time to wait for rollout to complete
+//
+// Returns:
+//   - error if timeout is reached or an error occurs during polling
+//
+// Note:
+//   - All existing pods must be replaced by new pods created after this function is called
+//   - Supports both single-node and multi-node deployments
+func WaitForAPIServerRollout(t library.LoggingT, podClient corev1client.PodInterface, labelSelector string, timeout time.Duration) error {
+	rolloutStartTime := time.Now()
```
wangke19: Lines 295-300: Getting initial pods outside the poll loop could create timing issues:

  rolloutStartTime := time.Now()  // Line 281
  // ... get initial pods ...

  If pods start rolling out between time.Now() and the first poll iteration, you might miss the transition.

  Suggestion: Move rolloutStartTime to be captured atomically with initial pod state.

test/ote/util.go
```diff
@@ -0,0 +1,374 @@
...
+
+// logConditionDetails logs detailed information about all conditions
+func logConditionDetails(t library.LoggingT, details map[string]conditionDetail) {
+	// Sort condition types for consistent output
+	conditionTypes := []string{"Available", "Progressing", "Degraded"}
+
+	for _, condType := range conditionTypes {
+		if detail, ok := details[condType]; ok {
+			if detail.Status == "Unknown" {
+				t.Logf("  %s: %s (%s)", detail.Type, detail.Status, detail.Reason)
```
wangke19, line 192: Lines 166-169, 186-189, 204-207: Message truncation is repeated three times.

  Suggestion: Extract to helper:
```
  func truncateMessage(msg string, maxLen int) string {
      if len(msg) > maxLen {
          return msg[:maxLen-3] + "..."
      }
      return msg
  }
```

test/ote/util.go
```diff
@@ -0,0 +1,374 @@
...
+
+	t.Logf("Waiting for ClusterOperator %s to reach status %v (timeout: %v)", coName, expectedStatus, timeout)
+
+	attempt := 0
+	consecutiveErrors := 0
+	maxConsecutiveErrors := 5
+	var lastStatus map[string]string
+	var lastConditionDetails map[string]conditionDetail
+
+	errCo := wait.PollUntilContextTimeout(context.Background(), 20*time.Second, timeout, false, func(ctx context.Context) (bool, error) {
```
wangke19: Context Propagation: Lines 40, 119, 313: Consider accepting context as a parameter to WaitForClusterOperatorStatus and WaitForAPIServerRollout instead of creating new background
   contexts, allowing callers to control cancellation.

test/ote/util.go
```diff
@@ -0,0 +1,385 @@
...
+		if !reflect.DeepEqual(lastStatus, gottenStatus) && lastStatus != nil {
+			logConditionChanges(t, attempt, coName, lastConditionDetails, conditionDetails)
+		}
+		lastStatus = gottenStatus
+		lastConditionDetails = conditionDetails
+
+		eq := reflect.DeepEqual(expectedStatus, gottenStatus)
+		if eq {
+			// Check if this is the stable healthy state
+			isHealthyState := reflect.DeepEqual(expectedStatus, map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"})
```
wangke19: Define a variable for `map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"}`
```suggestion
			
			var HealthyConditions = map[string]string{
      "Available": "True",
      "Progressing": "False",
      "Degraded": "False",
  }

```

test/ote/util.go
```diff
@@ -0,0 +1,385 @@
...
+}
+
+// WaitForClusterOperatorHealthy is a convenience wrapper for waiting for a ClusterOperator
+// to reach the standard healthy state (Available=True, Progressing=False, Degraded=False).
+// Example usage:
+//
+//	err := WaitForClusterOperatorHealthy(ctx, t, coClient, "kube-apiserver", 10*time.Minute, 1.0)
+func WaitForClusterOperatorHealthy(ctx context.Context, t library.LoggingT, coClient configv1client.ClusterOperatorInterface, coName string, timeout time.Duration, waitMultiplier float64) error {
+	return WaitForClusterOperatorStatus(ctx, t, coClient, coName,
+		map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"},
```
wangke19: https://github.com/openshift/library-go/pull/2050/changes#r2664131931

test/ote/util.go
```diff
@@ -0,0 +1,385 @@
...
+
+// getClusterOperatorConditionStatusWithDetails retrieves detailed status information for specified conditions
+func getClusterOperatorConditionStatusWithDetails(ctx context.Context, coClient configv1client.ClusterOperatorInterface, coName string, conditionsToQuery map[string]string) (map[string]string, map[string]conditionDetail, error) {
+	co, err := coClient.Get(ctx, coName, metav1.GetOptions{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get ClusterOperator %s: %w", coName, err)
+	}
+
+	statusMap := make(map[string]string)
+	detailsMap := make(map[string]conditionDetail)
```
wangke19: Variable naming inconsistency
```suggestion
	status := make(map[string]string)
	details := make(map[string]conditionDetail)
```

test/ote/util.go
```diff
@@ -0,0 +1,385 @@
...
+				t.Logf("[Attempt %d] ClusterOperator %s reached healthy state, waiting %v for stability check", attempt, coName, stableDelay)
+				time.Sleep(stableDelay)
+
+				gottenStatus, conditionDetails, err := getClusterOperatorConditionStatusWithDetails(pollCtx, coClient, coName, expectedStatus)
+				if err != nil {
+					t.Logf("Error during stability check: %v", err)
+					return false, nil
+				}
+
+				eq := reflect.DeepEqual(expectedStatus, gottenStatus)
```
wangke19: The variable eq is only used once - inline it:
```suggestion
  if reflect.DeepEqual(expectedStatus, gottenStatus) {
      t.Logf("ClusterOperator %s is stably available/non-progressing/non-degraded", coName)
      return true, nil
  }

```

test/ote/util.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+
+	return statusMap, detailsMap, nil
+}
+
+// logConditionDetails logs detailed information about all conditions
+func logConditionDetails(t library.LoggingT, details map[string]conditionDetail) {
+	// Sort condition types for consistent output
+	conditionTypes := []string{"Available", "Progressing", "Degraded"}
```
wangke19: This is a constant list and could be a package-level constant for reuse.

test/ote/util.go
```diff
@@ -0,0 +1,392 @@
...
+//	status, err := GetClusterOperatorConditionStatus(ctx, coClient, "kube-apiserver",
+//	map[string]string{"Available": "", "Progressing": "", "Degraded": ""})
+//	Returns: map[string]string{"Available": "True", "Progressing": "False", "Degraded": "False"}
+func GetClusterOperatorConditionStatus(ctx context.Context, coClient configv1client.ClusterOperatorInterface, coName string, conditionsToQuery map[string]string) (map[string]string, error) {
+	co, err := coClient.Get(ctx, coName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ClusterOperator %s: %w", coName, err)
+	}
+
+	statusMap := make(map[string]string)
```
wangke19:
```suggestion
	status := make(map[string]string)
```

test/ote/util.go
```diff
@@ -0,0 +1,392 @@
+package ote
```
p0lyn0mial, line 1: did we copy the code from existing repo or is it brand new code ?

p0lyn0mial, line 1: I have just realized that these function could be also defined directly in the origin repo. Thoughts ?

p0lyn0mial, line 1: Which tests will use these functions ?

gangwgr: it is new code not existing. i created some new functions to use in component repo

gangwgr: I feel not needed there

gangwgr: Needed for this https://github.com/openshift/cluster-kube-apiserver-operator/pull/1988 and some qe tests cases which we later migrate in component repo

test/ote/util.go
```diff
@@ -0,0 +1,392 @@
...
+//   - labelSelector: Label selector to identify API server pods (e.g., "apiserver=true")
+//   - timeout: Maximum time to wait for rollout to complete
+//
+// Returns:
+//   - error if timeout is reached or an error occurs during polling
+//
+// Note:
+//   - All existing pods must be replaced by new pods created after this function is called
+//   - Supports both single-node and multi-node deployments
+func WaitForAPIServerRollout(ctx context.Context, t library.LoggingT, podClient corev1client.PodInterface, labelSelector string, timeout time.Duration) error {
```
p0lyn0mial, line 299: how does this function differ from https://github.com/openshift/library-go/blob/master/test/library/apiserver/apiserver.go#L36

why do we need both ?

gangwgr: WaitForAPIServerToStabilizeOnTheSameRevision (Existing Function)
  - Polls until all API server pods are on the same revision (e.g., all on revision 10)
  - Does NOT care when pods were created - just checks they're all on the same revision

  WaitForAPIServerRollout 
- Waits for ALL pods to be RECREATED with new pods
  - Captures the current pod creation times when called
  - Polls until ALL existing pods are replaced by new pods created after the function was called
  - All new pods must be in Running state
  - Cares about pod creation time - ensures every pod is new
   -When we need to guarantee that ALL pods have been recreated with fresh configuration (e.g., after secret rotation, config changes that require pod restart)
   
   Key Difference:
  WaitForAPIServerToStabilizeOnTheSameRevision:
  - Old pods on revision 10 → New pods on revision 10 (PASSES - same revision)

  WaitForAPIServerRollout:
  - Old pods created at 10:00am → Old pods on same revision 10 (FAILS - pods not recreated)
  - Old pods created at 10:00am → New pods created at 10:15am (PASSES - all pods recreated)

PR #2052, opened 2025-11-16
STOR-2523: Implement new controller: DeploymentVersionController

mpatlasov:
https://issues.redhat.com/browse/STOR-2523
/cc @openshift/storage

This controller updates "release.openshift.io/version" of the Deployment specified by controllerDeploymentName with the current version of the operator, read from the "OPERATOR_IMAGE_VERSION" environment variable. The controller expects this environment variable to be set to a reasonable value. The controller checks whether a rollout of the Deployment with the current version is in progress and updates "release.openshift.io/version" only when the deployment has completed. The controller relies on `.status.conditions` with the `Progressing` type to ensure that the deployment has completed, as explained in the Kubernetes docs: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment The controller expects the operator to publish the Deployment with "release.openshift.io/desired-version" equal to the current version of the operator. The lack of this annotation is treated as an old Deployment from some previous version of the operator that did not use this controller. Currently, this controller is supposed to be used only for Deployments residing in HyperShift control planes.

### DIFF
--- a/pkg/operator/hypershift/deploymentversion/version_controller.go
+++ b/pkg/operator/hypershift/deploymentversion/version_controller.go
@@ -72,6 +72,8 @@ func NewDeploymentVersionController(
 	}
 	return factory.New().WithSync(
 		c.Sync,
+	).WithSyncDegradedOnError(
+		operatorClient,
 	).WithInformers(
 		c.deploymentInformer.Informer(),
 	).ResyncEvery(


pkg/operator/hypershift/deploymentversion/version_controller.go
```diff
@@ -0,0 +1,180 @@
...
+	var actualVersion string
+	for k, v := range deployment.Annotations {
+		if k == desiredVersionAnnotation {
+			klog.V(4).Infof("DeploymentVersionController: desiredVersion found: %s", v)
+			desiredVersion = v
+		} else if k == versionAnnotation {
+			klog.V(4).Infof("DeploymentVersionController: actualVersion found: %s", v)
+			actualVersion = v
+		}
+	}
```
jsafrane: Why not this? :

```suggestion
    desiredVersion := deployment.Annotations[desiredVersionAnnotation]
    actualVersion := deployment.Annotations[versionAnnotation)
```

pkg/operator/hypershift/deploymentversion/version_controller_test.go
```diff
@@ -0,0 +1,326 @@
...
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			// Prepare OPERATOR_IMAGE_VERSION env var for DeploymentVersionController
+			var version string
+			if test.envVarVersion == "" {
+				version = theVersion
+			} else {
+				version = test.envVarVersion
+			}
+			os.Setenv(operatorImageVersionEnvVarName, version)
```
jsafrane:
```suggestion
			t.Setenv(operatorImageVersionEnvVarName, version)
```

pkg/operator/hypershift/deploymentversion/version_controller_test.go
```diff
@@ -0,0 +1,326 @@
...
+		{
+			name:               "test sync: normal workflow: version update is expected (stale version set)",
+			inputDeployment:    withOldVersionAnnotationDeployment,
+			expectedDeployment: withVersionAnnotationDeployment,
+		},
+		{
+			name:               "test sync: desiredVersion mismatch",
+			inputDeployment:    withVersionAnnotationDeployment,
+			expectedDeployment: withVersionAnnotationDeployment,
+			envVarVersion:      newVersion,
```
jsafrane, line 169: This case troubles me a bit. The Deployment misses the desired version annotation and the controller is silent about it (except for logs). Someone intentionally created this controller and they probably expect to get an error if something is off.

What if we make the controller Degraded on such error? And all errors that are returned from `Sync()`?
After some inertia, it will propagate to ClusterOperator degraded condition and the user will see the problem.

mpatlasov: This is not "error" test case. This case is only to discard old stale Deployment objects from consideration. Another controller is responsible for publishing current up-to-date Deployment, so the situation when we observe "wrong" object must be transient. That's completely fine that we are silent (except `klog.V(4).Infof()`) in this case, imho.

pkg/operator/hypershift/deploymentversion/version_controller.go
mpatlasov: fixed

pkg/operator/hypershift/deploymentversion/version_controller_test.go
mpatlasov: fixed

pkg/operator/hypershift/deploymentversion/version_controller_test.go
mpatlasov: Resolved (added WithSyncDegradedOnError() to degrade controller on errors)

pkg/operator/hypershift/deploymentversion/version_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	controllerDeploymentName string
+	deploymentInformer       appsinformersv1.DeploymentInformer
+	operatorClient           v1helpers.OperatorClientWithFinalizers
+	controlPlaneKubeClient   kubernetes.Interface
+	eventRecorder            events.Recorder
+}
+
+func NewDeploymentVersionController(
+	name string,
+	controlPlaneNamespace string,
```
p0lyn0mial, line 57: Sorry for the late response.

OK. I’d say that his controller runs on HCP because it uses `controlPlaneNamespace` and `controllerDeploymentName`

pkg/operator/hypershift/deploymentversion/version_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	controllerDeploymentName string,
+	deploymentInformer appsinformersv1.DeploymentInformer,
+	operatorClient v1helpers.OperatorClientWithFinalizers,
+	controlPlaneKubeClient kubernetes.Interface,
```
p0lyn0mial, line 61: `controlPlaneKubeClient` is misleading, because you pass the namespace anyway (`c.controlPlaneKubeClient.AppsV1().Deployments(c.controlPlaneNamespace).Update(...)`), so to me it’s a `managementClusterKubeClient`

pkg/operator/hypershift/deploymentversion/version_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	operatorClient           v1helpers.OperatorClientWithFinalizers
+	controlPlaneKubeClient   kubernetes.Interface
+	eventRecorder            events.Recorder
+}
+
+func NewDeploymentVersionController(
+	name string,
+	controlPlaneNamespace string,
+	controllerDeploymentName string,
+	deploymentInformer appsinformersv1.DeploymentInformer,
```
p0lyn0mial, line 59: `deploymentInformer` is also misleading because it also points to the management cluster (`managementClusterDeploymentInformer`)

pkg/operator/hypershift/deploymentversion/version_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	operatorClient v1helpers.OperatorClientWithFinalizers,
```
p0lyn0mial, line 60: on which cluster is `operatorClient` stored?

mpatlasov: > on which cluster is operatorClient stored?

It's in Guest cluster.

PR #2053, opened 2025-11-18
OCPBUGS-62422: staticpod/prune: Remove tmp cert file pruning

tchap:
This is no longer needed as pruning happens in relevant sync looks that actually write these certificates atomically.

### DIFF
--- a/pkg/operator/staticpod/controller/prune/prune_controller.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller.go
@@ -51,7 +51,6 @@ const (
 func NewPruneController(
 	targetNamespace string,
 	podResourcePrefix string,
-	certDir string,
 	command []string,
 	configMapGetter corev1client.ConfigMapsGetter,
 	podGetter corev1client.PodsGetter,
@@ -62,7 +61,6 @@ func NewPruneController(
 	c := &PruneController{
 		targetNamespace:   targetNamespace,
 		podResourcePrefix: podResourcePrefix,
-		certDir:           certDir,
 		command:           command,
 
 		operatorClient:  operatorClient,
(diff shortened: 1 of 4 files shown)


tchap:
I am marking this as

/hold

to actually test this somewhere.

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -51,7 +51,6 @@ const (
 func NewPruneController(
 	targetNamespace string,
 	podResourcePrefix string,
-	certDir string,
```
tchap, line 54: This is no longer needed, but it also breaks the API 🤷🏻‍♂️

tchap:
Well, I am not sure extra testing is worth it as we are only removing stuff. The WIP PR I created only contains `Nothing to prune` log entries.

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -58,7 +55,8 @@ func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.IntSliceVar(&o.ProtectedRevisions, "protected-revisions", o.ProtectedRevisions, "list of revision IDs to preserve (not delete)")
 	fs.StringVar(&o.ResourceDir, "resource-dir", o.ResourceDir, "directory for all files supporting the static pod manifest")
 	fs.StringVar(&o.StaticPodName, "static-pod-name", o.StaticPodName, "name of the static pod")
-	fs.StringVar(&o.CertDir, "cert-dir", o.CertDir, "directory for all certs")
+	fs.String("cert-dir", "", "directory for all certs")
```
p0lyn0mial: since we are removing the CertDir why do we need this ?

pkg/operator/staticpod/prune/cmd.go
```diff
@@ -58,7 +55,8 @@ func (o *PruneOptions) AddFlags(fs *pflag.FlagSet) {
...
+	fs.MarkDeprecated("cert-dir", "certificate directory is now being pruned in relevant sync loops") //nolint:errcheck
```
p0lyn0mial: same question here.

pkg/operator/staticpod/prune/cmd.go
tchap: Yup, not sure what I was thinking at that time, I was not sure how the pod is spawned, I guess, but since the controller is spawning this and the flag is removed there, we can remove it completely. Thanks.

PR #2058, opened 2025-11-24
OCPBUGS-62517: Fix DeploymentController to comply with OpenShift Available API contract

jianzhangbjz:
The `DeploymentController` was violating the OpenShift API contract which states: '[A component must not report Available=False during the course of a normal upgrade'.](https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L156)
This fix ensures Available remains True during normal rolling updates and only goes False for actual failures.
Assisted-by: Claude code

### DIFF
--- a/pkg/operator/deploymentcontroller/deployment_controller.go
+++ b/pkg/operator/deploymentcontroller/deployment_controller.go
@@ -268,20 +268,33 @@ func (c *DeploymentController) syncManaged(ctx context.Context, opSpec *opv1.Ope
 		})
 
 	// Set Available condition
+	// Per OpenShift API contract (config/v1/types_cluster_operator.go:156):
+	// "A component must not report Available=False during the course of a normal upgrade."
+	// Available should only be False when the deployment has actually failed, not during
+	// normal rolling updates or when temporarily waiting for pods to become ready.
 	if slices.Contains(c.conditions, opv1.OperatorStatusTypeAvailable) {
 		availableCondition := applyoperatorv1.
 			OperatorCondition().WithType(c.instanceName + opv1.OperatorStatusTypeAvailable)
+
 		if deployment.Status.AvailableReplicas > 0 {
+			// Deployment has available replicas - clearly available
 			availableCondition = availableCondition.
 				WithStatus(opv1.ConditionTrue).
 				WithMessage("Deployment is available").
 				WithReason("AsExpected")
-
-		} else {
+		} else if isDeploymentAvailableDuringRollout(deployment) {
+			// Deployment is progressing normally (rolling update, initial deployment)
+			// Per API contract, remain Available=True during normal operations
 			availableCondition = availableCondition.
-				WithStatus(opv1.ConditionFalse).
+				WithStatus(opv1.ConditionTrue).
 				WithMessage("Waiting for Deployment").
 				WithReason("Deploying")
+		} else {
+			// Deployment appears to have failed - no available replicas and not progressing normally
+			availableCondition = availableCondition.
+				WithStatus(opv1.ConditionFalse).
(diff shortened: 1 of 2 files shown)


pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -268,20 +268,33 @@ func (c *DeploymentController) syncManaged(ctx context.Context, opSpec *opv1.Ope
...
 				WithStatus(opv1.ConditionTrue).
 				WithMessage("Deployment is available").
 				WithReason("AsExpected")
-
+		} else if isDeploymentAvailableDuringRollout(deployment) {
+			// Deployment is progressing normally (rolling update, initial deployment)
+			// Per API contract, remain Available=True during normal operations
+			availableCondition = availableCondition.
+				WithStatus(opv1.ConditionTrue).
+				WithMessage("Deployment is rolling out").
```
grokspawn: Maybe we want to set the message here "Waiting for Deployment" to be consistent with previous semantics and any tooling?

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -389,3 +402,95 @@ func hasFinishedProgressing(deployment *appsv1.Deployment) bool {
...
+		}
+	}
+
+	// Check if deployment is actively being updated (spec change being rolled out)
+	// This is the primary indicator of a "normal upgrade" in progress
+	if deployment.Generation != deployment.Status.ObservedGeneration {
+		// Spec has changed, deployment controller is working on rolling it out
+		// Per API contract, remain Available during this normal upgrade
+		return true
+	}
```
jsafrane, line 435: This comment + `return true` is wrong. Consider initial Deployment creation during cluster installation - the status is not populated yet (`ObservedGeneration` is 0) and there is no replica yet. The overall status of such operator is definitely not `Available = true`.

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -389,3 +402,95 @@ func hasFinishedProgressing(deployment *appsv1.Deployment) bool {
...
+			if cond.Status == corev1.ConditionTrue && cond.Reason == "NewReplicaSetAvailable" {
+				hasFinishedRollout = true
+			}
+		}
+	}
+
+	// If we're actively rolling out, remain Available per API contract
+	if isRollingOut {
+		return true
+	}
```
jsafrane, line 461: Again, during installation the status should be `Available = false`.

pkg/operator/deploymentcontroller/deployment_controller.go
```diff
@@ -389,3 +402,95 @@ func hasFinishedProgressing(deployment *appsv1.Deployment) bool {
...
+	if isProgressing {
+		return true
+	}
+
+	// Special case: brand new deployment with no status conditions yet
+	// This happens during initial deployment before Kubernetes has had a chance to update status
+	if len(deployment.Status.Conditions) == 0 && deployment.Status.ObservedGeneration == 0 {
+		// Deployment just created, is progressing normally
+		return true
+	}
```
jsafrane, line 491: This is just wrong.

1. The code is not reachable
2. The status should be `Available = False` during installation and `Available = true` during upgrade. This code has no clue about that.

jsafrane (commented):
I am afraid that this will need far more thoughts and work. And a lot of unit tests.

jianzhangbjz:
> Do you have correct PDBs set? Node drain should leave always at least one Deployment pod running, which keeps the controller Available.

Thanks for your info! Add PDB in https://github.com/operator-framework/operator-controller/pull/2362, closed this PR first.

PR #2060, opened 2025-11-27
NO-JIRA: Fix static pod pruning logic for non-contiguous set of revisions

grandeit:
#### Problem
The `PruneController` contains a logic bug in `revisionsToKeep()` that prevents pruning when the protected revision set is non-contiguous but spans from revision `1` to `LatestAvailableRevision`.

**Scenario that triggers the bug:**
Node has very old `LastFailedRevision: 5`
Cluster is now at `LatestAvailableRevision: 100`
Limits are `failedRevisionLimit: 5`, `succeededRevisionLimit: 5`
Protected set becomes {1,2,3,4,5,96,97,98,99,100} (10 revisions)

**The buggy logic sees:**
First element: `1`
Last element: `100`
Returns `keepAll = true` -> No pruning happens.
This causes **a lot** of `revision-status-* ConfigMaps` (and their owned `ConfigMaps`) to accumulate until a later failed revision eventually removes the first revision from the set.

#### Solution
Check if the set has exactly LatestAvailableRevision elements before triggering the keepAll optimization. This ensures that the set has no gaps and is in-fact contiguous.

#### Testing
Added test case: "prunes non-contiguous set (keeps 1-10 and 96-100, prunes 11-95)" that verifies:

Two nodes with `LastFailedRevision: 5` and `LastFailedRevision: 10`
`CurrentRevision: 100` on both nodes
`LatestAvailableRevision: 100`

Protected set: `{1,2,3,4,5,6,7,8,9,10,96,97,98,99,100}` (15 revisions)
Revisions `11 - 95` are pruned.
`keepAll` optimization does not trigger.

### DIFF
--- a/pkg/operator/staticpod/controller/prune/prune_controller.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller.go
@@ -135,7 +135,7 @@ func (c *PruneController) revisionsToKeep(status *operatorv1.StaticPodOperatorSt
 		}
 	}
 
-	if keep.Len() > 0 && sets.List(keep)[0] == 1 && sets.List(keep)[keep.Len()-1] == status.LatestAvailableRevision {
+	if keep.Len() > 0 && int32(keep.Len()) == status.LatestAvailableRevision && sets.List(keep)[0] == 1 && sets.List(keep)[keep.Len()-1] == status.LatestAvailableRevision {
 		return true, nil
 	}
 
--- a/pkg/operator/staticpod/controller/prune/prune_controller_test.go
+++ b/pkg/operator/staticpod/controller/prune/prune_controller_test.go
@@ -3,11 +3,12 @@ package prune
 import (
 	"context"
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
+
+	clocktesting "k8s.io/utils/clock/testing"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


grandeit:
Friendly ping to some active reviewers:
@p0lyn0mial @JoelSpeed @damdo 👋

JoelSpeed:
@grandeit Correct me if I'm wrong, but, doesn't the static pod controller here upgrade through sequential versions? 

Therefore if I have protected version 5, this implies to me that some pod is still stuck on this version, meanwhile, others are on a later version say 100. That stuck pod needs to upgrade through every version 5 through 100 to catch up. Therefore, if we were to prune the intermediate versions, that pod would become perma-stuck as we would no longer have the intermediate versions for it to iterate through?

tchap:
This looks correct, thanks!

/lgtm

pkg/operator/staticpod/controller/prune/prune_controller.go
```diff
@@ -135,7 +135,7 @@ func (c *PruneController) revisionsToKeep(status *operatorv1.StaticPodOperatorSt
 		}
 	}
 
-	if keep.Len() > 0 && sets.List(keep)[0] == 1 && sets.List(keep)[keep.Len()-1] == status.LatestAvailableRevision {
+	if keep.Len() > 0 && int32(keep.Len()) == status.LatestAvailableRevision && sets.List(keep)[0] == 1 && sets.List(keep)[keep.Len()-1] == status.LatestAvailableRevision {
 		return true, nil
```
bertinatto, line 139: could you add a comment documenting this logic?

bertinatto (commented):
Thanks for the PR! Could you create a [proof PR](https://github.com/openshift/library-go/blob/master/README.md?plain=1#L6) to exercise this code path?

The old code skipped pruning whenever the keep list started at 1 and ended at Latest, even if there was a gap (example: keep 1–10 and 96–100, Latest=100). This PR starts deleting that gap. I wonder if the old skip was accidental, or did we want to leave the middle alone... it seems like it was accidental, but I'd like to be extra cautious here.

PR #2061, opened 2025-11-27
NO-JIRA: certrotation: Fix using a wrong annotation value

tchap:
This is actually not affecting the logic, but a wrong error is returned.

Improved the unit test to check events and annotation keys.

### DIFF
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -188,7 +188,7 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 		return notBefore, notAfter, fmt.Sprintf("bad expiry: %q", notAfterString)
 	}
 	notBeforeString := annotations[CertificateNotBeforeAnnotation]
-	if len(notAfterString) == 0 {
+	if len(notBeforeString) == 0 {
 		return notBefore, notAfter, "missing notBefore"
 	}
 	notBefore, err = time.Parse(time.RFC3339, notBeforeString)
--- a/pkg/operator/certrotation/signer_test.go
+++ b/pkg/operator/certrotation/signer_test.go
@@ -2,75 +2,127 @@ package certrotation
 
 import (
 	"context"
-	clocktesting "k8s.io/utils/clock/testing"
+	"crypto/x509"
+	"encoding/pem"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/google/go-cmp/cmp"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	kubefake "k8s.io/client-go/kubernetes/fake"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
+	clocktesting "k8s.io/utils/clock/testing"
 
 	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
 func TestEnsureSigningCertKeyPair(t *testing.T) {
+	verifyActionsOnCreated := func(t *testing.T, actions []clienttesting.Action) {
+		t.Helper()
+		if len(actions) != 1 {
+			t.Fatalf("Expected 1 action, got %s", spew.Sdump(actions))
+		}
+
+		action := actions[0]
+		if !action.Matches("create", "secrets") {
+			t.Fatalf("Expected secret create action, got %s", spew.Sdump(action))
+		}
+	}
+
+	verifyActionsOnUpdated := func(t *testing.T, actions []clienttesting.Action) {
+		t.Helper()
+		if len(actions) != 1 {
+			t.Fatalf("Expected 1 action, got %s", spew.Sdump(actions))
+		}
+
+		action := actions[0]
+		if !action.Matches("update", "secrets") {
+			t.Fatalf("Expected secret update action, got %s", spew.Sdump(action))
+		}
+	}
+
+	verifyActionsEmpty := func(t *testing.T, actions []clienttesting.Action) {
+		t.Helper()
+		if len(actions) != 0 {
+			t.Fatalf("Expected no action, got %s", spew.Sdump(actions))
+		}
+	}
+
+	verifyControllerUpdatedSecret := func(t *testing.T, expected, actual bool) {
+		t.Helper()
+		if actual != expected {
+			t.Errorf("Expected controller updated secret flag to be %v, got %v", expected, actual)
+		}
+	}
+
+	verifySecret := func(t *testing.T, actual *corev1.Secret) {
+		t.Helper()
+
+		if certType, _ := CertificateTypeFromObject(actual); certType != CertificateTypeSigner {
+			t.Errorf("expected certificate type 'signer', got: %v", certType)
+		}
+		if len(actual.Data["tls.crt"]) == 0 || len(actual.Data["tls.key"]) == 0 {
+			t.Error(actual.Data)
+		}
+		verifyKeyPair(t, actual.Data["tls.crt"], actual.Data["tls.key"])
+
+		expectedAnnotationKeys := sets.New[string](
+			"auth.openshift.io/certificate-not-after",
+			"auth.openshift.io/certificate-not-before",
+			"auth.openshift.io/certificate-issuer",
+			"certificates.openshift.io/refresh-period",
+			"openshift.io/owning-component",
+		)
+		if keys := sets.KeySet(actual.Annotations); !keys.Equal(expectedAnnotationKeys) {
+			t.Errorf("Annotation keys don't match:\n%s", cmp.Diff(expectedAnnotationKeys, keys))
+		}
+
+		ownershipValue, found := actual.Annotations[annotations.OpenShiftComponent]
+		if !found {
+			t.Errorf("expected secret to have ownership annotations, got: %v", actual.Annotations)
+		}
+		if ownershipValue != "test" {
+			t.Errorf("expected ownership annotation to be 'test', got: %v", ownershipValue)
+		}
+		if len(actual.OwnerReferences) != 1 {
+			t.Errorf("expected to have exactly one owner reference")
+		}
+		if actual.OwnerReferences[0].Name != "operator" {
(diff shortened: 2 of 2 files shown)


pkg/operator/certrotation/signer.go
```diff
@@ -188,7 +188,7 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 		return notBefore, notAfter, fmt.Sprintf("bad expiry: %q", notAfterString)
 	}
 	notBeforeString := annotations[CertificateNotBeforeAnnotation]
-	if len(notAfterString) == 0 {
+	if len(notBeforeString) == 0 {
```
p0lyn0mial, line 191: are we missing a unit test :) ?

tchap: Of course you were going to say this 🙂 🥲

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,117 @@ package certrotation
...
 	"k8s.io/client-go/tools/cache"
+	clocktesting "k8s.io/utils/clock/testing"
 
 	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
 func TestEnsureSigningCertKeyPair(t *testing.T) {
+	verifyActionFunc := func(check func(t *testing.T, action clienttesting.Action)) func(t *testing.T, client *kubefake.Clientset, controllerUpdatedSecret bool) {
+		return func(t *testing.T, client *kubefake.Clientset, controllerUpdatedSecret bool) {
```
tchap: This code is just copied and refactored so that it can be used for both create and update.

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,117 @@ package certrotation
...
+				t.Error(actual.Data)
+			}
+
+			expectedAnnotationKeys := sets.New[string](
+				"auth.openshift.io/certificate-not-after",
+				"auth.openshift.io/certificate-not-before",
+				"auth.openshift.io/certificate-issuer",
+				"certificates.openshift.io/refresh-period",
+				"openshift.io/owning-component",
+			)
```
tchap: Added checking for annotation keys. Getting the precise values would require mocking time and what not, which is not currently possible. But it could be done if we changed some public functions.

pkg/operator/certrotation/signer_test.go
```diff
@@ -81,40 +123,48 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
...
+						annotations.OpenShiftComponent:            "test",
+					},
+				},
+				Type: corev1.SecretTypeTLS,
+				Data: map[string][]byte{"tls.crt": {}, "tls.key": {}},
+			},
+			RefreshOnlyWhenExpired: false,
+			verifyActions:          verifyActionsOnUpdated,
+			expectedEvents: []*corev1.Event{
+				{Reason: "SignerUpdateRequired", Message: `"signer" in "ns" requires a new signing cert/key pair: missing notBefore`},
```
tchap, line 191: This actually ensures the reason as fixed in this PR works now.

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,117 @@ package certrotation
...
+			if actual.OwnerReferences[0].Name != "operator" {
+				t.Errorf("expected owner reference to be 'operator', got %v", actual.OwnerReferences[0].Name)
+			}
+		}
+	}
+
+	verifyActionsOnCreated := verifyActionFunc(func(t *testing.T, action clienttesting.Action) {
+		t.Helper()
+		if !action.Matches("create", "secrets") {
+			t.Error(action)
```
p0lyn0mial: not `t.Fatal` ? does it make sense to continue the flow if we got unexpected action?

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,117 @@ package certrotation
...
+		}
+	})
+
+	verifyActionsOnUpdated := verifyActionFunc(func(t *testing.T, action clienttesting.Action) {
+		t.Helper()
+		if !action.Matches("update", "secrets") {
+			t.Error(action)
```
p0lyn0mial: not `t.Fatal` ? does it make sense to continue the flow if we got unexpected action?

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,117 @@ package certrotation
...
+
 	tests := []struct {
 		name string
 
 		initialSecret          *corev1.Secret
 		RefreshOnlyWhenExpired bool
 
-		verifyActions func(t *testing.T, client *kubefake.Clientset, controllerUpdatedSecret bool)
-		expectedError string
+		verifyActions  func(t *testing.T, client *kubefake.Clientset, controllerUpdatedSecret bool)
```
p0lyn0mial: would it make sense to change the signature to `[]func(t *testing.T, client *kubefake.Clientset, controllerUpdatedSecret bool)` ?

and then the usage would be:

`[verifyFn1, verifyFn2]`

pkg/operator/certrotation/signer_test.go
tchap: Maybe not, I will change that.

pkg/operator/certrotation/signer_test.go
tchap: I will rather rename it to `verifyAction`, because there is at most one.

tchap: Not sure I am making it better or worse now, but I updated it to use `verifyAction`. Not specifying any function means that no action is expected.

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,99 @@ package certrotation
...
 	clienttesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/cache"
+	clocktesting "k8s.io/utils/clock/testing"
 
 	"github.com/openshift/api/annotations"
 	"github.com/openshift/library-go/pkg/operator/events"
 )
 
 func TestEnsureSigningCertKeyPair(t *testing.T) {
+	newVerifyActionFunc := func(check func(t *testing.T, action clienttesting.Action)) func(t *testing.T, action clienttesting.Action) {
```
p0lyn0mial: could we change the signature of this method to accept the required object ?
in that case this function would be responsible for validating that object.

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,99 @@ package certrotation
...
+				t.Errorf("expected owner reference to be 'operator', got %v", actual.OwnerReferences[0].Name)
+			}
+		}
+	}
+
+	verifyActionOnCreated := newVerifyActionFunc(func(t *testing.T, action clienttesting.Action) {
+		t.Helper()
+		if !action.Matches("create", "secrets") {
+			t.Fatalf("Expected a Create action, got %s", spew.Sdump(action))
+		}
```
p0lyn0mial, line 56: then in this function we would validate the action, extract the object and call `newVerifyActionFunc` (we could rename the function to newVerifyObjectFunc`)

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,99 @@ package certrotation
...
+	})
+
+	verifyActionOnUpdated := newVerifyActionFunc(func(t *testing.T, action clienttesting.Action) {
+		t.Helper()
+		if !action.Matches("update", "secrets") {
+			t.Fatalf("Expected an Update action, got %s", spew.Sdump(action))
+		}
```
p0lyn0mial, line 63: same here.

pkg/operator/certrotation/signer_test.go
```diff
@@ -208,7 +228,41 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
 				t.Errorf("missing %q", test.expectedError)
 			}
 
-			test.verifyActions(t, client, updated)
+			actions := client.Actions()
+			if test.verifyAction == nil {
+				if len(actions) > 0 {
```
p0lyn0mial: could we change this code to the previous version where we used the `verifyEmptyAction` function that was used on the test cases

pkg/operator/certrotation/signer_test.go
```diff
@@ -208,7 +228,41 @@ func TestEnsureSigningCertKeyPair(t *testing.T) {
...
+				}
+
+				if len(actions) > 1 {
+					t.Fatalf("Expected 1 action, got %d:\n%s", len(actions), spew.Sdump(actions))
+				} else {
+					test.verifyAction(t, actions[0])
+				}
+			}
+
+			if events := pruneEventFieldsForComparison(recorder.Events()); !cmp.Equal(events, test.expectedEvents) {
```
p0lyn0mial, line 284: we usually don't verify the events.
is this the only way to verify the new test cases ?

UPDATE: I just saw https://github.com/openshift/library-go/pull/2061#discussion_r2568729869

not ideal but changing it would require more work.

pkg/operator/certrotation/signer_test.go
```diff
@@ -2,75 +2,99 @@ package certrotation
...
+		return func(t *testing.T, action clienttesting.Action) {
+			t.Helper()
+			check(t, action)
+
+			// We can use CreateAction for all cases as UpdateAction is the very same interface.
+			actual := action.(clienttesting.CreateAction).GetObject().(*corev1.Secret)
+			if certType, _ := CertificateTypeFromObject(actual); certType != CertificateTypeSigner {
+				t.Errorf("expected certificate type 'signer', got: %v", certType)
+			}
+			if len(actual.Data["tls.crt"]) == 0 || len(actual.Data["tls.key"]) == 0 {
```
p0lyn0mial: I know this is pre-existing (we don't have to change it now) but I would expect the tests to at least check if the cert and the key are valid.

pkg/operator/certrotation/signer_test.go
p0lyn0mial: why have you decided to remove the validation of the `updated` var?

pkg/operator/certrotation/signer_test.go
p0lyn0mial: ic, normally i would prefer to create a valid secret and then remove the `notBefore` annotation but i understand that would require more work.

tchap:
^^^

/hold

pkg/operator/certrotation/signer_test.go
tchap: It's being validated again.

pkg/operator/certrotation/signer_test.go
tchap: I will actually add this, I had it implemented in #2062

PR #2062, opened 2025-11-28
NO-JIRA: certrotation/signer: Refactor tests to use mocks

tchap:
Aligned `signer.go` and the associated tests so that proper mocks can be used while testing. This also required some changes to `pkg/crypto`, but some additional public functions were only added, so there is no breaking change regarding that package API.

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -649,7 +649,7 @@ func MakeSelfSignedCAConfigForDuration(name string, caLifetime time.Duration) (*
 	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime)
 }
 
-func UnsafeMakeSelfSignedCAConfigForDurationAtTime(name string, currentTime func() time.Time, caLifetime time.Duration) (*TLSCertificateConfig, error) {
+func MakeSelfSignedCAConfigForDurationAtTime(name string, currentTime func() time.Time, caLifetime time.Duration) (*TLSCertificateConfig, error) {
 	subject := pkix.Name{CommonName: name}
 	return makeSelfSignedCAConfigForSubjectAndDuration(subject, currentTime, caLifetime)
 }
--- a/pkg/operator/certrotation/signer.go
+++ b/pkg/operator/certrotation/signer.go
@@ -221,7 +221,7 @@ func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *core
 // setSigningCertKeyPairSecret creates a new signing cert/key pair and sets them in the secret
 func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, clock clock.Clock, validity time.Duration) (*crypto.TLSCertificateConfig, error) {
 	signerName := fmt.Sprintf("%s_%s@%d", signingCertKeyPairSecret.Namespace, signingCertKeyPairSecret.Name, clock.Now().Unix())
-	ca, err := crypto.UnsafeMakeSelfSignedCAConfigForDurationAtTime(signerName, clock.Now, validity)
+	ca, err := crypto.MakeSelfSignedCAConfigForDurationAtTime(signerName, clock.Now, validity)
 	if err != nil {
 		return nil, err
 	}


pkg/operator/certrotation/signer.go
```diff
@@ -149,14 +153,14 @@ func ensureOwnerReference(meta *metav1.ObjectMeta, owner *metav1.OwnerReference)
 	return false
 }
 
-func needNewSigningCertKeyPair(secret *corev1.Secret, refresh time.Duration, refreshOnlyWhenExpired bool) (bool, string) {
+func needNewSigningCertKeyPair(secret *corev1.Secret, now time.Time, refresh time.Duration, refreshOnlyWhenExpired bool) (bool, string) {
```
p0lyn0mial: could this function accept `clock.Clock` ?

pkg/operator/certrotation/signer.go
```diff
@@ -212,9 +216,9 @@ func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *core
 }
 
 // setSigningCertKeyPairSecret creates a new signing cert/key pair and sets them in the secret
-func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, validity time.Duration) (*crypto.TLSCertificateConfig, error) {
-	signerName := fmt.Sprintf("%s_%s@%d", signingCertKeyPairSecret.Namespace, signingCertKeyPairSecret.Name, time.Now().Unix())
-	ca, err := crypto.MakeSelfSignedCAConfigForDuration(signerName, validity)
+func setSigningCertKeyPairSecret(signingCertKeyPairSecret *corev1.Secret, now time.Time, validity time.Duration) (*crypto.TLSCertificateConfig, error) {
+	signerName := fmt.Sprintf("%s_%s@%d", signingCertKeyPairSecret.Namespace, signingCertKeyPairSecret.Name, now.Unix())
+	ca, err := crypto.UnsafeMakeSelfSignedCAConfigForDurationAtTime(signerName, func() time.Time { return now }, validity)
```
p0lyn0mial: Unsafe ?!

i'd prefer changing `MakeSelfSignedCAConfigForDuration` to accept `clock.Clock`

pkg/crypto/crypto.go
```diff
@@ -676,14 +676,18 @@ func makeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, currentTime
 }
 
 func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA) (*TLSCertificateConfig, error) {
+	return MakeCAConfigForDurationAtTime(name, time.Now, caLifetime, issuer)
+}
+
+func MakeCAConfigForDurationAtTime(name string, currentTime func() time.Time, caLifetime time.Duration, issuer *CA) (*TLSCertificateConfig, error) {
```
p0lyn0mial: what is the purpose of this function? (atm used only by `MakeCAConfigForDuration`)

pkg/operator/certrotation/signer.go
```diff
@@ -58,6 +58,10 @@ type RotatedSigningCASecret struct {
 // EnsureSigningCertKeyPair manages the entire lifecycle of a signer cert as a secret, from creation to continued rotation.
 // It always returns the currently used CA pair, a bool indicating whether it was created/updated within this function call and an error.
 func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*crypto.CA, bool, error) {
+	return c.ensureSigningCertKeyPair(ctx, time.Now())
```
p0lyn0mial: I'd prefer changing `RotatedSigningCASecret` to accept `clock.Clock` and then perhaps this method could provide a default implementation if not set.

pkg/operator/certrotation/signer.go
```diff
@@ -201,8 +205,8 @@ func getValidityFromAnnotations(annotations map[string]string) (notBefore time.T
 
 // setSigningCertKeyPairSecretAndTLSAnnotations generates a new signing certificate and key pair,
 // stores them in the specified secret, and adds predefined TLS annotations to that secret.
-func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *corev1.Secret, validity, refresh time.Duration, tlsAnnotations AdditionalAnnotations) error {
-	ca, err := setSigningCertKeyPairSecret(signingCertKeyPairSecret, validity)
+func setSigningCertKeyPairSecretAndTLSAnnotations(signingCertKeyPairSecret *corev1.Secret, now time.Time, validity, refresh time.Duration, tlsAnnotations AdditionalAnnotations) error {
```
p0lyn0mial: could this function accept `clock.Clock` ?

p0lyn0mial (commented):
This approach requires changing the signatures of existing methods, which would have a ripple effect. 

I’m also fine with the previous approach: https://github.com/openshift/library-go/pull/2061

pkg/operator/certrotation/signer.go
tchap: I don't know why the method is called like that, makes no sense TBH.

pkg/crypto/crypto.go
tchap: Yeah, I got confused, I guess, I will remove it. Thanks.

pkg/operator/certrotation/signer.go
tchap: I wanted to use the same timestamp for all cert operations, but realistically it will be basically the same even when all the functions call Now on their own...

pkg/operator/certrotation/signer.go
tchap: I would remove the `Unsafe` prefix from the method.

tchap: I renamed the method in a separate commit. I am afraid larger changes may break other users. We can consider it in a separate PR, I would say.

pkg/crypto/crypto.go
```diff
@@ -649,7 +649,7 @@ func MakeSelfSignedCAConfigForDuration(name string, caLifetime time.Duration) (*
 	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime)
 }
 
-func UnsafeMakeSelfSignedCAConfigForDurationAtTime(name string, currentTime func() time.Time, caLifetime time.Duration) (*TLSCertificateConfig, error) {
```
p0lyn0mial, line 652: do we know why this method was considered unsafe ?

p0lyn0mial: now we have two functions that differ by one param. 
call sites of the unsafe function would have to be fixed due to the name change. 

why not to change the first one to accept clock.Clock and remove the other ? is there something wrong with this approach? (except fixing all references to this method).

pkg/operator/certrotation/signer.go
```diff
@@ -53,11 +54,16 @@ type RotatedSigningCASecret struct {
...
 	Client        corev1client.SecretsGetter
 	EventRecorder events.Recorder
+	Clock         clock.Clock
 }
 
 // EnsureSigningCertKeyPair manages the entire lifecycle of a signer cert as a secret, from creation to continued rotation.
 // It always returns the currently used CA pair, a bool indicating whether it was created/updated within this function call and an error.
 func (c RotatedSigningCASecret) EnsureSigningCertKeyPair(ctx context.Context) (*crypto.CA, bool, error) {
+	if c.Clock == nil {
+		c.Clock = clock.RealClock{}
```
p0lyn0mial, line 64: we should test the new behaviour.

PR #2063, opened 2025-12-07
OCPSTRAT-485: Add ConfigMap revision pruner for API server operators

grandeit:
Add the new controller ConfigMapRevisionPruneController that prunes old revisioned ConfigMaps no longer in use by operand pods. This complements the existing SecretRevisionPruneController.

Key features:
- Prunes ConfigMaps by prefix (e.g., revision-status-, installer-status-)
- Preserves the 5 most recent revisions per prefix independently
- Only prunes revisions below the minimum pod revision in use

Also adds WithRevisionPruneController to APIServerControllerSet, allowing API server operators to configure ConfigMap revision pruning alongside the existing secret pruning capability.

With this, it will be possible to add revision pruning to the `openshift-apiserver` and `oauth-apiserver` in the future. As described in [RFE-3492](https://issues.redhat.com/browse/RFE-3492) and [RFE-4143](https://issues.redhat.com/browse/RFE-4143).

### DIFF
--- a/pkg/operator/revisionpruner/prune_controller.go
+++ b/pkg/operator/revisionpruner/prune_controller.go
@@ -1,0 +1,154 @@
+package revisionpruner
+
+import (
+	"context"
+	"maps"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/operator/revisionpruner/prune_controller.go
bertinatto, line 1: Since the other package is called `secretspruner`, probably a better name here would be `configmappruner`

bertinatto (commented):
Hey @grandeit, thanks a lot for this PR!

The changes here seem OK, but I’d be in favor of consolidating this controller with the `secretspruner` controller. Both controllers share the same pruning criteria for secrets and ConfigMaps, so having a single pruner would eliminate duplicated logic. The only real difference between the controller introduced in this PR and the `secretpruner` controller is the resource type being pruned (Secret vs. ConfigMap); the rest of the logic is the same.

Additionally, both ConfigMap and Secret revisions are created by the same controller (`revisioncontroller`), so it would be more consistent to also have a single controller responsible for pruning them.

Thoughts?

PR #2064, opened 2025-12-09
resourceapply: add improved version which copies over spec.conversion.webhook.clientConfig.caBundle

chrischdi:
This PR adds an improved version of ApplyCustomResourceDefinitionV1, analog to the pattern used in other packages, named `ApplyCustomResourceDefinitionV1Improved`.

The only change is to also copy over the caBundle of a conversion webhook, to preserve an injected ca bundle and to not cause indefinite reconciliations between some controller and service-ca-bundle operator.

This was found for https://github.com/openshift/cluster-capi-operator which makes use of this library.

CVO already today does this: https://github.com/openshift/cluster-version-operator/blob/0e6c916f99e05983190202575bb530200560acb9/lib/resourcemerge/apiext.go#L34

I adjusted the code though to match what library-go does for MutatingWebhookConfigurations.

### DIFF
--- a/pkg/operator/resource/resourceapply/apiextensions.go
+++ b/pkg/operator/resource/resourceapply/apiextensions.go
@@ -13,6 +13,39 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// ApplyCustomResourceDefinitionV1Improved applies the required CustomResourceDefinition to the cluster, preserving an injected ca bundle.
+func ApplyCustomResourceDefinitionV1Improved(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, recorder events.Recorder, required *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, bool, error) {
+	existing, err := client.CustomResourceDefinitions().Get(ctx, required.Name, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		requiredCopy := required.DeepCopy()
+		actual, err := client.CustomResourceDefinitions().Create(
+			ctx, resourcemerge.WithCleanLabelsAndAnnotations(requiredCopy).(*apiextensionsv1.CustomResourceDefinition), metav1.CreateOptions{})
+		resourcehelper.ReportCreateEvent(recorder, required, err)
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := false
+	existingCopy := existing.DeepCopy()
+
+	copyCustomResourceDefinitionConversionWebhookCABundle(existing, required)
+	resourcemerge.EnsureCustomResourceDefinitionV1(&modified, existingCopy, *required)
+	if !modified {
+		return existing, false, nil
+	}
+
+	if klog.V(2).Enabled() {
+		klog.Infof("CustomResourceDefinition %q changes: %s", existing.Name, JSONPatchNoError(existing, existingCopy))
+	}
+
+	actual, err := client.CustomResourceDefinitions().Update(ctx, existingCopy, metav1.UpdateOptions{})
+	resourcehelper.ReportUpdateEvent(recorder, required, err)
+
+	return actual, true, err
+}
+
 // ApplyCustomResourceDefinitionV1 applies the required CustomResourceDefinition to the cluster.
 func ApplyCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, recorder events.Recorder, required *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, bool, error) {
 	existing, err := client.CustomResourceDefinitions().Get(ctx, required.Name, metav1.GetOptions{})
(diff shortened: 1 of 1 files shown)


pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -13,6 +13,39 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// ApplyCustomResourceDefinitionV1Improved applies the required CustomResourceDefinition to the cluster, preserving an injected ca bundle.
+func ApplyCustomResourceDefinitionV1Improved(ctx context.Context, client apiextclientv1.CustomResourceDefinitionsGetter, recorder events.Recorder, required *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, bool, error) {
```
p0lyn0mial, line 17: Please double-check but I think that the `*Improved` versions let callers reuse a cache to avoid unnecessary writes when nothing has changed.

p0lyn0mial, line 17: Please also add a unit test

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -13,6 +13,39 @@ import (
...
+	modified := false
+	existingCopy := existing.DeepCopy()
+
+	copyCustomResourceDefinitionConversionWebhookCABundle(existing, required)
+	resourcemerge.EnsureCustomResourceDefinitionV1(&modified, existingCopy, *required)
+	if !modified {
+		return existing, false, nil
+	}
+
+	if klog.V(2).Enabled() {
```
p0lyn0mial, line 39: why not to use  `klog.V(2).Infof(...)` ?

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -55,3 +88,29 @@ func DeleteCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1
 	resourcehelper.ReportDeleteEvent(recorder, required, err)
 	return nil, true, nil
 }
+
+// copyCustomResourceDefinitionConversionWebhookCABundle populates spec.conversion.webhook.clientConfig.caBundle fields from
+// existing resource if it was set before and is not set in present. This provides upgrade compatibility with service-ca-bundle operator.
+func copyCustomResourceDefinitionConversionWebhookCABundle(existing *apiextensionsv1.CustomResourceDefinition, required *apiextensionsv1.CustomResourceDefinition) {
```
p0lyn0mial, line 94: could you rename the params to `from` and `to` ? it would be easier to read the code.

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -13,6 +13,39 @@ import (
...
+		return actual, true, err
+	}
+	if err != nil {
+		return nil, false, err
+	}
+
+	modified := false
+	existingCopy := existing.DeepCopy()
+
+	copyCustomResourceDefinitionConversionWebhookCABundle(existing, required)
```
p0lyn0mial, line 33: you pass the original pointer to the `required` obj which is modified. you need to make a copy of that obj and pass the pointer to the copy

pkg/operator/resource/resourceapply/apiextensions.go
```diff
@@ -55,3 +88,29 @@ func DeleteCustomResourceDefinitionV1(ctx context.Context, client apiextclientv1
 	resourcehelper.ReportDeleteEvent(recorder, required, err)
 	return nil, true, nil
 }
+
+// copyCustomResourceDefinitionConversionWebhookCABundle populates spec.conversion.webhook.clientConfig.caBundle fields from
+// existing resource if it was set before and is not set in present. This provides upgrade compatibility with service-ca-bundle operator.
```
p0lyn0mial, line 93: since this function `provides upgrade compatibility with service-ca-bundle operator` why shouldn't the `ApplyCustomResourceDefinitionV1` receive this improvement ?

chrischdi: I thought about creating a separate function to not "surprisingly" change behavior of the existing function.

I'm okay with doing the adjustment to `ApplyCustomResourceDefinitionV1`

damdo: I think creating an "Improved" function has generally been the approach around here right?
To not surprise importers with a "breaking change" in behaviour.
That said if we think we are ok with that, +1

p0lyn0mial: I think that the *Improved versions also let callers reuse a cache to avoid unnecessary writes when nothing has changed. Usually by accepting the `cache ResourceCache` param.
 
 I'm ok with a new function but I think the new function should also take the `cache ResourceCache` param. 
 
 We should also consider deprecating the old function. WDYT?

damdo: Sounds reasonable thanks

pkg/operator/resource/resourceapply/apiextensions.go
bertinatto: That's right, `*Improved` variants typically use the [ResourceCache](https://github.com/openshift/library-go/blob/HEAD/pkg/operator/resource/resourceapply/resource_cache.go#L32-L35). While the suffix isn’t ideal, it’s probably best to continue following the existing convention for consistency. Let's try to come up with a different suffix for now (`ApplyCustomResourceDefinitionV1WithCABundle`?).

bertinatto (commented):
Could you also create a proof PR where you bump library-go in your operator to use this version?

chrischdi:
As I'm not at the company anymore, one of the team would need to takeover, cc @damdo

damdo:
Thanks @chrischdi for your work
Yes, someone else will take over in a bit

chrischdi:
/close

cc @damdo FYI :-)

damdo:
Thanks @chrischdi 
We dropped this as we refactored the installer and not use this anymore, TY!

PR #2067, opened 2025-12-24
Allow manifest selection based on feature gates

JoelSpeed:
This will allow CVO to start including/excluding manifests based on the enabled feature gates, as opposed to currently where we only include based on the enabled feature set.

This means that we will need fewer simul-merge PRs where we merge a feature gate promotion but then have to adjust manifests to change their feature gating level.

The annotation itself is `release.openshift.io/feature-gate` which takes a comma separated list of feature gates, each optionally prefix by `-` to indicate a negation (this gate must be disabled). The requirements are ANDed.

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -26,6 +26,7 @@ const (
 	CapabilityAnnotation  = "capability.openshift.io/name"
 	DefaultClusterProfile = "self-managed-high-availability"
 	featureSetAnnotation  = "release.openshift.io/feature-set"
+	featureGateAnnotation = "release.openshift.io/feature-gate"
 )
 
 var knownFeatureSets = sets.Set[string]{}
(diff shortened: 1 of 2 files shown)


pkg/manifest/manifest.go
```diff
@@ -202,7 +237,7 @@ func (m *Manifest) Include(excludeIdentifier *string, requiredFeatureSet *string
 // to capabilities filtering. When set to true a manifest will not be excluded simply because it contains an unknown
 // capability. This is necessary to allow updates to an OCP version containing newly defined capabilities.
 func (m *Manifest) IncludeAllowUnknownCapabilities(excludeIdentifier *string, requiredFeatureSet *string, profile *string,
-	capabilities *configv1.ClusterVersionCapabilitiesStatus, overrides []configv1.ComponentOverride, allowUnknownCapabilities bool) error {
+	capabilities *configv1.ClusterVersionCapabilitiesStatus, overrides []configv1.ComponentOverride, enabledFeatureGates sets.Set[string], allowUnknownCapabilities bool) error {
```
bertinatto, line 250: Note to self: this is used here: https://github.com/openshift/cluster-version-operator/blob/d245c43ce077fe452771d9eee3af4ca57e79459b/lib/manifest/manifest.go#L46

PR #2069, opened 2026-01-02
OCPSTRAT-2876: Manifest inclusion filtering based on major version

JoelSpeed:
Currently based on #2067 

This teaches the manifest inclusion to filter based on a major-version annotation. This will allow us to concurrently build multiple major versions from a single branch while folks can include manifests (in particular feature gates) that apply only to 4, or 5, or 6 etc

### DIFF
--- a/pkg/manifest/manifest.go
+++ b/pkg/manifest/manifest.go
@@ -235,13 +235,18 @@ func checkFeatureGates(enabledGates sets.Set[string], annotations map[string]str
 }
 
 // checkMajorVersion validates if manifest should be included based on major version requirements
-func checkMajorVersion(clusterMajorVersion uint64, annotations map[string]string) error {
+func checkMajorVersion(gvk schema.GroupVersionKind, clusterMajorVersion uint64, annotations map[string]string) error {
 	if annotations == nil {
 		return nil // No annotations, include by default
 	}
 	majorVersionRequirements, ok := annotations[majorVersionAnnotation]
 	if !ok {
 		return nil // No requirements, include by default
+	}
+
+	if !isFeatureGate(gvk) && !isCustomResource(gvk) {
+		// Has a requirement, but is not of the expected kind
+		return fmt.Errorf("major version filtering is only supported for feature gates and custom resources")
 	}
 
 	requirements := strings.Split(majorVersionRequirements, ",")
@@ -338,7 +343,7 @@ func (m *Manifest) IncludeAllowUnknownCapabilities(excludeIdentifier *string, re
 
 	// Major version filtering
 	if majorVersion != nil {
-		err := checkMajorVersion(*majorVersion, annotations)
+		err := checkMajorVersion(m.GVK, *majorVersion, annotations)
 		if err != nil {
 			return err
 		}
@@ -542,4 +547,12 @@ func addIfNotDuplicateResource(manifest Manifest, resourceIds map[resourceId]boo
 		return nil
 	}
 	return fmt.Errorf("duplicate resource: (%s)", manifest.id)
+}
+
+func isFeatureGate(gvk schema.GroupVersionKind) bool {
+	return gvk.Group == "config.openshift.io" && gvk.Kind == "FeatureGate"
+}
+
+func isCustomResource(gvk schema.GroupVersionKind) bool {
+	return gvk.Group == "apiextensions.k8s.io" && gvk.Kind == "CustomResourceDefinition"
 }
--- a/pkg/manifest/manifest_test.go
+++ b/pkg/manifest/manifest_test.go
@@ -1380,6 +1380,7 @@ func TestMajorVersionFiltering(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			manifest := createTestManifest(tt.annotations)
+			manifest.GVK = schema.GroupVersionKind{Group: "config.openshift.io", Kind: "FeatureGate"}
 
 			// Test through the Include method which calls checkMajorVersion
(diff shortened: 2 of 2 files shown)


JoelSpeed:
It is yes

I was having some back and forth last week with @wking and @everettraven about how we want this to look. I'm currently pausing on this for the moment while i focus on getting the CVO feature gate annotation work wrapped. But this is on the requirements list for 5.0 so I'll come back to this soon

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+		{
+			name:                "no major version annotation should include",
+			annotations:         map[string]string{},
+			clusterMajorVersion: ptr.To(uint64(4)),
+			expectIncluded:      true,
+		},
+		{
+			name:                "nil major version should include all",
+			annotations:         map[string]string{"release.openshift.io/major-version": "4"},
+			clusterMajorVersion: nil,
```
JianLi-RH, line 1233: awsome

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+		},
+		{
+			name:                  "single inclusion - non-matching version",
+			annotations:           map[string]string{"release.openshift.io/major-version": "5"},
+			clusterMajorVersion:   ptr.To(uint64(4)),
+			expectIncluded:        false,
+			expectedErrorContains: "major version 4 does not match any required versions",
+		},
+		{
+			name:                "multiple inclusions - first matches",
```
JianLi-RH, line 1250: awsome

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+			annotations:         map[string]string{"release.openshift.io/major-version": "4,5,6"},
+			clusterMajorVersion: ptr.To(uint64(4)),
+			expectIncluded:      true,
+		},
+		{
+			name:                "multiple inclusions - middle matches",
```
JianLi-RH, line 1256: awsome

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+			annotations:         map[string]string{"release.openshift.io/major-version": "3,4,5"},
+			clusterMajorVersion: ptr.To(uint64(4)),
+			expectIncluded:      true,
+		},
+		{
+			name:                "multiple inclusions - last matches",
```
JianLi-RH, line 1262: awsome

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+		},
+		{
+			name:                  "invalid exclusion version format",
+			annotations:           map[string]string{"release.openshift.io/major-version": "-abc"},
+			clusterMajorVersion:   ptr.To(uint64(4)),
+			expectIncluded:        false,
+			expectedErrorContains: "invalid excluded major version \"abc\" in annotation: strconv.ParseUint: parsing \"abc\": invalid syntax",
+		},
+		{
+			name:                "zero version",
```
JianLi-RH, line 1347: awsome

pkg/manifest/manifest_test.go
```diff
@@ -1211,3 +1211,302 @@ func TestManifestIncludeWithFeatureGates(t *testing.T) {
...
+						t.Errorf("Expected error to contain %q, but got: %v", tt.expectedErrorContains, err)
+					}
+				}
+			}
+		})
+	}
+}
+
+// TestMajorVersionFilteringInteractionWithOtherFilters tests that major version filtering works alongside other filters
+func TestMajorVersionFilteringInteractionWithOtherFilters(t *testing.T) {
```
JianLi-RH, line 1406: this is really great job!

pkg/manifest/manifest.go
```diff
@@ -232,12 +234,69 @@ func checkFeatureGates(enabledGates sets.Set[string], annotations map[string]str
...
+		}
+	}
+
+	switch {
+	case includedVersions.Intersection(excludedVersions).Len() > 0:
+		return fmt.Errorf("inclusion and exclusion requirements overlap: %v", includedVersions.Intersection(excludedVersions).UnsortedList())
+	case includedVersions.Has(clusterMajorVersion):
+		return nil // Found matching version, include this manifest
+	case excludedVersions.Has(clusterMajorVersion):
+		return fmt.Errorf("major version %d matches excluded version %d", clusterMajorVersion, clusterMajorVersion)
```
JianLi-RH, line 285: It seems the second variable in the error string should be `excludedVersions`.

I got logs:
```
2026-03-19T03:56:10.602708782Z I0319 03:56:10.602688       1 payload.go:217] excluding Filename: "0000_03_version-4-5-6-false.yaml" Group: "apiextensions.k8s.io" Kind: "CustomResourceDefinition" Name: "version-4-5-6.test.io": major version 4 matches excluded version 4
```

@JoelSpeed please help condifrm.

JoelSpeed: I think it is actually correct. From the case condition we know that the `clusterMajorVersion` is in the `excludedVersions` set, so even if we tried to change this to print out the excluded versions, it would just show us a list of excluded versions rather than a specific excluded version that matched.

The error message that is printed out is always correct, the 4 here was an excluded version

JianLi-RH: um, it looks strange, but anyway from the message we can know the version is excluded

PR #2076, opened 2026-01-08
OCPBUGS-65626: add service account to guard pod

ehearne-redhat:
This change adds a bespoke service account to the guard pod. It also handles service account cleanup, and includes additional fields in tests and basic service account testing.

The reason for the change is that we should opt to use a bespoke service account rather than default.

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -12,7 +12,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/policy/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -48,6 +47,8 @@ type GuardController struct {
 	podGetter  corev1client.PodsGetter
 	pdbGetter  policyclientv1.PodDisruptionBudgetsGetter
 	pdbLister  policylisterv1.PodDisruptionBudgetLister
+	saLister   corelisterv1.ServiceAccountLister
+	saGetter   corev1client.ServiceAccountsGetter
 
 	// installerPodImageFn returns the image name for the installer pod
 	installerPodImageFn   func() string
@@ -70,6 +71,7 @@ func NewGuardController(
 	operatorClient operatorv1helpers.StaticPodOperatorClient,
 	podGetter corev1client.PodsGetter,
 	pdbGetter policyclientv1.PodDisruptionBudgetsGetter,
+	saGetter corev1client.ServiceAccountsGetter,
 	eventRecorder events.Recorder,
 	createConditionalFunc func() (bool, bool, error),
 	extraNodeSelector labels.Selector,
@@ -102,6 +104,8 @@ func NewGuardController(
 		podGetter:                     podGetter,
 		pdbGetter:                     pdbGetter,
 		pdbLister:                     kubeInformersForTargetNamespace.Policy().V1().PodDisruptionBudgets().Lister(),
+		saGetter:                      saGetter,
+		saLister:                      kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Lister(),
 		installerPodImageFn:           getInstallerPodImageFromEnv,
 		createConditionalFunc:         createConditionalFunc,
 		extraNodeSelector:             extraNodeSelector,
@@ -116,6 +120,7 @@ func NewGuardController(
 		WithInformers(
 			kubeInformersForTargetNamespace.Core().V1().Pods().Informer(),
 			kubeInformersClusterScoped.Core().V1().Nodes().Informer(),
+			kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Informer(),
 		).
 		WithSync(c.sync).
 		WithSyncDegradedOnError(operatorClient).
@@ -178,6 +183,9 @@ var pdbTemplate []byte
 //go:embed manifests/guard-pod.yaml
 var podTemplate []byte
 
+//go:embed manifests/guard-pod-sa.yaml
+var serviceAccountTemplate []byte
+
 func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext) error {
 	klog.V(5).Info("Syncing guards")
 
@@ -227,10 +235,25 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			for _, pod := range pods {
 				_, _, err = resourceapply.DeletePod(ctx, c.podGetter, syncCtx.Recorder(), pod)
 				if err != nil {
-					klog.Errorf("Unable to delete Pod: %v", err)
+					klog.Errorf("unable to delete Pod: %v", err)
 					errs = append(errs, err)
 				}
 			}
+		}
+
+		// delete the guard service account
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if err == nil {
+			_, _, err = resourceapply.DeleteServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				errs = append(errs, err)
+			}
+		} else if !apierrors.IsNotFound(err) {
+			errs = append(errs, err)
 		}
 	} else {
 		nodes, err := c.nodeLister.List(c.masterNodesSelector)
@@ -275,7 +298,7 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 					return fmt.Errorf("Unable to apply PodDisruptionBudget changes: %v", err)
 				}
 			}
-		} else if errors.IsNotFound(err) {
+		} else if apierrors.IsNotFound(err) {
 			_, _, err = resourceapply.ApplyPodDisruptionBudget(ctx, c.pdbGetter, syncCtx.Recorder(), pdb)
 			if err != nil {
 				klog.Errorf("Unable to create PodDisruptionBudget: %v", err)
@@ -289,6 +312,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		operands := map[string]*corev1.Pod{}
 		for _, pod := range pods {
 			operands[pod.Spec.NodeName] = pod
+		}
+
+		// Create the guard service account once for all guard pods
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		// There are no secrets, no image pull secrets, no annotations.
+		// The guard pod uses automountServiceAccountToken: false.
+		// There's nothing meaningful to update. So update logic isn't needed right now.
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if apierrors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				return fmt.Errorf("unable to create service account %v for Guard Pods: %w", serviceAccount.Name, err)
+			}
+		} else if err != nil {
+			return err
 		}
 
 		for _, node := range nodes {
@@ -328,6 +369,7 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			pod.Spec.NodeName = node.Name
 			pod.Spec.Containers[0].Image = c.installerPodImageFn()
 			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Host = operands[node.Name].Status.PodIP
+			pod.Spec.ServiceAccountName = serviceAccount.Name
 			// The readyz port as string type is expected to be convertible into int!!!
 			readyzPort, err := strconv.Atoi(c.readyzPort)
 			if err != nil {
@@ -362,6 +404,10 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 				}
 				if actual.Status.Phase != "" && actual.Status.Phase != corev1.PodPending && actual.Status.Phase != corev1.PodRunning {
 					klog.V(5).Infof("Pod phase is neither pending nor running, deleting %v so the guard can be re-created", pod.Name)
+					delete = true
+				}
+				if actual.Spec.ServiceAccountName != pod.Spec.ServiceAccountName {
+					klog.V(5).Infof("Service Account changed, deleting %v so the guard can be re-created", pod.Name)
 					delete = true
 				}
 				if delete {
--- a/pkg/operator/staticpod/controller/guard/guard_controller_test.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller_test.go
@@ -9,6 +9,7 @@ import (
 	clocktesting "k8s.io/utils/clock/testing"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -270,8 +271,10 @@ func TestRenderGuardPod(t *testing.T) {
 		node                  *corev1.Node
 		guardExists           bool
 		guardPod              *corev1.Pod
+		guardServiceAccount   *corev1.ServiceAccount
 		createConditionalFunc func() (bool, bool, error)
 		withArbiter           bool
+		expectSADeleted       bool
 	}{
 		{
 			name: "Operand pod missing",
@@ -400,6 +403,7 @@ func TestRenderGuardPod(t *testing.T) {
 					ControlPlaneTopology: configv1.HighlyAvailableTopologyMode,
 				},
 			},
+			expectSADeleted: true,
 			operandPod: &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "operand1",
@@ -427,6 +431,13 @@ func TestRenderGuardPod(t *testing.T) {
 					PodIP: "1.1.1.1",
 				},
 			},
+			guardServiceAccount: &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "guard-sa",
+					Namespace: "test",
+					Labels:    map[string]string{"app": "guard"},
+				},
+			},
 			node: fakeMasterNode("master1"),
 		},
 		{
@@ -439,6 +450,7 @@ func TestRenderGuardPod(t *testing.T) {
 					ControlPlaneTopology: configv1.TopologyMode("HighlyAvailableArbiter"),
 				},
 			},
+			expectSADeleted: true,
 			operandPod: &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "operand1",
@@ -464,6 +476,13 @@ func TestRenderGuardPod(t *testing.T) {
 				},
 				Status: corev1.PodStatus{
 					PodIP: "1.1.1.1",
+				},
+			},
+			guardServiceAccount: &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "guard-sa",
+					Namespace: "test",
+					Labels:    map[string]string{"app": "guard"},
 				},
 			},
 			//createConditionalFunc: func() (bool, bool, error) { return true, true, nil },
@@ -522,6 +541,13 @@ func TestRenderGuardPod(t *testing.T) {
 					Phase: corev1.PodSucceeded,
 				},
 			},
+			guardServiceAccount: &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "guard-sa",
+					Namespace: "test",
+					Labels:    map[string]string{"app": "guard"},
+				},
+			},
 			node:        fakeMasterNode("master1"),
 			guardExists: true,
 		},
@@ -574,6 +600,13 @@ func TestRenderGuardPod(t *testing.T) {
 					Phase: corev1.PodSucceeded,
 				},
 			},
+			guardServiceAccount: &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "guard-sa",
+					Namespace: "test",
+					Labels:    map[string]string{"app": "guard"},
+				},
+			},
 			createConditionalFunc: func() (bool, bool, error) { return true, true, nil },
 			node:                  fakeArbiterNode("arbiter"),
 			guardExists:           true,
@@ -622,6 +655,9 @@ func TestRenderGuardPod(t *testing.T) {
 			if test.guardPod != nil {
 				kubeClient.Tracker().Add(test.guardPod)
 			}
+			if test.guardServiceAccount != nil {
+				kubeClient.Tracker().Add(test.guardServiceAccount)
+			}
 			kubeInformers := informers.NewSharedInformerFactoryWithOptions(kubeClient, 1*time.Minute)
 			eventRecorder := events.NewRecorder(kubeClient.CoreV1().Events("test"), "test-operator", &corev1.ObjectReference{}, clocktesting.NewFakePassiveClock(time.Now()))
 
@@ -651,6 +687,8 @@ func TestRenderGuardPod(t *testing.T) {
 				podGetter:               kubeClient.CoreV1(),
 				pdbGetter:               kubeClient.PolicyV1(),
 				pdbLister:               kubeInformers.Policy().V1().PodDisruptionBudgets().Lister(),
+				saGetter:                kubeClient.CoreV1(),
+				saLister:                kubeInformers.Core().V1().ServiceAccounts().Lister(),
 				installerPodImageFn:     getInstallerPodImageFromEnv,
 				createConditionalFunc:   createConditionalFunc,
 			}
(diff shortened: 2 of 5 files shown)


ehearne-redhat:
@wangke19 @p0lyn0mial this PR is ready to review. Could you please take a look? :)

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -225,11 +230,25 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			errs = append(errs, err)
 		} else {
 			for _, pod := range pods {
+				// get service account to delete
+				saName := pod.Spec.ServiceAccountName
```
ingvagabund: If a pod gets deleted before a SA is the SA will stay undeleted. Might work better to label the SA and delete in the same manner as pods are. + unit test this.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -320,6 +339,15 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 			klog.V(5).Infof("Rendering guard pod for operand %v on node %v", operands[node.Name].Name, node.Name)
 
+			// Create Service Account
+			serviceAccount := &corev1.ServiceAccount{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      getGuardPodName(c.podResourcePrefix, node.Name),
+					Namespace: c.targetNamespace,
+				},
+			}
+			resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
```
ingvagabund: error needs to be handled and retried + unit test

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -364,6 +393,10 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 					klog.V(5).Infof("Pod phase is neither pending nor running, deleting %v so the guard can be re-created", pod.Name)
 					delete = true
 				}
+				if actual.Spec.ServiceAccountName != pod.Spec.ServiceAccountName {
```
ingvagabund, line 409: Will `pod.Spec.ServiceAccountName` SA get eventually deleted?

ehearne-redhat:
Hey @ingvagabund - thanks so much for your review! I have amended the changes and added unit tests. Tests are passing locally. 

Please take a look when you have the chance :)

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -200,6 +200,26 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+		podMap[pod.Name] = true
+	}
+
+	serviceAccounts, _ := c.saLister.ServiceAccounts(c.targetNamespace).List(labels.SelectorFromSet(labels.Set{"app": "guard"}))
+
+	for _, sa := range serviceAccounts {
+		// If service account exists but pod map does not have a key for it,
+		// it is a zombie service account, and must be deleted.
+		if !podMap[sa.Name] {
+			c.saGetter.ServiceAccounts(c.targetNamespace).Delete(ctx, sa.Name, metav1.DeleteOptions{})
```
ingvagabund: Worth logging if the deletion fails so it's easier to track the errors for future debugging.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -200,6 +200,26 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 	}
 
+	// Do a check for zombie service accounts if a guard pod gets manually
+	// deleted by the user.
+	pods, _ := c.podLister.Pods(c.targetNamespace).List(labels.SelectorFromSet(labels.Set{"app": "guard"}))
+
+	podMap := map[string]bool{}
+
+	for _, pod := range pods {
+		podMap[pod.Name] = true
```
ingvagabund: Better to collect `pod.Spec.ServiceAccountName` just in case the SA naming convention changes. It will also make the code easier to interpret.

ingvagabund:
The PR looks good in overall. Thank you. Just few more nits.

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     app: guard
   ownerReferences: # Value set by operator
 spec:
+  serviceAccountName: # Value set by operator
```
p0lyn0mial, line 12: does the guard pod need an SA ?

could we just set `automountServiceAccountToken: false` for simplicity ?

ehearne-redhat: It does need a serviceaccount as indicated by [OCPSTRAT-2401](https://issues.redhat.com/browse/OCPSTRAT-2401) .

ehearne-redhat: Guard pods were detected using `default` service accounts, so this PR aims to rectify this. Are there any concerns with using a bespoke service account for guard pods?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -195,6 +200,29 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 		return nil
 	}
 
+	// Do a check for zombie service accounts if a guard pod gets manually
```
p0lyn0mial: Assuming the guard pod doesn't need an SA, would it make sense to add a single shared SA for all pods alongside the PDB ?

p0lyn0mial: I think that could simply the code and remove the concern of using the default SA.

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
p0lyn0mial: > It does need a serviceaccount as indicated by [OCPSTRAT-2401](https://issues.redhat.com/browse/OCPSTRAT-2401) .

ok, would it make sense to also add `automountServiceAccountToken: false` in that case ?

p0lyn0mial: > Guard pods were detected using default service accounts, so this PR aims to rectify this

ic, how about adding a single shared SA for all pods alongside the PDB ?
I think that could simply the code.

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: @p0lyn0mial the monitor tests only picked up the guard pods as being part of core openshift that use default service account. 

Could we introduce those suggestions in a separate PR? I'm happy to work on that separately if possible.

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
ehearne-redhat: Yes - absolutely. Could I introduce those changes in a new PR? I can create a new ticket in Jira and work on it separately if possible.

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: You can open a new PR, but if we decide to create a single shared SA, then we won't need most of the changes introduced in this PR. Thoughts?

ehearne-redhat: I see what you mean. I will re-work the PR with your suggestions here. :)

p0lyn0mial: OK. Feel free to add a new commit on top - so that we can remove it if we don't like it.

ehearne-redhat: Upon looking into your suggestion further, I found that `installer-sa` has a CRB attached with `cluster-admin` role. 

The guard pod does not require any additional permissions. So, it probably wouldn't be possible to simplify service account attachment to pods `installer-pod`, `prune-pod` and `guard-pod` as the first two pods require permissions to run, but `guard-pod` does not. 

Adding permissions to `guard-pod` would break the principle of least privilege.

ehearne-redhat: Unless you meant an alternative that I didn't catch - please let me know if so. :)

p0lyn0mial: mhm, I suggest adding one static SA instance (even with a fixed name) instead of creating many SAs per pod. 

Something similar to how PDB creation/deletion works now. There is only a single instance of PDB.

We do not need any permissions for this SA. This new SA should be used by the guard pods.

Does it make sense ? Could this work ?

ehearne-redhat: @p0lyn0mial I believe this should work. I have added that change now. Let me know if you require evidence PRs for review. :)

pkg/operator/staticpod/controller/guard/manifests/guard-sa.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: # Value set by operator
+  name: guard-sa
+  labels:
+    app: guard
```
p0lyn0mial: should we add a new line at the end of the file ?

pkg/operator/staticpod/controller/guard/manifests/guard-pod-sa.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
p0lyn0mial, line 1: would it make sense to rename the file to `guard-pod-sa` ?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -328,6 +354,7 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			pod.Spec.NodeName = node.Name
 			pod.Spec.Containers[0].Image = c.installerPodImageFn()
 			pod.Spec.Containers[0].ReadinessProbe.HTTPGet.Host = operands[node.Name].Status.PodIP
+			pod.Spec.ServiceAccountName = "guard-sa"
```
p0lyn0mial: read the name from the `serviceAccount` variable.

pkg/operator/staticpod/controller/guard/manifests/guard-pod.yaml
p0lyn0mial, line 12: let's also add `automountServiceAccountToken: false`

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -102,6 +105,8 @@ func NewGuardController(
 		podGetter:                     podGetter,
 		pdbGetter:                     pdbGetter,
 		pdbLister:                     kubeInformersForTargetNamespace.Policy().V1().PodDisruptionBudgets().Lister(),
+		saGetter:                      saGetter,
+		saLister:                      kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Lister(),
```
p0lyn0mial, line 108: add `saLister` to `factory.New().WithInformers` below

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +308,15 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			operands[pod.Spec.NodeName] = pod
 		}
 
+		// Create the guard service account once for all guard pods
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+		_, _, err = resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
```
p0lyn0mial: this will make a LIVE call to the API on every sync. Let's use the saLister - similarly to what we do for the PDB

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -232,6 +240,15 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 				}
 			}
 		}
+
+		// delete the guard service account
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+		_, _, err = resourceapply.DeleteServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
```
p0lyn0mial: let's use the `saLister` otherwise we issue a new LIVE call to the API server - similarly to what we do for the PDB

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -102,6 +106,8 @@ func NewGuardController(
 		podGetter:                     podGetter,
 		pdbGetter:                     pdbGetter,
 		pdbLister:                     kubeInformersForTargetNamespace.Policy().V1().PodDisruptionBudgets().Lister(),
+		saGetter:                      saGetter,
+		saLister:                      saLister,
```
p0lyn0mial: let's take the lister from `kubeInformersForTargetNamespace`

p0lyn0mial: then let's add the informer to `factory.New().WithInformers()` below.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +317,23 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+		// Create the guard service account once for all guard pods
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if errors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				klog.Errorf("Unable to create service account %v for Guard Pods: %v", serviceAccount.Name, err)
```
p0lyn0mial: won't this ere be logged because we use `WithSyncDegradedOnError` ?
if yes, maybe we should skip the additional logging. (applies to the other places)

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +317,23 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+				return fmt.Errorf("Unable to create service account %v for Guard Pods: %v", serviceAccount.Name, err)
```
p0lyn0mial: also, I think that in go in general errors should not be capitalized.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +317,23 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			operands[pod.Spec.NodeName] = pod
 		}
 
+		// Create the guard service account once for all guard pods
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
```
p0lyn0mial, line 325: maybe we should add a comment which would explain why we don't want to perform updates ?

p0lyn0mial, line 325: And maybe we should create a simple unit test that fails when guard-pod-sa gets updated with a message saying that maybe the code needs to also perform an update ?

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: OK, I will remove the redundant `klog.Errorf` from my changes.

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: Yes, this is true. I was just following the convention from the other errors. I will remove those capital starts.

pkg/operator/staticpod/controllers.go
```diff
@@ -242,7 +242,9 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
 	podClient := b.kubeClient.CoreV1()
 	eventsClient := b.kubeClient.CoreV1()
 	pdbClient := b.kubeClient.PolicyV1()
+	saClient := b.kubeClient.CoreV1()
 	operandInformers := b.kubeNamespaceInformers.InformersFor(b.operandNamespace)
+	saLister := operandInformers.Core().V1().ServiceAccounts().Lister()
```
p0lyn0mial: we take the lister from `kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Lister()` this is not needed.

pkg/operator/staticpod/controllers.go
```diff
@@ -398,6 +400,8 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
 			b.staticPodOperatorClient,
 			podClient,
 			pdbClient,
+			saClient,
+			saLister,
```
p0lyn0mial: we take the lister from kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Lister() this is not needed.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -70,6 +72,8 @@ func NewGuardController(
 	operatorClient operatorv1helpers.StaticPodOperatorClient,
 	podGetter corev1client.PodsGetter,
 	pdbGetter policyclientv1.PodDisruptionBudgetsGetter,
+	saGetter corev1client.ServiceAccountsGetter,
+	saLister corelisterv1.ServiceAccountLister,
```
p0lyn0mial: we take the lister from kubeInformersForTargetNamespace.Core().V1().ServiceAccounts().Lister() this is not needed.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -227,11 +237,28 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+		// delete the guard service account
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if err == nil {
+			_, _, err = resourceapply.DeleteServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				klog.Errorf("unable to delete Service Account: %v", err)
```
p0lyn0mial: use `%w` for err wrapping.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -227,11 +237,28 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+				errs = append(errs, err)
+			}
+		} else if !apierrors.IsNotFound(err) {
+			klog.Errorf("unable to get service account %v from lister: %v", serviceAccount.Name, err)
```
p0lyn0mial: use `%w` for err wrapping.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +318,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		// By using saLister, we avoid redundant API calls when the API can be called just once
+		// with saLister. saLister reads from an in-memory cache populated by the informer and
+		// does not query the Kubernetes API for every request.
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if errors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				return fmt.Errorf("Unable to create service account %v for Guard Pods: %v", serviceAccount.Name, err)
```
p0lyn0mial: use `%w` for err wrapping.

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: ping

pkg/operator/staticpod/controller/guard/guard_controller_test.go
```diff
@@ -875,6 +938,81 @@ func TestGuardPodTemplate(t *testing.T) {
 	}
 }
 
+func TestGuardServiceAccountManifestStability(t *testing.T) {
```
p0lyn0mial, line 941: could this test be updated to something like:
```
      sa := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)

      expected := &corev1.ServiceAccount{
              Name:   "guard-sa",
              Labels: map[string]string{"app": "guard"},
          },
      }

      if !equality.Semantic.DeepEqual(sa, expected) {
          t.Fatalf("guard-pod-sa.yaml manifest has changed. " +
              "The controller only creates the SA, it does not update existing ones. " +
              "If the manifest changed, add update logic and update the expected value in this test.")
      }
```

?

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: Something along the lines of: There are no secrets, no image pull secrets, no annotations. The guard pod uses automountServiceAccountToken: false. There's nothing meaningful to update. So update logic isn't needed right now.

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: I'm getting this message when I try to use error wrapping
`k8s.io/klog/v2.Errorf does not support error-wrapping directive %w`

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: Same here
`k8s.io/klog/v2.Errorf does not support error-wrapping directive %w`

pkg/operator/staticpod/controller/guard/guard_controller.go
ehearne-redhat: This one I can do. :)

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: won't this ere be logged because we use WithSyncDegradedOnError ?
if yes, maybe we should skip the additional logging. (applies to the other places)

p0lyn0mial: we could use fmt if we want to enhance the context and then add it to the errs

pkg/operator/staticpod/controller/guard/guard_controller_test.go
```diff
@@ -270,8 +271,10 @@ func TestRenderGuardPod(t *testing.T) {
 		node                  *corev1.Node
 		guardExists           bool
 		guardPod              *corev1.Pod
+		guardServiceAccount   *corev1.ServiceAccount
 		createConditionalFunc func() (bool, bool, error)
 		withArbiter           bool
+		expectSADeleted       bool // true if service account should be deleted (cleanup mode)
```
p0lyn0mial: let's rm the comment, not sure it is helpful.

pkg/operator/staticpod/controller/guard/guard_controller.go
p0lyn0mial: won't this ere be logged because we use WithSyncDegradedOnError ?
if yes, maybe we should skip the additional logging. (applies to the other places

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +317,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		// There are no secrets, no image pull secrets, no annotations.
+		// The guard pod uses automountServiceAccountToken: false.
+		// There's nothing meaningful to update. So update logic isn't needed right now.
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if errors.IsNotFound(err) {
+			_, _, err = resourceapply.ApplyServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				return fmt.Errorf("Unable to create service account %v for Guard Pods: %w", serviceAccount.Name, err)
```
p0lyn0mial: isn't the go convention to start err with a lowercase letter ?

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -291,6 +317,24 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+		// Create the guard service account once for all guard pods
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		// There are no secrets, no image pull secrets, no annotations.
+		// The guard pod uses automountServiceAccountToken: false.
+		// There's nothing meaningful to update. So update logic isn't needed right now.
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if errors.IsNotFound(err) {
```
p0lyn0mial: let's use `apierrors ` and remove`"k8s.io/apimachinery/pkg/api/errors"` from the imports.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -227,11 +236,28 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+		// delete the guard service account
+		serviceAccount := resourceread.ReadServiceAccountV1OrDie(serviceAccountTemplate)
+		serviceAccount.Namespace = c.targetNamespace
+
+		// Check if the service account exists using the lister to avoid unnecessary API calls
+		_, err = c.saLister.ServiceAccounts(serviceAccount.Namespace).Get(serviceAccount.Name)
+		if err == nil {
+			_, _, err = resourceapply.DeleteServiceAccount(ctx, c.saGetter, syncCtx.Recorder(), serviceAccount)
+			if err != nil {
+				klog.Errorf("unable to delete Service Account: %v", err)
```
p0lyn0mial: the other log msg use "service account" not "Service Account", let's be consistent.

ehearne-redhat: I believe this error log has been removed based on previous comment above. The consistency has been resolved anyway. :)

PR #2080, opened 2026-01-19
pkg/crypto: Add missing cipher suites to OpenSSL-to-IANA ciphers mapping

damdo:
### Summary

Expands the `openSSLToIANACiphersMap` in `pkg/crypto/crypto.go`
to include additional TLS cipher suites that were previously missing.
This enables the `OpenSSLToIANACipherSuites()` function to translate
a broader set of OpenSSL cipher names to their IANA equivalents.

### Why

The existing map was missing several commonly used cipher suites,
particularly DHE (Diffie-Hellman Ephemeral) variants.
This caused `OpenSSLToIANACipherSuites()` to drop these ciphers
when translating cipher lists from OpenSSL format to IANA format.
Causing operators using this pkg to log these suites as unsupported.


pkg/crypto/crypto.go
```diff
@@ -159,9 +157,17 @@ var openSSLToIANACiphersMap = map[string]string{
 	"ECDHE-RSA-CHACHA20-POLY1305":   "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",   // 0xCC,0xA8
 	"ECDHE-ECDSA-AES128-SHA256":     "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",       // 0xC0,0x23
 	"ECDHE-RSA-AES128-SHA256":       "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",         // 0xC0,0x27
+	"ECDHE-ECDSA-AES256-SHA384":     "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384",       // 0xC0,0x24
```
p0lyn0mial, line 160: OK, it looks like most of the added ciphers also appear in the TLSProfiles (Old, Intermediate, Modern), so these should be added to the mapping.

pkg/crypto/crypto.go
```diff
@@ -170,9 +176,10 @@ var openSSLToIANACiphersMap = map[string]string{
 	"ECDHE-RSA-AES256-SHA":   "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA",   // 0xC0,0x14
 
 	// SSL 3
-	"AES128-SHA":   "TLS_RSA_WITH_AES_128_CBC_SHA",  // 0x00,0x2F
-	"AES256-SHA":   "TLS_RSA_WITH_AES_256_CBC_SHA",  // 0x00,0x35
-	"DES-CBC3-SHA": "TLS_RSA_WITH_3DES_EDE_CBC_SHA", // 0x00,0x0A
+	"AES128-SHA":             "TLS_RSA_WITH_AES_128_CBC_SHA",        // 0x00,0x2F
+	"AES256-SHA":             "TLS_RSA_WITH_AES_256_CBC_SHA",        // 0x00,0x35
+	"DES-CBC3-SHA":           "TLS_RSA_WITH_3DES_EDE_CBC_SHA",       // 0x00,0x0A
+	"ECDHE-RSA-DES-CBC3-SHA": "TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA", // 0xC0,0x12
```
p0lyn0mial, line 182: However, I cannot find this particular cipher. Could you explain why we would like to add it? I think this cipher is generally considered weak.

damdo: Good question.

The reason I'm adding these is to allow a translation between one format (OpenSSL) and the other (IANA).

The ciphers I put here do not imply these are safe or not weak, it just means that whomever uses `crypto.OpenSSLToIANACipherSuites()` is able to translate as many formats as possible from one domain to the other.

I did a scan of both formats and added all the ciphers that are in the intersection of the two.

IMO it shouldn't be the responsibility of the OpenSSLToIANACipherSuites() translation function to decide which ciphers are vulnerable and which ones are not, but it should be at the specific profile level.

This is also specified in the godoc signature of this function, which says `// Unknown ciphers are left out`.

If we don't want users to be able to specify a certain ciphers in the Custom profile, we should block them at the API level, or error at some equivalent level.

p0lyn0mial: ok, thanks, this makes sense.

ardaguclu:
I've opened revert PR https://github.com/openshift/library-go/pull/2118

damdo:
I see, thanks @ardaguclu it sounds like those operators have similar logic in how they treat unknown ciphers.

It feels to me that we should treat it as a bug in those operators as these are actually the ciphers specified in the profiles defined at the APIServer level, which are part of the desired configuration for operators secure serving. (cc. @joelanford as the PQC Staff Eng. for awareness)

If/When you have the bug (or bugs if you are creating more) for these components, could you please post it here, so we can track it? :)

Thanks!

ardaguclu:
This is kubernetes limitation. You can only add the cipher suites defined in https://github.com/kubernetes/kubernetes/blob/dc1ec1211e4f54064ba6dafd8aac46ac3d4379b4/staging/src/k8s.io/component-base/cli/flag/ciphersuites_flag.go#L34-L43 kubernetes. API Server won't allow the others.

damdo:
Ack thanks for the context. Please do keep us posted on how it goes, ty.

ardaguclu:
@damdo I agree with you. This needs to be fixed by either modifying in upstream or openshift/api. I assume that there is a Jira epic and dedicated team in control plane group for this?

PR #2081, opened 2026-01-19
OCPBUGS-38662: node controller: Don't degrade when node is upgrading

tchap:
Treat nodes as `Degraded` only when they are not upgrading.
This prevents `Degraded=True` in operators from blipping during upgrade.

There is a helper function that returns a node upgrade state enum value called `GetNodeUpgradeState`. The upgrade state is decided based on relevant MCO annotations on the Node object. This can be used by other controllers in library-go as well.

The idea is then to not mark nodes as Degraded when they are Rebooting or Working, which explicitly excludes draining, because that is unrelated to the node Ready status anyway. This seems the best we can do.

---

### Obsolete Description to Keep Context

Treat nodes as `Degraded` only when they are not rebooting for upgrade.
This prevents `Degraded=True` in operators from blipping during upgrade.

Based on https://github.com/openshift/machine-config-operator/blob/b4ea81d6885c61c91e1503d5f9f5ebed696092b1/pkg/daemon/update.go#L2812

Now the issue is that the annotation can be removed when the reboot is finished before the node becomes `Ready`. But this should be like seconds before it's `Ready`. On top of that, there is a 2-minute inertia for `Degraded` on the operator status, so this should be plenty to cover the gap. Another potential issue is that failing to update the annotations is only logged, it does not prevent the reboot. So we may have rebooting nodes missing the annotation.

An alternative would be ignore nodes with `machineconfiguration.openshift.io/state == Working`, but that seems to be too broad a condition as it covers basically the whole upgrade process, so that can show too positive view of the system. But it is a possibility.

There may be other ways to implement this for sure, but that would require pulling in other objects like `machineconfignode` for the nodes, where you can check the timestamp of the last reboot, for example. This doesn't seem particularly necessary as the inertia should cover vast majority of our use cases. So I focused on what we already have, which are the `Node` objects.

As discussed, I also ended up adding a `Degraded` inertia here, which means that the rebooting node is only ignored for certain period of time, which is by default 10 minutes. In case something goes very wrong during the upgrade...

### DIFF
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -19,6 +19,11 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+const (
+	machineConfigDaemonPostConfigAction = "machineconfiguration.openshift.io/post-config-action"
+	machineConfigDaemonStateRebooting   = "Rebooting"
+)
+
 // NodeController watches for new master nodes and adds them to the node status list in the operator config status.
 type NodeController struct {
 	controllerInstanceName string
@@ -145,28 +150,43 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext) 
 		}
 	}
 
-	// detect and report master nodes that are not ready
-	notReadyNodes := []string{}
+	// Detect and report master nodes that are not ready.
+	// Nodes currently rebooting for upgrade do not cause Degraded condition to be set.
+	var notReadyNodes []string
+	var notReadyRebootingNodes []string
 	for _, node := range nodes {
 		nodeReadyCondition := nodeConditionFinder(&node.Status, coreapiv1.NodeReady)
 
 		// If a "Ready" condition is not found, that node should be deemed as not Ready by default.
-		if nodeReadyCondition == nil {
+		var notReady bool
+		switch {
+		case nodeReadyCondition == nil:
 			notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready, no Ready condition found in status block", node.Name))
-			continue
-		}
+			notReady = true
 
-		if nodeReadyCondition.Status != coreapiv1.ConditionTrue {
+		case nodeReadyCondition.Status != coreapiv1.ConditionTrue:
 			notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready since %s because %s (%s)", node.Name, nodeReadyCondition.LastTransitionTime, nodeReadyCondition.Reason, nodeReadyCondition.Message))
+			notReady = true
+		}
+		if notReady && nodeRebootingForUpgrade(node) {
+			notReadyRebootingNodes = append(notReadyRebootingNodes, node.Name)
 		}
 	}
 
-	if len(notReadyNodes) > 0 {
+	switch {
+	case len(notReadyNodes) > len(notReadyRebootingNodes):
 		degradedCondition = degradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("MasterNodesReady").
 			WithMessage(fmt.Sprintf("The master nodes not ready: %s", strings.Join(notReadyNodes, ", ")))
-	} else {
+
+	case len(notReadyRebootingNodes) > 0:
+		degradedCondition = degradedCondition.
+			WithStatus(operatorv1.ConditionFalse).
+			WithReason("MasterNodesReady").
+			WithMessage(fmt.Sprintf("The master nodes rebooting for upgrade: %s", strings.Join(notReadyRebootingNodes, ", ")))
+
+	default:
 		degradedCondition = degradedCondition.
 			WithStatus(operatorv1.ConditionFalse).
 			WithReason("MasterNodesReady").
@@ -195,4 +215,8 @@ func nodeConditionFinder(status *coreapiv1.NodeStatus, condType coreapiv1.NodeCo
 	}
 
 	return nil
+}
+
+func nodeRebootingForUpgrade(node *coreapiv1.Node) bool {
+	return node.Annotations[machineConfigDaemonPostConfigAction] == machineConfigDaemonStateRebooting
 }
--- a/pkg/operator/staticpod/controller/node/node_controller.go
+++ b/pkg/operator/staticpod/controller/node/node_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	coreapiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -19,18 +20,34 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+// DefaultRebootingNodeDegradedInertia is the default period during which a node rebooting for upgrade is not considered Degraded.
+// The value is pretty large because bare metal nodes can take a long time to reboot for upgrade.
+const DefaultRebootingNodeDegradedInertia = 2 * time.Hour
+
 const (
 	machineConfigDaemonPostConfigAction = "machineconfiguration.openshift.io/post-config-action"
-	machineConfigDaemonStateRebooting   = "Rebooting"
+
+	machineConfigDaemonStateRebooting = "Rebooting"
 )
+
+// NodeControllerOption can be passed to NewNodeController to configure the controller.
+type NodeControllerOption func(*NodeController)
+
+// SetRebootingNodeDegradedInertia sets the period during which a node rebooting for upgrade is not considered Degraded.
+func SetRebootingNodeDegradedInertia(inert time.Duration) NodeControllerOption {
+	return func(c *NodeController) {
+		c.rebootingNodeDegradedInertia = inert
+	}
+}
 
 // NodeController watches for new master nodes and adds them to the node status list in the operator config status.
 type NodeController struct {
-	controllerInstanceName string
-	operatorClient         v1helpers.StaticPodOperatorClient
-	nodeLister             corelisterv1.NodeLister
-	extraNodeSelector      labels.Selector
-	masterNodesSelector    labels.Selector
+	controllerInstanceName       string
+	operatorClient               v1helpers.StaticPodOperatorClient
+	nodeLister                   corelisterv1.NodeLister
+	extraNodeSelector            labels.Selector
+	masterNodesSelector          labels.Selector
+	rebootingNodeDegradedInertia time.Duration
 }
 
 // NewNodeController creates a new node controller.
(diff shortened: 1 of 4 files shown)


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -62,6 +73,11 @@ func NewNodeController(
 		)
 }
 
+// SetRebootingNodeDegradedInertia sets the period during which a node rebooting for upgrade is not considered Degraded.
+func (c *NodeController) SetRebootingNodeDegradedInertia(inert time.Duration) {
+	c.rebootingNodeDegradedInertia = inert
+}
```
tchap: This is actually useless, the constructor does not return `NodeController` 😐

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -196,3 +243,11 @@ func nodeConditionFinder(status *coreapiv1.NodeStatus, condType coreapiv1.NodeCo
 
 	return nil
 }
+
+func nodeRebootingForUpgrade(node *coreapiv1.Node) bool {
+	return node.Annotations[machineConfigDaemonPostConfigAction] == machineConfigDaemonStateRebooting
```
p0lyn0mial: could we start a conversation with  the mco team or the forum-ocp-updates to check what information is available during an upgrade that we could use?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -19,13 +20,32 @@ import (
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+// DefaultRebootingNodeDegradedInertia is the default period during which a node rebooting for upgrade is not considered Degraded.
+const DefaultRebootingNodeDegradedInertia = 10 * time.Minute
```
tchap: Actually need to decide on the default value based on how long it usually takes to restart.

Filip proposed something around an hour, but needs to be verified.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -145,33 +172,55 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+			// If a "Ready" condition is not found, that node should be deemed as not Ready by default.
+			degradedMsg = fmt.Sprintf("node %q not ready, no Ready condition found in status block", node.Name)
 
-		if nodeReadyCondition.Status != coreapiv1.ConditionTrue {
-			notReadyNodes = append(notReadyNodes, fmt.Sprintf("node %q not ready since %s because %s (%s)", node.Name, nodeReadyCondition.LastTransitionTime, nodeReadyCondition.Reason, nodeReadyCondition.Message))
+		case nodeReadyCondition.Status != coreapiv1.ConditionTrue:
+			degradedMsg = fmt.Sprintf("node %q not ready since %s because %s (%s)", node.Name, nodeReadyCondition.LastTransitionTime, nodeReadyCondition.Reason, nodeReadyCondition.Message)
+		}
+		if len(degradedMsg) > 0 {
+			if nodeRebootingForUpgrade(node) && !shouldDegradeRebootingNode(nodeReadyCondition, c.rebootingNodeDegradedInertia) {
```
p0lyn0mial: I’m wondering what the chances are that this is the only controller that becomes degraded during an upgrade. Something tells me there might be many more controllers that are unhappy. 

I would prefer a more global solution to this problem. 

Also, silencing a degraded condition for two hours seems excessive. 

What about using `StatusSyncer.WithDegradedInertia`? I think it has the potential to be a broader solution. 

What is the maximum time after a node becomes degraded once the kubelet stops posting status updates?

Thoughts ?

tjungblu: @p0lyn0mial just digging this up again with Ondra. I've been using the degraded inertia on CEO for a month now and it has been quite good at shutting up those blips. 

Would you be open to discuss this further in https://github.com/openshift/cluster-kube-apiserver-operator/pull/2052 for KAS-O?

pkg/operator/staticpod/controller/node/node_state.go
```diff
@@ -0,0 +1,116 @@
...
+		return NodeUpgradeStateDone
+
+	case machineConfigStateWorking:
+		if node.Annotations[machineConfigPostConfigAction] == machineConfigPostConfigActionRebooting {
+			return NodeUpgradeStateRebooting
+		}
+
+		desiredDrain := node.Annotations[machineConfigDesiredDrain]
+		lastAppliedDrain := node.Annotations[machineConfigLastAppliedDrain]
+		if desiredDrain != "" && lastAppliedDrain != desiredDrain {
```
atiratree: This can be either `uncordon` or `drain`. Maybe we should not report draining, if the desired action is not `drain`.

pkg/operator/staticpod/controller/node/node_state.go
```diff
@@ -0,0 +1,116 @@
...
+			return NodeUpgradeStateDraining
+		}
```
atiratree, line 108: It might be also interesting to report cordoned (`.spec.schedulable` state) for non-static pod workloads since they will most likely will not be able to schedule.

pkg/operator/staticpod/controller/node/node_state_test.go
```diff
@@ -0,0 +1,218 @@
...
+		{
+			name: "rebooting takes precedence over drain annotations",
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-node",
+					Annotations: map[string]string{
+						machineConfigState:            machineConfigStateWorking,
+						machineConfigPostConfigAction: machineConfigPostConfigActionRebooting,
+						machineConfigDesiredDrain:     "drain-rendered-worker-xyz789",
+						machineConfigLastAppliedDrain: "drain-rendered-worker-xyz789",
```
atiratree:
```suggestion
						machineConfigLastAppliedDrain: "drain-rendered-worker-xyz666",
```

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -145,33 +166,55 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
+				upgradingNodes = append(upgradingNodes, fmt.Sprintf("node %q", node.Name))
+			} else {
+				degradedNodes = append(degradedNodes, degradedMsg)
+			}
+		}
+	}
+
+	degradedCondition = degradedCondition.WithReason("MasterNodesReady")
+
+	if len(degradedNodes) > 0 {
```
atiratree, line 196: it might be also useful to consider the total number of healthy and unhealthy nodes. E.g. by defining a maxUnavailable, and if we go over that number we could report degraded right away. We could set a default to 1 and let it be configurable in the NodeController constructor.

pkg/operator/staticpod/controller/node/node_controller_test.go
```diff
@@ -489,6 +525,234 @@ func TestNodeControllerDegradedConditionType(t *testing.T) {
...
+				expectedCondition.Type = condition.NodeControllerDegradedConditionType
+				expectedCondition.Reason = "MasterNodesReady"
+				expectedCondition.Status = operatorv1.ConditionTrue
+				expectedCondition.Message = `The master nodes not ready: node "test-node-1" not ready since 2018-01-12 22:51:48.324359102 +0000 UTC because TestReason (test message)`
+
+				return validateNodeControllerDegradedCondition(conditions, expectedCondition)
+			},
+		},
+		{
+			name: "scenario 15: one healthy but rebooting master node is reported",
```
atiratree, line 564:
```suggestion
			name: "scenario 15: one healthy but rebooting ready master node is reported",
```

tchap: I think that `ready` is already implied by the word `healthy`...

pkg/operator/staticpod/controller/node/node_state.go
tchap: Fixed in 819da93d8b84157d8534d98c591b8dc7df3b3641

pkg/operator/staticpod/controller/node/node_state.go
tchap: I think the caller can check this on their own TBH. This function is mainly for hiding the complexity regarding the annotations. The node is also cordoned while draining, so actually these are overlapping, meaning that the caller would have to know that `Draining` also implies `Cordoned`, but for that they can simply check the field themselves.

pkg/operator/staticpod/controller/node/node_controller.go
tchap: No problem with that, but we can do that in a separate PR, I think.

pkg/operator/staticpod/controller/node/node_state.go
atiratree: Sure, it is okay for now. We might revisit this later, when we need this kind of functionality again.

pkg/operator/staticpod/controller/node/node_controller_test.go
atiratree: ok, I got it mixed with the other node

pkg/operator/staticpod/controller/node/node_controller.go
p0lyn0mial: yeah, could we try the #2025 PR first? It is less invasive and covers all controllers.

tchap: I don't think this is really about doing one OR the other. Because we still need to detect upgrades somehow. They take longer than 10 mins without any problems, particularly on bare metal. Increasing the threshold to 10 or 15 minutes generally should still be used mostly for degradation when NOT upgrading.

p0lyn0mial: > I don't think this is really about doing one OR the other. Because we still need to detect upgrades somehow.

Isn’t this a platform issue, then? Should it be resolved at the platform level so that all controllers and operators are aware of the ongoing upgrade? (I think ehere was a KEP around this)

/cc @benluddy

tchap: @p0lyn0mial Yes, absolutely. There is for example https://github.com/kubernetes/enhancements/pull/5769 . But we need to do something about this before this is available. The KEP is not even merged yet. So we basically decided to do (passing what @atiratree wrote on Slack and I followed by updating this PR):

1. Degrade immediately if we know that something is definitely wrong (0 replicas available, etcd lost quorum, no leader present after the election timeout)
2. Degrade with a small inertia in general if 1. is okay (15m which @tjungblu experimented with and should be sufficient to cover usual disruption cases)
3. Degrade with a large inertia if 1. is okay and if we can detect that node is being shutdown/upgraded (1-2h? similar to what @tchap proposes in his PR)
 
Point 3 is relevant in our case to node controller and static pod controller set. And yes, we can share some helper functions like I implemented, but currently the controllers need to decide how to handle node upgrades, I guess.

sub-mod:
@sdodson @tchap any updates on this PR ? Thanks

tchap:
I am closing this in favor of setting the 10-minute inertia as discussed on the arch call, see https://github.com/openshift/cluster-kube-controller-manager-operator/pull/947

PR #2085, opened 2026-01-23
CNTRLPLANE-2247:Adding KMS e2e cases

gangwgr:
Adding KMS e2e cases and KMS mock plugin
```
 oc get po -n openshift-kms-plugin                                        
NAME               READY   STATUS    RESTARTS   AGE
kms-plugin-4n25r   1/1     Running   0          26s
kms-plugin-kzmq8   1/1     Running   0          26s
kms-plugin-nm676   1/1     Running   0          26s
rgangwar@rgangwar-mac library-go % oc logs pod/kms-plugin-4n25r -n openshift-kms-plugin     
Defaulted container "kms-plugin" out of: kms-plugin, init-softhsm (init)
+ export 'SOFTHSM2_CONF=/var/lib/softhsm/softhsm2.conf'
Validating SoftHSM config...
+ echo 'Validating SoftHSM config...'
+ cat /var/lib/softhsm/softhsm2.conf
directories.tokendir = /var/lib/softhsm/tokens
objectstore.backend = file
slots.removable = false
log.level = DEBUG
+ echo 'Validating token directory...'
Validating token directory...
+ ls -la /var/lib/softhsm/tokens/
total 0
drwx------    3 root     root            50 Jan 27 06:29 .
drwxrwxrwx    3 root     root            41 Jan 27 06:29 ..
drwx------    2 root     root           162 Jan 27 06:29 3949b7a1-0391-c0c8-ecaa-36b42ff8f37c
+ echo 'Starting mock KMS v2 plugin...'
Starting mock KMS v2 plugin...
+ exec mock-kms-plugin '-listen-addr=unix:///var/run/kmsplugin/kms.sock' '-config-file-path=/etc/softhsm-config.json'
rgangwar@rgangwar-mac library-go % 
```

### DIFF
--- a/test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
+++ b/test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
@@ -59,9 +59,10 @@ type yamlTemplateData struct {
 func DeployUpstreamMockKMSPlugin(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, image string) func() {
 	t.Helper()
 
-	if err := destroyNamespaceIfNotExists(ctx, t, kubeClient, namespace); err != nil {
-		t.Fatalf("Failed to cleanup existing namespace %q: %v", namespace, err)
-	}
+	// TEMPORARILY DISABLED FOR TESTING - namespace deletion before deploy
+	// if err := destroyNamespaceIfNotExists(ctx, t, kubeClient, namespace); err != nil {
+	// 	t.Fatalf("Failed to cleanup existing namespace %q: %v", namespace, err)
+	// }
 
 	t.Logf("Deploying upstream mock KMS v2 plugin in namespace %q using image %s", namespace, image)
 	daemonSetName, err := applyUpstreamMockKMSPluginManifests(ctx, t, kubeClient, namespace, image)
@@ -77,9 +78,10 @@ func DeployUpstreamMockKMSPlugin(ctx context.Context, t testing.TB, kubeClient k
 		// Before destroying the namespace, collect the logs of the pods in namespace
 		collectPodLogs(ctx, t, kubeClient, namespace)
 
-		if err := destroyNamespaceIfNotExists(ctx, t, kubeClient, namespace); err != nil {
-			t.Errorf("Failed to cleanup namespace %q: %v", namespace, err)
-		}
+		// TEMPORARILY DISABLED FOR TESTING - namespace deletion in cleanup
+		// if err := destroyNamespaceIfNotExists(ctx, t, kubeClient, namespace); err != nil {
+		// 	t.Errorf("Failed to cleanup namespace %q: %v", namespace, err)
+		// }
 	}
 }
 


test/library/encryption/kms_scenarios.go
```diff
@@ -0,0 +1,1478 @@
+package encryption
```
p0lyn0mial: @gangwgr I think that we could reuse already existing machinery/code. Essentially this should work:

Tests in general are defined in operator repos, library-go contains helpers.
I think that we will need tests for the **three operators** that will share common code from library-go

Let's start with `kas-o` first. Let's tart with `OnOff` scenario first.

1. I think that we should create a new job which would run [e2e-encryption-kms](https://github.com/openshift/cluster-kube-apiserver-operator/tree/850daefcfb8727f7bd404ef716787fd5e6effc8d/test). I think that this job must be optional initially and use TP.

2. We add [OnOff](https://github.com/openshift/cluster-kube-apiserver-operator/blob/850daefcfb8727f7bd404ef716787fd5e6effc8d/test/e2e-encryption/encryption_test.go#L40) scenario to `e2e-encryption-kms` with KMS [provider](https://github.com/openshift/cluster-kube-apiserver-operator/blob/850daefcfb8727f7bd404ef716787fd5e6effc8d/test/e2e-encryption/encryption_test.go#L14C5-L14C13)

3. We add a function to library-go that knows how to deploy the KMS-plugin we need.

4. We add TestEncryptionTypeKMS to library-go, this should be based on [TestEncryptionTypeAESGCM](https://github.com/openshift/cluster-kube-apiserver-operator/blob/8b6ce4dca27486b2b99c394439a50073be8af521/vendor/github.com/openshift/library-go/test/library/encryption/scenarios.go#L46C6-L46C30). This function will use the function from `3` to ensure the KMS plugin has been installed. The rest is the same as in `TestEncryptionTypeAESGCM`

5. We extend [TestEncryptionType](https://github.com/openshift/cluster-kube-apiserver-operator/blob/8b6ce4dca27486b2b99c394439a50073be8af521/vendor/github.com/openshift/library-go/test/library/encryption/scenarios.go#L53C6-L53C24) by the new KMS type.

ardaguclu: This looks nice to me. So in the next iteration, we can add more complex test scenarios based on this work.

gangwgr: @p0lyn0mial update pr, removed additional cases just keeping helper function and TestEncryptionTypeKMS tests that resources are encrypted using KMS. which follows the same pattern as TestEncryptionTypeAESGCM but ensures the KMS plugin is deployed first

test/library/encryption/kms_scenarios.go
```diff
@@ -0,0 +1,463 @@
...
+cd /tmp
+if [ ! -d "kubernetes" ]; then
+    git clone --depth 1 --branch v1.29.0 https://github.com/kubernetes/kubernetes.git
+fi
+
+cd kubernetes/staging/src/k8s.io/kms/internal/plugins/mock
+
+# Build the mock plugin
+echo "Building mock KMS plugin..."
+go build -o ${PLUGIN_DIR}/mock-kms-plugin .
```
p0lyn0mial: why not to reuse [the docker image](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kms/internal/plugins/_mock/Dockerfile) from upstream to build the plugin ?

I think we could open a new PR to library-go that: 

1. creates a new dir at `test/library/encryption/kms/k8s-mock-plugin/`
2. adds a small bash script (for example `build-from-k8s.sh`) to that dir that downloads the docker img from upstream and builds an image.
3. adds a README.md file that explains how to run the script
4. I think that pushing the img to the remote repo should not be part of the script but the README should contain some info about how to push the img.

Then we would use this image during deployment. This approach makes the test more stable and faster as we don't have to download the entire k8s repo during testing.


For backport we just move the entire `test/library/encryption/kms` dir.

test/library/encryption/kms_scenarios.go
```diff
@@ -0,0 +1,463 @@
...
+}
+
+func ensureKMSDaemonSet(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, config KMSPluginDeploymentConfig) error {
+	t.Helper()
+
+	privileged := true
+	hostPathType := corev1.HostPathDirectoryOrCreate
+	terminationGracePeriod := int64(30)
+
+	ds := &appsv1.DaemonSet{
```
p0lyn0mial: I think using a `DaemonSet` should be ok.
However I think that it would be easier to have resources defined directly in YAML files.

I think that we could (in a new PR)

1. create a new directory `test/library/encryption/kms/assets`
2. move all resources definition there.

Then in a new PR:
1. create a new file `test/library/encryption/kms/assets/plugin_deployer.go` which `//go:embed assets/*` and exposes a function for installing DaemonSet

then again, for backport we just move the entire `test/library/encryption/kms` dir.

test/library/encryption/kms_scenarios.go
p0lyn0mial: EDIT: updated the link to the docker image.

p0lyn0mial: This might be a tmp solution. Long term we could https://github.com/openshift/origin/tree/main/test/extended/util/image

p0lyn0mial: The docker img might be available in an upstream registry. Worth checking.

test/library/encryption/kms/assets/daemonset.yaml
```diff
@@ -0,0 +1,103 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: mock-kms-plugin
+  template:
+    metadata:
+      labels:
+        app: mock-kms-plugin
+    spec:
+      hostNetwork: true
```
ardaguclu: I don't think we need this. I believe that as long as our namespace and daemonset are privileged, apiservers can communicate with kms plugin via hostPath without requiring access to hostNetwork.

test/library/encryption/kms/k8s-mock-plugin/build-from-k8s.sh
```diff
@@ -0,0 +1,103 @@
...
+
+echo "=== Building Kubernetes Mock KMS v2 Plugin ==="
+echo "Image tag: ${IMAGE_TAG}"
+echo "Kubernetes version: ${K8S_VERSION}"
+echo ""
+
+# Detect container runtime (docker or podman)
+CONTAINER_RUNTIME=""
+if command -v docker &> /dev/null; then
+    CONTAINER_RUNTIME="docker"
```
ardaguclu: Do we really need this?. Doesn't our CI rely on podman?

test/library/encryption/kms_scenarios.go
ardaguclu: > The docker img might be available in an upstream registry. Worth checking.

When I searched upstream maintained image in a couple of days ago, I couldn't find it. Upstream CI builds and pushes it on fly https://github.com/kubernetes/kubernetes/blob/a90cf500049d228f5ee4d217c4714028eb9c5f35/test/e2e/testing-manifests/auth/encrypt/run-e2e.sh#L37

I might be mistaken though.

test/library/encryption/kms/k8s-mock-plugin/build-from-k8s.sh
gangwgr: we used podman only as I remember

test/library/encryption/kms/k8s-mock-plugin/build-from-k8s.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
```
p0lyn0mial: I decided to slightly simplify the script, PTAL at https://github.com/openshift/library-go/pull/2089

test/library/encryption/kms/assets/daemonset.yaml
```diff
@@ -0,0 +1,109 @@
...
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
ardaguclu: nit:

```suggestion
        node-role.kubernetes.io/control-plane: ""
```

test/library/encryption/kms/assets/daemonset.yaml
```diff
@@ -0,0 +1,109 @@
...
+      priorityClassName: system-node-critical
+      serviceAccountName: kms-plugin
+      tolerations:
+        - operator: Exists
+      initContainers:
+        - name: init-softhsm
+          image: {{ .Image }}
```
ardaguclu: I assume this image will be just base image?

test/library/encryption/kms/assets/plugin_deployer.go
```diff
@@ -0,0 +1,369 @@
...
+		pods, err := kubeClient.CoreV1().Pods(config.Namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: KMSPluginLabelSelector,
+		})
+		if err != nil {
+			return false, nil
+		}
+
+		readyPods := 0
+		for _, pod := range pods.Items {
+			if pod.Status.Phase == corev1.PodRunning {
```
ardaguclu:
```suggestion
			if pod.Status.Phase != corev1.PodRunning {
        return false
    }
    for _, cond := range pod.Status.Conditions {
        if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
            return true
        }
    }
    return false
```

test/library/encryption/kms/assets/daemonset.yaml
gangwgr: yes

test/library/encryption/kms/assets/daemonset.yaml
ardaguclu: @gangwgr what about this comment?. Do we need this or not in our CI?

gangwgr: not needed it worked without hostNetwork

test/library/encryption/kms/assets/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
p0lyn0mial: @gangwgr please open a new PR for the manifests.

test/library/encryption/kms/assets/plugin_deployer.go
```diff
@@ -0,0 +1,377 @@
...
+	return KMSPluginConfig{
+		Namespace:         DefaultKMSPluginNamespace,
+		SocketPath:        DefaultKMSPluginSocketPath,
+		Image:             DefaultKMSPluginImage,
+		DeploymentTimeout: 5 * time.Minute,
+	}
+}
+
+// DeployKMSPlugin deploys the mock KMS v2 plugin using embedded YAML assets.
+func DeployKMSPlugin(t testing.TB, kubeClient kubernetes.Interface, config KMSPluginConfig) error {
```
p0lyn0mial: let's open a new PR for the deployer :)

test/library/encryption/scenarios.go
```diff
@@ -58,6 +86,9 @@ func TestEncryptionType(t *testing.T, scenario BasicScenario, provider configv1.
 		TestEncryptionTypeAESGCM(t, scenario)
 	case configv1.EncryptionTypeIdentity, "":
 		TestEncryptionTypeIdentity(t, scenario)
+	case configv1.EncryptionTypeKMS:
+		// For KMS, use TestEncryptionTypeKMS directly with KMSScenarioConfig
```
p0lyn0mial: why ?

test/library/encryption/scenarios.go
```diff
@@ -50,6 +50,34 @@ func TestEncryptionTypeAESGCM(t *testing.T, scenario BasicScenario) {
...
+	// KMSPluginDeployFunc deploys the KMS plugin before enabling KMS encryption.
+	// This function should deploy the plugin and wait for it to be ready.
+	// If nil, assumes KMS plugin is already deployed.
+	KMSPluginDeployFunc func(t testing.TB) error
+}
+
+// TestEncryptionTypeKMS tests that resources are encrypted using KMS.
+// This follows the same pattern as TestEncryptionTypeAESGCM but ensures
+// the KMS plugin is deployed first.
+func TestEncryptionTypeKMS(t *testing.T, scenario BasicScenario, kmsConfig KMSScenarioConfig) {
```
p0lyn0mial: this should be:

```
func TestEncryptionTypeKMS(t *testing.T, scenario BasicScenario) {
	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
	clientSet := SetAndWaitForEncryptionType(e, configv1.EncryptionTypeKMS, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
	scenario.AssertFunc(e, clientSet, configv1.EncryptionTypeKMS, scenario.Namespace, scenario.LabelSelector)
	AssertEncryptionConfig(e, clientSet, scenario.EncryptionConfigSecretName, scenario.EncryptionConfigSecretNamespace, scenario.TargetGRs)
}
```

test/library/encryption/kms_scenarios.go
```diff
@@ -0,0 +1,39 @@
+package encryption
```
p0lyn0mial: let's  remove this file, it is not needed.

test/library/encryption/scenarios.go
```diff
@@ -50,6 +50,13 @@ func TestEncryptionTypeAESGCM(t *testing.T, scenario BasicScenario) {
 	AssertEncryptionConfig(e, clientSet, scenario.EncryptionConfigSecretName, scenario.EncryptionConfigSecretNamespace, scenario.TargetGRs)
 }
 
+func TestEncryptionTypeKMS(t *testing.T, scenario BasicScenario) {
+	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	clientSet := SetAndWaitForEncryptionType(e, configv1.EncryptionTypeKMS, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
+	scenario.AssertFunc(e, clientSet, configv1.EncryptionTypeKMS, scenario.Namespace, scenario.LabelSelector)
+	AssertEncryptionConfig(e, clientSet, scenario.EncryptionConfigSecretName, scenario.EncryptionConfigSecretNamespace, scenario.TargetGRs)
```
ardaguclu, line 57: Should we update https://github.com/openshift/library-go/blob/11a9d18b8fdd34062c3b2e1fae008a83af256450/test/library/encryption/assertion.go#L31?

p0lyn0mial: yeah, i think we will have to.

so this PR lgtm overall but I think we should test it before merging.

@gangwgr can you open a new pr in kas-o where we add onoff scenario and pull:

- this PR
- arda's library-go PR


we need to wait until @bertinatto merges his changes (hostPath).

test/library/encryption/kms/assets/k8s_mock_kms_plugin_daemonset.yaml
```diff
@@ -19,8 +19,8 @@ spec:
       tolerations:
         - operator: Exists
       initContainers:
-        # Init container script adapted from upstream Kubernetes:
-        # https://github.com/kubernetes/kubernetes/blob/5a6a1fe908cfc64cf8ffe6264e8d830a85579ed9/staging/src/k8s.io/kms/internal/plugins/_mock/kms.yaml
+        # Init container installs SoftHSM on each node and imports the shared key.
```
ardaguclu: We need to revert the changes in this file.

PR #2086, opened 2026-01-23
CNTRLPLANE-2241: Add KMS encryption mode into library-go encryption controllers

ardaguclu:
This PR adds the implementation proposed in https://github.com/openshift/enhancements/pull/1900. 

This PR introduces KMS as a new mode in encryption controllers to be functioning along side with the other modes. Basically, the idea is to track the KMS configuration to detect any configuration changes. So that key_controller can create new Secret to initiate migration. KMS Secret will not contain any confidential data. It will simply store the KMS configuration. 

For Tech Preview v1, we will only support static unix domain socket (i.e. `unix:///var/run/kmsplugin/kms.sock`). Consequently, in this version, we don't support KMS -> KMS migrations.

Ref: Previous attemps:
* https://github.com/openshift/library-go/pull/2045
* https://github.com/openshift/library-go/pull/1876

Notes:
* The changes here will be backported to 4.21
* Substantial portion of the changed lines are unit and integration tests.

### DIFF
--- a/pkg/operator/encryption/secrets/secrets.go
+++ b/pkg/operator/encryption/secrets/secrets.go
@@ -68,7 +68,12 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 
 	keyMode := state.Mode(s.Annotations[encryptionSecretMode])
 	switch keyMode {
-	case state.AESCBC, state.AESGCM, state.SecretBox, state.Identity, state.KMS:
+	case state.AESCBC, state.AESGCM, state.SecretBox, state.Identity:
+		key.Mode = keyMode
+	case state.KMS:
+		if key.KMSConfiguration == nil {
+			return state.KeyState{}, fmt.Errorf("KMSConfiguration can not be nil, when mode is KMS")
+		}
 		key.Mode = keyMode
 	default:
 		return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid mode: %s", s.Namespace, s.Name, keyMode)


pkg/operator/encryption/crypto/keys.go
```diff
@@ -1,28 +1,39 @@
...
 }
+
+func NewKMSKey(externalKey []byte) []byte {
+	// Calculate MD5 checksum of KMS config for identification.
+	// MD5 provides short fixed length to be used in EncryptionConfiguration provider name
+	// e.g. kms-secrets-1-XUFAKrxLKna5cZnZEQH8Ug==
+	hash := md5.Sum(externalKey)
+	checksum := hash[:]
+	return checksum
+}
```
bertinatto: Given that we should be FIPS compatible, we might be flagged by security scanners. I think it's probably better to use something else

pkg/operator/encryption/kms/kms.go
```diff
@@ -0,0 +1,69 @@
...
+// FromProviderName extracts the key ID and KMS Secret from a KMS provider name.
+// Expected format: kms-{resourceName}-{keyID}-{keySecret}
+func FromProviderName(providerName string) (keyID string, kmsKey string, err error) {
+	matches := providerNameRegex.FindStringSubmatch(providerName)
+	if len(matches) != 4 {
+		return "", "", fmt.Errorf("provider name %q has invalid format, expected kms-{resource}-{keyID}-{checksumBase64}", providerName)
+	}
+
+	checksumBase64 := matches[3]
+	return matches[2], checksumBase64, nil
```
bertinatto: nit: you can probably just use `matches[3]` instead of creating a new var

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +210,25 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 					Keys: []apiserverconfigv1.Key{key.Key},
 				},
 			})
+		case state.KMS:
+			kmsConfig, err := kms.FromBytes(key.KMSConfig)
+			if err != nil {
+				klog.Infof("skipping provider as kms config is invalid: %v", err)
```
bertinatto: nit: I see this follows the existing pattern, but should we consider a warning or even error here?

pkg/operator/encryption/kms/kms.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// ToBytes serializes the KMS configuration to JSON.
+func (k *KMS) ToBytes() ([]byte, error) {
+	jsonBytes, err := json.Marshal(k)
+	if err != nil {
+		return nil, err
+	}
+	return jsonBytes, nil
+}
```
bertinatto: `return json.Marshal(k)`? Or perhaps keep it as is and wrap the error, since we do that in `FromBytes()`?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,22 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 					Mode: s,
 				}
 
+			case provider.KMS != nil:
+				s := state.KMS
+				keyID, kmsKey, err := kms.FromProviderName(provider.KMS.Name)
```
bertinatto: IIUC `provider.KMS.Name` is the [name of the KMS plugin](https://github.com/openshift/library-go/blob/99f27e58edae079609513b90251c93268a1c0602/vendor/k8s.io/apiserver/pkg/apis/apiserver/v1/types_encryption.go#L139), i.e., users can use whatever they want, right? If that's the case, the regex could break if we have a plugin name like `hashicorp-vaul`

pkg/operator/encryption/kms/kms.go
```diff
@@ -0,0 +1,69 @@
...
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"time"
+
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+)
+
+const (
+	DefaultEndpoint = "unix:///var/run/kmsplugin/kms.sock"
```
bertinatto: I guess this is OK for v1, but I wonder if we should make this more flexible in the future?

pkg/operator/encryption/crypto/keys.go
ardaguclu: That is a good point and I think you are right. Updated the code to use `sha256` by truncating first 16 (to have short length). Let me know your thoughts.

pkg/operator/encryption/kms/kms.go
ardaguclu: Updated.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: It should indeed be Warning to be compatible with `klog.Warningf("skipping invalid KMS provider %s: %v", provider.KMS.Name, err)`. Updated.

pkg/operator/encryption/kms/kms.go
ardaguclu: Wrapped the error as in `FromBytes()`.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Provider name of EncryptionConfiguration is totally our under control (managed by state_controller). We generate it in correct format in https://github.com/openshift/library-go/pull/2086/changes#diff-9aecf0d61d4bb32f2825a75a37e02e2c1b612c000427f3f576d42853ac8d23aeR224 to be parsed in here. Even if someone manually update it, state_controller will overwrite.

pkg/operator/encryption/kms/kms.go
ardaguclu: I agree with you. We definitely need to make this more flexible in the next version.

pkg/operator/encryption/crypto/keys.go
bertinatto: I think this is fine because we just want a constant-length string generated from random-length string while avoiding security scanner warnings

ardaguclu:
/hold cancel
I believe that finally this is ready for review

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -113,6 +117,10 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Annotations[EncryptionSecretMigratedResources] = string(bs)
 	}
 
+	if len(ks.KMSConfig) > 0 {
+		s.Annotations[EncryptionSecretKMSConfig] = string(ks.KMSConfig)
```
p0lyn0mial: Since v1 assumes the KMS configuration is always the same and static, and we don’t know what future v2 will look like.

I’m wondering if we could slightly simplify the code by always putting the well-known configuration in the annotation when the current mode is KMS. This could simplify the code and potential places we would have to change.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -341,6 +354,21 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 		return 0, "", false
 	}
 
+	if currentMode == state.KMS {
+		// We are here because Encryption Mode is not changed
+		secret := crypto.ModeToNewKeyFunc[state.KMS](kmsConfig)
+		if latestKey.Key.Secret != base64.StdEncoding.EncodeToString(secret) {
```
p0lyn0mial: i'm not sure that all changes/updates to a kms config should trigger a new key generation. rolling out a new key is expensive in terms of required shutdowns. also changing a timeout seems like an update that should not require a new key. something we need to figure out for v2.

ardaguclu: I completely agree with you. From the perspective of this line, it doesn't matter what `secret` includes. This line in any case need to compare the _secret_ to detect new key or not.

https://github.com/openshift/library-go/pull/2086/changes#diff-de68f84fb611b486200c440adbdd1997edb7ab5ffbc520e485b39f73b18d339eR296-R304 this is where we decide which data should trigger new key or not (by forming the source of the _secret_)

pkg/operator/encryption/secrets/secrets.go
ardaguclu: I'll investigate the possible implications tomorrow and get back to you. I think it would be simpler as you said.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: However, in v1, we don't support kms configurational changes. We always embed the same well-known endpoint. In order to keep the simplicity, I've removed secret comparison for KMS. Latest version of this PR embeds well known endpoint and keeps tracks of it.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Initially I tried to adopt generic approach that would work also for v2 (except the API changes). But I think starting simple by dropping the unsupported parts in v1 alleviates both of us cognitive load. 

Thus, generic KMSConfig is dropped, instead default endpoint is embedded. There isn't KMSConfig comparison anymore, because we don't support this already.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -256,7 +257,8 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 }
 
 func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, internalReason, externalReason string) (*corev1.Secret, error) {
-	bs := crypto.ModeToNewKeyFunc[currentMode]()
+	kmsConfig := []byte(kms.DefaultEndpoint)
```
ardaguclu: This should include the selective set of kms fields from APIServer config that will trigger new key. For now it is statically set.

pkg/operator/encryption/crypto/keys.go
```diff
@@ -2,27 +2,36 @@ package crypto
...
 }
 
-func NewIdentityKey() []byte {
+func NewIdentityKey(_ []byte) []byte {
 	return make([]byte, 16) // the key is not used to perform encryption but must be a valid AES key
 }
+
+func NewKMSKey(externalKey []byte) []byte {
+	// Used as fixed-length identifier in EncryptionConfiguration provider names.
+	// Example: kms-secrets-1-a7K9mJ2xH4nQ8pL5vR3wTg==
```
ardaguclu: We can use hexadecimal instead base64

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +210,21 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 					Keys: []apiserverconfigv1.Key{key.Key},
 				},
 			})
+		case state.KMS:
+			kmsConfig := string(key.KMSConfig)
```
ardaguclu: In v1, `key.KMSConfig` equals to well known endpoint.

pkg/operator/encryption/state/types.go
```diff
@@ -40,6 +40,8 @@ type KeyState struct {
 	InternalReason string
 	// the user via unsupportConfigOverrides.encryption.reason triggered this key.
 	ExternalReason string
+	// Encoded KMSConfig that stores the KMS related fields
+	KMSConfig []byte
```
p0lyn0mial: nit could this be renamed to `KMSConfiguration` ?

p0lyn0mial: could `KMSConfiguration` be a struct ? why []bytes ?


could `KMSConfiguration` be:
```
KMSConfiguration *apiserverconfigv1.KMSConfiguration
```
?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -266,6 +268,9 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 		InternalReason: internalReason,
 		ExternalReason: externalReason,
 	}
+	if currentMode == state.KMS {
+		ks.KMSConfig = kmsConfig
```
p0lyn0mial: could this be:
```
ks.KMSConfiguration = &apiserverv1.KMSConfiguration{
			APIVersion: "v2",
			Name:        fmt.Sprintf("%d", keyID),
			Endpoint:   kms.DefaultEndpoint,
			Timeout:    &metav1.Duration{Duration: kms.DefaultTimeout},
		}
```

?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -256,7 +257,8 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 }
 
 func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, internalReason, externalReason string) (*corev1.Secret, error) {
-	bs := crypto.ModeToNewKeyFunc[currentMode]()
+	kmsConfig := []byte(kms.DefaultEndpoint)
+	bs := crypto.ModeToNewKeyFunc[currentMode](kmsConfig)
```
p0lyn0mial: since the key is not used why it can't be 0 as for the identity ?

p0lyn0mial (commented):
thanks, left a few additional comments. ptal when you have time.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: It will definitely be used in v2. You mean we can remove it for v1, right?

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: KMSConfig stores Plugin specific configurations (although this is obsolete, I just reference https://github.com/openshift/api/blob/81371d13d1fcad175a48627cf11524a94a80c377/config/v1/types_kmsencryption.go#L36 to show an example of a configuration). So it is bound to APIServer config rather than `apiserverv1.KMSConfiguration`.

For instance KMS config will likely store AppRole, etc. for Vault in v2.

pkg/operator/encryption/state/types.go
ardaguclu: I tried to answer in https://github.com/openshift/library-go/pull/2086#discussion_r2772912065. KMSConfiguration refers to plugin specific configuration rather than `apiserverconfigv1.KMSConfiguration`

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We need to trigger a new migration, if `keyArn` (or `appRole`, etc.) is changed, even though these are not represented in `apiserverv1.KMSConfiguration`.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: > since the key is not used why it can't be 0 as for the identity ?

I think, for v1, we can remove indeed.

pkg/operator/encryption/state/types.go
ardaguclu: I store as bytes array to keep the generality. Because each plugin type will have their own specific fields.

pkg/operator/encryption/controllers/key_controller.go
p0lyn0mial: I would like to keep this pr simple for v1. 
Once we start developing v2 we might change the approach.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +210,21 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
...
+			provider := apiserverconfigv1.ProviderConfiguration{
+				// This generated KMSConfiguration can be convertible to KeyState,
+				// since provider name contains the key id and secret.
+				KMS: &apiserverconfigv1.KMSConfiguration{
```
p0lyn0mial: here, as far as i can tell a KMSConfig is actually `apiserverconfigv1.KMSConfiguration`.

as far as i can tell we will never store any additional data beyond `apiserverconfigv1.KMSConfiguration` .

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: yeah, agree. It makes sense. I'm updating.

pkg/operator/encryption/controllers/key_controller.go
p0lyn0mial: would it make sense to focus on v1 in this PR ? 
we might change the approach for v2 once we start prototyping.

as far as i can tell a KMSConfig is actually apiserverconfigv1.KMSConfiguration
xref: https://github.com/openshift/library-go/pull/2086/changes#r2773184416

p0lyn0mial: > keyArn (or appRole, etc.) is changed, even though these are not represented in apiserverv1.KMSConfiguration.

aren't these changes for a kms plugin ?

 i think we need to distinguish two configs: one for a kms plugin and the other one for kms configuration (encryption-cfg)

ardaguclu: You are right actually. We may even need a third one to track key_id from Status endpoint. But for v1, I think we can have one.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: I'm marking this PR as WIP to make necessary changes based on this ^^.

ardaguclu: PR has been updated to store `apiserverconfigv1.KMSConfiguration`.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,19 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 					Mode: s,
 				}
 
+			case provider.KMS != nil:
+				// Since in v1 we don't support kms -> kms migrations, we can use empty key.
+				// Because we don't need to compare secrets to detect change.
+				secret := crypto.ModeToNewKeyFunc[state.KMS]()
```
p0lyn0mial: would it make sense to define a var for it similar to `emptyStaticIdentityKey` ?(defined at the top of this file)

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,19 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
...
+				keyId := strings.Split(provider.KMS.Name, "-")[0]
+				// Name and Secret must match to find our backed Secret
+				ks = state.KeyState{
+					Key: apiserverconfigv1.Key{
+						Name:   keyId,
+						Secret: base64.StdEncoding.EncodeToString(secret),
+					},
+					Mode: state.KMS,
```
p0lyn0mial, line 123: shouldn't we also populate `ks.KMSConfiguration` ?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,19 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 					Mode: s,
 				}
 
+			case provider.KMS != nil:
+				// Since in v1 we don't support kms -> kms migrations, we can use empty key.
+				// Because we don't need to compare secrets to detect change.
+				secret := crypto.ModeToNewKeyFunc[state.KMS]()
+				keyId := strings.Split(provider.KMS.Name, "-")[0]
```
p0lyn0mial: could we create a private function for adding/removing the suffix ?

p0lyn0mial (commented):
left a few more questions. 
overall it lgtm

pkg/operator/encryption/crypto/keys.go
```diff
@@ -12,6 +12,7 @@ var (
 		state.AESGCM:    NewAES256Key,
 		state.SecretBox: NewAES256Key, // secretbox requires a 32 byte key so we can reuse the same function here
 		state.Identity:  NewIdentityKey,
+		state.KMS:       NewIdentityKey,
```
p0lyn0mial: would it make sense to rename `NewIdentityKey` to something more generic ?(now used by state.Identity and state.KSM)

pkg/operator/encryption/encryptionconfig/config.go
p0lyn0mial: I’m still wondering about encapsulation here. 
I think other parts of the code won’t even be aware that the name has changed. 

Only the code in encryptionconfig/config.go will encode and decode the new name, right? 

Do we need to ensure that the name doesn’t already contain "-"? 
For simplicity, we could add a comment in the key generation function describing this requirement.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: yes, it would be nicer. Updated.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Once we set correct keyId and secret to KeyState, backed secret will be found which includes any other field we need (including the `KMSConfiguration`). KeyState is enriched in here https://github.com/openshift/library-go/blob/b6adacbfccda0d306d5a725c554bee0319d26f20/pkg/operator/encryption/encryptionconfig/config.go#L115-L120.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: That would be better. Updated.

pkg/operator/encryption/crypto/keys.go
ardaguclu: Good point. Updated.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: > Only the code in encryptionconfig/config.go will encode and decode the new name, right?

That is correct. Rest of the code only deals with `KeyState`. They do not even know about the details of EncryptionConfiguration. 

> Do we need to ensure that the name doesn’t already contain "-"?

Kube-apiserver already fails, if it encounters a provider name without `-`. Because uniqueness has broken. 

> For simplicity, we could add a comment in the key generation function describing this requirement.

I've added a comment about the reasons behind this provider name generation mechanism. Please let me know your thoughts. Thanks.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +231,16 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
...
+	// Ideally we should have used keyId simply in kms provider name.
+	// However, this is an upstream constraint that every provider name must be unique.
+	// To maintain uniqueness while still allowing access to the keyId, we generate provider name in this format.
+	return fmt.Sprintf("%s-%s", keyID, resource)
+}
+
+func getKeyIDFromProviderName(providerName string) string {
+	// We just need to obtain the keyID to find our backed secret
+	// e.g. "1-secrets"
+	return strings.Split(providerName, "-")[0]
```
benluddy: How should this handle an unexpected provider name?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +205,14 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 					Keys: []apiserverconfigv1.Key{key.Key},
 				},
 			})
+		case state.KMS:
+			// In order to preserve the uniqueness, we should insert resource name
```
benluddy, line 216: OK, "kms v2 providers name must always be unique across all kms providers (v1 and v2)":

https://github.com/kubernetes/kubernetes/blob/870e2928bc1d9a48a8ef75289ea57ec6cdeb5cb3/staging/src/k8s.io/apiserver/pkg/apis/apiserver/validation/validation_encryption.go#L438-L444

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +231,16 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 
 	return providers
 }
+
+func createKMSProviderName(keyID, resource string) string {
+	// Ideally we should have used keyId simply in kms provider name.
+	// However, this is an upstream constraint that every provider name must be unique.
+	// To maintain uniqueness while still allowing access to the keyId, we generate provider name in this format.
```
benluddy: OK, this is at least less strange than smuggling identity names in phony AES-GCM provider configs. Later, I would be interested in coming up with alternatives that let the controllers correlate provider configs to their source secret without smuggling.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -341,6 +350,19 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
...
+	if currentMode == state.KMS {
+		// We are here because Encryption Mode is not changed
+
+		// For now in v1, we don't support configurational changes. Therefore,
+		// it is pointless comparing the secrets.
+
+		// For KMS mode, we don't do time-based rotation. Therefore, we shortcut here
+		// KMS keys are rotated externally by the KMS system.
+		// Moreover, we don't trigger new key when external reason is changed.
+		// Because it would lead to duplicate providers which is not allowed.
```
benluddy, line 365: IIUC, even if we did roll out a new encryption config in response to the "reason" update, we're not even capable of triggering a KEK rotation unless a specific provider happened to expose that capability independently.

pkg/operator/encryption/kms/helpers.go
```diff
@@ -2,12 +2,18 @@ package kms
...
 
 	"github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	DefaultEndpoint = "unix:///var/run/kmsplugin/kms.sock"
+	DefaultTimeout  = 10 * time.Second
+)
```
benluddy: Is there a real need to export these constants? IMO, duplicating the values in tests is good because it requires that changes in observable behavior are accompanied by changes to test expectations. And other code (doing related things like setting up volume mounts) should be observing effective values rather than using hardcoded defaults.

pkg/operator/encryption/encryptionconfig/config.go
benluddy: Are there going to be side effects from this naming scheme visible in things like metric labels?

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Although it may not be possible, it is better to handle the invalid provider name gracefully. I've updated the code. Please let me know your thoughts.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: All the metrics that is about provider_name will be affected https://github.com/kubernetes/kubernetes/blob/90a76aaa9afbe9cdb3df5f0a4356018b6637648b/staging/src/k8s.io/apiserver/pkg/storage/value/encrypt/envelope/metrics/metrics.go#L169.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: I agree with you. It was error-prone and not nice. Since now we carry KMSConfigurations in backed Secrets, I think we won't need to carry anything in provider_names.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Good point. That is correct.

pkg/operator/encryption/kms/helpers.go
ardaguclu: The only reason is to be accessed by key_controller. If we move these constants in there, we don't need to export them.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -266,6 +267,14 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 		InternalReason: internalReason,
 		ExternalReason: externalReason,
 	}
+	if currentMode == state.KMS {
+		ks.KMSConfiguration = &apiserverv1.KMSConfiguration{
+			APIVersion: "v2",
+			Name:       fmt.Sprintf("%d", keyID), // this will be updated by inserting resource name
```
p0lyn0mial: given that the name will be changed only in the `EncryptionConfiguration` and will not be visible to the `KeyState`, is this comment still relevant? 

maybe we should add a comment about the expected key format, which must include the `keyID` (at least for now)?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,22 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 					Mode: s,
 				}
 
+			case provider.KMS != nil:
+				// Name and Secret must match to find our backed Secret
+				keyId, err := getKeyIDFromProviderName(provider.KMS.Name)
+				if err != nil {
+					klog.Warningf("skipping invalid provider name %s: %v", provider.KMS.Name, err)
```
p0lyn0mial: should we also log the index and the resource ?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -106,6 +108,22 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
...
+					continue // should never happen
+				}
+				ks = state.KeyState{
+					Key: apiserverconfigv1.Key{
+						Name: keyId,
+						// Since in v1 we don't support kms -> kms migrations, we can use empty key.
```
p0lyn0mial: Not sure if I understand this comment. 

For KMS the Secret doesn't matter and can/will always be empty - because the actual encryption key material is in the external KMS instance.

pkg/operator/encryption/encryptionconfig/config.go
p0lyn0mial: ok, thanks.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +210,14 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 					Keys: []apiserverconfigv1.Key{key.Key},
 				},
 			})
+		case state.KMS:
+			// In order to preserve the uniqueness, we should insert resource name
+			kmsCopy := key.KMSConfiguration.DeepCopy()
```
p0lyn0mial, line 217: yes, i think we need a DeepCopy here since this method is called per resource.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +236,20 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 
 	return providers
 }
+
+func createKMSProviderName(keyID, resource string) string {
+	// Ideally we should have used keyId simply in kms provider name.
+	// However, this is an upstream constraint that every provider name must be unique.
+	// To maintain uniqueness while still allowing access to the keyId, we generate provider name in this format.
```
p0lyn0mial: nit: the comment on the `getKeyIDFromProviderName` uses `keyID` notation thus i would rename `keyId`

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +236,20 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
...
+	return fmt.Sprintf("%s-%s", keyID, resource)
+}
+
+func getKeyIDFromProviderName(providerName string) (string, error) {
+	// We just need to obtain the keyID to find our backed secret
+	// e.g. "1-secrets"
+	parsed := strings.Split(providerName, "-")
+	if len(parsed) != 2 {
+		return "", fmt.Errorf("invalid provider name: %s", providerName)
```
p0lyn0mial: should why explain why it is invalid ?

pkg/operator/encryption/kms/helpers.go
p0lyn0mial: could we just inline these in key_controller?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +236,20 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 
 	return providers
 }
+
+func createKMSProviderName(keyID, resource string) string {
+	// Ideally we should have used keyId simply in kms provider name.
+	// However, this is an upstream constraint that every provider name must be unique.
+	// To maintain uniqueness while still allowing access to the keyId, we generate provider name in this format.
+	return fmt.Sprintf("%s-%s", keyID, resource)
```
p0lyn0mial: can kube resources contain "-" (e.g. CRDs) ? If yes, than should we adopt a different naming scheme here ?

pkg/operator/encryption/controllers/state_controller_test.go
```diff
@@ -800,8 +1202,10 @@ func validateSecretWithEncryptionConfig(actualSecret *corev1.Secret, expectedEnc
 		return fmt.Errorf("failed to verfy the encryption config, due to %v", err)
 	}
 
-	if !equality.Semantic.DeepEqual(expectedEncryptionCfg, actualEncryptionCfg) {
-		return fmt.Errorf("%s", diff.Diff(expectedEncryptionCfg, actualEncryptionCfg))
+	// Ignore KMS Name field since it contains a random suffix
+	ignoreKMSName := cmpopts.IgnoreFields(apiserverconfigv1.KMSConfiguration{}, "Name")
```
p0lyn0mial: do we have to ignore checking the kms plugin name ? is there a way to create it from the existing data ?

p0lyn0mial: isn't the name deterministic ?

pkg/operator/encryption/encryptionconfig/config_test.go
```diff
@@ -578,6 +682,21 @@ func keyToAESConfiguration(key *corev1.Secret) *apiserverconfigv1.AESConfigurati
...
 }
 
+func keyToKMSConfiguration(key *corev1.Secret, resource string) *apiserverconfigv1.KMSConfiguration {
+	keyID, ok := state.NameToKeyID(key.Name)
+	if !ok {
+		panic(fmt.Sprintf("invalid test secret name %q", key.Name))
+	}
+	return &apiserverconfigv1.KMSConfiguration{
+		APIVersion: "v2",
+		Name:       fmt.Sprintf("%d-test%d", keyID, keyID),
```
p0lyn0mial: could this produce the name the prod code is going to produce ?

test/e2e-encryption/encryption_test.go
```diff
@@ -719,3 +815,9 @@ func (p *provider) EncryptedGRs() []schema.GroupResource {
 func (p *provider) ShouldRunEncryptionControllers() (bool, error) {
 	return true, nil
 }
+
+// kmsProviderName computes the KMS provider name with checksum for a given resource and key ID.
```
p0lyn0mial: checksum ? (outdated ?)

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -478,6 +722,12 @@ func TestGetCurrentModeAndExternalReason(t *testing.T) {
 			name:             "reading empty config works",
 			apiServerObjects: []runtime.Object{&configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}}},
 		},
+
+		{
+			name:              "kms encryption mode",
+			apiServerObjects:  []runtime.Object{&configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}, Spec: configv1.APIServerSpec{Encryption: configv1.APIServerEncryption{Type: "KMS"}}}},
+			expectedKMSConfig: []byte("unix:///var/run/kmsplugin/kms.sock"),
```
p0lyn0mial: does this test validate this field ?

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Comment is not relevant anymore. I've removed it.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: In v2, we'll have to compare this field (or KMSConfiguration annotation) to detect configurational changes. But it sounds like this comment creates confusions. So I've removed it.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Updated the error message

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Nice catch!. CRDs can contain "-" which may create a problem for us in the long term. It appears that resource names can not contain "_" (underscore) in kubernetes. Hence, I've used underscore in provider name (e.g. `1_secrets`, `2_configmaps`) and updated the rest of the code with this.

pkg/operator/encryption/controllers/state_controller_test.go
ardaguclu: This was forgotten from the previous version. I've removed IgnoreFields.

ardaguclu: This was forgotten from the previous version. I've removed IgnoreFields.

pkg/operator/encryption/encryptionconfig/config_test.go
ardaguclu: This was forgotten from the previous version. I've updated the name.

pkg/operator/encryption/controllers/key_controller_test.go
ardaguclu: This was forgotten from the previous version. I've removed this field.

ardaguclu:
Unrelated?
/test unit

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +234,20 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
...
+	// Ideally we should have used keyID simply in kms provider name.
+	// However, this is an upstream constraint that every provider name must be unique.
+	// To maintain uniqueness while still allowing access to the keyID, we generate provider name in this format.
+	return fmt.Sprintf("%s_%s", keyID, resource)
+}
+
+func getKeyIDFromProviderName(providerName string) (string, error) {
+	// We just need to obtain the keyID to find our backed secret
+	// e.g. "1_secrets"
+	parsed := strings.Split(providerName, "_")
```
p0lyn0mial: should we use `strings.SplitN` ?

pkg/operator/encryption/testing/helpers.go
```diff
@@ -94,6 +95,30 @@ func CreateExpiredMigratedEncryptionKeySecretWithRawKey(targetNS string, grs []s
 	return CreateMigratedEncryptionKeySecretWithRawKey(targetNS, grs, keyID, rawKey, time.Now().Add(-(time.Hour*24*7 + time.Hour)))
 }
 
+func CreateEncryptionKeySecretWithKMSConfig(targetNS string, grs []schema.GroupResource, keyID uint64) *corev1.Secret {
+	emptyKey := make([]byte, 16)
+	secret := CreateEncryptionKeySecretWithRawKeyWithMode(targetNS, grs, keyID, emptyKey, "KMS")
+	kmsConfig := &apiserverconfigv1.KMSConfiguration{
+		APIVersion: "v2",
+		Name:       fmt.Sprintf("%d_%s", keyID, grs[0].Resource),
```
p0lyn0mial: shouldn't this be `fmt.Sprintf("%d", keyID)` ? 
the resource suffix is added later.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -192,6 +208,14 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 					Keys: []apiserverconfigv1.Key{key.Key},
 				},
 			})
+		case state.KMS:
+			// In order to preserve the uniqueness, we should insert resource name
+			kmsCopy := key.KMSConfiguration.DeepCopy()
```
p0lyn0mial, line 217: should we check `key.KMSConfiguration` for `nil` ?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -113,6 +121,14 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Annotations[EncryptionSecretMigratedResources] = string(bs)
 	}
 
+	if ks.KMSConfiguration != nil {
+		ks, err := json.Marshal(ks.KMSConfiguration)
```
p0lyn0mial: nit: `ks` shadows the outer `ks`, we could rename to `ksJSON`

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Although this is not possible, it makes sense gracefully handling the empty KMSConfiguration by skipping it with a message. Please let me know your thoughts.

p0lyn0mial: yes, that would match the existing pattern. good idea.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -209,6 +209,10 @@ func stateToProviders(resource string, desired state.GroupResourceState) []apise
 				},
 			})
 		case state.KMS:
+			if key.KMSConfiguration == nil {
+				klog.Infof("skipping key %s in KMS mode as its KMSConfiguration is nil", key.Key.Name)
```
p0lyn0mial: should we add resource and index too?

ardaguclu: I've added resource name in the log. 

Index is not the index in EncryptionConfiguration providers. It is an index in the Backed Secrets. We have keyID already as discriminator. Would you still want me to insert index name?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -210,3 +238,20 @@ func stateToProviders(desired state.GroupResourceState) []apiserverconfigv1.Prov
 
 	return providers
 }
+
+func createKMSProviderName(keyID, resource string) string {
```
p0lyn0mial, line 242: just a comment: as of today there a no restrictions on length for the provider name. the only restriction i found is that we are not allowed to use `:`

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -341,6 +353,19 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 		return 0, "", false
 	}
 
+	if currentMode == state.KMS {
+		// We are here because Encryption Mode is not changed
+
+		// For now in v1, we don't support configurational changes. Therefore,
```
bertinatto: optional: someone reading `v1` might get confused, maybe it's better to specify TP as well

pkg/operator/encryption/state/types.go
```diff
@@ -60,6 +62,7 @@ const (
 	AESGCM    Mode = "aesgcm"
 	SecretBox Mode = "secretbox" // available from the first release, see defaultMode below
 	Identity  Mode = "identity"  // available from the first release, see defaultMode below
+	KMS       Mode = "KMS"       // only supports KMS v2
```
bertinatto, line 65: nit: all the others are lowercase (`"KMS"`)

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -58,9 +58,17 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
...
+		if err := json.Unmarshal([]byte(v), kmsConfiguration); err != nil {
+			return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s annotation: %v", s.Namespace, s.Name, EncryptionSecretKMSConfig, err)
+		}
+		key.KMSConfiguration = kmsConfiguration
+	}
+
 	keyMode := state.Mode(s.Annotations[encryptionSecretMode])
 	switch keyMode {
-	case state.AESCBC, state.AESGCM, state.SecretBox, state.Identity:
+	case state.AESCBC, state.AESGCM, state.SecretBox, state.Identity, state.KMS:
```
bertinatto: Do you also need to add a check to test if `keyMode` is KMS and `key.KMSConfiguration` is `nil`?

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: You are right, it is confusing. So I've updated by mentioning about "Tech Preview v1".

pkg/operator/encryption/state/types.go
ardaguclu: Unfortunately this definition comes from https://github.com/openshift/api/blob/43090db3afe48e7dff0f6e39a866474d15cd0db5/config/v1/types_apiserver.go#L234 API. I agree that it should be lowercase, but we can't change it without API changes.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: This function (i.e. `ToKeyState`) is pure conversion from Secret to KeyState without any logic. The caller checks the emptiness of KMSConfiguration in https://github.com/openshift/library-go/pull/2086/changes#diff-9aecf0d61d4bb32f2825a75a37e02e2c1b612c000427f3f576d42853ac8d23aeR212-R215 (and silently ignores it). In my opinion, it would be better to leave the decision to user of KMSConfiguration (i.e. `stateToProviders`).

ardaguclu: We can add a check in here though. Once we add it, since `ToKeyState` is used by all controllers, any misconfiguration (KMS mode is enabled, but somehow KMSConfiguration is nil) will cause the failure of the controllers. On the other hand, this condition must not happen anyway. I'd like to hear opinions from @p0lyn0mial

p0lyn0mial: We can be defensive and add additional checks. What kind of checks would we like to add ?

ardaguclu: Perhaps something like this;

```go
	case state.KMS:
		if key.KMSConfiguration == nil {
			return state.KeyState{}, fmt.Errorf("KMSConfiguration can not be nil, when mode is KMS")
		}
		key.Mode = keyMode
```

ardaguclu: Updated. Please let me know your opinions.

PR #2087, opened 2026-01-23
NO-JIRA: Enable test-e2e-encryption

ardaguclu:
It appears that [integration tests for encryption controllers](https://github.com/openshift/library-go/blob/master/test/e2e-encryption/encryption_test.go) are never being executed ([example job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_library-go/2086/pull-ci-openshift-library-go-master-e2e-aws-encryption/2014621776604041216/artifacts/e2e-aws-encryption/test/build-log.txt)). This PR updates the Makefile target to enable them.

### DIFF
--- a/test/e2e-encryption/encryption_test.go
+++ b/test/e2e-encryption/encryption_test.go
@@ -12,11 +12,12 @@ import (
 
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -25,7 +26,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/yaml"
 
@@ -67,11 +67,11 @@ func TestEncryptionIntegration(tt *testing.T) {
 	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
 	require.NoError(t, err)
 	kubeInformers := v1helpers.NewKubeInformersForNamespaces(kubeClient, "openshift-config-managed")
-	apiextensionsClient, err := v1beta1.NewForConfig(kubeConfig)
+	apiextensionsClient, err := v1.NewForConfig(kubeConfig)
 	require.NoError(t, err)
 
 	// create ExtensionTest operator CRD
-	var operatorCRD apiextensionsv1beta1.CustomResourceDefinition
+	var operatorCRD apiextensionsv1.CustomResourceDefinition
 	require.NoError(t, yaml.Unmarshal([]byte(encryptionTestOperatorCRD), &operatorCRD))
 	crd, err := apiextensionsClient.CustomResourceDefinitions().Create(ctx, &operatorCRD, metav1.CreateOptions{})
 	if errors.IsAlreadyExists(err) {
@@ -80,6 +80,23 @@ func TestEncryptionIntegration(tt *testing.T) {
 		require.NoError(t, err)
 	}
 	defer apiextensionsClient.CustomResourceDefinitions().Delete(ctx, crd.Name, metav1.DeleteOptions{})
+
+	t.Logf("Waiting for CRD to be ready")
+	err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
+		oCRD, crdErr := apiextensionsClient.CustomResourceDefinitions().Get(ctx, operatorCRD.Name, metav1.GetOptions{})
+		if crdErr != nil {
+			return false, crdErr
+		}
+
+		for _, condition := range oCRD.Status.Conditions {
+			if condition.Type == apiextensionsv1.Established {
+				operatorCRD = *oCRD
+				return condition.Status == apiextensionsv1.ConditionTrue, nil
+			}
+		}
+		return false, nil
+	})
+	require.NoError(t, err)
 
 	// create operator client and create instance with ManagementState="Managed"
 	operatorGVR := schema.GroupVersionResource{Group: operatorCRD.Spec.Group, Version: "v1", Resource: operatorCRD.Spec.Names.Plural}
@@ -99,7 +116,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 	require.NoError(t, err)
 	dynamicClient, err := dynamic.NewForConfig(kubeConfig)
 	require.NoError(t, err)
-	err = wait.PollImmediate(time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+	err = wait.PollUntilContextTimeout(ctx, time.Second, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
 		_, err := dynamicClient.Resource(operatorGVR).Create(ctx, &unstructured.Unstructured{
 			Object: map[string]interface{}{
 				"apiVersion": "operator.openshift.io/v1",
@@ -121,7 +138,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 	require.NoError(t, err)
 
 	// create APIServer clients
-	fakeConfigClient := configv1clientfake.NewSimpleClientset(&configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}})
+	fakeConfigClient := configv1clientfake.NewClientset(&configv1.APIServer{ObjectMeta: metav1.ObjectMeta{Name: "cluster"}})
 	fakeConfigInformer := configv1informers.NewSharedInformerFactory(fakeConfigClient, 10*time.Minute)
 	fakeApiServerClient := fakeConfigClient.ConfigV1().APIServers()
 
@@ -158,6 +175,26 @@ func TestEncryptionIntegration(tt *testing.T) {
 	kubeInformers.Start(stopCh)
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
+
+	t.Logf("Waiting for informers to sync")
+	for gvr, synced := range operatorInformer.WaitForCacheSync(stopCh) {
+		if !synced {
+			t.Fatalf("informer for %v not synced", gvr)
+		}
+	}
+	for gvr, synced := range fakeConfigInformer.WaitForCacheSync(stopCh) {
+		if !synced {
+			t.Fatalf("informer for %v not synced", gvr)
+		}
+	}
+	for ns, nsSynced := range kubeInformers.WaitForCacheSync(stopCh) {
+		for gvr, synced := range nsSynced {
+			if !synced {
+				t.Fatalf("informer for %v in namespace %v not synced", gvr, ns)
+			}
+		}
+	}
+	t.Logf("informers are sync'ed")
 
 	waitForConfigEventuallyCond := func(cond func(s string) bool) {
 		t.Helper()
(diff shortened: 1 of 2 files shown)


test/e2e-encryption/encryption_test.go
```diff
@@ -378,16 +395,16 @@ func TestEncryptionIntegration(tt *testing.T) {
 	t.Logf("Delete the openshift-config-managed config")
 	deployer.DeleteOperandConfig()
 	waitForConfigs(
-		// 7 is migrated and hence only one needed, but we rotate through identity
-		"kubeapiservers.operator.openshift.io=identity,aescbc:7;kubeschedulers.operator.openshift.io=identity,aescbc:7",
-		// 7 is migrated, plus one backed key (5). 6 is deleted, and therefore is not preserved (would be if the operand config was not deleted)
+		// 7 is migrated, backed key (5) is immediately included when rotating through identity. 6 is deleted, not preserved (would be if operand config was not deleted)
+		"kubeapiservers.operator.openshift.io=identity,aescbc:7,aescbc:5;kubeschedulers.operator.openshift.io=identity,aescbc:7,aescbc:5",
```
ardaguclu, line 409: This one is missed in https://github.com/openshift/library-go/pull/615

test/e2e-encryption/encryption_test.go
```diff
@@ -84,6 +89,19 @@ func TestEncryptionIntegration(tt *testing.T) {
 	// create operator client and create instance with ManagementState="Managed"
 	operatorGVR := schema.GroupVersionResource{Group: operatorCRD.Spec.Group, Version: "v1", Resource: operatorCRD.Spec.Names.Plural}
 	operatorGVK := schema.GroupVersionKind{Group: operatorCRD.Spec.Group, Version: "v1", Kind: operatorCRD.Spec.Names.Kind}
+	dynamicClient, err := dynamic.NewForConfig(kubeConfig)
+	require.NoError(t, err)
+
+	// wait for CRD to be ready by polling the resource
+	t.Logf("Waiting for CRD to be ready")
+	err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, EncryptionTestTimeout, true, func(ctx context.Context) (bool, error) {
```
p0lyn0mial: why not to use `wait.ForeverTestTimeout` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -84,6 +89,19 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+		_, err := dynamicClient.Resource(operatorGVR).List(ctx, metav1.ListOptions{})
```
p0lyn0mial: how does it ensure the CRD is ready ?

p0lyn0mial: should we get the CRD and check some condition ?

test/e2e-encryption/encryption_test.go
```diff
@@ -97,9 +115,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 
 	operatorClient, operatorInformer, err := genericoperatorclient.NewClusterScopedOperatorClient(clk, kubeConfig, operatorGVR, operatorGVK, extractApplySpec, extractApplyStatus)
 	require.NoError(t, err)
-	dynamicClient, err := dynamic.NewForConfig(kubeConfig)
-	require.NoError(t, err)
-	err = wait.PollImmediate(time.Second, wait.ForeverTestTimeout, func() (bool, error) {
+	err = wait.PollUntilContextTimeout(ctx, time.Second, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
```
p0lyn0mial, line 119: here we use `wait.ForeverTestTimeout` :)

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +175,17 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	// wait for operator client to be able to read the resource
+	t.Logf("Waiting for operator client to sync")
```
p0lyn0mial: is this waiting for an informer to sync ?

if yes, should we wait until the informer is synced ?

test/e2e-encryption/encryption_test.go
```diff
@@ -45,6 +45,9 @@ import (
 	"github.com/openshift/library-go/test/library"
 )
 
+// Encryption controllers run every minute. 2 minutes is sufficient for the required condition to be met.
+var EncryptionTestTimeout = 2 * time.Minute
```
p0lyn0mial: why not to use `wait.ForeverTestTimeout` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -297,15 +318,11 @@ func TestEncryptionIntegration(tt *testing.T) {
...
-			}
-			spec.UnsupportedConfigOverrides.Raw = []byte(fmt.Sprintf(`{"encryption":{"reason":%q}}`, reason))
-			_, _, err = operatorClient.UpdateOperatorSpec(context.TODO(), rv, spec)
-			return err
-		})
+		applyConfig := applyoperatorv1.OperatorSpec().
+			WithUnsupportedConfigOverrides(runtime.RawExtension{
+				Raw: []byte(fmt.Sprintf(`{"encryption":{"reason":%q}}`, reason)),
+			})
+		err = operatorClient.ApplyOperatorSpec(context.TODO(), "encryption-test", applyConfig)
```
p0lyn0mial: don't we have a `ctx` already defined ?

test/e2e-encryption/encryption_test.go
p0lyn0mial: yeah, i think you are right.

test/e2e-encryption/encryption_test.go
ardaguclu: Good point. Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: Yes, it is better to check the established condition of the CRD. Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: I defined this timeout due to the fact that controllers run per minute https://github.com/openshift/library-go/blob/d07df3e9f6043b6d36b263841cba56e11d2ce89b/pkg/operator/encryption/controllers/key_controller.go#L110 to give sufficient time for the expected conditions to be met. However, I believe that if we wait for the sync'ed signal of informers, we won't need this. So that we can use `wait.ForeverTestTimeout` (30 seconds). I've updated the PR with that.

test/e2e-encryption/encryption_test.go
```diff
@@ -81,6 +81,26 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+	err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
+		oCRD, crdErr := apiextensionsClient.CustomResourceDefinitions().Get(ctx, operatorCRD.Name, metav1.GetOptions{})
+		if crdErr != nil {
+			if errors.IsNotFound(crdErr) {
+				return false, nil
+			}
+			return false, crdErr
+		}
+
+		for _, condition := range oCRD.Status.Conditions {
```
p0lyn0mial, line 91: could we use `apiextensions.IsCRDConditionTrue(oCRD, apiextensions.Established)` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -81,6 +81,26 @@ func TestEncryptionIntegration(tt *testing.T) {
 	}
 	defer apiextensionsClient.CustomResourceDefinitions().Delete(ctx, crd.Name, metav1.DeleteOptions{})
 
+	t.Logf("Waiting for CRD to be ready")
+	err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
+		oCRD, crdErr := apiextensionsClient.CustomResourceDefinitions().Get(ctx, operatorCRD.Name, metav1.GetOptions{})
+		if crdErr != nil {
+			if errors.IsNotFound(crdErr) {
```
p0lyn0mial: A CRD after creation must be persisted in etcd. I think we can fail on any error.

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +179,20 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	t.Logf("Waiting for operator informer to sync")
+	if !cache.WaitForCacheSync(stopCh, operatorClient.Informer().HasSynced) {
```
p0lyn0mial: should we `infSynced := operatorInformer.WaitForCacheSync(stopCh)` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +179,20 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+		t.Fatalf("failed to sync operator informer")
+	}
+	kubeInformers.WaitForCacheSync(stopCh)
```
p0lyn0mial: `kubeInformers := operatorInformer.WaitForCacheSync(stopCh)` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +179,20 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	t.Logf("Waiting for operator informer to sync")
```
p0lyn0mial: should we also wait for `fakeConfigInformer` ?

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +179,20 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+	if !cache.WaitForCacheSync(stopCh, operatorClient.Informer().HasSynced) {
+		t.Fatalf("failed to sync operator informer")
+	}
+	kubeInformers.WaitForCacheSync(stopCh)
+	err = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond, wait.ForeverTestTimeout, true, func(ctx context.Context) (bool, error) {
+		_, _, _, err := operatorClient.GetOperatorState()
```
p0lyn0mial: is this required ?

test/e2e-encryption/encryption_test.go
ardaguclu: It appears that `IsCRDConditionTrue` accepts internal CustomResourceDefinition but in our case, we have external CustomResourceDefinition.

test/e2e-encryption/encryption_test.go
ardaguclu: Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: Updated by using `operatorInformer.WaitForCacheSync(stopCh)`. Please let me know your opinions about this. Should I use returned value (`infSynced`) in somewhere?

test/e2e-encryption/encryption_test.go
ardaguclu: Shouldn't we wait `kubeInformers` sync'ed rather than`operatorInformer`?

test/e2e-encryption/encryption_test.go
ardaguclu: Updated.

test/e2e-encryption/encryption_test.go
ardaguclu: Yeah, I think it becomes redundant. Removed

test/e2e-encryption/encryption_test.go
p0lyn0mial: I think we should wait for all factories.

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +176,12 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	t.Logf("Waiting for informers to sync")
+	operatorInformer.WaitForCacheSync(stopCh)
```
p0lyn0mial: i think that we should check the response from `WaitForCacheSync` otherwise we don't know if an informer synced :)

ardaguclu: Good point :). I've updated the PR accordingly.

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +176,22 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	t.Logf("Waiting for informers to sync")
+	if !operatorInformer.WaitForCacheSync(stopCh)[operatorGVR] {
```
p0lyn0mial: could we just simply check all returned gvrs ? 
effectively these are the ones used by the test.

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +176,22 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+		t.Fatalf("informer for %v not synced", operatorGVR)
+	}
+	for typ, synced := range fakeConfigInformer.WaitForCacheSync(stopCh) {
+		if !synced {
+			t.Fatalf("informer for %v not synced", typ)
+		}
+	}
+	for typ, synced := range kubeInformers.WaitForCacheSync(stopCh)["openshift-config-managed"] {
```
p0lyn0mial: could we just simply check all returned gvrs ? 
effectively these are the ones used by the test.

test/e2e-encryption/encryption_test.go
```diff
@@ -159,6 +176,22 @@ func TestEncryptionIntegration(tt *testing.T) {
 	operatorInformer.Start(stopCh)
 	go controllers.Run(ctx, 1)
 
+	t.Logf("Waiting for informers to sync")
+	if !operatorInformer.WaitForCacheSync(stopCh)[operatorGVR] {
+		t.Fatalf("informer for %v not synced", operatorGVR)
+	}
+	for typ, synced := range fakeConfigInformer.WaitForCacheSync(stopCh) {
```
p0lyn0mial: nit: could we rename `typ` to `grv` ?

p0lyn0mial:
/hold 

for `ci/prow/e2e-aws-encryption`

PR #2090, opened 2026-01-26
CNTRLPLANE-2241: kms: add helper for TechPreview V1

bertinatto:
/assign @ardaguclu @benluddy @p0lyn0mial 

Proof PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2015

### DIFF
--- a/pkg/operator/encryption/kms/helpers.go
+++ b/pkg/operator/encryption/kms/helpers.go
@@ -1,0 +1,66 @@
+package kms
+
+import (
+	"fmt"
+
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AddKMSPluginVolumeAndMountToPodSpec conditionally adds the KMS plugin volume mount to the specified container.
+// It assumes the pod spec does not already contain the KMS volume or mount; no deduplication is performed.
+// Deprecated: this is a temporary solution to get KMS TP v1 out. We should come up with a different approach afterwards.
+func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName string, featureGateAccessor featuregates.FeatureGateAccess) error {
+	if podSpec == nil {
+		return fmt.Errorf("pod spec cannot be nil")
+	}
+
+	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
+		return nil
+	}
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get feature gates: %w", err)
+	}
+
+	if !featureGates.Enabled(features.FeatureGateKMSEncryption) {
+		return nil
+	}
+
+	containerIndex := -1
+	for i, container := range podSpec.Containers {
+		if container.Name == containerName {
+			containerIndex = i
+			break
+		}
+	}
+
+	if containerIndex < 0 {
+		return fmt.Errorf("container %s not found", containerName)
+	}
+
+	container := &podSpec.Containers[containerIndex]
+	container.VolumeMounts = append(container.VolumeMounts,
+		corev1.VolumeMount{
+			Name:      "kms-plugin-socket",
+			MountPath: "/var/run/kmsplugin",
+		},
+	)
+
+	directoryOrCreate := corev1.HostPathDirectoryOrCreate
+	podSpec.Volumes = append(podSpec.Volumes,
+		corev1.Volume{
+			Name: "kms-plugin-socket",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/var/run/kmsplugin",
+					Type: &directoryOrCreate,
+				},
+			},
+		},
+	)
+
+	return nil
+}
--- a/pkg/operator/encryption/kms/helpers_test.go
+++ b/pkg/operator/encryption/kms/helpers_test.go
@@ -1,0 +1,197 @@
+package kms
+
+import (
+	"errors"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddKMSPluginVolume(t *testing.T) {
+	directoryOrCreate := corev1.HostPathDirectoryOrCreate
+
+	tests := []struct {
+		name                string
+		featureGateAccessor featuregates.FeatureGateAccess
+		actualPodSpec       *corev1.PodSpec
+		expectedPodSpec     *corev1.PodSpec
+		containerName       string
+		expectError         bool
+	}{
+		{
+			name: "nil pod: returns error",
+			featureGateAccessor: featuregates.NewHardcodedFeatureGateAccess(
+				[]configv1.FeatureGateName{features.FeatureGateKMSEncryption},
+				[]configv1.FeatureGateName{},
+			),
+			actualPodSpec:   nil,
+			expectedPodSpec: nil,
+			containerName:   "kube-apiserver",
+			expectError:     true,
+		},
+		{
+			name: "container not found: returns error",
+			featureGateAccessor: featuregates.NewHardcodedFeatureGateAccess(
+				[]configv1.FeatureGateName{features.FeatureGateKMSEncryption},
+				[]configv1.FeatureGateName{},
+			),
+			actualPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "other-container"},
+				},
+			},
+			expectedPodSpec: nil,
+			containerName:   "kube-apiserver",
+			expectError:     true,
+		},
+		{
+			name: "feature gates not observed: no volume added",
+			featureGateAccessor: featuregates.NewHardcodedFeatureGateAccessForTesting(
+				nil,
+				nil,
+				make(chan struct{}),
+				nil,
+			),
+			actualPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+			},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+			},
+			containerName: "kube-apiserver",
+			expectError:   false,
+		},
+		{
+			name: "feature gate accessor error: returns error",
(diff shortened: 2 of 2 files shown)


pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+					Type: &directoryOrCreate,
+				},
+			},
+		},
+	)
+
+	for i, container := range pod.Spec.Containers {
+		if container.Name != "kube-apiserver" {
+			continue
+		}
```
benluddy: Should container name or a selection function be passed in as a parameter so that this will work for openshift-apiserver and oauth-apiserver?

bertinatto: It should! Fixed

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+				},
+			},
+		},
+	)
+
+	for i, container := range pod.Spec.Containers {
+		if container.Name != containerName {
+			continue
+		}
+		pod.Spec.Containers[i].VolumeMounts = append(container.VolumeMounts,
```
p0lyn0mial: should we check if the mount doesn't already exist ? :) 

otherwise we might add it multiple times.
also worth adding a test case for this ^

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
+		return nil
+	}
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get feature gates: %w", err)
+	}
+
+	if !featureGates.Enabled(features.FeatureGateKMSEncryptionProvider) {
```
p0lyn0mial: we can change it to `KMSEncryption` now.

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil
+	}
+
+	directoryOrCreate := corev1.HostPathDirectoryOrCreate
+	pod.Spec.Volumes = append(pod.Spec.Volumes,
```
p0lyn0mial: should we check if the volume doesn't already exist ? :)

otherwise we might add it multiple times.
also worth adding a test case for this ^

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get feature gates: %w", err)
+	}
+
+	if !featureGates.Enabled(features.FeatureGateKMSEncryptionProvider) {
+		return nil
+	}
+
```
p0lyn0mial, line 31: what if `containerName` doesn't exist ?
should we check for existence before doing any modifications ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AddKMSPluginVolumeToPod conditionally adds the KMS plugin volume mount to the kube-apiserver container.
+// FIXME: this is a temporary solution to get KMS TP v1 out. We should come up with a different approach afterwards.
+func AddKMSPluginVolumeToPod(pod *corev1.Pod, containerName string, featureGateAccessor featuregates.FeatureGateAccess) error {
+	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
+		return nil
```
p0lyn0mial, line 20: i think this is what we usually return when FG hasn't been observed, right ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AddKMSPluginVolumeToPod conditionally adds the KMS plugin volume mount to the kube-apiserver container.
+// FIXME: this is a temporary solution to get KMS TP v1 out. We should come up with a different approach afterwards.
+func AddKMSPluginVolumeToPod(pod *corev1.Pod, containerName string, featureGateAccessor featuregates.FeatureGateAccess) error {
+	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
```
p0lyn0mial, line 19: should we ensure `pod != nil` ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	if !featureGates.Enabled(features.FeatureGateKMSEncryptionProvider) {
+		return nil
+	}
+
+	directoryOrCreate := corev1.HostPathDirectoryOrCreate
+	pod.Spec.Volumes = append(pod.Spec.Volumes,
+		corev1.Volume{
+			Name: "kms-socket",
```
p0lyn0mial: nit: `kms-plugin-socket` ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	)
+
+	for i, container := range pod.Spec.Containers {
+		if container.Name != containerName {
+			continue
+		}
+		pod.Spec.Containers[i].VolumeMounts = append(container.VolumeMounts,
+			corev1.VolumeMount{
+				Name:      "kms-socket",
```
p0lyn0mial: nit: `kms-plugin-socket` ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AddKMSPluginVolumeToPod conditionally adds the KMS plugin volume mount to the kube-apiserver container.
+// FIXME: this is a temporary solution to get KMS TP v1 out. We should come up with a different approach afterwards.
```
p0lyn0mial: maybe we should also already `Deprecate` it :) ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -0,0 +1,53 @@
...
+func AddKMSPluginVolumeToPod(pod *corev1.Pod, containerName string, featureGateAccessor featuregates.FeatureGateAccess) error {
```
p0lyn0mial: nit: `AddKMSPluginVolumeAndMountToPod`

pkg/operator/encryption/kms/helpers.go
ardaguclu: That makes sense :)

pkg/operator/encryption/kms/helpers.go
bertinatto: Good point.  This would be a problem if:

1) This function is called multiple times per sync;
2) Another routine adds a mount with the same name;

In both cases, I believe this function should fail rather than preventing the addition of the duplicate, and that's the currently behavior. Do you think it fail silently instead?

Alternatively, I could add a check with a different error, but I'm not sure it's worth it.

No strong opinions, though. I'm happy to change if you disagree.

bertinatto: BTW, in `pkg/operator/v1helpers/helpers.go` we do the same thing

pkg/operator/encryption/kms/helpers.go
bertinatto: Done

pkg/operator/encryption/kms/helpers.go
bertinatto: Right, in that case it's not possible to make a decision

pkg/operator/encryption/kms/helpers.go
bertinatto: See my previous comment

pkg/operator/encryption/kms/helpers.go
bertinatto: Added a check

pkg/operator/encryption/kms/helpers.go
bertinatto: I guess it won't hurt. Done

pkg/operator/encryption/kms/helpers.go
bertinatto: Done

pkg/operator/encryption/kms/helpers.go
bertinatto: Changed it, but deprecated usually means "don't use this, use X instead", so it implies that an alternative exists

pkg/operator/encryption/kms/helpers.go
bertinatto: Renamed

pkg/operator/encryption/kms/helpers.go
bertinatto: Done

ardaguclu:
This looks good to me. But I'll defer the tag to @p0lyn0mial

pkg/operator/encryption/kms/helpers.go
p0lyn0mial: > Do you think it fail silently instead?

My idea was to simply check whether the volume/mount exists before adding it to the list. 
I understand that the assumption is that this function will receive a pod that does not contain the volume/mounts. 

If we don’t want to add these checks, would it make sense to at least document this assumption in the function’s docs?

pkg/operator/encryption/kms/helpers_test.go
```diff
@@ -0,0 +1,204 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddKMSPluginVolume(t *testing.T) {
+	tests := []struct {
```
p0lyn0mial, line 18: This is a personal preference. 

I really like it when test definitions include inputs and expected outputs. It allows me to review test cases without looking at the logic. 

If you feel the same way, we could slightly update the unit test to include `actualPod` and `expectedPod` (with the expected volumes).

pkg/operator/encryption/kms/helpers.go
bertinatto: Makes sense, added a doc

pkg/operator/encryption/kms/helpers_test.go
bertinatto: Ack; included `actualPod` and `expectedPod`

PR #2093, opened 2026-01-27
OCPBUGS-38662: guard controller: Handle conflict gracefully

tchap:
Currently the controller degrades on any error, including an apply conflict. This can happen during upgrades and should not be the case.

The cause is probably the fact that we delete and create the guard pod in the very same sync call. But instead of handling this specifically, the apply call now simply ignores conflict errors and sync is automatically invoked again as the changes propagate.

This is to remove degrading like
```
GuardControllerDegraded: Unable to apply pod openshift-kube-scheduler-guard-ip-10-0-106-117.us-east-2.compute.internal changes: Operation cannot be fulfilled on pods \"openshift-kube-scheduler-guard-ip-10-0-106-117.us-east-2.compute.internal\": the object has been modified; please apply your changes to the latest version and try again
```

### DIFF
--- a/pkg/operator/staticpod/controller/guard/guard_controller.go
+++ b/pkg/operator/staticpod/controller/guard/guard_controller.go
@@ -380,7 +380,9 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			_, _, err = resourceapply.ApplyPod(ctx, c.podGetter, syncCtx.Recorder(), pod)
 			if err != nil {
 				klog.Errorf("Unable to apply pod %v changes: %v", pod.Name, err)
-				errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+				if !apierrors.IsConflict(err) {
+					errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+				}
 			}
 		}
 	}


pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -380,7 +380,9 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			_, _, err = resourceapply.ApplyPod(ctx, c.podGetter, syncCtx.Recorder(), pod)
 			if err != nil {
 				klog.Errorf("Unable to apply pod %v changes: %v", pod.Name, err)
-				errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+				if !apierrors.IsConflict(err) {
+					errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+				}
```
tchap, line 385: I am not sure logging such an error is useful, but I left the log statement there.

pkg/operator/staticpod/controller/guard/guard_controller.go
```diff
@@ -380,7 +380,9 @@ func (c *GuardController) sync(ctx context.Context, syncCtx factory.SyncContext)
 			_, _, err = resourceapply.ApplyPod(ctx, c.podGetter, syncCtx.Recorder(), pod)
 			if err != nil {
 				klog.Errorf("Unable to apply pod %v changes: %v", pod.Name, err)
-				errs = append(errs, fmt.Errorf("Unable to apply pod %v changes: %v", pod.Name, err))
+				if !apierrors.IsConflict(err) {
```
p0lyn0mial, line 383: How is this PR handling conflicts gracefully? 🙂

In my opinion, we should not hide errors.

A conflict error is still an error.
For example, it might indicate a conflict with a different actor in the system.

What we could do instead (if we aren’t already) is consider retrying certain types of requests when a conflict error occurs.

tchap: I think that ignoring is very graceful 🙂 

Or we can simply return when we delete the pod and wait for another round. I think that that's the cause. Because the deployment is using `Recreate` strategy, there should not be multiple pods at once.

tchap: I will investigate further how to handle this in a better way and what the root cause actually is...

tchap: It's actually racing with Kubelet that is updating the status after pod creation, from I can tell. This seems to be normal.

What do you mean retrying requests? This is actually retrying the request, but once an up-to-date object is fetched.

PR #2095, opened 2026-01-27
OCPBUGS-38662: resourcesync: Handle conflict gracefully

tchap:
Do not degrade on conflict when applying changes.
Let the machinery retry automatically when the state is sync'd.

### DIFF
--- a/pkg/operator/resourcesynccontroller/resourcesync_controller.go
+++ b/pkg/operator/resourcesynccontroller/resourcesync_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	"net/http"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/operator/resourcesynccontroller/resourcesync_controller.go
```diff
@@ -226,7 +226,7 @@ func (c *ResourceSyncController) Sync(ctx context.Context, syncCtx factory.SyncC
 		}
 
 		_, _, err := resourceapply.SyncPartialConfigMap(ctx, c.configMapGetter, syncCtx.Recorder(), source.Namespace, source.Name, destination.Namespace, destination.Name, source.syncedKeys, []metav1.OwnerReference{})
-		if err != nil {
+		if err != nil && !apierrors.IsConflict(err) {
```
p0lyn0mial, line 229: How is this PR handling conflicts gracefully? 🙂 

In my opinion, we should not hide errors. 

A conflict error is still an error. 
For example, it might indicate a conflict with a different actor in the system. 

What we could do instead (if we aren’t already) is consider retrying certain types of requests when a conflict error occurs.

PR #2103, opened 2026-02-02
CNTRLPLANE-2247: kms/k8s_mock_kms_plugin_deployer: for simplicity just ensure at least one pod is scheduled before checking readiness

p0lyn0mial:


test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
```diff
@@ -129,6 +129,10 @@ func waitForDaemonSetReady(ctx context.Context, t testing.TB, kubeClient kuberne
 		t.Logf("DaemonSet %s/%s status: desired=%d, ready=%d, available=%d",
 			namespace, daemonSetName, ds.Status.DesiredNumberScheduled, ds.Status.NumberReady, ds.Status.NumberAvailable)
 
+		// for simplicity just ensure at least one pod is scheduled before checking readiness
+		if ds.Status.DesiredNumberScheduled == 0 {
```
p0lyn0mial, line 133: I think that this field is not precomputed at creation time. It is written only after the DaemonSet controller runs.

p0lyn0mial: Thus `ds.Status.NumberReady == ds.Status.DesiredNumberScheduled` might be satisfied and return a false status (the DS ready)

PR #2112, opened 2026-02-03
NO-JIRA: Do not delete kms plugin namespace after test exits

ardaguclu:
This PR removes the kms plugin namespace deletion. Because it appears that cluster requires the accessibility of kms plugin to complete. 

This PR also removes the daemonset logs collection. Because it appears that our CI can smartly collect kms namespace logs (e.g. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/logs/multi-pr-openshift-cluster-kube-apiserver-operator-2018-openshift-cluster-openshift-apiserver-operator-643-openshift-cluster-authentication-operator-832-e2e-gcp-operator-encryption-kms/2018674377842757632/artifacts/e2e-gcp-operator-encryption-kms/gather-extra/artifacts/pods/)

### DIFF
--- a/test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
+++ b/test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
@@ -4,15 +4,11 @@ import (
 	"bytes"
 	"context"
 	"embed"
-	"fmt"
-	"io"
-	"os"
 	"path/filepath"
 	"testing"
 	"text/template"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


test/library/encryption/kms/k8s_mock_kms_plugin_deployer.go
```diff
@@ -73,63 +69,7 @@ func DeployUpstreamMockKMSPlugin(ctx context.Context, t testing.TB, kubeClient k
 	}
 	t.Logf("Upstream mock KMS v2 plugin deployed successfully!")
 
-	return func() {
-		// Before destroying the namespace, collect the logs of the pods in namespace
-		collectPodLogs(ctx, t, kubeClient, namespace)
-
-		if err := destroyNamespaceIfNotExists(ctx, t, kubeClient, namespace); err != nil {
```
p0lyn0mial, line 80: should we also remove destroying the ns at the beginning of the `DeployUpstreamMockKMSPlugin` function ?

multiple tests share the same cluster so if the next test removes the ns it will also affect the kas.

ardaguclu: Good point. `destroyNamespaceIfNotExists` removed entirely.

PR #2113, opened 2026-02-04
CNTRLPLANE-2247: test/library/encryption/kms/k8s_mock_kms_plugin_deployer: cleanup the signature of the DeployUpstreamMockKMSPlugin fn

p0lyn0mial:
we changed the function in https://github.com/openshift/library-go/pull/2112 but we forgot to update the signature.


ardaguclu:
I intentionally didn't change the signature in case we'll need a cleanup function in the future. But we define it when it is needed
/lgtm

PR #2114, opened 2026-02-04
HPCASE-208:Add TLSAdherencePolicy helper

richardsonnick:
Add ShouldHonorClusterTLSProfile helper for TLS adherence policy

Adds a helper function that encapsulates the logic for determining
whether a component should honor the cluster-wide TLS security profile
from apiserver.config.openshift.io/cluster.

This function handles:
- Empty/unset values (treated as LegacyExternalAPIServerComponentsOnly)
- LegacyExternalAPIServerComponentsOnly (returns false)
- StrictAllComponents (returns true)
- Unknown enum values (returns true for forward compatibility)

Component implementors should use this helper rather than checking
tlsAdherence field values directly, allowing coordinated changes to
the default semantic across all implementations.

See: https://github.com/openshift/enhancements/pull/XXXX

Relevant APIServer changes: https://github.com/openshift/api/pull/2680

### DIFF
--- a/pkg/crypto/tls_adherence.go
+++ b/pkg/crypto/tls_adherence.go
@@ -1,0 +1,23 @@
+package crypto
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// ShouldHonorClusterTLSProfile returns true if the component should honor the
+// cluster-wide TLS security profile settings from apiserver.config.openshift.io/cluster.
+//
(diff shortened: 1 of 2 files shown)


richardsonnick:
This will be a draft diff until https://github.com/openshift/api/pull/2680 is landed. Since 2680 adds the TLSAdherencePolicy type

PR #2116, opened 2026-02-04
NO-JIRA: Add ECDSA P-256 support for certificates and CAs

fabiendupont:
Extends the crypto package to generate ECDSA P-256 key pairs and certificates in addition to existing RSA support, enabling OpenShift components to use modern elliptic curve cryptography.

### Changes

**Key generation:**
- `KeyAlgorithm` type for algorithm selection (`AlgorithmRSA`, `AlgorithmECDSA`)
- ECDSA P-256 key pair generation with proper SubjectKeyIdentifier hashing (SHA-1 per RFC 5280)
- `newKeyPairWithAlgorithm()` for unified key generation dispatch

**Server certificates:**
- `CA.MakeServerCertWithAlgorithm()` and `CA.MakeServerCertForDurationWithAlgorithm()`

**CA certificates:**
- `MakeSelfSignedCAConfigForDurationWithAlgorithm()` for ECDSA root CAs
- `MakeCAConfigForDurationWithAlgorithm()` for ECDSA intermediate CAs

**Template cleanup:**
- Removed hardcoded `SignatureAlgorithm: x509.SHA256WithRSA` from certificate templates
- `x509.CreateCertificate` infers the correct algorithm from the signing key (SHA256WithRSA for RSA, ECDSAWithSHA256 for ECDSA P-256)
- This is backward compatible and enables correct cross-algorithm signing (e.g. RSA CA signing ECDSA leaf certs)

All existing APIs remain unchanged — new functionality is opt-in through `*WithAlgorithm` variants.

### Test coverage
- Key generation, signature algorithm detection, and ECDSA PEM encoding
- Server certs: RSA CA + ECDSA leaf, ECDSA CA + RSA leaf, ECDSA CA + ECDSA leaf
- `MakeServerCertForDurationWithAlgorithm` (production code path from `certrotation/target.go`)
- Self-signed ECDSA CAs and mixed-algorithm intermediate CA chains
- Backwards compatibility: all existing RSA tests pass unchanged

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -4,12 +4,14 @@ import (
 	"bytes"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/sha1"
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -463,6 +465,16 @@ const (
 	keyBits = 2048
 )
 
+// KeyAlgorithm represents the type of key pair to generate
+type KeyAlgorithm int
+
+const (
+	// AlgorithmRSA generates 2048-bit RSA key pairs (default for backwards compatibility)
+	AlgorithmRSA KeyAlgorithm = iota
+	// AlgorithmECDSA generates P-256 ECDSA key pairs
+	AlgorithmECDSA
+)
+
 type CA struct {
 	Config *TLSCertificateConfig
 
@@ -666,29 +678,34 @@ func MakeSelfSignedCAConfigForSubject(subject pkix.Name, lifetime time.Duration)
 	if lifetime > DefaultCACertificateLifetimeDuration {
 		warnAboutCertificateLifeTime(subject.CommonName, DefaultCACertificateLifetimeDuration)
 	}
-	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, lifetime)
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, lifetime, AlgorithmRSA)
 }
 
 func MakeSelfSignedCAConfigForDuration(name string, caLifetime time.Duration) (*TLSCertificateConfig, error) {
 	subject := pkix.Name{CommonName: name}
-	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime)
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime, AlgorithmRSA)
+}
+
+func MakeSelfSignedCAConfigForDurationWithAlgorithm(name string, caLifetime time.Duration, algorithm KeyAlgorithm) (*TLSCertificateConfig, error) {
+	subject := pkix.Name{CommonName: name}
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime, algorithm)
 }
 
 func UnsafeMakeSelfSignedCAConfigForDurationAtTime(name string, currentTime func() time.Time, caLifetime time.Duration) (*TLSCertificateConfig, error) {
 	subject := pkix.Name{CommonName: name}
-	return makeSelfSignedCAConfigForSubjectAndDuration(subject, currentTime, caLifetime)
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, currentTime, caLifetime, AlgorithmRSA)
 }
 
-func makeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, currentTime func() time.Time, caLifetime time.Duration) (*TLSCertificateConfig, error) {
+func makeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, currentTime func() time.Time, caLifetime time.Duration, algorithm KeyAlgorithm) (*TLSCertificateConfig, error) {
 	// Create CA cert
-	rootcaPublicKey, rootcaPrivateKey, publicKeyHash, err := newKeyPairWithHash()
+	rootcaPublicKey, rootcaPrivateKey, publicKeyHash, err := newKeyPairWithAlgorithm(algorithm)
 	if err != nil {
 		return nil, err
 	}
 	// AuthorityKeyId and SubjectKeyId should match for a self-signed CA
 	authorityKeyId := publicKeyHash
 	subjectKeyId := publicKeyHash
-	rootcaTemplate := newSigningCertificateTemplateForDuration(subject, caLifetime, currentTime, authorityKeyId, subjectKeyId)
+	rootcaTemplate := newSigningCertificateTemplateForDuration(subject, caLifetime, currentTime, authorityKeyId, subjectKeyId, algorithm)
 	rootcaCert, err := signCertificate(rootcaTemplate, rootcaPublicKey, rootcaTemplate, rootcaPrivateKey)
 	if err != nil {
 		return nil, err
@@ -701,14 +718,22 @@ func makeSelfSignedCAConfigForSubjectAndDuration(subject pkix.Name, currentTime 
 }
 
 func MakeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA) (*TLSCertificateConfig, error) {
+	return makeCAConfigForDuration(name, caLifetime, issuer, AlgorithmRSA)
+}
+
+func MakeCAConfigForDurationWithAlgorithm(name string, caLifetime time.Duration, issuer *CA, algorithm KeyAlgorithm) (*TLSCertificateConfig, error) {
+	return makeCAConfigForDuration(name, caLifetime, issuer, algorithm)
+}
+
+func makeCAConfigForDuration(name string, caLifetime time.Duration, issuer *CA, algorithm KeyAlgorithm) (*TLSCertificateConfig, error) {
 	// Create CA cert
-	signerPublicKey, signerPrivateKey, publicKeyHash, err := newKeyPairWithHash()
+	signerPublicKey, signerPrivateKey, publicKeyHash, err := newKeyPairWithAlgorithm(algorithm)
 	if err != nil {
 		return nil, err
 	}
 	authorityKeyId := issuer.Config.Certs[0].SubjectKeyId
 	subjectKeyId := publicKeyHash
-	signerTemplate := newSigningCertificateTemplateForDuration(pkix.Name{CommonName: name}, caLifetime, time.Now, authorityKeyId, subjectKeyId)
+	signerTemplate := newSigningCertificateTemplateForDuration(pkix.Name{CommonName: name}, caLifetime, time.Now, authorityKeyId, subjectKeyId, algorithm)
 	signerCert, err := issuer.SignCertificate(signerTemplate, signerPublicKey)
 	if err != nil {
 		return nil, err
@@ -816,40 +841,44 @@ func (ca *CA) MakeAndWriteServerCert(certFile, keyFile string, hostnames sets.Se
 type CertificateExtensionFunc func(*x509.Certificate) error
 
 func (ca *CA) MakeServerCert(hostnames sets.Set[string], lifetime time.Duration, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
-	serverPublicKey, serverPrivateKey, publicKeyHash, _ := newKeyPairWithHash()
-	authorityKeyId := ca.Config.Certs[0].SubjectKeyId
-	subjectKeyId := publicKeyHash
-	serverTemplate := newServerCertificateTemplate(pkix.Name{CommonName: sets.List(hostnames)[0]}, sets.List(hostnames), lifetime, time.Now, authorityKeyId, subjectKeyId)
-	for _, fn := range fns {
-		if err := fn(serverTemplate); err != nil {
-			return nil, err
-		}
-	}
-	serverCrt, err := ca.SignCertificate(serverTemplate, serverPublicKey)
-	if err != nil {
-		return nil, err
-	}
-	server := &TLSCertificateConfig{
-		Certs: append([]*x509.Certificate{serverCrt}, ca.Config.Certs...),
-		Key:   serverPrivateKey,
-	}
-	return server, nil
+	return ca.makeServerCert(hostnames, lifetime, AlgorithmRSA, fns...)
 }
 
 func (ca *CA) MakeServerCertForDuration(hostnames sets.Set[string], lifetime time.Duration, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
-	serverPublicKey, serverPrivateKey, publicKeyHash, _ := newKeyPairWithHash()
+	return ca.makeServerCert(hostnames, lifetime, AlgorithmRSA, fns...)
+}
+
+// MakeServerCertWithAlgorithm creates a server certificate with the specified key algorithm
+func (ca *CA) MakeServerCertWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
+	return ca.makeServerCert(hostnames, lifetime, algorithm, fns...)
+}
+
+// MakeServerCertForDurationWithAlgorithm creates a server certificate with specified duration and algorithm
+func (ca *CA) MakeServerCertForDurationWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
+	return ca.makeServerCert(hostnames, lifetime, algorithm, fns...)
+}
+
+func (ca *CA) makeServerCert(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
+	serverPublicKey, serverPrivateKey, publicKeyHash, err := newKeyPairWithAlgorithm(algorithm)
+	if err != nil {
+		return nil, err
+	}
+
 	authorityKeyId := ca.Config.Certs[0].SubjectKeyId
 	subjectKeyId := publicKeyHash
-	serverTemplate := newServerCertificateTemplateForDuration(pkix.Name{CommonName: sets.List(hostnames)[0]}, sets.List(hostnames), lifetime, time.Now, authorityKeyId, subjectKeyId)
+	serverTemplate := newServerCertificateTemplate(pkix.Name{CommonName: sets.List(hostnames)[0]}, sets.List(hostnames), lifetime, time.Now, authorityKeyId, subjectKeyId, algorithm)
+
 	for _, fn := range fns {
 		if err := fn(serverTemplate); err != nil {
 			return nil, err
 		}
 	}
+
 	serverCrt, err := ca.SignCertificate(serverTemplate, serverPublicKey)
 	if err != nil {
 		return nil, err
 	}
+
 	server := &TLSCertificateConfig{
 		Certs: append([]*x509.Certificate{serverCrt}, ca.Config.Certs...),
 		Key:   serverPrivateKey,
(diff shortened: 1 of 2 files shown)


pkg/crypto/crypto.go
```diff
@@ -837,6 +849,70 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.Set[string], lifetime tim
 	return server, nil
 }
 
+// MakeServerCertWithAlgorithm creates a server certificate with the specified key algorithm
+func (ca *CA) MakeServerCertWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
```
p0lyn0mial, line 852: Hmm, I might be wrong, but it looks like `MakeServerCert` and `MakeServerCertWithAlgorithm` are very similar. There are two differences. The first is the key algorithm, and the second is that `MakeServerCertWithAlgorithm` sets the signature algorithm. 


I’m wondering if it would make sense to create a private common function for both. I think something like the following could work: 

```
func (ca *CA) MakeServerCertWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
	sigFn := func(template *x509.Certificate) error {
		template.SignatureAlgorithm = signatureAlgorithmForKey(ca.Config.Key)
		return nil
	}
	fns = append([]CertificateExtensionFunc{sigFn}, fns...)
	return ca.makeServerCertWithAlgorithm(hostnames, lifetime, algorithm, fns...)
}
```

and

```
func (ca *CA) MakeServerCert(hostnames sets.Set[string], lifetime time.Duration, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
	return ca.makeServerCertWithAlgorithm(hostnames, lifetime, AlgorithmRSA, fns...)
}
```

pkg/crypto/crypto.go
```diff
@@ -1001,6 +1077,55 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
...
+
+// newKeyPairWithAlgorithm generates a new key pair using the specified algorithm
+func newKeyPairWithAlgorithm(algo KeyAlgorithm) (crypto.PublicKey, crypto.PrivateKey, []byte, error) {
+	switch algo {
+	case AlgorithmECDSA:
+		return newECDSAKeyPairWithHash()
+	case AlgorithmRSA:
+		return newKeyPairWithHash()
+	default:
+		return newKeyPairWithHash() // Default to RSA for backwards compatibility
```
p0lyn0mial: Since this is a private method I think we could return an error for unsupported algorithms.

pkg/crypto/crypto.go
```diff
@@ -837,6 +849,70 @@ func (ca *CA) MakeServerCertForDuration(hostnames sets.Set[string], lifetime tim
...
+
+	server := &TLSCertificateConfig{
+		Certs: append([]*x509.Certificate{serverCrt}, ca.Config.Certs...),
+		Key:   serverPrivateKey,
+	}
+	return server, nil
+}
+
+// MakeServerCertForDurationWithAlgorithm creates a server certificate with specified duration and algorithm
+func (ca *CA) MakeServerCertForDurationWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
```
p0lyn0mial: am I right that the differences between `MakeServerCertForDuration` and `MakeServerCertForDurationWithAlgorithm` mirror those between `MakeServerCert` and `MakeServerCertWithAlgorithm` (key algorithm and signature algorithm)? If so, could we refactor them in the same way?

fabiendupont: Good feedback. Makes sense.

pkg/crypto/crypto.go
```diff
@@ -1021,13 +1077,62 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
...
+		return newECDSAKeyPairWithHash()
+	case AlgorithmRSA:
+		return newKeyPairWithHash()
+	default:
+		return nil, nil, nil, fmt.Errorf("unsupported key algorithm: %d", algo)
+	}
+}
+
+// signatureAlgorithmForKey returns the appropriate x509.SignatureAlgorithm for the given private key
+func signatureAlgorithmForKey(key crypto.PrivateKey) x509.SignatureAlgorithm {
```
p0lyn0mial: this function is used only in tests. i think that it asserts the signature algorithm on the generated certs.

why not to move this function to the `crypto_test.go` file then ?

pkg/crypto/crypto.go
```diff
@@ -1064,8 +1169,6 @@ func newServerCertificateTemplateForDuration(subject pkix.Name, hosts []string,
 	template := &x509.Certificate{
 		Subject: subject,
 
-		SignatureAlgorithm: x509.SHA256WithRSA,
```
p0lyn0mial, line 1067: OK. I think that the `SignatureAlgorithm` will be set by `x509.CreateCertificate` and our unit tests assert this. Is that correct ?

pkg/crypto/crypto.go
```diff
@@ -1021,13 +1077,62 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
...
+
+// newKeyPairWithAlgorithm generates a new key pair using the specified algorithm
+func newKeyPairWithAlgorithm(algo KeyAlgorithm) (crypto.PublicKey, crypto.PrivateKey, []byte, error) {
+	switch algo {
+	case AlgorithmECDSA:
+		return newECDSAKeyPairWithHash()
+	case AlgorithmRSA:
+		return newKeyPairWithHash()
+	default:
+		return nil, nil, nil, fmt.Errorf("unsupported key algorithm: %d", algo)
```
p0lyn0mial, line 1101: It looks like the error message will take the form of `unsupported key algorithm: 2`, which isn't meaningful to someone debugging. Is there a way to make it more descriptive ?

pkg/crypto/crypto.go
```diff
@@ -1021,13 +1077,62 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
...
+	publicKey, privateKey, err := newECDSAKeyPair()
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	pubBytes, err := x509.MarshalPKIXPublicKey(publicKey)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	hash := sha1.New()
+	hash.Write(pubBytes)
```
p0lyn0mial: I might be wrong but I think that `x509.MarshalPKIXPublicKey` returns `SubjectPublicKeyInfo`:

```
// MarshalPKIXPublicKey converts a public key to PKIX, ASN.1 DER form.
// The encoded public key is a SubjectPublicKeyInfo structure
// (see RFC 5280, Section 4.1).
```

which has the following form:
```
SubjectPublicKeyInfo  ::=  SEQUENCE  {
       algorithm            AlgorithmIdentifier,
       subjectPublicKey     BIT STRING
  }
```

The RFC 5280 Section 4.2.1.2. says to hash only the `subjectPublicKey BIT STRING` value which means we hash too much at the moment (algorithm + subjectPublicKey). 
Should we fix it ?

pkg/crypto/crypto.go
fabiendupont: Done. After removing the hardcoded `SignatureAlgorithm` from the templates (which made the `sigFn` in `MakeServerCertWithAlgorithm` / `MakeServerCertForDurationWithAlgorithm` unnecessary), this function has no production callers. Moved it to `crypto_test.go` where it's used to assert the correct algorithm on generated certs.

pkg/crypto/crypto.go
fabiendupont: Yes, that's correct. When `SignatureAlgorithm` is zero-valued in the template, `x509.CreateCertificate` infers it from the signing key — `SHA256WithRSA` for RSA keys, `ECDSAWithSHA256` for ECDSA P-256 keys. Our tests assert the resulting `SignatureAlgorithm` on the generated certs for all combinations (RSA CA + RSA leaf, RSA CA + ECDSA leaf, ECDSA CA + RSA leaf, ECDSA CA + ECDSA leaf).

pkg/crypto/crypto.go
fabiendupont: Good point. This `default` branch can only be reached if a new `KeyAlgorithm` constant is added to the const block (e.g. `AlgorithmEd25519`) without adding a corresponding case in this switch.

Since `newKeyPairWithAlgorithm` is private and the callers always pass one of the defined constants, a developer hitting this error would know they forgot to wire up their new algorithm.

I added a comment to clarify that, and kept `%d` so the numeric value is visible for debugging.

pkg/crypto/crypto.go
fabiendupont: Good catch. Fixed — we now unmarshal the DER-encoded `SubjectPublicKeyInfo` and hash only the `subjectPublicKey` BIT STRING value, per RFC 5280 Section 4.2.1.2 method 1.

pkg/crypto/crypto.go
```diff
@@ -666,22 +678,27 @@ func MakeSelfSignedCAConfigForSubject(subject pkix.Name, lifetime time.Duration)
...
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, lifetime, AlgorithmRSA)
 }
 
 func MakeSelfSignedCAConfigForDuration(name string, caLifetime time.Duration) (*TLSCertificateConfig, error) {
 	subject := pkix.Name{CommonName: name}
-	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime)
+	return makeSelfSignedCAConfigForSubjectAndDuration(subject, time.Now, caLifetime, AlgorithmRSA)
+}
+
+func MakeSelfSignedCAConfigForDurationWithAlgorithm(name string, caLifetime time.Duration, algorithm KeyAlgorithm) (*TLSCertificateConfig, error) {
```
p0lyn0mial, line 689: I might be wrong but
`MakeSelfSignedCAConfigForDurationWithAlgorithm` & `MakeCAConfigForDurationWithAlgorithm` use `newSigningCertificateTemplateForDuration` which sets

 `KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign`, 

I found RFC 5480 where Section 3 says:
```
If the keyUsage extension is present in a Certification Authority
   (CA) certificate that indicates id-ecPublicKey in
   SubjectPublicKeyInfo, then any combination of the following values
   MAY be present:

     digitalSignature;
     nonRepudiation;
     keyAgreement;
     keyCertSign; and
     cRLSign.
```

which tells me that `x509.KeyUsageKeyEncipherment`  (in `newSigningCertificateTemplateForDuration`) should not be used for ECDSA only for RSA.

pkg/crypto/crypto.go
```diff
@@ -816,40 +841,72 @@ func (ca *CA) MakeAndWriteServerCert(certFile, keyFile string, hostnames sets.Se
...
-	serverPublicKey, serverPrivateKey, publicKeyHash, _ := newKeyPairWithHash()
+	return ca.makeServerCert(hostnames, lifetime, AlgorithmRSA, fns...)
+}
+
+func (ca *CA) MakeServerCertForDuration(hostnames sets.Set[string], lifetime time.Duration, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
+	return ca.makeServerCertForDuration(hostnames, lifetime, AlgorithmRSA, fns...)
+}
+
+// MakeServerCertWithAlgorithm creates a server certificate with the specified key algorithm
+func (ca *CA) MakeServerCertWithAlgorithm(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
```
p0lyn0mial, line 852: I think the issue described at https://github.com/openshift/library-go/pull/2116/changes#r2826380341 is also applicable to the server, client cert template.

pkg/crypto/crypto.go
fabiendupont: Added a keyUsageForAlgorithm helper that returns:
  - RSA: KeyEncipherment | DigitalSignature
  - ECDSA: DigitalSignature only

  Per RFC 5480 Section 3, KeyEncipherment is not valid for ECDSA keys. The CA, server, and intermediate cert templates now receive the algorithm and set KeyUsage accordingly. Tests assert the correct KeyUsage for both RSA and ECDSA certs.

  The public client cert template functions (NewClientCertificateTemplate / NewClientCertificateTemplateForDuration) are unchanged since client cert creation currently always uses RSA.

pkg/crypto/crypto.go
```diff
@@ -1021,13 +1078,75 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
...
+		// This can only be reached if a new KeyAlgorithm constant is added
+		// to the const block above without a corresponding case here.
+		return nil, nil, nil, fmt.Errorf("unsupported key algorithm: %d", algo)
+	}
+}
+
+// keyUsageForAlgorithm returns the appropriate KeyUsage for the given algorithm.
+// RSA keys use KeyEncipherment (for RSA key transport in TLS) + DigitalSignature.
+// ECDSA keys use only DigitalSignature per RFC 5480 Section 3.
+func keyUsageForAlgorithm(algorithm KeyAlgorithm) x509.KeyUsage {
```
p0lyn0mial: nit: could we rename this to `baseKeyUsageForAlgorithm` (or something similar) to indicate the returned value is a starting point that callers may need to extend?

pkg/crypto/crypto_test.go
```diff
@@ -20,6 +23,19 @@ import (
...
+// signatureAlgorithmForKey returns the appropriate x509.SignatureAlgorithm for the given private key.
+// Used in tests to assert that x509.CreateCertificate infers the correct signature algorithm.
+func signatureAlgorithmForKey(key crypto.PrivateKey) x509.SignatureAlgorithm {
+	switch key.(type) {
+	case *ecdsa.PrivateKey:
+		return x509.ECDSAWithSHA256
+	case *rsa.PrivateKey:
+		return x509.SHA256WithRSA
+	default:
+		return x509.SHA256WithRSA
```
p0lyn0mial: why not to `t.Fatal` for unrecognized values?

p0lyn0mial (commented):
Added a few more nits, overall lgtm. 

Before merging, should we pull the changes to the `service-ca-operator` to make sure the backward compatibility path works? 

Also, are we planning some e2e tests for the `service-ca-operator` to cover the new path?

pkg/crypto/crypto.go
fabiendupont: Done. Renamed to `baseKeyUsageForAlgorithm` to make it clear the returned value is a starting point that callers extend (e.g. `| KeyUsageCertSign` for CAs).

pkg/crypto/crypto_test.go
fabiendupont: Done. `signatureAlgorithmForKey` now takes `testing.TB` and calls `t.Fatalf` on unrecognized key types instead of silently defaulting to RSA.

pkg/crypto/crypto_test.go
```diff
@@ -20,6 +23,21 @@ import (
...
+func signatureAlgorithmForKey(t testing.TB, key crypto.PrivateKey) x509.SignatureAlgorithm {
+	t.Helper()
+	switch key.(type) {
+	case *ecdsa.PrivateKey:
+		return x509.ECDSAWithSHA256
+	case *rsa.PrivateKey:
+		return x509.SHA256WithRSA
+	default:
+		t.Fatalf("unrecognized private key type: %T", key)
+		return 0
```
p0lyn0mial: nit: this is unreachable code - i think this could be removed.

fabiendupont: Unfortunately Go requires it — `t.Fatalf` does not have a "never returns" type, so the compiler needs a return statement to satisfy the function signature. Added a `// unreachable, but required by the compiler` comment and moved the `t.Fatalf` + `return` outside the switch to make it clearer.

p0lyn0mial: mhm, ok, thanks for checking.

p0lyn0mial:
/lgtm

/hold
putting this on hold until the `service-ca-operator` PR gets reviewed and we are confident the changes introduced in this PR are the ones we need.

@fabiendupont thanks for the pr and all your work :)

pkg/crypto/crypto.go
```diff
@@ -816,40 +841,72 @@ func (ca *CA) MakeAndWriteServerCert(certFile, keyFile string, hostnames sets.Se
...
+func (ca *CA) makeServerCertForDuration(hostnames sets.Set[string], lifetime time.Duration, algorithm KeyAlgorithm, fns ...CertificateExtensionFunc) (*TLSCertificateConfig, error) {
+	serverPublicKey, serverPrivateKey, publicKeyHash, err := newKeyPairWithAlgorithm(algorithm)
+	if err != nil {
+		return nil, err
+	}
+
 	authorityKeyId := ca.Config.Certs[0].SubjectKeyId
 	subjectKeyId := publicKeyHash
-	serverTemplate := newServerCertificateTemplateForDuration(pkix.Name{CommonName: sets.List(hostnames)[0]}, sets.List(hostnames), lifetime, time.Now, authorityKeyId, subjectKeyId)
+	serverTemplate := newServerCertificateTemplateForDuration(pkix.Name{CommonName: sets.List(hostnames)[0]}, sets.List(hostnames), lifetime, time.Now, authorityKeyId, subjectKeyId, algorithm)
```
benluddy: The difference between `MakeServerCert[WithAlgorithm]` and `MakeServerCertForDuration[WithAlgorithm]` appears to come down to whether they call `newServerCertificateTemplate` or `newServerCertificateTemplateForDuration`, and those two have evolved in a strange way.

The first accepted the lifetime as number of days, applied a default, then converted it to a time.Duration before delegating to the ...ForDuration func. Last year, this was changed so that both accept time.Duration. The remaining differences are (1) the first form applies a default lifetime and (2) has a side effect of writing to stderr.

I _think_ this means that `MakeServerCert[WithAlgorithm]` will apply a default when the provided lifetime is <= 0, whereas `MakeServerCertForDuration[WithAlgorithm]` will not. I doubt this difference is obvious to anyone who might be using this package. **I agree with what Fabien has done in this PR** (i.e., MakeServerCertWithAlgorithm is consistent with MakeServerCert, and MakeServerCertForDuration is consistent with MakeServerCertForDurationWithAlgorithm), but this feels like something we should improve later (cc @p0lyn0mial @sanchezl).

pkg/crypto/crypto_test.go
```diff
@@ -151,7 +168,11 @@ func newSigningCertificateTemplate(subject pkix.Name, lifetime time.Duration, cu
 		warnAboutCertificateLifeTime(subject.CommonName, DefaultCACertificateLifetimeDuration)
 	}
 
-	return newSigningCertificateTemplateForDuration(subject, lifetime, currentTime, nil, nil)
+	algo := AlgorithmRSA
+	if len(algorithm) > 0 {
+		algo = algorithm[0]
+	}
+	return newSigningCertificateTemplateForDuration(subject, lifetime, currentTime, nil, nil, algo)
```
benluddy: Instead of making the function variadic with a default (and tossing all but the first KeyAlgorithm), can we make the KeyAlgorithm a normal parameter and update the callers to always provide a value?

pkg/crypto/crypto_test.go
```diff
@@ -573,3 +595,400 @@ func TestTLSProfileCipherSuitesHaveMappings(t *testing.T) {
...
+	require.IsType(t, &ecdsa.PublicKey{}, publicKey, "public key should be ECDSA")
+	require.IsType(t, &ecdsa.PrivateKey{}, privateKey, "private key should be ECDSA")
+
+	// Verify curve is P-256
+	require.Equal(t, elliptic.P256(), publicKey.Curve, "should use P-256 curve")
+	require.Equal(t, elliptic.P256(), privateKey.Curve, "should use P-256 curve")
+
+	// Verify public key matches private key
+	require.True(t, publicKey.X.Cmp(privateKey.PublicKey.X) == 0, "public key X should match")
+	require.True(t, publicKey.Y.Cmp(privateKey.PublicKey.Y) == 0, "public key Y should match")
```
benluddy: Would https://pkg.go.dev/crypto/ecdsa#PublicKey.Equal be suitable for this test?

benluddy (commented):
This looks reasonable. I spoke with @sanchezl about potential overlap with https://github.com/openshift/enhancements/pull/1882 and I don't think it would be harmful to grow a second way to do this in the future or have to do some refactoring.

pkg/crypto/crypto.go
p0lyn0mial: sgtm, the distinction is almost invisible.

pkg/crypto/crypto.go
```diff
@@ -1021,13 +1078,75 @@ func newRSAKeyPair() (*rsa.PublicKey, *rsa.PrivateKey, error) {
 	return &privateKey.PublicKey, privateKey, nil
 }
 
+// newECDSAKeyPair generates a new P-256 ECDSA key pair
+func newECDSAKeyPair() (*ecdsa.PublicKey, *ecdsa.PrivateKey, error) {
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
hasbro17, line 1055: Not a review, just a note for the PKI feature team that will refactor this down the line https://github.com/openshift/enhancements/pull/1882

> Support RSA (with configurable key sizes: 2048, 3072, 4096) and ECDSA (with configurable curves: P-256, P-384, P-521) algorithms in the initial implementation

/cc @rh-roman @sanchezl We'll have to thread the key size here and for `newKeyPairWithHash()` for the RSA case. 
https://github.com/openshift/library-go/blob/5d9eb6295ff6424951225a1006c00cbbb7ef3d27/pkg/crypto/crypto.go#L1005

And through newKeyPairWithAlgorithm() as well.

pkg/crypto/crypto.go
fabiendupont: Done. Consolidated the two paths:

- Removed `makeServerCertForDuration` — `MakeServerCertForDuration` and `MakeServerCertForDurationWithAlgorithm` now delegate to `makeServerCert`.
- Removed the `newServerCertificateTemplate` lifetime-defaulting wrapper — `newServerCertificateTemplateForDuration` is now the sole `newServerCertificateTemplate`.
- Updated `TestValidityPeriodOfServerCertificate` to remove the test cases for `0` and negative lifetime that relied on the removed defaulting behavior.

pkg/crypto/crypto_test.go
fabiendupont: Done. `newSigningCertificateTemplate` now takes `algorithm KeyAlgorithm` as a required parameter. All callers updated to explicitly pass `AlgorithmRSA` (or `tt.caAlgorithm` for the mixed test).

pkg/crypto/crypto_test.go
fabiendupont: Done. Replaced the manual `X.Cmp`/`Y.Cmp` checks with `publicKey.Equal(&privateKey.PublicKey)`.

PR #2118, opened 2026-02-05
Revert "pkg/crypto: Add missing cipher suites to OpenSSL-to-IANA ciphers mapping"

ardaguclu:
Reverts openshift/library-go#2080

Based on this https://github.com/openshift/library-go/pull/2080#issuecomment-3851637780


p0lyn0mial:
cc @damdo

PR #2119, opened 2026-02-05
pkg/crypto: OpenSSL-to-IANA ciphers mapping: add missing, add comment for unsupported ones

damdo:
A modified revert of openshift/library-go#2118, to partially reintroduce https://github.com/openshift/library-go/pull/2080 and add comments on the reasoning regarding missing cipher mappings and explaining how to maintain the mappings list.

We want to have a mapping for all OpenSSL ciphers defined in https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go 
so then we can translate them to IANA ciphers, which is what go's crypto/tls understands.

We want to make sure both the openshift/api and these mappings are also compatible with go's crypto/tls ciphers: https://github.com/golang/go/blob/d4febb45179fa99ee1d5783bcb693ed7ba14115c/src/crypto/tls/cipher_suites.go#L682-L724 

More details here: https://github.com/openshift/library-go/pull/2080#issuecomment-3852356977
And here: https://redhat-internal.slack.com/archives/C098FU5MRAB/p1770309657097269

--

openshift/api PR counterpart: https://github.com/openshift/api/pull/2697


damdo:
@bertinatto here it is: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2034

damdo:
/hold

Until https://github.com/openshift/api/pull/2697 merges

damdo:
@bertinatto @p0lyn0mial @ardaguclu CI has finished on https://github.com/openshift/cluster-kube-apiserver-operator/pull/2034 are we happy with the results?

sanchezl (approved):
/lgtm

PR #2121, opened 2026-02-09
NO-JIRA: Define static encryption providers as array to be used by all

ardaguclu:
SSIA

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -34,6 +34,8 @@ var (
 	// plus 10 additional minutes for actual migration
 	waitPollTimeout       = 69*time.Minute + 10*time.Minute
 	defaultEncryptionMode = string(configv1.EncryptionTypeIdentity)
+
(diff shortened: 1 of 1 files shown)


test/library/encryption/helpers.go
```diff
@@ -34,6 +34,8 @@ var (
 	// plus 10 additional minutes for actual migration
 	waitPollTimeout       = 69*time.Minute + 10*time.Minute
 	defaultEncryptionMode = string(configv1.EncryptionTypeIdentity)
+
+	SupportedStaticEncryptionProviders = []configv1.EncryptionType{configv1.EncryptionTypeAESGCM, configv1.EncryptionTypeAESCBC}
```
p0lyn0mial, line 38: should we also add `secretbox` ? Is it supported ?

ardaguclu: I think secretBox is not supported by our controllers https://github.com/openshift/library-go/blob/7bced6e899b65175945fc309a2e125de49cac3dc/pkg/operator/encryption/controllers/key_controller.go#L290?

p0lyn0mial: ah, ok, thanks for checking.

PR #2122, opened 2026-02-10
OCPBUGS-74511: remove RouteExternalCertificate feature gate

jcmoraisjr:
RouteExternalCertificate featuregate is now enabled by default. This update is removing the opts used to hold its value, hardcoding the behavior when the value is true.

This update needs to be in sync with o/kubernetes and o/openshift-apiserver.

### DIFF
--- a/pkg/manifestclienttest/testdata/must-gather-01/cluster-scoped-resources/config.openshift.io/featuregates.yaml
+++ b/pkg/manifestclienttest/testdata/must-gather-01/cluster-scoped-resources/config.openshift.io/featuregates.yaml
@@ -59,7 +59,6 @@ items:
       - name: MixedCPUsAllocation
       - name: NetworkLiveMigration
       - name: NodeSwap
-      - name: RouteExternalCertificate
       - name: SigstoreImageVerification
       - name: VSphereControlPlaneMachineSet
       - name: VSphereStaticIPs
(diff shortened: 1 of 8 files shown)


jcmoraisjr:
/hold
needs to be merged after https://github.com/openshift/kubernetes/pull/2585 and https://github.com/openshift/openshift-apiserver/pull/604

bertinatto (commented):
/approve

Is the feature enabled by default since 4.19?

bertinatto:
> @bertinatto this change should be immediately followed by changes in both [o/kubernetes](https://github.com/openshift/kubernetes/pull/2585) and [o/openshift-apiserver](https://github.com/openshift/openshift-apiserver/pull/604) due to the removal of the whole struct. Can you support us in a way we bump and merge all of them quickly, so users trying to bump library-go are not impacted? Otherwise we can change the approach and instead make smaller cleanups on o/kubernetes and o/openshift-apiserver before changing library-go.
> 
> There is also a new comment [here](https://github.com/openshift/kubernetes/pull/2585#discussion_r2877797946). Any preferable approach?

It looks like you want to merge the library-go PR first, then bump it in o/k and o/openshift-apiserver.

The o/k and o/openshift-apiserver PRs can have:

1) A commit bumping library-go (o/k has a special script for that, check the README)
2) Any other commits you may need to drop the feature (including dead references to library-go code that you're removing)

I recommend you wait for @jacobsee and @jubittajohn to land the kube 1.35 update first.

I can approve PRs in o/k and o/library-go. You'll need find an approver for o/openshift-apiserver.

jcmoraisjr:
@bertinatto [o/openshift-apiserver](https://github.com/openshift/openshift-apiserver/pull/604) is still on k 1.34, but I missed the hold tag and it ended up approved and merged. I think I cannot update library-go there due to kube version mismatch, so folks should handle it when bumping kube libs.

I'm planning to unhold this one now, immediately followed by bumping its version on [o/kubernetes](https://github.com/openshift/kubernetes/pull/2585) and merging there as well. Sounds good, should we proceed?

PR #2128, opened 2026-02-18
OCPBUGS-65896: controller/workload: Rework Degraded/Progressing conditions

tchap:
Reworks how the workload controller computes Degraded, Progressing, and Available conditions to fix false positives during scaling and improve failure detection.

**Progressing/Degraded no longer consult Deployment generation.** Previously, a generation mismatch (e.g. from a scale-up) would immediately set Progressing=True and could trigger Degraded. Now the controller relies solely on the Deployment's own Progressing condition:
- Progressing=True when the deployment has not yet reported `NewReplicaSetAvailable` and hasn't timed out.
- `ProgressDeadlineExceeded` is detected from the Deployment's Progressing condition and surfaces as both Progressing=False and DeploymentDegraded=True.

**DeploymentDegraded now detects failing pods** (when not mid-rollout). A pod is considered failing when it is not Ready past its ProgressDeadlineSeconds since creation, or (when MinReadySeconds > 0) its Ready condition is flapping — currently Ready but the last transition is too recent to count as available, checked only after ProgressDeadlineSeconds + MinReadySeconds have elapsed. Terminating pods are excluded to avoid false positives from the old ReplicaSet during rollouts. Diagnostic messages use `deployment.PodContainersStatus` to describe container states.

**WorkloadDegraded SyncError is set via defer** to ensure it is applied regardless of the return path.

This affects `authentication-operator` and `openshift-apiserver-operator`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized status error messaging and clearer Progressing state handling with improved timeout detection and explicit timed-out messaging.
  * More accurate Degraded reporting: better detection of failing pods and concise "N of M unavailable" messages; preserves "AsExpected" when appropriate.
  * Stricter version recording gated on observed generation and replica alignment; pod-list errors now surface as SyncError in status.

* **Tests**
  * Expanded and rewritten status scenarios covering timeouts, multi-pod failures, pod-list errors, recovery paths, and version-recording validations.

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -16,6 +17,7 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
@@ -217,10 +219,7 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	}()
 
 	if !preconditionsReady {
-		var message string
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
-		}
+		message := errMessage(errs)
 		if len(message) == 0 {
 			message = "the operator didn't specify what preconditions are missing"
 		}
@@ -248,24 +247,14 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		return kerrors.NewAggregate(errs)
 	}
 
-	if len(errs) > 0 {
-		message := ""
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
+	defer func() {
+		if len(errs) > 0 {
+			workloadDegradedCondition = workloadDegradedCondition.
+				WithStatus(operatorv1.ConditionTrue).
+				WithReason("SyncError").
+				WithMessage(errMessage(errs))
 		}
-		workloadDegradedCondition = workloadDegradedCondition.
-			WithStatus(operatorv1.ConditionTrue).
-			WithReason("SyncError").
-			WithMessage(message)
-	} else if workload == nil {
-		workloadDegradedCondition = workloadDegradedCondition.
-			WithStatus(operatorv1.ConditionTrue).
-			WithReason("NoDeployment").
-			WithMessage(fmt.Sprintf("deployment/%s: could not be retrieved", c.targetNamespace))
-	} else {
-		workloadDegradedCondition = workloadDegradedCondition.
-			WithStatus(operatorv1.ConditionFalse)
-	}
+	}()
 
 	if workload == nil {
 		message := fmt.Sprintf("deployment/%s: could not be retrieved", c.targetNamespace)
@@ -284,42 +273,53 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("NoDeployment").
 			WithMessage(message)
 
+		workloadDegradedCondition = workloadDegradedCondition.
+			WithStatus(operatorv1.ConditionTrue).
+			WithReason("NoDeployment").
+			WithMessage(message)
+
 		return kerrors.NewAggregate(errs)
 	}
+
+	workloadDegradedCondition = workloadDegradedCondition.WithStatus(operatorv1.ConditionFalse)
 
 	if workload.Status.AvailableReplicas == 0 {
 		deploymentAvailableCondition = deploymentAvailableCondition.
 			WithStatus(operatorv1.ConditionFalse).
 			WithReason("NoPod").
-			WithMessage(fmt.Sprintf("no %s.%s pods available on any node.", workload.Name, c.targetNamespace))
+			WithMessage(fmt.Sprintf("no %s.%s pods available on any node", workload.Name, c.targetNamespace))
 	} else {
 		deploymentAvailableCondition = deploymentAvailableCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("AsExpected")
 	}
 
-	desiredReplicas := int32(1)
-	if workload.Spec.Replicas != nil {
-		desiredReplicas = *(workload.Spec.Replicas)
+	desiredReplicas := ptr.Deref(workload.Spec.Replicas, 1)
+
+	// Update is done when the deployment controller has reported NewReplicaSetAvailable.
+	// Checking the current vs. observed generation here is not possible since we don't want to be Progressing on scaling.
+	progressTimedOutMessage, workloadIsBeingUpdatedTooLong := hasDeploymentTimedOutProgressing(workload.Status)
+	workloadIsBeingUpdated := !hasDeploymentProgressed(workload.Status) && !workloadIsBeingUpdatedTooLong
+
+	var progressDeadlineExceededMessage string
+	if workloadIsBeingUpdatedTooLong {
+		progressDeadlineExceededMessage = fmt.Sprintf("deployment/%s.%s has timed out progressing: %s", workload.Name, c.targetNamespace, progressTimedOutMessage)
 	}
 
-	// If the workload is up to date, then we are no longer progressing
-	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
-	// Update is done when all pods have been updated to the latest revision
-	// and the deployment controller has reported NewReplicaSetAvailable
-	workloadIsBeingUpdated := !workloadAtHighestGeneration || !hasDeploymentProgressed(workload.Status)
-	workloadIsBeingUpdatedTooLong := v1helpers.IsUpdatingTooLong(previousStatus, *deploymentProgressingCondition.Type)
-	if !workloadAtHighestGeneration {
-		deploymentProgressingCondition = deploymentProgressingCondition.
-			WithStatus(operatorv1.ConditionTrue).
-			WithReason("NewGeneration").
-			WithMessage(fmt.Sprintf("deployment/%s.%s: observed generation is %d, desired generation is %d.", workload.Name, c.targetNamespace, workload.Status.ObservedGeneration, workload.ObjectMeta.Generation))
-	} else if workloadIsBeingUpdated {
+	switch {
+	case workloadIsBeingUpdated:
 		deploymentProgressingCondition = deploymentProgressingCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("PodsUpdating").
-			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
-	} else {
+			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest revision and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
+
+	case workloadIsBeingUpdatedTooLong:
+		deploymentProgressingCondition = deploymentProgressingCondition.
+			WithStatus(operatorv1.ConditionFalse).
+			WithReason("ProgressDeadlineExceeded").
+			WithMessage(progressDeadlineExceededMessage)
+
+	default:
 		// Terminating pods don't account for any of the other status fields but
 		// still can exist in a state when they are accepting connections and would
 		// contribute to unexpected behavior when we report Progressing=False.
@@ -332,22 +332,42 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("AsExpected")
 	}
 
-	// During a rollout the default maxSurge (25%) will allow the available
-	// replicas to temporarily exceed the desired replica count. If this were
-	// to occur, the operator should not report degraded.
-	workloadHasAllPodsAvailable := workload.Status.AvailableReplicas >= desiredReplicas
-	if !workloadHasAllPodsAvailable && (!workloadIsBeingUpdated || workloadIsBeingUpdatedTooLong) {
-		numNonAvailablePods := desiredReplicas - workload.Status.AvailableReplicas
+	switch {
+	case workloadIsBeingUpdatedTooLong:
 		deploymentDegradedCondition = deploymentDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
-			WithReason("UnavailablePod")
-		podContainersStatus, err := deployment.PodContainersStatus(workload, c.podsLister)
+			WithReason("ProgressDeadlineExceeded").
+			WithMessage(progressDeadlineExceededMessage)
+
+	// The following case handles Degraded when not progressing, particularly on scaling.
+	case !workloadIsBeingUpdated && workload.Status.AvailableReplicas < desiredReplicas:
+		hasFailing, err := hasFailingPods(workload, c.podsLister, time.Now())
 		if err != nil {
-			podContainersStatus = []string{fmt.Sprintf("failed to get pod containers details: %v", err)}
+			errs = append(errs, err)
 		}
-		deploymentDegradedCondition = deploymentDegradedCondition.
-			WithMessage(fmt.Sprintf("%v of %v requested instances are unavailable for %s.%s (%s)", numNonAvailablePods, desiredReplicas, workload.Name, c.targetNamespace, strings.Join(podContainersStatus, ", ")))
-	} else {
+		if hasFailing || workload.Status.AvailableReplicas == 0 {
+			containerMessages, err := deployment.PodContainersStatus(workload, c.podsLister)
+			if err != nil {
+				errs = append(errs, err)
+			}
+			var failureDescription string
+			if len(containerMessages) > 0 {
+				failureDescription = ` (` + strings.Join(containerMessages, ", ") + `)`
+			}
+
+			numUnavailable := desiredReplicas - workload.Status.AvailableReplicas
+			message := fmt.Sprintf("%d of %d requested instances are unavailable for %s.%s%s", numUnavailable, desiredReplicas, workload.Name, c.targetNamespace, failureDescription)
+			deploymentDegradedCondition = deploymentDegradedCondition.
+				WithStatus(operatorv1.ConditionTrue).
+				WithReason("UnavailablePod").
+				WithMessage(message)
+		} else {
+			deploymentDegradedCondition = deploymentDegradedCondition.
+				WithStatus(operatorv1.ConditionFalse).
+				WithReason("AsExpected")
+		}
+
+	default:
 		deploymentDegradedCondition = deploymentDegradedCondition.
 			WithStatus(operatorv1.ConditionFalse).
 			WithReason("AsExpected")
@@ -356,8 +376,11 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	// if the deployment is all available and at the expected generation, then update the version to the latest
 	// when we update, the image pull spec should immediately be different, which should immediately cause a deployment rollout
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
-	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
-	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
+	if operatorConfigAtHighestGeneration &&
+		workload.ObjectMeta.Generation == workload.Status.ObservedGeneration &&
+		workload.Status.AvailableReplicas == desiredReplicas &&
+		workload.Status.UpdatedReplicas == desiredReplicas {
+
 		c.versionRecorder.SetVersion(c.constructOperandNameFor(workload.Name), c.targetOperandVersion)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -303,23 +301,24 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 		desiredReplicas = *(workload.Spec.Replicas)
 	}
 
-	// If the workload is up to date, then we are no longer progressing
-	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
-	// Update is done when all pods have been updated to the latest revision
-	// and the deployment controller has reported NewReplicaSetAvailable
-	workloadIsBeingUpdated := !workloadAtHighestGeneration || !hasDeploymentProgressed(workload.Status)
-	workloadIsBeingUpdatedTooLong := v1helpers.IsUpdatingTooLong(previousStatus, *deploymentProgressingCondition.Type)
```
tchap, line 309: We don't use `IsUpdatingTooLong` any more, we based the check on `ProgressDeadlineExceeded` from the deployment controller.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -303,23 +301,24 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+	// Update is done when the deployment controller has reported NewReplicaSetAvailable.
+	// Checking the current vs. observed generation here is not possible since we don't want to be Progressing on scaling.
+	progressTimedOutMessage, workloadIsBeingUpdatedTooLong := hasDeploymentTimedOutProgressing(workload.Status)
+	workloadIsBeingUpdated := !hasDeploymentProgressed(workload.Status) && !workloadIsBeingUpdatedTooLong
+	switch {
+	case workloadIsBeingUpdated:
 		deploymentProgressingCondition = deploymentProgressingCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("PodsUpdating").
 			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest generation and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
```
atiratree:
```suggestion
			WithMessage(fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest revision and %d/%d pods are available", workload.Name, c.targetNamespace, workload.Status.UpdatedReplicas, desiredReplicas, workload.Status.AvailableReplicas, desiredReplicas))
```

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -332,22 +331,21 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 
-	// During a rollout the default maxSurge (25%) will allow the available
-	// replicas to temporarily exceed the desired replica count. If this were
-	// to occur, the operator should not report degraded.
-	workloadHasAllPodsAvailable := workload.Status.AvailableReplicas >= desiredReplicas
-	if !workloadHasAllPodsAvailable && (!workloadIsBeingUpdated || workloadIsBeingUpdatedTooLong) {
-		numNonAvailablePods := desiredReplicas - workload.Status.AvailableReplicas
+	// Degraded is set when the deployment is not Available or timed out progressing.
+	switch {
+	case *deploymentAvailableCondition.Status == operatorv1.ConditionFalse:
```
atiratree: Let's use `.workload.Status.AvailableReplicas` as the source of truth here., so we don't make a mistake in the future. I have seen these conditions being reshuffled couple of times in the past.

atiratree: FYI: this will cause the workload to be degraded during the initial rollout, which is fine, I guess? 

Please, at least add a comment about this here.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -356,8 +354,11 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	// if the deployment is all available and at the expected generation, then update the version to the latest
 	// when we update, the image pull spec should immediately be different, which should immediately cause a deployment rollout
 	// which should immediately result in a deployment generation diff, which should cause this block to be skipped until it is ready.
-	workloadHasAllPodsUpdated := workload.Status.UpdatedReplicas == desiredReplicas
-	if workloadAtHighestGeneration && workloadHasAllPodsAvailable && workloadHasAllPodsUpdated && operatorConfigAtHighestGeneration {
+	if operatorConfigAtHighestGeneration &&
+		workload.ObjectMeta.Generation == workload.Status.ObservedGeneration &&
+		workload.Status.AvailableReplicas >= desiredReplicas &&
```
atiratree:
```suggestion
		workload.Status.AvailableReplicas == desiredReplicas &&
```

do we need the `>=`?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: let's also set `.spec.progressDeadlineSeconds` to 15m on the consumer workloads to honor the superseded `progressingConditionTimeout` value

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -577,7 +578,7 @@ func TestUpdateOperatorStatus(t *testing.T) {
 			},
 		},
 		{
-			name: "progressing==false for a longer time shouldn't make the otherwise fine workload degraded",
+			name: "stable workload with long-false progressing condition",
```
atiratree: false progressing condition of the input or the output?

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -676,6 +677,173 @@ func TestUpdateOperatorStatus(t *testing.T) {
...
+					{
+						Type:    fmt.Sprintf("%sDeployment%s", defaultControllerName, operatorv1.OperatorStatusTypeProgressing),
+						Status:  operatorv1.ConditionFalse,
+						Reason:  "AsExpected",
+						Message: "",
+					},
+				}
+				return areCondidtionsEqual(expectedConditions, actualStatus.Conditions)
+			},
+		},
```
atiratree, line 1354: Can you please also add a test for unavailable workload that has successfully progressed (completed) in the past?

pkg/operator/apiserver/controller/workload/workload.go
tchap: I don't think we can do that. `Sync` from the delegate returns the synchronized deployment already. So we would have to send another update request, I guess.

pkg/operator/apiserver/controller/workload/workload.go
tchap: Not sure why it was there, maybe for surge. But we can use `==` no problem.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: This is actually more of a redundant test left from the previous version, I will remove it.

pkg/operator/apiserver/controller/workload/workload.go
tchap: We can align this in auth when updating library-go. I will actually open a PR there once this is accepted to run CI anyway.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: let's duplicate the message creation here as well for the same reason of not depending on the available condition

pkg/operator/apiserver/controller/workload/workload.go
atiratree: we can wait until the old revision pods are gone after the surge

pkg/operator/apiserver/controller/workload/workload.go
atiratree: Yup, I meant that we should update it in the consumer repos like auth.

atiratree: Might be better to open a proof PR before the library-go changes are merged.

atiratree (commented):
Just one nit, otherwise LGTM, thanks!

pkg/operator/apiserver/controller/workload/workload.go
tchap: Opened https://github.com/openshift/cluster-authentication-operator/pull/843

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -4,7 +4,6 @@ import (
 	"context"
```
p0lyn0mial, line 4: Am I right that those are the three (at least) main behavioral changes introduced in this PR ? 

```
1. Scaling the workload (up/down) won't report Progressing=True 

2. A stuck rollout eventually stops reporting Progressing=True 

3. A workload missing some replicas (2/3) no longer triggers Degraded=True
```

tchap: Yeah, right. We only degrade on Available == 0.

Also checking the Progressing timeout now depends on the native Deployment `ProgressDeadlineExceeded` instead of using a helper and once Progressing deadline is hit, we go Degraded.

p0lyn0mial: I would like to get @benluddy opinion on that.

tchap: Ok, I added another commit that actually makes Degraded work as expected, i.e. Degraded on not all replicas available. This is done by actually checking relevant pods and counting the failing ones so that starting new pods is not included into the Degrading condition.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -386,6 +403,66 @@ func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
...
+			case c.RestartCount > 1:
+				msg = fmt.Sprintf("container %s is crashlooping in pod %s", c.Name, pod.Name)
+			case c.State.Waiting != nil && isFailureWaitingReason(c.State.Waiting.Reason):
+				msg = fmt.Sprintf("container %s is waiting (%s) in pod %s", c.Name, c.State.Waiting.Reason, pod.Name)
+			}
+			if len(msg) > 0 {
+				failureMessages = append(failureMessages, msg)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
tchap: This would lead to false positives as this is an expected temporary state, rejected.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -386,6 +403,66 @@ func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
...
+			return cond.Message, cond.Status == corev1.ConditionFalse && cond.Reason == "ProgressDeadlineExceeded"
+		}
+	}
+	return "", false
+}
+
+func hasFailingPods(workload *appsv1.Deployment, podsLister corev1listers.PodLister) ([]string, error) {
+	pods, err := podsLister.Pods(workload.Namespace).List(labels.SelectorFromSet(workload.Spec.Template.Labels))
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
tchap: Fixed.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -386,6 +403,70 @@ func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
...
+	}
+	pods, err := podsLister.Pods(workload.Namespace).List(selector)
+	if err != nil {
+		return nil, err
+	}
+
+	var failureMessages []string
+	for _, pod := range pods {
+		for _, c := range slices.Concat(pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses) {
```
atiratree: Hm, can we simplify this? I think it would be preferable to not take responsibility for the container lifecycle. It has its intricacies and is still evolving. 

Could we simulate what the deployment controller is doing during rollout also when we initiate scaling?

The main signal that the deployment controller is using is the `Ready` condition

We could use it to do the following:
- if the pod is not ready - wait for `ProgressDeadlineSeconds` until it becomes ready (check start is creationTimestamp).
- if the pod is ready, but it became available too late or the ready condition is flapping (ready transition time is after ProgressDeadlineSeconds), the operator should become degraded. Once it becomes available, the deployment controller will set AvailableReplicas.
- We can skip the checks above if `workloadIsBeingUpdated` to leave the responsibility on the deployment controller.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -331,20 +331,35 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 			WithReason("ProgressDeadlineExceeded").
-			WithMessage(fmt.Sprintf("deployment/%s.%s has timed out progressing: %s", workload.Name, c.targetNamespace, progressTimedOutMessage))
+			WithMessage(progressDeadlineExceededMessage)
+
+	case workload.Status.AvailableReplicas < desiredReplicas:
+		failureMessages, err := hasFailingPods(workload, c.podsLister)
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if len(failureMessages) > 0 || (workload.Status.AvailableReplicas == 0 && !workloadIsBeingUpdated) {
```
atiratree: `workload.Status.AvailableReplicas == 0 `

During initial rollout this is fine, but later not so much. I wonder if we can do something about it?

tchap: What solution are you thinking about? I can only think about removing this and simply always degrade on zero replicas available, which kinda makes sense, only introduces some initial noise...

pkg/operator/apiserver/controller/workload/workload.go
tchap: Attempted to do this 0e2cd124e41b5f184e4b1a5d82d18fa5d16639c4 , feel free to review 🙏🏻

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -332,22 +332,43 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
+	switch {
+	case workloadIsBeingUpdatedTooLong:
 		deploymentDegradedCondition = deploymentDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
-			WithReason("UnavailablePod")
-		podContainersStatus, err := deployment.PodContainersStatus(workload, c.podsLister)
+			WithReason("ProgressDeadlineExceeded").
+			WithMessage(progressDeadlineExceededMessage)
+
+	case !workloadIsBeingUpdated && workload.Status.AvailableReplicas < desiredReplicas:
```
atiratree, line 343: can you please drop a comment here, that this is for checking progress during scaling operations?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -332,22 +332,43 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 		if err != nil {
-			podContainersStatus = []string{fmt.Sprintf("failed to get pod containers details: %v", err)}
+			errs = append(errs, err)
 		}
-		deploymentDegradedCondition = deploymentDegradedCondition.
-			WithMessage(fmt.Sprintf("%v of %v requested instances are unavailable for %s.%s (%s)", numNonAvailablePods, desiredReplicas, workload.Name, c.targetNamespace, strings.Join(podContainersStatus, ", ")))
-	} else {
+		if hasFailing || workload.Status.AvailableReplicas == 0 {
+			var failureDescription string
+			if hasFailing {
```
atiratree: why is this `if` important?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -386,6 +410,64 @@ func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
...
+
+	for _, pod := range pods {
+		if pod.DeletionTimestamp != nil {
+			continue
+		}
+
+		readyCond := findPodReadyCondition(pod)
+		deadline := pod.CreationTimestamp.Time.Add(progressDeadline)
+
+		if (readyCond == nil || readyCond.Status != corev1.ConditionTrue) && now.After(deadline) {
```
atiratree, line 444: this looks pretty good, but it will not cover the flapping ready condition. I mean it will detect the transition, but there is `minReadySeconds` period of time, when the pod is ready but not available and we should probably stay degraded for this period?

I guess the suggested check is only relevant after `ProgressDeadlineSeconds + minReadySeconds`, because during normal rollout/scaleup it is expected.

pkg/operator/apiserver/controller/workload/workload.go
tchap: We only assemble the extended message on `hasFailing` because that means there are actually some pods. When there are no pods, `PodContainersStatus` returns something like `no pods found with label XYZ`. So we just want to skip that. I will add a comment there.

pkg/operator/apiserver/controller/workload/workload.go
tchap: Added the flapping check in https://github.com/openshift/library-go/pull/2128/changes/e1a9fe8aeeb03db294e1db82af9b2b2bd7d4b0b1

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -628,54 +746,675 @@ func TestUpdateOperatorStatus(t *testing.T) {
...
+		{
+			// MinReadySeconds=60, ProgressDeadlineSeconds defaults to 600.
+			// Combined deadline: 600+60 = 660s (11m).
+			// Pod created 15m ago → past combined deadline → check is relevant.
+			// LastTransitionTime 10s ago < MinReadySeconds (60s) → flapping → degraded.
+			name: "scenario: pod with flapping Ready condition detected as failing",
```
atiratree: nit:
```suggestion
			name: "scenario: pod with flapping Ready condition after successful rollout detected as failing",
```
would give more info on what we are doing here, WDYT? Also, we might consider adding this to the other 2 tests?

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -628,54 +746,675 @@ func TestUpdateOperatorStatus(t *testing.T) {
...
+					Namespace:  "openshift-apiserver",
+					Generation: 5,
+				},
+				Spec: appsv1.DeploymentSpec{
+					Replicas:        ptr.To[int32](3),
+					MinReadySeconds: 60,
+				},
+				Status: appsv1.DeploymentStatus{
+					AvailableReplicas:  2,
+					UpdatedReplicas:    3,
```
atiratree, line 1007: This can be confusing without knowing the state of the other pods in this test case. With degraded `AsExpected`, I expect that the newest pod was recently scaled up and did not manage to become available yet.

pkg/operator/apiserver/controller/workload/workload.go
atiratree: The pods might still exist with `AvailableReplicas == 0`. Even if they don't, shouldn't `PodContainersStatus` fn handle that?

pkg/operator/apiserver/controller/workload/workload.go
atiratree: Thx!

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: The tests are usually written so that only the breaking pods are mentioned in the test cases while the ok ones are represented by `AvailableReplicas`, but I can create some test cases also where the healthy pods are listed so that we see they are not being included.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: Amending...

pkg/operator/apiserver/controller/workload/workload_test.go
atiratree: I think the current state is fine, but one of the pods should be unavailable, right? For example created/ready for 30s.

tchap: Isn't this checked in `scenario: pod with flapping Ready condition detected as failing` ?

tchap: Added a pod that was created 10 seconds ago, so the numbers match and the test passes.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -619,101 +688,789 @@ func TestUpdateOperatorStatus(t *testing.T) {
...
+					ObservedGeneration: 5,
+					Conditions: []appsv1.DeploymentCondition{
+						{Type: appsv1.DeploymentProgressing, Status: corev1.ConditionTrue, Reason: "NewReplicaSetAvailable"},
+					},
+				},
+			},
+			pods: []*corev1.Pod{
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "apiserver-1", Namespace: "openshift-apiserver", CreationTimestamp: metav1.NewTime(time.Now().Add(-15 * time.Minute))},
```
atiratree:
```suggestion
					ObjectMeta: metav1.ObjectMeta{Name: "apiserver-1", Namespace: "openshift-apiserver", CreationTimestamp: metav1.NewTime(time.Now().Add(-20 * time.Minute))},
```

this is probably more clear - to ensure that the flapping condiition can reconcile back to a healthy one - well passed the deadline

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -619,101 +688,789 @@ func TestUpdateOperatorStatus(t *testing.T) {
...
+					Status: corev1.PodStatus{
+						Phase: corev1.PodRunning,
+						Conditions: []corev1.PodCondition{
+							{Type: corev1.PodReady, Status: corev1.ConditionTrue, LastTransitionTime: metav1.NewTime(time.Now().Add(-5 * time.Minute))},
+						},
+					},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "apiserver-2", Namespace: "openshift-apiserver", CreationTimestamp: metav1.NewTime(time.Now().Add(-10 * time.Second))},
```
atiratree, line 1028: please add a short comment here for posterity why the pod is present/needed

tchap: Amended.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: Incorporated.

atiratree:
I think, this looks like a reasonable to way to monitor progressing and degraded even during scalling operations. The deployment controller does not support this, but we do use the Deployment primitives to achieve this.
/lgtm

atiratree:
Btw @tchap can you also try to run this again? https://github.com/openshift/cluster-authentication-operator/pull/843

p0lyn0mial:
Could we split this PR into smaller PRs? It looks like the following changes were included:
 
1. Some cleanup, which could be moved to a separate refactoring PR
2. A rework of the progressing condition
3. A rewording of the degraded condition
4. Changes related to the version

This breakdown would help me better understand the proposed changes. Thanks.

PR #2132, opened 2026-02-24
OCPBUGS-77056: Release lock before WaitForCacheSync in secret monitor

bentito:
## Summary

Routes using `spec.tls.externalCertificate` trigger a call to
`AddSecretEventHandler` for each route at router startup. The previous
implementation held the global write mutex (`s.lock`) for the **entire** duration
of the call, including the `WaitForCacheSync` round-trip to the API server and
etcd (~100–600ms per secret). This caused all N secret registrations to run
strictly serially:

| Scale | etcd latency | Startup time |
|---|---|---|
| 112 secrets | ~587ms/secret | 65s |
| 1,738 secrets | ~1,414ms/secret | **40.9 minutes** |

The customer reports ~2,000 external cert routes causing router pods to fail
their 120s startup probe repeatedly and enter CrashLoopBackOff.

## Root Cause

In `addSecretEventHandler`, `WaitForCacheSync` was called inside
`defer s.lock.Unlock()`, meaning the global write lock was held while waiting
for each secret's informer to complete its initial LIST to etcd. No other
registration could begin until the previous one finished.

## Fix

Write the new `monitoredItem` into `s.monitors` **before** releasing the lock
(so concurrent callers for the same secret safely reuse the same informer),
then release the lock **before** `WaitForCacheSync`. This allows concurrent
`AddSecretEventHandler` calls for different secrets to perform their etcd
round-trips in parallel, reducing startup time from O(N × api\_latency) to
O(api\_latency).

On sync failure the map entry is cleaned up so a future caller can retry with
a fresh informer.

## Testing

All existing unit tests in `pkg/secret/...` pass.

A concurrent registration test should be added (tracked separately).

Fixes: OCPBUGS-77056

### DIFF
--- a/pkg/route/secretmanager/manager.go
+++ b/pkg/route/secretmanager/manager.go
@@ -85,27 +85,28 @@ func (m *manager) RegisterRoute(ctx context.Context, namespace, routeName, secre
 	m.handlersLock.Unlock()
 
 	// Add a secret event handler for the specified namespace and secret, with the handler functions.
+	// This call releases the monitor lock internally during WaitForCacheSync, allowing concurrent
+	// registrations for different secrets. However, it blocks until the cache is synced so that
+	// GetSecret works immediately after RegisterRoute returns.
 	klog.V(5).Infof("trying to add handler for key %s with secret %s", key, secretName)
 
-	go func() {
-		handlerReg, err := m.monitor.AddSecretEventHandler(ctx, namespace, secretName, handler)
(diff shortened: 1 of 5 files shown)


bentito:
Closing this PR. As we iterated on the solution, we discovered that the primary bottleneck for startup time was actually the synchronous and repetitive SubjectAccessReview (SAR) API checks in the router, rather than the WaitForCacheSync lock in library-go. We have implemented a router-only fix (openshift/router#745) that caches and throttles these SAR checks, making this lock release unnecessary."

PR #2144, opened 2026-03-18
OCPBUGS-66225: fix config observer degraded blip

tjungblu:
Use GetOperatorStateWithQuorum when retrying UpdateSpec so we get a fresh resourceVersion instead of reusing the stale one from the lister.

This was observed in the image registry operator recently. (https://redhat.atlassian.net/browse/OCPBUGS-66225?focusedCommentId=16463951)

### DIFF
--- a/pkg/operator/v1helpers/helpers.go
+++ b/pkg/operator/v1helpers/helpers.go
@@ -123,11 +123,22 @@ type UpdateOperatorSpecFunc func(spec *operatorv1.OperatorSpec) error
 func UpdateSpec(ctx context.Context, client OperatorClient, updateFuncs ...UpdateOperatorSpecFunc) (*operatorv1.OperatorSpec, bool, error) {
 	updated := false
 	var operatorSpec *operatorv1.OperatorSpec
+	previousResourceVersion := ""
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		oldSpec, _, resourceVersion, err := client.GetOperatorState()
 		if err != nil {
 			return err
 		}
+		if resourceVersion == previousResourceVersion {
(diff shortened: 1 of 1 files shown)


pkg/operator/configobserver/config_observer_controller.go
```diff
@@ -176,7 +176,7 @@ func (c ConfigObserver) sync(ctx context.Context, syncCtx factory.SyncContext) e
 	}
 
 	if err := c.updateObservedConfig(ctx, syncCtx, existingConfig, mergedObservedConfig); err != nil {
-		errs = []error{err}
+		errs = append(errs, fmt.Errorf("error while updating observed config: %w", err))
```
tjungblu: while reading this seemed to me like a bug because it was swallowing all previous errors, so just fixing this along the actual error below.

pkg/operator/v1helpers/helpers.go
```diff
@@ -123,11 +123,22 @@ type UpdateOperatorSpecFunc func(spec *operatorv1.OperatorSpec) error
 func UpdateSpec(ctx context.Context, client OperatorClient, updateFuncs ...UpdateOperatorSpecFunc) (*operatorv1.OperatorSpec, bool, error) {
 	updated := false
 	var operatorSpec *operatorv1.OperatorSpec
+	previousResourceVersion := ""
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		oldSpec, _, resourceVersion, err := client.GetOperatorState()
 		if err != nil {
 			return err
 		}
+		if resourceVersion == previousResourceVersion {
```
tjungblu, line 132: it's effectively the same now as below in UpdateStatus

PR #2147, opened 2026-03-20
pkg/operator/staticpod/startupmonitor: fix 100% cpu for loop

alrs:
This fixes an empty `for` loop that would grab 100% cpu. The loop remains, but it now has to wait 300 years before looping.

### DIFF
--- a/pkg/operator/staticpod/startupmonitor/cmd.go
+++ b/pkg/operator/staticpod/startupmonitor/cmd.go
@@ -3,6 +3,7 @@ package startupmonitor
 import (
 	"context"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/operator/staticpod/startupmonitor/cmd.go
```diff
@@ -201,6 +202,7 @@ func (o *Options) suicide(installerLock Locker) {
 	installerLock.Unlock()
 	klog.Info("Waiting for SIGTERM...")
 	for {
+		<-time.After(time.Duration(math.MaxInt64))
```
tchap, line 205: A more idiomatic way is to do `select {}` and be done with it.

Having said that, the right approach IMO is to pass a context into `suicide` and block on `<-ctx.Done()`. You can just pass the signal context then and be done with it.

PR #2162, opened 2026-04-14
CNTRLPLANE-3237: Add KMS preflight checker scaffolding

p0lyn0mial:
## Summary by CodeRabbit

* **New Features**
  * Added a KMS preflight command to run a one-shot Pod that validates KMS provider connectivity and behavior.
  * Command accepts provider and timeout flags, validates inputs, and reports pass/fail with timing.
  * Preflight runs status, encrypt, and decrypt checks and logs detailed outcomes and latencies.
  * Uses an embedded, templated Pod manifest that launches a lightweight container with minimal CPU/memory.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
+++ b/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{.PodName}}
+  namespace: {{.Namespace}}
+spec:
+  # This is a one-shot preflight check, not a long-running pod.
+  # The operator creates it, waits for completion, and inspects the result.
+  restartPolicy: Never
+  containers:
+    - name: kms-preflight-check
+      image: {{.OperatorImage}}
+      command: [{{.Command}}]
+      args:
+        - --kms-call-timeout={{.KMSCallTimeout}}
+      # TODO: figure out good resource request values.
+      resources:
+        requests:
+          memory: 50Mi
+          cpu: 5m
--- a/pkg/operator/encryption/kms/preflight/checker.go
+++ b/pkg/operator/encryption/kms/preflight/checker.go
@@ -1,0 +1,108 @@
+package preflight
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+	kmsservice "k8s.io/kms/pkg/service"
+)
+
+// healthzOK is the value the KMS plugin returns when healthy.
+// See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kms/apis/v2/api.proto#L39
+const healthzOK = "ok"
+
+// checker runs the preflight check against a KMS plugin by calling
+// Status, Encrypt, and Decrypt on the kmsservice.Service interface.
+// this is the same interface the apiserver uses.
+type checker struct {
+	service        kmsservice.Service
+	randReader     io.Reader
+	statusTimeout  time.Duration
+	statusInterval time.Duration
+}
+
+func newChecker(service kmsservice.Service) *checker {
+	return &checker{
+		service:        service,
+		randReader:     rand.Reader,
+		statusTimeout:  30 * time.Second,
+		statusInterval: 2 * time.Second,
+	}
+}
+
+func (c *checker) check(ctx context.Context) error {
+	if err := c.checkStatus(ctx); err != nil {
+		return err
+	}
+	return c.checkEncryptDecrypt(ctx)
+}
+
+// checkStatus polls the KMS plugin status endpoint until it reports healthy.
+// The plugin may not be immediately available after startup, so we retry
+// before reporting a failure.
+func (c *checker) checkStatus(ctx context.Context) error {
+	klog.Infof("[1/3] Checking KMS plugin status endpoint (interval %v, timeout %v)", c.statusInterval, c.statusTimeout)
+	return wait.PollUntilContextTimeout(ctx, c.statusInterval, c.statusTimeout, true, func(ctx context.Context) (bool, error) {
+		start := time.Now()
+		resp, err := c.service.Status(ctx)
+		elapsed := time.Since(start)
+		if err != nil {
+			klog.Infof("  not ready: %v, latency=%v", err, elapsed)
+			return false, nil
+		}
+		// we only check healthz here.
+		// version and keyID validation is the apiserver's responsibility,
+		// the preflight check just confirms the plugin is reachable and healthy.
+		if resp.Healthz != healthzOK {
+			klog.Infof("  not ready: healthz=%q, latency=%v", resp.Healthz, elapsed)
+			return false, nil
+		}
+		klog.Infof("  Status: healthz=%q, version=%q, keyID=%q, latency=%v", resp.Healthz, resp.Version, resp.KeyID, elapsed)
+		return true, nil
+	})
+}
+
+func (c *checker) checkEncryptDecrypt(ctx context.Context) error {
+	klog.Info("[2/3] Checking KMS plugin encrypt endpoint")
+	plainText := make([]byte, 32)
+	if _, err := io.ReadFull(c.randReader, plainText); err != nil {
+		return fmt.Errorf("failed to generate random plaintext: %w", err)
+	}
+	// unique per run, used for tracing
+	uid := fmt.Sprintf("preflight-check-%x", plainText[:4])
+	start := time.Now()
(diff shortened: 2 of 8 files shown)


pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -0,0 +1,25 @@
...
+      # TODO: figure out good resource request values.
+      resources:
+        requests:
+          memory: 50Mi
+          cpu: 5m
+      volumeMounts:
+        - name: kms-plugin-socket
+          mountPath: /var/run/kmsplugin
+  volumes:
+    # TODO: check with Fabio on the actual socket sharing mechanism.
```
p0lyn0mial: cc @bertinatto

pkg/operator/encryption/kms/preflight/pod_template.go
```diff
@@ -0,0 +1,61 @@
...
+	PodName       string
+	Namespace     string
+	OperatorImage string
+	Command       string
+}
+
+// GeneratePodTemplate renders the KMS preflight pod YAML template and calls
+// withKMSPluginContainerFn to inject the KMS plugin container into the pod.
+//
+// The base template defines the checker container and the shared emptyDir volume
```
p0lyn0mial: This is a tmp solution until we have the magic fn exposed by the plugin lifecycle.

ardaguclu:
Skeleton of pre-flight checker makes sense to me

jp8x7rgh6j-cmd (approved):

pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -0,0 +1,74 @@
...
+			return o.run(ctx)
+		},
+	}
+
+	o.addFlags(cmd.Flags())
+	return cmd
+}
+
+func (o *options) addFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&o.kmsProviderName, "kms-provider-name", "", "name of the KMS provider")
```
ardaguclu, line 42: What does this refer to?, same kms plugin (i.e. `kms-1.sock`) is represented as multiple providers (i.e. `1_secrets`, `1_configmaps`). What value this is supposed to get.

pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -0,0 +1,74 @@
...
+	fs.DurationVar(&o.kmsCallTimeout, "kms-call-timeout", 0, "timeout for each gRPC call to the KMS plugin")
+}
+
+func (o *options) validate() error {
+	if len(o.kmsProviderName) == 0 {
+		return fmt.Errorf("--kms-provider-name is required")
+	}
+	if o.kmsCallTimeout <= 0 {
```
ardaguclu, line 50: It’s preferable for this value to match the default timeout in the encryption controllers, right? We set it to 20 seconds and it works properly, but it may not work correctly when set to 10 seconds

ardaguclu (commented):
This is a nicely written code, thank you. Dropped a few questions.

pkg/operator/encryption/kms/preflight/cmd.go
p0lyn0mial: ok, i will remove this for now. we will figure it out later.

pkg/operator/encryption/kms/preflight/cmd.go
p0lyn0mial: yes, `kas-o` will pass a value (preferably the default one) when calling `GeneratePodTemplate`. does it make sense ?

pkg/operator/encryption/kms/preflight/checker.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	// unique per run, used for tracing
+	uid := fmt.Sprintf("preflight-check-%x", plaintext[:4])
+	start := time.Now()
+	encResp, err := c.service.Encrypt(ctx, uid, plaintext)
+	elapsed := time.Since(start)
+	if err != nil {
+		return fmt.Errorf("encrypt call failed, latency=%v: %w", elapsed, err)
+	}
+	if bytes.Equal(encResp.Ciphertext, plaintext) {
```
ardaguclu: These (i.e. `encResp.Ciphertext` and `plaintext`) should not equal. Am I missing something?

pkg/operator/encryption/kms/preflight/checker.go
```diff
@@ -0,0 +1,108 @@
...
+			return false, nil
+		}
+		klog.Infof("  Status: healthz=%q, version=%q, keyID=%q, latency=%v", resp.Healthz, resp.Version, resp.KeyID, elapsed)
+		return true, nil
+	})
+}
+
+func (c *checker) checkEncryptDecrypt(ctx context.Context) error {
+	klog.Info("[2/3] Checking KMS plugin encrypt endpoint")
+	plaintext := make([]byte, 32)
```
ardaguclu: nit:

```suggestion
	plainText := make([]byte, 32)
```

pkg/operator/encryption/kms/preflight/checker.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("encrypt call failed, latency=%v: %w", elapsed, err)
+	}
+	if bytes.Equal(encResp.Ciphertext, plaintext) {
+		return fmt.Errorf("encrypt returned plaintext unchanged, expected encrypted data")
+	}
+	var annotations []string
+	for k, v := range encResp.Annotations {
+		annotations = append(annotations, fmt.Sprintf("%s=%q", k, v))
+	}
+	klog.Infof("  Encrypt: keyID=%q, ciphertext=%d bytes, annotations=[%s], latency=%v", encResp.KeyID, len(encResp.Ciphertext), strings.Join(annotations, ", "), elapsed)
```
ardaguclu, line 93: Would this be simpler;

```go
klog.InfoS("Encrypt", 
    "keyID", encResp.KeyID, 
    "ciphertextBytes", len(encResp.Ciphertext), 
    "annotations", encResp.Annotations, 
    "latency", elapsed,
)
```

pkg/operator/encryption/kms/preflight/pod_template.go
```diff
@@ -0,0 +1,62 @@
...
+	PodName        string
+	Namespace      string
+	OperatorImage  string
+	Command        string
+	KMSCallTimeout string
+}
+
+// GeneratePodTemplate renders the KMS preflight pod YAML template.
+// The pod has a single container that connects to the KMS plugin via
+// a hostPath-mounted unix socket at /var/run/kmsplugin.
```
ardaguclu, line 27: Don't kms plugin and checker run on the same pod?. So we don't need hostPath-mounted unix socket. Or documentation is wrong?.

ardaguclu: If we use hostPath-mounted uds to use TPv1 functionality. Maybe it is better to just drop this from documentation.

p0lyn0mial: For now this is what we have. Once we have the plugin lifecycle we will have to change this code to reuse the function that knows how to install the plugin. Does it make sense ?

pkg/operator/encryption/kms/preflight/checker.go
p0lyn0mial: yes, thus the check, if they are then we return. does it make sense.

pkg/operator/encryption/kms/preflight/pod_template.go
ardaguclu: Yes, it makes sense. Do you think that we should remove this from the documentation or we'll update the documentation when we start using plugin lifecycle?

pkg/operator/encryption/kms/preflight/checker.go
p0lyn0mial: we usually don't use multiline formatting.

pkg/operator/encryption/kms/preflight/checker.go
p0lyn0mial: +1

pkg/operator/encryption/kms/preflight/checker.go
ardaguclu: ah, it makes sense. Ignore my comment :)

ardaguclu:
Rest of my comments are not blocker
/lgtm
/hold
in case you may want more review

pkg/operator/encryption/kms/preflight/pod_template.go
p0lyn0mial: We will update once we start using the plugin lifecycle.

pkg/operator/encryption/kms/preflight/checker.go
p0lyn0mial: np

p0lyn0mial:
> in case you may want more review

I accept post-review comments :)

/hold cancel

PR #2169, opened 2026-04-20
[WIP] CNTRLPLANE-3236: Kms plugin lifecycle

bertinatto:
proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2104

## Release Notes

* **New Features**
  * Added support for KMS provider configurations, enabling Vault-based KMS providers with configurable address, namespace, and transit settings.
  * Implemented automatic KMS plugin sidecar injection into API server pods when KMS encryption is enabled.
  * Enhanced encryption configuration to store and retrieve KMS provider settings.

* **Tests**
  * Expanded test coverage for KMS provider configuration handling and sidecar container injection.

### DIFF
--- a/pkg/operator/encryption/kms/helpers.go
+++ b/pkg/operator/encryption/kms/helpers.go
@@ -2,16 +2,24 @@ package kms
 
 import (
 	"fmt"
+	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/encryption/encoding"
 	corev1 "k8s.io/api/core/v1"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 )
+
+var kmsEndpointRegexp = regexp.MustCompile(`^unix:///var/run/kmsplugin/kms-(\d+)\.sock$`)
 
 const providerConfigDataKeyPrefix = "kms-provider-config-"
 const credentialDataKeyPrefix = "kms-secret-data-"
+const credentialsDir = "/etc/kubernetes/static-pod-resources/secrets/encryption-config"
 
 // ToProviderConfigSecretDataKeyFor constructs the data key for storing a KMS provider config in the encryption-config Secret.
 // The keyID must be a valid non-negative integer string.
@@ -112,4 +120,51 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName 
 	)
 
 	return nil
+}
+
+func findFirstKMSConfiguration(config *apiserverv1.EncryptionConfiguration) *apiserverv1.KMSConfiguration {
+	for _, resource := range config.Resources {
+		for _, provider := range resource.Providers {
+			if provider.KMS != nil {
+				return provider.KMS
+			}
+		}
+	}
+	return nil
+}
+
+func parseKeyIDFromEndpoint(endpoint string) (string, error) {
+	matches := kmsEndpointRegexp.FindStringSubmatch(endpoint)
+	if matches == nil {
+		return "", fmt.Errorf("unexpected KMS endpoint format: %s", endpoint)
+	}
+	return matches[1], nil
+}
+
+func parseProviderConfig(secret *corev1.Secret, kmsConfiguration *apiserverv1.KMSConfiguration) (*configv1.KMSConfig, error) {
+	keyID, err := parseKeyIDFromEndpoint(kmsConfiguration.Endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse key ID from endpoint: %w", err)
+	}
+	providerConfigKey, err := ToProviderConfigSecretDataKeyFor(keyID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create provider config secret key ID from endpoint: %w", err)
+	}
+	providerConfigData, ok := secret.Data[providerConfigKey]
+	if !ok {
+		return nil, fmt.Errorf("missing provider config key %s in encryption-config secret", providerConfigKey)
+	}
+	kmsConfig, err := encoding.DecodeKMSConfig(providerConfigData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode provider config: %w", err)
+	}
+	return kmsConfig, nil
+}
+
+func parseSecretDataPath(kmsConfiguration *apiserverv1.KMSConfiguration) (string, error) {
+	keyID, err := parseKeyIDFromEndpoint(kmsConfiguration.Endpoint)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse key ID from endpoint: %w", err)
+	}
+	return filepath.Join(credentialsDir, credentialDataKeyPrefix+keyID), nil
 }
--- a/pkg/operator/encryption/kms/helpers.go
+++ b/pkg/operator/encryption/kms/helpers.go
@@ -11,6 +11,7 @@ import (
 )
 
 const providerConfigDataKeyPrefix = "kms-provider-config-"
+const credentialDataKeyPrefix = "kms-secret-data-"
 
 // ToProviderConfigSecretDataKeyFor constructs the data key for storing a KMS provider config in the encryption-config Secret.
 // The keyID must be a valid non-negative integer string.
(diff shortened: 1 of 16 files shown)


pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,55 @@
...
+		v.Credentials.Data["VAULT_SECRET_ID"],
+		kmsConfig.Endpoint,
+		v.Config.VaultAddress,
+		v.Config.VaultNamespace,
+		v.Config.TransitMount,
+		v.Config.TransitKey,
+		v.Credentials.Data["VAULT_ROLE_ID"],
+	)
+
+	privileged := true
```
ardaguclu, line 45: Is there any particular reason to run the container privileged?

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,55 @@
...
+	return corev1.Container{
+		Name:    containerName,
+		Image:   v.Config.Image,
+		Command: []string{"/bin/sh", "-c"},
```
ardaguclu, line 49: I think we shouldn't have this `Command` anymore

pkg/operator/encryption/kms/helpers.go
```diff
@@ -64,3 +82,42 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
 
 	return nil
 }
+
+func ParseKeyIDFromEndpoint(endpoint string) (string, error) {
```
ardaguclu, line 86: Would it be better to directly extract key id from;

```go
regexp.MustCompile(`^unix:///var/run/kmplugin/kms-(\d+)\.sock$`)
```

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,40 @@
...
+	case config.Vault != nil:
+		return &plugins.VaultSidecarProvider{
+			Config: config.Vault,
+		}, nil
+	default:
+		return nil, fmt.Errorf("unsupported KMS provider configuration")
+	}
+}
+
+func AddSidecarContainer(podSpec *corev1.PodSpec, provider KMSSidecarProvider, containerName string, kmsConfig *apiserverv1.KMSConfiguration,
```
ardaguclu, line 27: Maybe this just returns `Container` and the caller uses its instead of implicit assignment.

ardaguclu (commented):
I dropped some comments but all of them are about mechanical changes. I think, core semantics is nice and aligns what we discussed.

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: I needed this to set up the UDS in the hostpath dir. Added a FIXME to remove this once I move to emptyDir

pkg/operator/encryption/kms/helpers.go
bertinatto: Done

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: I still need it because I'm temporarily doing `echo "%s" > /tmp/secret-id` before calling the plugin binary, so technically I need to run 2 commands, that's why I call `/bin/sh -c` instead of using the entrypoint. Once I can read the secret ID from the encryption configuration, then I'll be able to drop the `echo` command and the `Command` field.

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: Implemented emptyDir and removed this. I haven't tested it yet, let's see

pkg/operator/encryption/kms/helpers.go
ardaguclu: In my opinion, this commit is ready to merge whether before the API changes or ,ideally, after the API changes.

But it looks like this commit depends on https://github.com/openshift/library-go/pull/2163 to merge first.

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,51 @@
...
+
+	args := fmt.Sprintf(`
+	echo "%s" > /tmp/secret-id
+	exec /vault-kube-kms \
+	-listen-address=%s \
+	-vault-address=%s \
+	-vault-namespace=%s \
+	-transit-mount=%s \
+	-transit-key=%s \
+	-log-level=debug-extended \
```
p0lyn0mial: probably we don't want to run the vault plugin with verbose logs by default in production.
maybe this could be set via `unsupportedConfigOverrides` ? 

/cc @ardaguclu

ardaguclu: Agree. Log-level can be updated just as timeout value via unsupportedConfigOverrides (this mechanism does not exist today for KMS and we'll add it separately)

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +109,24 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
...
+	matches := kmsEndpointRegexp.FindStringSubmatch(endpoint)
+	if matches == nil {
+		return "", fmt.Errorf("unexpected KMS endpoint format: %s", endpoint)
+	}
+	return matches[1], nil
+}
+
+func GetKMSProviderConfig(secretData map[string][]byte, keyID string) (*configv1.KMSConfig, error) {
```
bertinatto: `GetKMSProviderConfig` -> `GetKMSConfigForKey`

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// FIXME: this is fragile. TBD
+	args := fmt.Sprintf(`set -e
+	CREDS=$(cat %s)
+	SECRET_ID=${CREDS#*\"VAULT_SECRET_ID\":\"}
+	SECRET_ID=${SECRET_ID%%%%\"*}
+	ROLE_ID=${CREDS#*\"VAULT_ROLE_ID\":\"}
+	ROLE_ID=${ROLE_ID%%%%\"*}
+	printf '%%s' "$SECRET_ID" > /tmp/secret-id
```
ardaguclu, line 36: Vault mock plugin does not need any credentials. I think we can update this PR to generate a side-car plugin container spec with the data coming from kms-provider-config. This is testable because we'll set vault kms plugin image from API directly.

pkg/operator/encryption/kms/helpers.go
```diff
@@ -2,16 +2,24 @@ package kms
...
 
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/encryption/encoding"
 	corev1 "k8s.io/api/core/v1"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 )
 
+var kmsEndpointRegexp = regexp.MustCompile(`^unix:///var/run/kmsplugin/kms-(\d+)\.sock$`)
```
ardaguclu, line 18: Agree that we should move this https://github.com/openshift/library-go/blob/6769d5fe19ec9b07192f20e71e6cf8c9cd9bc4ff/pkg/operator/encryption/controllers/key_controller.go#L43 to here.

pkg/operator/encryption/kms/helpers.go
```diff
@@ -113,3 +121,50 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
 
 	return nil
 }
+
+func findFirstKMSConfiguration(config *apiserverv1.EncryptionConfiguration) *apiserverv1.KMSConfiguration {
```
ardaguclu, line 125: I think we don't need to wire the first config anymore. kms-plugin-configs are coming from encryption-config Secret already. So it is just a matter of a loop over the keys.

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,192 @@
...
+	EncryptionConfigNamespace  string
+	EncryptionConfigSecretName string
+	APIServerContainerName     string
+}
+
+type SidecarProvider interface {
+	BuildSidecarContainer(name string, kmsConfiguration *apiserverv1.KMSConfiguration) (corev1.Container, error)
+}
+
+func newSidecarProvider(providerConfig *configv1.KMSConfig, secretDataPath string) (SidecarProvider, error) {
```
ardaguclu, line 25:
```suggestion
func newSidecarProvider(providerConfig *configv1.KMSConfig) (SidecarProvider, error) {
```
We can wire secret when it is needed

ardaguclu (commented):
In my opinion, this PR works nice to inject multiple kms plugins as sidecars by reading from kms-plugin-config-* data from encryption config Secret. We can remove all the credentials dependencies. (of course, meanwhile I'll work on passing credentials)

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,63 @@
...
+		return corev1.Container{}, fmt.Errorf("vault config cannot be nil")
+	}
+
+	// FIXME: this is fragile. TBD
+	args := fmt.Sprintf(`set -e
+	CREDS=$(cat %s)
+	SECRET_ID=${CREDS#*\"VAULT_SECRET_ID\":\"}
+	SECRET_ID=${SECRET_ID%%%%\"*}
+	ROLE_ID=${CREDS#*\"VAULT_ROLE_ID\":\"}
+	ROLE_ID=${ROLE_ID%%%%\"*}
```
tjungblu, line 35: @bertinatto your demo got me thinking:
* you can just directly write the args string from the kmsConfig, without the indirection of the env vars
* the secret_id mount is obviously difficult to get rid off...

for the secret_id we may create a dedicated secret text file that can be mounted directly? I don't know how this will end up in the static pod though - I assume we want to avoid going to the API with this directly due to the catch-22 of the encryption/decryption of secrets 😄

bertinatto: yeah, that'd be great, but unfortunately we can't mount the secret because this is a static pod; it doesn't work indeed :(

pkg/operator/encryption/kms/helpers.go
bertinatto: oh, I didn't recall this existed before. Are you suggesting I remove the one in key_controller.go and keep this one?

ardaguclu: Yes we can remove the one from key_controller. I think kms/helpers.go is better place.

bertinatto: hm, so that one is a string fmt, and this one is a regex

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: I created a new PR for the vault mock: https://github.com/openshift/library-go/pull/2209

pkg/operator/encryption/kms/helpers.go
bertinatto: ack, I can get the key ids from the secret's keys. The plugin also needs the endpoint path, do you suggest I reconstruct it using the information from the kms-plugin-config? Or is it better to plumb that information thru?

ardaguclu: encryption-config key contains uds path keyed by secret_KEYID, oauth_KEYID, etc. format.

PR #2171, opened 2026-04-21
OCPSTRAT-3036: Rebase 1.36.2

jubittajohn:
## Summary by CodeRabbit

* **Dependencies**
  * Go toolchain bumped to 1.26.0 and a broad refresh of library dependencies.

* **Chores**
  * CI build image tag updated.

* **Bug Fixes**
  * Standardized event/log formatting to avoid accidental format-string interpretation.

* **Tests**
  * Added informer "has synced" checker helpers to improve test reliability and prevent hanging waits.


miheer:
@jubittajohn can we bump this to K8s 1.36.1 ?

jubittajohn:
@miheer 
The PR is already bumped to `1.36.1`. https://github.com/openshift/library-go/pull/2171/changes#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L41 (We always try to bump to the latest available version). Sorry for missing to update the title and description to match it.

pkg/crypto/crypto_test.go
```diff
@@ -552,15 +552,29 @@ func TestServerCertRegeneration(t *testing.T) {
...
 // in the OpenShift TLS security profiles have corresponding mappings in
 // openSSLToIANACiphersMap. This ensures that when TLS profiles are translated
 // from OpenSSL format to IANA format, no ciphers are silently dropped.
+// Ciphers that Go's crypto/tls does not support are excluded from this check.
 func TestTLSProfileCipherSuitesHaveMappings(t *testing.T) {
 	var missingMappings []string
 
 	for profileType, profileSpec := range configv1.TLSProfiles {
 		for _, cipher := range profileSpec.Ciphers {
+			if _, unsupported := ciphersUnsupportedByGo[cipher]; unsupported {
```
p0lyn0mial, line 565: Instead of maintaining a separate `ciphersUnsupportedByGo` skip list, maybe we should rewrite the test to use `OpenSSLToIANACipherSuites` ?

That's the function operators actually call to translate cipher suites for go binaries. For example, see how cluster-kube-apiserver-operator uses it [here](https://github.com/openshift/cluster-kube-apiserver-operator/blob/ae0c9631284bca8cfcbc305674781e097f7238ae/vendor/github.com/openshift/library-go/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go#L108).

This change could be made in a new PR and would reduce the diff of this PR.

pkg/operator/staticpod/controller/guard/guard_controller_test.go
```diff
@@ -88,7 +100,14 @@ func (i FakeInfrastructureSharedInformer) GetController() cache.Controller    {
 func (i FakeInfrastructureSharedInformer) Run(stopCh <-chan struct{})         {}
 func (i FakeInfrastructureSharedInformer) RunWithContext(ctx context.Context) {}
 func (i FakeInfrastructureSharedInformer) HasSynced() bool                    { return i.HasSynced_ }
```
p0lyn0mial: could we update `FakeInfrastructureSharedInformer` to store `synced chan struct{}` and then update `HasSynced` to: 

```
HasSynced() bool {
      select {
      case <-i.synced:
          return true
      default:
          return false
      }
  }
```

pkg/operator/staticpod/controller/guard/guard_controller_test.go
```diff
@@ -88,7 +100,14 @@ func (i FakeInfrastructureSharedInformer) GetController() cache.Controller    {
...
-func (i FakeInfrastructureSharedInformer) LastSyncResourceVersion() string    { return "" }
+func (i FakeInfrastructureSharedInformer) HasSyncedChecker() cache.DoneChecker {
```
p0lyn0mial: then we could simplify this method to just `return &fakeSharedIndexInformerDone{synced: i.synced}`

p0lyn0mial: we could do the same in the other places.

pkg/operator/apiserver/controller/apiservice/apigroup.go
```diff
@@ -78,7 +78,7 @@ func checkDiscoveryForByAPIServices(ctx context.Context, recorder events.Recorde
 		err := checkDiscoveryForAPIService(ctx, restclient, apiService)
 		if err != nil {
 			groupVersionString := fmt.Sprintf("%s.%s", apiService.Spec.Group, apiService.Spec.Version)
-			recorder.Warningf("OpenShiftAPICheckFailed", fmt.Sprintf("%q failed with %v", groupVersionString, err))
+			recorder.Warningf("OpenShiftAPICheckFailed", "%s failed with %v", groupVersionString, err)
```
p0lyn0mial: `"%q failed with %v"` ?

pkg/operator/configobserver/oauth/observe_tokenconfig.go
```diff
@@ -61,7 +61,7 @@ func ObserveAccessTokenMaxAgeSeconds(genericlisters configobserver.Listers, reco
 	}
 
 	if existingAccessTokenMaxAgeSeconds != observedAccessTokenMaxAgeSeconds {
-		recorder.Eventf("ObserveAccessTokenMaxAgeSeconds", "%s changed from %d to %d", fieldAccessTokenMaxAgeSeconds,
+		recorder.Eventf("ObserveAccessTokenMaxAgeSeconds", "%s changed from %f to %f", fieldAccessTokenMaxAgeSeconds,
```
p0lyn0mial, line 64: `"%s changed from %d to %d"` ?

pkg/crypto/crypto_test.go
jubittajohn: `OpenSSLToIANACipherSuites` silently drops unknown ciphers by design. That means it can't distinguish between ciphers excluded because Go doesn't support them (like the 3 CBC SHA384/SHA256 suites) and ciphers missing from `openSSLToIANACiphersMap` by accident. If `openshift/api` adds a new cipher Go does support but someone forgets the mapping, this test wouldn't catch it. The skip list makes that distinction explicit I guess.

pkg/operator/configobserver/oauth/observe_tokenconfig.go
jubittajohn: "%s changed from %d to %d" returns the following error
```
pkg/operator/configobserver/oauth/observe_tokenconfig.go:64:71: (github.com/openshift/library-go/pkg/operator/events.Recorder).Eventf format %d has arg existingAccessTokenMaxAgeSeconds of wrong type float64
```

pkg/operator/apiserver/controller/apiservice/apigroup.go
jubittajohn: Updated

pkg/crypto/crypto_test.go
p0lyn0mial: OK, good point about OpenSSLToIANACipherSuites silently dropping unknown ciphers. Thanks.

should make ciphersUnsupportedByGo self validating ? it would be nice to know if a future go release adds a support for a cipher in the ciphersUnsupportedByGo. in that case we should remove the cipher from the ciphersUnsupportedByGo and start validating it. WDYT?

jubittajohn: That makes sense to me. It seems like a worthwhile improvement. Do you think it would be okay to merge the rebase PRs first and then address this in a follow-up PR?

p0lyn0mial: since this PR is still a WIP, we could open a new PR and merge it even before this one.

jubittajohn: This is addressed in this PR: https://github.com/openshift/library-go/pull/2320 which is now merged. Therefore the commit in discussion can be dropped from this rebase PR and this thread can be closed. Thank you Lukasz for the feedbacks and review.

tmshort:
What's the status of this PR? A lot of components need k8s v1.36 support

kannon92:
@jubittajohn 
seems rebase is close to landing, can the WIP be removed?

/hold

/retitle [OCPSTRAT-3036](https://redhat.atlassian.net/browse/OCPSTRAT-3036): Rebase 1.36.2

kannon92:
> What's the status of this PR? A lot of components need k8s v1.36 support

#tmp-k8s-1-36 in internal slack is a better place to ask status of rebase.

jubittajohn:
@kannon92 The reason I had WIP in the title is that it references my local branches for the api and client-go changes: https://github.com/openshift/library-go/pull/2171/changes/c0fa52b616c6e8568a8e6d7902407a7ab07f02eb#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R154-R158.
Once we start landing the Kubernetes rebase changes, the library bumps should land in the following order because of their dependencies: api → client-go → library-go.

That said, I'm not sure whether I should keep WIP in the title until it references the actual merged commits instead of my local branches.

jubittajohn:
>What's the status of this PR? A lot of components need k8s v1.36 support

+1 ing Kevin's message. `library-go` PR should land as a part of the Kube rebase process. The current status and all of the work around this can be tracked in the #tmp-k8s-1-36 slack channel

jubittajohn:
/retitle WIP: [OCPSTRAT-3036](https://redhat.atlassian.net/browse/OCPSTRAT-3036): Rebase 1.36.2
Moving it back to WIP since the PR can't be merged in its current state, and we don't want it to be merged accidentally.

PR #2172, opened 2026-04-22
CNTRLPLANE-3237: Bundle encryption-config Secret to be accessible by all controllers

ardaguclu:
This was requested in https://github.com/openshift/library-go/pull/2163#discussion_r3116059379 and https://github.com/openshift/library-go/pull/2163#discussion_r3116281961. The main motivation is to make all fields in the encryption-config Secret accessible in a round-trippable way, so both existing and future controllers (e.g., pre-flight checker) can use them.

This PR bundles encryption-config Secret with all the data. Previously it was only carrying EncryptionConfiguration but this is not the case anymore. Every controller needs to access all the fields of encryption-config which is supposed to include EncryptionConfiguration, KMSProviderConfig, SecretRefs, ConfigMapRefs, etc.

## Summary by CodeRabbit

* **Refactor**
  * Centralized and replaced the internal encryption configuration representation across the operator so encryption secrets and state are handled via a single, consistent data wrapper and naming scheme.

* **Tests**
  * Updated and expanded tests to exercise the new encryption data pathways, builders, and secret conversions to preserve behavior and prevent regressions.


ardaguclu:
Opened https://github.com/openshift/cluster-kube-apiserver-operator/pull/2119 to test
/uncc @tkashem 
/cc @p0lyn0mial @bertinatto 

edited: All the tests passed in https://github.com/openshift/cluster-kube-apiserver-operator/pull/2119

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -259,7 +260,7 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 	return nil // we made this key earlier
 }
 
-func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, internalReason, externalReason string) (*corev1.Secret, error) {
+func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, apiServerEncryption *v1.APIServerEncryption, internalReason, externalReason string) (*corev1.Secret, error) {
```
p0lyn0mial: `apiServerEncryption` is not used. let's add it when it is needed.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -271,40 +272,43 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
+			},
+			// TODO: apiserverEncryption.KMSConfig will be wired to KMS.ProviderConfig
 		}
 	}
 	return secrets.FromKeyState(c.instanceName, ks)
 }
 
-func (c *keyController) getCurrentModeAndExternalReason(ctx context.Context) (state.Mode, string, error) {
+func (c *keyController) getCurrentModeAndExternalReason(ctx context.Context) (state.Mode, string, *v1.APIServerEncryption, error) {
```
p0lyn0mial: `*v1.APIServerEncryption` is unused. could we change the signature of this method when it is needed?

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -185,7 +185,7 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
 	desiredEncryptedConfig := encryptionconfig.FromEncryptionState(desiredEncryptionState)
```
p0lyn0mial: seems like we should rename `desiredEncryptedConfig` since we changed the type returned from the `FromEncryptionState` method.

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -20,23 +20,32 @@ var (
...
+func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) *state.EncryptionSecretData {
 	resourceConfigs := make([]apiserverconfigv1.ResourceConfiguration, 0, len(encryptionState))
+	kmsConfigs := make(map[string]*state.KMSConfig)
 
 	for gr, grKeys := range encryptionState {
 		resourceConfigs = append(resourceConfigs, apiserverconfigv1.ResourceConfiguration{
 			Resources: []string{gr.String()}, // we are forced to lose data here because this API is broken
 			Providers: stateToProviders(gr.Resource, grKeys),
 		})
+		for _, key := range grKeys.ReadKeys {
```
p0lyn0mial: could we not introduce behavioral changes in this PR? Could we just focus on the refactoring?

pkg/operator/encryption/encryptionconfig/secret.go
```diff
@@ -44,12 +44,19 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*apiserverconfigv1.Encry
 	if !ok {
 		return nil, fmt.Errorf("unexpected wrong type %T", encryptionConfigObj)
 	}
-	return encryptionConfig, nil
+	return &state.EncryptionSecretData{
+		EncryptionConfig: encryptionConfig,
+		KMSConfig:        kmsConfigsFromEncryptionConfig(encryptionConfig),
```
p0lyn0mial: could we not introduce behavioral changes in this PR? Could we just focus on the refactoring?

pkg/operator/encryption/encryptionconfig/config.go
```diff
@@ -20,23 +20,32 @@ var (
 	emptyStaticKey = base64.StdEncoding.EncodeToString(crypto.NewEmptyKey())
 )
 
-// FromEncryptionState converts state to config.
-func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) *apiserverconfigv1.EncryptionConfiguration {
+// FromEncryptionState converts encryption state to state.EncryptionSecretData.
+func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) *state.EncryptionSecretData {
```
p0lyn0mial: I think that we could trace all callers and update the local val that previously assumed EC

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -20,23 +20,32 @@ var (
 	emptyStaticKey = base64.StdEncoding.EncodeToString(crypto.NewEmptyKey())
 )
```
p0lyn0mial, line 21: would it make sense to rename the pkg to `encryptiondata` or something similar ?
we could do that in a new PR.

pkg/operator/encryption/state/types.go
```diff
@@ -40,8 +40,20 @@ type KeyState struct {
...
+}
+
+// KMSConfig stores all KMS encryption mode related configurations
+type KMSConfig struct {
+	// Encoded EncryptionConfig that stores the KMS related fields
+	EncryptionConfig *apiserverconfigv1.KMSConfiguration
+}
+
+// EncryptionSecretData stores the EncryptionConfiguration and per-key KMS configurations keyed by keyID.
+type EncryptionSecretData struct {
```
p0lyn0mial: shouldn't this type be added to `encryptionconfig/encryptiondata` pkg?

p0lyn0mial: and once we move to `encryptiondata` then maybe we could rename it to `Config` ? (`encryptiondata.Data`) ?

p0lyn0mial:
experimented a bit with your PR in https://github.com/openshift/library-go/pull/2174

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated.

pkg/operator/encryption/controllers/migration_controller.go
ardaguclu: Updated

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Just for clarification: When you asked me there is any behavioral changes, I replied that no there isn't. But I didn't consider behavioral additions as behavioral changes. Probably I should.

To sum up: All behavioral changes are removed from this PR.

pkg/operator/encryption/encryptionconfig/secret.go
ardaguclu: Updated

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: Updated

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: I've renamed the package name to `encryptiondata` in second commit.

pkg/operator/encryption/state/types.go
ardaguclu: Updated. New usage is `encryptiondata.Config`

ardaguclu:
> experimented a bit with your PR in #2174

@p0lyn0mial I've updated this PR based on your suggestions and mimicking some changes from https://github.com/openshift/library-go/pull/2174. All the behavioral changes are removed. Package name is renamed to `encryptiondata` in second commit. 

I'll update this fake bump https://github.com/openshift/cluster-kube-apiserver-operator/pull/2119 to test the latest changes.

pkg/operator/encryption/encryptionconfig/config.go
ardaguclu: KMS struct changes are removed from this PR as well. Current state of the PR is pure mechanical renames.

pkg/operator/encryption/controllers/state_controller.go
```diff
@@ -144,14 +143,14 @@ func (c *stateController) generateAndApplyCurrentEncryptionConfigSecret(ctx cont
...
-	changed, err := c.applyEncryptionConfigSecret(ctx, desiredEncryptionConfig, recorder)
+	desiredSecretData := encryptiondata.FromEncryptionState(desiredEncryptionState)
+	changed, err := c.applyEncryptionConfigSecret(ctx, desiredSecretData, recorder)
 	if err != nil {
 		return err
 	}
 
 	if changed {
-		currentEncryptionConfig, _ := encryptionconfig.ToEncryptionState(currentConfig, encryptionSecrets)
+		currentEncryptionConfig, _ := encryptiondata.ToEncryptionState(currentConfig, encryptionSecrets)
```
bertinatto, line 153: I'm getting very confused with the naming :)

Here, the param `currentConfig` is a `*encryptiondata.Config` var, right? 

Most of the time throughout the code it's called `secretData`, or `desiredSecretData` (which makes sense) - but here it's different; also in migration_controller.go:185 it's called `desiredEncryptionState`. There are possibly others.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -20,8 +20,14 @@ var (
...
 
-// FromEncryptionState converts state to config.
-func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) *apiserverconfigv1.EncryptionConfiguration {
+// Config stores the EncryptionConfiguration and additional operator-managed
+// encryption state that doesn't fit into the upstream type.
+type Config struct {
+	Encryption *apiserverconfigv1.EncryptionConfiguration
+}
```
bertinatto, line 27: Based on the discussion we had in the other PR, I thought we were going to bundle the kms-related fields in the `KeyState` struct, no? We could also start bundling the data in that struct before doing a repo-wide refactoring, right?

pkg/operator/encryption/deployer/unionrevisionedpod.go
```diff
@@ -50,18 +50,18 @@ func (d *UnionRevisionLabelPodDeployer) DeployedEncryptionConfigSecret(ctx conte
...
 
 	for _, secret := range seenSecrets {
-		currentEncryptionCfg, err := encryptionconfig.FromSecret(secret)
+		currentEncryptionCfg, err := encryptiondata.FromSecret(secret)
 		if err != nil {
 			return nil, false, err
 		}
 
-		if !reflect.DeepEqual(goldenEncryptionCfg.Resources, currentEncryptionCfg.Resources) {
+		if !reflect.DeepEqual(goldenEncryptionCfg.Encryption.Resources, currentEncryptionCfg.Encryption.Resources) {
```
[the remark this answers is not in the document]
ardaguclu: I think this problem already exists in main branch. This is not specific to the changes here. However, I agree that we need to gracefully handle nil case. Maybe in a separate follow up PR.

pkg/operator/encryption/encryptiondata/config_test.go
```diff
@@ -428,7 +430,7 @@ func TestToEncryptionState(t *testing.T) {
 
 	for _, scenario := range scenarios {
 		t.Run(scenario.name, func(t *testing.T) {
-			actualOutput, _ := ToEncryptionState(scenario.input, nil)
+			actualOutput, _ := encryptiondata.ToEncryptionState(scenario.input, nil)
```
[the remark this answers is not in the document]
ardaguclu: This PR solely focuses on mechanical changes for refactoring. I'm not sure we want to change this here. But I agree that we need to handle the error in a separate PR maybe?

pkg/operator/encryption/encryptiondata/testing/helpers.go
```diff
@@ -0,0 +1,132 @@
...
+				{
+					Identity: &apiserverconfigv1.IdentityConfiguration{},
+				},
+			},
+		}
+		for i, key := range keysResource.Keys {
+			desiredMode := ""
+			if len(keysResource.Modes) == len(keysResource.Keys) {
+				desiredMode = keysResource.Modes[i]
+			}
```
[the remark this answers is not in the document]
ardaguclu: Isn't this problem existing in main branch as well?. This is not specific to the changes here.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -36,7 +42,9 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 		return resourceConfigs[i].Resources[0] < resourceConfigs[j].Resources[0] // each resource has its own keys
 	})
 
-	return &apiserverconfigv1.EncryptionConfiguration{Resources: resourceConfigs}
+	return &Config{
```
p0lyn0mial: nit: could we make it one-liner ?

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -44,12 +44,18 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*apiserverconfigv1.Encry
 	if !ok {
 		return nil, fmt.Errorf("unexpected wrong type %T", encryptionConfigObj)
 	}
-	return encryptionConfig, nil
+	return &Config{
```
p0lyn0mial: nit: could we make it one-liner ?

p0lyn0mial (commented):
lgtm, just two nits :)

pkg/operator/encryption/encryptiondata/config.go

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: Updated

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -63,12 +69,12 @@ func ToEncryptionState(encryptionConfig *apiserverconfigv1.EncryptionConfigurati
 	}
 	backedKeys = state.SortRecentFirst(backedKeys)
 
-	if encryptionConfig == nil {
+	if secretData == nil || secretData.Encryption == nil {
```
p0lyn0mial, line 72: in the future, when there are more places that check nil we could intro

```
func (c *Config) HasEncryptionConfiguration() bool {
      return c != nil && c.Encryption != nil
  }
```

then this would become:

```
if !secretData.HasEncryptionConfiguration() {
      return nil, backedKeys
  }
```

ardaguclu: Good idea

pkg/operator/encryption/deployer/unionrevisionedpod.go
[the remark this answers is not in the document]
ardaguclu: @p0lyn0mial we may need to fix this in a followup PR to prevent any panic. This is a problem in main branch.

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: https://github.com/openshift/library-go/pull/2179 this is for bundling kms related configs in KeyState

pkg/operator/encryption/controllers/state_controller.go
p0lyn0mial: we could make the names consistent, for example: `desiredSecretData`

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: yes, this will come in a new pr.

pkg/operator/encryption/controllers/state_controller.go
ardaguclu: @p0lyn0mial asked me to update all the parameters accordingly and I updated https://github.com/openshift/library-go/pull/2172#discussion_r3130441588. However, it seems that there are more to be updated. This is my bad actually.

ardaguclu: According to my analysis, this PR correctly updates the namings due to the signature change in FromEncryptionState function. However, `ToEncryptionState` returns `map[schema.GroupResource]state.GroupResourceState, []state.KeyState`. 

Local variable `currentEncryptionConfig` is wrong from the beginning.

PR #2175, opened 2026-04-23
SPLAT-2651: Added vSphere cloud config modules from 3CMO

vr4manta:
[SPLAT-2651](https://redhat.atlassian.net/browse/SPLAT-2651)

### Changes
- Added 3CMO vSphere cloud provider config module for sharing across other projects
- Vendor updates

### Notes
Here are some questions asked about this PR: 

> Why the code needs to be in library-go and not in CCCMO?

Because multiple operators have this logic including installer.  Why copy/pasta when you can share?

> Why the code copies code from the cloud provider and does not vendor it? There must be a simple reason, it's just not written anywhere.

From what I can tell w/ the history of 3CMO, is they originally copied it from upstream to reduce logging and other stuff.  This was a similar issue I ran into when I first added multi vcenter support as day 0.  Also, now that we have changed from yaml to json for one of the commits, there is some diff from upstream.

## Summary by CodeRabbit

* **New Features**
  * vSphere cloud provider configuration support with INI↔YAML parsing and YAML marshaling.

* **Tests**
  * Added comprehensive tests for parsing, conversion, validation, and error cases (malformed and empty inputs).

* **Chores**
  * Added new dependency requirements to support parsing and YAML handling.


vr4manta:
/hold
Testing w/ other PRs

pkg/cloudprovider/vsphere/config.go
```diff
@@ -0,0 +1,47 @@
+package vsphere_cloud_config
```
JoelSpeed: Does this affect the import path when you're importing into other packages? Maybe just make this vsphere to match the dir

pkg/cloudprovider/vsphere/yaml.go
```diff
@@ -0,0 +1,159 @@
+package vsphere_cloud_config
+
+import (
+	"fmt"
+
+	yaml "gopkg.in/yaml.v2"
```
JoelSpeed: Out of interest, why this yaml library? Might be worth standardising on the sigs K8s yaml library? (Uses `json` instead of `yaml` tags)

pkg/cloudprovider/vsphere/config.go
vr4manta: good point, i did copy over as is.  I'll update to match directory.

pkg/cloudprovider/vsphere/yaml.go
vr4manta: This was what it was using in 3CMO.  I can try migrating and verify no major changes.

vr4manta: Changing the package did result in a lot of changes in behavior.  The biggest is how yaml would exclude if empty, but k8s will result in empty array objects.  I converted to pointer and modified logic of dependent projects to check for nil when processing the config.  3CMO i had to make a few minor changes that should be reviewed as well.  The installer seems to be ok since we were just setting and not reading or transforming.

pkg/cloudprovider/vsphere/yaml.go
```diff
@@ -0,0 +1,159 @@
...
+// Labels tags categories and tags which correspond to "built-in node labels: zones and region"
+type Labels struct {
+	Zone   string `yaml:"zone,omitempty" json:"zone,omitempty"`
+	Region string `yaml:"region,omitempty" json:"region,omitempty"`
+}
+
+// CommonConfig is used to read and store information from the cloud configuration file
+type CommonConfig struct {
+	// Global values...
+	Global Global `json:"global,omitempty"`
```
JoelSpeed, line 112: Missing the yaml tag, but, I'd probably say remove the yaml tags and just have json tags, WDYT?

JoelSpeed (commented):
Should be consistent with either all or no yaml tags, IMO no yaml tags, WDYT @vr4manta ?

pkg/cloudprovider/vsphere/yaml.go
vr4manta: For whatever reason, the original never had a yaml tag: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/main/pkg/cloud/vsphere/vsphere_cloud_config/yaml.go#L112

So are you saying just remove all the yaml tags from this file?  I can try and see what happens.  As long as it still outputs as it has in the base and we are not changing too much functionality, then I'm cool with that.

PR #2178, opened 2026-04-24
OCPBUGS-80958: Kube apiserver node replace

bhperry:
When master node is deleted and recreated with the same name, ensure the node status is reset by comparing UID.

## Related PRs:
https://github.com/openshift/api/pull/2821
https://github.com/openshift/client-go/pull/371
https://github.com/openshift/cluster-kube-apiserver-operator/pull/2199

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Static Pod Operator node status pruning and replacement logic to be UID-aware, preserving existing entries only when the stored UID matches the current node’s UID and handling unknown/mismatched UIDs correctly.
* **New Features**
  * Newly observed node statuses now record the node UID for more reliable tracking.
* **Tests**
  * Added UID-focused test coverage, including deterministic fake node UID assignment and assertions for preserve/replace/remove behavior.
* **Chores**
  * Updated module resolution for OpenShift components.


pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -84,22 +85,28 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
-			}
-		}
-		if found {
-			newTargetNodeState := applyoperatorv1.NodeStatus().WithNodeName(originalOperatorStatus.NodeStatuses[i].NodeName)
+		currentUID, found := nodeUIDs[nodeState.NodeName]
+		if found && (nodeState.NodeUID == currentUID || nodeState.NodeUID == "") {
+			newTargetNodeState := applyoperatorv1.NodeStatus().
+				WithNodeName(nodeState.NodeName).
+				WithNodeUID(currentUID)
 			newTargetNodeStates = append(newTargetNodeStates, newTargetNodeState)
```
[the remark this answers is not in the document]
bhperry: Correct. Any existing nodes that are already in a bad state will need to be manually fixed by removing the status entry.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -84,22 +85,31 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 	// remove entries for missing nodes
 	for i, nodeState := range originalOperatorStatus.NodeStatuses {
-		found := false
-		for _, node := range nodes {
-			if nodeState.NodeName == node.Name {
-				found = true
+		currentUID, found := nodeUIDs[nodeState.NodeName]
+		if found && (nodeState.NodeUID == currentUID || nodeState.NodeUID == "") {
+			if nodeState.NodeUID == "" {
+				syncCtx.Recorder().Warningf("MasterNodeUnknownUID", "Node %s has no recorded UID. If this node was replaced its status may need to be manually removed from operator objects.")
```
atiratree: We can't have this, during an upgrade to a version that introduces NodeUID field, this value will be empty for all nodes. Suggesting a manual resolution during an upgrade is not very user friendly.

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -84,22 +85,31 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 			}
-		}
-		if found {
-			newTargetNodeState := applyoperatorv1.NodeStatus().WithNodeName(originalOperatorStatus.NodeStatuses[i].NodeName)
+			newTargetNodeState := applyoperatorv1.NodeStatus().
+				WithNodeName(nodeState.NodeName).
+				WithNodeUID(currentUID)
```
atiratree, line 101: Do we set or plan to set an empty UID knowingly anywhere in our code base? If not we can just always simply overwrite the empty values, no?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -84,22 +85,31 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
...
 			newTargetNodeStates = append(newTargetNodeStates, newTargetNodeState)
 		} else {
-			syncCtx.Recorder().Warningf("MasterNodeRemoved", "Observed removal of master node %s", nodeState.NodeName)
+			if found {
```
atiratree: I would expect the original `MasterNodeRemoved` to catch the UID change.

The node reprovisioning must have been very quick while the node controller was down or slow to sync?

The bug that we are fixing is most likely rare.

pkg/operator/staticpod/controllers.go
```diff
@@ -287,6 +288,7 @@ func (b *staticPodOperatorControllerBuilder) ToControllers() (manager.Controller
 			configMapClient,
 			secretClient,
 			podClient,
+			nodeClient,
```
atiratree: We should strive to use listers where we can, especially when we poll nodes in a loop.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -494,6 +498,14 @@ func (c *InstallerController) manageInstallationPods(ctx context.Context, operat
 			nodeChoiceReason = fmt.Sprintf("node %s is the next node in the line", currNodeState.NodeName)
 		}
 
+		// Verify the node UID matches the actual node before updating its status.
+		// A UID mismatch means the node was replaced and the node controller will handle resetting status.
+		if uidChanged, uidErr := c.hasNodeUIDChanged(ctx, currNodeState.NodeName, currNodeState.NodeUID); uidErr != nil {
+			return true, 0, nil, nil, uidErr
+		} else if uidChanged {
+			return true, c.installerBackOff(1), nil, nil, nil
```
atiratree: The sync should be triggered by an informer so we don't need the backoff/requeue.

pkg/operator/staticpod/controller/node/node_controller.go
bhperry: Fair point. The case where this would be an issue should be really rare anyway.

pkg/operator/staticpod/controller/node/node_controller.go
bhperry: No, UID should never be set empty. After the upgrade and operators sync it should be set at all times.

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: true, will return `false, 0` then instead

pkg/operator/staticpod/controller/node/node_controller.go
bhperry: Right, it is quite rare. The only way I've been able to reproduce the bug is by stopping the kube-apiserver-operator, replacing the node, and then starting the operator back up again. So I think the most likely real world case where it would occur is if the node that the operator is running on goes down and the pod isn't able to be scheduled until the new node (with the same name) comes up.

Are you suggesting that I remove the MasterNodeReplaced warning and just use MasterNodeRemoved in either case? I guess as long as we include node name and uid on MasterNodeRemoved as well as on MasterNodeObserved then an admin reading the logs would be able to see what happened.

pkg/operator/staticpod/controllers.go
bhperry: Yes this was just to test. I was thinking if I switched it to an informer here then getter should be ok, but I agree it shouldn't use an API get for each node.

What I was trying to avoid is listing ALL nodes when we actually only need the ones that are in the `status.nodeStatuses`. In large cluster that could be quite a bit of overhead.

pkg/operator/staticpod/controller/node/node_controller.go
atiratree: +1, MasterNodeRemoved with the old UID should be enough

pkg/operator/staticpod/controllers.go
atiratree: with shared informers the cache is already present and shared with the node controller, thx

bhperry:
/hold

depends on api and client-go change

pkg/operator/v1helpers/test_helpers.go
```diff
@@ -301,7 +301,11 @@ func (n *fakeNodeLister) List(selector labels.Selector) ([]*corev1.Node, error)
 }
 
 func (n *fakeNodeLister) Get(name string) (*corev1.Node, error) {
-	panic("implement me")
+	node, err := n.client.CoreV1().Nodes().Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return node, nil
```
bertinatto: nit: `return  n.client.CoreV1().Nodes().Get(context.TODO(), name, metav1.GetOptions{})`?

pkg/operator/staticpod/controller/node/node_controller.go
```diff
@@ -116,16 +118,19 @@ func (c *NodeController) sync(ctx context.Context, syncCtx factory.SyncContext)
 	for _, node := range nodes {
 		found := false
 		for _, nodeState := range originalOperatorStatus.NodeStatuses {
+			// NodeUID is intentionally ignored here. Nodes that are replaced with the same name will get added in the next sync
+			// to ensure previous nodeStatus from a stale cache isn't merged in with the fresh status.
 			if nodeState.NodeName == node.Name {
 				found = true
+				break
```
bertinatto, line 125: nice catch

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1123,6 +1135,25 @@ func (c InstallerController) ensureRequiredResourcesExist(ctx context.Context, r
...
+	node, err := c.nodeLister.Get(nodeName)
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Skipping node status update for %s: node not found", nodeName)
+		return true, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	if string(node.UID) != nodeUid {
+		klog.Warningf("Skipping node status update for %s: node UID changed (stored=%s, actual=%s)", nodeName, nodeUid, node.UID)
```
bertinatto, line 1153: it seems like this warning should be moved to the caller? `hasNodeUIDChanged()`

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -599,6 +609,7 @@ func prepareNodeStatusApplyConfigurationFor(nodeStatuses []operatorv1.NodeStatus
 func nodeStatusToNodeStatusApplyConfigurations(nodeStatus operatorv1.NodeStatus) *applyoperatorv1.NodeStatusApplyConfiguration {
 	nodeStatusApplyConfiguration := applyoperatorv1.NodeStatus().
 		WithNodeName(nodeStatus.NodeName).
+		WithNodeUID(nodeStatus.NodeUID).
```
bertinatto: for existing nodes, `NodeUID` is an empty string, correct? In that case, this will try to set `NodeUID` to an empty string in the api, which might fail validation because the field, [although optional, doesn't accept an empty string as a valid value](https://github.com/openshift/api/blob/master/operator/v1/types.go#L275-L279). Could you check that?

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1123,6 +1135,25 @@ func (c InstallerController) ensureRequiredResourcesExist(ctx context.Context, r
 	return fmt.Errorf("missing required resources: %v", aggregatedErr)
 }
 
+func (c *InstallerController) hasNodeUIDChanged(nodeName, nodeUid string) (bool, error) {
+	if nodeUid == "" {
+		return false, nil
+	}
+	node, err := c.nodeLister.Get(nodeName)
+	if apierrors.IsNotFound(err) {
+		klog.Warningf("Skipping node status update for %s: node not found", nodeName)
```
bertinatto, line 1146: it seems like the warning shouldn't assume the reason the function is being called

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: Warning is placed here because there are two cases where we skip. Node was deleted, or node exists but UID doesn't match. Different message for each case.

Perhaps the name of the function should be `skipNodeStatusUpdate` since that is really the purpose it serves?

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: See https://github.com/openshift/library-go/pull/2178#discussion_r3732311290

This function evolved since I first wrote it, so I am updating the name to clarify its intent.

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: Correct, it will start empty for existing nodes. You might be right on that, I will check.

pkg/operator/staticpod/controller/installer/installer_controller.go
bertinatto: renaming the function sounds like a good idea. `skipNodeStatusUpdate` reads like it's actually going to do the skipping, but that's not what it does. I think `shouldSkipNodeStatusUpdate` (or a variation of that) would be a better name

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -1164,6 +1195,17 @@ func (c *InstallerController) Sync(ctx context.Context, syncCtx factory.SyncCont
...
+	for _, nodeStatus := range originalOperatorStatus.NodeStatuses {
+		if shouldSkip, err := c.skipNodeStatusUpdate(nodeStatus.NodeName, nodeStatus.NodeUID); err != nil {
+			return errors.Wrapf(err, "error validating uid for node %s", nodeStatus.NodeName)
+		} else if shouldSkip {
+			// Avoid applying node statuses when a stale entry is found
+			// Sync will retrigger when the old node status is removed by the node controller
+			return nil
+		}
+	}
+
```
bertinatto, line 1210: I'm nervous about this because it takes only a single stale `nodeStatus` to prevent the installer pod from running and copying **all** revisioned resources to the static pods. On top of it, we have a bare informer that will not trigger a resync when the node controller does fix the stale `nodeStatus`. The controller also doesn't resync on any interval, so we'd rely on the other informers only.

pkg/operator/staticpod/controller/installer/installer_controller.go
bertinatto: sounds good

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: Bare informer is for the Node kind only. When the node controller removes the entry from nodeStatuses that triggers an update through the static pod operator informer which will trigger a resync in the installer controller.

I originally considered having the mismatched node-status removed here as well so the other entries can update, but I didn't want to cause contention between the installer and node controllers. Node controller already handles adding and removing from the list of nodeStatuses, so it felt like a better separation of concerns for that to be handled only there. Without skipping stale entries, the installer and node controllers end up fighting back and forth over it. Entry gets removed in node and added back in installer in an infinite loop. No way to update everything except for the stale entry as far as I know.

Any time we see a stale entry in the installer controller, we know the node controller is either about to remove it or already did and the cache just hasn't synced yet. As soon as the cache updates with that change the static pod operator informer will trigger a resync and update all valid entries.

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: Updated to only apply when not empty. Good catch.

pkg/operator/staticpod/controller/installer/installer_controller.go
bertinatto: > Bare informer is for the Node kind only. When the node controller removes the entry from nodeStatuses that triggers an update through the static pod operator informer which will trigger a resync in the installer controller.

you're right, I didn't realize we have a non-bare `operatorClient` informer that will make this controller resync when an entry in nodeStatuses changes, so on that front we should be good.

> I originally considered having the mismatched node-status removed here as well so the other entries can update, but I didn't want to cause contention between the installer and node controllers. Node controller already handles adding and removing from the list of nodeStatuses, so it felt like a better separation of concerns for that to be handled only there. Without skipping stale entries, the installer and node controllers end up fighting back and forth over it. Entry gets removed in node and added back in installer in an infinite loop. No way to update everything except for the stale entry as far as I know.
> 
> Any time we see a stale entry in the installer controller, we know the node controller is either about to remove it or already did and the cache just hasn't synced yet. As soon as the cache updates with that change the static pod operator informer will trigger a resync and update all valid entries.

I believe you do the right thing by not removing the entry in nodeStatus from this controller. Node controller should definitely own that task.  My main concern is related to the fact that **all** new revisioned resources for **all** nodes are gated if there's a single stale entry in nodeStatuses.

I understand it's just a matter of waiting for the node controller to do its job, and then this controller would be able to copy all revisioned resources, but this dependency on the node controller makes me wonder if we can ever get into a chicken-and-egg problem.

From what I investigated, this should not be an issue, so let's go ahead and merge this and keep an eye on it. Thanks for bearing with me.

bertinatto (commented):
/lgtm

Thanks! @bhperry could you bump library-go in KASO to make sure we don't let other changes pile up (it's easier to catch regressions).

pkg/operator/staticpod/controller/installer/installer_controller.go
bhperry: No problem at all. I agree it's good to call out and keep an eye on this behavior.

PR #2182, opened 2026-04-27
OCPBUGS-62792: Add unit tests for pkg/controller/controllercmd/cmd.go

jubittajohn:
This unit test was written by Claude

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage to improve reliability of configuration handling and TLS certificate behavior in the controller. Tests validate generation and validity windows of self-signed certificates, ensure preconfigured certificates are respected (no unintended regeneration), and verify configuration file detection and parsing. These tests increase confidence in startup/configuration flows and certificate rotation behavior.

### DIFF
--- a/pkg/controller/controllercmd/cmd_test.go
+++ b/pkg/controller/controllercmd/cmd_test.go
@@ -1,0 +1,312 @@
+package controllercmd
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+)
+
+func TestAddDefaultRotationToConfig_SelfSignedCertLifetime(t *testing.T) {
+	// Skip if service-serving-cert files exist, as we need to test the self-signed cert path
+	if hasServiceServingCerts("/var/run/secrets/serving-cert") {
+		t.Skip("skipping: service-serving-cert exists, cannot test self-signed cert generation")
+	}
+
+	c := &ControllerCommandConfig{
+		componentName: "test-controller",
+		basicFlags:    &ControllerFlags{},
+	}
+
+	config := &operatorv1alpha1.GenericOperatorConfig{
+		ServingInfo: configv1.HTTPServingInfo{},
+	}
+	config.ServingInfo.CertFile = ""
+	config.ServingInfo.KeyFile = ""
+
+	_, _, err := c.AddDefaultRotationToConfig(config, []byte{})
+	if err != nil {
+		t.Fatalf("AddDefaultRotationToConfig failed: %v", err)
+	}
+
+	if config.ServingInfo.CertFile == "" {
+		t.Fatal("expected CertFile to be set, got empty string")
+	}
+	if config.ServingInfo.KeyFile == "" {
+		t.Fatal("expected KeyFile to be set, got empty string")
+	}
+
+	certPEM, err := os.ReadFile(config.ServingInfo.CertFile)
+	if err != nil {
+		t.Fatalf("failed to read generated certificate: %v", err)
+	}
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		t.Fatal("failed to parse certificate PEM")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		t.Fatalf("failed to parse certificate: %v", err)
+	}
+
+	actualLifetime := cert.NotAfter.Sub(cert.NotBefore)
+	expectedLifetime := 30 * 24 * time.Hour
+
+	// Allow 1 minute tolerance
+	tolerance := 1 * time.Minute
+	lowerBound := expectedLifetime - tolerance
+	upperBound := expectedLifetime + tolerance
+
+	if actualLifetime < lowerBound || actualLifetime > upperBound {
+		t.Errorf("certificate lifetime is incorrect: expected %v (30 days), got %v", expectedLifetime, actualLifetime)
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+)
+
+func TestAddDefaultRotationToConfig_SelfSignedCertLifetime(t *testing.T) {
```
tchap, line 15: Generally I find it very helpful to add a docstring to each test explaining what is being tested and potentially why we need the test. Just the overall context, you know.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tmpDir := t.TempDir()
+			if err := tt.setup(tmpDir); err != nil {
+				t.Fatalf("setup failed: %v", err)
+			}
+
+			result := hasServiceServingCerts(tmpDir)
```
tchap, line 127: Honestly this function is weird. It returns `true` when the `os.Stat` calls fail. I don't find that acceptable.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+)
+
+func TestAddDefaultRotationToConfig_SelfSignedCertLifetime(t *testing.T) {
+	// Skip if service-serving-cert files exist, as we need to test the self-signed cert path
+	if hasServiceServingCerts("/var/run/secrets/serving-cert") {
+		t.Skip("skipping: service-serving-cert exists, cannot test self-signed cert generation")
+	}
```
tchap, line 19: To avoid this, we should rather create a private method and test that:

```golang
func (c *ControllerCommandConfig) AddDefaultRotationToConfig(config *operatorv1alpha1.GenericOperatorConfig, configContent []byte) (map[string][]byte, []string, error) {
	return c.addDefaultRotationToConfigAtPath("/var/run/secrets/serving-cert", config, configContent)
}

func (c *ControllerCommandConfig) addDefaultRotationToConfigAtPath(certDir string, config *operatorv1alpha1.GenericOperatorConfig, configContent []byte) (map[string][]byte, []string, error) {
...
}
```

Then we can simply use a temp directory.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+	c := &ControllerCommandConfig{
+		componentName: "test-controller",
+		basicFlags:    &ControllerFlags{},
+	}
+
+	config := &operatorv1alpha1.GenericOperatorConfig{
+		ServingInfo: configv1.HTTPServingInfo{},
+	}
+	config.ServingInfo.CertFile = ""
+	config.ServingInfo.KeyFile = ""
```
tchap, line 30: Hasn't this just been initialized to an empty object?

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+}
+
+func TestAddDefaultRotationToConfig(t *testing.T) {
+	tests := []struct {
+		name              string
+		configFile        string
+		existingCertFile  string
+		existingKeyFile   string
+		expectGenerated   bool
```
tchap, line 141: It would be really nice to add docs to the fields so that it's obvious what setting that field does.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+					t.Errorf("expected CertFile to remain %q, got %q", tt.existingCertFile, config.ServingInfo.CertFile)
+				}
+				if config.ServingInfo.KeyFile != tt.existingKeyFile {
+					t.Errorf("expected KeyFile to remain %q, got %q", tt.existingKeyFile, config.ServingInfo.KeyFile)
+				}
+			}
+
+			if len(observedFiles) != tt.expectObservedLen {
+				t.Errorf("expected %d observed files, got %d: %v", tt.expectObservedLen, len(observedFiles), observedFiles)
+			}
```
tchap, line 211: I find this whole test a bit confusing. We are setting `config.ServingInfo.CertFile = tt.existingCertFile` and then we check on return that it equals the same value. Not sure what that is supposed to test. I would be happier if we could clearly state expected inputs and outputs in the table and skip deducing them. These are the params, these are the objects returned. Clear and simple to read and understand. Checking observedFile len is not very telling.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+}
+
+func TestConfig(t *testing.T) {
+	tests := []struct {
+		name           string
+		configContent  string
+		expectError    bool
+		expectNilUnstr bool
+		validateConfig func(*testing.T, *operatorv1alpha1.GenericOperatorConfig)
```
tchap, line 241: Again, pls add comments.

pkg/controller/controllercmd/cmd_test.go
```diff
@@ -0,0 +1,312 @@
...
+				t.Error("expected unstructured to be non-nil")
+			}
+
+			if tt.validateConfig != nil {
+				tt.validateConfig(t, config)
+			}
+
+			if tt.configContent != "" && content == nil {
+				t.Error("expected content to be set when config file exists")
+			}
```
tchap, line 309: Can we have something more declarative rather then this bunch of ifs pls?

PR #2179, opened 2026-04-27
CNTRLPLANE-3237: Encapsulate all kms config under a single field in KeyState

ardaguclu:
This PR is continuation of https://github.com/openshift/library-go/pull/2172 to add https://github.com/openshift/library-go/pull/2163#discussion_r3116059379. 

Thanks to that we'll add all kms specific fields under single KMSConfig field to better manage.

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal KMS configuration state management to improve code organization and maintainability.

PR #2185, opened 2026-04-28
CNTRLPLANE-3237: Encapsulate more functions

ardaguclu:
This PR addresses some changes requested in https://github.com/openshift/library-go/pull/2179#discussion_r3152292197 and https://github.com/openshift/library-go/pull/2179#discussion_r3152329065

## Summary by CodeRabbit

* **Refactor**
  * Standardized nil-safe checks for encryption and KMS configuration across the operator.
  * Made KMS configuration optional-at-read, only materializing it when present.
  * Added small helper predicates to centralize presence checks, improving consistency and reducing duplicate logic.

### DIFF
--- a/pkg/operator/encryption/state/types.go
+++ b/pkg/operator/encryption/state/types.go
@@ -40,8 +40,12 @@ type KeyState struct {
 	InternalReason string
 	// the user via unsupportConfigOverrides.encryption.reason triggered this key.
 	ExternalReason string
-	// stores the all KMS encryption mode related configurations
+	// stores all the KMS encryption mode related configurations
 	KMS *KMSConfig
+}
+
+func (k *KeyState) HasKMSEncryptionConfig() bool {
+	return k != nil && k.KMS != nil && k.KMS.EncryptionConfig != nil
 }
 
 // KMSConfig stores all KMS encryption mode related configurations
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/state/types.go
```diff
@@ -44,6 +44,10 @@ type KeyState struct {
 	KMS *KMSConfig
 }
 
+func (k KeyState) HasKMSEncryptionConfig() bool {
```
p0lyn0mial: should this be a pointer receiver so that
```
var c *KeyState
c.HasKMSEncryptionConfig()
```

won't panic ?

pkg/operator/encryption/state/types.go
```diff
@@ -44,6 +44,10 @@ type KeyState struct {
...
+	return k.KMS != nil && k.KMS.EncryptionConfig != nil
+}
+
 // KMSConfig stores all KMS encryption mode related configurations
 type KMSConfig struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
```
p0lyn0mial, line 53: should we also address https://github.com/openshift/library-go/pull/2179#discussion_r3148218082 ?

pkg/operator/encryption/state/types.go
ardaguclu: Updated

pkg/operator/encryption/state/types.go
ardaguclu: Updated

PR #2186, opened 2026-04-28
CNTRLPLANE-3237: Introduce KMSProviderConfig in KeyState

ardaguclu:
This PR solely focuses on populating the kms-provider-config into Key Secret as well as `KeyState` for the work that is described https://github.com/openshift/enhancements/pull/1960

Currently kms-provider-config is not wired to encryption-config Secret that to be used to carry the data to plugin lifecycle.

## Summary by CodeRabbit

* **New Features**
  * Persist and validate KMS provider configuration inside encryption secrets; added support to read provider config from a dedicated config Secret.
  * Added KMS provider API types and Vault provider support (Transit mount/key, AppRole auth, optional TLS).

* **Tests**
  * Expanded tests to cover KMS provider config serialization, deserialization, presence/absence checks, and roundtrip consistency.


pkg/operator/encryption/state/types.go
```diff
@@ -48,6 +49,27 @@ type KeyState struct {
 type KMSConfig struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
 	EncryptionConfig *apiserverconfigv1.KMSConfiguration
```
p0lyn0mial: we could rename this to `Encryption` - could be a new PR.

pkg/operator/encryption/state/types.go
```diff
@@ -48,6 +49,27 @@ type KeyState struct {
...
+
+	// ProviderConfig stores KMS provider specific configurations
+	ProviderConfig *KMSProviderConfig
```
p0lyn0mial: then this could be `Provider`

pkg/operator/encryption/state/types.go
```diff
@@ -48,6 +49,27 @@ type KeyState struct {
...
+// managed KMS provider specific details such as secret/configmap mappings.
+type KMSProviderConfig struct {
+	// KMSProviderConfig is directly derived from the KMSConfig in o/api.
+	// Therefore, all the KMS provider details in o/api is represented here.
+	*v1.KMSConfig
+
+	// TODO: These fields are mimicked from https://github.com/openshift/api/pull/2805
+	// When API PR merges, *v1.KMSConfig will inherently support store fields,
+	// so that we can remove these internal structs.
+	KMSPluginImage string `json:"kmsPluginImage,omitempty"`
```
p0lyn0mial: I would wait until the API definitions land before introducing these changes.

pkg/operator/encryption/state/types.go
```diff
@@ -48,6 +49,27 @@ type KeyState struct {
 type KMSConfig struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
 	EncryptionConfig *apiserverconfigv1.KMSConfiguration
+
+	// ProviderConfig stores KMS provider specific configurations
+	ProviderConfig *KMSProviderConfig
+}
+
+// KMSProviderConfig stores KMS provider details.
+// This type does not only store the API definitions but also carries internally
```
p0lyn0mial: what other fields will be stored here ?

pkg/operator/encryption/state/types.go
```diff
@@ -48,6 +49,27 @@ type KeyState struct {
...
+// managed KMS provider specific details such as secret/configmap mappings.
+type KMSProviderConfig struct {
+	// KMSProviderConfig is directly derived from the KMSConfig in o/api.
+	// Therefore, all the KMS provider details in o/api is represented here.
+	*v1.KMSConfig
```
p0lyn0mial: how about pulling in the o/api PR in this PR ?

ardaguclu:
/hold
this PR contains fake o/api bump

pkg/operator/encryption/state/types.go
ardaguclu: Done. Second commit uses the o/api PR

pkg/operator/encryption/state/types.go
ardaguclu: Credential Secret/ConfigMap mapping will be stored additionally.

pkg/operator/encryption/state/types.go
ardaguclu: `Provider` is nicer indeed.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -136,6 +147,14 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Data[EncryptionSecretKMSEncryptionConfig] = kmsEncCfgJSON
 	}
 
+	if ks.KMS != nil && ks.KMS.ProviderConfig != nil {
```
p0lyn0mial: `HasKMSProviderConfig` ?

ardaguclu:
/retitle WIP: [CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Introduce KMSProviderConfig in KeyState

ardaguclu:
/retitle [CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Introduce KMSProviderConfig in KeyState

pkg/operator/encryption/secrets/secrets.go
ardaguclu: done

pkg/operator/encryption/api/api.go
```diff
@@ -0,0 +1,284 @@
+// TODO: This API definitions are copied from https://github.com/openshift/api/pull/2805
```
p0lyn0mial: nit: could we create a new `api` pkg for it ? for example `pkg/operator/encryption/api` ?

ardaguclu: It would be indeed better

pkg/operator/encryption/state/tmp.go
```diff
@@ -0,0 +1,284 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="serverName must be a valid DNS hostname: contain no more than 253 characters, contain only lowercase alphanumeric characters, '-' or '.', and start and end with an alphanumeric character"
+	// +optional
+	ServerName string `json:"serverName,omitempty"`
+}
+
+// GetKMSProviderConfig converts the APIServer config into the internal KMSProviderConfig.
+// All downstream controllers operate solely on this internal representation.
+// TODO: these hardcoded data will be replaced by API when it is ready
+func GetKMSProviderConfig() *KMSProviderConfig {
```
p0lyn0mial: wait, didn't we want to reference the "api" type ?

ardaguclu: yes, correct :). I was too quick and it was a mistake

pkg/operator/encryption/api/api.go
ardaguclu: Moved

pkg/operator/encryption/api/api.go
```diff
@@ -0,0 +1,281 @@
...
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="serverName must be a valid DNS hostname: contain no more than 253 characters, contain only lowercase alphanumeric characters, '-' or '.', and start and end with an alphanumeric character"
+	// +optional
+	ServerName string `json:"serverName,omitempty"`
+}
+
+// GetKMSProviderConfig returns KMSConfigAPI that is the representation of API
+// TODO: these hardcoded data will be replaced by API when it is ready
+func GetKMSProviderConfig() *KMSConfigAPI {
```
p0lyn0mial: why do we need this ?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -272,14 +272,15 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 		ExternalReason: externalReason,
 	}
 	if currentMode == state.KMS {
+		kmsProviderConfig, _ := api.GetKMSProviderConfigFromSecret(context.TODO(), c.secretClient)
```
bertinatto: Even though this is temporary, maybe we can log the error to help debugging?

ardaguclu: Good point. Updated.

pkg/operator/encryption/testing/helpers.go
```diff
@@ -106,6 +108,8 @@ func CreateEncryptionKeySecretWithKMSConfig(targetNS string, grs []schema.GroupR
 	}
 	kmsConfigJSON, _ := json.Marshal(kmsConfig)
 	secret.Data[encryptionSecretKMSEncryptionConfigForTest] = kmsConfigJSON
+	providerConfigJSON, _ := json.Marshal(&state.KMSProviderConfig{KMSConfigAPI: encryptionapi.GetKMSProviderConfig()})
```
bertinatto: Same question about logging. Maybe `klog.Errorf`?

pkg/operator/encryption/api/api.go
```diff
@@ -0,0 +1,281 @@
+// TODO: This API definitions are copied from https://github.com/openshift/api/pull/2805
+// This package will be removed when the referenced PR merges.
+package api
+
+// KMSConfig defines the configuration for the KMS instance
+// that will be used with KMS encryption
+// +kubebuilder:validation:XValidation:rule="self.type == 'Vault' ? has(self.vault) : !has(self.vault)",message="vault config is required when kms provider type is Vault, and forbidden otherwise"
+// +union
```
bertinatto: Can we get rid of these annotations?

pkg/operator/encryption/api/api.go
```diff
@@ -0,0 +1,281 @@
...
+	// The secret must contain two keys: "roleID" for the AppRole Role ID and "secretID" for the AppRole Secret ID.
+	//
+	// The namespace for the secret is openshift-config.
+	//
+	// +required
+	Secret VaultSecretReference `json:"secret,omitzero"`
+}
+
+// VaultKMSConfig defines the KMS plugin configuration specific to Vault KMS
+type VaultKMSConfig struct {
```
bertinatto: Would it be possible to copy only this struct? I think we don't need all the rest.

ardaguclu:
/retest
/hold 
https://github.com/openshift/api/pull/2805 is tagged

pkg/operator/encryption/state/types.go
```diff
@@ -48,10 +49,17 @@ func (k *KeyState) HasKMSEncryption() bool {
...
+	return k != nil && k.KMSConfig != nil && k.KMSConfig.Provider != nil
+}
+
 // KMSConfig stores all KMS encryption mode related configurations
 type KMSConfig struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
 	Encryption *apiserverconfigv1.KMSConfiguration
+
+	// Provider stores KMS provider specific configurations
+	Provider *v1.KMSConfig
```
bertinatto, line 62: +1

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -136,6 +148,14 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
...
 	}
 
+	if ks.HasKMSProvider() {
+		providerJSON, err := json.Marshal(ks.KMSConfig.Provider)
+		if err != nil {
+			return nil, err
+		}
+		s.Data[EncryptionSecretKMSProviderConfig] = providerJSON
+	}
+
```
bertinatto, line 161: I believe I may have this somewhere else, but do we need to check for a state where the keyState has the `Encryption` field set, but not the `Provider` field? It doesn't seem we need to?

ardaguclu: Normally this can't happen but there is a possibility that Secrets are manually updated. If that happens, instead of leaving the cluster in unpredictable state, I think it is better to catch this and error out to degrade the cluster.

So in my opinion we should check that Encryption and Provider are not nil, when encryption mode is KMS

ardaguclu: I've updated the code to return error.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -278,6 +279,7 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 				Endpoint:   fmt.Sprintf(kmsEndpointFormat, keyID),
 				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
 			},
+			Provider: &v1.KMSConfig{},
```
ardaguclu: This will be carried from APIServer config when https://github.com/openshift/api/pull/2805 merges.

p0lyn0mial: shouldn't we store the config from the api here ?

ardaguclu: yes, see https://github.com/openshift/library-go/pull/2186#discussion_r3159221488. I can update now, if you are fine with that

ardaguclu: In any case, we'll wait for the merge of API PR.

p0lyn0mial: what happens in the future when the API is changed and we have code that updates the content of the secret ? especially what happens during an upgrade ?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -76,6 +77,17 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 			// encryption mode.
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig)
 		}
+		if v, ok := s.Data[EncryptionSecretKMSProviderConfig]; ok && len(v) > 0 {
+			providerConfig := &v1.KMSConfig{}
+			if err := json.Unmarshal(v, providerConfig); err != nil {
+				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSProviderConfig, err)
+			}
+			key.KMSConfig.Provider = providerConfig
```
p0lyn0mial, line 84: can `KMSConfig` be `nil` ?

ardaguclu: No, we populate it here https://github.com/openshift/library-go/blob/0722f82a14b720ed875880fba8f91b9be5e34287/pkg/operator/encryption/secrets/secrets.go#L71-L73, otherwise we return error

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Although there are many variations of this behavior, I think this won't be an issue. 

For instance, cluster is upgraded to new version with new set of API definitions. Key controller will carry new API definitions. Plugin lifecycle already knows how to use them for side-car container.

p0lyn0mial: do we have to update the client as well ? otherwise we won't be able to read the new api, right ?

ardaguclu: Yes, that is correct. We have to update clients.

p0lyn0mial: Do we have to do it before we merge this PR ?

pkg/operator/encryption/secrets/secrets.go
p0lyn0mial: nit: could we add a small comment explaining why it cannot be nil?

p0lyn0mial: nit: could we change 

```
 key.KMSConfig = &state.KMSConfig{ 
 	Encryption: kmsConfiguration, 
 } 
 ```
  
 to a one-liner ?

File: pkg/operator/encryption/controllers/key_controller.go (hunk 16 above)
Comment (p0lyn0mial) [inline pkg/operator/encryption/controllers/key_controller.go]: should we modify `getCurrentModeAndExternalReason` to return api configuration ? then we could pass the api configuration directly to this function.

in the future we will use the api configuration to update the key secret.

Comment (p0lyn0mial) [reply pkg/operator/encryption/controllers/key_controller.go]: i think that needsNewKey could also use the api config in the future

Comment (ardaguclu) [reply pkg/operator/encryption/controllers/key_controller.go]: We have to merge this PR first, so that client (i.e. plugin lifecycle) can get the data to use.

Comment (ardaguclu) [reply pkg/operator/encryption/controllers/key_controller.go]: yes, these are all correct. With respect to https://github.com/openshift/library-go/pull/2186#discussion_r3159329515, `getCurrentModeAndExternalReason` should return api configuration. I can do it now, or after the API PR merges.

Comment (ardaguclu) [reply pkg/operator/encryption/controllers/key_controller.go]: > i think that needsNewKey could also use the api config in the future

Yes, in order to detect kms -> kms migration.

Comment (ardaguclu) [reply pkg/operator/encryption/controllers/key_controller.go]: I've updated the code by getting data from `getCurrentModeAndExternalReason`. Please let me know your thoughts.

File: pkg/operator/encryption/secrets/secrets.go (hunk 17 above)
Comment (ardaguclu) [reply pkg/operator/encryption/secrets/secrets.go:84]: Updated.

File: pkg/operator/encryption/controllers/key_controller.go (hunk 16 above)
Comment (ardaguclu) [reply pkg/operator/encryption/controllers/key_controller.go]: Updated the code. Please let me know your thoughts.

**Comment by ardaguclu** [comment]
Failure https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_library-go/2186/pull-ci-openshift-library-go-master-e2e-aws-encryption/2049436822190166016 is expected because now we require non-nil KMSConfig. This should be fixed when API PR merges.

File: pkg/operator/encryption/controllers/key_controller.go [hunk 18]
```diff
@@ -271,42 +272,47 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
 				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
 			},
+			Provider: apiServerEncryption.KMS,
 		}
 	}
 	return secrets.FromKeyState(c.instanceName, ks)
 }
 
-func (c *keyController) getCurrentModeAndExternalReason(ctx context.Context) (state.Mode, string, error) {
+func (c *keyController) getCurrentModeAndExternalReason(ctx context.Context) (state.Mode, string, *v1.APIServerEncryption, error) {
```
p0lyn0mial: `getCurrentModeReasonAndEncryptionConfig` ?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -271,42 +272,47 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+		return "", "", nil, err
 	}
 
 	encryptionConfig, err := structuredUnsupportedConfigFrom(operatorSpec.UnsupportedConfigOverrides.Raw, c.unsupportedConfigPrefix)
 	if err != nil {
-		return "", "", err
+		return "", "", nil, err
 	}
 
+	encryption := &apiServer.Spec.Encryption
```
p0lyn0mial: do we have to return a pointer ?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -271,42 +272,47 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
 	reason := encryptionConfig.Encryption.Reason
-	switch currentMode := state.Mode(apiServer.Spec.Encryption.Type); currentMode {
+	switch currentMode := state.Mode(encryption.Type); currentMode {
 	case state.AESCBC, state.AESGCM, state.KMS, state.Identity: // secretbox is disabled for now
```
p0lyn0mial: could we check encryption.KMS for nil here ?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -271,42 +272,47 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 		ExternalReason: externalReason,
 	}
 	if currentMode == state.KMS {
+		if apiServerEncryption == nil || apiServerEncryption.KMS == nil {
```
p0lyn0mial: then we could drop this checks.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -163,7 +164,7 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
 }
 
 func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext factory.SyncContext, encryptedGRs []schema.GroupResource) error {
-	currentMode, externalReason, err := c.getCurrentModeAndExternalReason(ctx)
+	currentMode, externalReason, apiServerEncryption, err := c.getCurrentModeAndExternalReason(ctx)
```
p0lyn0mial: `encryptionConfiguration` ?

p0lyn0mial: `apiEncryptionConfiguration` ?

ardaguclu: Updated for apiEncryptionConfiguration

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Dropped

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Yes, better. Updated

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: No we don't have to. Updated.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -68,14 +69,24 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
...
 			// encryption.apiserver.operator.openshift.io-kms-encryption-config data field is required for KMS
 			// encryption mode.
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig)
 		}
+		if v, ok := s.Data[EncryptionSecretKMSProviderConfig]; ok && len(v) > 0 {
+			providerConfig := &v1.KMSConfig{}
+			if err := json.Unmarshal(v, providerConfig); err != nil {
+				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSProviderConfig, err)
+			}
+			// KMSConfig is non-nil here; initialized by the encryption config branch above
```
bertinatto: If the secret has a `EncryptionSecretKMSProviderConfig` key, but lacks a `EncryptionSecretKMSEncryptionConfig` key, then `KMSConfig` could be nil (and we'd panic here).

Probably not something that can happen, but maybe we can be defensive about it and initialize `key.KMSConfig` before the if-else clauses - and add a comment that the allocation would be wasted if the secret doesn't have the KMS keys (which is fine).

bertinatto: Something like:

```diff
diff --git a/pkg/operator/encryption/secrets/secrets.go b/pkg/operator/encryption/secrets/secrets.go
index f16198107..aa0fa417d 100644
--- a/pkg/operator/encryption/secrets/secrets.go
+++ b/pkg/operator/encryption/secrets/secrets.go
@@ -64,12 +64,14 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
        case state.AESCBC, state.AESGCM, state.SecretBox, state.Identity:
                key.Mode = keyMode
        case state.KMS:
+               // This allocation will be wasted if KMS keys are not present in the secret, but that should not happen
+               key.KMSConfig = &state.KMSConfig{}
                if v, ok := s.Data[EncryptionSecretKMSEncryptionConfig]; ok && len(v) > 0 {
                        kmsConfiguration := &apiserverconfigv1.KMSConfiguration{}
                        if err := json.Unmarshal(v, kmsConfiguration); err != nil {
                                return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSEncryptionConfig, err)
                        }
-                       key.KMSConfig = &state.KMSConfig{Encryption: kmsConfiguration}
+                       key.KMSConfig.Encryption = kmsConfiguration
                } else {
                        // encryption.apiserver.operator.openshift.io-kms-encryption-config data field is required for KMS
                        // encryption mode.
@@ -80,7 +82,6 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
                        if err := json.Unmarshal(v, providerConfig); err != nil {
                                return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSProviderConfig, err)
                        }
-                       // KMSConfig is non-nil here; initialized by the encryption config branch above
                        key.KMSConfig.Provider = providerConfig
                } else {
                        // encryption.apiserver.operator.openshift.io-kms-provider-config data field is required for KMS
```

ardaguclu: That would be better. I'll make changes after https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_library-go/2186/pull-ci-openshift-library-go-master-e2e-aws-encryption/2049479863575777280 finishes

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -136,6 +151,14 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Data[EncryptionSecretKMSEncryptionConfig] = kmsEncCfgJSON
 	}
 
+	if ks.HasKMSProvider() {
+		providerJSON, err := json.Marshal(ks.KMSConfig.Provider)
+		if err != nil {
+			return nil, err
```
bertinatto: nit: we could annotate the error here and on the check above (for the EncryptionConfiguration) to say what exactly failed to be serialized

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -128,6 +139,10 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Annotations[EncryptionSecretMigratedResources] = string(bs)
 	}
 
+	if ks.Mode == state.KMS && (!ks.HasKMSEncryption() || !ks.HasKMSProvider()) {
+		return nil, fmt.Errorf("%s or %s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig, EncryptionSecretKMSProviderConfig)
+	}
```
bertinatto: nit: we could move this up and avoid some processing if we're in a bad state

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -278,35 +279,42 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
 				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
 			},
+			Provider: apiServerEncryption.KMS,
 		}
 	}
 	return secrets.FromKeyState(c.instanceName, ks)
 }
 
-func (c *keyController) getCurrentModeAndExternalReason(ctx context.Context) (state.Mode, string, error) {
+func (c *keyController) getCurrentModeReasonAndEncryptionConfig(ctx context.Context) (state.Mode, string, v1.APIServerEncryption, error) {
```
bertinatto, line 288: do we have a reason for this not to be a pointer?

ardaguclu: It was. I updated after https://github.com/openshift/library-go/pull/2186#discussion_r3161025817

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Updated. Please let me know about the state.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Updated.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Updated.

bertinatto (commented):
/lgtm

ardaguclu:
It is better to wait for the result of https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2130/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-gcp-operator-encryption-kms/2049478757034496000
/hold

bertinatto (commented):
/lgtm

ardaguclu:
It failed but this requires changes in test step. I will fix it in a followup PR
/hold cancel

PR #2187, opened 2026-04-28
CNTRLPLANE-3237: Rename to Encryption by dropping Config suffix

ardaguclu:
This PR addresses the change requested in https://github.com/openshift/library-go/pull/2186#discussion_r3152782018

## Summary by CodeRabbit

* **Refactor**
  * Internal KMS encryption configuration was renamed and reorganized for consistency; behavior unchanged for users.
  * Secret serialization and state conversions for KMS-backed keys were updated to match the new structure.
* **Tests**
  * Test fixtures updated to reflect the configuration rename; no functional changes expected in production.

### DIFF
--- a/pkg/operator/encryption/encryptiondata/config.go
+++ b/pkg/operator/encryption/encryptiondata/config.go
@@ -219,12 +219,12 @@ func stateToProviders(resource string, desired state.GroupResourceState) []apise
 				},
 			})
 		case state.KMS:
-			if key.KMS == nil || key.KMS.EncryptionConfig == nil {
-				klog.Infof("skipping key %s for %s in KMS mode as its KMS.EncryptionConfig is nil", key.Key.Name, resource)
+			if !key.HasKMSEncryption() {
+				klog.Infof("skipping key %s for %s in KMS mode as its either KMSConfig or KMSConfig.Encryption is nil", key.Key.Name, resource)
 				continue // this should never happen
 			}
 			// In order to preserve the uniqueness, we should insert resource name
-			kmsCopy := key.KMS.EncryptionConfig.DeepCopy()
+			kmsCopy := key.KMSConfig.Encryption.DeepCopy()
 			kmsCopy.Name = createKMSProviderName(key.Key.Name, resource)
 			provider := apiserverconfigv1.ProviderConfiguration{
 				KMS: kmsCopy,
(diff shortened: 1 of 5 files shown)


pkg/operator/encryption/state/types.go
```diff
@@ -44,14 +44,14 @@ type KeyState struct {
 	KMS *KMSConfig
```
p0lyn0mial: would it make sense to rename KMS to KMSConfig ?

p0lyn0mial: then it would be referenced as `ks.KMSConfig.Encryption`

pkg/operator/encryption/state/types.go
```diff
@@ -44,14 +44,14 @@ type KeyState struct {
...
 }
 
-func (k *KeyState) HasKMSEncryptionConfig() bool {
-	return k != nil && k.KMS != nil && k.KMS.EncryptionConfig != nil
+func (k *KeyState) HasKMSEncryption() bool {
+	return k != nil && k.KMS != nil && k.KMS.Encryption != nil
```
p0lyn0mial: do we need to check `k != nil` ?

ardaguclu: Just to be defensive, it is impossible that k can be nil. Would you want me to remove that?

p0lyn0mial: ok, this is what we did for config as well, https://github.com/openshift/library-go/pull/2185/changes

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -219,12 +219,12 @@ func stateToProviders(resource string, desired state.GroupResourceState) []apise
 				},
 			})
 		case state.KMS:
-			if key.KMS == nil || key.KMS.EncryptionConfig == nil {
-				klog.Infof("skipping key %s for %s in KMS mode as its KMS.EncryptionConfig is nil", key.Key.Name, resource)
+			if !key.HasKMSEncryption() {
+				klog.Infof("skipping key %s for %s in KMS mode as its KMSConfig.Encryption is nil", key.Key.Name, resource)
```
bertinatto: This predates your PR, but the log message could be wrong. We don't really know if `KMSConfig.Encryption` is nil, right? It could be `key`, `key.KMSConfig`, or `key.KMSConfig.Encryption`

bertinatto (commented):
Just a small nit, otherwise LGTM

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: I've updated the log message. Please let me know your thoughts.

bertinatto (commented):
/lgtm

PR #2191, opened 2026-04-29
NO-JIRA: Set required KMSConfig in e2e tests for KMS mode

ardaguclu:
After the API changes have merged for KMS, KMSConfig can not be nil when encryption mode is set to KMS. This PR fixes this to enable e2e tests again.

Currently apart from the Image field (which points to mock KMS plugin), all the fields are dummy. But we'll have to update them with the correct values, once we onboard Vault enterprise in our CI. 

## Summary by CodeRabbit

* **Tests**
  * Improved encryption test infrastructure to better simulate KMS configurations across encryption modes.

---

Note: Internal testing improvements only; no user-facing changes.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -24,6 +24,7 @@ import (
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 
 	"github.com/openshift/library-go/test/library"
+	"github.com/openshift/library-go/test/library/encryption/kms"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


test/library/encryption/helpers.go
```diff
@@ -66,6 +67,11 @@ func SetAndWaitForEncryptionType(t testing.TB, encryptionType configv1.Encryptio
 	if needsUpdate {
 		t.Logf("Updating encryption type in the config file for APIServer to %q", encryptionType)
 		apiServer.Spec.Encryption.Type = encryptionType
+		if encryptionType == configv1.EncryptionTypeKMS {
+			apiServer.Spec.Encryption.KMS = defaultTestKMSConfig()
+		} else {
+			apiServer.Spec.Encryption.KMS = nil
+		}
```
bertinatto: Would it make sense for `defaultTestKMSConfig()` to accept `encryptionType` and return the correct `KMSConfig` according to that param? In practice this if-else clause would move inside `defaultTestKMSConfig()`

ardaguclu: Updated.

bertinatto (commented):
/lgtm

PR #2193, opened 2026-04-30
CNTRLPLANE-3234: health reporter

ibihim:
## What

Create package with cobra command to run a health monitor for KMS plugin.

## Why

We need some introspection into the state of the KMS plugin.
Preparing to create proper signaling for users.

## Notes

```bash                                                                                                                
kubectl --context kind-kms-health-claude -n kms-health-test \
  get cm kms-health-status -o json | jq '.data | map_values(fromjson)'
```

Output:

```json                                                                                                                               
{                    
  "not-ok": {
    "timestamp": "2026-04-29T21:48:46Z",                                                                                                   
    "observerPod": "kms-health-monitor-55789c9cc6-62vw7",
    "detail": "unhealthy:test-forced",                                                                                                     
    "keyIDHash": "127fb4517dfd...77458d3"                                                                                                  
  },                 
  "ok": {                                                                                                                                  
    "timestamp": "2026-04-29T22:01:43Z",                                                                                                 
    "observerPod": "kms-health-monitor-55789c9cc6-62vw7",                                                                                  
    "keyIDHash": "127fb4517dfd...77458d3"
  },                                                                                                                                       
  "rpc-error": {                                                                                                                         
    "timestamp": "2026-04-29T21:49:04Z",                                                                                                   
    "observerPod": "kms-health-monitor-55789c9cc6-62vw7",                                                                                
    "detail": "DeadlineExceeded"                                                                                                           
  }
}                                
```

## Summary by CodeRabbit

* **New Features**
  * New KMS health monitor CLI: validates flags, runs periodic (jittered) plugin probes, and reports aggregated KMS plugin health as operator status conditions in the cluster.
  * Supports selecting target operator and explicit kubeconfig for reporting.

* **Tests**
  * Added unit tests for the health checker and the status writer, covering healthy, unhealthy, and error scenarios.


pkg/operator/encryption/kms/health/monitor.go
```diff
@@ -0,0 +1,78 @@
...
+// Run blocks until ctx is cancelled.
+func (m *Monitor) Run(ctx context.Context) {
+	for {
+		status := m.probe.Probe(ctx)
+		status.ObserverPod = m.observerPod
+
+		// Cancel inline rather than via defer: defers in unbounded loops
+		// accumulate until the function returns.
+		writeCtx, writeCancel := context.WithTimeout(ctx, m.writeTimeout)
+		err := m.writer.Write(writeCtx, status)
```
ibihim: TODO:
- Compare hysteresis like pkg/monitor/health?
- (nit) IIFE better than non-defer cancel?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,238 @@
+package health
```
ardaguclu, line 1: What this package is supposed to do is very similar to https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/kms/preflight/checker.go. I think it is better to mimic from it. 

The only difference is health monitor will periodically write the status to somewhere.

ardaguclu (commented):
I know it is not ready for review yet. Nevertheless I dropped a comment that can be useful. Also design needs to go here https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/kms-encryption-foundations.md

pkg/operator/encryption/kms/health/cmd.go
ibihim: What are the differences? [Both](https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/kms/preflight/cmd.go) are cobra commands. They have flags, validations and a run method.

I create a reader, a writer and put them into a monitor. Writer is an interface, because I want to pivot to operator conditions.

p0lyn0mial: yeah, it seems there is room for reuse. it would be better to have a single binary with custom logic.

pkg/operator/encryption/kms/health/writer.go
```diff
@@ -0,0 +1,61 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/utils/clock"
+)
+
+type Writer struct {
+	operatorClient v1helpers.OperatorClientWithFinalizers
+	nodeName       string
+}
+
+func (w *Writer) Apply(ctx context.Context, conditions []PluginHealthCondition) error {
```
ardaguclu, line 22: As far as I can see, `writer.go` aims to update the OperatorCondition. I guess there should be a built-in functionality for this behavior in library-go already?.

ibihim: Would make sense. Let me try to find it.

ibihim: Good instinct, and you're right that there's a built-in for the simple case: `v1helpers.UpdateStatus(ctx, client, v1helpers.UpdateConditionFn(cond))` collapses this into one call, and `SetOperatorCondition` even handles the lastTransitionTime-only-on-change bit. That's the RMW version of what I'm doing by hand.

Here is why the current solution is still the way to go:

1. **The setup:** one monitor runs per node, but there is only one cluster-scoped operator CR. So every node writes its own condition into the same `status.conditions` array.

```
            one operator CR
     status.conditions: [ _a, _b, _c ]
            ▲       ▲       ▲
         node-a  node-b  node-c
```

2. **Why RMW gets chatty here:** `UpdateStatus` reads the whole object, edits the array, writes it back. It's only ~3 control-plane writers on a 30s interval, so `RetryOnConflict` would survive it fine. But they'd keep invalidating each other's reads for no real reason, and it's avoidable.

3. **Why SSA is cleaner:** with a per-node field manager, each node owns only its own entry (`conditions` is `listType=map` keyed by `type`). The server merges the rest, so node-a writing never conflicts with node-b. No read, no 409s between nodes.

So it is the same Apply vs Update trade-off as everywhere else: field ownership versus whole-object optimistic concurrency.

Separately, you did catch a real thing: the manual `WithLastTransitionTime(metav1.Now())` is redundant. `ApplyOperatorStatus` already canonicalizes it via `SetApplyConditionsLastTransitionTime`, so I will drop that line.

pkg/operator/encryption/kms/health/checker.go
```diff
@@ -0,0 +1,86 @@
...
+	keyID   string
+	service kmsservice.Service
+}
+
+type Checker struct {
+	plugins []plugin
+	now     func() time.Time
+}
+
+func NewChecker(ctx context.Context, udsPaths []string, timeout time.Duration) (*Checker, error) {
```
ardaguclu, line 24: I've implemented an alternative https://github.com/openshift/library-go/pull/2271 to suggest the changes. Please let me know your thoughts.

ibihim: Oh wow, now I am a reviewer.

ardaguclu: That is just to give you an idea. You shouldn't review that one, but there are some bits to show you how we can reuse what we have in library-go.

PR #2194, opened 2026-04-30
CNTRLPLANE-3237: use runtime.Encode/Decode with registered types

bertinatto:
Encode `apiserverconfigv1.KMSConfiguration` inside an `apiserverconfigv1.EncryptionConfiguration` envelope and `configv1.KMSConfig` inside an `configv1.APIServer` envelope.

 This allows for proper use of `runtime.{Encode|Decode}` with registered types instead of raw JSON.

Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2133

```
$ oc get -n openshift-config-managed secrets/encryption-key-openshift-kube-apiserver-1 -o json | jq -r '.data."encryption.apiserver.operator.openshift.io-kms-encryption-config"' | base64 -d | jq
{
  "kind": "EncryptionConfiguration",
  "apiVersion": "apiserver.config.k8s.io/v1",
  "resources": [
    {
      "resources": null,
      "providers": [
        {
          "kms": {
            "apiVersion": "v2",
            "name": "1",
            "endpoint": "unix:///var/run/kmsplugin/kms-1.sock",
            "timeout": "10s"
          }
        }
      ]
    }
  ]
}

$ oc get -n openshift-config-managed secrets/encryption-key-openshift-kube-apiserver-1 -o json | jq -r '.data."encryption.apiserver.operator.openshift.io-kms-provider-config"' | base64 -d | jq
{
  "kind": "APIServer",
  "apiVersion": "config.openshift.io/v1",
  "metadata": {},
  "spec": {
    "servingCerts": {},
    "clientCA": {
      "name": ""
    },
    "encryption": {
      "kms": {
        "type": "Vault",
        "vault": {
          "kmsPluginImage": "quay.io/bertinatto/vault@sha256:33599dd6eee61dcf9a60138759fafda3d88593a3c2072585156882c6b5bd3fa5",
          "vaultAddress": "https://vault.example.com:8200",
          "authentication": {
            "type": "AppRole",
            "appRole": {
              "secret": {
                "name": "vault-approle-secret"
              }
            }
          },
          "transitKey": "my-transit-key"
        }
      }
    },
    "audit": {}
  },
  "status": {}
}
```
/assign @ardaguclu @p0lyn0mial @benluddy 

## Summary by CodeRabbit

* **Refactor**
  * KMS/encryption serialization switched to new encoding helpers for more robust, API-compliant handling and clearer encode/decode error messages.

* **Tests**
  * Tests updated to use the new encoding flow, include explicit KMS timeouts, and fail loudly on encoding errors to surface issues earlier.

### DIFF
--- a/pkg/operator/encryption/secrets/secrets.go
+++ b/pkg/operator/encryption/secrets/secrets.go
@@ -8,13 +8,13 @@ import (
 	"strconv"
 	"time"
 
-	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
+	"github.com/openshift/library-go/pkg/operator/encryption/encoding"
 	"github.com/openshift/library-go/pkg/operator/encryption/state"
 )
 
@@ -66,8 +66,8 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 	case state.KMS:
 		key.KMSConfig = &state.KMSConfig{}
 		if v, ok := s.Data[EncryptionSecretKMSEncryptionConfig]; ok && len(v) > 0 {
-			kmsConfiguration := &apiserverconfigv1.KMSConfiguration{}
-			if err := json.Unmarshal(v, kmsConfiguration); err != nil {
+			kmsConfiguration, err := encoding.DecodeKMSConfiguration(v)
+			if err != nil {
 				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSEncryptionConfig, err)
 			}
 			key.KMSConfig.Encryption = kmsConfiguration
@@ -77,11 +77,11 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig)
 		}
 		if v, ok := s.Data[EncryptionSecretKMSProviderConfig]; ok && len(v) > 0 {
-			providerConfig := &configv1.KMSConfig{}
-			if err := json.Unmarshal(v, providerConfig); err != nil {
+			kmsConfig, err := encoding.DecodeKMSConfig(v)
+			if err != nil {
 				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSProviderConfig, err)
 			}
-			key.KMSConfig.Provider = providerConfig
+			key.KMSConfig.Provider = kmsConfig
 		} else {
 			// encryption.apiserver.operator.openshift.io-kms-provider-config data field is required for KMS
 			// encryption mode.
@@ -144,19 +144,19 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 	}
 
 	if ks.HasKMSEncryption() {
-		kmsEncCfgJSON, err := json.Marshal(ks.KMSConfig.Encryption)
+		encryptionConfigurationData, err := encoding.EncodeKMSConfiguration(ks.KMSConfig.Encryption)
 		if err != nil {
-			return nil, fmt.Errorf("failed to marshal KMS encryption config: %w", err)
+			return nil, err
 		}
-		s.Data[EncryptionSecretKMSEncryptionConfig] = kmsEncCfgJSON
+		s.Data[EncryptionSecretKMSEncryptionConfig] = encryptionConfigurationData
 	}
 
 	if ks.HasKMSProvider() {
-		providerJSON, err := json.Marshal(ks.KMSConfig.Provider)
+		providerData, err := encoding.EncodeKMSConfig(ks.KMSConfig.Provider)
 		if err != nil {
-			return nil, fmt.Errorf("failed to marshal KMS provider config: %w", err)
+			return nil, err
 		}
-		s.Data[EncryptionSecretKMSProviderConfig] = providerJSON
+		s.Data[EncryptionSecretKMSProviderConfig] = providerData
 	}
 
 	return s, nil
--- a/pkg/operator/encryption/secrets/secrets.go
+++ b/pkg/operator/encryption/secrets/secrets.go
@@ -8,7 +8,7 @@ import (
 	"strconv"
 	"time"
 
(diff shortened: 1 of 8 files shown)


pkg/operator/encryption/secrets/secrets.go
```diff
@@ -184,3 +197,63 @@ func ListKeySecrets(ctx context.Context, secretClient corev1client.SecretsGetter
 	}
 	return encryptionSecrets, nil
 }
+
+func DecodeKMSConfiguration(data []byte) (*apiserverconfigv1.KMSConfiguration, error) {
+	// KMSConfiguration is not a runtime.Object, so use EncryptionConfiguration as envolope
+	encryptionConfiguration := &apiserverconfigv1.EncryptionConfiguration{}
```
ardaguclu: Can we introduce a new generic function similar to https://github.com/openshift/library-go/blob/d6efc4ecf8552e97236e262cc423e90d4c30d769/pkg/operator/encryption/encryptiondata/secret.go#L37-L46 for decoding. So that FromSecret https://github.com/openshift/library-go/blob/d6efc4ecf8552e97236e262cc423e90d4c30d769/pkg/operator/encryption/encryptiondata/secret.go#L31 can also reuse it. 

DecodeKMSConfiguration can return KMSConfiguration after calling that function. WDYT?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -184,3 +197,63 @@ func ListKeySecrets(ctx context.Context, secretClient corev1client.SecretsGetter
 	}
 	return encryptionSecrets, nil
 }
+
+func DecodeKMSConfiguration(data []byte) (*apiserverconfigv1.KMSConfiguration, error) {
```
ardaguclu: nit: Documentation

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -184,3 +197,63 @@ func ListKeySecrets(ctx context.Context, secretClient corev1client.SecretsGetter
...
+		return nil, err
+	}
+	// This should never happen, unless the secret was incorrectly written in FromKeyState()
+	if len(encryptionConfiguration.Resources) != 1 || len(encryptionConfiguration.Resources[0].Providers) != 1 {
+		return nil, fmt.Errorf("invalid KMS encryption config")
+	}
+	return encryptionConfiguration.Resources[0].Providers[0].KMS, nil
+}
+
+func DecodeKMSConfig(data []byte) (*configv1.KMSConfig, error) {
```
ardaguclu: nit: Documentation

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -184,3 +197,63 @@ func ListKeySecrets(ctx context.Context, secretClient corev1client.SecretsGetter
...
+	encryptionConfiguration := &apiserverconfigv1.EncryptionConfiguration{
+		Resources: []apiserverconfigv1.ResourceConfiguration{
+			{
+				Providers: []apiserverconfigv1.ProviderConfiguration{
+					{KMS: encryption},
+				},
+			},
+		},
+	}
+	encoder := codecs.LegacyCodec(apiserverconfigv1.SchemeGroupVersion)
```
ardaguclu: Can we move https://github.com/openshift/library-go/blob/d6efc4ecf8552e97236e262cc423e90d4c30d769/pkg/operator/encryption/encryptiondata/secret.go#L55-L59 to a separate generic function. So that `ToSecret` and this function can reuse it.

ardaguclu (commented):
Mechanism looks reasonable to me. Just dropped a few comments.

pkg/operator/encryption/secrets/secrets.go
p0lyn0mial: +1

we have existing code at:

1. https://github.com/openshift/library-go/blob/d11d3518a3cd31d846000bdd96beabe1c530e816/pkg/operator/encryption/encryptiondata/secret.go#L37
2. https://github.com/openshift/library-go/blob/d11d3518a3cd31d846000bdd96beabe1c530e816/pkg/operator/encryption/encryptiondata/secret.go#L55

how about we add a new pkg say `pkg/operator/encryption/codec` which would contain single shared scheme, codecs, init() and four Decode/Encode functions.

then FromSecret/ToSecret could call codec.EncodeEncryptionConfiguration / codec.DecodeEncryptionConfiguration

pkg/operator/encryption/secrets/secrets.go
p0lyn0mial: +1

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -184,3 +197,63 @@ func ListKeySecrets(ctx context.Context, secretClient corev1client.SecretsGetter
...
+}
+
+func DecodeKMSConfig(data []byte) (*configv1.KMSConfig, error) {
+	// KMSConfig is not a runtime.Object, so use APIServer as envolope
+	apiServer := &configv1.APIServer{}
+	err := runtime.DecodeInto(codecs.UniversalDecoder(configv1.SchemeGroupVersion), data, apiServer)
+	if err != nil {
+		return nil, err
+	}
+	return apiServer.Spec.Encryption.KMS, nil
```
p0lyn0mial: can KMS (*KMSConfig) be nil ? if yes then we can return nil, nil from this function.

pkg/operator/encryption/secrets/secrets.go
p0lyn0mial: maybe that new pkg could be even internal.

p0lyn0mial: experimented with this idea a bit at https://github.com/p0lyn0mial/library-go/commits/kms-k8s-codecs-key-secret-rev/

bertinatto: moved all these new functions to a new package (I called it `encoding` instead of `codecs` because I like it more, but I can change if you guys prefer). Also, I added an extra commit that migrates the encryption-secret functions (`ToSecret` and `FromSecret`) to use the new package

pkg/operator/encryption/secrets/secrets.go
bertinatto: Added docs in all funcs

pkg/operator/encryption/secrets/secrets.go
bertinatto: Added

pkg/operator/encryption/secrets/secrets.go
bertinatto: Done

pkg/operator/encryption/secrets/secrets.go
bertinatto: No, it cannot be nil:
```
func (k *KeyState) HasKMSEncryption() bool {
	return k != nil && k.KMSConfig != nil && k.KMSConfig.Encryption != nil
}
```

But since we moved these to a new package that could be used outside KMS, I added nil checks in the Encoding functions to be on the safe side.

ardaguclu:
/lgtm
/hold
For e2e-aws-encryption

PR #2197, opened 2026-05-05
controller/workload: Clean up status reporting helpers and test infrastructure

tchap:
This is being split from https://github.com/openshift/library-go/pull/2128

- Replace inline error message loops with `errors.Join`
- Use `ptr.Deref` for replicas
- Fix `areCondidtionsEqual` typo
- Remove unnecessary Template/Labels from test fixtures.

## Summary by CodeRabbit

* **Refactor**
  * Improved error aggregation and default replica handling in the workload controller, yielding clearer operator status messages and more reliable default scaling behavior.

* **Tests**
  * Strengthened tests and simplified fixtures, fixing expectations and adding status comparison helpers to validate status updates more reliably.

### DIFF
--- a/pkg/operator/apiserver/controller/workload/workload.go
+++ b/pkg/operator/apiserver/controller/workload/workload.go
@@ -16,6 +16,7 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/ptr"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
@@ -218,8 +219,8 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 
 	if !preconditionsReady {
 		var message string
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
+		if err := errors.Join(errs...); err != nil {
+			message = err.Error()
 		}
 		if len(message) == 0 {
 			message = "the operator didn't specify what preconditions are missing"
@@ -249,9 +250,9 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 	}
 
 	if len(errs) > 0 {
-		message := ""
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
+		var message string
+		if err := errors.Join(errs...); err != nil {
+			message = err.Error()
 		}
 		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
@@ -298,10 +299,7 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
 			WithReason("AsExpected")
 	}
 
-	desiredReplicas := int32(1)
-	if workload.Spec.Replicas != nil {
-		desiredReplicas = *(workload.Spec.Replicas)
-	}
+	desiredReplicas := ptr.Deref(workload.Spec.Replicas, 1)
 
 	// If the workload is up to date, then we are no longer progressing
 	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
--- a/pkg/operator/apiserver/controller/workload/workload_test.go
+++ b/pkg/operator/apiserver/controller/workload/workload_test.go
@@ -95,7 +95,7 @@ func TestUpdateOperatorStatus(t *testing.T) {
 						Message: "deployment/: could not be retrieved",
 					},
 				}
-				return areCondidtionsEqual(expectedConditions, actualStatus.Conditions)
+				return areConditionsEqual(expectedConditions, actualStatus.Conditions)
 			},
 		},
 		{
@@ -112,7 +112,7 @@ func TestUpdateOperatorStatus(t *testing.T) {
 					{
 						Type:    fmt.Sprintf("%sWorkloadDegraded", defaultControllerName),
 						Status:  operatorv1.ConditionTrue,
-						Message: "nasty error\n",
+						Message: "nasty error",
 						Reason:  "SyncError",
 					},
 					{
@@ -129,7 +129,7 @@ func TestUpdateOperatorStatus(t *testing.T) {
 						Message: "deployment/: could not be retrieved",
 					},
 				}
-				return areCondidtionsEqual(expectedConditions, actualStatus.Conditions)
+				return areConditionsEqual(expectedConditions, actualStatus.Conditions)
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -386,6 +377,17 @@ func hasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
...
 }
 
+func errMessage(errs []error) string {
+	var b strings.Builder
+	for i, err := range errs {
+		if i > 0 {
+			b.WriteString("\n")
+		}
+		b.WriteString(err.Error())
+	}
```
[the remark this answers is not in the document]
tchap: This can't really happen and if it does, it's an issue that should break. Skipping.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -756,6 +765,37 @@ func (f *fakePodLister) Pods(namespace string) corev1listers.PodNamespaceLister
...
+
+func (f *fakeVersionRecorder) SetVersion(operandName, version string) {
+	f.setVersionCalls = append(f.setVersionCalls, setVersionCall{operandName, version})
+}
+
+func (f *fakeVersionRecorder) UnsetVersion(_ string)                  {}
+func (f *fakeVersionRecorder) GetVersions() map[string]string         { return nil }
+func (f *fakeVersionRecorder) VersionChangedChannel() <-chan struct{} { return nil }
+
+func expectVersionRecorded(r *fakeVersionRecorder) error {
```
p0lyn0mial: unused ?

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -756,6 +765,37 @@ func (f *fakePodLister) Pods(namespace string) corev1listers.PodNamespaceLister
...
+	expected := []setVersionCall{{OperandName: "apiserver", Version: "v1.0.0-test"}}
+	if d := cmp.Diff(expected, r.setVersionCalls); d != "" {
+		return fmt.Errorf("unexpected SetVersion calls (-want +got):\n%s", d)
+	}
+	return nil
+}
+
+func expectVersionNotRecorded(r *fakeVersionRecorder) error {
```
p0lyn0mial: unused ?

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -249,14 +247,10 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 	if len(errs) > 0 {
-		message := ""
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
-		}
 		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("SyncError").
-			WithMessage(message)
+			WithMessage(errMessage(errs))
```
p0lyn0mial: why not to simply use `errors.Join()`? 
we just need to check `nil`.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -704,16 +703,20 @@ func TestUpdateOperatorStatus(t *testing.T) {
 				syncErrrors:             scenario.errors,
 			}
 
+			recorder := &fakeVersionRecorder{}
```
p0lyn0mial: could we simply use `status.NewVersionGetter()` ?
that would allow us to use `GetVersions()`.

pkg/operator/apiserver/controller/workload/workload.go
tchap: Good point, amending.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: Yeah, good point, overlooked that one.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: Yeah, I will need this later, but I will remove this for now.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -726,28 +728,34 @@ func TestUpdateOperatorStatus(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
+			if scenario.validateVersionRecorder != nil {
```
p0lyn0mial: do we have a case that sets `validateVersionRecorder` ?

tchap: I kinda wanted to add all the helpers initially in this PR, but it seems it will be more clear to add it when actually needed...

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -249,14 +250,10 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
 	if len(errs) > 0 {
-		message := ""
-		for _, err := range errs {
-			message = message + err.Error() + "\n"
-		}
 		workloadDegradedCondition = workloadDegradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
 			WithReason("SyncError").
-			WithMessage(message)
+			WithMessage(errors.Join(errs...).Error())
```
p0lyn0mial: can `Error()` handle `nil` values (no errs) ?

pkg/operator/apiserver/controller/workload/workload_test.go
p0lyn0mial: mhm, then let's intro `validateVersionRecorder` at a later time.

pkg/operator/apiserver/controller/workload/workload.go
tchap: It can't, but here we check right above that `len(errs) > 0`.

pkg/operator/apiserver/controller/workload/workload_test.go
tchap: Removed.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -57,6 +58,7 @@ func TestUpdateOperatorStatus(t *testing.T) {
 
 		workload                        *appsv1.Deployment
 		pods                            []*corev1.Pod
+		podListErr                      error
```
p0lyn0mial: do we have a case that sets `podListErr` ?

tchap: Removed.

pkg/operator/apiserver/controller/workload/workload.go
p0lyn0mial: mhm, what if `errs := []error{nil}` ?

tchap: I would consider that a bug and a reason to panic, honestly 🥲 But ok, I can make it super defensive.

tchap: Amended.

PR #2202, opened 2026-05-06
CNTRLPLANE-3237: Error out when encryption-config Secret contains mismatching KMS providers

ardaguclu:
As promised in https://github.com/openshift/library-go/pull/2163#discussion_r3188194716, this PR errors out when there is a discrepancy between same KMS providers in different resources. 

## Summary by CodeRabbit

* **New Features**
  * Added KMS provider configuration support with cross-resource consistency validation.

* **Bug Fixes**
  * Improved error handling when deriving encryption configuration to avoid proceeding on conversion failures.
  * Early-failure on inconsistent KMS provider usage to prevent applying conflicting encryption settings.

* **Tests**
  * Added and updated tests covering KMS provider validation and related error paths.

### DIFF
--- a/pkg/operator/encryption/encryptiondata/config.go
+++ b/pkg/operator/encryption/encryptiondata/config.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	"k8s.io/klog/v2"
@@ -36,7 +37,7 @@ func (c *Config) HasEncryptionConfiguration() bool {
 }
 
 // FromEncryptionState converts encryption state to Config.
-func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) *Config {
+func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) (*Config, error) {
 	resourceConfigs := make([]apiserverconfigv1.ResourceConfiguration, 0, len(encryptionState))
 	var kmsProviders map[string]*configv1.KMSConfig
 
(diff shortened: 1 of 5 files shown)


ardaguclu:
/hold
until https://github.com/openshift/library-go/pull/2163 merges

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -56,7 +57,11 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 				if kmsProviders == nil {
 					kmsProviders = map[string]*configv1.KMSConfig{}
 				}
-				if _, exists := kmsProviders[key.Key.Name]; !exists {
+				if provider, exists := kmsProviders[key.Key.Name]; exists {
+					if !reflect.DeepEqual(provider, key.KMSConfig.Provider) {
```
bertinatto: since we're comparing api types, should we use the equality package from upstream? There's also `maps.Equal`, but I think equality is more appropriate

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -56,7 +57,11 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
...
+						return nil, fmt.Errorf("KMS provider config mismatch for keyID %s: configs from different resources must be identical", key.Key.Name)
```
bertinatto, line 64: in what scenarios can this happen? When the user manually change the secrets? Or when there's a bug in our code?

ardaguclu: > When the user manually change the secrets?

I think mostly this can cause

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Good idea, I'll use it

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: > Or when there's a bug in our code?

that too, that's why we want to add an extra check.

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: +1

p0lyn0mial: please add a comment explaining why we need an extra check.

ardaguclu: Updated to use `equality.Semantic.DeepEqual`.

pkg/operator/encryption/encryptiondata/config_test.go
```diff
@@ -705,6 +708,54 @@ func newFakeIdentityKeyForTest() []byte {
 	return make([]byte, 16)
 }
 
+func TestFromEncryptionStateKMSProviderConfigMismatch(t *testing.T) {
```
p0lyn0mial: nit: could you rework to table tests ?

p0lyn0mial:
LGTM

p0lyn0mial:
I'm not tagging so that @bertinatto has a chance to look at the PR.

pkg/operator/encryption/encryptiondata/config_test.go
ardaguclu: Updated for table tests

bertinatto (commented):
/lgtm

PR #2203, opened 2026-05-06
OCPBUGS-65896: controller/workload: Rework Progressing condition

tchap:
Replace the generation-based progressing check and the 15-minute v1helpers.IsUpdatingTooLong timer with the deployment controller's native ProgressDeadlineExceeded condition. This avoids false Progressing=True signals on scaling operations (where generation changes but no rollout occurs) and uses a more accurate stall detection mechanism.

Extract progressing condition logic into a new DeploymentProgressingCondition helper in pkg/apps/deployment, along with exported HasDeploymentProgressed and HasDeploymentTimedOutProgressing helpers. The degraded condition now reports ProgressDeadlineExceeded when the deployment has timed out, instead of the generic UnavailablePod reason.

[Testing PR for CAO](https://github.com/openshift/cluster-authentication-operator/pull/889)

This is split from https://github.com/openshift/library-go/pull/2128

## Notes for Review

I published the helpers in a separate package because there are multiple other components that will be able to use them later on. I also reviewed the test cases as much as I could. Claude says all combinations are covered.

I also made sure not to touch Degraded in this PR, it's gonna come next.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment progress detection, timeout messaging, and handling of stalled rollouts (including default-replica scenarios).
  * More accurate workload status transitions and degradation handling during rollouts and timeouts.

* **Refactor**
  * Centralized progression logic so controllers use a single progression-driven path and trigger updates at the appropriate revision.

* **Tests**
  * Added extensive tests for deployment progression, timeouts, rollout scenarios, revision/generation mismatches, and pod-listing error paths.

### DIFF
--- a/pkg/apps/deployment/progressing.go
+++ b/pkg/apps/deployment/progressing.go
@@ -1,0 +1,65 @@
+package deployment
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// DeploymentProgressingCondition computes an operator Progressing condition from
+// the deployment's own status conditions and replica counts.
+func DeploymentProgressingCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+	desiredReplicas := ptr.Deref(deployment.Spec.Replicas, 1)
+	timedOutMessage, timedOut := HasDeploymentTimedOutProgressing(deployment.Status)
+
+	switch {
+	case !HasDeploymentProgressed(deployment.Status) && !timedOut:
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "PodsUpdating",
+			Message: fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest revision and %d/%d pods are available", deployment.Name, deployment.Namespace, deployment.Status.UpdatedReplicas, desiredReplicas, deployment.Status.AvailableReplicas, desiredReplicas),
+		}
+
+	case timedOut:
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "ProgressDeadlineExceeded",
+			Message: fmt.Sprintf("deployment/%s.%s has timed out progressing: %s", deployment.Name, deployment.Namespace, timedOutMessage),
+		}
+
+	default:
+		return operatorv1.OperatorCondition{
+			Type:   operatorv1.OperatorStatusTypeProgressing,
+			Status: operatorv1.ConditionFalse,
+			Reason: "AsExpected",
+		}
+	}
+}
+
+// HasDeploymentProgressed returns true if the deployment reports NewReplicaSetAvailable
+// via the DeploymentProgressing condition.
+func HasDeploymentProgressed(status appsv1.DeploymentStatus) bool {
+	for _, cond := range status.Conditions {
+		if cond.Type == appsv1.DeploymentProgressing {
+			return cond.Status == corev1.ConditionTrue && cond.Reason == "NewReplicaSetAvailable"
+		}
+	}
+	return false
+}
+
+// HasDeploymentTimedOutProgressing returns true if the deployment reports ProgressDeadlineExceeded.
+// The function returns the Progressing condition message as the first return value.
+func HasDeploymentTimedOutProgressing(status appsv1.DeploymentStatus) (string, bool) {
+	for _, cond := range status.Conditions {
+		if cond.Type == appsv1.DeploymentProgressing {
+			return cond.Message, cond.Status == corev1.ConditionFalse && cond.Reason == "ProgressDeadlineExceeded"
+		}
+	}
+	return "", false
+}
--- a/pkg/apps/deployment/progressing.go
+++ b/pkg/apps/deployment/progressing.go
@@ -1,0 +1,65 @@
+package deployment
+
+import (
+	"fmt"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// DeploymentProgressingCondition computes an operator Progressing condition from
+// the deployment's own status conditions and replica counts.
+func DeploymentProgressingCondition(deployment *appsv1.Deployment) operatorv1.OperatorCondition {
+	desiredReplicas := ptr.Deref(deployment.Spec.Replicas, 1)
+	timedOutMessage, timedOut := HasDeploymentTimedOutProgressing(deployment.Status)
+
+	switch {
+	case timedOut:
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "ProgressDeadlineExceeded",
+			Message: fmt.Sprintf("deployment/%s.%s has timed out progressing: %s", deployment.Name, deployment.Namespace, timedOutMessage),
+		}
+
+	case !HasDeploymentProgressed(deployment.Status):
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "PodsUpdating",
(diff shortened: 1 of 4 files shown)


pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -301,34 +301,16 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
-		// so that the old pods terminate before the new ones are started.
-		deploymentProgressingCondition = deploymentProgressingCondition.
-			WithStatus(operatorv1.ConditionFalse).
-			WithReason("AsExpected")
-	}
+	// Update is done when the deployment controller has reported NewReplicaSetAvailable.
+	// Checking the current vs. observed generation here is not possible since we don't want to be Progressing on scaling.
+	progressingCond := deployment.DeploymentProgressingCondition(workload)
+	_, workloadIsBeingUpdatedTooLong := deployment.HasDeploymentTimedOutProgressing(workload.Status)
+	workloadIsBeingUpdated := !deployment.HasDeploymentProgressed(workload.Status) && !workloadIsBeingUpdatedTooLong
```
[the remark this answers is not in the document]
tchap: This will be fixed in a subsequent PR.

pkg/operator/apiserver/controller/workload/workload_test.go
```diff
@@ -61,7 +61,6 @@ func TestUpdateOperatorStatus(t *testing.T) {
 		operatorPreconditionsNotReady   bool
 		preconditionError               error
 		errors                          []error
-		previousConditions              []operatorv1.OperatorCondition
```
tchap, line 64: Previous conditions are not really consulted and we replace all conditions, so this is not very useful.

pkg/apps/deployment/progressing.go
```diff
@@ -0,0 +1,65 @@
...
+	desiredReplicas := ptr.Deref(deployment.Spec.Replicas, 1)
+	timedOutMessage, timedOut := HasDeploymentTimedOutProgressing(deployment.Status)
+
+	switch {
+	case timedOut:
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  "ProgressDeadlineExceeded",
+			Message: fmt.Sprintf("deployment/%s.%s has timed out progressing: %s", deployment.Name, deployment.Namespace, timedOutMessage),
```
everettraven, line 25: I think the more common printing of a namespace-name pairing follows the `<namespace>/<name>` format. Since that is pretty standard, it would be helpful for debugging purposes to follow that pattern instead of a net-new pattern.

```suggestion
			Message: fmt.Sprintf("deployment %s/%s has timed out progressing: %s", deployment.Namespace, deployment.Name, timedOutMessage),
```

pkg/apps/deployment/progressing.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+	case !HasDeploymentProgressed(deployment.Status):
+		return operatorv1.OperatorCondition{
+			Type:    operatorv1.OperatorStatusTypeProgressing,
+			Status:  operatorv1.ConditionTrue,
+			Reason:  "PodsUpdating",
+			Message: fmt.Sprintf("deployment/%s.%s: %d/%d pods have been updated to the latest revision and %d/%d pods are available", deployment.Name, deployment.Namespace, deployment.Status.UpdatedReplicas, desiredReplicas, deployment.Status.AvailableReplicas, desiredReplicas),
```
everettraven, line 33: Same comment re: namespace-name pairing formatting.

pkg/operator/apiserver/controller/workload/workload.go
```diff
@@ -299,42 +299,22 @@ func (c *Controller) updateOperatorStatus(ctx context.Context, previousStatus *o
...
-	workloadAtHighestGeneration := workload.ObjectMeta.Generation == workload.Status.ObservedGeneration
-	// Update is done when all pods have been updated to the latest revision
-	// and the deployment controller has reported NewReplicaSetAvailable
-	workloadIsBeingUpdated := !workloadAtHighestGeneration || !hasDeploymentProgressed(workload.Status)
-	workloadIsBeingUpdatedTooLong := v1helpers.IsUpdatingTooLong(previousStatus, *deploymentProgressingCondition.Type)
-	if !workloadAtHighestGeneration {
-		deploymentProgressingCondition = deploymentProgressingCondition.
-			WithStatus(operatorv1.ConditionTrue).
-			WithReason("NewGeneration").
-			WithMessage(fmt.Sprintf("deployment/%s.%s: observed generation is %d, desired generation is %d.", workload.Name, c.targetNamespace, workload.Status.ObservedGeneration, workload.ObjectMeta.Generation))
```
everettraven, line 314: It looks like this check ends up getting removed entirely, which I'm not sure we want?

It seems reasonable to me that we validate that the observed state (status) is reflective of the desired generation of the deployment.

For the linked bug, have we pinpointed _why_ the generation on the deployment is changing but no rollouts are occurring?

tchap: The trigger for this issue is that operators using this package get Progressing on scaling, which happens when the operator uses Deployment in a DaemonSet-like manner, i.e. it sets desired replicas to the number of nodes, which changes on scaling. And that triggers a new generation, I guess. So we can't be checking generation to check Progressing. So the current way of defining Progressing is in the helper function in the `deployment` package now.

pkg/apps/deployment/progressing.go
tchap: Amended the commit.

pkg/apps/deployment/progressing.go
tchap: Amended the commit.

pkg/operator/apiserver/controller/workload/workload.go
everettraven: If we aren't checking the generation, then we are risking using a stale observed state to make our decisions around the state of the operator's operands. This means we will likely end up with false-positive and/or false-negative scenarios when checking the state of the deployment.

If the goal here is to strictly identify when the deployment spec has changed due to a scaling operation and prevent changes to the progressing condition, we should instead devise a mechanism for that to be reported to this controller. Without doing some more digging on usage, it seems reasonable to me (for the use case in question) to use a replicas-only change as a signal that there is a reaction based on scaling operations happening.

tchap: Yeah, I think that I went through this option in the past, but the interface is not very handy for this case. The `Delegate` interface actually does the sync and only returns the current workload. But truth to be told, I can probably just store the previous workload object and compare. This is not bullet-proof as restarting the operator will obviously cause that field to be reset and in that case we should probably always mark as Progressing, but that does seem like an edge case to me. Thoughts?

everettraven: IIRC there is an annotation that gets set based on a hash of the spec of the Deployment.

Is there a way we can leverage a similar bit of logic but omit replicas from the hashing by always zeroing it out to use that as a signal that nothing but the replicas field has been modified?

tchap: This can be surely implemented in various ways using an annotation. I think that I proposed this originally and Filip rejected that solution in the original PR, that's how we ended up here. He basically said that there were already too many annotations. There is no annotation I know of that we could use since this is a special case. The existing hash annotation is based on the related configmaps, secrets and proxy settings. So we would probably have to add an annotation just for this use case...

I am wondering how much of a problem there is with the current implementation. Ok, you can decide based on stale data, but how do we actually mark Progressing? The operator is progressing unless we are in the end state or we timed out. That's super inclusive. In other words, if we see a stale state, what happens? We will stay progressing a bit longer, I guess, because the lag can only be from Progressing=True to Progressing=False, if I can imagine this correctly.

tchap: Hmm, but I guess the same can happen in the other direction. If the k8s Deployment controller is not picking updates, we won't be progressing ever.

everettraven: > I think that I proposed this originally and Filip rejected that solution in the original PR, that's how we ended up here. He basically said that there were already too many annotations. There is no annotation I know of that we could use since this is a special case. The existing hash annotation is based on the related configmaps, secrets and proxy settings. So we would probably have to add an annotation just for this use case...

Happen to have a link handy? I'd be curious to see the arguments against an annotation with a hash that would allow us to determine whether or not the change is likely based on a cluster scaling operation.

I think if we need to have another annotation to solve this problem it is probably fine.

>I am wondering how much of a problem there is with the current implementation. Ok, you can decide based on stale data, but how do we actually mark Progressing? The operator is progressing unless we are in the end state or we timed out. That's super inclusive. In other words, if we see a stale state, what happens? We will stay progressing a bit longer, I guess, because the lag can only be from Progressing=True to Progressing=False, if I can imagine this correctly.
>Hmm, but I guess the same can happen in the other direction. If the k8s Deployment controller is not picking updates, we won't be progressing ever.

In practice, I'm not sure how often it would actually happen that we are observing an invalid state.

My concern is around the fact that we would have no meaningful signal and it would likely be difficult for us to pinpoint that case. 

Having false-negatives and false-positives as potential end-results here is a big red flag to me. We should be avoiding as much as possible telling a customer that something is happening when it isn't and that something isn't happening when it is.

I'd like to see an approach where we can clearly attribute a change to cluster scaling while preserving the ability to verify that the state we are observing matches the generation of our desired configuration.

PR #2208, opened 2026-05-08
CNTRLPLANE-3237: Complete renaming of KMS structs to better clarity

ardaguclu:
This PR reflects the API changes merged in https://github.com/openshift/api/pull/2833 as well as renamings that we agreed upon;

* KMSConfig to KMSState
* configv1.KMSConfig to configv1.KMSPluginConfig
* kms-provider-config to kms-plugin-config

Apart from the mechanical renamings, [KMSPluginConfig is updated](https://github.com/openshift/api/blob/047c8957c266b27a4220a6369560850e325ae586/config/v1/types_apiserver.go#L243) to value type from reference type. That means, there is no nil check anymore for KMSPluginConfig. This PR updates the KMSPluginConfig usages to value type.

There should not be any functional changes.

## Summary by CodeRabbit

* **Dependencies**
  * Updated OpenShift API and client module revisions.

* **Refactor**
  * Migrated KMS encryption representation from a provider to a plugin model; internal secret/data key naming and operator state were adjusted while preserving external encryption behavior.

* **Tests**
  * Updated unit and end-to-end tests and test helpers to validate and exercise the KMS plugin configuration flow.


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -307,9 +307,6 @@ func (c *keyController) getCurrentModeReasonAndEncryptionConfig(ctx context.Cont
 	case state.AESCBC, state.AESGCM, state.Identity: // secretbox is disabled for now
 		return currentMode, reason, encryption, nil
 	case state.KMS:
-		if encryption.KMS == nil {
```
ardaguclu, line 310: encryption.KMS is value type. It will always be populated after https://github.com/openshift/api/pull/2836 merges.

pkg/operator/encryption/state/types.go
```diff
@@ -42,24 +42,24 @@ type KeyState struct {
...
 
 func (k *KeyState) HasKMSEncryption() bool {
-	return k != nil && k.KMSConfig != nil && k.KMSConfig.Encryption != nil
+	return k != nil && k.KMSState != nil && k.KMSState.Encryption != nil
 }
 
-func (k *KeyState) HasKMSProvider() bool {
-	return k != nil && k.KMSConfig != nil && k.KMSConfig.Provider != nil
+func (k *KeyState) HasKMSPlugin() bool {
+	return k != nil && k.KMSState != nil && k.KMSState.Plugin != (configv1.KMSPluginConfig{})
```
ardaguclu: `k.KMSState.Plugin` is value type. We compare the emptiness against `(configv1.KMSPluginConfig{})`

pkg/operator/encryption/state/types.go
```diff
@@ -42,24 +42,24 @@ type KeyState struct {
 	// the user via unsupportConfigOverrides.encryption.reason triggered this key.
 	ExternalReason string
 	// stores all the KMS encryption mode related configurations
-	KMSConfig *KMSConfig
+	KMS *KMSState
```
p0lyn0mial, line 45: didn't we want `KMS *KMS` ?

ardaguclu: Unfortunately KMS struct name collides with KMS encryption mode in the same file

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -28,28 +28,28 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*Config, error) {
...
-			kmsProviders = map[string]*configv1.KMSConfig{}
+		if kmsPlugins == nil {
+			kmsPlugins = map[string]configv1.KMSPluginConfig{}
 		}
-		kmsProviders[keyID] = providerConfig
+		kmsPlugins[keyID] = pluginConfig
 	}
 
-	return &Config{Encryption: encryptionConfig, KMSProviders: kmsProviders}, nil
+	return &Config{Encryption: encryptionConfig, KMSPlugins: kmsPlugins}, nil
```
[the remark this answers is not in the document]
ardaguclu: This entry recently added. It hasn't been used by anyone. So this is backwards compatible change.

pkg/operator/encryption/encoding/encoding.go
```diff
@@ -75,12 +75,9 @@ func DecodeKMSConfiguration(data []byte) (*apiserverconfigv1.KMSConfiguration, e
...
 
-// EncodeKMSConfig serializes a configv1.KMSConfig into a configv1.APIServer wrapper.
-// We use a configv1.APIServer as an envelope type because configv1.KMSConfig is not a runtime.Object.
-func EncodeKMSConfig(kmsConfig *configv1.KMSConfig) ([]byte, error) {
-	if kmsConfig == nil {
-		return nil, fmt.Errorf("KMSConfig object cannot be nil")
-	}
+// EncodeKMSPluginConfig serializes a configv1.KMSPluginConfig into a configv1.APIServer wrapper.
+// We use a configv1.APIServer as an envelope type because configv1.KMSPluginConfig is not a runtime.Object.
+func EncodeKMSPluginConfig(kmsConfig configv1.KMSPluginConfig) ([]byte, error) {
```
bertinatto, line 80: I see that we're removing the nil check here because kmsConfig is no longer a pointer, but we're not doing any validation at admission time yet, are we?

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -827,8 +827,8 @@ func TestGetCurrentModeReasonAndEncryptionConfig(t *testing.T) {
 			if externalReason != scenario.expectedReasonFromCfg {
 				t.Errorf("unexpected reason read from the config: %q, expected: %q", externalReason, scenario.expectedReasonFromCfg)
 			}
-			if currentMode == "KMS" && encryption.KMS == nil {
-				t.Errorf("expected non-nil KMS config when mode is KMS")
+			if currentMode == "KMS" && encryption.KMS == (configv1.KMSPluginConfig{}) {
+				t.Errorf("expected non-empty KMS config when mode is KMS")
```
bertinatto, line 831: this is something I hadn't seen before. Apparently this will no longer work if we add a non-comparable field to KMSPluginConfig. I think it's fine for now

pkg/operator/encryption/encoding/encoding.go
bertinatto: I guess this shouldn't block this PR, but it'd be nice if we can follow up at the api front

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -39,32 +39,32 @@ func (c *Config) HasEncryptionConfiguration() bool {
...
-						return nil, fmt.Errorf("KMS provider config mismatch for keyID %s: configs from different resources must be identical", key.Key.Name)
+					// an identical plugin config, since they originate from the same Key Secret.
+					if !equality.Semantic.DeepEqual(plugin, key.KMS.Plugin) {
+						return nil, fmt.Errorf("KMS plugin config mismatch for keyID %s: configs from different resources must be identical", key.Key.Name)
 					}
 				} else {
-					kmsProviders[key.Key.Name] = key.KMSConfig.Provider
+					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
```
bertinatto, line 69: This file has a more functions:

```
func createKMSProviderName(keyID, resource string) string {
	// Ideally we should have used keyID simply in kms provider name.
	// However, this is an upstream constraint that every provider name must be unique.
	// To maintain uniqueness while still allowing access to the keyID, we generate provider name in this format.
	return fmt.Sprintf("%s_%s", keyID, resource)
}

func getKeyIDFromProviderName(providerName string) (string, error) {
	// We just need to obtain the keyID to find our backed secret
	// e.g. "1_secrets"
	parsed := strings.SplitN(providerName, "_", 2)
	if len(parsed) != 2 {
		return "", fmt.Errorf("invalid provider name %q: expected format keyID_resourceName", providerName)
	}
	return parsed[0], nil
}

```

should we rename these as well?

bertinatto: if we need to rename them, we can probably do that in a follow up

test/e2e-encryption/encryption_test.go
```diff
@@ -530,7 +530,7 @@ func TestEncryptionIntegration(tt *testing.T) {
 	)
 	waitForMigration("12")
 	waitForConditionStatus("Encrypted", operatorv1.ConditionTrue)
-	verifyKMSProviders()
+	verifyKMSPlugins()
```
bertinatto, line 533: I think there's a `kmsProviderName()` down below

bertinatto (commented):
I think I found more stuff that could be renamed, but we don't have to do that in a single PR

/lgtm

pkg/operator/encryption/encoding/encoding.go
ardaguclu: I think this https://github.com/openshift/api/pull/2836 will fix it

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Thank you for review and approval. I've opened this https://github.com/openshift/library-go/pull/2211 for the missing bits.

PR #2209, opened 2026-05-08
 CNTRLPLANE-3236: add initial bits for plugin lifecycle

bertinatto:
/assign @ardaguclu 

Proof: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2146

## Summary by CodeRabbit

* **New Features**
  * KMS support: automatic detection and decoding of active KMS endpoint and plugin config from encryption secrets, Vault KMS sidecar provider, and automatic injection of the KMS sidecar plus socket and resource mounts into API server pods.

* **Tests**
  * Added unit tests covering endpoint/key parsing, plugin-config decoding, Vault sidecar container construction, and sidecar injection success and failure scenarios.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -1,0 +1,161 @@
+package pluginlifecycle
+
+import (
+	"context"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
+)
+
+// sidecarProvider abstracts the construction of a KMS plugin sidecar container for a specific provider (e.g. Vault).
+type sidecarProvider interface {
+	// Name returns the identifier used to name the sidecar container and locate its volume mounts.
+	Name() string
+	// BuildSidecarContainer returns a fully configured sidecar container ready to be injected into the API server pod
+	BuildSidecarContainer() (corev1.Container, error)
+}
+
+// newSidecarProvider creates a provider-specific SidecarProvider for the given keyID, UDS endpoint, and plugin configuration.
+func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSPluginConfig) (sidecarProvider, error) {
+	switch pluginConfig.Type {
+	case configv1.VaultKMSProvider:
+		return newVaultSidecarProvider("vault-kms-plugin", keyID, udsPath, pluginConfig)
+	default:
+		return nil, fmt.Errorf("unsupported KMS plugin configuration")
+	}
+}
+
+// AddKMSPluginSidecarToPodSpec discovers KMS plugins from the encryption-config secret and injects a sidecar container for each one into the pod spec.
+// It uses an uncached client to avoid injecting sidecars based on a stale encryption configuration.
+func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter) error {
+	if podSpec == nil {
+		return fmt.Errorf("pod spec cannot be nil")
+	}
+
+	if containerName == "" {
+		return fmt.Errorf("container name cannot be empty")
+	}
+
+	encryptionConfigurationSecret, err := secretClient.Secrets(encryptionConfigNamespace).Get(ctx, encryptionConfigSecretName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Infof("skipping KMS sidecar injection: %s/%s secret not found", encryptionConfigNamespace, encryptionConfigSecretName)
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get %s/%s secret: %w", encryptionConfigNamespace, encryptionConfigSecretName, err)
+	}
+
+	encryptionConfig, err := encryptiondata.FromSecret(encryptionConfigurationSecret)
+	if err != nil {
+		return fmt.Errorf("failed to extract encryption config from %s/%s secret: %w", encryptionConfigNamespace, encryptionConfigSecretName, err)
+	}
+
+	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(encryptionConfig)
+	if err != nil {
+		return fmt.Errorf("failed to get KMS configurations: %w", err)
+	}
+	if len(kmsConfigurations) == 0 {
+		klog.V(4).Infof("skipping KMS sidecar injection: no KMS plugins found in EncryptionConfiguration")
+		return nil
+	}
+
+	klog.V(4).Infof("injecting %d KMS sidecar(s)", len(kmsConfigurations))
+
+	for _, kmsConfiguration := range kmsConfigurations {
+		// ExtractUniqueAndSortedKMSConfigurations function rewrites the .Name field to include only the key ID
+		keyID := kmsConfiguration.Name
+		udsPath := kmsConfiguration.Endpoint
+
+		pluginConfig, ok := encryptionConfig.KMSPlugins[keyID]
+		if !ok {
+			return fmt.Errorf("missing plugin config for keyID %s", keyID)
+		}
+
+		sidecarProvider, err := newSidecarProvider(keyID, udsPath, pluginConfig)
+		if err != nil {
+			return fmt.Errorf("failed to create a sidecar provider for keyID %s: %w", keyID, err)
+		}
+
+		if err := appendSidecarContainer(podSpec, sidecarProvider); err != nil {
+			return err
+		}
+
+		if err := ensureSocketVolumeMount(podSpec, sidecarProvider.Name()); err != nil {
+			return err
+		}
+	}
+
+	if err := ensureSocketVolumeMount(podSpec, containerName); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func appendSidecarContainer(podSpec *corev1.PodSpec, provider sidecarProvider) error {
+	container, err := provider.BuildSidecarContainer()
+	if err != nil {
+		return fmt.Errorf("failed to build sidecar container: %w", err)
+	}
+
+	podSpec.Containers = append(podSpec.Containers, container)
+	return nil
+}
+
+func ensureSocketVolumeMount(podSpec *corev1.PodSpec, containerName string) error {
+	containerIndex := -1
+	for i, container := range podSpec.Containers {
+		if container.Name == containerName {
+			containerIndex = i
+			break
+		}
+	}
+
+	if containerIndex < 0 {
+		return fmt.Errorf("container %s not found", containerName)
+	}
+
+	container := &podSpec.Containers[containerIndex]
+	foundMount := false
+	for _, m := range container.VolumeMounts {
+		if m.Name == "kms-plugin-socket" {
+			foundMount = true
+			break
+		}
+	}
+	if !foundMount {
+		container.VolumeMounts = append(container.VolumeMounts,
+			corev1.VolumeMount{
+				Name:      "kms-plugin-socket",
+				MountPath: "/var/run/kmsplugin",
+			},
+		)
+	}
+
+	// The volume mount in the container requires a volume in the podSpec
+	ensureSocketVolume(podSpec)
+	return nil
+}
+
+func ensureSocketVolume(podSpec *corev1.PodSpec) {
+	for _, volume := range podSpec.Volumes {
+		if volume.Name == "kms-plugin-socket" {
+			return
+		}
+	}
+
+	podSpec.Volumes = append(podSpec.Volumes,
+		corev1.Volume{
+			Name: "kms-plugin-socket",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	)
+}
--- a/pkg/operator/encryption/kms/pluginlifecycle/vault.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/vault.go
@@ -1,0 +1,50 @@
+package pluginlifecycle
+
+import (
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
+func newVaultSidecarProvider(name, keyID, udsPath string, pluginConfig configv1.KMSPluginConfig) (*vault, error) {
+	return &vault{
+		name:    name,
+		keyID:   keyID,
+		udsPath: udsPath,
+		config:  &pluginConfig.Vault,
+	}, nil
+}
+
+// vault implements SidecarProvider for HashiCorp Vault KMS.
+type vault struct {
+	name    string
+	keyID   string
+	udsPath string
+	config  *configv1.VaultKMSPluginConfig
+}
+
+// Name returns the sidecar name appended by the key id.
+func (v *vault) Name() string {
+	return fmt.Sprintf("%s-%s", v.name, v.keyID)
+}
+
+// BuildSidecarContainer returns a container spec for the Vault KMS plugin sidecar
+// configured with the Vault address, namespace, transit mount, and transit key.
+func (v *vault) BuildSidecarContainer() (corev1.Container, error) {
+	return corev1.Container{
+		Name:  v.Name(),
+		Image: v.config.KMSPluginImage,
+		Args: []string{
+			// TODO(bertinatto): this path is a placeholder for the mock plugin. This path doesn't exist. It will need to be validated
+			// once we support the actual Vault plugin. It'll likely be different for KASO vs. aggregated API server operators.
+			fmt.Sprintf("-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-%s", v.keyID),
+			fmt.Sprintf("-listen-address=%s", v.udsPath),
+			fmt.Sprintf("-vault-address=%s", v.config.VaultAddress),
+			fmt.Sprintf("-vault-namespace=%s", v.config.VaultNamespace),
+			fmt.Sprintf("-transit-mount=%s", v.config.TransitMount),
+			fmt.Sprintf("-transit-key=%s", v.config.TransitKey),
+		},
+	}, nil
+}
--- a/pkg/operator/encryption/kms/pluginlifecycle/vault_test.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/vault_test.go
@@ -1,0 +1,147 @@
+package pluginlifecycle
+
+import (
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func TestVaultSidecarProvider_BuildSidecarContainer(t *testing.T) {
+	tests := []struct {
+		name            string
+		pluginConfig    configv1.KMSPluginConfig
+		containerName   string
+		keyID           string
+		udsPath         string
+		inputPodSpec    *corev1.PodSpec
+		expectedPodSpec *corev1.PodSpec
+		wantErr         string
+	}{
+		{
+			name: "builds container with correct args",
+			pluginConfig: configv1.KMSPluginConfig{
+				Type: configv1.VaultKMSProvider,
+				Vault: configv1.VaultKMSPluginConfig{
+					KMSPluginImage: "quay.io/test/vault:v2",
+					VaultAddress:   "https://vault.example.com:8200",
+					VaultNamespace: "my-namespace",
+					TransitKey:     "my-key",
+					TransitMount:   "transit",
+				},
+			},
+			containerName: "kms-plugin",
+			keyID:         "555",
+			udsPath:       "unix:///var/run/kmsplugin/kms-555.sock",
+			inputPodSpec:  &corev1.PodSpec{},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "kms-plugin-555",
+						Image: "quay.io/test/vault:v2",
+						Args: []string{
+							"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-555",
+							"-listen-address=unix:///var/run/kmsplugin/kms-555.sock",
+							"-vault-address=https://vault.example.com:8200",
+							"-vault-namespace=my-namespace",
+							"-transit-mount=transit",
+							"-transit-key=my-key",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "appends to existing containers",
+			pluginConfig: configv1.KMSPluginConfig{
+				Type: configv1.VaultKMSProvider,
+				Vault: configv1.VaultKMSPluginConfig{
+					KMSPluginImage: "quay.io/test/vault:v2",
+					VaultAddress:   "https://vault.example.com:8200",
+					VaultNamespace: "my-namespace",
+					TransitKey:     "my-key",
+					TransitMount:   "transit",
+				},
+			},
+			containerName: "kms-plugin",
+			keyID:         "555",
+			udsPath:       "unix:///var/run/kmsplugin/kms-555.sock",
+			inputPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "kube-apiserver",
+						Image: "registry.k8s.io/kube-apiserver:v1.30.0",
+					},
+				},
+			},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "kube-apiserver",
+						Image: "registry.k8s.io/kube-apiserver:v1.30.0",
+					},
+					{
+						Name:  "kms-plugin-555",
+						Image: "quay.io/test/vault:v2",
+						Args: []string{
+							"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-555",
+							"-listen-address=unix:///var/run/kmsplugin/kms-555.sock",
+							"-vault-address=https://vault.example.com:8200",
+							"-vault-namespace=my-namespace",
+							"-transit-mount=transit",
+							"-transit-key=my-key",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "empty optional fields",
+			pluginConfig: configv1.KMSPluginConfig{
+				Type: configv1.VaultKMSProvider,
+				Vault: configv1.VaultKMSPluginConfig{
+					KMSPluginImage: "quay.io/test/vault:v2",
+					VaultAddress:   "https://vault.example.com:8200",
+				},
+			},
+			containerName: "kms-plugin",
+			keyID:         "999",
+			udsPath:       "unix:///var/run/kmsplugin/kms.sock",
+			inputPodSpec:  &corev1.PodSpec{},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:  "kms-plugin-999",
+						Image: "quay.io/test/vault:v2",
+						Args: []string{
+							"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-999",
+							"-listen-address=unix:///var/run/kmsplugin/kms.sock",
+							"-vault-address=https://vault.example.com:8200",
+							"-vault-namespace=",
+							"-transit-mount=",
+							"-transit-key=",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			provider, err := newVaultSidecarProvider(tt.containerName, tt.keyID, tt.udsPath, tt.pluginConfig)
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
+				return
+			}
+			require.NoError(t, err)
+
+			container, err := provider.BuildSidecarContainer()
+			require.NoError(t, err)
+
+			tt.inputPodSpec.Containers = append(tt.inputPodSpec.Containers, container)
+			require.Equal(t, tt.expectedPodSpec, tt.inputPodSpec)
+		})
+	}
+}
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
@@ -1,0 +1,333 @@
+package pluginlifecycle
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/encoding"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+	fake "k8s.io/client-go/kubernetes/fake"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func TestInjectIntoPodSpec(t *testing.T) {
+	secretClient := func(secrets ...*corev1.Secret) corev1client.SecretsGetter {
+		objs := make([]runtime.Object, len(secrets))
+		for i, s := range secrets {
+			objs[i] = s
+		}
+		return fake.NewSimpleClientset(objs...).CoreV1()
+	}
+
+	vaultConfig := &configv1.KMSPluginConfig{
+		Type: configv1.VaultKMSProvider,
+		Vault: configv1.VaultKMSPluginConfig{
+			KMSPluginImage: "quay.io/test/vault:v1",
+			VaultAddress:   "https://vault.example.com:8200",
+			VaultNamespace: "my-namespace",
+			TransitKey:     "my-key",
+			TransitMount:   "transit",
+			Authentication: configv1.VaultAuthentication{
+				Type: configv1.VaultAuthenticationTypeAppRole,
+				AppRole: configv1.VaultAppRoleAuthentication{
+					Secret: configv1.VaultSecretReference{Name: "vault-kms-credentials"},
+				},
+			},
+		},
+	}
+	pluginConfigBytes, err := encoding.EncodeKMSPluginConfig(*vaultConfig)
+	require.NoError(t, err)
+
+	pluginConfigKey, err := kms.ToPluginConfigSecretDataKeyFor("555")
+	require.NoError(t, err)
+
+	encryptionConfig := &apiserverv1.EncryptionConfiguration{
+		Resources: []apiserverv1.ResourceConfiguration{
+			{
+				Resources: []string{"secrets"},
+				Providers: []apiserverv1.ProviderConfiguration{
+					{
+						KMS: &apiserverv1.KMSConfiguration{
+							APIVersion: "v2",
+							Name:       "555_secrets",
+							Endpoint:   "unix:///var/run/kmsplugin/kms-555.sock",
+						},
+					},
+				},
+			},
+		},
+	}
+	encryptionConfigBytes, err := encoding.EncodeEncryptionConfiguration(encryptionConfig)
+	require.NoError(t, err)
+
+	encryptionConfigSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "encryption-config",
+			Namespace: "openshift-kube-apiserver",
+		},
+		Data: map[string][]byte{
+			"encryption-config": encryptionConfigBytes,
+			pluginConfigKey:     pluginConfigBytes,
+		},
+	}
+
+	sidecarArgs := []string{
+		"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-555",
+		"-listen-address=unix:///var/run/kmsplugin/kms-555.sock",
+		"-vault-address=https://vault.example.com:8200",
+		"-vault-namespace=my-namespace",
+		"-transit-mount=transit",
+		"-transit-key=my-key",
+	}
+
+	socketMount := corev1.VolumeMount{
+		Name:      "kms-plugin-socket",
+		MountPath: "/var/run/kmsplugin",
+	}
+	socketVolume := corev1.Volume{
+		Name: "kms-plugin-socket",
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
+		},
+	}
+	resourceDirVolume := corev1.Volume{
+		Name: "resource-dir",
+		VolumeSource: corev1.VolumeSource{
+			HostPath: &corev1.HostPathVolumeSource{
+				Path: "/etc/kubernetes/static-pod-resources",
+			},
+		},
+	}
+
+	tests := []struct {
+		name            string
+		actualPodSpec   *corev1.PodSpec
+		expectedPodSpec *corev1.PodSpec
+		secretClient    corev1client.SecretsGetter
+		wantErr         string
+	}{
+		{
+			name: "single provider: sidecar and volumes injected",
+			actualPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+				Volumes: []corev1.Volume{resourceDirVolume},
+			},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:         "kube-apiserver",
+						VolumeMounts: []corev1.VolumeMount{socketMount},
+					},
+					{
+						Name:         "vault-kms-plugin-555",
+						Image:        "quay.io/test/vault:v1",
+						Args:         sidecarArgs,
+						VolumeMounts: []corev1.VolumeMount{socketMount},
+					},
+				},
+				Volumes: []corev1.Volume{resourceDirVolume, socketVolume},
+			},
+			secretClient: secretClient(encryptionConfigSecret),
+		},
+		{
+			name: "two providers during migration: both sidecars injected in keyID order",
+			actualPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+				Volumes: []corev1.Volume{resourceDirVolume},
+			},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:         "kube-apiserver",
+						VolumeMounts: []corev1.VolumeMount{socketMount},
+					},
+					{
+						Name:  "vault-kms-plugin-777",
+						Image: "quay.io/test/vault:v2",
+						Args: []string{
+							"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-777",
+							"-listen-address=unix:///var/run/kmsplugin/kms-777.sock",
+							"-vault-address=https://vault2.example.com:8200",
+							"-vault-namespace=other-namespace",
+							"-transit-mount=transit2",
+							"-transit-key=other-key",
+						},
+						VolumeMounts: []corev1.VolumeMount{socketMount},
+					},
+					{
+						Name:  "vault-kms-plugin-555",
+						Image: "quay.io/test/vault:v1",
+						Args: []string{
+							"-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-555",
+							"-listen-address=unix:///var/run/kmsplugin/kms-555.sock",
+							"-vault-address=https://vault.example.com:8200",
+							"-vault-namespace=my-namespace",
+							"-transit-mount=transit",
+							"-transit-key=my-key",
+						},
+						VolumeMounts: []corev1.VolumeMount{socketMount},
+					},
+				},
+				Volumes: []corev1.Volume{resourceDirVolume, socketVolume},
+			},
+			secretClient: func() corev1client.SecretsGetter {
+				vaultConfig2 := &configv1.KMSPluginConfig{
+					Type: configv1.VaultKMSProvider,
+					Vault: configv1.VaultKMSPluginConfig{
+						KMSPluginImage: "quay.io/test/vault:v2",
+						VaultAddress:   "https://vault2.example.com:8200",
+						VaultNamespace: "other-namespace",
+						TransitKey:     "other-key",
+						TransitMount:   "transit2",
+					},
+				}
+				pluginConfig2Bytes, err := encoding.EncodeKMSPluginConfig(*vaultConfig2)
+				require.NoError(t, err)
+
+				pluginConfigKey2, err := kms.ToPluginConfigSecretDataKeyFor("777")
+				require.NoError(t, err)
+
+				multiEncConfig := &apiserverv1.EncryptionConfiguration{
+					Resources: []apiserverv1.ResourceConfiguration{
+						{
+							Resources: []string{"secrets"},
+							Providers: []apiserverv1.ProviderConfiguration{
+								{
+									KMS: &apiserverv1.KMSConfiguration{
+										APIVersion: "v2",
+										Name:       "555_secrets",
+										Endpoint:   "unix:///var/run/kmsplugin/kms-555.sock",
+									},
+								},
+								{
+									KMS: &apiserverv1.KMSConfiguration{
+										APIVersion: "v2",
+										Name:       "777_secrets",
+										Endpoint:   "unix:///var/run/kmsplugin/kms-777.sock",
+									},
+								},
+							},
+						},
+					},
+				}
+				multiEncConfigBytes, err := encoding.EncodeEncryptionConfiguration(multiEncConfig)
+				require.NoError(t, err)
+
+				return secretClient(&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "encryption-config",
+						Namespace: "openshift-kube-apiserver",
+					},
+					Data: map[string][]byte{
+						"encryption-config": multiEncConfigBytes,
+						pluginConfigKey:     pluginConfigBytes,
+						pluginConfigKey2:    pluginConfig2Bytes,
+					},
+				})
+			}(),
+		},
+		{
+			name:          "nil pod spec",
+			actualPodSpec: nil,
+			secretClient:  secretClient(encryptionConfigSecret),
+			wantErr:       "pod spec cannot be nil",
+		},
+		{
+			name: "missing encryption-config secret: pod spec unchanged",
+			actualPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+			},
+			expectedPodSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: "kube-apiserver"},
+				},
+			},
+			secretClient: secretClient(),
+		},
+		{
+			name: "missing encryption-config key in secret: pod spec unchanged",
+			actualPodSpec: &corev1.PodSpec{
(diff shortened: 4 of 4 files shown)


pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil
+	}
+
+	encryptionConfiguration, err := encoding.DecodeEncryptionConfiguration(encryptionConfigurationBytes)
+	if err != nil {
+		return fmt.Errorf("failed to decode EncryptionConfiguration from %s/%s secret: %w", opConfig.EncryptionConfigNamespace, opConfig.EncryptionConfigSecretName, err)
+	}
+
+	// TODO: don't get only the first? Deduplicate and do this in a loop
+	kmsConfiguration := findFirstKMSConfiguration(encryptionConfiguration)
```
ardaguclu: If we deduplicate based on uds path, we don't need to get the first provider

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,36 @@
...
+func (v *VaultMockSidecarProvider) BuildSidecarContainer(name string, kmsConfig *apiserverv1.KMSConfiguration) (corev1.Container, error) {
+	if v.Config == nil {
+		return corev1.Container{}, fmt.Errorf("vault config cannot be nil")
+	}
+
+	return corev1.Container{
+		Name:  name,
+		Image: v.Config.KMSPluginImage,
+		Args: []string{
+			"-approle-secret-id-path=/var/run/this-file-does-not-exist",
```
ardaguclu: That works for me

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+func InjectIntoPodSpec(podSpec *corev1.PodSpec, secretLister corev1listers.SecretLister, opConfig OperatorConfig) error {
+	if podSpec == nil {
+		return fmt.Errorf("pod spec cannot be nil")
+	}
+
+	encryptionConfigurationSecret, err := secretLister.Secrets(opConfig.EncryptionConfigNamespace).Get(opConfig.EncryptionConfigSecretName)
+	if err != nil {
```
ardaguclu: If not notfound error, propagate err

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	encryptionConfigurationSecret, err := secretLister.Secrets(opConfig.EncryptionConfigNamespace).Get(opConfig.EncryptionConfigSecretName)
+	if err != nil {
+		klog.V(4).Infof("KMS is disabled: could not get %s secret: %v", opConfig.EncryptionConfigSecretName, err)
+		return nil
+	}
+
+	encryptionConfigurationBytes, ok := encryptionConfigurationSecret.Data["encryption-config"]
+	if !ok {
```
ardaguclu: Return err, this should not happen

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,187 @@
...
+
+	encryptionConfiguration, err := encoding.DecodeEncryptionConfiguration(encryptionConfigurationBytes)
+	if err != nil {
+		return fmt.Errorf("failed to decode EncryptionConfiguration from %s/%s secret: %w", opConfig.EncryptionConfigNamespace, opConfig.EncryptionConfigSecretName, err)
+	}
+
+	// TODO: don't get only the first? Deduplicate and do this in a loop
+	kmsConfiguration := findFirstKMSConfiguration(encryptionConfiguration)
+	if kmsConfiguration == nil {
+		// TODO: should error out instead?
```
ardaguclu: I think yes we should

pkg/operator/encryption/kms/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +96,42 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
...
+	}
+	return matches[1], nil
+}
+
+func parseProviderConfig(secret *corev1.Secret, kmsConfiguration *apiserverv1.KMSConfiguration) (*configv1.KMSConfig, error) {
+	keyID, err := parseKeyIDFromEndpoint(kmsConfiguration.Endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse key ID from endpoint: %w", err)
+	}
+	providerConfigKey, err := ToProviderConfigSecretDataKeyFor(keyID)
```
ardaguclu: We first need to merge this https://github.com/openshift/library-go/pull/2208. After we will rename these to pluginConfigKey

pkg/operator/encryption/kms/sidecar.go
bertinatto: updated, PTAL and let me know if it makes sense

pkg/operator/encryption/kms/helpers.go
bertinatto: yup, sounds good

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +96,53 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
 
 	return nil
 }
+
+func findKMSConfiguration(config *apiserverv1.EncryptionConfiguration) (*apiserverv1.KMSConfiguration, error) {
+	byEndpoint := map[string]*apiserverv1.KMSConfiguration{}
+	for _, resource := range config.Resources {
```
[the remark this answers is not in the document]
bertinatto: fixed

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+func NewVaultSidecarProvider(name string, pluginConfig *configv1.KMSPluginConfig) (*Vault, error) {
+	return &Vault{
+		name:   name,
+		Config: &pluginConfig.Vault,
+	}, nil
+}
+
+// Vault implements SidecarProvider for HashiCorp Vault KMS.
+// Currently, only the vault mock plugin is supported.
+type Vault struct {
```
ardaguclu: We only need UDS path from the KMSConfiguration. Maybe it is better to store the UDS here as well, so that we don't need to pass `KMSConfiguration` into the functions. 

I experimented this in https://github.com/openshift/library-go/commit/ab41f9a4f4535941729003d81d290f7d7f39c89c

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +96,56 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
...
+func parseKeyIDFromEndpoint(endpoint string) (string, error) {
+	matches := kmsEndpointRegexp.FindStringSubmatch(endpoint)
+	if matches == nil {
+		return "", fmt.Errorf("unexpected KMS endpoint format: %s", endpoint)
+	}
+	return matches[1], nil
+}
+
+func parsePluginConfig(secret *corev1.Secret, kmsConfiguration *apiserverv1.KMSConfiguration) (*configv1.KMSPluginConfig, error) {
+	keyID, err := parseKeyIDFromEndpoint(kmsConfiguration.Endpoint)
```
ardaguclu: We should parse from provider name. There is existed util function for it. We just need to move it to here to be reusable https://github.com/openshift/library-go/commit/28678070d9e2b7b0bebb3211eb7d6522a607d71f

ardaguclu: But it is also good to parse keyID from endpoint as well. So that keyID from provider.name and keyID from endpoint must match. I experimented https://github.com/openshift/library-go/commit/49405a2289fe210cdf13a5d2b481ef49b3c66160

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,207 @@
...
+	APIServerContainerName     string
+}
+
+// SidecarProvider abstracts the construction of a KMS plugin sidecar container
+// for a specific provider (e.g. Vault).
+type SidecarProvider interface {
+	// Name returns the identifier used to name the sidecar container and locate its volume mounts.
+	Name() string
+	// BuildSidecarContainer returns a fully configured sidecar container for the given KMS configuration.
+	BuildSidecarContainer(kmsConfiguration *apiserverv1.KMSConfiguration) (corev1.Container, error)
```
ardaguclu: If we embed UDS in VaultConfig. We don't need to pass kmsConfiguration anymore.

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+)
+
+// NewVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
+func NewVaultSidecarProvider(name string, pluginConfig *configv1.KMSPluginConfig) (*Vault, error) {
+	return &Vault{
+		name:   name,
+		Config: &pluginConfig.Vault,
```
p0lyn0mial: does it have to be exported ?

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+)
+
+// NewVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
+func NewVaultSidecarProvider(name string, pluginConfig *configv1.KMSPluginConfig) (*Vault, error) {
+	return &Vault{
```
p0lyn0mial: should we validate `pluginConfig` ?

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (v *Vault) Name() string {
+	return v.name
+}
+
+// BuildSidecarContainer returns a container spec for the Vault KMS plugin sidecar
+// configured with the Vault address, namespace, transit mount, and transit key.
+func (v *Vault) BuildSidecarContainer(kmsConfig *apiserverv1.KMSConfiguration) (corev1.Container, error) {
+	if v.Config == nil {
```
p0lyn0mial: should we do the validation in the constructor ?

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func newSidecarProvider(pluginConfig *configv1.KMSPluginConfig) (SidecarProvider, error) {
+	switch pluginConfig.Type {
+	case configv1.VaultKMSProvider:
+		return plugins.NewVaultSidecarProvider("vault-kms-plugin", pluginConfig)
```
p0lyn0mial: does the name have to be unique ?

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+		return corev1.Container{}, fmt.Errorf("vault config cannot be nil")
+	}
+
+	return corev1.Container{
+		Name:  v.Name(),
+		Image: v.Config.KMSPluginImage,
+		Args: []string{
+			"-approle-secret-id-path=/var/run/this-file-does-not-exist",
```
p0lyn0mial: is this because the wrapper requires this flag to be passed ?

p0lyn0mial: if yes, please add a comment explaining why and stating this is a temporary situation.

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
...
+		name:   name,
+		Config: &pluginConfig.Vault,
+	}, nil
+}
+
+// Vault implements SidecarProvider for HashiCorp Vault KMS.
+// Currently, only the vault mock plugin is supported.
+type Vault struct {
+	name   string
+	Config *configv1.VaultKMSPluginConfig
```
p0lyn0mial: does it have to be exported ? why ?

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,49 @@
+package plugins
```
p0lyn0mial: what is the purpose of this package ?

maybe we should create a new package `pluginlifecycle` (or something similar) where we would put all related code ?

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	encryptionConfigurationBytes, ok := encryptionConfigurationSecret.Data["encryption-config"]
+	if !ok {
+		klog.V(4).Infof("KMS is disabled: could not get encryption-config key in secret %s", encryptionConfigurationSecret.Name)
+		return err
```
p0lyn0mial: what err are we returning here ?

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	encryptionConfiguration, err := encoding.DecodeEncryptionConfiguration(encryptionConfigurationBytes)
+	if err != nil {
+		return fmt.Errorf("failed to decode EncryptionConfiguration from %s/%s secret: %w", opConfig.EncryptionConfigNamespace, opConfig.EncryptionConfigSecretName, err)
+	}
+
+	// TODO: currently only Vault is supported, so there should be only one KMSConfiguration
```
p0lyn0mial: what if we are in `KMS1 -> KMS2` migration scenario?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +96,56 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
 
 	return nil
 }
+
+func findKMSConfiguration(config *apiserverv1.EncryptionConfiguration) (*apiserverv1.KMSConfiguration, error) {
```
p0lyn0mial: @ardaguclu is there a room for a small refactor ? would it make sense to have a small utility function that knows the details like the fact that there are duplicates ....

p0lyn0mial: maybe this function could return (*KMSConfig, boolean (found), err) ?

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,207 @@
...
+	case configv1.VaultKMSProvider:
+		return plugins.NewVaultSidecarProvider("vault-kms-plugin", pluginConfig)
+	default:
+		return nil, fmt.Errorf("unsupported KMS plugin configuration")
+	}
+}
+
+// InjectIntoPodSpec reads the encryption configuration secret, determines the active KMS plugin, and injects the
+// corresponding sidecar container and required volume mounts into the given pod spec. It is a no-op when KMS is not configured.
+func InjectIntoPodSpec(podSpec *corev1.PodSpec, secretLister corev1listers.SecretLister, opConfig OperatorConfig) error {
```
p0lyn0mial: I think that in general this code should assume there can be multiple providers defined (KMS1 - > KMS2).

pkg/operator/encryption/kms/helpers.go
ardaguclu: Yes, that would be better

pkg/operator/encryption/kms/sidecar.go
p0lyn0mial: could OperatorConfig be replaced by `namespace, secretName, apiServerContainer string` ?

p0lyn0mial: this function cannot be consumed by the preflight checker which will be run before we have encryption configuration secret.

p0lyn0mial:
I'm not saying that we should update the PR now but in the future there will be a need to also inject https://github.com/openshift/enhancements/pull/2005/changes (probably not for the preflight checker)

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: imported your commit

pkg/operator/encryption/kms/helpers.go
bertinatto: imported your commit

pkg/operator/encryption/kms/sidecar.go
bertinatto: imported your commits

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: it doesn't need to be exported anymore

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: added validation to pluginConfig

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: the constructor is the one responsible for settings this field

pkg/operator/encryption/kms/sidecar.go
bertinatto: Arda's commit changes this so that it's unique

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: yes, the wrapper needs something; but I can use the file path intended to be used by the real vault plugin instead

pkg/operator/encryption/kms/sidecar.go
bertinatto: it should cover the migration scenario now

pkg/operator/encryption/kms/sidecar.go
bertinatto: this should be covered now

bertinatto: the function signature would be very long, but it can be added. Let me know

bertinatto: should the changes be introduced in this PR or in the PR that introduces the function in the preflight checker? If they should be introduced now, what specific changes does it need?

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: the idea was to store the vendor-specific implementation for providers listed in https://github.com/openshift/api/blob/master/config/v1/types_kmsencryption.go#L7-L32

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: will we also need the `KMSConfiguration.Timeou` in addition to `KMSConfiguration.Endpoint`?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -152,7 +152,7 @@ func ToEncryptionState(secretData *Config, keySecrets []*corev1.Secret) (map[sch
 
 			case provider.KMS != nil:
 				// Name and Secret must match to find our backed Secret
-				keyID, err := getKeyIDFromPluginName(provider.KMS.Name)
+				keyID, err := kms.GetKeyIDFromProviderName(provider.KMS.Name)
```
ardaguclu:
```suggestion
				keyID, err := kms.GetKeyIDFromPluginName(provider.KMS.Name)
```

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +93,94 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
...
+		}
+	}
+	return byKeyID, nil
+}
+
+// GetKeyIDFromProviderName extracts the keyID from a KMS provider name. The provider name
+// follows the format "{keyID}_{resourceName}" (e.g., "555_secrets"), where the keyID portion
+// identifies the encryption key and the resource portion ensures uniqueness across resource
+// groups as required by the upstream EncryptionConfiguration schema.
+func GetKeyIDFromProviderName(providerName string) (string, error) {
```
ardaguclu:
```suggestion
func GetKeyIDFromPluginName(pluginName string) (string, error) {
```

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,275 @@
...
+	BuildSidecarContainer() (corev1.Container, error)
+}
+
+// newSidecarProvider creates a provider-specific SidecarProvider for the given keyID, UDS endpoint,
+// and plugin configuration. The keyID is incorporated into the sidecar container name to ensure
+// uniqueness when multiple providers coexist during a KMS-to-KMS migration.
+//
+// For example, a Vault provider with keyID "555" and endpoint "unix:///var/run/kmsplugin/kms-555.sock"
+// produces a sidecar container named "vault-kms-plugin-555" that listens on that socket.
+func newSidecarProvider(keyID string, udsPath string, pluginConfig *configv1.KMSPluginConfig) (SidecarProvider, error) {
```
ardaguclu:
```suggestion
func newSidecarPlugin(keyID string, udsPath string, pluginConfig *configv1.KMSPluginConfig) (SidecarProvider, error) {
```

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: Should we return error if encryption-config key entry is absent in encryption-config Secret?

ardaguclu (commented):
Mostly nitpicks
/lgtm
/hold
in case you want to fix them now or followup PR

ardaguclu:
verify needs to be fixed as well

pkg/operator/encryption/kms/sidecar.go
bertinatto: I did consider renaming this, but "provider" here is not meant to be a "kms provider", it's meant to be "something that provides a sidecar" (think of it as a [Stringer](https://pkg.go.dev/fmt#Stringer) interface). I'm happy to rename if you disagree

ardaguclu: ah, that makes sense to me

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: in that case the secret would be malformed, so I think we should return an error. Fixed

pkg/operator/encryption/encryptiondata/config.go
bertinatto: fixed

pkg/operator/encryption/kms/helpers.go
bertinatto: changed

pkg/operator/encryption/kms/sidecar.go
ardaguclu: I think it is better to pass one by one instead OperatorConfig. Because it is hard to understand what fields are exactly used. This will be longer but more readable imo.

ardaguclu:
Just one nit.
/lgtm

pkg/operator/encryption/kms/plugins/vault.go
ardaguclu: I think validation includes non-empty vault-address, transit-key, etc.

pkg/operator/encryption/kms/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: this validation is also done when those fields are set in the api, does it make sense to validate them again here?

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: I updated this to use the actual path where the secret-id is available. This should work for both the mock and the real plugin

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: this field is no longer exported

pkg/operator/encryption/kms/sidecar.go
bertinatto: added context to `err`

pkg/operator/encryption/kms/helpers.go
bertinatto: the function returns a map now, which is later checked with `if len(kmsEndpointsByKeyID) == 0`; does that help?

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,223 @@
...
+		return nil
+	}
+
+	encryptionConfigurationSecret, err := secretLister.Secrets(encryptionConfigNamespace).Get(encryptionConfigSecretName)
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Infof("skipping KMS sidecar injection: %s/%s secret not found", encryptionConfigNamespace, encryptionConfigSecretName)
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get %s/%s secret", encryptionConfigNamespace, encryptionConfigSecretName)
```
ardaguclu: Embed err

ardaguclu:
@bertinatto if all the comments from @p0lyn0mial are addressed, maybe we can merge this PR. WDYT?

bertinatto:
> @bertinatto if all the comments from @p0lyn0mial are addressed, maybe we can merge this PR. WDYT?

I think there are some open questions about the preflight checker not being able to use `InjectIntoPodSpec` as is, and the use of a `plugins` package. I don't have a preference, I'll leave that to you. BTW, I lost your tag because I pushed an update.

pkg/operator/encryption/kms/plugins/vault.go
ardaguclu: KMSPluginConfig is validated in API side with all necessary data. I agree that we don't need to validate same

pkg/operator/encryption/kms/sidecar.go
ardaguclu: I think, according to this PR, preflight checker can easily do;

```go
sidecarProvider, err := newSidecarProvider(keyID, udsPath, pluginConfig)
container, err := provider.BuildSidecarContainer()
if err != nil {
   return fmt.Errorf("failed to build sidecar container: %w", err)
}
```
keyID and udsPath can be arbitrary. pluginConfig is retrieved from API. The rest volume mounts, etc. are static manifests. 

Note: `newSidecarProvider` should be exported.

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NewVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
+func NewVaultSidecarProvider(name, keyID, udsPath string, pluginConfig *configv1.KMSPluginConfig) (*Vault, error) {
```
p0lyn0mial: what is the point of exposing this function ? It seems like calling code cares about `InjectIntoPodSpec`

pkg/operator/encryption/kms/plugins/vault.go
p0lyn0mial: I'd just put it behind a `pluginlifecycle` pkg from wich I would only expose `InjectIntoPodSpec`

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/openshift/library-go/pkg/operator/encryption/encoding"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/plugins"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+)
+
+// SidecarProvider abstracts the construction of a KMS plugin sidecar container for a specific provider (e.g. Vault).
+type SidecarProvider interface {
```
p0lyn0mial: if we had `pluginlifecycle` then there is no need to expose this interface.

pkg/operator/encryption/kms/sidecar.go
```diff
@@ -0,0 +1,222 @@
...
+	switch pluginConfig.Type {
+	case configv1.VaultKMSProvider:
+		return plugins.NewVaultSidecarProvider("vault-kms-plugin", keyID, udsPath, pluginConfig)
+	default:
+		return nil, fmt.Errorf("unsupported KMS plugin configuration")
+	}
+}
+
+// InjectIntoPodSpec discovers KMS plugins from the encryption-config secret and injects a sidecar container for each one into the pod spec. No-op when the KMS feature gate is disabled or no KMS plugins are configured.
+func InjectIntoPodSpec(podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretLister corev1listers.SecretLister, featureGateAccessor featuregates.FeatureGateAccess) error {
```
p0lyn0mial: in fact, `SidecarProvider` is not used here.

pkg/operator/encryption/kms/plugins/vault.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Name returns the sidecar name appended by the key id.
+func (v *Vault) Name() string {
+	return fmt.Sprintf("%s-%s", v.name, v.keyID)
+}
+
+// BuildSidecarContainer returns a container spec for the Vault KMS plugin sidecar
+// configured with the Vault address, namespace, transit mount, and transit key.
+func (v *Vault) BuildSidecarContainer() (corev1.Container, error) {
+	if v.config == nil {
```
p0lyn0mial: didn't we check in the in the constructor ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,222 @@
...
+	featureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get feature gates: %w", err)
+	}
+
+	if !featureGates.Enabled(features.FeatureGateKMSEncryption) {
+		return nil
+	}
+
+	encryptionConfigurationSecret, err := secretLister.Secrets(encryptionConfigNamespace).Get(encryptionConfigSecretName)
```
p0lyn0mial: maybe we should use a live client. all encryption controllers use a live client. 
maybe we could use https://github.com/openshift/library-go/blob/fb92a809fef516581af274227a103075d121174c/pkg/operator/encryption/preconditions.go#L25 as a precondition ?

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +93,51 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
...
+	for _, resource := range config.Resources {
+		for _, provider := range resource.Providers {
+			if provider.KMS == nil {
+				continue
+			}
+			keyID, err := GetKeyIDFromPluginName(provider.KMS.Name)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse key ID from provider name %q: %w", provider.KMS.Name, err)
+			}
+			byKeyID[keyID] = provider.KMS.Endpoint
```
p0lyn0mial: @ardaguclu didn't we want to validate the provider for the same key ? just in case we intro a bug ?

pkg/operator/encryption/kms/sidecar.go
p0lyn0mial: could we find a better name for this `InjectIntoPodSpec` ? previously we had `AddKMSPluginVolumeAndMountToPodSpec` which was more descriptive.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,222 @@
...
+	if apierrors.IsNotFound(err) {
+		klog.V(4).Infof("skipping KMS sidecar injection: %s/%s secret not found", encryptionConfigNamespace, encryptionConfigSecretName)
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get %s/%s secret: %w", encryptionConfigNamespace, encryptionConfigSecretName, err)
+	}
+
+	encryptionConfigurationBytes, ok := encryptionConfigurationSecret.Data["encryption-config"]
```
p0lyn0mial: could we just use `encryptiondata.FromSecret` ?

ardaguclu: That is a very good idea

pkg/operator/encryption/kms/helpers.go
ardaguclu: Yes, I think we should validate if there is any mismatches between the same providers.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: That would save a lot of code indeed.

pkg/operator/encryption/kms/sidecar.go
p0lyn0mial: what is `containerName` used for ? should we validate if the str is not empty ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: Note: I think `kmsEndpointsByKeyID` might still be needed to extract uds paths per each key.

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: it's exposed because it's supposed to be called outside this package?

pkg/operator/encryption/kms/plugins/vault.go
bertinatto: this was a validation in case it was created without the constructor. Removed

pkg/operator/encryption/kms/sidecar.go
bertinatto: validated empty string

bertinatto: renamed

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,211 @@
...
+
+		if err := appendSidecarContainer(podSpec, sidecarProvider); err != nil {
+			return err
+		}
+
+		if err := ensureSocketVolumeMount(podSpec, sidecarProvider.Name()); err != nil {
+			return err
+		}
+
+		if err := ensureResourceDirVolumeMount(podSpec, sidecarProvider.Name()); err != nil {
```
p0lyn0mial: should we also check for the existence of resource-dir volume?

pkg/operator/encryption/kms/sidecar.go
bertinatto: renamed `plugin` to `pluginlifecycle` and moved this file there.

the interface is now unexported. Should it be removed instead?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,211 @@
...
+	switch pluginConfig.Type {
+	case configv1.VaultKMSProvider:
+		return newVaultSidecarProvider("vault-kms-plugin", keyID, udsPath, pluginConfig)
+	default:
+		return nil, fmt.Errorf("unsupported KMS plugin configuration")
+	}
+}
+
+// AddKMSPluginSidecarToPodSpec discovers KMS plugins from the encryption-config secret and injects a sidecar container for each one into the pod spec. No-op when the KMS feature gate is disabled or no KMS plugins are configured.
+func AddKMSPluginSidecarToPodSpec(podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretLister corev1listers.SecretLister) error {
```
p0lyn0mial: what is `containerName` used for ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
p0lyn0mial: maybe we should use lister and rely on the precondition function ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,211 @@
...
+	container, err := provider.BuildSidecarContainer()
+	if err != nil {
+		return fmt.Errorf("failed to build sidecar container: %w", err)
+	}
+
+	podSpec.Containers = append(podSpec.Containers, container)
+	return nil
+}
+
+func ensureResourceDirVolumeMount(podSpec *corev1.PodSpec, containerName string) error {
```
p0lyn0mial: why is this needed ?

bertinatto: this is where the credentials are propagated to. The Volume already exists in the kube-apiserver pod, but it's not used in the aggregated apiservers which use a Deployment. It's not required for the mock plugin, so maybe I can remove it

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: removed the volume mount

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: switched to a client, PTAL

pkg/operator/encryption/kms/helpers.go
```diff
@@ -90,3 +93,51 @@ func AddKMSPluginVolumeAndMountToPodSpec(podSpec *corev1.PodSpec, containerName
 
 	return nil
 }
+
+// KMSendpointsByKeyID walks through all resource configurations in the EncryptionConfiguration
+// and collects every unique KMS provider, deduplicating by endpoint.
+func KMSendpointsByKeyID(config *apiserverv1.EncryptionConfiguration) (map[string]string, error) {
```
p0lyn0mial: do we need that function ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -287,13 +287,3 @@ func createKMSPluginName(keyID, resource string) string {
 	// To maintain uniqueness while still allowing access to the keyID, we generate plugin name in this format.
 	return fmt.Sprintf("%s_%s", keyID, resource)
 }
-
-func getKeyIDFromPluginName(pluginName string) (string, error) {
```
p0lyn0mial, line 291: then we can revert this change back.

pkg/operator/encryption/kms/helpers.go
ardaguclu: I think we don't need newly added functions in kms/helpers.go anymore

bertinatto: yeah, removed

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -0,0 +1,51 @@
...
+	return fmt.Sprintf("%s-%s", v.name, v.keyID)
+}
+
+// BuildSidecarContainer returns a container spec for the Vault KMS plugin sidecar
+// configured with the Vault address, namespace, transit mount, and transit key.
+func (v *vault) BuildSidecarContainer() (corev1.Container, error) {
+	return corev1.Container{
+		Name:  v.Name(),
+		Image: v.config.KMSPluginImage,
+		Args: []string{
```
p0lyn0mial, line 39: are all these flags required or some are optional ?
if all are required then should we validate the values ? (just in case ?)

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -0,0 +1,51 @@
...
+			fmt.Sprintf("-approle-secret-id-path=/var/run/secrets/vault-kms/secret-id-%s", v.keyID),
```
p0lyn0mial, line 42: will this file be written on disk by the installer for kas ? 
how will it be mounted for the other servers ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if err := ensureSocketVolumeMount(podSpec, containerName); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func appendSidecarContainer(podSpec *corev1.PodSpec, provider sidecarProvider) error {
+	if podSpec == nil {
```
p0lyn0mial: nit: this is redundant because we already check it in `AddKMSPluginSidecarToPodSpec`

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+			corev1.Volume{
+				Name: "kms-plugin-socket",
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+			},
+		)
+	}
+
+	return nil
```
p0lyn0mial: this function always returns `nil`

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+	// The volume mount in the container requires a volume in the podSpec
+	return ensureSocketVolume(podSpec)
+}
+
+func ensureSocketVolume(podSpec *corev1.PodSpec) error {
+	foundVolume := false
+	for _, volume := range podSpec.Volumes {
+		if volume.Name == "kms-plugin-socket" {
+			foundVolume = true
+			break
```
p0lyn0mial: should we return early ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+	case configv1.VaultKMSProvider:
+		return newVaultSidecarProvider("vault-kms-plugin", keyID, udsPath, pluginConfig)
+	default:
+		return nil, fmt.Errorf("unsupported KMS plugin configuration")
+	}
+}
+
+// AddKMSPluginSidecarToPodSpec discovers KMS plugins from the encryption-config secret and injects a sidecar container for each one into the pod spec.
+// It uses an uncached client to avoid injecting sidecars based on a stale encryption configuration.
+func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter) error {
```
p0lyn0mial, line 36: would it make sense to add a test for verifying that calling AddKMSPluginSidecarToPodSpec twice doesn't duplicate sidecars or volumes ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+		// ExtractUniqueAndSortedKMSConfigurations function rewrites the .Name field to include only the key ID
+		keyID := kmsConfiguration.Name
+		udsPath := kmsConfiguration.Endpoint
+
+		pluginConfig, ok := encryptionConfig.KMSPlugins[keyID]
+		if !ok {
+			return fmt.Errorf("missing plugin config for keyID %s", keyID)
+		}
+
+		sidecarProvider, err := newSidecarProvider(keyID, udsPath, &pluginConfig)
```
p0lyn0mial: nit: do we have to pass a * to pluginConfig ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get %s/%s secret: %w", encryptionConfigNamespace, encryptionConfigSecretName, err)
+	}
+
+	encryptionConfig, err := encryptiondata.FromSecret(encryptionConfigurationSecret)
+	if err != nil {
+		return fmt.Errorf("failed to extract encryption config from %s/%s secret: %w", encryptionConfigNamespace, encryptionConfigSecretName, err)
+	}
+	if !encryptionConfig.HasEncryptionConfiguration() {
```
p0lyn0mial: would it make to drop this check ? already done in ExtractUniqueAndSortedKMSConfigurations .
(btw: i think at this point the encryption configuration is required).

pkg/operator/encryption/kms/pluginlifecycle/vault_test.go
```diff
@@ -0,0 +1,156 @@
...
+					VaultAddress:   "https://vault.example.com:8200",
+					VaultNamespace: "my-namespace",
+					TransitKey:     "my-key",
+					TransitMount:   "transit",
+				},
+			},
+			containerName: "kms-plugin",
+			keyID:         "555",
+			udsPath:       "unix:///var/run/kmsplugin/kms-555.sock",
+			inputPodSpec:  &corev1.PodSpec{},
```
p0lyn0mial, line 37: do these unit tests have to operate on the podSpec level ?

pkg/operator/encryption/kms/pluginlifecycle/vault_test.go
```diff
@@ -0,0 +1,156 @@
...
+			if tt.wantErr != "" {
+				require.ErrorContains(t, err, tt.wantErr)
+				return
+			}
+			require.NoError(t, err)
+
+			container, err := provider.BuildSidecarContainer()
+			require.NoError(t, err)
+
+			tt.inputPodSpec.Containers = append(tt.inputPodSpec.Containers, container)
```
p0lyn0mial, line 143: why it is important to append and test against the podSpec ?

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: yes, for kas we can rely on the files written by the installer pod on `/etc/kubernetes/static-pod-resources/...` (resource-dir volume already exists in the pod spec).

the other operators don't have that, so they will need a different approach (like provide a secret volume sourced from the revisioned secret).

p0lyn0mial: > the other operators don't have that, so they will need a different approach (like provide a secret volume sourced from the revisioned secret)

should AddKMSPluginSidecarToPodSpec know about it ? otherwise we cannot switch the other operators to use the function provided in this pr, right ?

p0lyn0mial: > yes, for kas we can rely on the files written by the installer pod on /etc/kubernetes/static-pod-resources/... (resource-dir volume already exists in the pod spec).

should we at least check the mount / path is on the pod ?

bertinatto: I trimmed down this PR to contain only the required changes to deploy the vault mock plugin in kaso, I thought we wanted smaller PRs. I can certainly add to `AddKMSPluginSidecarToPodSpec` what I have in mind

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: it seems wasteful to pass by value and copy every single field of the struct, but I can change

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: I think it can be dropped

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: does this function need to be idempotent? We had this discussion for [AddKMSPluginVolumeAndMountToPodSpec](https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/kms/helpers.go#L38-L38). Let me know

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: they have been validate during admission, do we need to validate them again? I think at this point it's safe to assume we have good values here (I think that's the point of having the validation at the api level), but I'm happy to validate again if you disagree

pkg/operator/encryption/kms/pluginlifecycle/vault_test.go
bertinatto: no, I don't think I'd need a podSpec, it was mostly personal taste

pkg/operator/encryption/kms/pluginlifecycle/vault.go
p0lyn0mial: ah, ok, that's fine then.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
p0lyn0mial: i think the code already is idempotent, for example `ensureSocketVolume` that's why i thought we could cover it in a unit test.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
p0lyn0mial: it is defined as a non pointer field elsewhere, for example https://github.com/openshift/library-go/blob/fcabbdecd6d9ad6caad5088f899d26231301c615/pkg/operator/encryption/encryptiondata/config.go#L28

maybe we should unify ? otherwise you have to always & or check for nil val.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
p0lyn0mial: fair point. okay, maybe we could skip simple validation here. thanks.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -0,0 +1,51 @@
...
+// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
+func newVaultSidecarProvider(name, keyID, udsPath string, pluginConfig *configv1.KMSPluginConfig) (*vault, error) {
+	if pluginConfig == nil {
+		return nil, fmt.Errorf("plugin config cannot be nil")
+	}
+	return &vault{
+		name:    name,
+		keyID:   keyID,
+		udsPath: udsPath,
+		config:  &pluginConfig.Vault,
```
p0lyn0mial, line 16: like here, it is defined as a non pointer and we always have to do & and then later on when you see a ponter you wonder if it can be nil

p0lyn0mial:
overall this pr lgtm, thanks for all the work. i think i will wait unit tomorrow so that @ardaguclu can also take a look.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
p0lyn0mial: > yes, for kas we can rely on the files written by the installer pod on /etc/kubernetes/static-pod-resources/... (resource-dir volume already exists in the pod spec).

>> should we at least check the mount / path is on the pod ?

@bertinatto what about ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: removed the pointer

p0lyn0mial:
should we open a test pr in kas-o to validate this PR ?

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: I changed it to be a non-pointer, but I'm not sure I follow. `pluginConfig` is (was) a pointer here, the `&` was for the `.Vault` field because `config` is a pointer. Am I missing something?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: `appendSidecarContainer` breaks the idempotency. I think we should skip if podSpec includes a container with the same name. But this can be achieved in followup PR.

> does this function need to be idempotent?

I think it would be safer. Library-go does not control how many times this function will be run (from operators, preflight checker, etc.).

ardaguclu:
Dropped a comment about idempotency which can be fixed in followup PR. Thank you
/lgtm
/hold cancel

PR #2212, opened 2026-05-11
CNTRLPLANE-3237: Carry referenced Secret data from API to plugin lifecycle for KMS

ardaguclu:
This PR carries [Vault credentials](https://github.com/openshift/api/blob/49ae21b8502fc80a351dc17ca4f0a16056797ab4/config/v1/types_kmsencryption.go#L96-L101) (according to the new format in https://github.com/openshift/api/pull/2836 -- `role-id`, `secret-id` instead of `roleID`, `secretID`) based on https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/kms-encryption-foundations.md.

This PR mimics from https://github.com/openshift/library-go/pull/2169/changes/5221c8280c1ba7448f318cde82784274d4f2604b commit. 

## Release Notes

* **New Features**
  * Enhanced KMS encryption support with improved credential and secret management. The system now fetches referenced secrets from the configuration namespace and validates required credential data keys are present for KMS plugins.

* **Bug Fixes**
  * Improved validation of KMS plugin configuration to ensure all required credentials are properly available and accessible.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -284,7 +284,9 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
 			Plugin: apiServerEncryption.KMS,
 		}
 
-		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+		if secretName, expectedKeys, err := referencedSecretName(apiServerEncryption.KMS); err != nil {
+			return nil, err
+		} else if len(secretName) > 0 {
 			refSecret, err := c.secretClient.Secrets(openshiftConfigNS).Get(ctx, secretName, metav1.GetOptions{})
 			if err != nil {
 				return nil, fmt.Errorf("failed to get secret %s in %s: %w", secretName, openshiftConfigNS, err)
@@ -404,18 +406,19 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 // referencedSecretName returns the name of the secret referenced by the KMS plugin
 // config and the specific data keys to carry from that secret. Only the listed keys
 // are copied into the Key Secret; any other data in the referenced secret is ignored.
-func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
+func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string, error) {
 	switch plugin.Type {
 	case configv1.VaultKMSProvider:
 		switch plugin.Vault.Authentication.Type {
 		case configv1.VaultAuthenticationTypeAppRole:
 			// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
 			// These are the only keys carried into the encryption key secret.
-			return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
+			return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}, nil
+		default:
+			return "", nil, fmt.Errorf("unsupported Vault authentication type %q", plugin.Vault.Authentication.Type)
 		}
-		return "", nil
 	default:
-		return "", nil
+		return "", nil, fmt.Errorf("unsupported KMS provider type %q", plugin.Type)
 	}
 }
 
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -42,6 +42,7 @@ const (
 	encryptionSecretMigrationInterval = time.Hour * 24 * 7 // one week
 	kmsEndpointFormat                 = "unix:///var/run/kmsplugin/kms-%d.sock"
 	defaultKMSTimeout                 = 10 * time.Second
+	openshiftConfigNS                 = "openshift-config"
 )
 
 // keyController creates new keys if necessary. It
@@ -117,6 +118,7 @@ func NewKeyController(
 			apiServerInformer.Informer(),
 			operatorClient.Informer(),
 			kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
+			// TODO: add informer for openshift-config namespace to watch referenced Secrets for KMS plugin secret data changes
 			deployer,
 		).ToController(
 		c.controllerInstanceName,
@@ -223,7 +225,7 @@ func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext fact
 
 	sort.Sort(sort.StringSlice(reasons))
 	internalReason := strings.Join(reasons, ", ")
-	keySecret, err := c.generateKeySecret(newKeyID, currentMode, apiEncryptionConfiguration, internalReason, externalReason)
+	keySecret, err := c.generateKeySecret(ctx, newKeyID, currentMode, apiEncryptionConfiguration, internalReason, externalReason)
 	if err != nil {
 		return fmt.Errorf("failed to create key: %v", err)
 	}
@@ -260,7 +262,7 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 	return nil // we made this key earlier
 }
 
-func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, apiServerEncryption configv1.APIServerEncryption, internalReason, externalReason string) (*corev1.Secret, error) {
+func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, currentMode state.Mode, apiServerEncryption configv1.APIServerEncryption, internalReason, externalReason string) (*corev1.Secret, error) {
 	bs := crypto.ModeToNewKeyFunc[currentMode]()
 	ks := state.KeyState{
 		Key: apiserverv1.Key{
@@ -280,6 +282,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode, 
 				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
 			},
 			Plugin: apiServerEncryption.KMS,
+		}
+
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			refSecret, err := c.secretClient.Secrets(openshiftConfigNS).Get(ctx, secretName, metav1.GetOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in %s: %w", secretName, openshiftConfigNS, err)
+			}
+			for _, key := range expectedKeys {
+				v, ok := refSecret.Data[key]
+				if !ok {
+					return nil, fmt.Errorf("secret %s in %s is missing required key %q", secretName, openshiftConfigNS, key)
+				}
+				if err := ks.KMS.PluginSecretData.Set(secretName, key, v); err != nil {
+					return nil, err
+				}
+			}
 		}
 	}
 	return secrets.FromKeyState(c.instanceName, ks)
@@ -381,6 +399,20 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 	// we check for encryptionSecretMigratedTimestamp set by migration controller to determine when migration completed
 	// this also generates back pressure for key rotation when migration takes a long time or was recently completed
 	return latestKeyID, "rotation-interval-has-passed", time.Since(latestKey.Migrated.Timestamp) > encryptionSecretMigrationInterval
+}
+
+// referencedSecretName returns the name of the secret referenced by the KMS plugin
+// config and the specific data keys to carry from that secret. Only the listed keys
+// are copied into the Key Secret; any other data in the referenced secret is ignored.
+func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
+	switch plugin.Type {
+	case configv1.VaultKMSProvider:
+		// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
+		// These are the only keys carried into the encryption key secret.
+		return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
+	default:
+		return "", nil
+	}
 }
 
 // TODO make this un-settable once set
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -407,9 +407,13 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
 	switch plugin.Type {
 	case configv1.VaultKMSProvider:
-		// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
-		// These are the only keys carried into the encryption key secret.
-		return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
+		switch plugin.Vault.Authentication.Type {
+		case configv1.VaultAuthenticationTypeAppRole:
+			// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
+			// These are the only keys carried into the encryption key secret.
+			return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
+		}
+		return "", nil
 	default:
 		return "", nil
 	}
--- a/pkg/operator/encryption/encryptiondata/config.go
+++ b/pkg/operator/encryption/encryptiondata/config.go
@@ -30,6 +30,48 @@ type Config struct {
 	// KMSPlugins maps keyID to plugin-specific configuration,
 	// carried from Key Secrets into the encryption-config Secret.
 	KMSPlugins map[string]configv1.KMSPluginConfig
+	// KMSPluginsSecretData maps keyID to secret data carried from
+	// Key Secrets into the encryption-config Secret.
+	// Structure: keyID → secretName → dataKey → value.
+	KMSPluginsSecretData KMSPluginsSecretData
+}
+
+// KMSPluginsSecretData maps keyID to secret data carried from Key Secrets into
+// the encryption-config Secret. Structure: keyID → secretName → dataKey → value.
+type KMSPluginsSecretData struct {
+	ByKeyID map[string]state.KMSSecretData
+}
+
+// SetFromRawKey stores a value for the given keyID, splitting rawKey
+// on "_" into secretName and dataKey.
+func (d *KMSPluginsSecretData) SetFromRawKey(keyID, rawKey string, value []byte) error {
+	if len(keyID) == 0 {
+		return fmt.Errorf("keyID must not be empty")
+	}
+	if d.ByKeyID == nil {
+		d.ByKeyID = map[string]state.KMSSecretData{}
+	}
+	sd := d.ByKeyID[keyID]
+	if err := sd.SetFromRawKey(rawKey, value); err != nil {
+		return err
+	}
+	d.ByKeyID[keyID] = sd
+	return nil
+}
+
+// FlatEntriesByKeyID returns the stored data as a map of keyID to flat entries,
+// where each flat entry is keyed by "secretName_dataKey".
+func (d *KMSPluginsSecretData) FlatEntriesByKeyID() map[string]map[string][]byte {
+	if d.ByKeyID == nil {
+		return nil
+	}
+	result := map[string]map[string][]byte{}
+	for keyID, sd := range d.ByKeyID {
+		if flat := sd.FlatEntries(); flat != nil {
+			result[keyID] = flat
+		}
+	}
+	return result
 }
 
 func (c *Config) HasEncryptionConfiguration() bool {
@@ -40,6 +82,7 @@ func (c *Config) HasEncryptionConfiguration() bool {
 func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupResourceState) (*Config, error) {
 	resourceConfigs := make([]apiserverconfigv1.ResourceConfiguration, 0, len(encryptionState))
 	var kmsPlugins map[string]configv1.KMSPluginConfig
+	var kmsPluginsSecretData KMSPluginsSecretData
 
 	for gr, grKeys := range encryptionState {
 		resourceConfigs = append(resourceConfigs, apiserverconfigv1.ResourceConfiguration{
@@ -47,11 +90,11 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 			Providers: stateToProviders(gr.Resource, grKeys),
 		})
 
-		// Collect KMS plugin configs from read keys (which already include the write key).
-		// We iterate over encryptionState which is keyed by GroupResource, so the same
-		// keyID is seen once per resource (e.g. key "1" for secrets and key "1" for configmaps).
-		// Since all resources share the same Key Secret, the plugin config is identical
-		// across duplicates and we only need to keep the first occurrence.
+		// Collect KMS plugin configs and secret data from read keys (which already
+		// include the write key). We iterate over encryptionState which is keyed by
+		// GroupResource, so the same keyID is seen once per resource. Since all
+		// resources share the same Key Secret, the plugin config and secret data are
+		// identical across duplicates and we only need to keep the first occurrence.
 		for _, key := range grKeys.ReadKeys {
 			if key.HasKMSPlugin() {
 				if kmsPlugins == nil {
@@ -67,6 +110,19 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
+			if key.HasKMSSecretData() {
+				if existing, exists := kmsPluginsSecretData.ByKeyID[key.Key.Name]; exists {
+					if !equality.Semantic.DeepEqual(existing, key.KMS.PluginSecretData) {
+						return nil, fmt.Errorf("KMS secret data mismatch for keyID %s: secret data from different resources must be identical", key.Key.Name)
+					}
+				} else {
+					for rawKey, value := range key.KMS.PluginSecretData.FlatEntries() {
+						if err := kmsPluginsSecretData.SetFromRawKey(key.Key.Name, rawKey, value); err != nil {
+							return nil, fmt.Errorf("failed to copy secret data for keyID %s: %w", key.Key.Name, err)
+						}
+					}
+				}
+			}
 		}
 	}
 
@@ -76,8 +132,9 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 	})
 
 	return &Config{
-		Encryption: &apiserverconfigv1.EncryptionConfiguration{Resources: resourceConfigs},
-		KMSPlugins: kmsPlugins,
+		Encryption:           &apiserverconfigv1.EncryptionConfiguration{Resources: resourceConfigs},
+		KMSPlugins:           kmsPlugins,
+		KMSPluginsSecretData: kmsPluginsSecretData,
 	}, nil
 }
 
--- a/pkg/operator/encryption/encryptiondata/secret.go
+++ b/pkg/operator/encryption/encryptiondata/secret.go
@@ -41,11 +41,15 @@ func keyIDFromPluginConfigSecretDataKey(dataKey string) (string, bool, error) {
 	return keyID, true, nil
 }
 
-// EncryptionConfSecretName is the name of the final encryption config secret that is revisioned per apiserver rollout.
-const EncryptionConfSecretName = "encryption-config"
-
-// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
-const EncryptionConfSecretKey = "encryption-config"
+const (
+	// EncryptionConfSecretName is the name of the final encryption config secret that is revisioned per apiserver rollout.
+	EncryptionConfSecretName = "encryption-config"
+	// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
+	EncryptionConfSecretKey = "encryption-config"
+	// encryptionConfigSecretDataPrefix is the data key prefix for KMS plugin secret
+	// data entries in the encryption-config Secret. Full key: "kms-plugin-secret-{secretName}_{dataKey}-{keyID}".
+	encryptionConfigSecretDataPrefix = "kms-plugin-secret-"
+)
 
 func FromSecret(encryptionConfigSecret *corev1.Secret) (*Config, error) {
 	data, ok := encryptionConfigSecret.Data[EncryptionConfSecretKey]
@@ -80,7 +84,26 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*Config, error) {
 		kmsPlugins[keyID] = pluginConfig
 	}
 
-	return &Config{Encryption: encryptionConfig, KMSPlugins: kmsPlugins}, nil
+	// Extract secret data entries from the encryption-config Secret.
+	// Data keys follow the format "kms-plugin-secret-{secretName}_{dataKey}-{keyID}"
+	// (e.g. "kms-plugin-secret-app-role_role-id-1"). keyIDFromSecretDataKey
+	// returns the keyID (e.g. "1") and the combined key (e.g. "app-role_role-id"),
+	// which is then split on "_" to recover secretName and dataKey.
+	var kmsPluginsSecretData KMSPluginsSecretData
+	for key, value := range encryptionConfigSecret.Data {
+		keyID, rawKey, found, err := parseSecretDataKey(key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse secret data key %s: %w", key, err)
+		}
+		if !found {
+			continue
+		}
+		if err := kmsPluginsSecretData.SetFromRawKey(keyID, rawKey, value); err != nil {
+			return nil, fmt.Errorf("failed to set key %s: %w", key, err)
+		}
+	}
+
+	return &Config{Encryption: encryptionConfig, KMSPlugins: kmsPlugins, KMSPluginsSecretData: kmsPluginsSecretData}, nil
 }
 
 func ToSecret(ns, name string, secretData *Config) (*corev1.Secret, error) {
@@ -122,6 +145,15 @@ func ToSecret(ns, name string, secretData *Config) (*corev1.Secret, error) {
 			return nil, err
 		}
 		s.Data[dataKey] = encodedPlugin
+	}
+
+	// Write secret data entries to the encryption-config Secret.
+	// Each entry from FlatEntries() (e.g. "app-role_role-id") is combined with the keyID
+	// (e.g. "1") to produce "kms-plugin-secret-app-role_role-id-1".
+	for keyID, flatEntries := range secretData.KMSPluginsSecretData.FlatEntriesByKeyID() {
+		for flatKey, value := range flatEntries {
+			s.Data[encryptionConfigSecretDataPrefix+flatKey+"-"+keyID] = value
+		}
 	}
 
 	return s, nil
@@ -168,4 +200,20 @@ func ExtractUniqueAndSortedKMSConfigurations(secretData *Config) ([]*apiserverco
 		return iKeyID > jKeyID
 	})
 	return result, nil
+}
+
+func parseSecretDataKey(dataKey string) (keyID, rawKey string, found bool, err error) {
+	rest, found := strings.CutPrefix(dataKey, encryptionConfigSecretDataPrefix)
+	if !found {
+		return "", "", false, nil
+	}
+	i := strings.LastIndex(rest, "-")
+	if i < 1 {
+		return "", "", false, nil
+	}
+	keyID = rest[i+1:]
+	if _, err := strconv.ParseUint(keyID, 10, 64); err != nil {
+		return "", "", false, fmt.Errorf("invalid keyID %q: must be a non-negative integer", keyID)
+	}
+	return keyID, rest[:i], true, nil
 }
--- a/pkg/operator/encryption/secrets/secrets.go
+++ b/pkg/operator/encryption/secrets/secrets.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -76,16 +77,25 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 			// encryption mode.
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig)
 		}
-		if v, ok := s.Data[EncryptionSecretKMSPluginConfig]; ok && len(v) > 0 {
+		if v, ok := s.Data[encryptionSecretKMSPluginConfig]; ok && len(v) > 0 {
 			kmsConfig, err := encoding.DecodeKMSPluginConfig(v)
 			if err != nil {
-				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, EncryptionSecretKMSPluginConfig, err)
+				return state.KeyState{}, fmt.Errorf("secret %s/%s has invalid %s data: %w", s.Namespace, s.Name, encryptionSecretKMSPluginConfig, err)
 			}
 			key.KMS.Plugin = kmsConfig
 		} else {
 			// encryption.apiserver.operator.openshift.io-kms-plugin-config data field is required for KMS
 			// encryption mode.
-			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSPluginConfig)
+			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", encryptionSecretKMSPluginConfig)
+		}
+		for dataKey, value := range s.Data {
+			rawKey, found := strings.CutPrefix(dataKey, encryptionSecretKMSSecretDataPrefix)
+			if !found {
+				continue
+			}
+			if err := key.KMS.PluginSecretData.SetFromRawKey(rawKey, value); err != nil {
+				return state.KeyState{}, fmt.Errorf("secret %s/%s has malformed secret data key %q: %w", s.Namespace, s.Name, dataKey, err)
+			}
 		}
 		key.Mode = keyMode
 	default:
@@ -106,7 +116,7 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 	}
 
 	if ks.Mode == state.KMS && (!ks.HasKMSEncryption() || !ks.HasKMSPlugin()) {
-		return nil, fmt.Errorf("%s or %s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig, EncryptionSecretKMSPluginConfig)
+		return nil, fmt.Errorf("%s or %s can not be empty, when mode is KMS", EncryptionSecretKMSEncryptionConfig, encryptionSecretKMSPluginConfig)
 	}
 
 	s := &corev1.Secret{
@@ -156,7 +166,13 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		if err != nil {
 			return nil, err
 		}
-		s.Data[EncryptionSecretKMSPluginConfig] = pluginData
+		s.Data[encryptionSecretKMSPluginConfig] = pluginData
+	}
+
+	if ks.HasKMSSecretData() {
+		for flatKey, value := range ks.KMS.PluginSecretData.FlatEntries() {
+			s.Data[encryptionSecretKMSSecretDataPrefix+flatKey] = value
+		}
 	}
 
 	return s, nil
--- a/pkg/operator/encryption/secrets/types.go
+++ b/pkg/operator/encryption/secrets/types.go
@@ -54,9 +54,14 @@ const (
 	// encryption configuration for KMS mode in the encryption-key secret.
 	EncryptionSecretKMSEncryptionConfig = "encryption.apiserver.operator.openshift.io-kms-encryption-config"
 
-	// EncryptionSecretKMSPluginConfig is the data field key that stores the serialized KMS plugin
+	// encryptionSecretKMSPluginConfig is the data field key that stores the serialized KMS plugin
 	// configuration for KMS mode in the encryption-key secret.
-	EncryptionSecretKMSPluginConfig = "encryption.apiserver.operator.openshift.io-kms-plugin-config"
+	encryptionSecretKMSPluginConfig = "encryption.apiserver.operator.openshift.io-kms-plugin-config"
+
+	// encryptionSecretKMSSecretDataPrefix is the data field key prefix for secret data values
+	// fetched from the referenced secret in openshift-config. The full data key is
+	// constructed as prefix + secretName + separator + dataKey.
+	encryptionSecretKMSSecretDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-secret-"
 )
 
 // MigratedGroupResources is the data structured stored in the
--- a/pkg/operator/encryption/state/types.go
+++ b/pkg/operator/encryption/state/types.go
@@ -1,6 +1,8 @@
 package state
 
 import (
+	"fmt"
+	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -15,6 +17,11 @@ const (
 	KubernetesDescriptionScaryValue = `WARNING: DO NOT EDIT.
 Altering of the encryption secrets will render you cluster inaccessible.
 Catastrophic data loss can occur from the most minor changes.`
+
+	// secretDataKeySeparator separates the secret name from the data key.
+	// Underscore is used because it is forbidden in Kubernetes secret/configmap
+	// names, preventing collisions.
+	secretDataKeySeparator = "_"
 )
 
 // GroupResourceState represents, for a single group resource, the write and read keys in a
@@ -53,6 +60,10 @@ func (k *KeyState) HasKMSPlugin() bool {
 	return k != nil && k.KMS != nil && k.KMS.Plugin != (configv1.KMSPluginConfig{})
 }
 
+func (k *KeyState) HasKMSSecretData() bool {
+	return k != nil && k.KMS != nil && len(k.KMS.PluginSecretData.Entries) > 0
+}
+
 // KMSState stores all KMS encryption mode related configurations
 type KMSState struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
@@ -60,6 +71,58 @@ type KMSState struct {
 
 	// Plugin stores KMS plugin specific configurations
 	Plugin configv1.KMSPluginConfig
+
+	// PluginSecretData stores data key-value pairs fetched from referenced secrets.
+	PluginSecretData KMSSecretData
+}
+
+// KMSSecretData stores data key-value pairs fetched from referenced secrets.
+// Entries maps secret names to their data key-value pairs.
+type KMSSecretData struct {
+	Entries map[string]map[string][]byte
+}
+
+func (d *KMSSecretData) Set(secretName, dataKey string, value []byte) error {
+	if len(secretName) == 0 || len(dataKey) == 0 || len(value) == 0 {
+		return fmt.Errorf("secretName, dataKey, and value must not be empty")
+	}
+	if strings.Contains(secretName, "_") {
+		return fmt.Errorf("secret name %q must not contain underscores", secretName)
+	}
+	if d.Entries == nil {
+		d.Entries = map[string]map[string][]byte{}
+	}
+	if d.Entries[secretName] == nil {
+		d.Entries[secretName] = map[string][]byte{}
+	}
+	d.Entries[secretName][dataKey] = value
+	return nil
+}
+
+// SetFromRawKey splits a combined key of the form "secretName_dataKey"
+// and stores the value.
+func (d *KMSSecretData) SetFromRawKey(rawKey string, value []byte) error {
+	parts := strings.SplitN(rawKey, secretDataKeySeparator, 2)
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid combined key %q: expected format {secretName}%s{dataKey}", rawKey, secretDataKeySeparator)
+	}
+	return d.Set(parts[0], parts[1], value)
+}
+
+// FlatEntries returns the stored data as a flat map keyed by "secretName_dataKey".
+// "_" separates secretName from dataKey because "_" is forbidden in
+// Kubernetes secret names, making the split unambiguous.
+func (d *KMSSecretData) FlatEntries() map[string][]byte {
+	if d.Entries == nil {
+		return nil
+	}
+	result := map[string][]byte{}
+	for secretName, keys := range d.Entries {
+		for dataKey, value := range keys {
+			result[secretName+secretDataKeySeparator+dataKey] = value
+		}
+	}
+	return result
 }
 
 type MigrationState struct {
--- a/pkg/operator/encryption/state/types_test.go
+++ b/pkg/operator/encryption/state/types_test.go
@@ -1,0 +1,181 @@
+package state
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestKMSSecretDataSet(t *testing.T) {
+	tests := []struct {
+		name       string
+		secretName string
+		dataKey    string
+		value      []byte
+		wantErr    bool
+		expected   KMSSecretData
+	}{
+		{
+			name:       "valid set",
+			secretName: "vault-approle-secret",
+			dataKey:    "role-id",
+			value:      []byte("test-role-id"),
+			expected: KMSSecretData{Entries: map[string]map[string][]byte{
+				"vault-approle-secret": {"role-id": []byte("test-role-id")},
+			}},
+		},
+		{
+			name:       "secret name with underscore returns error",
+			secretName: "vault_approle_secret",
+			dataKey:    "role-id",
+			value:      []byte("test-role-id"),
+			wantErr:    true,
+		},
+		{
+			name:       "secret name with trailing underscore returns error",
+			secretName: "vault-approle-secret_",
+			dataKey:    "role-id",
+			value:      []byte("test-role-id"),
+			wantErr:    true,
+		},
+		{
+			name:       "secret name with leading underscore returns error",
+			secretName: "_vault-approle-secret",
+			dataKey:    "role-id",
+			value:      []byte("test-role-id"),
+			wantErr:    true,
+		},
+		{
+			name:    "empty secret name returns error",
+			dataKey: "role-id",
+			value:   []byte("test-role-id"),
+			wantErr: true,
+		},
+		{
+			name:       "empty data key returns error",
+			secretName: "vault-approle-secret",
+			value:      []byte("test-role-id"),
+			wantErr:    true,
+		},
+		{
+			name:       "empty value returns error",
+			secretName: "vault-approle-secret",
+			dataKey:    "role-id",
+			value:      []byte{},
+			wantErr:    true,
+		},
+		{
+			name:       "nil value returns error",
+			secretName: "vault-approle-secret",
+			dataKey:    "role-id",
+			wantErr:    true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var d KMSSecretData
+			err := d.Set(tt.secretName, tt.dataKey, tt.value)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("expected error, got nil")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if diff := cmp.Diff(tt.expected, d); diff != "" {
+				t.Errorf("unexpected result (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestKMSSecretDataSetFromRawKey(t *testing.T) {
+	tests := []struct {
+		name     string
+		rawKey   string
+		value    []byte
+		wantErr  bool
+		expected KMSSecretData
+	}{
+		{
+			name:   "valid raw key",
+			rawKey: "vault-approle-secret_role-id",
+			value:  []byte("test-role-id"),
+			expected: KMSSecretData{Entries: map[string]map[string][]byte{
+				"vault-approle-secret": {"role-id": []byte("test-role-id")},
+			}},
+		},
+		{
+			name:    "missing separator returns error",
+			rawKey:  "no-separator-here",
+			value:   []byte("v"),
+			wantErr: true,
+		},
+		{
+			name:   "multiple underscores splits on first only",
+			rawKey: "vault_approle_secret_role-id",
+			value:  []byte("v"),
+			expected: KMSSecretData{Entries: map[string]map[string][]byte{
+				"vault": {"approle_secret_role-id": []byte("v")},
+			}},
+		},
+		{
+			name:    "empty raw key returns error",
+			rawKey:  "",
+			value:   []byte("v"),
+			wantErr: true,
+		},
+		{
+			name:    "empty value returns error",
+			rawKey:  "vault-approle-secret_role-id",
+			value:   []byte{},
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var d KMSSecretData
+			err := d.SetFromRawKey(tt.rawKey, tt.value)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("expected error, got nil")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if diff := cmp.Diff(tt.expected, d); diff != "" {
+				t.Errorf("unexpected result (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestKMSSecretDataFlatEntries(t *testing.T) {
+	d := KMSSecretData{Entries: map[string]map[string][]byte{
+		"vault-approle-secret": {
+			"role-id":   []byte("test-role-id"),
+			"secret-id": []byte("test-secret-id"),
+		},
+	}}
+
+	expected := map[string][]byte{
+		"vault-approle-secret_role-id":   []byte("test-role-id"),
+		"vault-approle-secret_secret-id": []byte("test-secret-id"),
+	}
+
+	if diff := cmp.Diff(expected, d.FlatEntries()); diff != "" {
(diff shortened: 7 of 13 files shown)


ardaguclu:
/retitle [CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Carry KMS credentials from referenced Secret

pkg/operator/encryption/kms/credentials.go
```diff
@@ -0,0 +1,48 @@
...
+func fetchVaultCredentials(ctx context.Context, secretClient corev1client.SecretsGetter, plugin configv1.KMSPluginConfig) (map[string]string, error) {
+	secretName := plugin.Vault.Authentication.AppRole.Secret.Name
+	if len(secretName) == 0 {
+		return nil, fmt.Errorf("vault AppRole secret reference name is empty")
+	}
+
+	credSecret, err := secretClient.Secrets("openshift-config").Get(ctx, secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get credential secret %s in openshift-config: %w", secretName, err)
+	}
```
bertinatto: instead of returning an error right away, should we give the user time to create the secret, once they set the KMS fields in the api?

pkg/operator/encryption/kms/credentials.go
```diff
@@ -0,0 +1,48 @@
...
+
+	credentials := map[string]string{}
+	for _, key := range vaultAppRoleRequiredKeys {
+		value, ok := credSecret.Data[key]
+		if !ok {
+			return nil, fmt.Errorf("credential secret %s in openshift-config is missing required key %q", secretName, key)
+		}
```
bertinatto: yeah, if the secret is malformed, we should probably error out

pkg/operator/encryption/kms/credentials.go
```diff
@@ -0,0 +1,48 @@
...
+		credentials[secrets.EncryptionSecretKMSCredentialPrefix+"-"+secretName+"-"+key] = string(value)
```
bertinatto: the credentials secret is not subject to the validation done at admission time, like the KMS fields in the `cluster/apiserver` CR. I think we should probably validate that the user is setting valid values for these 2 secret keys, wdyt?

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,22 @@ func (k *KeyState) HasKMSPlugin() bool {
 	return k != nil && k.KMS != nil && k.KMS.Plugin != (configv1.KMSPluginConfig{})
 }
 
+func (k *KeyState) HasKMSCredentials() bool {
+	return k != nil && k.KMS != nil && len(k.KMS.Credentials) > 0
```
bertinatto: for vault this would actually need to be 2, but I think it's fine as is because you're currently checking that in `fetchVaultCredentials`

pkg/operator/encryption/kms/credentials.go
```diff
@@ -0,0 +1,48 @@
...
+
+var vaultAppRoleRequiredKeys = []string{"role-id", "secret-id"}
+
+func fetchVaultCredentials(ctx context.Context, secretClient corev1client.SecretsGetter, plugin configv1.KMSPluginConfig) (map[string]string, error) {
+	secretName := plugin.Vault.Authentication.AppRole.Secret.Name
+	if len(secretName) == 0 {
+		return nil, fmt.Errorf("vault AppRole secret reference name is empty")
+	}
+
+	credSecret, err := secretClient.Secrets("openshift-config").Get(ctx, secretName, metav1.GetOptions{})
```
bertinatto: this secret shouldn't change a lot, so I wonder if we should use a cached client (lister) instead

pkg/operator/encryption/kms/credentials.go
```diff
@@ -3,13 +3,55 @@ package kms
...
+// form used by Config.KMSCredentials and the encryption-config Secret by
+// stripping the "encryption.apiserver.operator.openshift.io-kms-secret-" prefix.
+func ToShortCredentialKeys(credentials map[string]string) map[string]string {
+	short := make(map[string]string, len(credentials))
+	for credKey, value := range credentials {
+		short[strings.TrimPrefix(credKey, secrets.EncryptionSecretKMSCredentialPrefix+"-")] = value
+	}
+	return short
+}
+
```
bertinatto: nit: maybe keep these functions together with their plugin-config counterparts in helpers.go

pkg/operator/encryption/kms/credentials.go
ardaguclu: I think, it is prerequisite for cluster-admin to create secret beforehand. We should degrade if it is not found. This error blocks the new key creation. So this won't impact on cluster's stability.

pkg/operator/encryption/kms/credentials.go
ardaguclu: Key controller can not have semantic knowledge about what values should be. Preflight checker will validate this.

pkg/operator/encryption/state/types.go
ardaguclu: yeah, I wanted to keep here generic.

pkg/operator/encryption/kms/credentials.go
ardaguclu: Good idea. Let me update it.

ardaguclu: However, this requires adding `openshift-config` to the key controller's informer watch and store `SecretLister` in keyController struct. If you are ok with this signature change, I can update.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -67,6 +73,21 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
+			if key.HasKMSCredentials() {
+				if kmsCredentials == nil {
+					kmsCredentials = map[string]map[string]string{}
+				}
+				// Convert from Key Secret format (e.g. "encryption.apiserver.operator.openshift.io-kms-secret-app-role-role-id")
+				// to the short form used by Config and the encryption-config Secret (e.g. "app-role-role-id").
+				shortCreds := kms.ToShortCredentialKeys(key.KMS.Credentials)
```
ardaguclu: I'm not happy with this conversion but it is required to convert `encryption.apiserver.operator.openshift.io-kms-secret-app-role-role-id` to `kms-secret-app-role-role-id`

ardaguclu: Maybe we should break the convention and use `kms-secret-app-role-role-id` in everywhere including the key Secrets.

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,22 @@ func (k *KeyState) HasKMSPlugin() bool {
...
 	// Encoded EncryptionConfig that stores the KMS related fields
 	Encryption *apiserverconfigv1.KMSConfiguration
 
 	// Plugin stores KMS plugin specific configurations
 	Plugin configv1.KMSPluginConfig
+
+	// Credentials stores fetched credential values from the referenced secret
+	// in openshift-config. Keys use the format
+	// "encryption.apiserver.operator.openshift.io-kms-secret-{secretName}-{keyName}".
+	Credentials map[string]string
```
p0lyn0mial: wait, didn't we want to store Secrets/ConfigMap data fields ? xref: https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/kms-encryption-foundations.md#proposal

p0lyn0mial: why not to make it more generic:
```
SecretData map[string][]byte
```

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +282,12 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 			},
 			Plugin: apiServerEncryption.KMS,
 		}
+
+		credentials, err := kms.FetchCredentials(ctx, c.secretClient, apiServerEncryption.KMS)
```
p0lyn0mial: by making the data struct more generic e.g. `SecretData` we make this code more generic too:

```
if secretName := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
	refSecret, err := c.secretClient.Secrets("openshift-config").Get(ctx, secretName, metav1.GetOptions{})
	...
	sd := make(map[string][]byte, len(refSecret.Data))
	for k, v := range refSecret.Data {
		sd[k] = v
	}
	ks.KMS.SecretData = sd
}
```

p0lyn0mial: where `referencedSecretName` could be private and simple:

```
func referencedSecretName(plugin configv1.KMSPluginConfig) string {
	switch plugin.Type {
	case configv1.VaultKMSProvider:
		return plugin.Vault.Authentication.AppRole.Secret.Name
	default:
		return ""
	}
}
```

p0lyn0mial: do we already have a const for `"openshift-config"` namespace ?

pkg/operator/encryption/kms/credentials.go
```diff
@@ -0,0 +1,90 @@
+package kms
```
p0lyn0mial: mhm, not sure if we need a new file, seems all methods could be private and close to the source.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -30,6 +31,10 @@ type Config struct {
 	// KMSPlugins maps keyID to plugin-specific configuration,
 	// carried from Key Secrets into the encryption-config Secret.
 	KMSPlugins map[string]configv1.KMSPluginConfig
+	// KMSCredentials maps keyID to credential data carried from
+	// Key Secrets into the encryption-config Secret.
+	// Inner map keys use the short form (e.g. "app-role-role-id").
+	KMSCredentials map[string]map[string]string
```
p0lyn0mial: `KMSSecretData map[string]map[string][]byte`

pkg/operator/encryption/secrets/types.go
```diff
@@ -57,6 +57,11 @@ const (
 	// EncryptionSecretKMSPluginConfig is the data field key that stores the serialized KMS plugin
 	// configuration for KMS mode in the encryption-key secret.
 	EncryptionSecretKMSPluginConfig = "encryption.apiserver.operator.openshift.io-kms-plugin-config"
+
+	// EncryptionSecretKMSCredentialPrefix is the data field key prefix for credential values
+	// fetched from the referenced secret in openshift-config. The full data key is
+	// constructed by appending "-{secretName}-{dataKey}" to this prefix.
+	EncryptionSecretKMSCredentialPrefix = "encryption.apiserver.operator.openshift.io-kms-secret"
```
p0lyn0mial: maybe this should be `.....kms-plugin-secret"` to align with `"encryption.apiserver.operator.openshift.io-kms-plugin-config"`

pkg/operator/encryption/controllers/key_controller.go
p0lyn0mial: it seems like we miss some dynamic validation, for example we should go degraded when the src secret is removed but referenced.

p0lyn0mial: seems like we miss an informer that would trigger the sync loop.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -87,6 +88,14 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 			// encryption mode.
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSPluginConfig)
 		}
+		for dataKey, value := range s.Data {
+			if strings.HasPrefix(dataKey, EncryptionSecretKMSCredentialPrefix) {
```
p0lyn0mial: i'd like to experiment a bit with how we store data here, e.g. with our without the prefix.

p0lyn0mial (commented):
I did a first pass and left a few comments. Let me know what you think. Thanks.

pkg/operator/encryption/kms/credentials.go
ardaguclu: I think we agreed that we will use secretClient?

pkg/operator/encryption/state/types.go
ardaguclu: That is correct. Credentials store key-value map in KMSState which will be converted to each data field.

ardaguclu:
/retitle [CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Carry referenced Secret data from API to plugin lifecycle for KMS

pkg/operator/encryption/state/types.go
ardaguclu: Updated

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated with one difference. We can not carry all the data fields in the given Secret. We need to carry only the selective set of data keys. So `referencedSecretName` function will also return an array for this. Please let me know your thoughts.

pkg/operator/encryption/kms/credentials.go
ardaguclu: Updated.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Tracking the current secret references, keeping their content up-to-date, degrading if they are removed will be handled in another PR. This PR only focuses on carrying this data. If Secret is not present, or one one of the selected set of data fields are missing, we will return error (I assume that means we'll degrade).

ardaguclu: Added a TODO for informer.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Sounds good to me. Thanks for spending time on this and review.

test/e2e-encryption/encryption_test.go
```diff
@@ -305,6 +305,41 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+
+			// Verify actual values match the source secret
+			require.Equal(t, "test-role-id", secretData["vault-approle-secret-role-id"], "role-id secret data mismatch for keyID %s", keyID)
+			require.Equal(t, "test-secret-id", secretData["vault-approle-secret-secret-id"], "secret-id secret data mismatch for keyID %s", keyID)
+
+			// Verify Key Secret also carries the secret data
+			keySecret, err := kubeClient.CoreV1().Secrets("openshift-config-managed").Get(ctx, fmt.Sprintf("encryption-key-%s-%s", component, keyID), metav1.GetOptions{})
+			require.NoError(t, err)
+			require.Equal(t, "test-role-id", string(keySecret.Data[secrets.EncryptionSecretKMSSecretDataPrefix+"-vault-approle-secret-role-id"]), "key secret %s role-id secret data mismatch", keyID)
+			require.Equal(t, "test-secret-id", string(keySecret.Data[secrets.EncryptionSecretKMSSecretDataPrefix+"-vault-approle-secret-secret-id"]), "key secret %s secret-id secret data mismatch", keyID)
```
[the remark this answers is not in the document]
ardaguclu: Nice catch. Updated.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in openshift-config: %w", secretName, err)
+			}
+			sd := make(map[string][]byte, len(expectedKeys))
+			for _, key := range expectedKeys {
+				v, ok := refSecret.Data[key]
+				if !ok {
+					return nil, fmt.Errorf("secret %s in openshift-config is missing required key %q", secretName, key)
+				}
+				sd[secretName+"-"+key] = v
```
p0lyn0mial: what if `secretName` already contains `-` ? is this a problem ?
what if `key` already contains `-` ? is this a problem ?

p0lyn0mial: what if:

secretName="vault-approle", key="secret-role-id" gives "vault-approle-secret-role-id"
secretName="vault-approle-secret", key="role-id" gives "vault-approle-secret-role-id"

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -383,6 +401,15 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 	return latestKeyID, "rotation-interval-has-passed", time.Since(latestKey.Migrated.Timestamp) > encryptionSecretMigrationInterval
 }
 
+func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
+	switch plugin.Type {
+	case configv1.VaultKMSProvider:
+		return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
```
p0lyn0mial: it is important that we document that we only care respect []string{"role-id", "secret-id"}

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -116,7 +116,9 @@ func NewKeyController(
 		WithInformers(
 			apiServerInformer.Informer(),
 			operatorClient.Informer(),
+
 			kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
+			// TODO: add informer for openshift-config namespace to watch referenced Secrets for KMS plugin secret data changes
```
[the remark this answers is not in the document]
ardaguclu: Informer will be added in a followup PR.

test/e2e-encryption/encryption_test.go
```diff
@@ -305,6 +305,41 @@ func TestEncryptionIntegration(tt *testing.T) {
...
+			}
+		}
+
+		for keyID := range expectedKeyIDs {
+			secretData, ok := cfg.KMSSecretData[keyID]
+			require.True(t, ok, "expected secret data for keyID %s in encryption-config secret", keyID)
+			require.NotEmpty(t, secretData, "expected non-empty secret data for keyID %s", keyID)
+
+			// Verify actual values match the source secret
+			require.Equal(t, "test-role-id", secretData["vault-approle-secret-role-id"], "role-id secret data mismatch for keyID %s", keyID)
```
p0lyn0mial: does the comparison work ? we are comparing []byte with string. i guess it works because in go string actually is []byte

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			refSecret, err := c.secretClient.Secrets("openshift-config").Get(ctx, secretName, metav1.GetOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in openshift-config: %w", secretName, err)
+			}
+			sd := make(map[string][]byte, len(expectedKeys))
+			for _, key := range expectedKeys {
+				v, ok := refSecret.Data[key]
+				if !ok {
+					return nil, fmt.Errorf("secret %s in openshift-config is missing required key %q", secretName, key)
```
p0lyn0mial: worth adding unit tests to covert his scenario ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -67,6 +72,18 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
+			if key.HasKMSSecretData() {
+				if kmsSecretData == nil {
+					kmsSecretData = map[string]map[string][]byte{}
+				}
+				if existing, exists := kmsSecretData[key.Key.Name]; exists {
```
p0lyn0mial: do we expect duplicates here ? why ?

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial, line 104: same question here, do we expect duplicates here ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -30,6 +30,10 @@ type Config struct {
 	// KMSPlugins maps keyID to plugin-specific configuration,
 	// carried from Key Secrets into the encryption-config Secret.
 	KMSPlugins map[string]configv1.KMSPluginConfig
+	// KMSSecretData maps keyID to secret data carried from
+	// Key Secrets into the encryption-config Secret.
+	// Inner map keys use the short form (e.g. "app-role-role-id").
+	KMSSecretData map[string]map[string][]byte
```
p0lyn0mial: maybe we should rename to `KMSPluginsSecretData` ?

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: if we don't expect duplicates then maybe we check if there are none.

in either case we should add tests.

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: if we don't expect duplicates then maybe we check if there are none.

in either case we should add tests.

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -56,7 +61,29 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*Config, error) {
 		kmsPlugins[keyID] = pluginConfig
 	}
 
-	return &Config{Encryption: encryptionConfig, KMSPlugins: kmsPlugins}, nil
+	// Extract secret data entries from the encryption-config Secret.
+	// Data keys follow the format "kms-plugin-secret-{secretDataKey}-{keyID}"
+	// (e.g. "kms-plugin-secret-app-role-role-id-1"). KeyIDFromSecretDataKey
+	// returns the keyID (e.g. "1") and the secretDataKey (e.g. "app-role-role-id").
+	var kmsSecretData map[string]map[string][]byte
+	for key, value := range encryptionConfigSecret.Data {
```
p0lyn0mial, line 93: do we have tests that cover changes in this file ?

p0lyn0mial: seems like not everything is covered.

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -100,6 +127,19 @@ func ToSecret(ns, name string, secretData *Config) (*corev1.Secret, error) {
 		s.Data[dataKey] = encodedPlugin
 	}
 
+	// Write secret data entries to the encryption-config Secret.
+	// Each secret dataKey (e.g. "app-role-role-id") is combined with the keyID
```
p0lyn0mial: what if the keyID=123 ?

p0lyn0mial: what if dataKey=token-123 ?

p0lyn0mial: please add unit test to cover all edge cases.

pkg/operator/encryption/controllers/key_controller.go
p0lyn0mial: in the future we might ref more than one secret.

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -100,6 +127,19 @@ func ToSecret(ns, name string, secretData *Config) (*corev1.Secret, error) {
...
+	// (e.g. "1") to produce the data key "kms-plugin-secret-app-role-role-id-1".
+	for keyID, keySecretData := range secretData.KMSSecretData {
+		for secretDataKey, value := range keySecretData {
+			dataKey, err := toSecretDataKeyFor(secretDataKey, keyID)
+			if err != nil {
+				return nil, err
+			}
+			s.Data[dataKey] = value
```
p0lyn0mial: if we have the same dataKey from two different secrets then I think we have a collision here.

p0lyn0mial (commented):
did another pass, haven't finished. seems like there is a potential of a collision when there is more than one secret in the future ....

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -17,11 +18,15 @@ import (
...
-// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
-const EncryptionConfSecretKey = "encryption-config"
+const (
+	// EncryptionConfSecretName is the name of the final encryption config secret that is revisioned per apiserver rollout.
+	EncryptionConfSecretName = "encryption-config"
+	// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
+	EncryptionConfSecretKey = "encryption-config"
+	// encryptionConfigSecretDataPrefix is the data key prefix for KMS plugin secret
+	// data entries in the encryption-config Secret. Full key: "kms-plugin-secret-{secretDataKey}-{keyID}".
+	encryptionConfigSecretDataPrefix = "kms-plugin-secret-"
```
p0lyn0mial, line 51: worth updating the KEP.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: From the encryption controller's perspective this is not a problem. It concatenates secretName and key and propagates to the plugin lifecycle as is. 

But this collision can be a problem in plugin lifecycle side. I'd like to discuss more.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: WDYM?. It will be `kms-plugin-secret-app-role-role-id-123`?

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: We expect duplicates. Because each resource will carry their own kms-plugin. Secret's kms-plugin must match to configmap's kms-plugin. 

I agree that we should cover this in unit tests.

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Yes, we expect duplicates. `TestFromEncryptionStateKMSPluginConfigValidation` function covers this behavior.

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,21 @@ func (k *KeyState) HasKMSPlugin() bool {
...
 type KMSState struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
 	Encryption *apiserverconfigv1.KMSConfiguration
 
 	// Plugin stores KMS plugin specific configurations
 	Plugin configv1.KMSPluginConfig
+
+	// SecretData stores fetched values from the referenced secret
+	// in openshift-config. Keys use the format "{secretName}-{keyName}".
+	SecretData map[string][]byte
```
p0lyn0mial: maybe this should be:
```
SecretData map[string]map[string][]byte  // secretName to dataKey to value
```

ardaguclu:
/hold
until collision solution is agreed and implemented

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: "degraded when KMS referenced secret is missing a required key" test covers this scenario.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Documented.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: More tests are added under config_test.go

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We have decided to pursue in that format;

```
kms-plugin-secret-{secretName}_{dataKey}-{keyID}
encryption.apiserver.operator.openshift.io-kms-plugin-secret-{secretName}_{dataKey}
```

So that PR is updated based on this.

ardaguclu: We have fixed that problem in https://github.com/openshift/library-go/pull/2212#discussion_r3257158958

test/e2e-encryption/encryption_test.go
ardaguclu: Correct. Fixed.

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Renamed.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: More tests are added as well for the edge cases (e.g. app-role_role-id-3-5 extracts 5 as keyID)

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: I think current state of this PR has fixed that collison problem.

pkg/operator/encryption/state/types.go
ardaguclu: KeyState stores SecretData map[string]map[string][]byte: secretName -> dataKey -> data
Config stores SecretData map[string]map[string]map[string][]byte: keyID -> secretName -> dataKey -> data

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 			},
 			Plugin: apiServerEncryption.KMS,
 		}
+
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			refSecret, err := c.secretClient.Secrets(openshiftConfigNS).Get(ctx, secretName, metav1.GetOptions{})
```
p0lyn0mial, line 290: validate if the `secretName` doesn't contain `_`

pkg/operator/encryption/encryptiondata/secret.go
p0lyn0mial: ping

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -67,6 +72,18 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
+			if key.HasKMSSecretData() {
+				if kmsPluginsSecretData == nil {
+					kmsPluginsSecretData = map[string]map[string]map[string][]byte{}
```
p0lyn0mial: I think that this only creates an empty top-level map. I think that inner maps are still `nil`. Are we missing unit tests ?

pkg/operator/encryption/secrets/types.go
```diff
@@ -57,6 +57,16 @@ const (
...
+
+	// EncryptionSecretKMSSecretDataPrefix is the data field key prefix for secret data values
+	// fetched from the referenced secret in openshift-config. The full data key is
+	// constructed as prefix + secretName + separator + dataKey.
+	EncryptionSecretKMSSecretDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-secret-"
+
+	// SecretDataKeySeparator separates the secret name from the data key in
+	// Key Secret data keys. Underscore is used because it is forbidden in
+	// Kubernetes secret/configmap names, preventing collisions.
+	SecretDataKeySeparator = "_"
```
p0lyn0mial: does it have to be exported ?

pkg/operator/encryption/secrets/types.go
```diff
@@ -57,6 +57,16 @@ const (
 	// EncryptionSecretKMSPluginConfig is the data field key that stores the serialized KMS plugin
 	// configuration for KMS mode in the encryption-key secret.
 	EncryptionSecretKMSPluginConfig = "encryption.apiserver.operator.openshift.io-kms-plugin-config"
+
+	// EncryptionSecretKMSSecretDataPrefix is the data field key prefix for secret data values
+	// fetched from the referenced secret in openshift-config. The full data key is
+	// constructed as prefix + secretName + separator + dataKey.
+	EncryptionSecretKMSSecretDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-secret-"
```
p0lyn0mial: wondering if we should expose that one too ?

pkg/operator/encryption/secrets/types.go
```diff
@@ -57,6 +57,16 @@ const (
 	// EncryptionSecretKMSPluginConfig is the data field key that stores the serialized KMS plugin
 	// configuration for KMS mode in the encryption-key secret.
 	EncryptionSecretKMSPluginConfig = "encryption.apiserver.operator.openshift.io-kms-plugin-config"
```
p0lyn0mial: wondering if we should expose that one too ?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -87,6 +88,24 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
...
 			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSPluginConfig)
 		}
+		for dataKey, value := range s.Data {
+			rawKey, found := strings.CutPrefix(dataKey, EncryptionSecretKMSSecretDataPrefix)
+			if !found || len(rawKey) == 0 {
+				continue
+			}
+			parts := strings.SplitN(rawKey, SecretDataKeySeparator, 2)
+			if len(parts) != 2 {
+				continue
```
p0lyn0mial: shouldn't we return an err ? at this point the prefix was found, no ?

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -14,14 +15,19 @@ import (
...
-// EncryptionConfSecretName is the name of the final encryption config secret that is revisioned per apiserver rollout.
-const EncryptionConfSecretName = "encryption-config"
-
-// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
-const EncryptionConfSecretKey = "encryption-config"
+const (
+	// EncryptionConfSecretName is the name of the final encryption config secret that is revisioned per apiserver rollout.
+	EncryptionConfSecretName = "encryption-config"
+	// EncryptionConfSecretKey is the map data key used to store the raw bytes of the final encryption config.
+	EncryptionConfSecretKey = "encryption-config"
```
p0lyn0mial, line 48: does it have to be exported ?

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: ok, thanks, i will have a look.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -159,6 +178,16 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Data[EncryptionSecretKMSPluginConfig] = pluginData
 	}
 
+	if ks.HasKMSSecretData() {
+		for secretName, secretData := range ks.KMS.SecretData {
+			for dataKey, value := range secretData {
+				// Write referenced secret data to the Key Secret using the format:
+				// "encryption.apiserver.operator.openshift.io-kms-plugin-secret-{secretName}_{dataKey}"
```
p0lyn0mial: we had a long discussion about it. Worth document "why" we are doing this.

p0lyn0mial (commented):
did another pass. there are a few things i would still like to check.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: https://github.com/openshift/enhancements/pull/2014

pkg/operator/encryption/secrets/secrets.go
ardaguclu: yes, we should return error

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in openshift-config: %w", secretName, err)
+			}
+			sd := make(map[string][]byte, len(expectedKeys))
+			for _, key := range expectedKeys {
```
p0lyn0mial, line 294: how about:
```
for _, key := range expectedKeys {
	v, ok := refSecret.Data[key]
	if !ok {
		return nil, fmt.Errorf("secret %s in openshift-config is missing required key %q", secretName, key)
	}
	ks.KMS.SecretData.Set(secretName, key, v)
}
```
?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -87,6 +88,24 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
...
+				continue
+			}
+			secretName, secretKey := parts[0], parts[1]
+			if key.KMS.SecretData == nil {
+				key.KMS.SecretData = map[string]map[string][]byte{}
+			}
+			if key.KMS.SecretData[secretName] == nil {
+				key.KMS.SecretData[secretName] = map[string][]byte{}
+			}
+			key.KMS.SecretData[secretName][secretKey] = value
```
p0lyn0mial: I think this could be replaced by `key.KMS.SecretData.Set(...)`

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,21 @@ func (k *KeyState) HasKMSPlugin() bool {
...
 type KMSState struct {
 	// Encoded EncryptionConfig that stores the KMS related fields
 	Encryption *apiserverconfigv1.KMSConfiguration
 
 	// Plugin stores KMS plugin specific configurations
 	Plugin configv1.KMSPluginConfig
+
+	// SecretData stores fetched values from the referenced secret
+	// in openshift-config. Outer key is the secret name, inner key is the data key.
+	SecretData map[string]map[string][]byte
```
p0lyn0mial: `PluginSecretData KMSSecretData`


```
// KMSSecretData stores data key-value pairs fetched from referenced secrets.
// Entries maps secret names to their data key-value pairs.
type KMSSecretData struct {
	Entries map[string]map[string][]byte
}

func (d *KMSSecretData) Set(secretName, dataKey string, value []byte) {
	if d.Entries == nil {
		d.Entries = map[string]map[string][]byte{}
	}
	if d.Entries[secretName] == nil {
		d.Entries[secretName] = map[string][]byte{}
	}
	d.Entries[secretName][dataKey] = value
}
```

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Inner mappings are directly set by KeyState;

```go
kmsPluginsSecretData[key.Key.Name] = key.KMS.SecretData
```

pkg/operator/encryption/secrets/types.go
ardaguclu: `secrets/secrets.go` requires it to generate dataKey. `encryptiondata/secret.go` requires it to parse this dataKey.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: yes, it is used by `observe_encryption_config.go`

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: I've updated kmsPluginsSecretData to `kmsPluginsSecretData = map[string]state.KMSSecretData{}`.

pkg/operator/encryption/secrets/types.go
ardaguclu: Marked as private

pkg/operator/encryption/secrets/types.go
ardaguclu: Marked as private

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Updated to return error

pkg/operator/encryption/secrets/secrets.go
ardaguclu: Briefly documented the reasons behind this concatenation.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: That is better. Updated.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: That is better. Updated.

pkg/operator/encryption/state/types.go
ardaguclu: Good idea. I've added a `Get` function too. 

Furthermore, I've used this struct in Config struct -- `KMSPluginsSecretData map[string]state.KMSSecretData` . Thanks to that, we get rid of nested maps. Please let me know your thoughts.

pkg/operator/encryption/secrets/types.go
ardaguclu: key_controller uses it to validate the SecretName.

ardaguclu: After introducing helper functions in secrets.go (join, split, validate). I am able to unexport this variable.

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -145,3 +193,26 @@ func ExtractUniqueAndSortedKMSConfigurations(secretData *Config) ([]*apiserverco
 	})
 	return result, nil
 }
+
+func toSecretDataKeyFor(secretDataKey, keyID string) (string, error) {
```
p0lyn0mial: could this function accept `(secretName, dataKey, keyID)` ?

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -145,3 +193,26 @@ func ExtractUniqueAndSortedKMSConfigurations(secretData *Config) ([]*apiserverco
...
+	if _, err := strconv.ParseUint(keyID, 10, 64); err != nil {
+		return "", fmt.Errorf("invalid keyID %q: must be a non-negative integer", keyID)
+	}
+	return encryptionConfigSecretDataPrefix + secretDataKey + "-" + keyID, nil
+}
+
+func keyIDFromSecretDataKey(dataKey string) (string, string, bool, error) {
```
p0lyn0mial: could this function return `keyID, secretName, dataKey, found, and error` ?
maybe we should add `parse` in the name then ?

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,43 @@ func (k *KeyState) HasKMSPlugin() bool {
...
+	PluginSecretData KMSSecretData
+}
+
+// KMSSecretData stores data key-value pairs fetched from referenced secrets.
+// Entries maps secret names to their data key-value pairs.
+type KMSSecretData struct {
+	Entries map[string]map[string][]byte
+}
+
+func (d *KMSSecretData) Get() map[string]map[string][]byte {
```
p0lyn0mial: mhm, why do we need this method ? 
we do expose `Entries` to be used in `range` statements.

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +53,43 @@ func (k *KeyState) HasKMSPlugin() bool {
 	return k != nil && k.KMS != nil && k.KMS.Plugin != (configv1.KMSPluginConfig{})
 }
 
+func (k *KeyState) HasKMSSecretData() bool {
+	return k != nil && k.KMS != nil && len(k.KMS.PluginSecretData.Get()) > 0
```
p0lyn0mial: use `Entries`

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,23 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 			},
 			Plugin: apiServerEncryption.KMS,
 		}
+
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			if err := secrets.ValidateSecretDataKey(secretName); err != nil {
```
p0lyn0mial: could we move this check directly to `ks.KMS.PluginSecretData.Set` :) ?

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: Good idea. Updated.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: Good idea. Updated.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -30,6 +30,10 @@ type Config struct {
 	// KMSPlugins maps keyID to plugin-specific configuration,
 	// carried from Key Secrets into the encryption-config Secret.
 	KMSPlugins map[string]configv1.KMSPluginConfig
+	// KMSPluginsSecretData maps keyID to secret data carried from
+	// Key Secrets into the encryption-config Secret.
+	// Structure: keyID → secretName → dataKey → value.
+	KMSPluginsSecretData map[string]state.KMSSecretData
```
p0lyn0mial: would:

```
type Config struct {
	Encryption *apiserverconfigv1.EncryptionConfiguration
	// KMSPlugins maps keyID to plugin-specific configuration,
	// carried from Key Secrets into the encryption-config Secret.
	KMSPlugins map[string]configv1.KMSPluginConfig
	// KMSPluginsSecretData maps keyID to secret data carried from
	// Key Secrets into the encryption-config Secret.
	// Structure: keyID → secretName → dataKey → value.
	KMSPluginsSecretData KMSPluginsSecretData
}

// KMSPluginsSecretData maps keyID to secret data carried from Key Secrets into
// the encryption-config Secret. Structure: keyID → secretName → dataKey → value.
type KMSPluginsSecretData struct {
	ByKeyID map[string]state.KMSSecretData
}

// SetFromCombinedKey stores a value for the given keyID, splitting combinedKey
// on "_" into secretName and dataKey.
func (d *KMSPluginsSecretData) SetFromCombinedKey(keyID, combinedKey string, value []byte) error {
	if d.ByKeyID == nil {
		d.ByKeyID = map[string]state.KMSSecretData{}
	}
	sd := d.ByKeyID[keyID]
	if err := sd.SetFromCombinedKey(combinedKey, value); err != nil {
		return err
	}
	d.ByKeyID[keyID] = sd
	return nil
}

// FlatEntriesByKeyID returns the stored data as a map of keyID to flat entries,
// where each flat entry is keyed by "secretName_dataKey".
func (d *KMSPluginsSecretData) FlatEntriesByKeyID() map[string]map[string][]byte {
	result := map[string]map[string][]byte{}
	for keyID, sd := range d.ByKeyID {
		result[keyID] = sd.FlatEntries()
	}
	return result
}
```

make sense ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -67,6 +102,18 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 					kmsPlugins[key.Key.Name] = key.KMS.Plugin
 				}
 			}
+			if key.HasKMSSecretData() {
```
p0lyn0mial, line 113:
```
if key.HasKMSSecretData() {
				if existing, exists := kmsPluginsSecretData.ByKeyID[key.Key.Name]; exists {
					if !equality.Semantic.DeepEqual(existing.FlatEntries(), key.KMS.PluginSecretData.FlatEntries()) {
						return nil, fmt.Errorf("KMS secret data mismatch for keyID %s: secret data from different resources must be identical", key.Key.Name)
					}
				} else {
					for rawKey, value := range key.KMS.PluginSecretData.FlatEntries() {
						if err := kmsPluginsSecretData.SetFromRawKey(key.Key.Name, rawKey, value); err != nil {
							return nil, fmt.Errorf("failed to copy secret data for keyID %s: %w", key.Key.Name, err)
						}
					}
				}
			}
```

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -145,3 +178,19 @@ func ExtractUniqueAndSortedKMSConfigurations(secretData *Config) ([]*apiserverco
 	})
 	return result, nil
 }
+
+func parseSecretDataKey(dataKey string) (keyID, rawKey string, found bool, err error) {
```
p0lyn0mial, line 205: do we have unit tests for this function ?

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Nicer. Added.

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: There are tests indirectly covering the functionality but there wasn't any directly. I've added tests for this function. But since this is unexported function, I had to rename secret_test.go package name to encryptionconfigdata (from encryptionconfigdata_test) which I believe more correct.

ardaguclu:
Opened https://github.com/openshift/cluster-kube-apiserver-operator/pull/2162 to test

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +60,63 @@ func (k *KeyState) HasKMSPlugin() bool {
...
+
+// KMSSecretData stores data key-value pairs fetched from referenced secrets.
+// Entries maps secret names to their data key-value pairs.
+type KMSSecretData struct {
+	Entries map[string]map[string][]byte
+}
+
+func (d *KMSSecretData) Set(secretName, dataKey string, value []byte) error {
+	if strings.Contains(secretName, "_") {
+		return fmt.Errorf("secret name %q must not contain underscores", secretName)
```
p0lyn0mial, line 90: do we have a test for this case ?

pkg/operator/encryption/encryptiondata/secret.go
p0lyn0mial: let's also add a case for `numeric segments in rawKey (e.g., kms-plugin-secret-secret-a_port-8080-1)`

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 			},
 			Plugin: apiServerEncryption.KMS,
 		}
+
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			refSecret, err := c.secretClient.Secrets(openshiftConfigNS).Get(ctx, secretName, metav1.GetOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in openshift-config: %w", secretName, err)
```
p0lyn0mial: do we have a test case for this scenario ?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -76,16 +77,25 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
...
 			key.KMS.Plugin = kmsConfig
 		} else {
 			// encryption.apiserver.operator.openshift.io-kms-plugin-config data field is required for KMS
 			// encryption mode.
-			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", EncryptionSecretKMSPluginConfig)
+			return state.KeyState{}, fmt.Errorf("%s can not be empty, when mode is KMS", encryptionSecretKMSPluginConfig)
+		}
+		for dataKey, value := range s.Data {
+			rawKey, found := strings.CutPrefix(dataKey, encryptionSecretKMSSecretDataPrefix)
+			if !found || len(rawKey) == 0 {
```
p0lyn0mial: do we expect the rawKey to be 0 when found == true ?

ardaguclu: if datakey is `"encryption.apiserver.operator.openshift.io-kms-plugin-secret-`, found is true but len(rawKey) is 0.

ardaguclu: Maybe we should update this like this;

```go
if !found {
 continue
}
if len(raw) == 0 {
return fmt.Errorf()
}
```

ardaguclu: And we should do the same for `parseSecretDataKey` and `KeyIDFromPluginConfigSecretDataKey`

p0lyn0mial: does an empty suffix mean data corruption or a bug in the writer?

p0lyn0mial: do we expect `kms-plugin-config-` to be produced by our code ?

p0lyn0mial: will an empty rawKey will be caught by SetFromRawKey?

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +60,63 @@ func (k *KeyState) HasKMSPlugin() bool {
...
+}
+
+// KMSSecretData stores data key-value pairs fetched from referenced secrets.
+// Entries maps secret names to their data key-value pairs.
+type KMSSecretData struct {
+	Entries map[string]map[string][]byte
+}
+
+func (d *KMSSecretData) Set(secretName, dataKey string, value []byte) error {
+	if strings.Contains(secretName, "_") {
```
p0lyn0mial, line 89: should we also check if `secretName`, `dataKey` are not empty ? maybe even `value` ?

pkg/operator/encryption/state/types.go
```diff
@@ -53,13 +60,63 @@ func (k *KeyState) HasKMSPlugin() bool {
...
+		return fmt.Errorf("invalid combined key %q: expected format {secretName}%s{dataKey}", rawKey, secretDataKeySeparator)
+	}
+	return d.Set(parts[0], parts[1], value)
+}
+
+// FlatEntries returns the stored data as a flat map keyed by "secretName_dataKey".
+// "_" separates secretName from dataKey because "_" is forbidden in
+// Kubernetes secret names, making the split unambiguous.
+func (d *KMSSecretData) FlatEntries() map[string][]byte {
+	result := map[string][]byte{}
```
p0lyn0mial, line 119: should we check if `d.Entries` is `nill` and if yes then simply return `nil` as well ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -67,6 +102,19 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
...
+				if existing, exists := kmsPluginsSecretData.ByKeyID[key.Key.Name]; exists {
+					if !equality.Semantic.DeepEqual(existing.FlatEntries(), key.KMS.PluginSecretData.FlatEntries()) {
```
p0lyn0mial: should we compare `FlatEntries()` why ?

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -30,6 +30,40 @@ type Config struct {
...
+		return err
+	}
+	d.ByKeyID[keyID] = sd
+	return nil
+}
+
+// FlatEntriesByKeyID returns the stored data as a map of keyID to flat entries,
+// where each flat entry is keyed by "secretName_dataKey".
+func (d *KMSPluginsSecretData) FlatEntriesByKeyID() map[string]map[string][]byte {
+	result := map[string]map[string][]byte{}
```
p0lyn0mial, line 68: same question here.

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -30,6 +30,40 @@ type Config struct {
...
+// KMSPluginsSecretData maps keyID to secret data carried from Key Secrets into
+// the encryption-config Secret. Structure: keyID → secretName → dataKey → value.
+type KMSPluginsSecretData struct {
+	ByKeyID map[string]state.KMSSecretData
+}
+
+// SetFromRawKey stores a value for the given keyID, splitting rawKey
+// on "_" into secretName and dataKey.
+func (d *KMSPluginsSecretData) SetFromRawKey(keyID, rawKey string, value []byte) error {
+	if d.ByKeyID == nil {
```
p0lyn0mial, line 51: should we also validate keyID ? (not empty)

pkg/operator/encryption/encryptiondata/secret.go
p0lyn0mial: ah, maybe we should add one more case

`s.Data[encryptionConfigSecretDataPrefix+flatKey+"-"+keyID] = value`

where `flatKey` ends with `-`, e.g.: `kms-plugin-secret-secret-a_port-8080--1`

wdyt ?

pkg/operator/encryption/secrets/secrets.go
ardaguclu: > does an empty suffix mean data corruption or a bug in the writer?

I was thinking more about user manually updates these Secrets.

> do we expect kms-plugin-config- to be produced by our code ?

All prefixes and suffixes are produced by our code

> will an empty rawKey will be caught by SetFromRawKey?

Yes, it will return `fmt.Errorf("invalid combined key %q: expected format {secretName}%s{dataKey}", rawKey, secretDataKeySeparator)` error

ardaguclu: So since `SetFromRawKey` errors out for empty rawKey, we can remove `len(rawKey) == 0` from that condition

pkg/operator/encryption/encryptiondata/config.go
ardaguclu: Because you asked me to do that :) https://github.com/openshift/library-go/pull/2212#discussion_r3276327675.

I think it doesn't matter whether we compare `existing` or `existing.FlatEntries`.

pkg/operator/encryption/secrets/secrets.go
p0lyn0mial: sgtm

pkg/operator/encryption/encryptiondata/config.go
p0lyn0mial: :D

p0lyn0mial: let's compare `existing`

pkg/operator/encryption/state/types_test.go
```diff
@@ -0,0 +1,181 @@
...
+			rawKey:  "no-separator-here",
+			value:   []byte("v"),
+			wantErr: true,
+		},
+		{
+			name:   "multiple underscores splits on first only",
+			rawKey: "vault_approle_secret_role-id",
+			value:  []byte("v"),
+			expected: KMSSecretData{Entries: map[string]map[string][]byte{
+				"vault": {"approle_secret_role-id": []byte("v")},
```
ardaguclu, line 122: @p0lyn0mial if datakey contains `_`, this will be the expected output. I assume that we agree with that?

p0lyn0mial: I think so. `vault` is the secret name and `approle_secret_role-id` is the data key, right ?

p0lyn0mial:
/lgtm
/approve

/hold

so that @bertinatto can take a look.

pkg/operator/encryption/controllers/key_controller.go
bertinatto: nit: can use the `openshiftConfigNS` new constant to refer to `openshift-config` namespace

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
...
+			}
+			for _, key := range expectedKeys {
+				v, ok := refSecret.Data[key]
+				if !ok {
+					return nil, fmt.Errorf("secret %s in openshift-config is missing required key %q", secretName, key)
```
bertinatto: same here

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -281,6 +283,22 @@ func (c *keyController) generateKeySecret(keyID uint64, currentMode state.Mode,
 			},
 			Plugin: apiServerEncryption.KMS,
 		}
+
+		if secretName, expectedKeys := referencedSecretName(apiServerEncryption.KMS); len(secretName) > 0 {
+			refSecret, err := c.secretClient.Secrets(openshiftConfigNS).Get(ctx, secretName, metav1.GetOptions{})
+			if err != nil {
+				return nil, fmt.Errorf("failed to get secret %s in openshift-config: %w", secretName, err)
+			}
+			for _, key := range expectedKeys {
```
bertinatto, line 294: this lgtm, just asking for my knowledge: I believe I mentioned about validating the secret before, but you mentioned the encryption controllers should not have semantic knowledge about the secret (which also makes sense). `referencedSecretName` now has that knowledge, and I'm fine with that, but just wondering if I missed any discussions about it

pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -56,7 +61,26 @@ func FromSecret(encryptionConfigSecret *corev1.Secret) (*Config, error) {
...
+	for key, value := range encryptionConfigSecret.Data {
+		keyID, rawKey, found, err := parseSecretDataKey(key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse secret data key %s: %w", key, err)
+		}
+		if !found {
+			continue
+		}
+		if err := kmsPluginsSecretData.SetFromRawKey(keyID, rawKey, value); err != nil {
+			return nil, fmt.Errorf("failed to parse key %s: %w", key, err)
```
bertinatto: nit: failed to parse or to set?

bertinatto (commented):
/lgtm

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -383,6 +401,20 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
...
 
+// referencedSecretName returns the name of the secret referenced by the KMS plugin
+// config and the specific data keys to carry from that secret. Only the listed keys
+// are copied into the Key Secret; any other data in the referenced secret is ignored.
+func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
+	switch plugin.Type {
+	case configv1.VaultKMSProvider:
+		// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
+		// These are the only keys carried into the encryption key secret.
+		return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
```
p0lyn0mial: @ardaguclu I think we should also check `VaultAuthenticationType` https://github.com/openshift/api/blob/master/config/v1/types_kmsencryption.go#L94

ardaguclu: Good point. Yes, we should

p0lyn0mial: I think that as of today it must be set to `AppRole`

p0lyn0mial: https://github.com/openshift/api/blob/master/config/v1/types_kmsencryption.go#L87

p0lyn0mial: make sure we have a unit test for that case too :)

test/e2e-encryption/encryption_test.go
```diff
@@ -452,6 +489,18 @@ func TestEncryptionIntegration(tt *testing.T) {
 	)
 	waitForConditionStatus("Encrypted", operatorv1.ConditionTrue)
 
+	t.Logf("Create vault AppRole vault AppRole secret")
```
bertinatto: nit: `vault AppRole` is duplicate

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: Updated

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: Updated with unit tests

test/e2e-encryption/encryption_test.go
ardaguclu: Updated

pkg/operator/encryption/controllers/key_controller.go
p0lyn0mial: I think we basically have two options:

Option 1 would be to restrict fields to only well-known ones  essentially what we have today.
Option 2 would be to copy arbitrary fields.

Copying arbitrary fields seems too permissive. What’s the point of copying fields we won’t need?

The expected fields will have to be documented for users, just like we document them for, for example, identity providers.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -383,6 +401,24 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
...
+// are copied into the Key Secret; any other data in the referenced secret is ignored.
+func referencedSecretName(plugin configv1.KMSPluginConfig) (string, []string) {
+	switch plugin.Type {
+	case configv1.VaultKMSProvider:
+		switch plugin.Vault.Authentication.Type {
+		case configv1.VaultAuthenticationTypeAppRole:
+			// The Vault AppRole secret must contain "role-id" and "secret-id" keys.
+			// These are the only keys carried into the encryption key secret.
+			return plugin.Vault.Authentication.AppRole.Secret.Name, []string{"role-id", "secret-id"}
+		}
```
p0lyn0mial, line 419: should we return an error on unrecognized types ?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -383,6 +401,24 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
...
+		return "", nil
+	default:
+		return "", nil
```
p0lyn0mial: should we return an error here too ? i think it is expected to be called only for KMS providers, no ?

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We should return error. Updated.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We should return error. Updated.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: I agree about the reasons of working on selective set of data keys.  

I think it is better to logically store provider specific details (i.e. Vault) in kms/vault.go. So that they won't be spread across all over the code base.

ardaguclu:
Test passed https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2164/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-gcp-operator-encryption-kms/2057681540460056576 with the combination of https://github.com/openshift/library-go/pull/2240.

I'll release the hold, since https://github.com/openshift/library-go/pull/2240 can be merged separately.
/hold cancel

Thank you @p0lyn0mial @bertinatto @gangwgr for all your collaboration

PR #2214, opened 2026-05-12
NO-JIRA: Error out when there is duplicate kms-plugin-config in encryption-config Secret

ardaguclu:
kms-plugin-configs are stored in encryption-config Secret as deduplicated. So theoretically it is not possible to have multiple plugins for same key id. But it is better to add this check.

## Summary by CodeRabbit

* **Bug Fixes**
  * KMS encryption configuration now validates for duplicate key identifiers and returns a clear error message when duplicates are detected, preventing silent configuration inconsistencies.


ardaguclu:
/retitle NO-JIRA: Error out when there is duplicate kms-plugin-config in encryption-config Secret

PR #2218, opened 2026-05-12
NO-JIRA: add two more encryption rotation scenarios

tjungblu:
This adds two new encryption tests:
* Forced rotation during first migration
* Forced rotation during a rotation

## Summary by CodeRabbit

* **Tests**
  * Added operator condition–based detection for encryption migration progress with a secret-polling fallback
  * Added rotation scenarios to validate behavior during active migrations (first-migration and overlapping-rotation cases)
  * Added helpers to verify encryption stability and track key rotations
  * Enhanced test flow to apply APIServer encryption settings immediately before waiting for migration completion

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/client-go/util/retry"
 
 	configv1 "github.com/openshift/api/config/v1"
+	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 
 	"github.com/openshift/library-go/test/library"
@@ -56,14 +57,22 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
+// GetOperatorConditionsFuncType fetches operator conditions (e.g. for EncryptionMigrationControllerProgressing).
+type GetOperatorConditionsFuncType func(t testing.TB) ([]operatorv1.OperatorCondition, error)
+
 func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
-	t.Helper()
-
-	t.Logf("Starting encryption e2e test for %q mode", provider.Type)
-
 	clientSet := GetClients(t)
 	lastMigratedKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube, namespace, labelSelector)
 	require.NoError(t, err)
+
+	ApplyAPIServerEncryptionType(t, clientSet, provider)
+	WaitForEncryptionKeyBasedOn(t, clientSet.Kube, lastMigratedKeyMeta, provider.Type, defaultTargetGRs, namespace, labelSelector)
+	return clientSet
+}
+
+func ApplyAPIServerEncryptionType(t testing.TB, clientSet ClientSet, provider EncryptionProvider) {
+	t.Helper()
+	t.Logf("Starting encryption e2e test for %q mode", provider.Type)
 
 	apiServer, err := clientSet.ApiServerConfig.Get(context.TODO(), "cluster", metav1.GetOptions{})
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -222,6 +222,12 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
 			continue // no write key to migrate to
 		}
 
+		// TODO(thomas): wire up a hook to define this from a provider configuration
+		// TODO(thomas): when we're on KMS provider we can grab the keyId from an annotation
+		// TODO(thomas): a new rotation controller can update the annotation for KMS
+
+		// TODO(thomas): alternative, when we detect a new KeyId, we can also remove the migrated-* annotations,
+		// that way the controller will definitely try to reconcile it again.
```
tjungblu: I'll remove that, sorry

test/library/encryption/scenarios.go
```diff
@@ -254,3 +265,84 @@ func TestEncryptionRotation(t testing.TB, scenario RotationScenario) {
...
+	expectedFirstWriteKey, err := determineNextEncryptionKeyName(prevMeta.Name, labelSelector)
+	require.NoError(e, err)
+
+	ApplyAPIServerEncryptionType(e, clientSet, scenario.EncryptionProvider, scenario.TargetGRs)
+
+	if !WaitForEncryptionMigrationInProgressWindow(e, clientSet.Kube, scenario.GetOperatorConditionsFunc, expectedFirstWriteKey, scenario.TargetGRs, ns, labelSelector) {
+		t.Skipf("initial migration finished before an in-progress window was observed; set GetOperatorConditionsFunc or use a cluster where migration stays visible longer")
+	}
+
+	require.NoError(e, ForceKeyRotation(e, scenario.UnsupportedConfigFunc, fmt.Sprintf("test-rotation-during-first-migration-%s", rand.String(4))))
```
p0lyn0mial, line 326: `ForceKeyRotation` has no effect on KMS. What is out plan here ? Do we want to run the new scenarios for all encryption providers ?

test/library/encryption/helpers.go
```diff
@@ -274,6 +288,27 @@ func ForceKeyRotation(t testing.TB, updateUnsupportedConfig UpdateUnsupportedCon
 	})
 }
 
+// ClearForcedKeyRotationReason clears encryption.reason under UnsupportedConfigOverrides (same merge path as
+// ForceKeyRotation). Call when a test finishes so the next test in sequence does not inherit a non-empty
+// reason and the key controller does not keep seeing an external rotation request.
+func ClearForcedKeyRotationReason(t testing.TB, updateUnsupportedConfig UpdateUnsupportedConfigFunc) error {
```
p0lyn0mial, line 292: instead of a new fn why no to change teh ForceKeyRotation to register a cleanup routine via t.Cleanup ?

test/library/encryption/scenarios.go
```diff
@@ -254,3 +265,84 @@ func TestEncryptionRotation(t testing.TB, scenario RotationScenario) {
...
+	// n=2: one write-key revision from turning encryption on, one from ForceKeyRotation.
+	WaitForNRotations(e, clientSet.Kube, scenario.EncryptionProvider, scenario.TargetGRs, ns, labelSelector, prevMeta, 2)
```
p0lyn0mial: also for KMS, rotation will not create a new key.

p0lyn0mial (commented):
I did a first quick pass. I’d like to clarify the scope. Are we creating new cases that will support all providers, or just KMS?

test/library/encryption/scenarios.go
tjungblu: ForceKeyRotation will be implemented differently later on

test/library/encryption/helpers.go
tjungblu: why would I couple those two functionalities together? what if I just want to force a key rotation and not cleanup?

test/library/encryption/scenarios.go
tjungblu: yep, so this would be implemented differently in KMS - I assume this would use this API that you and Ben cobbled together so far

PR #2219, opened 2026-05-12
NO-JIRA: simplify waiting in GetLastKeyMeta

tjungblu:
Simplifies the changes made in #1188 to just do a dumb 5m context timeout wait with 30s polling in between.

## Summary by CodeRabbit

* **Tests**
  * Improved test utilities for retrieving encryption key metadata to be more resilient to transient listing errors.
  * Polling now retries periodically over a longer window and logs interim failures instead of aborting early.
  * Enhances reliability of key discovery during test runs.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -201,29 +201,16 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
 
 	// in theory the max time we tolerate disruption on an SNO cluster is 60 seconds
 	// so we set the timeout to 5 min just in case
-	pollTimeout := time.Minute * 5
-
-	// set the number of step to high value
-	// we should stop on timeout otherwise the backoff returns after 5 steps
-	// and we never wait the timeout value
-	backOff := retry.DefaultBackoff
-	backOff.Steps = 9999
-
-	// in theory the max time we tolerate disruption on an SNO cluster is 60 seconds
-	// so we set the timeout to 5 min just in case
-	err := onErrorWithTimeout(pollTimeout, backOff, func(err error) bool {
-		if !transientAPIError(err) {
-			t.Logf("error = %v is not retriable, failed to get the metadata from the last encryption key", err)
-			return false
-		}
-		return true // retry
-	}, func() (err error) {
-		selectedSecrets, err = secretsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+	err := wait.PollUntilContextTimeout(t.Context(), 30*time.Second, time.Minute*5, true, func(ctx context.Context) (result bool, err error) {
+		selectedSecrets, err = secretsClient.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 		if err != nil {
-			t.Logf("failed to list secrets, err = %v", err.Error())
+			t.Logf("GetLastKeyMeta: Failed to list secrets: %v", err)
+			return false, nil
 		}
-		return
+
+		return true, nil
 	})
+
 	if err != nil {
 		return EncryptionKeyMeta{}, err
 	}


test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
-		}
-		return true // retry
-	}, func() (err error) {
-		selectedSecrets, err = secretsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+	err := wait.PollUntilContextTimeout(t.Context(), 30*time.Second, time.Minute*5, true, func(ctx context.Context) (bool, error) {
+		labelSecrets, err := secretsClient.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 		if err != nil {
-			t.Logf("failed to list secrets, err = %v", err.Error())
+			t.Logf("GetLastKeyMeta: Failed to list secrets: %v", err)
+			return false, err
```
ardaguclu: Should we return err only when isTransientAPIError true? and the rest we should return false, nil?

tjungblu: that's actually the thing that drove me to change it in the first place, isTransientAPIError doesn't cover everything. For example, auth can be offline for those said 5 minutes, but 401s are not covered. I also don't want to change the function because this is used in too many places.

ardaguclu: ~~You are right. The error types not listed in `isTransientAPIError` does not make sense~~

Just one note: test/library/encryption/errors.go file won' be used anymore.

test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
 		}
-		return
+
+		selectedSecrets = labelSecrets
+		return true, nil
 	})
```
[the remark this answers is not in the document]
ardaguclu: Do you know any error that requires the retry?

test/library/encryption/helpers.go
ardaguclu: KMS changes are disruptive for the cluster. After [this comment](https://github.com/openshift/library-go/pull/2219/changes#r3231959532), I think we should retry for such errors. Because for instance, on SNO, there can be a period of time that kube-apiserver is not functioning?, we should retry. Otherwise, test will flake.

tjungblu: it is used in several places, ForceKeyRotation for example

test/library/encryption/helpers.go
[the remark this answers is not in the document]
tjungblu: is this exhaustive for all errors that can happen?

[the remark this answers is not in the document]
tjungblu: no that last comment was for you @coderabbitai

test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
 	// so we set the timeout to 5 min just in case
-	pollTimeout := time.Minute * 5
-
-	// set the number of step to high value
-	// we should stop on timeout otherwise the backoff returns after 5 steps
-	// and we never wait the timeout value
-	backOff := retry.DefaultBackoff
-	backOff.Steps = 9999
-
-	// in theory the max time we tolerate disruption on an SNO cluster is 60 seconds
```
p0lyn0mial, line 212: would it make sense to preserve this comment ? (it explains why 5m)

test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
-		if !transientAPIError(err) {
-			t.Logf("error = %v is not retriable, failed to get the metadata from the last encryption key", err)
-			return false
-		}
-		return true // retry
-	}, func() (err error) {
-		selectedSecrets, err = secretsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+	err := wait.PollUntilContextTimeout(t.Context(), 30*time.Second, time.Minute*5, true, func(ctx context.Context) (bool, error) {
+		labelSecrets, err := secretsClient.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
 		if err != nil {
```
p0lyn0mial, line 206: let's use transientAPIError - makes the test less brittle

test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
-			t.Logf("failed to list secrets, err = %v", err.Error())
+			t.Logf("GetLastKeyMeta: Failed to list secrets: %v", err)
+			return false, nil
 		}
-		return
+
+		selectedSecrets = labelSecrets
```
p0lyn0mial: why do we need this ?

test/library/encryption/helpers.go
```diff
@@ -201,29 +201,17 @@ func GetLastKeyMeta(t testing.TB, kubeClient kubernetes.Interface, namespace, la
...
-	err := onErrorWithTimeout(pollTimeout, backOff, func(err error) bool {
-		if !transientAPIError(err) {
-			t.Logf("error = %v is not retriable, failed to get the metadata from the last encryption key", err)
-			return false
-		}
-		return true // retry
-	}, func() (err error) {
-		selectedSecrets, err = secretsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+	err := wait.PollUntilContextTimeout(t.Context(), 30*time.Second, time.Minute*5, true, func(ctx context.Context) (bool, error) {
+		labelSecrets, err := secretsClient.List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
```
p0lyn0mial: selectedSecrets ?

test/library/encryption/helpers.go
tjungblu: the comment is still there

p0lyn0mial: oh, ok.

test/library/encryption/helpers.go
tjungblu: as mentioned earlier on slack and on the other thread with the ai rabbit: it doesn't cover all transient errors

test/library/encryption/helpers.go
tjungblu: it' s compilation error due to the scoping of the function. 

> Unused variable 'selectedSecrets'

I gave the return type a variable name and removed the assignment.

test/library/encryption/helpers.go
p0lyn0mial: Okay, but right now we have a mix. Some tests use transientAPIError and some don’t.

Are there any known transient errors we’re currently missing? Would it make sense to add them to transientAPIError?

In general, do we currently have issues with transient errors?

I think it would be better not to hide any errors. At the moment, we treat all errors as transient. Using transientAPIError instead would raise the bar for the entire platform, and I think we’re already doing a good job since the tests are generally stable :) 

WDYT?

PR #2221, opened 2026-05-13
CNTRLPLANE-3237:: pkg/operator/encryption/kms: cleanup

p0lyn0mial:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for KMS plugin secret keys to reject invalid or negative key IDs and improve error handling.

* **Refactor**
  * Internal handling of plugin secret key formatting/parsing consolidated and localized to reduce external coupling.

* **Tests**
  * Expanded and reorganized tests for plugin secret key construction/parsing and plugin configuration handling, including new negative cases.


pkg/operator/encryption/encryptiondata/secret_test.go
```diff
@@ -5,12 +5,136 @@ import (
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			data := baseSecret(t)
+			for k, v := range tt.extraKeys {
+				data[k] = v
+			}
+			secret := &corev1.Secret{Data: data}
+			_, err := encryptiondata.FromSecret(secret)
+			if tt.wantError && err == nil {
```
bertinatto: should we unify whether we use `github.com/stretchr/testify/require` in library-go unit tests? Right now it's a mix of require and manual if/t.Fatal checks

pkg/operator/encryption/encryptiondata/secret_test.go
```diff
@@ -5,12 +5,136 @@ import (
...
+				Encryption: &apiserverconfigv1.EncryptionConfiguration{
+					Resources: []apiserverconfigv1.ResourceConfiguration{{
+						Resources: []string{"secrets"},
+					}},
+				},
+				KMSPlugins: map[string]configv1.KMSPluginConfig{
+					tt.keyID: {},
+				},
+			}
+			_, err := encryptiondata.ToSecret("ns", "name", cfg)
```
bertinatto: IIUC, the idea of this new test is to test `toPluginConfigSecretDataKeyFor` via `ToSecret`. In general I think it's better to unit test the function directly because then you can check exactly what it produces for a given input. For example, the old test verified that for a given key `42`, the function would produce a data key `kms-plugin-config-42`, but now it only checks that the function didn't return any errors.

to test `toPluginConfigSecretDataKeyFor` we'd need to change the package directive in this file to `encryptiondata` though (it's currently `encryptiondata_test`)

pkg/operator/encryption/encryptiondata/secret_test.go
```diff
@@ -5,12 +5,136 @@ import (
...
+			extraKeys: map[string][]byte{"": {}},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			data := baseSecret(t)
+			for k, v := range tt.extraKeys {
+				data[k] = v
+			}
```
bertinatto: nit: 	`maps.Copy(data, tt.extraKeys)`

ardaguclu:
Changes look good to me. Comments from @bertinatto make sense.

pkg/operator/encryption/encryptiondata/secret_test.go
p0lyn0mial: yes, we could create a new test file which would be in the same pkg as the production code to test the priv methods. then we could just move the prev tests. wdty ?

ardaguclu:
Comments are all about test changes. I think, those can be handled separately. 
/lgtm

p0lyn0mial:
/hold 

I think I will move the tests to a private file.

p0lyn0mial:
> I think I will move the tests to a private file.

done.

/hold cancel

pkg/operator/encryption/encryptiondata/secret_test.go
p0lyn0mial: replaced require with standard assertions.

bertinatto (commented):
/lgtm

pkg/operator/encryption/encryptiondata/secret_test.go
```diff
@@ -1,4 +1,4 @@
-package encryptiondata_test
+package encryptiondata
```
bertinatto, line 1: +1

PR #2223, opened 2026-05-14
CNTRLPLANE-3236: kms: fix args passed to the vault mock plugin

bertinatto:
Previously, `-vault-namespace` and `-transit-mount` reflected whatever values their counterparts api fields had. However, those fields are optional in the api, but since they are not pointers, their values could be be empty strings. As a result, this patch checks if those fields contain empty strings and _optionally_ pass them through to the KMS plugin via CLI args.

In addition, this patch also plumbs through the role-id param, which is a required api field.

## Improvements

* Vault KMS plugin sidecar container arguments are now dynamically constructed based on your specific configuration settings
* Optional Vault configuration parameters (vault-namespace, transit-mount) are included only when you explicitly configure them
* Refined vault-kms-plugin sidecar command-line argument ordering to ensure proper initialization sequence during deployment


ardaguclu:
Changes look good to me
/lgtm
/hold
feel free to unhold if tests work

PR #2225, opened 2026-05-14
CNTRLPLANE-3238: test/kms: add reusable Vault KMS helpers and KMSTestProvider abstraction

gangwgr:
Move the Vault KMS encryption config and AppRole secret setup from cluster-kube-apiserver-operator into library-go so any operator can reuse them.

Add KMSTestProvider type that pairs an encryption config with its setup function, enabling the same test cases to run against multiple KMS provider without duplicating test logic.

## Summary by CodeRabbit

* **Tests**
  * Enhanced Vault KMS encryption tests with Enterprise namespace support and AppRole authentication credentials setup
  * Updated encryption provider configuration structure with wrapper type supporting optional per-provider setup hooks
  * Reorganized supported static encryption providers list for encryption test scenario handling

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -36,9 +36,9 @@ var (
 	waitPollTimeout       = 69*time.Minute + 10*time.Minute
 	defaultEncryptionMode = string(configv1.EncryptionTypeIdentity)
 
-	SupportedStaticEncryptionProviders = []configv1.APIServerEncryption{
-		{Type: configv1.EncryptionTypeAESGCM},
-		{Type: configv1.EncryptionTypeAESCBC},
+	SupportedStaticEncryptionProviders = []EncryptionProvider{
+		{APIServerEncryption: configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESGCM}},
+		{APIServerEncryption: configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESCBC}},
 	}
 )
 
@@ -56,8 +56,9 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
-func SetAndWaitForEncryptionType(t testing.TB, provider configv1.APIServerEncryption, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
+func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
 	t.Helper()
+
 	t.Logf("Starting encryption e2e test for %q mode", provider.Type)
 
 	clientSet := GetClients(t)
@@ -66,10 +67,13 @@ func SetAndWaitForEncryptionType(t testing.TB, provider configv1.APIServerEncryp
 
 	apiServer, err := clientSet.ApiServerConfig.Get(context.TODO(), "cluster", metav1.GetOptions{})
 	require.NoError(t, err)
-	needsUpdate := !equality.Semantic.DeepEqual(apiServer.Spec.Encryption, provider)
+	needsUpdate := !equality.Semantic.DeepEqual(apiServer.Spec.Encryption, provider.APIServerEncryption)
 	if needsUpdate {
-		t.Logf("Updating encryption configuration for APIServer from %#v to %#v", apiServer.Spec.Encryption, provider)
-		apiServer.Spec.Encryption = provider
+		if provider.Setup != nil {
+			provider.Setup(t)
+		}
+		t.Logf("Updating encryption configuration for APIServer from %#v to %#v", apiServer.Spec.Encryption, provider.APIServerEncryption)
+		apiServer.Spec.Encryption = provider.APIServerEncryption
 		_, err = clientSet.ApiServerConfig.Update(context.TODO(), apiServer, metav1.UpdateOptions{})
 		require.NoError(t, err)
 	} else {
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -1,11 +1,62 @@
 package kms
 
 import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/clock"
+
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	library "github.com/openshift/library-go/test/library/encryption"
 )
 
-// DefaultFakeKMSPluginConfig is a fake Vault KMS configuration used by tests.
-// The values are not real and are likely to change as the KMS integration evolves.
+const (
+	defaultVaultNamespace         = "vault-kms"
+	defaultVaultCredentialsSecret = "vault-credentials"
+	defaultVaultAppRoleSecretName = "vault-approle-secret"
+	defaultVaultKMSPluginImage    = "quay.io/openshifttest/mock-kms-plugin@sha256:03bb07a2c08b509653c4c70217a06a4b389c10b4d87922f50ee5eac82db5e140"
+	defaultVaultAddress           = "https://vault.vault-kms.svc:8200"
+	defaultVaultEnterpriseNS      = "admin"
+	defaultVaultTransitMount      = "transit"
+	defaultVaultTransitKey        = "kms-key"
+	defaultAppRoleTargetNamespace = "openshift-config"
+)
+
+// DefaultVaultEncryptionProvider is a ready-to-use Vault KMS EncryptionProvider for e2e tests.
+// It bundles the default config with the AppRole secret setup.
+var DefaultVaultEncryptionProvider = library.EncryptionProvider{
+	APIServerEncryption: DefaultVaultKMSPluginConfig,
+	Setup:               ensureDefaultVaultAppRoleSecret,
+}
+
+// DefaultVaultKMSPluginConfig is the standard Vault KMS encryption config
+// used by CI e2e tests.
+var DefaultVaultKMSPluginConfig = configv1.APIServerEncryption{
+	Type: configv1.EncryptionTypeKMS,
+	KMS: configv1.KMSPluginConfig{
+		Type: configv1.VaultKMSProvider,
+		Vault: configv1.VaultKMSPluginConfig{
+			KMSPluginImage: defaultVaultKMSPluginImage,
+			VaultAddress:   defaultVaultAddress,
+			VaultNamespace: defaultVaultEnterpriseNS,
+			TransitMount:   defaultVaultTransitMount,
+			TransitKey:     defaultVaultTransitKey,
+			Authentication: configv1.VaultAuthentication{
+				Type: configv1.VaultAuthenticationTypeAppRole,
+				AppRole: configv1.VaultAppRoleAuthentication{
+					Secret: configv1.VaultSecretReference{Name: defaultVaultAppRoleSecretName},
+				},
+			},
+		},
+	},
+}
+
+// DefaultFakeKMSPluginConfig is a fake Vault KMS configuration used by unit tests.
 var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
 	Type: configv1.VaultKMSProvider,
 	Vault: configv1.VaultKMSPluginConfig{
@@ -19,4 +70,32 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
 		},
 		TransitKey: "test-transit-key",
 	},
+}
+
+// ensureDefaultVaultAppRoleSecret reads credentials from the vault-credentials secret
+// (created by a CI step) and applies the AppRole secret in openshift-config
+// using the default configuration constants.
+func ensureDefaultVaultAppRoleSecret(t testing.TB) {
+	t.Helper()
+	ctx := t.Context()
+	cs := library.GetClients(t)
+
+	creds, err := cs.Kube.CoreV1().Secrets(defaultVaultNamespace).Get(ctx, defaultVaultCredentialsSecret, metav1.GetOptions{})
+	require.NoError(t, err, "failed to read %s/%s secret (was the vault-install CI step run?)", defaultVaultNamespace, defaultVaultCredentialsSecret)
+
+	required := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      defaultVaultAppRoleSecretName,
+			Namespace: defaultAppRoleTargetNamespace,
+		},
+		Type: corev1.SecretTypeOpaque,
+		Data: map[string][]byte{
+			"roleID":   creds.Data["role-id"],
+			"secretID": creds.Data["secret-id"],
+		},
+	}
+	recorder := events.NewInMemoryRecorder("vault-approle-secret-setup", clock.RealClock{})
+	_, changed, err := resourceapply.ApplySecret(ctx, cs.Kube.CoreV1(), recorder, required)
+	require.NoError(t, err, "failed to apply AppRole secret")
+	t.Logf("Applied AppRole secret %s in %s (changed=%v)", defaultVaultAppRoleSecretName, defaultAppRoleTargetNamespace, changed)
 }
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -25,19 +25,27 @@ type BasicScenario struct {
 	AssertFunc                      func(t testing.TB, clientSet ClientSet, expectedMode configv1.EncryptionType, namespace, labelSelector string)
 }
 
+// EncryptionProvider pairs an encryption config with an optional setup function
+// that ensures prerequisites (secrets, credentials, infrastructure) are in place.
+type EncryptionProvider struct {
+	configv1.APIServerEncryption
+	// Setup is called once before the provider is first used. May be nil.
+	Setup func(t testing.TB)
+}
+
 func TestEncryptionTypeIdentity(t testing.TB, scenario BasicScenario) {
 	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
-	clientSet := SetAndWaitForEncryptionType(e, configv1.APIServerEncryption{Type: configv1.EncryptionTypeIdentity}, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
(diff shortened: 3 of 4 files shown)


test/library/encryption/kms/vault.go
```diff
@@ -1,11 +1,54 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	configv1 "github.com/openshift/api/config/v1"
+	library "github.com/openshift/library-go/test/library/encryption"
+)
+
+const (
+	DefaultVaultNamespace         = "vault-kms"
+	DefaultVaultCredentialsSecret = "vault-credentials"
+	DefaultVaultAppRoleSecretName = "vault-approle-secret"
```
ardaguclu:
```suggestion
	DefaultAppRoleSecretName = "vault-approle-secret"
```

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +63,40 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
+	t.Helper()
+	ctx := context.Background()
+	cs := library.GetClients(t)
+
+	creds, err := cs.Kube.CoreV1().Secrets(DefaultVaultNamespace).Get(ctx, DefaultVaultCredentialsSecret, metav1.GetOptions{})
+	require.NoError(t, err, "failed to read %s/%s secret (was the vault-install CI step run?)", DefaultVaultNamespace, DefaultVaultCredentialsSecret)
+
+	roleID, ok := creds.Data["role-id"]
+	require.Truef(t, ok && len(roleID) > 0, "missing or empty key %q in %s/%s", "role-id", DefaultVaultNamespace, DefaultVaultCredentialsSecret)
+	secretID, ok := creds.Data["secret-id"]
```
ardaguclu: We should be careful to not leak

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +63,40 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
+		Data: map[string][]byte{
+			"roleID":   roleID,
+			"secretID": secretID,
+		},
+	}
+	_, err = cs.Kube.CoreV1().Secrets(DefaultAppRoleTargetNamespace).Create(ctx, appRoleSecret, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		existing, getErr := cs.Kube.CoreV1().Secrets(DefaultAppRoleTargetNamespace).Get(ctx, DefaultVaultAppRoleSecretName, metav1.GetOptions{})
+		require.NoError(t, getErr, "failed to get existing AppRole secret for update")
+		appRoleSecret.ResourceVersion = existing.ResourceVersion
```
ardaguclu: Why do we need ResourceVersion?

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +63,40 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
+		_, err = cs.Kube.CoreV1().Secrets(DefaultAppRoleTargetNamespace).Update(ctx, appRoleSecret, metav1.UpdateOptions{})
```
ardaguclu: Why do we update?

test/library/encryption/kms/vault.go
gangwgr: Better to have this DefaultVaultAppRoleSecretName, because all are specific to vault

test/library/encryption/kms/vault.go
gangwgr: Validatio uses require.Contains which only prints key names, never secret values

test/library/encryption/kms/vault.go
gangwgr: not needed removed

test/library/encryption/kms/vault.go
gangwgr: updated, if secret already exists, just log and return

test/library/encryption/kms/provider.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"testing"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// KMSTestProvider pairs a KMS encryption config with its prerequisite setup.
+// This allows the same test cases to run against any KMS provider (Vault, AWS, etc.)
+// by simply iterating over a slice of providers.
+type KMSTestProvider struct {
```
p0lyn0mial: how will this provider be used ? why do we need it ?

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +63,37 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
 		TransitKey: "test-transit-key",
 	},
 }
+
+// EnsureVaultAppRoleSecret reads credentials from the vault-credentials secret
+// (created by a CI step) and creates the AppRole secret in openshift-config.
+// If the secret already exists it is left untouched.
+func EnsureVaultAppRoleSecret(t testing.TB) {
```
p0lyn0mial: will this function create an AppRole Secret for the default configuration ? 

if yes, then maybe we should rename to `EnsureDefaultVaultAppRoleSecret`

test/library/encryption/kms/vault.go
```diff
@@ -1,11 +1,54 @@
...
+		Vault: configv1.VaultKMSPluginConfig{
+			KMSPluginImage: DefaultVaultKMSPluginImage,
+			VaultAddress:   DefaultVaultAddress,
+			VaultNamespace: DefaultVaultEnterpriseNS,
+			TransitMount:   DefaultVaultTransitMount,
+			TransitKey:     DefaultVaultTransitKey,
+			Authentication: configv1.VaultAuthentication{
+				Type: configv1.VaultAuthenticationTypeAppRole,
+				AppRole: configv1.VaultAppRoleAuthentication{
+					Secret: configv1.VaultSecretReference{Name: DefaultVaultAppRoleSecretName},
```
p0lyn0mial: could we just simply use `DefaultVaultCredentialsSecret` ? why do we need a new secret ?

ardaguclu: We enforce the users to create a Secret under openshift-config namespace. Vault is deployed to dedicated vault-kms namespace. So we need this action

test/library/encryption/kms/provider.go
gangwgr: Right now KAS-O tests we only have Vault KMS. But whenwe have  a second provider (e.g. AWS KMS) and want to add add that, we want to run same cases

Run the same TestKMSEncryptionOnOff test against each provider
Run TestEncryptionProvidersMigration across all providers
Each provider needs its own setup (Vault needs AppRole secret, AWS would need IAM credentials, etc.)
KMSTestProvider groups the config + setup together so the operator test just does:

```

var allProviders = []kms.KMSTestProvider{
    kms.DefaultVaultTestProvider,
    // future: kms.DefaultAWSTestProvider,
}

func testKMSEncryptionOnOff(t testing.TB) {
    kms.SetupAll(t, allProviders)
    for _, p := range allProviders {
        library.TestEncryptionTurnOnAndOff(t, library.OnOffScenario{
            EncryptionProvider: p.EncryptionConfig,
            ...
        })
    }
```
}

test/library/encryption/kms/vault.go
gangwgr: Yes for same as @ardaguclu said

test/library/encryption/kms/vault.go
gangwgr: yes, updated

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +63,38 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
+// If the secret already exists it is left untouched.
+func EnsureDefaultVaultAppRoleSecret(t testing.TB) {
+	t.Helper()
+	ctx := context.Background()
+	cs := library.GetClients(t)
+
+	creds, err := cs.Kube.CoreV1().Secrets(DefaultVaultNamespace).Get(ctx, DefaultVaultCredentialsSecret, metav1.GetOptions{})
+	require.NoError(t, err, "failed to read %s/%s secret (was the vault-install CI step run?)", DefaultVaultNamespace, DefaultVaultCredentialsSecret)
+
+	require.Contains(t, creds.Data, "role-id", "vault-credentials secret is missing the role-id key")
```
ardaguclu: I assume that require.Contains is smart enough to check that creds.Data map contains role-id key.

test/library/encryption/kms/provider.go
p0lyn0mial: then let's extend all providers, not only KMS.

for example https://github.com/openshift/library-go/blob/bd1ee2bc0d67c374f0c90acd889f934f55cbacf5/test/library/encryption/scenarios.go#L235

this field could be extended by a struct that holds `configv1.APIServerEncryption` and a custom setup function. WDYT

test/library/encryption/kms/vault.go
p0lyn0mial: ah, ok, thx.

test/library/encryption/kms/vault.go
gangwgr: yes

test/library/encryption/scenarios.go
```diff
@@ -103,12 +111,16 @@ type testStep struct {
 }
 
 func TestEncryptionTurnOnAndOff(t testing.TB, scenario OnOffScenario) {
+	if scenario.EncryptionProvider.Setup != nil {
+		scenario.EncryptionProvider.Setup(t)
```
p0lyn0mial: IIRC all tests use `SetAndWaitForEncryptionType`. So I think this should be the only place that calls the setup fn.

test/library/encryption/kms/vault.go
```diff
@@ -1,11 +1,61 @@
...
+
 	configv1 "github.com/openshift/api/config/v1"
+	library "github.com/openshift/library-go/test/library/encryption"
+)
+
+const (
+	DefaultVaultNamespace         = "vault-kms"
+	DefaultVaultCredentialsSecret = "vault-credentials"
+	DefaultVaultAppRoleSecretName = "vault-approle-secret"
+	DefaultVaultKMSPluginImage    = "quay.io/openshifttest/mock-kms-plugin@sha256:03bb07a2c08b509653c4c70217a06a4b389c10b4d87922f50ee5eac82db5e140"
```
p0lyn0mial: is this a Vault img ? or it will be in the future ?

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +70,38 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
 	},
 }
+
+// EnsureDefaultVaultAppRoleSecret reads credentials from the vault-credentials secret
+// (created by a CI step) and creates the AppRole secret in openshift-config
+// using the default configuration constants.
+// If the secret already exists it is left untouched.
+func EnsureDefaultVaultAppRoleSecret(t testing.TB) {
+	t.Helper()
+	ctx := context.Background()
```
p0lyn0mial: use t.Context()

test/library/encryption/kms/vault.go
gangwgr: it is vault fake image which we are current using, in future we will update with real vault plugin image

test/library/encryption/kms/vault.go
gangwgr: updated

test/library/encryption/scenarios.go
gangwgr: updated

test/library/encryption/helpers.go
```diff
@@ -56,8 +56,14 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
-func SetAndWaitForEncryptionType(t testing.TB, provider configv1.APIServerEncryption, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
+func SetAndWaitForEncryptionType(t testing.TB, ep EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
 	t.Helper()
+
+	if ep.Setup != nil {
```
p0lyn0mial: I think this should moved down around the line 77, no ?

test/library/encryption/kms/vault.go
p0lyn0mial: do we have to compare the content ? or is it ok to assume the default will not change ?

gangwgr:
https://github.com/openshift/library-go/pull/2225#discussion_r3259333158
Better to compare

test/library/encryption/helpers.go
gangwgr: updated

test/library/encryption/helpers.go
```diff
@@ -56,8 +56,11 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
-func SetAndWaitForEncryptionType(t testing.TB, provider configv1.APIServerEncryption, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
+func SetAndWaitForEncryptionType(t testing.TB, ep EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
 	t.Helper()
+
+	provider := ep.APIServerEncryption
```
p0lyn0mial: let's remove this var.

test/library/encryption/helpers.go
```diff
@@ -56,8 +56,11 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
-func SetAndWaitForEncryptionType(t testing.TB, provider configv1.APIServerEncryption, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
+func SetAndWaitForEncryptionType(t testing.TB, ep EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
```
p0lyn0mial: rename ep to provider.

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +69,46 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
...
+
+	creds, err := cs.Kube.CoreV1().Secrets(DefaultVaultNamespace).Get(ctx, DefaultVaultCredentialsSecret, metav1.GetOptions{})
+	require.NoError(t, err, "failed to read %s/%s secret (was the vault-install CI step run?)", DefaultVaultNamespace, DefaultVaultCredentialsSecret)
+
+	expectedData := map[string][]byte{
+		"roleID":   creds.Data["role-id"],
+		"secretID": creds.Data["secret-id"],
+	}
+
+	existing, err := cs.Kube.CoreV1().Secrets(DefaultAppRoleTargetNamespace).Get(ctx, DefaultVaultAppRoleSecretName, metav1.GetOptions{})
```
p0lyn0mial: could this be replaced by `resourceapply.ApplySecret` ?

gangwgr: yes, updated

test/library/encryption/scenarios.go
```diff
@@ -232,12 +242,14 @@ type RotationScenario struct {
...
 	GetRawResourceFunc    func(t testing.TB, clientSet ClientSet, namespace string) string
 	UnsupportedConfigFunc UpdateUnsupportedConfigFunc
-	EncryptionProvider    configv1.APIServerEncryption
+	EncryptionProvider    EncryptionProvider
 }
 
 // TestEncryptionRotation first encrypts data with aescbc key
 // then it forces a key rotation by setting the "encyrption.Reason" in the operator's configuration file
 func TestEncryptionRotation(t testing.TB, scenario RotationScenario) {
+	provider := scenario.EncryptionProvider
```
p0lyn0mial: rm this variable.

test/library/encryption/scenarios.go
```diff
@@ -194,13 +203,14 @@ func TestEncryptionProvidersMigration(t testing.TB, scenario ProvidersMigrationS
 
 	// step 2: migrate through each provider in sequence
 	for i, provider := range scenario.EncryptionProviders {
+		p := provider
```
p0lyn0mial: remove this variable

test/library/encryption/scenarios.go
```diff
@@ -103,12 +111,13 @@ type testStep struct {
 }
 
 func TestEncryptionTurnOnAndOff(t testing.TB, scenario OnOffScenario) {
+	provider := scenario.EncryptionProvider
```
p0lyn0mial: remove this variable

test/library/encryption/kms/vault.go
```diff
@@ -1,11 +1,62 @@
...
+// DefaultVaultEncryptionProvider is a ready-to-use Vault KMS EncryptionProvider for e2e tests.
+// It bundles the default config with the AppRole secret setup.
+var DefaultVaultEncryptionProvider = library.EncryptionProvider{
+	APIServerEncryption: DefaultVaultKMSEncryptionConfig,
+	Setup:               EnsureDefaultVaultAppRoleSecret,
+}
+
+// DefaultVaultKMSEncryptionConfig is the standard Vault KMS encryption config
+// used by CI e2e tests. Consumers can use this directly or copy/modify as needed.
+var DefaultVaultKMSEncryptionConfig = configv1.APIServerEncryption{
```
p0lyn0mial: `DefaultVaultKMSPluginConfig`

test/library/encryption/kms/vault.go
```diff
@@ -1,11 +1,62 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/clock"
+
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	library "github.com/openshift/library-go/test/library/encryption"
+)
+
+const (
```
p0lyn0mial, line 18: do we have to export all these consts ?

gangwgr: all needed within  this file, so updated

test/library/encryption/kms/vault.go
```diff
@@ -20,3 +71,31 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
 		TransitKey: "test-transit-key",
 	},
 }
+
+// EnsureDefaultVaultAppRoleSecret reads credentials from the vault-credentials secret
+// (created by a CI step) and applies the AppRole secret in openshift-config
+// using the default configuration constants.
+func EnsureDefaultVaultAppRoleSecret(t testing.TB) {
```
p0lyn0mial: maybe we could make this function private for now ?

gangwgr: done

PR #2231, opened 2026-05-18
NO-JIRA: Keep equality during roundtrips of EncryptionConfiguration

ardaguclu:
Since Encryption of Config struct is not set TypeMeta, it won't be equal to the EncryptionConfiguration returned from encryption-config Secret. This does not create a bug, because we do all comparisons against `Encryption.Resources` not `Encryption`.

However, conceptually this is not right. After the roundtrips, we must have same data. Therefore, this PR adds TypeMeta to EncryptionConfiguration resources to keep it consistent as well as a test demonstrating the issue (test fails without the patch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve resource metadata (kind and API version) when converting encryption configurations and state so settings remain intact across round-trip conversions and comparisons.

* **Tests**
  * Added and updated tests to validate round-trip conversion and ensure expected configurations now include and match the preserved resource metadata.

### DIFF
--- a/pkg/operator/encryption/encryptiondata/config.go
+++ b/pkg/operator/encryption/encryptiondata/config.go
@@ -8,6 +8,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


ardaguclu:
encryption-config Secret stores the encoded EncryptionConfiguration with TypeMeta. So this PR should be backwards compatible.

ardaguclu:
Opened this https://github.com/openshift/cluster-kube-apiserver-operator/pull/2157 to test

ardaguclu:
I've experimented this idea and it works well. However, there is no need to ask more review time from people for such hypothetical fixes. We can fix it anytime we need it.
/close

pkg/operator/encryption/encryptiondata/config.go
```diff
@@ -76,7 +77,10 @@ func FromEncryptionState(encryptionState map[schema.GroupResource]state.GroupRes
 	})
 
 	return &Config{
-		Encryption: &apiserverconfigv1.EncryptionConfiguration{Resources: resourceConfigs},
+		Encryption: &apiserverconfigv1.EncryptionConfiguration{
+			TypeMeta:  metav1.TypeMeta{Kind: "EncryptionConfiguration", APIVersion: "apiserver.config.k8s.io/v1"},
```
bertinatto, line 81: nice catch. I think having a constructor for `Config` (where TypeMeta would be enforced) would prevent this from being added again in the future.

PR #2232, opened 2026-05-18
CNTRLPLANE-3361: pluginlifecycle: treat transit-mount as required

flavianmissi:
the api has changed to make this field required, so we change to reflect it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sidecar startup now consistently includes and requires the -transit-mount parameter; argument ordering for sidecar startup has been standardized.

* **Tests**
  * Test expectations updated to match the new argument order.

* **Chores**
  * Mock plugin wrapper CLI now documents and enforces transit-mount as required during startup.


bertinatto (commented):
/lgtm

and 

/hold

for https://github.com/openshift/api/pull/2847. We need that first because otherwise we could pass `-transit-mount=""` to the plugin, which would then fail to start. Not a big deal, but it'd break KMS tests in other PRs.

ardaguclu:
I think this PR needs to be rebased and merged soon. Or @bertinatto maybe you can add this in one of your PRs.

flavianmissi:
@bertinatto the api change has merged. is it best to bump the api here before we merge this one as well?

flavianmissi:
Oh, I missed the rebase bits - will take care of it asap.

flavianmissi:
I've rebased and also bumped openshift/api for good measure. Unit tests passed locally.

bertinatto (commented):
/lgtm

IIUC since the api changes merged, it's currently not possible to have this field empty. As a result, I think it's safe to pass the transit-mount parameter to the plugin sidecar.

bertinatto:
> /hold
> 
> for [openshift/api#2847](https://github.com/openshift/api/pull/2847). We need that first because otherwise we could pass `-transit-mount=""` to the plugin, which would then fail to start. Not a big deal, but it'd break KMS tests in other PRs.

/hold cancel

PR #2233, opened 2026-05-18
CNTRLPLANE-3236: kms: plugin sidecar should be an init container

bertinatto:
In recent runs of KMS tests I noticed that, upon receiving the termination signal from the kubelet, the KMS plugin sidecar would terminate right away. This caused a problem for the `kube-apiserver` that was still gracefully handling in-flight requests before shutting down, as it still needed the plugin to be running and serving requests via the UDS path.

This patch uses the Sidecar Containers feature [1] from Kubernetes to handle this issue. Now the KMS plugin is an init container with restart policy set to Always. This prevents the plugin from shutting down before kube-apiserver.

[1] https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure KMS plugin socket mount is added to both init and application containers and that the socket volume is created reliably, improving sidecar injection and migration robustness.
  * Sidecar containers now use a RestartPolicy of Always for improved reliability.

* **Tests**
  * Tests updated to assert RestartPolicy and verify correct init-container injection, ordering, and replacement during migrations.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -102,14 +102,17 @@ func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, 
 			return err
 		}
 
-		if err := ensureSocketVolumeMount(podSpec, sidecarProvider.Name()); err != nil {
+		if err := ensureSocketVolumeMountInContainer(podSpec.InitContainers, sidecarProvider.Name()); err != nil {
 			return err
 		}
 	}
 
-	if err := ensureSocketVolumeMount(podSpec, containerName); err != nil {
+	if err := ensureSocketVolumeMountInContainer(podSpec.Containers, containerName); err != nil {
 		return err
 	}
+
+	// The volume mount in the kube-apiserver and KMS plugin containers requires a volume in the podSpec
+	ensureSocketVolume(podSpec)
 
 	return nil
 }
@@ -120,20 +123,20 @@ func ensureSidecarContainer(podSpec *corev1.PodSpec, provider sidecarProvider) e
 		return fmt.Errorf("failed to build sidecar container: %w", err)
 	}
 
-	for i, container := range podSpec.Containers {
+	for i, container := range podSpec.InitContainers {
 		if container.Name == sidecar.Name {
-			podSpec.Containers[i] = sidecar
+			podSpec.InitContainers[i] = sidecar
 			return nil
 		}
 	}
 
-	podSpec.Containers = append(podSpec.Containers, sidecar)
+	podSpec.InitContainers = append(podSpec.InitContainers, sidecar)
 	return nil
 }
 
-func ensureSocketVolumeMount(podSpec *corev1.PodSpec, containerName string) error {
+func ensureSocketVolumeMountInContainer(containers []corev1.Container, containerName string) error {
 	containerIndex := -1
-	for i, container := range podSpec.Containers {
+	for i, container := range containers {
 		if container.Name == containerName {
 			containerIndex = i
 			break
@@ -144,8 +147,8 @@ func ensureSocketVolumeMount(podSpec *corev1.PodSpec, containerName string) erro
 		return fmt.Errorf("container %s not found", containerName)
 	}
 
-	container := &podSpec.Containers[containerIndex]
 	foundMount := false
+	container := &containers[containerIndex]
 	for _, m := range container.VolumeMounts {
 		if m.Name == "kms-plugin-socket" {
 			foundMount = true
@@ -160,9 +163,6 @@ func ensureSocketVolumeMount(podSpec *corev1.PodSpec, containerName string) erro
 			},
 		)
 	}
-
-	// The volume mount in the container requires a volume in the podSpec
-	ensureSocketVolume(podSpec)
 	return nil
 }
 
--- a/pkg/operator/encryption/kms/pluginlifecycle/vault.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/vault.go
@@ -6,6 +6,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/utils/ptr"
 )
 
 // newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
@@ -55,10 +56,13 @@ func (v *vault) BuildSidecarContainer() (corev1.Container, error) {
 	}
 
 	return corev1.Container{
-		Name:                     v.Name(),
-		Image:                    v.config.KMSPluginImage,
-		Args:                     args,
-		ImagePullPolicy:          corev1.PullIfNotPresent,
+		Name:            v.Name(),
+		Image:           v.config.KMSPluginImage,
+		Args:            args,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		// We place the container in InitContainers with RestartPolicyAlways so the kubelet starts it before
+		// regular containers and keeps it running for the pod's lifetime.
+		RestartPolicy:            ptr.To(corev1.ContainerRestartPolicyAlways),
 		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		// TODO(bertinatto): the plugin sidecar needs to be measure under heavy load to figure out good defaults.
 		// For now follow what most sidecars in the kube-apiserver pod do. xref:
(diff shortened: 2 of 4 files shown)


pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -120,20 +123,20 @@ func ensureSidecarContainer(podSpec *corev1.PodSpec, provider sidecarProvider) e
...
 		if container.Name == sidecar.Name {
-			podSpec.Containers[i] = sidecar
+			podSpec.InitContainers[i] = sidecar
 			return nil
 		}
 	}
 
-	podSpec.Containers = append(podSpec.Containers, sidecar)
+	podSpec.InitContainers = append(podSpec.InitContainers, sidecar)
 	return nil
```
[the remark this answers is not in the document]
bertinatto: It's not possible to have duplicate containers in `Containers` because every time the injection function runs, it gets a fresh podSpec.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -54,11 +54,13 @@ func (v *vault) BuildSidecarContainer() (corev1.Container, error) {
 		args = append(args, fmt.Sprintf("-transit-mount=%s", v.config.TransitMount))
 	}
 
+	restartPolicyAlways := corev1.ContainerRestartPolicyAlways
 	return corev1.Container{
 		Name:                     v.Name(),
 		Image:                    v.config.KMSPluginImage,
 		Args:                     args,
 		ImagePullPolicy:          corev1.PullIfNotPresent,
+		RestartPolicy:            &restartPolicyAlways,
```
ardaguclu:
```suggestion
		RestartPolicy:            ptr.To(corev1.ContainerRestartPolicyAlways),
```

ardaguclu: Better to document why this value is set.

ardaguclu (commented):
I read this sidecar containers [documentation](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/). Running kms plugins as sidecar containers does not only fix the graceful shutdown problem described in the PR but also guarantees that kms plugins will be run before the kube-apiserver. That inherently resolves the transient disruptions that might occur if kube-apiserver is run before kms plugins. 

So overall idea makes sense to me. But we'll need inputs from @benluddy @p0lyn0mial about hidden implications of this approach (if there is any?).

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -54,11 +54,13 @@ func (v *vault) BuildSidecarContainer() (corev1.Container, error) {
...
 		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 		// TODO(bertinatto): the plugin sidecar needs to be measure under heavy load to figure out good defaults.
```
p0lyn0mial, line 67: could we ask the Vault team ?

p0lyn0mial, line 67: would it also make sense to ask the team about `readiness and liveness` probes for the plugin?

p0lyn0mial (commented):
does this PR affect `terminationGracePeriodSeconds` ( 
https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/bindata/assets/kube-apiserver/pod.yaml#L291C3-L291C32) ?

if the server uses most or all of {{.GracefulTerminationDuration}} to drain connections, the KMS plugin sidecar gets SIGTERM followed almost immediately by SIGKILL no time for graceful cleanup.

p0lyn0mial:
> I read this sidecar containers [documentation](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/). Running kms plugins as sidecar containers does not only fix the graceful shutdown problem described in the PR but also guarantees that kms plugins will be run before the kube-apiserver. That inherently resolves the transient disruptions that might occur if kube-apiserver is run before kms plugins.

> So overall idea makes sense to me. But we'll need inputs from @benluddy @p0lyn0mial about hidden implications of this approach (if there is any?).


An alternative is to capture SIGTERM or modify the plugin to allow for a grace period before exiting.

I don't know whether there are any implications to using sidecar containers. I don't have experience with them. Maybe @ingvagabund  does?

From what I understand, the sidecar container will be started before the server and can be restarted independently. During shutdown, it will be terminated after the server. All of these properties are useful for us.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: ask the team if they have any recommendation

bertinatto:
> An alternative is to capture SIGTERM or modify the plugin to allow for a grace period before exiting.

that's correct, we talked a bit about these alternatives during standup today. A positive thing about using a sidecar is that it's plugin-agnostic, as it doesn't need any extra machinery or changes in plugins. It only makes sure plugins start and stop in the right order.

also, we learned that Hypershift uses the sidecar approach as well, so the approach is battle-tested.

> From what I understand, the sidecar container will be started before the server and can be restarted independently. During shutdown, it will be terminated after the server. All of these properties are useful for us.

that's my understanding as well. IMO it's worth giving it a shot.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: sorry, I meant to say that I **asked** the team. I read it now again and it seemed I told you to ask them yourself :D

bertinatto:
> does this PR affect `terminationGracePeriodSeconds` ( https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/bindata/assets/kube-apiserver/pod.yaml#L291C3-L291C32) ?

I don't have an answer right now, I'll need to do some research and get back to you

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: `ptr` looks better indeed

bertinatto: also, added a comment explaining the reason for the `RestartPolicyAlways`

bertinatto:
KMS tests in the proof PR are green with this change: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2146/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-gcp-operator-encryption-kms/2056406860012982272

ingvagabund:
> I don't know whether there are any implications to using sidecar containers. I don't have experience with them. Maybe @ingvagabund does?

Unfortunately I do not have. What I can think of is synchronization. If the sidecar needs to be fully initialized before the main container starts you will need to implement some waiting/pulling mechanism. And be more resilient to the sidecar container restarts.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: asked the Vault plugin team on the Slack channel

bertinatto:
> does this PR affect `terminationGracePeriodSeconds` ( https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/bindata/assets/kube-apiserver/pod.yaml#L291C3-L291C32) ?
> 
> if the server uses most or all of {{.GracefulTerminationDuration}} to drain connections, the KMS plugin sidecar gets SIGTERM followed almost immediately by SIGKILL no time for graceful cleanup.

my understanding is that this PR does not affect `terminationGracePeriodSeconds`. The idea of this patch is to allow the plugin to be alive while the kube-apiserver is draining the requests. By the time the plugin gets a SIGKILL, the kube-apiserver is already gone, so in my understanding that should be fine because the plugin is supposed to be stateless and there shouldn't be any clean-up necessary. I sent a message to the Vault team to confirm.

p0lyn0mial:
> my understanding is that this PR does not affect terminationGracePeriodSeconds. The idea of this patch is to allow the plugin to be alive while the kube-apiserver is draining the requests. By the time the plugin gets a SIGKILL, the kube-apiserver is already gone, so in my understanding that should be fine because the plugin is supposed to be stateless and there shouldn't be any clean-up necessary. I sent a message to the Vault team to confirm.

we talked about it and there is a potential issue if the server uses most or all of {{.GracefulTerminationDuration}} to drain connections. then the plugin will not get any grace time. @bertinatto will have a chat with the vault team about it.

i have also added the todo to set the `terminationGracePeriodSeconds ` to the work items ([xref](https://docs.google.com/document/d/1htjUQmCk60AVkcOe1QzLSYNrBhO7T8hVsFDiTbR4AJM/edit?tab=t.0#heading=h.7q1jphnlk300))

p0lyn0mial:
/lgtm

/hold
for testing

PR #2235, opened 2026-05-19
test/kms: replace t.Context() with context.Background() for Ginkgo compatibility

gangwgr:
test/kms: replace t.Context() with context.Background() for Ginkgo compatibility

## Summary by CodeRabbit

* **Tests**
  * Improved context handling in encryption tests so Vault and Kubernetes operations use an explicit context for secret and key actions.
  * Updated encryption provider setup hooks to accept and receive explicit context parameters, improving control and isolation during test runs.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -58,6 +58,7 @@ type UpdateUnsupportedConfigFunc func(raw []byte) error
 
 func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
 	t.Helper()
+	ctx := context.TODO()
 
 	t.Logf("Starting encryption e2e test for %q mode", provider.Type)
 
@@ -65,16 +66,16 @@ func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defa
 	lastMigratedKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube, namespace, labelSelector)
 	require.NoError(t, err)
 
-	apiServer, err := clientSet.ApiServerConfig.Get(context.TODO(), "cluster", metav1.GetOptions{})
+	apiServer, err := clientSet.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
 	require.NoError(t, err)
 	needsUpdate := !equality.Semantic.DeepEqual(apiServer.Spec.Encryption, provider.APIServerEncryption)
 	if needsUpdate {
 		if provider.Setup != nil {
-			provider.Setup(t)
+			provider.Setup(ctx, t)
 		}
 		t.Logf("Updating encryption configuration for APIServer from %#v to %#v", apiServer.Spec.Encryption, provider.APIServerEncryption)
 		apiServer.Spec.Encryption = provider.APIServerEncryption
-		_, err = clientSet.ApiServerConfig.Update(context.TODO(), apiServer, metav1.UpdateOptions{})
+		_, err = clientSet.ApiServerConfig.Update(ctx, apiServer, metav1.UpdateOptions{})
 		require.NoError(t, err)
 	} else {
 		t.Logf("APIServer is already configured to use %q mode", provider.Type)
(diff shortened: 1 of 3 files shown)


test/library/encryption/kms/vault.go
```diff
@@ -85,7 +85,7 @@ var DefaultFakeKMSPluginConfig = configv1.KMSPluginConfig{
 // using the default configuration constants.
 func ensureDefaultVaultAppRoleSecret(t testing.TB) {
 	t.Helper()
-	ctx := t.Context()
+	ctx := context.Background()
```
p0lyn0mial: why ? isn't ginkgo providing a ctx already ?

ardaguclu: Interesting... [Run func](https://github.com/golang/go/blob/45f1313c189bc51de0484c76fa77efc424c6b969/src/testing/testing.go#L2225) populates the `ctx`. However, it seems like Ginkgo does not call run?.

ardaguclu: I think, [g.GinkgoTB](https://github.com/openshift/cluster-kube-apiserver-operator/blob/51b3e923b0fa74b36bd60b21db8c8202d358d0b6/test/e2e-encryption-kms/encryption_kms.go#L24) does not have anything about `Context`. 

It would be desirable to populate t.ctx and wire it to everywhere. So that when the context is cancelled, test cleans up. But as far as I understand, it is not that easy.

ardaguclu: From what I understand from the code base, ginkgo does not provide a context.

gangwgr: Yes @ardaguclu  is correct GinkgoTB doesn't wire up t.Context()

gangwgr: yes

p0lyn0mial: ok, could we then pass the context as an arg to the functions ?

that would require changing the test signatures too, e.g.:
```
g.It("...", func(ctx context.Context) {
      ensureDefaultVaultAppRoleSecret(ctx, g.GinkgoTB(), ...)
  })
```

gangwgr: Yes, ctx from g.It(func(ctx context.Context) {...})  it requires changing the EncryptionProvider.Setup signature from func(t testing.TB) to func(ctx context.Context, t testing.TB), plus all the TestEncryption* functions in library-go that call Setup.

test/library/encryption/scenarios.go
```diff
@@ -30,7 +31,10 @@ type BasicScenario struct {
 type EncryptionProvider struct {
 	configv1.APIServerEncryption
 	// Setup is called once before the provider is first used. May be nil.
-	Setup func(t testing.TB)
+	// The context is provided by the test framework (e.g. Ginkgo) and should be
```
ardaguclu: Context is not provided by the test framework

test/library/encryption/helpers.go
```diff
@@ -70,7 +70,7 @@ func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defa
 	needsUpdate := !equality.Semantic.DeepEqual(apiServer.Spec.Encryption, provider.APIServerEncryption)
 	if needsUpdate {
 		if provider.Setup != nil {
-			provider.Setup(t)
+			provider.Setup(context.TODO(), t)
```
ardaguclu: I think, it is good time to unify the context usage in that helpers.go.

We can have one ctx := context.TODO. We should pass same ctx to everywhere including the GetLastMetaKey, etc.

gangwgr:
/hold
testing in kas-o pr https://github.com/openshift/cluster-kube-apiserver-operator/pull/2153

ardaguclu:
I think there is no need to test this PR on kas-o. It wires the context.
/hold cancel

test/library/encryption/helpers.go
p0lyn0mial: sgtm

p0lyn0mial: can be done in a follow up PRs.

test/library/encryption/helpers.go
```diff
@@ -58,23 +58,24 @@ type UpdateUnsupportedConfigFunc func(raw []byte) error
 
 func SetAndWaitForEncryptionType(t testing.TB, provider EncryptionProvider, defaultTargetGRs []schema.GroupResource, namespace, labelSelector string) ClientSet {
 	t.Helper()
+	ctx := context.TODO()
```
p0lyn0mial, line 61: could we pass the ctx as an arg ?

gangwgr: @p0lyn0mial (https://redhat.enterprise.slack.com/team/UK6NS15HS) SetAndWaitForEncryptionType's signature would cascade to all test functions do we need this change https://github.com/openshift/library-go/pull/2236/changes?

PR #2236, opened 2026-05-21
test/encryption: pass context.Context as an explicit argument

gangwgr:
Instead of creating context.TODO() internally in SetAndWaitForEncryptionType, accept ctx as a parameter so callers control the context lifecycle. This also removes the misleading comment about Ginkgo from the EncryptionProvider struct.
As per discussion https://github.com/openshift/library-go/pull/2235/changes#r3276734982

## Summary by CodeRabbit

* **Tests**
  * Refactored encryption test helpers to consistently manage test execution context, improving reliability and control over test operations across multiple encryption scenarios.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -56,9 +56,8 @@ type EncryptionKeyMeta struct {
 
 type UpdateUnsupportedConfigFunc func(raw []byte) error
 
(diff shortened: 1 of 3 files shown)


p0lyn0mial:
/lgtm

worth checking the other methods/functions if they also accept a ctx as a param

PR #2241, opened 2026-05-25
WIP: CNTRLPLANE-3237: In-place field update for KMS mode regardless of the convergence

ardaguclu:
When a KMS sidecar revision is stuck (e.g. wrong container image), the cluster-admin needs to update APIServer.spec.encryption.kms with corrected values. Previously, all encryption controllers blocked on revision convergence, so the fix could never be applied.

This change allows the key_controller to update the latest key secret's KMS plugin config in-place for non-migration fields (image, TLS, AppRole) regardless of convergence state. If there is a change in migration-triggering fields, we should wait until convergence happens.

State controller carries in-place fields updates only when there is no change in the encryption-config data key.

## Summary by CodeRabbit

* **Bug Fixes**
  * KMS plugin configuration now updates in-place during key transitions without recreating secrets, improving reliability during mode changes.

* **Refactor**
  * Enhanced encryption configuration application to support in-place updates while transitioning between encryption modes.

* **Tests**
  * Added test scenarios for KMS plugin configuration propagation during state transitions.
  * Enhanced end-to-end validation of plugin configuration changes and synchronization across key updates.

PR #2243, opened 2026-05-26
test/encryption: add InvalidImageRecoveryScenario for KMS plugin imag…

gangwgr:
…e fix

Adds a test scenario that verifies the cluster can recover when a KMS encryption config is set with an invalid plugin image:

1. Enable KMS with invalid (non-existent) plugin image
2. Attempt to switch to aescbc (cluster remains stuck on KMS)
3. Operator reports degraded
4. Fix the config with valid KMS plugin image
5. Operator recovers and encryption works normally

## Summary by CodeRabbit

* **New Features**
  * Operator now applies in-place KMS plugin config updates for the latest key when safe and propagates updated plugin data into the managed encryption state; skips propagation when not applicable or when migration is required.

* **Tests**
  * Expanded unit and e2e coverage for in-place updates, migration detection, retry-based verification, recovery from invalid KMS plugin images, and a test ensuring config helpers do not mutate inputs; added degraded/recovery wait scenarios and timeouts.

* **Chores**
  * Vault AppRole secret keys standardized to hyphenated names ("role-id", "secret-id").

### DIFF
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -8,10 +8,10 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/rand"
-
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	configv1 "github.com/openshift/api/config/v1"
 )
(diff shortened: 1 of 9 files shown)


test/library/encryption/scenarios.go
```diff
@@ -277,3 +277,81 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+// InvalidImageRecoveryScenario tests that an invalid KMS plugin image causes
+// the cluster to degrade and that correcting the image restores normal operation.
+//
+// Arda's scenario:
+//  1. Enable KMS with invalid image
+//  2. Attempt to switch to aescbc (cluster should be stuck on KMS)
+//  3. See degraded
+//  4. Update with KMS with valid image
+//  5. See everything works
+type InvalidImageRecoveryScenario struct {
```
ardaguclu, line 290: I think this scenario nicely exercises the in-place field update case which will be added by https://github.com/openshift/library-go/pull/2241.

PR #2245, opened 2026-05-26
CNTRLPLANE-3513: prepare rotation test for KMS

tjungblu:
This extracts the force rotation and waiting for new key aspect out of the current rotation test.

## Summary by CodeRabbit

* **Tests**
  * Added modular test helpers and callbacks for triggering and awaiting encryption key rotations; rotation scenarios updated to use these hooks.
  * Improves reliability and flexibility across different key-management backends.
  * No end-user visible changes; internal testing only.

### DIFF
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -116,13 +116,17 @@ func ensureDefaultVaultAppRoleSecret(ctx context.Context, t testing.TB) {
 	t.Logf("Applied AppRole secret %s in %s (changed=%v)", defaultVaultAppRoleSecretName, defaultAppRoleTargetNamespace, changed)
 }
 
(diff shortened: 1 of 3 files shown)


test/library/encryption/kms/vault.go
```diff
@@ -116,13 +116,17 @@ func ensureDefaultVaultAppRoleSecret(ctx context.Context, t testing.TB) {
 	t.Logf("Applied AppRole secret %s in %s (changed=%v)", defaultVaultAppRoleSecretName, defaultAppRoleTargetNamespace, changed)
 }
 
+func ForceVaultKeyRotation() library.ForceRotationFunc {
```
tjungblu, line 119: if you wonder, this useless proxy function is just here because of cyclic dependencies between the kms and encryption package

PR #2246, opened 2026-05-28
Remove old plugin setup

gangwgr:
Remove old plugin setup

## Release Notes

* **Chores**
  * Removed mock KMS plugin deployment infrastructure files.
  * Cleaned up test asset generation utilities and documentation.

* **Tests**
  * Made vault namespace flag required in test configuration.
  * Updated default KMS plugin image configuration reference.

### DIFF
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -106,8 +106,8 @@ func ensureDefaultVaultAppRoleSecret(ctx context.Context, t testing.TB) {
 		},
 		Type: corev1.SecretTypeOpaque,
 		Data: map[string][]byte{
-			"roleID":   creds.Data["role-id"],
-			"secretID": creds.Data["secret-id"],
+			"role-id":   creds.Data["role-id"],
+			"secret-id": creds.Data["secret-id"],
 		},
 	}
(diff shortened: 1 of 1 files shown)


test/library/encryption/kms/k8s-mock-plugin/wrapper/main.go
```diff
@@ -1,119 +0,0 @@
-package main
```
ardaguclu, line 1: We need fake plugin. We will use it in kms to kms migrations.

test/library/encryption/kms/vault.go
```diff
@@ -75,7 +75,7 @@ var DefaultFakeKMSPluginConfig = configv1.APIServerEncryption{
 	KMS: configv1.KMSPluginConfig{
 		Type: configv1.VaultKMSProvider,
 		Vault: configv1.VaultKMSPluginConfig{
-			KMSPluginImage: WellKnownUpstreamMockKMSPluginImage,
+			KMSPluginImage: defaultVaultKMSPluginImage,
```
ardaguclu: We should still use fake plugin. Moving to original Vault should be in separate PR.

gangwgr: it is fake image in variable, same as old

PR #2249, opened 2026-05-29
test/encryption: add InvalidConfigRecoveryScenario for KMS plugin image

gangwgr:
test/encryption: add InvalidImageRecoveryScenario for KMS plugin image

## Summary by CodeRabbit

* **Tests**
  * Added new encryption recovery test scenarios for validating system behavior during invalid configuration states and recovery processes.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -136,13 +136,13 @@ func WaitForEncryptionKeyBasedOn(t testing.TB, kubeClient kubernetes.Interface, 
 	}
 
 	if prevKeyMeta.Mode == encryptionMode {
-		waitForNoNewEncryptionKey(t, kubeClient, prevKeyMeta, namespace, labelSelector)
+		WaitForNoNewEncryptionKey(t, kubeClient, prevKeyMeta, namespace, labelSelector)
 		return
 	}
 	WaitForNextMigratedKey(t, kubeClient, prevKeyMeta, defaultTargetGRs, namespace, labelSelector)
 }
 
-func waitForNoNewEncryptionKey(t testing.TB, kubeClient kubernetes.Interface, prevKeyMeta EncryptionKeyMeta, namespace, labelSelector string) {
+func WaitForNoNewEncryptionKey(t testing.TB, kubeClient kubernetes.Interface, prevKeyMeta EncryptionKeyMeta, namespace, labelSelector string) {
 	t.Helper()
 	// given that the happy path scenario needs ~30 min
 	// waiting 5 min to see if a new key hasn't been created seems to be enough.
@@ -371,5 +371,65 @@ func setUpTearDown(namespace string) func(testing.TB, bool) {
 				t.Logf("Last seen: %-15v Type: %-10v Reason: %-40v Source: %-55v Message: %v", now.Sub(ev.LastTimestamp.Time), ev.Type, ev.Reason, ev.Source.Component, ev.Message)
 			}
 		}
+	}
+}
+
+// WaitForPodImagePullBackOff polls pods in the given namespace until at least one pod
+// has a container stuck in ImagePullBackOff or ErrImagePull.
+// When a KMS plugin image is invalid, the operator cannot progress because the
+// sidecar container fails to start. The operator does not report Degraded in this
+// case — it only gets stuck in Progressing. This function detects that stuck state
+// by observing the pod-level image pull failure, which is the earliest signal that
+// the invalid config has taken effect.
+func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, labelSelector string, timeout time.Duration) {
+	t.Helper()
+	t.Logf("Waiting up to %s for a pod in %s (selector=%s) to enter ImagePullBackOff", timeout, namespace, labelSelector)
+	err := wait.PollUntilContextTimeout(ctx, waitPollInterval, timeout, true, func(ctx context.Context) (bool, error) {
+		pods, err := kubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			t.Logf("Error listing pods: %v", err)
+			return false, nil
+		}
+		for _, pod := range pods.Items {
+			for _, cs := range append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...) {
+				if cs.State.Waiting != nil && (cs.State.Waiting.Reason == "ImagePullBackOff" || cs.State.Waiting.Reason == "ErrImagePull") {
+					t.Logf("Pod %s container %s is in %s", pod.Name, cs.Name, cs.State.Waiting.Reason)
+					return true, nil
+				}
+			}
+		}
+		return false, nil
+	})
+	require.NoError(t, err, "timed out waiting for pod to enter ImagePullBackOff in namespace %s", namespace)
+}
+
+// WaitForCurrentKeyMigrated waits until the current (latest) encryption key has completed
+// migration for all target group resources. It fails if the key changes unexpectedly
+// (i.e. a different key than prevKeyMeta appears).
+func WaitForCurrentKeyMigrated(t testing.TB, kubeClient kubernetes.Interface, prevKeyMeta EncryptionKeyMeta, targetGRs []schema.GroupResource, namespace, labelSelector string) {
+	t.Helper()
+
+	t.Logf("Waiting up to %s for key %q to complete migration of %v", waitPollTimeout.String(), prevKeyMeta.Name, targetGRs)
+	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		currentKeyMeta, err := GetLastKeyMeta(t, kubeClient, namespace, labelSelector)
+		if err != nil {
+			return false, err
+		}
+		if currentKeyMeta.Name != prevKeyMeta.Name {
+			return false, fmt.Errorf("unexpected key observed %q, expected no new key", currentKeyMeta.Name)
+		}
+		if len(currentKeyMeta.Migrated) < len(targetGRs) {
+			return false, nil
+		}
+		for _, expectedGR := range targetGRs {
+			if !hasResource(expectedGR, currentKeyMeta.Migrated) {
+				return false, nil
+			}
+		}
+		t.Logf("Key %q has completed migration of %v", currentKeyMeta.Name, currentKeyMeta.Migrated)
+		return true, nil
+	}); err != nil {
+		newErr := fmt.Errorf("timed out waiting for key %q to complete migration of %v: %v", prevKeyMeta.Name, targetGRs, err)
+		require.NoError(t, newErr)
 	}
 }
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -8,8 +8,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -280,4 +280,76 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
 	}
 
 	// TODO: assert conditions - operator and encryption migration controller must report status as active not progressing, and not failing for all scenarios
+}
+
+// ApplyEncryption applies the given encryption config to apiserver/cluster
+// without waiting for completion.
+func ApplyEncryption(ctx context.Context, t testing.TB, encryption configv1.APIServerEncryption) {
+	t.Helper()
+	cs := GetClients(t)
+	apiServer, err := cs.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
+	require.NoError(t, err)
+	apiServer.Spec.Encryption = encryption
+	_, err = cs.ApiServerConfig.Update(ctx, apiServer, metav1.UpdateOptions{})
+	require.NoError(t, err)
+	t.Logf("Applied encryption config (type=%s)", encryption.Type)
+}
+
+type KMSInvalidEncryptionRecoveryScenario struct {
+	BasicScenario
+	InvalidProvider EncryptionProvider
+	ValidProvider   EncryptionProvider
+	WaitForStuck    func(ctx context.Context, t testing.TB)
+}
+
+// TestInvalidEncryptionRecovery validates recovery from an invalid encryption config:
+//  1. Apply invalid config — operator stuck (e.g. ImagePullBackOff, connection timeout)
+//  2. Switch to AESCBC — verify no new encryption key is created (revisions stuck)
+//  3. Apply valid config — verify recovery and successful encryption
+func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenario KMSInvalidEncryptionRecoveryScenario) {
+	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	clientSet := GetClients(e)
+
+	require.NotNil(t, scenario.InvalidProvider.Setup, "InvalidProvider.Setup must not be nil")
+	require.NotNil(t, scenario.ValidProvider.Setup, "ValidProvider.Setup must not be nil")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.InvalidProvider.Type, "InvalidProvider must use KMS encryption type")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.ValidProvider.Type, "ValidProvider must use KMS encryption type")
+
+	steps := []testStep{
+		{name: "ApplyInvalidConfig", testFunc: func(t testing.TB) {
+			scenario.InvalidProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.InvalidProvider.APIServerEncryption)
+		}},
+		{name: "WaitForStuck", testFunc: func(t testing.TB) {
+			scenario.WaitForStuck(ctx, t)
+		}},
+		{name: "SwitchToAESCBCAndVerifyNoNewKey", testFunc: func(t testing.TB) {
+			prevKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			ApplyEncryption(ctx, t, configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESCBC})
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, prevKeyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "ApplyValidConfigAndVerifyRecovery", testFunc: func(t testing.TB) {
+			scenario.ValidProvider.Setup(ctx, t)
+			prevKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			ApplyEncryption(ctx, t, scenario.ValidProvider.APIServerEncryption)
+			WaitForCurrentKeyMigrated(t, clientSet.Kube, prevKeyMeta,
+				scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
+			scenario.AssertFunc(t, clientSet, scenario.ValidProvider.Type,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+	}
+
+	for _, step := range steps {
+		t.Logf("=== STEP: %s ===", step.name)
+		step.testFunc(e)
+		if t.Failed() {
+			t.Errorf("stopping the test as %q step failed", step.name)
+			return
+		}
+	}
 }


test/library/encryption/kms/vault.go
```diff
@@ -33,6 +33,11 @@ const (
 	defaultVaultTransitKey        = "kms-key"
 	defaultAppRoleTargetNamespace = "openshift-config"
 	vaultCommandTimeout           = 30 * time.Second
+
+	// InvalidVaultKMSPluginImage is an OCI image reference that passes API validation
+	// (correct format, sha256 digest, sufficient length) but does not exist in any registry.
+	// Use this to test degradation when the KMS plugin image cannot be pulled.
+	InvalidVaultKMSPluginImage = "quay.io/openshifttest/mock-kms-plugin-nonexistent@sha256:0000000000000000000000000000000000000000000000000000000000000000"
```
ardaguclu: We don't need to define a new variable for this. We can just an invalid image where we need it.

test/library/encryption/kms/vault.go
```diff
@@ -47,6 +47,35 @@ var DefaultFakeVaultEncryptionProvider = library.EncryptionProvider{
 	Setup:               ensureDefaultVaultAppRoleSecret,
 }
 
+// InvalidImageVaultEncryptionProvider is a Vault KMS EncryptionProvider with a
+// non-existent plugin image. Use this for testing degradation when the image cannot be pulled.
+var InvalidImageVaultEncryptionProvider = library.EncryptionProvider{
```
ardaguclu: Do we really need this?. We can use the current valid encryption provider. After that we can modify it with invalid image.

test/library/encryption/kms/vault.go
```diff
@@ -47,6 +47,35 @@ var DefaultFakeVaultEncryptionProvider = library.EncryptionProvider{
...
+	APIServerEncryption: InvalidImageVaultKMSPluginConfig,
+	Setup:               ensureDefaultVaultAppRoleSecret,
+}
+
+// InvalidImageVaultKMSPluginConfig is identical to DefaultVaultKMSPluginConfig
+// but uses a non-existent image that will fail to pull.
+var InvalidImageVaultKMSPluginConfig = configv1.APIServerEncryption{
```
ardaguclu: I think we don't need this. We can use the current Vault definition.

test/library/encryption/helpers.go
```diff
@@ -346,3 +346,34 @@ func setUpTearDown(namespace string) func(testing.TB, bool) {
 		}
 	}
 }
+
+// WaitForPodImagePullBackOff polls pods in the given namespace until at least one pod
+// has an init container or container stuck in ImagePullBackOff. This is useful for
+// detecting that an invalid KMS plugin image is causing static pod failure.
+func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, labelSelector string, timeout time.Duration) {
```
ardaguclu, line 384: That is too specific error to be waiting. When kms plugin is unreachable, I'd expect that operator goes degraded. We should wait for the degradation instead of pod's specific condition. Because since apiserver can not communicate with the plugin, it should report unhealthy.

What is the result of your tests?. It didn't go degraded?.

test/library/encryption/scenarios.go
```diff
@@ -277,3 +277,64 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+//
+// Flow:
+//  1. Enable KMS with invalid (non-existent) plugin image
+//  2. Operator reports degraded (stuck NodeInstaller, ImagePullBackOff, etc.)
+//  3. Fix by applying KMS with valid image and wait for full encryption migration
+type InvalidImageRecoveryScenario struct {
+	BasicScenario
+	// InvalidImageProvider is the KMS EncryptionProvider configured with an invalid
+	// (non-existent or broken) KMS plugin image. Enabling this should cause degradation.
+	InvalidImageProvider EncryptionProvider
```
ardaguclu: I think we don't need this.

test/library/encryption/scenarios.go
```diff
@@ -277,3 +277,64 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+	apiServer, err := cs.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get APIServer config")
+	apiServer.Spec.Encryption = scenario.InvalidImageProvider.APIServerEncryption
+	_, err = cs.ApiServerConfig.Update(ctx, apiServer, metav1.UpdateOptions{})
+	require.NoError(t, err, "failed to update APIServer with invalid KMS image")
+
+	// step 2: wait for degraded — the operator is stuck because the invalid KMS image
+	// prevents the NodeInstaller from completing the rollout
+	t.Log("Waiting for operator to report degraded status")
+	scenario.WaitForDegraded(ctx, e)
```
ardaguclu: That should be sufficient instead of waiting for imagepullerr

test/library/encryption/helpers.go
ardaguclu: Also noone uses this function currently

gangwgr: In our tests, the operator does not go Degraded when the KMS plugin image is invalid. It just stays stuck in Progressing (NodeInstallerProgressing) because the static pod revision can never complete. The apiserver never actually starts with the bad plugin, so it never gets a chance to report unhealthy, the image pull fails before the pod runs.

gangwgr: if we are planning to fix that, so i removed imagepull func

test/library/encryption/scenarios.go
gangwgr: not sure it is sufficient, same reason https://github.com/openshift/library-go/pull/2249#discussion_r3332098592

test/library/encryption/scenarios.go
```diff
@@ -277,3 +276,46 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+// before calling TestEncryptionInvalidImageRecovery. This function then:
+//  1. Waits for the operator to report degraded
+//  2. Fixes the config with the valid KMS image and waits for full encryption migration
+type InvalidImageRecoveryScenario struct {
+	BasicScenario
+	// ValidImageProvider is the KMS EncryptionProvider configured with the correct
+	// KMS plugin image. Applying this after degradation should restore the cluster.
+	ValidImageProvider EncryptionProvider
+	// WaitForDegraded should block until the operator reports a degraded condition.
+	WaitForDegraded func(ctx context.Context, t testing.TB)
```
ardaguclu: Where is the implementation of this function?

ardaguclu: It will be assigned by the caller?

test/library/encryption/scenarios.go
```diff
@@ -277,3 +276,46 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+// by applying a valid KMS image via SetAndWaitForEncryptionType.
+func TestEncryptionInvalidImageRecovery(ctx context.Context, t testing.TB, scenario InvalidImageRecoveryScenario) {
+	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	require.NotNil(t, scenario.WaitForDegraded, "WaitForDegraded must not be nil")
+	require.NotNil(t, scenario.AssertFunc, "AssertFunc must not be nil")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.ValidImageProvider.Type, "ValidImageProvider must be KMS type")
+
+	// step 1: wait for degraded — the operator is stuck because the invalid KMS image
+	// prevents the NodeInstaller from completing the rollout
+	t.Log("Waiting for operator to report degraded status")
```
ardaguclu: We haven't set invalid image yet?

test/library/encryption/scenarios.go
gangwgr: Yes, it's assigned by the caller in kas-o

test/library/encryption/scenarios.go
gangwgr: we can set the invalid image by call in test

ardaguclu: Due to the nature of this Scenario, in my opinion we should set it here to be explicit about it.

gangwgr: we can add invalid image setup in library-go also, both works
do we need add here or keep it in caller?

ardaguclu: I think, we should have it here. This is the core part of the scenario

test/library/encryption/scenarios.go
```diff
@@ -277,3 +276,46 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+	// verify the cluster is still on KMS (the invalid image blocked any transition)
+	cs := GetClients(e)
+	t.Log("Verifying cluster remained on KMS encryption")
+	scenario.AssertFunc(e, cs, configv1.EncryptionTypeKMS, scenario.Namespace, scenario.LabelSelector)
+
+	// step 2: fix by applying KMS with valid image and wait for full recovery
+	t.Log("Recovering: applying KMS with valid plugin image and waiting for full migration")
+	SetAndWaitForEncryptionType(ctx, e, scenario.ValidImageProvider, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
+
+	t.Log("Invalid image recovery test passed")
```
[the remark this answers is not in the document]
gangwgr: @coderabbit The post-recovery operator health check is intentionally left to the caller (KAS-O test), not library-go. We discussed this before — the KAS-O test adds libraryapiserver.WaitForAPIServerToStabilizeOnTheSameRevision after calling TestEncryptionInvalidImageRecovery:

test/library/encryption/scenarios.go
gangwgr: added

ardaguclu:
@gangwgr you are right. It is not degrading, it stuck in progressing;

```
$ oc get co
NAME                                       VERSION                                                AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                             4.23.0-0-2026-06-01-074800-test-ci-ln-iqf6b22-latest   True        True          False      95m     APIServerDeploymentProgressing: deployment/apiserver.openshift-oauth-apiserver: 1/3 pods have been updated to the latest generation and 2/3 pods are available
kube-apiserver                             4.23.0-0-2026-06-01-074800-test-ci-ln-iqf6b22-latest   True        True          False      112m    NodeInstallerProgressing: 3 nodes are at revision 7; 0 nodes have achieved new revision 9
openshift-apiserver                        4.23.0-0-2026-06-01-074800-test-ci-ln-iqf6b22-latest   True        True          False      107m    APIServerDeploymentProgressing: deployment/apiserver.openshift-apiserver: 1/3 pods have been updated to the latest generation and 2/3 pods are available
```

but in reality pod is stuck in error;

```
$ oc get pods -n openshift-kube-apiserver
NAME                                                              READY   STATUS                  RESTARTS   AGE
kube-apiserver-ip-10-0-102-172.us-west-1.compute.internal         0/6     Init:ImagePullBackOff   0          8m11s
$ oc get pods -n openshift-apiserver
NAME                         READY   STATUS                  RESTARTS      AGE
apiserver-6c94754994-fks4q   0/3     Init:ImagePullBackOff   0             17m
$ oc get pods -n openshift-oauth-apiserver
NAME                         READY   STATUS                  RESTARTS   AGE
apiserver-65f95f8675-x77jh   0/2     Init:ImagePullBackOff   0          17m
```
So we should wait for ImagePullBackOff error as you did previously, after that patching with correct image and wait for the success.
/cc @p0lyn0mial

ardaguclu:
BTW, it depends on the field we use for in-place field updates. Image field gets this error but there might be another field that causes degradation.

test/library/encryption/scenarios.go
```diff
@@ -277,3 +277,57 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+
+// InvalidImageRecoveryScenario tests that an invalid KMS plugin image causes
+// the operator to degrade and that providing a valid KMS image restores
+// normal operation.
+//
+// Flow:
+//  1. Apply invalid KMS encryption config (causes operator to get stuck)
+//  2. Wait for degraded/stuck state
+//  3. Fix by applying valid KMS image and wait for full encryption migration
+type InvalidImageRecoveryScenario struct {
```
p0lyn0mial: wondering if we could create a more generic scenario that would accept an invalid configuration (not necessarily invalid image) and possibly inject the config at different stages: before encryption has been enabled, after encryption has been enabled, during migration etc.

WDYT ?

ardaguclu: I think, this is a good idea. Although currently we only have image field as in-place field (other fields in Vault are all trigerring migration), it would be better to have generic approach.

gangwgr: Yes, updated

ardaguclu: I think we can test the combinations in one scenario;
* Enable KMS with invalid image
* Update to aescbc and see that there is no switch to aescbc
* Patch KMS with correct image with previous configuration (because apiserver/cluster is on aescbc)
* See that everything works
* Patch again with invalid KMS image
* See that revision is stuck
* Patch with correct KMS image
* Immediately patch with incorrect KMS image -- during migration
* Patch with correct KMS image again

gangwgr: do we need multiple times invalid image? it will increase case duration

ardaguclu: I think we need to test all cases.

gangwgr: can we combine below also in one case?
1. Unreachable vault
2. Bad creds
3. invalid transit mount

ardaguclu: > Unreachable vault

This can be tested, if this is caused by invalid configuration

> Bad creds

Updating content of referenced Secrets hasn't been implemented yet. But when it is implemented, this can be a separate scenario.

> Invalid transit mount

Invalid transit mount needs to be caught by preflight checker.

test/library/encryption/scenarios.go
```diff
@@ -277,3 +277,121 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+
+// EncryptionConfigStep is a single named step in an EncryptionConfigScenario.
+type EncryptionConfigStep struct {
+	Name string
+	Run  func(ctx context.Context, t testing.TB)
+}
+
+// TestEncryptionConfigScenario executes each step in sequence, stopping
+// on the first failure.
+func TestEncryptionConfigScenario(ctx context.Context, t testing.TB, scenario EncryptionConfigScenario) {
```
ardaguclu: @gangwgr I've implemented my suggested changes in https://github.com/openshift/library-go/pull/2265 for all the changes in this PR. Could you please take a look at?

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,67 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+	InvalidProvider EncryptionProvider
+	ValidProvider   EncryptionProvider
+	WaitForStuck    func(ctx context.Context, t testing.TB)
+}
+
+// TestInvalidEncryptionRecovery validates recovery from an invalid encryption config:
+//  1. Apply invalid config — operator stuck (e.g. ImagePullBackOff, connection timeout)
+//  2. Switch to AESCBC — verify no new encryption key is created (revisions stuck)
+//  3. Apply valid config — verify recovery and successful encryption
+func TestInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenario InvalidEncryptionRecoveryScenario) {
```
ardaguclu: Let's fail if InvalidProvider.Setup or ValidProvider.Setup are nil

test/library/encryption/helpers.go
```diff
@@ -373,3 +373,58 @@ func setUpTearDown(namespace string) func(testing.TB, bool) {
 		}
 	}
 }
+
+// WaitForPodImagePullBackOff polls pods in the given namespace until at least one pod
+// has an init container or container stuck in ImagePullBackOff or ErrImagePull.
```
ardaguclu: Can you please document why we need to wait these errors?.

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,72 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+		{name: "SwitchToAESCBCAndVerifyNoNewKey", testFunc: func(t testing.TB) {
+			prevKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			ApplyEncryption(ctx, t, configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESCBC})
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, prevKeyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "ApplyValidConfigAndVerifyRecovery", testFunc: func(t testing.TB) {
+			if scenario.ValidProvider.Setup != nil {
```
ardaguclu: This one is redundant too

test/library/encryption/scenarios.go
ardaguclu: This scenario will only work with KMS encryption modes. Others do not support this functionality. Should we rename it to `TestKMSInvalidEncryptionRecovery`?

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,72 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+
+// TestInvalidEncryptionRecovery validates recovery from an invalid encryption config:
+//  1. Apply invalid config — operator stuck (e.g. ImagePullBackOff, connection timeout)
+//  2. Switch to AESCBC — verify no new encryption key is created (revisions stuck)
+//  3. Apply valid config — verify recovery and successful encryption
+func TestInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenario InvalidEncryptionRecoveryScenario) {
+	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	clientSet := GetClients(e)
+
+	require.NotNil(t, scenario.InvalidProvider.Setup, "InvalidProvider.Setup must not be nil")
```
ardaguclu, line 313: Does it make sense we validate encryption modes are KMS for invalidprovider and validprovider.

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,70 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+	cs := GetClients(t)
+	apiServer, err := cs.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
+	require.NoError(t, err)
+	apiServer.Spec.Encryption = encryption
+	_, err = cs.ApiServerConfig.Update(ctx, apiServer, metav1.UpdateOptions{})
+	require.NoError(t, err)
+	t.Logf("Applied encryption config (type=%s)", encryption.Type)
+}
+
+type InvalidEncryptionRecoveryScenario struct {
```
ardaguclu:
```suggestion
type KMSInvalidEncryptionRecoveryScenario struct {
```

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,70 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			ApplyEncryption(ctx, t, configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESCBC})
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, prevKeyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "ApplyValidConfigAndVerifyRecovery", testFunc: func(t testing.TB) {
+			scenario.ValidProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.ValidProvider.APIServerEncryption)
+			WaitForCurrentKeyMigrated(t, clientSet.Kube,
```
ardaguclu: We should pass preKeyMeta to this function and compare it with currentKeyMeta. They must be equal, otherwise we should fail.

test/library/encryption/helpers.go
```diff
@@ -373,3 +373,63 @@ func setUpTearDown(namespace string) func(testing.TB, bool) {
...
+func WaitForCurrentKeyMigrated(t testing.TB, kubeClient kubernetes.Interface, targetGRs []schema.GroupResource, namespace, labelSelector string) {
+	t.Helper()
+
+	t.Logf("Waiting up to %s for current encryption key to complete migration of %v", waitPollTimeout.String(), targetGRs)
+	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+		currentKeyMeta, err := GetLastKeyMeta(t, kubeClient, namespace, labelSelector)
+		if err != nil {
+			return false, err
+		}
+		if len(currentKeyMeta.Name) == 0 {
```
ardaguclu:
```go
		if currentKeyMeta.Name != prevKeyMeta.Name {
			return false, fmt.Errorf("unexpected key observed %q, expected no new key", currentKeyMeta.Name)
		}
```

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,83 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+			prevKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			ApplyEncryption(ctx, t, configv1.APIServerEncryption{Type: configv1.EncryptionTypeAESCBC})
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, prevKeyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+			currentKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			require.Equal(t, prevKeyMeta, currentKeyMeta, "encryption key must not change while operator is stuck")
```
ardaguclu: We don't need this equality check. WaitForNoNewEncryptionKey handles this already. We also do not need to get currentKeyMeta (line: 333) too

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,83 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+		}},
+		{name: "ApplyValidConfigAndVerifyRecovery", testFunc: func(t testing.TB) {
+			prevKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
```
ardaguclu, line 336: Can you move this under `ValidProvider.Setup`

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,83 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.ValidProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.ValidProvider.APIServerEncryption)
+			WaitForNextMigratedKey(t, clientSet.Kube, prevKeyMeta,
```
ardaguclu: We shouldn't wait for NextMigratedKey, should we?

test/library/encryption/scenarios.go
```diff
@@ -281,3 +281,83 @@ func TestEncryptionRotation(ctx context.Context, t testing.TB, scenario Rotation
...
+				scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
+			currentKeyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
```
ardaguclu: No need to have this and equality check

ardaguclu:
Thank you
/lgtm
/approve

PR #2252, opened 2026-06-01
CNTRLPLANE-3226: kms: wire Vault credentials into sidecar containers

bertinatto:
Proof PRs (with a previous version of this PR):

- https://github.com/openshift/cluster-kube-apiserver-operator/pull/2169
- https://github.com/openshift/cluster-openshift-apiserver-operator/pull/699
- https://github.com/openshift/cluster-authentication-operator/pull/903

## Summary by CodeRabbit

* **New Features**
  * KMS plugin sidecar injection for static pod deployments with per-key credential mounts.
  * Vault AppRole credentials delivered to sidecars via mounted, secret-backed credential paths.

* **Improvements**
  * Unified handling of socket and credential mounts across deployment modes.
  * Deterministic per-key credential wiring for multi-provider scenarios.
  * Secret data keys now follow a consistent per-key naming convention used for credential files.

* **Tests**
  * Updated tests to validate credential provisioning, mounts, and sidecar arguments.


pkg/operator/encryption/encryptiondata/secret.go
```diff
@@ -202,6 +202,12 @@ func ExtractUniqueAndSortedKMSConfigurations(secretData *Config) ([]*apiserverco
 	return result, nil
 }
 
+// SecretDataKeyName returns the data key used in the encryption-config Secret
+// for a KMS plugin secret entry: "kms-plugin-secret-{secretName}_{dataKey}-{keyID}".
+func SecretDataKeyName(secretName, dataKey, keyID string) string {
+	return fmt.Sprintf("%s%s_%s-%s", encryptionConfigSecretDataPrefix, secretName, dataKey, keyID)
+}
+
```
bertinatto, line 216: I need to reconstruct the secret key name because I need it for the secret ID file path. However, it seems like we don't have utility funcs that handle that, so I created this one. PTAL and let me know if I should use it in line 155 as well

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -23,98 +33,163 @@ type sidecarProvider interface {
...
+
+	return nil
+}
+
+// AddKMSPluginSidecarToPodSpec injects KMS plugin sidecar containers into an aggregated API server pod spec (e.g., openshift-apiserver, oauth-apiserver).
+//
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
-// It uses an uncached client to avoid injecting sidecars based on a stale encryption configuration.
+// The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
 func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
```
bertinatto, line 92: for the aggregated apiservers, the signature didn't change

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: [secretData.FlatEntries](https://github.com/openshift/library-go/blob/24982272db91388ba83f3f856b6c98a776bcfed8/pkg/operator/encryption/state/types.go#L115) gives you the correct key (i.e. secretName_dataKey). We can have this function like this;

```go
func SecretDataKeyName(flatKey, keyID string) string {
	return fmt.Sprintf("%s%s-%s", encryptionConfigSecretDataPrefix, flatKey, keyID)
}
```

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -2,29 +2,40 @@ package pluginlifecycle
...
 	return &vault{
 		name:    name,
 		keyID:   keyID,
 		udsPath: udsPath,
 		config:  &pluginConfig.Vault,
-	}, nil
+		// role-id is passed as a CLI arg value, so we just need to get its raw value
+		roleID: string(secretData.Entries[secretName]["role-id"]),
+		// secret-id is passed as a file path to avoid exposing it in the pod manifest, so we need to construct the file path accordingly
+		secretIDPath: filepath.Join(credentialsDir, encryptiondata.SecretDataKeyName(secretName, "secret-id", keyID)),
```
ardaguclu: I think we need a function in state/types to generate internal key;

```go
func GetDataKey(secretName, dataKey string) {
   return secretName+secretDataKeySeparator+dataKey
}
```

We can update [here](https://github.com/openshift/library-go/blob/24982272db91388ba83f3f856b6c98a776bcfed8/pkg/operator/encryption/state/types.go#L122) with GetDataKey.

After that we need another GetSecretDataKey in encrpytiondata/secret

```go
func GetSecretDataKey(flatKey, keyID string) string {
	return fmt.Sprintf("%s%s-%s", encryptionConfigSecretDataPrefix, flatKey, keyID)
}
```

```go
filepath.Join(credentialsDir, encryptiondata.GetSecretDataKey(GetDataKey(secretName, dataKey), keyID))
```

@p0lyn0mial WDYT?

pkg/operator/encryption/encryptiondata/secret.go
ardaguclu: `%s_%s` generation belongs to key Secret. Outer one belongs to encryptiondata secret. I think we can adopt an approach like this https://github.com/openshift/library-go/pull/2252/changes/9b2290876ef56ad9171fca342cfec81a25176cc5#r3339140739

pkg/operator/encryption/kms/pluginlifecycle/vault.go
ardaguclu: I can open a separate PR if you would prefer

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -2,29 +2,40 @@ package pluginlifecycle
...
+func newVaultSidecarProvider(name, keyID, udsPath string, pluginConfig configv1.KMSPluginConfig, secretData state.KMSSecretData, credentialsDir string) *vault {
+	secretName := pluginConfig.Vault.Authentication.AppRole.Secret.Name
 	return &vault{
 		name:    name,
 		keyID:   keyID,
 		udsPath: udsPath,
 		config:  &pluginConfig.Vault,
-	}, nil
+		// role-id is passed as a CLI arg value, so we just need to get its raw value
+		roleID: string(secretData.Entries[secretName]["role-id"]),
```
p0lyn0mial: When secretData.Entries is nil this code will panic.

I think we could add `Get` method on `KMSSecretData` which would prevent panicking. 

The `KMSSecretData` already has a `Set` method adding a corresponding `Get` seem like a natural extension.

@ardaguclu  WDYT ?

ardaguclu: This is good idea. Maybe like this signature;

```go
func (d *KMSSecretData) Get(secretName, dataKey string) (value []byte, ok bool, err error) {
```

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -23,98 +33,169 @@ type sidecarProvider interface {
...
 
 		pluginConfig, ok := encryptionConfig.KMSPlugins[keyID]
 		if !ok {
-			return fmt.Errorf("missing plugin config for keyID %s", keyID)
+			return nil, fmt.Errorf("missing plugin config for keyID %s", keyID)
 		}
 
-		sidecarProvider, err := newSidecarProvider(keyID, udsPath, pluginConfig)
+		var secretData state.KMSSecretData
+		if encryptionConfig.KMSPluginsSecretData.ByKeyID != nil {
```
p0lyn0mial: maybe we should also export a `Get` method on `KMSPluginsSecretData` ?

`KMSPluginsSecretData` already has `Set` so `Get` would be natural extension. We could even make `ByKeyID` private.

@ardaguclu WDYT ?

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -2,29 +2,40 @@ package pluginlifecycle
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/utils/ptr"
 )
 
-// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
-func newVaultSidecarProvider(name, keyID, udsPath string, pluginConfig configv1.KMSPluginConfig) (*vault, error) {
+// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin data.
+// It assumes the input data has been already been validated.
+func newVaultSidecarProvider(name, keyID, udsPath string, pluginConfig configv1.KMSPluginConfig, secretData state.KMSSecretData, credentialsDir string) *vault {
```
p0lyn0mial: should `newVaultSidecarProvider` take `vaultConfig configv1.VaultKMSPluginConfig` directly ?

p0lyn0mial: not sure if plugins should need to know: 
  - How credentials are stored (KMSPluginsSecretData structure, byKeyID map)
  - How on-disk file names are derived (KMSSecretDataKeyName encoding)
  - Which package owns what (encryptiondata vs state)

we could introduce a helper `credentialResolver` which would encapsulate ^ plus some validation.

I think we could implement something like:
```
// credentialResolver provides nil-safe, validated access to KMS plugin credentials
// and resolves on-disk file paths without exposing the Secret data key encoding.
type credentialResolver struct {
	pluginsSecretData encryptiondata.KMSPluginsSecretData
	credentialsDir    string
	keyID             string
}

func (r *credentialResolver) Value(secretName, dataKey string) (string, error) {
	sd, ok := r.pluginsSecretData.Get(r.keyID)
	if !ok {
		return "", fmt.Errorf("missing secret data for keyID %s", r.keyID)
	}
	v, ok := sd.Get(secretName, dataKey)
	if !ok {
		return "", fmt.Errorf("missing %s in secret %s for keyID %s", dataKey, secretName, r.keyID)
	}
	return string(v), nil
}

func (r *credentialResolver) FilePath(secretName, dataKey string) (string, error) {
	sd, ok := r.pluginsSecretData.Get(r.keyID)
	if !ok {
		return "", fmt.Errorf("missing secret data for keyID %s", r.keyID)
	}
	if _, ok := sd.Get(secretName, dataKey); !ok {
		return "", fmt.Errorf("missing %s in secret %s for keyID %s", dataKey, secretName, r.keyID)
	}
	return filepath.Join(r.credentialsDir, encryptiondata.KMSSecretDataKeyName(secretName, dataKey, r.keyID)), nil
}
```

then the usage in the vault plugin becomes:

```
secretName := vaultConfig.Authentication.AppRole.Secret.Name

	roleID, err := creds.Value(secretName, "role-id")
	if err != nil {
		return nil, err
	}

	secretIDPath, err := creds.FilePath(secretName, "secret-id")
	if err != nil {
		return nil, err
	}
```

WDYT ?

p0lyn0mial: opened https://github.com/openshift/library-go/pull/2257

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
p0lyn0mial: xref: https://github.com/openshift/library-go/pull/2260

pkg/operator/encryption/kms/pluginlifecycle/vault.go
p0lyn0mial: xref: https://github.com/openshift/library-go/pull/2260

pkg/operator/encryption/encryptiondata/secret.go
bertinatto: makes sense. The links above don't work for me, can you check them?

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: something like this? I'm not sure I got the names right: https://github.com/openshift/library-go/pull/2264

Edit: I got confused with "flat" vs "raw" keys. IIUC they mean the same thing?

pkg/operator/encryption/encryptiondata/secret.go
bertinatto: ok, I got the link:

```
https://github.com/openshift/library-go/pull/2252/changes/9b2290876ef56ad9171fca342cfec81a25176cc5#r3339140739
```

submitted https://github.com/openshift/library-go/pull/2264

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: something like this? https://github.com/openshift/library-go/pull/2252/changes/57bafcaf22e7e8f576e7c694b7473bfde4fe7ae9

ardaguclu:
This https://github.com/openshift/library-go/pull/2264 needs to be taken out from this PR. But overall the changes look good to me.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -29,66 +38,132 @@ type sidecarProvider interface {
...
+// AddKMSPluginSidecarToStaticPodSpec injects KMS plugin sidecar containers into a kube-apiserver static pod spec.
+//
+// Static pods access credentials through the resource-dir volume, which the static pod revision controller
+// populates on disk from the encryption-config Secret. Because those files are owned by root, each sidecar
+// is configured to run as UID 0.
+//
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
-// It uses an uncached client to avoid injecting sidecars based on a stale encryption configuration.
+// The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
+func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
```
p0lyn0mial, line 60: could we intro this in a new PR ? would make the # of lines to review a bit smaller.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -183,3 +266,35 @@ func ensureSocketVolume(podSpec *corev1.PodSpec) {
 		},
 	)
 }
+
+func setRunAsUser(containers []corev1.Container, containerName string, uid int64) error {
```
p0lyn0mial, line 292: why do we need this function ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -183,3 +266,35 @@ func ensureSocketVolume(podSpec *corev1.PodSpec) {
...
+			return nil
+		}
+	}
+	return fmt.Errorf("container %s not found", containerName)
+}
+
+func ensureCredentialsVolume(podSpec *corev1.PodSpec, secretName string) {
+	for _, volume := range podSpec.Volumes {
+		if volume.Name == credentialsVolumeName {
+			return
```
p0lyn0mial: should we check the existing volume has the expected secret name ?

similarly to what we did in 
https://github.com/openshift/library-go/pull/2253/changes#diff-e58dbdd0e5a800a3c3deb3dfdc741194ced2d0fe3245f4b971b604879d9fb5a4R160

if yes, then worth checking if the other existing functions are consistent with ^

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: we need this function because of the resources-dir volume: https://github.com/openshift/library-go/pull/2252/changes#diff-e58dbdd0e5a800a3c3deb3dfdc741194ced2d0fe3245f4b971b604879d9fb5a4R79

pkg/operator/encryption/encryptiondata/secret.go
tjungblu: do you need to rebase? seems like this is done here:
https://github.com/openshift/library-go/pull/2269/changes

pkg/operator/encryption/kms/pluginlifecycle/vault.go
p0lyn0mial: yes, with https://github.com/openshift/library-go/pull/2269/changes

you should be able to read the rawKey with `sd.FlatEntry` and then use `FormatKMSSecretDataKey`

note that we slightly changed the signature of the `FlatEntry` method in `https://github.com/openshift/library-go/pull/2270`

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: makes sense, updated `ensureCredentialsVolume` and `ensureCredentialsVolume` to use a shared function that checks for volume equality

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: do you have a suggestion of how to split this PR (preferably without adding more code)?

pkg/operator/encryption/encryptiondata/secret.go
bertinatto: done

pkg/operator/encryption/kms/pluginlifecycle/vault.go
bertinatto: updated to use `FlatEntry` and `FormatKMSSecretDataKey`

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: attempt here: https://github.com/openshift/library-go/pull/2275/

ardaguclu:
As far as I can see, all the comments have been addressed. If there is any other changes needed, this can be handled in followup PR by @bertinatto 
/lgtm

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+)
+
+// credentialResolver resolves KMS plugin credentials for a specific keyID
+// by looking up secret data and mapping it to values or file paths on disk.
+type credentialResolver struct {
```
ardaguclu, line 12: Soon, we'll have to discuss whether we need another resolver for ConfigMap or we should update credentialResolver to handle Secret and ConfigMap.

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -9,22 +9,39 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin configuration.
-func newVaultSidecarProvider(name, keyID, udsPath string, vaultConfig configv1.VaultKMSPluginConfig) (*vault, error) {
+// newVaultSidecarProvider creates a Vault sidecar provider from the given KMS plugin data.
+// It assumes the input data has been already been validated.
+func newVaultSidecarProvider(name, keyID, udsPath string, vaultConfig configv1.VaultKMSPluginConfig, creds *credentialResolver) (*vault, error) {
+	secretName := vaultConfig.Authentication.AppRole.Secret.Name
```
ardaguclu, line 15: Should we error out if secretName is empty?. On the other hand, API will require to have it. So maybe we should not.

bertinatto: I agree we don't need to validate, but I can add a check if needed. I replied somewhere else, in a previous incarnation of this PR, that I believe the whole point of having the api validation is to free the user from validating the fields, and I think that applies well here

ardaguclu: I wanted to note that we retrieved this definition from Secret not directly API. So theoretically, it can be manually malformed.

bertinatto: yeah, I agree. This is an api type whose values went through a round trip: user input in the api -> serialized values in secrets -> back to deserialized values in the api type. If there's a bug in the encryption controllers, I agree this can be malformed

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
p0lyn0mial, line 12: should we add unit tests for `credentialResolver` ?

pkg/operator/encryption/kms/pluginlifecycle/vault.go
```diff
@@ -9,22 +9,39 @@ import (
...
+
+	roleID, err := creds.Value(secretName, "role-id")
```
p0lyn0mial, line 17: Worth guarding against empty `roleID`/`secretIDPath` here. 

`KMSReferenceData.Set` rejects empty values today, but this function doesn't know about that invariant. If it ever breaks, the sidecar would launch with `-approle-role-id=` and fail at runtime.
A quick emptiness check here would fail fast with a clear message.

bertinatto: submitted: https://github.com/openshift/library-go/pull/2283

PR #2253, opened 2026-06-01
CNTRLPLANE-3361: kms: generalize ensureVolumeMountInContainer helper

bertinatto:
## Summary by CodeRabbit

* **Refactor**
  * Internal volume mount injection for the KMS plugin sidecar was generalized and made more maintainable.

* **Tests**
  * Added a test that detects and fails when an existing KMS plugin volume mount has conflicting settings.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -9,10 +9,16 @@ import (
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
+)
+
+const (
+	kmsPluginSocketVolumeName = "kms-plugin-socket"
+	kmsPluginSocketMountPath  = "/var/run/kmsplugin"
 )
 
 // sidecarProvider abstracts the construction of a KMS plugin sidecar container for a specific provider (e.g. Vault).
@@ -83,6 +89,7 @@ func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, 
 
 	klog.V(4).Infof("injecting %d KMS sidecar(s)", len(kmsConfigurations))
 
+	socketVolumeMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: false}
 	for _, kmsConfiguration := range kmsConfigurations {
 		// ExtractUniqueAndSortedKMSConfigurations function rewrites the .Name field to include only the key ID
 		keyID := kmsConfiguration.Name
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -147,22 +147,19 @@ func ensureSocketVolumeMountInContainer(containers []corev1.Container, container
...
 
-	foundMount := false
 	container := &containers[containerIndex]
 	for _, m := range container.VolumeMounts {
-		if m.Name == "kms-plugin-socket" {
-			foundMount = true
-			break
+		if m.Name == volumeName {
+			return nil
 		}
```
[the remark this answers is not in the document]
ardaguclu: I agree that we can keep the existing mechanism.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -102,12 +102,12 @@ func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec,
 			return err
 		}
 
-		if err := ensureSocketVolumeMountInContainer(podSpec.InitContainers, sidecarProvider.Name()); err != nil {
+		if err := ensureVolumeMountInContainer(podSpec.InitContainers, sidecarProvider.Name(), "kms-plugin-socket", "/var/run/kmsplugin", false); err != nil {
```
ardaguclu: nit: It is a good time to define kms-plugin-socket and /var/run/kmsplugin as constants?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -134,7 +134,7 @@ func ensureSidecarContainer(podSpec *corev1.PodSpec, provider sidecarProvider) e
 	return nil
 }
 
-func ensureSocketVolumeMountInContainer(containers []corev1.Container, containerName string) error {
+func ensureVolumeMountInContainer(containers []corev1.Container, containerName, volumeName, mountPath string, readOnly bool) error {
```
p0lyn0mial: how about passing `mount corev1.VolumeMount` instead of `volumeName, mountPath string, readOnly bool` ?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -147,22 +147,19 @@ func ensureSocketVolumeMountInContainer(containers []corev1.Container, container
...
 	}
 
-	foundMount := false
 	container := &containers[containerIndex]
 	for _, m := range container.VolumeMounts {
-		if m.Name == "kms-plugin-socket" {
-			foundMount = true
-			break
+		if m.Name == volumeName {
+			return nil
```
p0lyn0mial, line 163: let's check if the existing volume has the same settings and return an err if not.

p0lyn0mial: let's also add a new test case for ^

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: moved to constants

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: done

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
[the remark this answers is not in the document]
bertinatto: as per another comment, this was changed to fail if a volume mount with the same name, but different else is different, exists

PR #2256, opened 2026-06-02
get vault Service ClusterIP

gangwgr:
get vault Service ClusterIP

## Summary by CodeRabbit

* **Tests**
  * Improved encryption test setup to dynamically resolve the Vault service address from the cluster, requiring a reachable service IP and port and deriving protocol from port naming when applicable.
  * Test Vault configuration is populated at invocation time so each run uses current runtime values.
  * Existing AppRole secret wiring for tests is preserved to maintain prior authentication behaviour.

### DIFF
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -3,6 +3,7 @@ package kms
 import (
 	"context"
 	"fmt"
+	"net"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -35,11 +36,18 @@ const (
 	vaultCommandTimeout           = 30 * time.Second
 )
 
-// DefaultVaultEncryptionProvider is a ready-to-use Vault KMS EncryptionProvider for e2e tests.
-// It bundles the default config with the AppRole secret setup.
-var DefaultVaultEncryptionProvider = library.EncryptionProvider{
-	APIServerEncryption: DefaultVaultKMSPluginConfig,
-	Setup:               ensureDefaultVaultAppRoleSecret,
+// DefaultVaultEncryptionProvider returns a ready-to-use Vault KMS EncryptionProvider for e2e tests.
+// It resolves the Vault Service ClusterIP at call time to avoid DNS resolution issues,
+// and bundles the AppRole secret setup.
+func DefaultVaultEncryptionProvider(ctx context.Context, t testing.TB) library.EncryptionProvider {
+	cfg := DefaultVaultKMSPluginConfig
+	// Use the Service ClusterIP instead of DNS name because kube-apiserver pods
+	// cannot resolve cluster-local Service names (they use host network DNS).
+	cfg.KMS.Vault.VaultAddress = getVaultServiceAddress(ctx, t)
+	return library.EncryptionProvider{
+		APIServerEncryption: cfg,
+		Setup:               ensureDefaultVaultAppRoleSecret,
+	}
 }
 
 var DefaultFakeVaultEncryptionProvider = library.EncryptionProvider{
@@ -173,4 +181,30 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
 	require.NoError(t, err, "failed to parse key version from output: %q", string(output))
 
 	return version
+}
+
+// getVaultServiceAddress returns the Vault address using the Service's ClusterIP
+// instead of the DNS name, reading the port and scheme from the Service spec.
+func getVaultServiceAddress(ctx context.Context, t testing.TB) string {
+	t.Helper()
+	cs := library.GetClients(t)
+
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
+	require.NotEmpty(t, svc.Spec.Ports, "vault Service has no ports")
+
+	// The Vault Helm chart names the client port "https" (8200).
+	var port *corev1.ServicePort
+	for i := range svc.Spec.Ports {
+		if svc.Spec.Ports[i].Name == "https" {
+			port = &svc.Spec.Ports[i]
+			break
+		}
+	}
+	require.NotNil(t, port, "vault Service has no port named \"https\"")
+
+	addr := fmt.Sprintf("https://%s", net.JoinHostPort(svc.Spec.ClusterIP, strconv.Itoa(int(port.Port))))
+	t.Logf("Resolved Vault Service address: %s", addr)
+	return addr
 }


test/library/encryption/kms/vault.go
```diff
@@ -174,3 +180,15 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
 
 	return version
 }
+
+// GetVaultServiceAddress returns the Vault address using the Service's ClusterIP
```
p0lyn0mial: could we make this function private ?

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +180,15 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+func GetVaultServiceAddress(ctx context.Context, t testing.TB) string {
+	t.Helper()
+	cs := library.GetClients(t)
+
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
+	addr := fmt.Sprintf("https://%s", net.JoinHostPort(svc.Spec.ClusterIP, "8200"))
```
p0lyn0mial: I think this function should read the port and scheme (HTTPS) directly from the service.

test/library/encryption/kms/vault.go
```diff
@@ -35,11 +36,16 @@ const (
...
-// It bundles the default config with the AppRole secret setup.
-var DefaultVaultEncryptionProvider = library.EncryptionProvider{
-	APIServerEncryption: DefaultVaultKMSPluginConfig,
-	Setup:               ensureDefaultVaultAppRoleSecret,
+// DefaultVaultEncryptionProvider returns a ready-to-use Vault KMS EncryptionProvider for e2e tests.
+// It resolves the Vault Service ClusterIP at call time to avoid DNS resolution issues,
+// and bundles the AppRole secret setup.
+func DefaultVaultEncryptionProvider(ctx context.Context, t testing.TB) library.EncryptionProvider {
+	cfg := DefaultVaultKMSPluginConfig
+	cfg.KMS.Vault.VaultAddress = GetVaultServiceAddress(ctx, t)
```
p0lyn0mial: let's add a small comment explaining why we are doing this.

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +180,15 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
 }
+
+// GetVaultServiceAddress returns the Vault address using the Service's ClusterIP
+func GetVaultServiceAddress(ctx context.Context, t testing.TB) string {
+	t.Helper()
+	cs := library.GetClients(t)
+
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
```
p0lyn0mial, line 194: add one more guard `require.NotEmpty(t, svc.Spec.Ports, "vault Service has no ports")`

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +182,24 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+// getVaultServiceAddress returns the Vault address using the Service's ClusterIP
+// instead of the DNS name, reading the port and scheme from the Service spec.
+func getVaultServiceAddress(ctx context.Context, t testing.TB) string {
+	t.Helper()
+	cs := library.GetClients(t)
+
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
+	require.NotEmpty(t, svc.Spec.Ports, "vault Service has no ports")
```
p0lyn0mial, line 195: should we also add `require.Len(t, svc.Spec.Ports, 1, "vault Service must have exactly one port")` ?

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +182,24 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+
+	port := svc.Spec.Ports[0]
```
p0lyn0mial: and then after ensuring there will be only a single port we could read its name and port number.

test/library/encryption/kms/vault.go
gangwgr: i think we needed as I checked in helm chart step it create 2 ports 
```
ports:
  - name: {{ include "vault.scheme" . }}   # resolves to "https" when TLS enabled
    port: 8200
    targetPort: 8200
  - name: https-internal
    port: 8201
    targetPort: 8201
```

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +182,33 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+	// and "https-internal" (8201) for cluster communication. Find the client port.
+	var port *corev1.ServicePort
+	for i := range svc.Spec.Ports {
+		if svc.Spec.Ports[i].Name == "https" || svc.Spec.Ports[i].Name == "http" {
+			port = &svc.Spec.Ports[i]
+			break
+		}
+	}
+	if port == nil {
+		require.Len(t, svc.Spec.Ports, 1, "vault Service has multiple ports and none named https/http")
```
p0lyn0mial: if the port hasn't been found then we need to fail

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +182,33 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
+	require.NotEmpty(t, svc.Spec.Ports, "vault Service has no ports")
+
+	// The Vault Helm chart creates two ports: "https" (8200) for client traffic
+	// and "https-internal" (8201) for cluster communication. Find the client port.
+	var port *corev1.ServicePort
+	for i := range svc.Spec.Ports {
+		if svc.Spec.Ports[i].Name == "https" || svc.Spec.Ports[i].Name == "http" {
```
p0lyn0mial: this can give as a random order.

p0lyn0mial: you know what, maybe for simplicity let's assume we need to find a port with `https` name.

test/library/encryption/kms/vault.go
```diff
@@ -174,3 +182,29 @@ func getCurrentKeyVersion(ctx context.Context, t testing.TB) int {
...
+
+	svc, err := cs.Kube.CoreV1().Services(defaultVaultNamespace).Get(ctx, "vault", metav1.GetOptions{})
+	require.NoError(t, err, "failed to get vault Service in namespace %s", defaultVaultNamespace)
+	require.NotEmpty(t, svc.Spec.ClusterIP, "vault Service has no ClusterIP")
+	require.NotEmpty(t, svc.Spec.Ports, "vault Service has no ports")
+
+	// The Vault Helm chart names the client port "https" (8200).
+	var port *corev1.ServicePort
+	for i := range svc.Spec.Ports {
+		if svc.Spec.Ports[i].Name == "https" && svc.Spec.Ports[i].Port == 8200 {
```
gangwgr: added this check because, if both ports has https it can return random, as we know our port is 8200 for client so added this

p0lyn0mial: just check `svc.Spec.Ports[i].Name == "https"`

p0lyn0mial: drop `svc.Spec.Ports[i].Port == 8200`

gangwgr: done

PR #2261, opened 2026-06-02
NO-JIRA: Fix data race in InMemoryRecorder

tchap:
Events() reads the events slice, ComponentName() reads source,
and WithContext() writes ctx — all without holding the embedded
mutex that Event()/Warning()/ForComponent() use. Add locking to
all three for consistency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race conditions in the in-memory event recorder: reads are now concurrency-safe and return independent snapshots so callers can't observe or modify internal event storage during concurrent operations.
  * Protected component/context metadata accesses from concurrent reads/writes.
  * Improves stability under concurrent workloads.

### DIFF
--- a/pkg/operator/events/recorder_in_memory.go
+++ b/pkg/operator/events/recorder_in_memory.go
@@ -42,6 +42,8 @@ func NewInMemoryRecorder(sourceComponent string, clock clock.PassiveClock) InMem
 }
 
 func (r *inMemoryEventRecorder) ComponentName() string {
+	r.Lock()
+	defer r.Unlock()
 	return r.source
 }
 
@@ -55,6 +57,8 @@ func (r *inMemoryEventRecorder) ForComponent(component string) Recorder {
 }
 
 func (r *inMemoryEventRecorder) WithContext(ctx context.Context) Recorder {
+	r.Lock()
+	defer r.Unlock()
 	r.ctx = ctx
 	return r
 }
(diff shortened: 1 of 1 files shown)


pkg/operator/events/recorder_in_memory.go
```diff
@@ -65,7 +65,9 @@ func (r *inMemoryEventRecorder) WithComponentSuffix(suffix string) Recorder {
 
 // Events returns list of recorded events
 func (r *inMemoryEventRecorder) Events() []*corev1.Event {
```
p0lyn0mial, line 71: are there any other methods on this struct that miss the lock ?

pkg/operator/events/recorder_in_memory.go
```diff
@@ -65,6 +65,8 @@ func (r *inMemoryEventRecorder) WithComponentSuffix(suffix string) Recorder {
...
+	r.Lock()
+	defer r.Unlock()
 	return r.events
 }
```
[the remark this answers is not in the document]
p0lyn0mial: @coderabbitai fixed ?

pkg/operator/events/recorder_in_memory.go
tchap: Yeah, `ComponentName` and `WithContext`, but I don't think that these matter. They are for init only, you won't really get a race while running the recorder. `events` need to be protected for sure.

tchap: I can add the lock, it's weirdly inconsistent, though. Give me a sec.

p0lyn0mial: yeah, let's make it consistent.

tchap: Amended.

tchap:
/retitle NO-JIRA: Fix data race in InMemoryRecorder

PR #2273, opened 2026-06-03
OCPBUGS-85367: controllercmd,certsyncpod: add UserAgentSuffix to distinguish clients

tchap:
Add `WithUserAgentSuffix` to `ControllerBuilder` and expose `UserAgentSuffix` on `ControllerCommandConfig` so that static pod sidecar controllers using localhost can be identified by the kube-apiserver's pre-readiness request filter.

The `cert-syncer`, which does not use `controllercmd`, sets its suffix directly in `Complete()`.

---

Note that this does not generate super pretty `UserAgent` strings. When you run the test, you get something like

```
___TestWithUserAgentSuffix_in_github_com_openshift_library_go_pkg_controller_controllercmd.test/v0.0.0 (linux/amd64) kubernetes/$Format/cert-recovery
```

So to check for a particular component, you need to do `HasPrefix("cluster-kube-controller-manager-operator") && HasSuffix(/cert-syncer), for example.

## Release Notes

* **New Features**
  * Controllers now support a configurable REST client User-Agent suffix, which is appended to outgoing requests for clearer identification.

* **Tests**
  * Added coverage to verify the User-Agent suffix behavior, including the “empty suffix” and “appended suffix” cases.

* **Documentation / Chores**
  * Reorganized import ordering where applicable.

### DIFF
--- a/pkg/controller/controllercmd/builder.go
+++ b/pkg/controller/controllercmd/builder.go
@@ -3,11 +3,12 @@ package controllercmd
 import (
 	"context"
 	"fmt"
-	"k8s.io/utils/clock"
 	"os"
 	"strings"
 	"sync"
 	"time"
+
+	"k8s.io/utils/clock"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
Overall the approach seems reasonable. It would help me if you could provide a list of the components you'd like to change, or even just one component you'd like to modify.

tchap:
> Overall the approach seems reasonable. It would help me if you could provide a list of the components you'd like to change, or even just one component you'd like to modify.

There is already a testing PR for KCM-O: https://github.com/openshift/cluster-kube-controller-manager-operator/pull/940 . The Jira issue mentions KAS-O, KCM-O and the scheduler operator, but I haven't checked other components. All these three need a very similar PR as this testing one for KCM-O. Then we need to align the skip list in the API server. And by skip list I mean to extend the logic as in https://github.com/openshift/kubernetes/pull/376

PR #2275, opened 2026-06-03
CNTRLPLANE-3226: kms: split sidecar injection for static pods vs aggregated apiservers

bertinatto:
## Release Notes

* **New Features**
  * KMS plugin sidecar injection now supports static-pod API server configurations in addition to aggregated-pod specifications.
  * Enhanced credential mounting for KMS plugin sidecars with support for Vault AppRole authentication.

* **Refactor**
  * Consolidated core sidecar injection logic into shared helper functions for improved maintainability.
  * Enhanced volume provisioning with improved error handling for socket and credential volumes.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -14,6 +14,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	resourceDirVolumeName = "resource-dir"
+	credentialsVolumeName = "kms-plugin-credentials"
+	credentialsMountPath  = "/var/run/secrets/kms-plugin"
+	resourcesDir          = "/etc/kubernetes/static-pod-resources"
 )
 
 const (
@@ -29,7 +37,8 @@ type sidecarProvider interface {
 	BuildSidecarContainer() (corev1.Container, error)
 }
 
-// newSidecarProvider creates a provider-specific SidecarProvider for the given keyID, UDS endpoint, and plugin configuration.
+// newSidecarProvider creates a provider-specific sidecarProvider for the given keyID and plugin configuration,
+// wiring in credentials from secretData and the on-disk credentialsDir.
 func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSPluginConfig) (sidecarProvider, error) {
 	switch pluginConfig.Type {
 	case configv1.VaultKMSProvider:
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -167,19 +244,50 @@ func ensureVolumeMountInContainer(containers []corev1.Container, containerName s
...
+	return ensureVolume(podSpec, volume)
+}
+
+func setRunAsUser(containers []corev1.Container, containerName string, uid int64) error {
+	for i, c := range containers {
+		if c.Name == containerName {
+			if c.SecurityContext == nil {
+				containers[i].SecurityContext = &corev1.SecurityContext{}
+			}
+			containers[i].SecurityContext.RunAsUser = ptr.To(uid)
```
ardaguclu, line 288: According to my research, we don't have any option other than running the container with uid: 0. But maybe we should harden the SecurityContext with this. WDYT?

```suggestion
			containers[i].SecurityContext.RunAsUser = ptr.To(uid)

			if uid == 0 {
				containers[i].SecurityContext.ReadOnlyRootFilesystem = ptr.To(true)
				containers[i].SecurityContext.AllowPrivilegeEscalation = ptr.To(false)
				containers[i].SecurityContext.Capabilities = &corev1.Capabilities{
					Drop: []corev1.Capability{"ALL"},
				}
				containers[i].SecurityContext.SeccompProfile = &corev1.SeccompProfile{
					Type: corev1.SeccompProfileTypeRuntimeDefault,
				}
			}
```

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -39,56 +48,120 @@ func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSP
...
+		return nil
+	}
+
+	for _, name := range sidecarNames {
+		volumeMount := corev1.VolumeMount{Name: resourceDirVolumeName, MountPath: resourcesDir, ReadOnly: true}
+		if err := ensureVolumeMountInContainer(podSpec.InitContainers, name, volumeMount); err != nil {
+			return err
+		}
+		// The resource-dir files are owned by root, so the sidecar needs root to read credential files.
+		if err := setRunAsUser(podSpec.InitContainers, name, 0); err != nil {
```
ardaguclu, line 76: In a followup PR, you can rename this function and drop the uid argument. Nobody else uses it.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: This can be fixed in followup PR.

ardaguclu (commented):
Overall changes look good to me. I have 2 comments which can be handled in a followup PR.
/lgtm

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: This is a very good point, we should definitely tight this up. The plugin doesn't need any privileges to do its work, it only needs to run as root to have access to the secret ID file. I'll need to test these changes in a cluster to make sure it works, then I'll create a new PR  to update them. Thanks!

bertinatto: submitted: https://github.com/openshift/library-go/pull/2277

PR #2161, opened 2026-04-13
CNTRLPLANE-3237: Add new KMS foundational mechanism in encryption controller

ardaguclu:
This PR brings the minimal changes required for the KMS TPv2 work described here https://github.com/openshift/enhancements/pull/1960. 

Main aim is to store data in Data field of Secret instead of annotation, insert key id into unix domain socket path. This PR should work with plugin lifecycle smoothly. 

Changes here break KMS TP v1 functionality, so it must not be merged in release-4.22 code base (so that this can be merged after the branch cut of 4.23). 

This PR excludes;
* Track the changes in API side (whether new key is required or just in-place update)
* secret/configmap, kmsConfig propagation mechanism to Key Secret
* secret/configmap periodic sync

## Summary by CodeRabbit

* **Refactor**
  * KMS encryption keys now use per-key Unix socket endpoints instead of a single shared socket
  * KMS configuration is stored in secret data fields (not annotations) and related identifiers updated
  * Test/KMS assets adjusted to run multiple per-key mock KMS instances

* **Tests**
  * Test suites updated to validate per-key endpoints and the new KMS config storage format

PR #2279, opened 2026-06-08
pkg/operator/encryption/kms/health: define cobra flag interface

ibihim:
## What

Defines the cobra/pflag command interface for the kms-health-monitor
sidecar.

## Why

I was told to split up the PR https://github.com/openshift/library-go/pull/2161

## Notes

Used for ref https://github.com/ardaguclu/library-go/blob/21d59d1efa2b372350d178aa0696ecdc8b3a6380/pkg/operator/encryption/kms/health/cmd.go

## Summary by CodeRabbit

* **New Features**
  * Added a CLI tool for KMS health reporting with configurable KMS socket targets, probe interval, per-RPC read/write timeouts, node identification, and kubeconfig/in‑cluster selection; validates inputs and logs startup parameters.
* **Tests**
  * Added unit tests covering socket URI formats, multiple-socket acceptance, timing parameter validation, and node-name requirements.

### DIFF
--- a/pkg/operator/encryption/kms/health/cmd.go
+++ b/pkg/operator/encryption/kms/health/cmd.go
@@ -1,0 +1,110 @@
+package health
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+// kmsSocketPattern matches the socket path each co-located KMSv2 plugin is
+// mounted at, e.g. unix:///var/run/kmsplugin/kms-1.sock.
+var kmsSocketPattern = regexp.MustCompile(`^unix:///var/run/kmsplugin/kms-\d+\.sock$`)
+
+// options' flag-bound fields are exported so the struct can be logged as a
+// whole via klog.InfoS, which JSON-marshals its values.
+type options struct {
+	KMSSockets   []string
+	Interval     time.Duration
+	ReadTimeout  time.Duration
+	WriteTimeout time.Duration
+	NodeName     string
+	Kubeconfig   string
+
+	newOperatorClient func(*rest.Config) (v1helpers.OperatorClient, error)
+}
+
+func NewCommand(ctx context.Context, newOperatorClient func(*rest.Config) (v1helpers.OperatorClient, error)) *cobra.Command {
+	o := &options{
+		newOperatorClient: newOperatorClient,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "kms-health-reporter",
+		Short: "Observes co-located KMSv2 plugins and publishes status as an OperatorCondition.",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.validate(); err != nil {
+				return err
+			}
+			return o.run()
+		},
+	}
+	o.addFlags(cmd.Flags())
+	return cmd
+}
+
+func (o *options) addFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVar(&o.KMSSockets, "kms-sockets", nil, "KMS plugin endpoints in unix:// URI format (e.g. unix:///var/run/kmsplugin/kms-1.sock)")
+	fs.DurationVar(&o.Interval, "interval", 30*time.Second, "cadence between probe+emit cycles")
+	fs.DurationVar(&o.ReadTimeout, "read-timeout", 5*time.Second, "deadline for each Status RPC")
+	fs.DurationVar(&o.WriteTimeout, "write-timeout", 10*time.Second, "deadline for each condition update")
+	fs.StringVar(&o.NodeName, "node-name", "", "node name recorded in the condition used to help to identify the origin")
+	fs.StringVar(&o.Kubeconfig, "kubeconfig", "", "path to a kubeconfig; empty uses in-cluster config")
+}
+
+func (o *options) validate() error {
+	if len(o.KMSSockets) == 0 {
+		return fmt.Errorf("--kms-sockets is required, at least one")
+	}
+	for _, s := range o.KMSSockets {
+		if !kmsSocketPattern.MatchString(s) {
+			return fmt.Errorf("--kms-sockets entry %q must match %s", s, kmsSocketPattern)
+		}
+	}
+
+	if o.Interval <= 0 {
+		return fmt.Errorf("--interval must be positive")
+	}
+	if o.ReadTimeout <= 0 {
+		return fmt.Errorf("--read-timeout must be positive")
+	}
+	if o.WriteTimeout <= 0 {
+		return fmt.Errorf("--write-timeout must be positive")
+	}
+	if o.NodeName == "" {
+		return fmt.Errorf("--node-name is required")
+	}
+
+	return nil
+}
+
+func (o *options) run() error {
+	cfg, err := buildRESTConfig(o.Kubeconfig)
+	if err != nil {
+		return fmt.Errorf("build rest config: %w", err)
+	}
+
+	if _, err := o.newOperatorClient(cfg); err != nil {
+		return fmt.Errorf("build operator client: %w", err)
+	}
+
+	klog.InfoS("kms-health-reporter starting", "config", o)
+
+	return nil
+}
+
+func buildRESTConfig(kubeconfig string) (*rest.Config, error) {
+	if kubeconfig != "" {
+		return clientcmd.BuildConfigFromFlags("", kubeconfig)
+	}
+	return rest.InClusterConfig()
+}
--- a/pkg/operator/encryption/kms/health/cmd_test.go
+++ b/pkg/operator/encryption/kms/health/cmd_test.go
@@ -1,0 +1,116 @@
+package health
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// validOptions returns an options value that passes validate. Each test case
+// mutates a single field so the failure under test is unambiguous.
+func validOptions() *options {
+	return &options{
+		KMSSockets:   []string{"unix:///var/run/kmsplugin/kms-1.sock"},
+		Interval:     30 * time.Second,
+		ReadTimeout:  5 * time.Second,
+		WriteTimeout: 10 * time.Second,
+		NodeName:     "node-1",
+	}
+}
+
+func TestValidate(t *testing.T) {
+	tests := []struct {
+		name    string
+		mutate  func(*options)
+		wantErr bool
+	}{
+		{
+			name:   "valid",
+			mutate: func(*options) {},
+		},
+		{
+			name:    "no sockets",
+			mutate:  func(o *options) { o.KMSSockets = nil },
+			wantErr: true,
+		},
+		{
+			name:    "empty socket entry",
+			mutate:  func(o *options) { o.KMSSockets = []string{""} },
+			wantErr: true,
+		},
+		{
+			name:   "multiple valid sockets",
+			mutate: func(o *options) { o.KMSSockets = append(o.KMSSockets, "unix:///var/run/kmsplugin/kms-2.sock") },
+		},
+		{
+			name:    "socket missing unix scheme",
+			mutate:  func(o *options) { o.KMSSockets = []string{"/var/run/kmsplugin/kms-1.sock"} },
+			wantErr: true,
+		},
+		{
+			name:    "socket scheme without path",
+			mutate:  func(o *options) { o.KMSSockets = []string{"unix://"} },
+			wantErr: true,
+		},
+		{
+			name:    "socket wrong directory",
+			mutate:  func(o *options) { o.KMSSockets = []string{"unix:///tmp/kms-1.sock"} },
+			wantErr: true,
+		},
+		{
+			name:    "socket non-numeric index",
+			mutate:  func(o *options) { o.KMSSockets = []string{"unix:///var/run/kmsplugin/kms-x.sock"} },
+			wantErr: true,
+		},
+		{
+			name:    "socket missing .sock suffix",
+			mutate:  func(o *options) { o.KMSSockets = []string{"unix:///var/run/kmsplugin/kms-1"} },
+			wantErr: true,
+		},
+		{
+			name:    "socket with surrounding whitespace",
+			mutate:  func(o *options) { o.KMSSockets = []string{" unix:///var/run/kmsplugin/kms-1.sock "} },
+			wantErr: true,
+		},
+		{
+			name:    "interval zero",
+			mutate:  func(o *options) { o.Interval = 0 },
+			wantErr: true,
+		},
+		{
+			name:    "interval negative",
+			mutate:  func(o *options) { o.Interval = -time.Second },
+			wantErr: true,
+		},
+		{
+			name:    "read timeout zero",
+			mutate:  func(o *options) { o.ReadTimeout = 0 },
+			wantErr: true,
+		},
+		{
+			name:    "write timeout zero",
+			mutate:  func(o *options) { o.WriteTimeout = 0 },
+			wantErr: true,
+		},
+		{
+			name:    "node name empty",
+			mutate:  func(o *options) { o.NodeName = "" },
+			wantErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			o := validOptions()
+			tc.mutate(o)
+
+			err := o.validate()
+			if tc.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}


pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	startFunc := func(ctx context.Context, controllerContext *controllercmd.ControllerContext) error {
+		if err := o.validate(); err != nil {
+			return err
+		}
+		return o.run(ctx, controllerContext.KubeConfig)
+	}
+
+	cfg := controllercmd.NewControllerCommandConfig(
```
ardaguclu: @p0lyn0mial I suggested to use this `NewControllerCommandConfig`, since it provides graceful shutdown termination, logging, etc. But none of the controllers in library-go uses it. Would it make sense using this?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+		"kms-health-monitor",
+		version.Info{Major: "0", Minor: "0"},
```
ardaguclu:
```suggestion
		version.Get(),
```

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+		startFunc,
+		clock.RealClock{},
+	)
+	cfg.DisableLeaderElection = true
+	cfg.DisableServing = true
```
ardaguclu: Why do we disableServing?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (o *options) addFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVar(&o.kmsSockets, "kms-sockets", nil, "KMS plugin endpoints in unix:// URI format (e.g. unix:///var/run/kmsplugin/kms-1.sock)")
+	fs.DurationVar(&o.interval, "interval", o.interval, "cadence between probe+emit cycles")
+	fs.DurationVar(&o.readTimeout, "read-timeout", o.readTimeout, "deadline for each Status RPC")
+	fs.DurationVar(&o.writeTimeout, "write-timeout", o.writeTimeout, "deadline for each condition update")
+	fs.StringVar(&o.nodeName, "node-name", "", "node name recorded in the condition used to help to identify the origin")
+
+	fs.StringVar(&o.targetGroup, "target-group", "", "API group of the operator CR (e.g. operator.openshift.io)")
```
ardaguclu: Why do we need both GVR and GVK?. Can't we choose one and use it?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	fs.StringVar(&o.targetVersion, "target-version", "v1", "API version of the operator CR")
+	fs.StringVar(&o.targetResource, "target-resource", "", "resource name of the operator CR (e.g. kubeapiservers)")
+	fs.StringVar(&o.targetKind, "target-kind", "", "kind of the operator CR (e.g. KubeAPIServer)")
+}
+
+func (o *options) validate() error {
+	if len(o.kmsSockets) == 0 {
```
ardaguclu: I think instead of having these manual validations, we can simply use regex `unix:///var/run/kmsplugin/kms-%d.sock`.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if o.targetKind == "" {
+		return fmt.Errorf("--target-kind is required")
+	}
+
+	return nil
+}
+
+func (o *options) run(_ context.Context, _ *rest.Config) error {
+	_ = schema.GroupVersionResource{
```
ardaguclu: Please remove unused initializations

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(o.kmsSockets) == 0 {
+		return fmt.Errorf("--kms-sockets is required, at least one")
+	}
+	for _, s := range o.kmsSockets {
+		if strings.TrimSpace(s) == "" {
+			return fmt.Errorf("--kms-sockets cannot contain empty entries")
+		}
+		if !strings.HasPrefix(s, unixScheme) {
+			return fmt.Errorf("--kms-sockets entry %q must use unix:// scheme", s)
+		}
```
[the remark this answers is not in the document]
ardaguclu: I think it is better to use regex. Because we know the exact path.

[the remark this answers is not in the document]
ardaguclu: Format is `unix:///var/run/kms-plugin/kms-%d.sock`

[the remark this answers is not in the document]
ardaguclu: This doesn't make sense to me. Path is almost hardcoded, why do you still make it generic?

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: no strong opinion. won't `NewControllerCommandConfig` provide leader election by default ? 
I think we need to set `DisableLeaderElection`.
Wondering what else we need to disable ...


> But none of the controllers in library-go uses it

I think `NewControllerCommandConfig` is used by operators not controllers.

ardaguclu: Maybe we can start simpler by directly running as in https://github.com/openshift/library-go/pull/2193/changes#diff-1c70e24fc6168cec30f3288123a4078ae669ed05761705702c16214d4fb1b39dR36-R46

ardaguclu: On the other hand, `NewControllerCommandConfig` gives us initialized kubeconfig with more benefits. I'm leaning towards keeping it, until we see an issue. But I agree that we should disable leader election.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: +1

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	fs.DurationVar(&o.writeTimeout, "write-timeout", o.writeTimeout, "deadline for each condition update")
+	fs.StringVar(&o.nodeName, "node-name", "", "node name recorded in the condition used to help to identify the origin")
+
+	fs.StringVar(&o.targetGroup, "target-group", "", "API group of the operator CR (e.g. operator.openshift.io)")
+	fs.StringVar(&o.targetVersion, "target-version", "v1", "API version of the operator CR")
+	fs.StringVar(&o.targetResource, "target-resource", "", "resource name of the operator CR (e.g. kubeapiservers)")
+	fs.StringVar(&o.targetKind, "target-kind", "", "kind of the operator CR (e.g. KubeAPIServer)")
+}
+
+func (o *options) validate() error {
```
p0lyn0mial, line 64: could we add some unit tests for the `validate` method ?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	cmd.Use = "kms-health-monitor"
+	cmd.Short = "Observes co-located KMSv2 plugins and publishes status as an OperatorCondition."
+
+	o.addFlags(cmd.Flags())
+	return cmd
+}
+
+func (o *options) addFlags(fs *pflag.FlagSet) {
+	fs.StringSliceVar(&o.kmsSockets, "kms-sockets", nil, "KMS plugin endpoints in unix:// URI format (e.g. unix:///var/run/kmsplugin/kms-1.sock)")
+	fs.DurationVar(&o.interval, "interval", o.interval, "cadence between probe+emit cycles")
```
p0lyn0mial: I think that usually we set the default val in the flag registration call.

p0lyn0mial: Also, usually we provide the default value in the flag description.

pkg/operator/encryption/kms/health/cmd.go
ardaguclu: After some investigations, `NewControllerCommandConfig` is used by operators. This controller is supposed to run on apiservers. It is better to not use it.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	writeTimeout time.Duration
+	nodeName     string
+
+	targetGroup    string
+	targetVersion  string
+	targetResource string
+	targetKind     string
+}
+
+func NewCommand(ctx context.Context) *cobra.Command {
```
p0lyn0mial: since this binary will be exposed via operators would it make sense to change the signature to:

`NewCommand(ctx context.Context, newOperatorClient func(config *rest.Config) (operatorv1helpers.OperatorClient, error)) *cobra.Command`


that would allow us to get rid of:
```
targetGroup    string
targetVersion  string
targetResource string
targetKind     string
```

and would allow us to work with a "typed" client instead of dynamic.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+		"kms-health-monitor",
+		version.Info{Major: "0", Minor: "0"},
+		startFunc,
+		clock.RealClock{},
+	)
+	cfg.DisableLeaderElection = true
+	cfg.DisableServing = true
+
+	cmd := cfg.NewCommandWithContext(ctx)
+	cmd.Use = "kms-health-monitor"
```
p0lyn0mial: i think the kep calls it `kms-health-reporter`

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+		Version:  o.targetVersion,
+		Resource: o.targetResource,
+	}
+	_ = schema.GroupVersionKind{
+		Group:   o.targetGroup,
+		Version: o.targetVersion,
+		Kind:    o.targetKind,
+	}
+
+	klog.InfoS("kms-health-monitor starting",
```
p0lyn0mial: `kms-health-reporter starting`

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	if o.interval <= 0 {
+		return fmt.Errorf("--interval must be positive")
+	}
+	if o.readTimeout <= 0 {
+		return fmt.Errorf("--read-timeout must be positive")
+	}
+	if o.writeTimeout <= 0 {
+		return fmt.Errorf("--write-timeout must be positive")
+	}
+	if o.readTimeout+o.writeTimeout >= o.interval {
```
p0lyn0mial: why is this check required ?

the probe loop runs sequentially, the tick fires, do the work, if it takes longer than the interval, the next tick just starts late.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	fs.StringVar(&o.targetResource, "target-resource", "", "resource name of the operator CR (e.g. kubeapiservers)")
+	fs.StringVar(&o.targetKind, "target-kind", "", "kind of the operator CR (e.g. KubeAPIServer)")
+}
+
+func (o *options) validate() error {
+	if len(o.kmsSockets) == 0 {
+		return fmt.Errorf("--kms-sockets is required, at least one")
+	}
+	for _, s := range o.kmsSockets {
+		if strings.TrimSpace(s) == "" {
```
p0lyn0mial: isn't this check a bit redundant ? 
(whitespace only string will never pass the prefix check below)

ardaguclu: I think we should use regex and that will eliminate all these manual validations.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: could this just be `klog.InfoS("kms-health-reporter starting", "config", o)` ?

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: +1

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: i think we can get rid of these, have a look https://github.com/openshift/library-go/pull/2279#discussion_r3372819012 and let me know what you think.

pkg/operator/encryption/kms/health/cmd.go
ardaguclu: That is a good idea.

pkg/operator/encryption/kms/health/cmd.go
ardaguclu: Your suggestion is much cleaner

pkg/operator/encryption/kms/health/cmd.go
ibihim: Because you said so:

https://github.com/ardaguclu/library-go/blob/21d59d1efa2b372350d178aa0696ecdc8b3a6380/pkg/operator/encryption/kms/health/cmd.go#L54

pkg/operator/encryption/kms/health/cmd.go
ibihim: Because you said so: https://github.com/ardaguclu/library-go/blob/21d59d1efa2b372350d178aa0696ecdc8b3a6380/pkg/operator/encryption/kms/health/cmd.go#L69

ibihim: This is not funny.

ardaguclu: My suggestion was an improvement until Lukasz's suggestion is much more better.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: I’m fine either way. We can start by creating a `cobra.Command` and adjust later if we find `NewControllerCommandConfig` to be a better option.

pkg/operator/encryption/kms/health/cmd.go
ibihim: That is cool.

pkg/operator/encryption/kms/health/cmd.go
ibihim: OK, I will remove it.

pkg/operator/encryption/kms/health/cmd.go
ibihim: After removing the `NewControllerCommandConfig`, this  vanished. Right?

pkg/operator/encryption/kms/health/cmd.go
ibihim: After removing `NewControllerCommandConfig`, this isn't an issue anymore, right?

pkg/operator/encryption/kms/health/cmd.go
ibihim: Solved by addressing other comment.

pkg/operator/encryption/kms/health/cmd.go
ibihim: Technically we could. I consciously decided against it as the validation logic is very straightforward, easy to read... but when using regex, it might get some value.

pkg/operator/encryption/kms/health/cmd.go
ibihim: 👍

ibihim: I will add the regex, but a length check makes sense, right?

pkg/operator/encryption/kms/health/cmd.go
ibihim: Ok, will fix it.

pkg/operator/encryption/kms/health/cmd.go
ibihim: It was a hint for what is coming, but I can remove it.

pkg/operator/encryption/kms/health/cmd.go
ibihim: I thought it is easier to read, if set in a struct, but let me do it hardcoded.

ibihim: This is done [automagically by pflag.Flagset](https://github.com/openshift/library-go/blob/master/vendor/github.com/spf13/pflag/flag.go#L752-L758).

```bash
#!/usr/bin/env bash
# Run kms-health-reporter --help. The package has no main(), so we wire a
# throwaway one, go-run it, and clean up. Run from the repo root.
set -euo pipefail

pkg=github.com/openshift/library-go/pkg/operator/encryption/kms/health
dir=$(mktemp -d ./zzhelp.XXXXXX)
trap 'rm -rf "$dir"' EXIT

cat > "$dir/main.go" <<EOF
package main

import (
	"context"

	"$pkg"
	"k8s.io/client-go/rest"
)

func main() {
	cmd := health.NewCommand(context.Background(),
		func(*rest.Config) (v1helpers.OperatorClient, error) { return nil, nil })
	cmd.SetArgs([]string{"--help"})
	_ = cmd.Execute()
}
EOF

go run "./$dir"
```

```bash
./verify-help.sh
Observes co-located KMSv2 plugins and publishes status as an OperatorCondition.

Usage:
  kms-health-reporter [flags]

Flags:
  -h, --help                     help for kms-health-reporter
      --interval duration        cadence between probe+emit cycles (default 30s)
      --kms-sockets strings      KMS plugin endpoints in unix:// URI format (e.g. unix:///var/run/kmsplugin/kms-1.sock)
      --kubeconfig string        path to a kubeconfig; empty uses in-cluster config
      --node-name string         node name recorded in the condition used to help to identify the origin
      --read-timeout duration    deadline for each Status RPC (default 5s)
      --write-timeout duration   deadline for each condition update (default 10s)
```

ibihim: But I can add it by hand if you want.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: ok, let's don't forget to update the KEP to reflect that.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: yes.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: ah, great, no need to add it to the description then. thanks for checking.

pkg/operator/encryption/kms/health/cmd.go
ibihim: You are right, the loop is sequential, an overrun just delays the next tick. **Removed**: https://github.com/openshift/library-go/pull/2279/commits/0c437612b4697a7bb636b84a50e3e9d7bd9f6305

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,150 @@
...
+	if o.readTimeout <= 0 {
+		return fmt.Errorf("--read-timeout must be positive")
+	}
+	if o.writeTimeout <= 0 {
+		return fmt.Errorf("--write-timeout must be positive")
+	}
+	if o.readTimeout+o.writeTimeout >= o.interval {
+		return fmt.Errorf("--read-timeout (%v) + --write-timeout (%v) must be less than --interval (%v)",
+			o.readTimeout, o.writeTimeout, o.interval)
+	}
```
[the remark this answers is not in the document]
ibihim: The whole check was removed (see https://github.com/openshift/library-go/pull/2279#discussion_r3372863901): no sum, no overflow.

pkg/operator/encryption/kms/health/cmd.go
ibihim: Great idea, but the properties are unexported. Should I make them export to make it work with `klog.InfoS`?

pkg/operator/encryption/kms/health/cmd.go
ibihim: Added: https://github.com/openshift/library-go/pull/2279/commits/8078d6d08ccab28e4e789df42d40eb6d5e0824db

pkg/operator/encryption/kms/health/cmd.go
ibihim: With length check I meant the length of the slice containing the kms paths as those won't be validated by the regex.

pkg/operator/encryption/kms/health/cmd.go
p0lyn0mial: yes, let's export the options.

ibihim: Done. Note that the durations render as nanoseconds: `config={"KMSSockets":["unix:///var/run/kmsplugin/kms-1.sock"],"Interval":30000000000,...}`

p0lyn0mial: let's start simple, in the future we can add a custom marshaller or explicitly log individual fields.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -0,0 +1,110 @@
...
+	if len(o.KMSSockets) == 0 {
+		return fmt.Errorf("--kms-sockets is required, at least one")
+	}
+	for _, s := range o.KMSSockets {
+		if !kmsSocketPattern.MatchString(s) {
+			return fmt.Errorf("--kms-sockets entry %q must match %s", s, kmsSocketPattern)
+		}
+	}
+
+	if o.Interval <= 0 {
```
p0lyn0mial, line 74: in a follow-up maybe we should intro a cap for the max val for the timeouts ?

PR #2285, opened 2026-06-09
Added new test case for update image scenario

gangwgr:
Added new test case for update image scenario

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end scenario that verifies an in-place KMS configuration update does not create a new encryption key and validates propagation of the change.
  * Added a test helper that waits for at least one matching API server pod and ensures all matching pods satisfy a caller-defined container condition for more reliable rollout verification.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -403,6 +403,34 @@ func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient ku
 	require.NoError(t, err, "timed out waiting for pod to enter ImagePullBackOff in namespace %s", namespace)
 }
 
+// WaitForPodContainerCondition polls pods until at least one pod satisfies the given condition.
+// keyName is the current encryption key secret name, passed to match so callers can
+// correlate pod container names with the active key (e.g. vault-kms-plugin-42).
+func WaitForPodContainerCondition(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, labelSelector, keyName string, match func(pod corev1.Pod, keyName string) bool) {
+	t.Helper()
+	t.Logf("Waiting up to %s for a pod in %s (selector=%s, key=%s) to satisfy condition",
+		waitPollTimeout, namespace, labelSelector, keyName)
+	err := wait.PollUntilContextTimeout(ctx, waitPollInterval, waitPollTimeout, true, func(ctx context.Context) (bool, error) {
+		pods, err := kubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			t.Logf("Error listing pods: %v", err)
+			return false, nil
+		}
+		if len(pods.Items) == 0 {
+			t.Logf("No pods found yet in %s", namespace)
+			return false, nil
+		}
+		for _, pod := range pods.Items {
+			if match(pod, keyName) {
+				return true, nil
+			}
+		}
+		t.Logf("No pod yet satisfies condition")
+		return false, nil
+	})
+	require.NoError(t, err, "timed out waiting for a pod in %s to satisfy condition", namespace)
+}
+
 // WaitForCurrentKeyMigrated waits until the current (latest) encryption key has completed
 // migration for all target group resources. It fails if the key changes unexpectedly
 // (i.e. a different key than prevKeyMeta appears).
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -353,3 +353,66 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
 		}
 	}
 }
+
+// KMSInPlaceUpdateScenario tests that updating an in-place KMS config field
+// (e.g. kmsPluginImage) takes effect without creating a new encryption key.
+// The caller supplies Provider (initial valid config) and UpdatedProvider (same config
+// with one in-place field changed).
+type KMSInPlaceUpdateScenario struct {
+	BasicScenario
+	Provider        EncryptionProvider
+	UpdatedProvider EncryptionProvider
+	// WaitForPropagation is called after the in-place update to verify the change
+	// took effect. Receives the current encryption key so callers can match pod
+	// container names to the active key. Same pattern as WaitForStuck in
+	// KMSInvalidEncryptionRecoveryScenario.
+	WaitForPropagation func(ctx context.Context, t testing.TB, keyMeta EncryptionKeyMeta)
+}
+
+// TestKMSInPlaceUpdate validates in-place KMS config field updates:
+//  1. Apply valid provider and verify migration
+//  2. Update in-place field and verify no new encryption key is created
+//  3. WaitForPropagation — caller verifies the change took effect
+func TestKMSInPlaceUpdate(ctx context.Context, t testing.TB, scenario KMSInPlaceUpdateScenario) {
+	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	clientSet := GetClients(e)
+
+	require.NotNil(t, scenario.Provider.Setup, "Provider.Setup must not be nil")
+	require.NotNil(t, scenario.UpdatedProvider.Setup, "UpdatedProvider.Setup must not be nil")
+	require.NotNil(t, scenario.WaitForPropagation, "WaitForPropagation must not be nil")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.Provider.Type, "Provider must use KMS encryption type")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.UpdatedProvider.Type, "UpdatedProvider must use KMS encryption type")
+
+	steps := []testStep{
+		{name: "ApplyValidProviderAndVerifyMigration", testFunc: func(t testing.TB) {
+			SetAndWaitForEncryptionType(ctx, t, scenario.Provider, scenario.TargetGRs,
+				scenario.Namespace, scenario.LabelSelector)
+			scenario.AssertFunc(t, clientSet, scenario.Provider.Type,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "UpdateInPlaceField", testFunc: func(t testing.TB) {
+			keyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.UpdatedProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.UpdatedProvider.APIServerEncryption)
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, keyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "WaitForPropagation", testFunc: func(t testing.TB) {
+			keyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.WaitForPropagation(ctx, t, keyMeta)
+		}},
+	}
+
+	for _, step := range steps {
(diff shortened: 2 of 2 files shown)


test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,74 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.Provider.Type, "Provider must use KMS encryption type")
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.UpdatedProvider.Type, "UpdatedProvider must use KMS encryption type")
+
+	initialImage, err := kmsPluginImage(scenario.Provider.APIServerEncryption)
+	require.NoError(t, err)
+	updatedImage, err := kmsPluginImage(scenario.UpdatedProvider.APIServerEncryption)
+	require.NoError(t, err)
+	require.NotEqual(t, initialImage, updatedImage, "UpdatedProvider kmsPluginImage must differ from Provider")
+
+	var prevKeyMeta EncryptionKeyMeta
```
ardaguclu: We don't need this. Maybe it is better to call GetLastKeyMeta every time.

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,74 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+		{name: "UpdateKMSPluginImage", testFunc: func(t testing.TB) {
+			var err error
+			prevKeyMeta, err = GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.UpdatedProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.UpdatedProvider.APIServerEncryption)
+		}},
+		{name: "WaitForNoNewEncryptionKey", testFunc: func(t testing.TB) {
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, prevKeyMeta,
```
ardaguclu: Calling WaitForNoNewEncryptionKey definitely makes sense. But I wonder how we will ensure that the kms plugin container uses the new image.

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,62 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
 		}
 	}
 }
+
+// KMSPluginImageUpdateScenario tests that updating the KMS plugin image in-place
+// rolls out a new sidecar image without creating a new encryption key.
+// The caller supplies Provider (initial valid config) and UpdatedProvider (same config
+// with only kmsPluginImage changed).
+type KMSPluginImageUpdateScenario struct {
```
ardaguclu: As we did for invalidconfig scenario, I think it is better to rename this to more generic like `KMSInPlaceUpdateScenario`.

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,62 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+	BasicScenario
+	Provider        EncryptionProvider
+	UpdatedProvider EncryptionProvider
+	// WaitForUpdatedImage is called after the image update to verify the new image
+	// is running (e.g. by polling API server pod init containers). Caller-provided,
+	// same pattern as WaitForStuck in KMSInvalidEncryptionRecoveryScenario.
+	WaitForUpdatedImage func(ctx context.Context, t testing.TB)
```
ardaguclu: In order to keep it generic, we can rename it to `WaitForPropagation`.

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,62 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+	require.Equal(t, configv1.EncryptionTypeKMS, scenario.UpdatedProvider.Type, "UpdatedProvider must use KMS encryption type")
+
+	steps := []testStep{
+		{name: "ApplyValidProviderAndVerifyMigration", testFunc: func(t testing.TB) {
+			SetAndWaitForEncryptionType(ctx, t, scenario.Provider, scenario.TargetGRs,
+				scenario.Namespace, scenario.LabelSelector)
+			scenario.AssertFunc(t, clientSet, scenario.Provider.Type,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "UpdateKMSPluginImage", testFunc: func(t testing.TB) {
```
ardaguclu: `UpdateInPlaceField`

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,62 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+			keyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.UpdatedProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.UpdatedProvider.APIServerEncryption)
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, keyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "WaitForUpdatedImage", testFunc: func(t testing.TB) {
+			scenario.WaitForUpdatedImage(ctx, t)
```
ardaguclu: WaitForPropagation and callers will use custom wait command as we did for invalidconfigrecovery.

test/library/encryption/helpers.go
```diff
@@ -403,6 +403,50 @@ func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient ku
 	require.NoError(t, err, "timed out waiting for pod to enter ImagePullBackOff in namespace %s", namespace)
 }
 
+// WaitForPodInitContainerImage polls pods until each has a running init container
+// using the expected image. Init containers are scanned the same way as in
+// WaitForPodImagePullBackOff.
+func WaitForPodInitContainerImage(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, labelSelector, expectedImage string) {
```
ardaguclu: What about updating this function to be generic;

```go
 // WaitForPodContainerCondition polls pods until every pod has at least one
  // container status (init or regular, depending on the predicate) that satisfies
  // the given condition.
  func WaitForPodContainerCondition(
      ctx context.Context,
      t testing.TB,
      kubeClient kubernetes.Interface,
      namespace, labelSelector, conditionDesc string,
      match func(pod corev1.Pod) bool,
  ) {
      t.Helper()
      t.Logf("Waiting up to %s for pods in %s (selector=%s) to satisfy: %s",
          waitPollTimeout, namespace, labelSelector, conditionDesc)
      err := wait.PollUntilContextTimeout(ctx, waitPollInterval, waitPollTimeout, true,
          func(ctx context.Context) (bool, error) {
              pods, err := kubeClient.CoreV1().Pods(namespace).List(ctx,
                  metav1.ListOptions{LabelSelector: labelSelector})
              if err != nil {
                  t.Logf("Error listing pods: %v", err)
                  return false, nil
              }
              if len(pods.Items) == 0 {
                  t.Logf("No pods found yet in %s", namespace)
                  return false, nil
              }
              for _, pod := range pods.Items { 
                  if !match(pod) {
                      t.Logf("Pod %s does not yet satisfy condition: %s", pod.Name, conditionDesc)
                      return false, nil
                  }
              }
              return true, nil
          })
      require.NoError(t, err, "timed out waiting for pods in %s to satisfy: %s",
          namespace, conditionDesc)
  }
```

Callers will use it like this;

```go
 WaitForPropagation: func(ctx context.Context, t testing.TB) {
      WaitForPodContainerCondition(ctx, t, kubeClient, namespace, labelSelector,
          fmt.Sprintf("running init container with image %q", expectedImage),
          func(pod corev1.Pod) bool {
              if pod.Status.Phase != corev1.PodRunning {
                  return false
              }
              for _, cs := range pod.Status.InitContainerStatuses {
                  if cs.State.Running != nil && cs.Image == expectedImage {
                      return true
                  }
              }
              return false
          },
      )
  },
```

ardaguclu:
Thank  you
/lgtm
/approve
/hold
Can you please run this PR with https://github.com/openshift/library-go/pull/2241 on kas-o to see it passes?. If everything works, you can unhold this PR.

test/library/encryption/scenarios.go
```diff
@@ -353,3 +353,62 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+			keyMeta, err := GetLastKeyMeta(t, clientSet.Kube,
+				scenario.Namespace, scenario.LabelSelector)
+			require.NoError(t, err)
+			scenario.UpdatedProvider.Setup(ctx, t)
+			ApplyEncryption(ctx, t, scenario.UpdatedProvider.APIServerEncryption)
+			WaitForNoNewEncryptionKey(t, clientSet.Kube, keyMeta,
+				scenario.Namespace, scenario.LabelSelector)
+		}},
+		{name: "WaitForPropagation", testFunc: func(t testing.TB) {
+			scenario.WaitForPropagation(ctx, t)
```
ardaguclu: Maybe we should get last key as in line: 393 and pass it to waitforpropagation. So that if we find a matching container, we can compare container name with our key.

test/library/encryption/helpers.go
```diff
@@ -403,6 +403,32 @@ func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient ku
...
+		pods, err := kubeClient.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			t.Logf("Error listing pods: %v", err)
+			return false, nil
+		}
+		if len(pods.Items) == 0 {
+			t.Logf("No pods found yet in %s", namespace)
+			return false, nil
+		}
+		for _, pod := range pods.Items {
```
ardaguclu, line 423:
```suggestion
		for _, pod := range pods.Items {
			if match(pod, keyName) {
				return true, nil
			}
		}
       t.Logf("Pod %s does not yet satisfy condition: %s", pod.Name, conditionDesc)
		return false, nil

```

test/library/encryption/helpers.go
```diff
@@ -403,6 +403,34 @@ func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient ku
 	require.NoError(t, err, "timed out waiting for pod to enter ImagePullBackOff in namespace %s", namespace)
 }
 
+// WaitForPodContainerCondition polls pods until every pod satisfies the given condition.
+// keyName is the current encryption key secret name, passed to match so callers can
+// correlate pod container names with the active key (e.g. vault-kms-plugin-42).
+func WaitForPodContainerCondition(ctx context.Context, t testing.TB, kubeClient kubernetes.Interface, namespace, labelSelector, keyName, conditionDesc string, match func(pod corev1.Pod, keyName string) bool) {
```
ardaguclu: We don't need conditionDesc?

test/library/encryption/helpers.go
ardaguclu: Don't we need to continue iteration over the pods until we see match. So maybe we should use this ^^

gangwgr:
@ardaguclu holding pr to test with your pr, will unhold later
/hold

test/library/encryption/helpers.go
```diff
@@ -403,6 +403,34 @@ func WaitForPodImagePullBackOff(ctx context.Context, t testing.TB, kubeClient ku
...
+		}
+		if len(pods.Items) == 0 {
+			t.Logf("No pods found yet in %s", namespace)
+			return false, nil
+		}
+		for _, pod := range pods.Items {
+			if match(pod, keyName) {
+				return true, nil
+			}
+			t.Logf("Pod %s does not yet satisfy condition", pod.Name)
```
ardaguclu: This pod may not be the one we are looking for. Can we move this log just above the return false, nil

gangwgr: done

PR #2288, opened 2026-06-10
operator/encryption: add migration generation counter for KMS key re-migration

p0lyn0mial:
For KMS encryption mode, the remote key (e.g. Vault KEK) rotates independently without requiring a new local key secret. This commit introduces a generation counter mechanism that allows triggering re-migration of etcd data without creating a new key.

Two new annotations on the key secret:
- migration-generation: the requested migration generation
- migrated-generation: the generation migration was last completed for

When migration-generation > migrated-generation, MigratedFor() returns needsRemigration=true. The migration controller encodes the generation into the writeKey passed to EnsureMigration (e.g. "5-1"), causing the migrator to delete the old StorageVersionMigration CR and create a new one. After all resources complete re-migration, migrated-generation is updated to match.

The trigger mechanism that bumps migration-generation is left for a follow-up commit.

## Release Notes

* **New Features**
  * Added support for encryption key re-migration with generation tracking to handle scenarios such as KMS key rotation.
  * Enhanced the migration controller to distinguish between initial migration and re-migration of resources using generation-based lifecycle management.

* **Tests**
  * Added test coverage for migration generation scenarios and state serialization.

### DIFF
--- a/pkg/operator/encryption/state/types.go
+++ b/pkg/operator/encryption/state/types.go
@@ -50,6 +50,11 @@ type KeyState struct {
 	ExternalReason string
 	// stores all the KMS encryption mode related configurations
 	KMS *KMSState
+
+	// MigrationGeneration is the requested migration generation from the
+	// migration-generation annotation. A value greater than Migrated.Generation
+	// signals that re-migration is needed (e.g. after a remote KMS key rotation).
+	MigrationGeneration uint
 }
 
 func (k *KeyState) HasKMSEncryption() bool {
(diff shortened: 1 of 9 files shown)


pkg/operator/encryption/state/types.go
```diff
@@ -171,6 +162,9 @@ type MigrationState struct {
 	Timestamp time.Time
 	// the resources that were migrated at some point in time to this key.
 	Resources []schema.GroupResource
+	// Generation is the generation for which migration was last completed,
+	// from the migrated-generation annotation.
+	Generation uint
```
tjungblu, line 167: from our meeting: let's make this a string and we fill it with our KEKid.

PR #2292, opened 2026-06-10
CNTRLPLANE-3545: kms: expand credentials resolver to include configMap data

bertinatto:
## Summary by CodeRabbit

* **Refactor**
  * Renamed and reorganized internal credential/resolver interfaces and switched to a unified "reference data" concept; added support for resolving credentials from ConfigMap-backed entries.

* **Tests**
  * Updated and expanded tests to reflect the resolver API changes and to cover ConfigMap-backed credential path handling.

No direct user-facing changes; internal credential handling and test coverage improved.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/credentials.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/credentials.go
@@ -7,17 +7,17 @@ import (
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
 )
 
-// credentialResolver resolves KMS plugin credentials for a specific keyID
-// by looking up secret data and mapping it to values or file paths on disk.
-type credentialResolver struct {
+// referenceDataResolver resolves KMS plugin reference data for a specific keyID
+// by looking up secret and configMap data and mapping it to values or file paths on disk.
+type referenceDataResolver struct {
 	keyID                string
-	credentialsDir       string
+	referenceDataDir     string
 	pluginsSecretData    encryptiondata.KMSPluginsReferenceData
 	pluginsConfigMapData encryptiondata.KMSPluginsReferenceData
 }
 
 // SecretValue returns the value for the given secret name and data key.
-func (r *credentialResolver) SecretValue(secretName, dataKey string) (string, error) {
+func (r *referenceDataResolver) SecretValue(secretName, dataKey string) (string, error) {
 	sd, ok := r.pluginsSecretData.Get(r.keyID)
 	if !ok {
 		return "", fmt.Errorf("missing secret data for keyID %s", r.keyID)
@@ -30,7 +30,7 @@ func (r *credentialResolver) SecretValue(secretName, dataKey string) (string, er
 }
 
 // SecretFilePath returns the on-disk path where the data for the given secret name and data key is mounted.
-func (r *credentialResolver) SecretFilePath(secretName, dataKey string) (string, error) {
+func (r *referenceDataResolver) SecretFilePath(secretName, dataKey string) (string, error) {
 	sd, ok := r.pluginsSecretData.Get(r.keyID)
 	if !ok {
 		return "", fmt.Errorf("missing secret data for keyID %s", r.keyID)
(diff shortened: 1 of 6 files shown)


pkg/operator/encryption/kms/pluginlifecycle/credentials.go
```diff
@@ -10,13 +10,14 @@ import (
 // credentialResolver resolves KMS plugin credentials for a specific keyID
 // by looking up secret data and mapping it to values or file paths on disk.
 type credentialResolver struct {
```
p0lyn0mial: at this point this is not longer `credentialResolver`
more `resourceResolver` or `referenceResolver`

make sure we also reflect the change in the go doc.

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
```diff
@@ -41,3 +42,17 @@ func (r *credentialResolver) FilePath(secretName, dataKey string) (string, error
 	secretDataKey := encryptiondata.FormatKMSSecretDataKey(rawKey, r.keyID)
 	return filepath.Join(r.credentialsDir, secretDataKey), nil
 }
+
+// ConfigMapFilePath returns the on-disk path where the CA bundle for the given configmap name and data key is mounted.
```
p0lyn0mial: `CA bundle ` ? why this method is specific to `CA bundle` ?

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
```diff
@@ -41,3 +42,17 @@ func (r *credentialResolver) FilePath(secretName, dataKey string) (string, error
...
+	cd, ok := r.pluginsConfigMapData.Get(r.keyID)
+	if !ok {
+		return "", fmt.Errorf("missing configMap data for keyID %s", r.keyID)
+	}
+	rawKey, ok := cd.FlatEntry(configMapName, dataKey)
+	if !ok {
+		return "", fmt.Errorf("missing %s in configMap %s for keyID %s", dataKey, configMapName, r.keyID)
+	}
+	configMapDataKey := encryptiondata.FormatKMSConfigMapDataKey(rawKey, r.keyID)
+	return filepath.Join(r.credentialsDir, configMapDataKey), nil
```
p0lyn0mial: Should we add a defensive check for empty paths in `ConfigMapFilePath`? We could also do the same for `Secret*` method so call sites don't have to validate the result themselves.

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
bertinatto: replaced "CA bundle" with "data". Also changed "credential" from the method above to "data" for consistency

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
bertinatto: I noticed that `encryptiondata.FormatKMSConfigMapDataKey()` will never return an empty string, so the check will never fail.

should we change `encryptiondata.FormatKMSConfigMapDataKey()` to validate the inputs and return an error or an empty string if the inputs are empty?  Also, should we also validate its output [here](https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/encryptiondata/secret.go#L180-L180)?

bertinatto: CC @ardaguclu

ardaguclu: `rawKey` can not be empty. So it is impossible to have empty string from FormatKMSConfigMapDataKey

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
bertinatto: maybe `referenceDataResolver` to match `encryptiondata.KMSPluginsReferenceData`?

bertinatto: added a commit to rename to `referenceDataResolver`. Also, renamed `credentialsDir` to `referenceDataDir` for consistency

pkg/operator/encryption/kms/pluginlifecycle/credentials.go
bertinatto: does that mean I should not add the check?

ardaguclu: I think we don't need to add check

PR #2298, opened 2026-06-11
pkg/operator/encryption/kms/health: add KMS plugin health checker

ibihim:
## What

Adds a checker to the kms-health-reporter that probes each co-located KMSv2 plugin's Status endpoint and turns the responses into per-plugin health reports (healthy / unhealthy / error, one entry per plugin).

## Why

We want to track the state of all plugins. Currently we log it, later we add it to the status of each apiserver operator's CR status.

## Note

Compare to https://github.com/openshift/library-go/pull/2193

## Summary by CodeRabbit

* **New Features**
  * Continuous, cancellable KMS plugin health monitoring with periodic, jittered probes.
  * Per-plugin health reports including status, last-checked time, and key identifiers.

* **Bug Fixes**
  * Stricter KMS socket ID parsing and clearer initialization error handling.
  * More reliable kubeconfig/in-cluster config resolution.

* **Tests**
  * Added tests for concurrent probe execution and socket ID extraction.

### DIFF
--- a/pkg/operator/encryption/kms/health/cmd.go
+++ b/pkg/operator/encryption/kms/health/cmd.go
@@ -10,14 +10,29 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server"
+	k8senvelopekmsv2 "k8s.io/apiserver/pkg/storage/value/encrypt/envelope/kmsv2"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
+const providerName = "kms-health-reporter"
+
 // kmsSocketPattern matches the socket path each co-located KMSv2 plugin is
 // mounted at, e.g. unix:///var/run/kmsplugin/kms-1.sock.
-var kmsSocketPattern = regexp.MustCompile(`^unix:///var/run/kmsplugin/kms-\d+\.sock$`)
+var kmsSocketPattern = regexp.MustCompile(`^unix:///var/run/kmsplugin/kms-(\d+)\.sock$`)
+
+// keyIDFromSocket extracts the sequential key id captured by kmsSocketPattern,
+// e.g. "1" from unix:///var/run/kmsplugin/kms-1.sock.
+func keyIDFromSocket(socket string) (string, error) {
+	m := kmsSocketPattern.FindStringSubmatch(socket)
+	if m == nil {
+		return "", fmt.Errorf("socket %q must match %s", socket, kmsSocketPattern)
+	}
+	return m[1], nil
+}
 
 // options' flag-bound fields are exported so the struct can be logged as a
 // whole via klog.InfoS, which JSON-marshals its values.
@@ -45,7 +60,7 @@ func NewCommand(ctx context.Context, newOperatorClient func(*rest.Config) (v1hel
 			if err := o.validate(); err != nil {
 				return err
 			}
-			return o.run()
+			return o.run(ctx)
 		},
 	}
 	o.addFlags(cmd.Flags())
@@ -65,10 +80,15 @@ func (o *options) validate() error {
 	if len(o.KMSSockets) == 0 {
 		return fmt.Errorf("--kms-sockets is required, at least one")
 	}
+	socketSet := make(map[string]struct{}, len(o.KMSSockets))
 	for _, s := range o.KMSSockets {
 		if !kmsSocketPattern.MatchString(s) {
 			return fmt.Errorf("--kms-sockets entry %q must match %s", s, kmsSocketPattern)
 		}
+		if _, ok := socketSet[s]; ok {
+			return fmt.Errorf("--kms-sockets entry %q is duplicated", s)
+		}
+		socketSet[s] = struct{}{}
 	}
 
 	if o.Interval <= 0 {
@@ -87,8 +107,12 @@ func (o *options) validate() error {
 	return nil
 }
 
-func (o *options) run() error {
-	cfg, err := buildRESTConfig(o.Kubeconfig)
+func (o *options) run(ctx context.Context) error {
+	ctx = setupSignalContext(ctx)
+
+	// Empty kubeconfig falls back to the in-cluster config (service account
+	// token + KUBERNETES_SERVICE_HOST), which is the deployed path.
+	cfg, err := clientcmd.BuildConfigFromFlags("", o.Kubeconfig)
 	if err != nil {
 		return fmt.Errorf("build rest config: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -97,11 +111,45 @@ func (o *options) run() error {
...
+
+	for _, socket := range sockets {
+		keyID, err := keyIDFromSocket(socket)
+		if err != nil {
+			return nil, err
+		}
+
+		service, err := k8senvelopekmsv2.NewGRPCService(ctx, socket, providerName, timeout)
+		if err != nil {
+			return nil, fmt.Errorf("dial KMS plugin at %q: %w", socket, err)
```
ardaguclu: In which circumstances NewGPRCService returns errors?. I'm asking to understand we should return error here or, log the error and continue for others?.
It sounds like returning error is the right choice.

ardaguclu (commented):
This PR is nicely written without any AI generated redundant code. Just dropped a comment. Other than that looks good to me. Thank you.

pkg/operator/encryption/kms/health/cmd.go
ibihim: It seems to fail if the UDS path is malformed (which is a good reason, but is noop due to our validation):

https://github.com/ibihim/library-go/commit/c0f1eb12766e118b403b8706f2214d244f42a777

`grpc.Dial` is deprecated, but it is what is used here and the only thing that could fail, besides the `util.ParseEndpoint`. All other error paths are disabled at its call site, mostly by the hardcoded `insecure.NewCredentials()`. There is no connection test involved, the socket is first touched at the Status RPC, where a dead plugin surfaces as a per-plugin error.

tjungblu: surprisingly many cases:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/value/encrypt/envelope/kmsv2/grpc_service.go#L50

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -87,7 +101,7 @@ func (o *options) validate() error {
 	return nil
 }
 
-func (o *options) run() error {
+func (o *options) run(ctx context.Context) error {
```
p0lyn0mial, line 110: should we react to `SIGTERM` ?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -97,11 +111,45 @@ func (o *options) run() error {
...
+			return nil, fmt.Errorf("dial KMS plugin at %q: %w", socket, err)
+		}
+
+		plugins = append(plugins, plugin{keyID: keyID, service: service})
+	}
+
+	return plugins, nil
+}
+
 func buildRESTConfig(kubeconfig string) (*rest.Config, error) {
```
p0lyn0mial: do we need `buildRESTConfig` ? I think that `BuildConfigFromFlags` already falls back to in-cluster config when kubeconfig is empty.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -97,11 +111,45 @@ func (o *options) run() error {
...
+func buildPlugins(ctx context.Context, sockets []string, timeout time.Duration) ([]plugin, error) {
+	plugins := make([]plugin, 0, len(sockets))
+
+	for _, socket := range sockets {
+		keyID, err := keyIDFromSocket(socket)
+		if err != nil {
+			return nil, err
+		}
+
+		service, err := k8senvelopekmsv2.NewGRPCService(ctx, socket, providerName, timeout)
```
p0lyn0mial: should we make `providerName` unique per plugin ?

pkg/operator/encryption/kms/health/checker.go
```diff
@@ -0,0 +1,75 @@
...
+// See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kms/apis/v2/api.proto#L39
+const healthzOK = "ok"
+
+const (
+	statusHealthy   = "healthy"
+	statusUnhealthy = "unhealthy"
+	statusError     = "error"
+)
+
+type PluginHealthReport struct {
```
p0lyn0mial: does it have to be exported ?

pkg/operator/encryption/kms/health/checker.go
```diff
@@ -0,0 +1,75 @@
...
+	// KeyID is the controller's sequential key id; KEKID is the KMS provider's
+	// encryption key id. Distinct identifiers, easy to confuse.
+	KeyID       string    `json:"keyID"`
```
p0lyn0mial: do we need json tags on this struct ?

pkg/operator/encryption/kms/health/checker.go
```diff
@@ -0,0 +1,75 @@
...
+func newChecker(plugins []plugin) *checker {
+	return &checker{
+		plugins: plugins,
+		now:     time.Now,
+	}
+}
+
+// checkStatus never returns an error: a failed probe is encoded as a report
+// with Status "error" so the caller always gets one entry per plugin.
+func (c *checker) checkStatus(ctx context.Context) []PluginHealthReport {
```
p0lyn0mial: nit: feels more like `probeStatus` or `probeStatuses`  or `probeAll`

pkg/operator/encryption/kms/health/checker.go
```diff
@@ -0,0 +1,75 @@
...
+	LastChecked time.Time `json:"lastChecked"`
+	Detail      string    `json:"detail,omitempty"`
+}
+
+type plugin struct {
+	keyID   string
+	service kmsservice.Service
+}
+
+type checker struct {
```
p0lyn0mial: nit: `prober`

pkg/operator/encryption/kms/health/prober.go
```diff
@@ -0,0 +1,75 @@
...
+		now:     time.Now,
+	}
+}
+
+// checkStatus never returns an error: a failed probe is encoded as a report
+// with Status "error" so the caller always gets one entry per plugin.
+func (c *checker) checkStatus(ctx context.Context) []PluginHealthReport {
+	reports := make([]PluginHealthReport, 0, len(c.plugins))
+
+	// We could fan out if performance is an issue.
```
p0lyn0mial: we might reconsider this decision one the timeout is 30s or so then the lag would be 60s.

pkg/operator/encryption/kms/health/cmd.go
ibihim: I will update the logging.

ibihim: @tjungblu, yes, but did you check my test? Those cases never ever happen.

tjungblu: are you sure?
> 		grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),

will definitely check whether the other party responds before returning. It may timeout after some time however.

pkg/operator/encryption/kms/health/checker.go
ibihim: With the API change, rather not. Without it, yes.

tjungblu: maybe we can get Bryce to merge it today before shiftweek, let's see

pkg/operator/encryption/kms/health/prober.go
```diff
@@ -0,0 +1,88 @@
...
+// probeAll never returns an error: a failed probe is encoded as a report
+// with Status "error" so the caller always gets one entry per plugin.
+// Probes run concurrently so one hung plugin doesn't delay the others;
+// worst-case duration is one read-timeout, not the sum.
+func (p *prober) probeAll(ctx context.Context) []pluginHealthReport {
+	reports := make([]pluginHealthReport, len(p.plugins))
+
+	var wg sync.WaitGroup
+	for i, plugin := range p.plugins {
+		wg.Go(func() {
```
p0lyn0mial, line 59: that is a clean usage of go routines. didn't know you can use wg.Go

p0lyn0mial:
LGTM

let's fix `ci/prow/verify-deps` before merging.

p0lyn0mial:
oh, and let's squash the commits.

pkg/operator/encryption/kms/health/prober.go
```diff
@@ -0,0 +1,88 @@
...
+	return reports
+}
+
+func (p *prober) probe(ctx context.Context, plugin pluginClient) pluginHealthReport {
+	report := pluginHealthReport{
+		KeyID:       plugin.keyID,
+		LastChecked: p.now(),
+	}
+
+	resp, err := plugin.service.Status(ctx)
```
p0lyn0mial, line 74: just realized that resp is a pointer and can be `nil`. should we add a check ?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -87,8 +102,12 @@ func (o *options) validate() error {
...
 }
 
-func (o *options) run() error {
-	cfg, err := buildRESTConfig(o.Kubeconfig)
+func (o *options) run(ctx context.Context) error {
+	ctx = setupSignalContext(ctx)
+
+	// Empty kubeconfig falls back to the in-cluster config (service account
+	// token + KUBERNETES_SERVICE_HOST), which is the deployed path.
+	cfg, err := clientcmd.BuildConfigFromFlags("", o.Kubeconfig)
```
ardaguclu, line 115: nit: Ideally this should be in ~~`validate`~~ `complete` function not `run`. But this is definitely not a blocker.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -97,14 +116,58 @@ func (o *options) run() error {
...
 	return nil
 }
 
-func buildRESTConfig(kubeconfig string) (*rest.Config, error) {
-	if kubeconfig != "" {
-		return clientcmd.BuildConfigFromFlags("", kubeconfig)
+func buildPlugins(ctx context.Context, sockets []string, timeout time.Duration) ([]pluginClient, error) {
+	plugins := make([]pluginClient, 0, len(sockets))
+
+	for _, socket := range sockets {
```
p0lyn0mial, line 146: just realized that we don't check for duplicates, in theory a "user" can pass `--kms-sockets kms-1.sock,kms-1.sock` - we could add a check in a new pr.

pkg/operator/encryption/kms/health/prober.go
ibihim: I never used it, by I heard already of it, so here I am giving it a try.

ardaguclu:
LGTM

pkg/operator/encryption/kms/health/cmd.go
ibihim: more or less. I didn't do a deep analysis.

I brought a test for proof, you could come up with one of your own? :smile:

pkg/operator/encryption/kms/health/prober.go
ibihim: `NullPointerException`. Always worth checking.

pkg/operator/encryption/kms/health/cmd.go
ibihim: I will give it a try. Could be cool.

PR #2302, opened 2026-06-12
kms: rename credentials.go to reference_data_resolver.go

p0lyn0mial:
rename `credentials.go` to  `reference_data_resolver.go` to match the `referenceDataResolver` type defined inside which was renamed in #2292


bertinatto (commented):
/lgtm

PR #2305, opened 2026-06-15
Add new KMS to KMS migration test cases

gangwgr:
## Summary by CodeRabbit

* **Tests**
  * Added KMS-to-KMS migration scenario coverage, including a full primary → secondary → primary transition and verification that resources can be decrypted after the final identity switch.
  * Added step-driven KMS on/off flow tests that encrypt with the primary key, switch to identity to confirm decryption, then encrypt with the secondary key and confirm decryption again after switching back.
  * Improved the waiting/verification behavior for KMS-to-KMS encryption type transitions to better account for key recreation during updates.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -94,17 +94,26 @@ func SetAndWaitForEncryptionType(ctx context.Context, t testing.TB, provider Enc
 
 	apiServer, err := clientSet.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
 	require.NoError(t, err)
-	needsUpdate := !equality.Semantic.DeepEqual(apiServer.Spec.Encryption, provider.APIServerEncryption)
+	previousEncryption := apiServer.Spec.Encryption
+	needsUpdate := !equality.Semantic.DeepEqual(previousEncryption, provider.APIServerEncryption)
 	if needsUpdate {
 		if provider.Setup != nil {
 			provider.Setup(ctx, t)
 		}
-		t.Logf("Updating encryption configuration for APIServer from %#v to %#v", apiServer.Spec.Encryption, provider.APIServerEncryption)
+		t.Logf("Updating encryption configuration for APIServer from %#v to %#v", previousEncryption, provider.APIServerEncryption)
 		apiServer.Spec.Encryption = provider.APIServerEncryption
 		_, err = clientSet.ApiServerConfig.Update(ctx, apiServer, metav1.UpdateOptions{})
 		require.NoError(t, err)
 	} else {
 		t.Logf("APIServer is already configured to use %q mode", provider.Type)
+	}
+
+	// KMS-to-KMS migration: when both old and new are KMS but the config differs,
+	// the key controller creates a new key. We must wait for the next migrated key
+	// rather than asserting no new key is created.
+	if needsUpdate && provider.Type == configv1.EncryptionTypeKMS && previousEncryption.Type == configv1.EncryptionTypeKMS {
+		WaitForNextMigratedKey(t, clientSet.Kube, lastMigratedKeyMeta, defaultTargetGRs, namespace, labelSelector)
+		return clientSet
 	}
 
 	WaitForEncryptionKeyBasedOn(t, clientSet.Kube, lastMigratedKeyMeta, provider.Type, defaultTargetGRs, namespace, labelSelector)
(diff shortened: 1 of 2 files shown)


test/library/encryption/scenarios.go
```diff
@@ -354,6 +354,255 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
 }
 
+// TestKMSInvalidDuringMigration validates recovery when an invalid config is applied
+// while KMS encryption is already active and migrating. Unlike TestKMSInvalidEncryptionRecovery
+// (which starts with invalid config), this test first establishes valid KMS encryption,
+// then injects a fault mid-flight.
+//  1. Enable KMS with valid config and wait for migration to complete
+//  2. Apply invalid config (e.g. bad image) while encryption is active → wait stuck
+//  3. Apply valid config → verify migration completes on the same key (no new key minted)
+func TestKMSInvalidDuringMigration(ctx context.Context, t testing.TB, scenario KMSInvalidEncryptionRecoveryScenario) {
```
ardaguclu: Let's focus on kms to kms migration in that PR. So maybe we can remove this.

test/library/encryption/scenarios.go
```diff
@@ -354,6 +354,255 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+	}
+}
+
+// TestKMSMultiCycleRecovery validates that the operator recovers cleanly through
+// multiple invalid→valid cycles without leaving residual bad state.
+//  1. Apply invalid config → wait stuck
+//  2. Apply valid config → verify recovery and migration
+//  3. Apply invalid config again → wait stuck
+//  4. Apply valid config → verify recovery and migration again
+func TestKMSMultiCycleRecovery(ctx context.Context, t testing.TB, scenario KMSInvalidEncryptionRecoveryScenario) {
```
ardaguclu: Let's remove this scenario and focus on kms to kms migration

test/library/encryption/scenarios.go
```diff
@@ -354,6 +354,255 @@ func TestKMSInvalidEncryptionRecovery(ctx context.Context, t testing.TB, scenari
...
+
+// TestKMSToKMSRoundTrip validates round-trip migration: primary → secondary → primary.
+// Verifies that returning to the original KMS provider after migrating away works correctly
+// and resources remain accessible throughout.
+//  1. Create a test resource
+//  2. Encrypt with primary KMS provider and verify
+//  3. Migrate to secondary KMS provider and verify
+//  4. Migrate back to primary KMS provider and verify
+//  5. Switch to identity — verify resources are decrypted
+func TestKMSToKMSRoundTrip(ctx context.Context, t testing.TB, scenario KMSToKMSMigrationScenario) {
```
ardaguclu: This is almost identical to TestKMSToKMSMigration. Can we remove this and `scenario.PrimaryProvider` as last step to the other?

ardaguclu:
/hold
until we verify everything is correct

test/library/encryption/helpers.go
```diff
@@ -94,19 +94,28 @@ func SetAndWaitForEncryptionType(ctx context.Context, t testing.TB, provider Enc
...
 		require.NoError(t, err)
 	} else {
 		t.Logf("APIServer is already configured to use %q mode", provider.Type)
 	}
 
+	// KMS-to-KMS migration: when both old and new are KMS but the config differs,
+	// the key controller creates a new key. We must wait for the next migrated key
+	// rather than asserting no new key is created.
+	if needsUpdate && provider.Type == configv1.EncryptionTypeKMS && previousEncryption.Type == configv1.EncryptionTypeKMS {
+		WaitForNextMigratedKey(t, clientSet.Kube, lastMigratedKeyMeta, defaultTargetGRs, namespace, labelSelector)
```
ardaguclu, line 115: Are we sure that this logic works?. We still pass the same lastMigrateKeyMeta to `WaitForEncryptionKeyBasedOn` func. How don't we wait for no-new-key?

gangwgr: yes, this works the return clientSet on line 116. When we hit the KMS-to-KMS condition, we call WaitForNextMigratedKey and return immediately. The WaitForEncryptionKeyBasedOn call on line 119 is only reached for non-KMS-to-KMS cases (e.g., aescbc→kms or same config). So we never end up in the "no new key" path during KMS-to-KMS migration.

ardaguclu: ah, I missed `return clientSet`. Thanks of clarification

PR #2307, opened 2026-06-16
NO-JIRA: add preflight pod deployer

tjungblu:
## Summary by CodeRabbit

* **New Features**
  * Updated the in-cluster KMS preflight to a templated `apps/v1` Deployment with consistent labeling, templated service account, injected pod metadata, control-plane scheduling, and `emptyDir` plugin socket mounting.
  * Added explicit tolerations for both legacy and current control-plane taints.
  * Added a KMS call timeout argument to the preflight container.
* **Bug Fixes**
  * Improved status handling (Pending with no matching pods; errors with multiple matches) and Cleanup behavior when the Deployment is missing.
* **Refactor**
  * Switched to typed Deployment decoding with clearer validation errors.
* **Tests**
  * Added comprehensive unit tests for deploy/status/cleanup, idempotency, and selector behavior.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
+++ b/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
@@ -3,18 +3,44 @@ kind: Pod
 metadata:
   name: {{.PodName}}
   namespace: {{.Namespace}}
+  annotations:
+    encryption.apiserver.operator.openshift.io/kms-preflight-config-hash: {{.ConfigHash}}
 spec:
   # This is a one-shot preflight check, not a long-running pod.
   # The operator creates it, waits for completion, and inspects the result.
   restartPolicy: Never
+  priorityClassName: system-cluster-critical
+  nodeSelector:
+    node-role.kubernetes.io/master: ""
+  tolerations:
+    # Ensure pod can be scheduled on master nodes
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoExecute"
   containers:
     - name: kms-preflight-check
       image: {{.OperatorImage}}
       command: [{{.Command}}]
       args:
         - --kms-call-timeout={{.KMSCallTimeout}}
+      env:
+      - name: POD_NAME
+        valueFrom: { fieldRef: { fieldPath: metadata.name } }
+      - name: POD_NAMESPACE
+        valueFrom: { fieldRef: { fieldPath: metadata.namespace } }
+      - name: CONFIG_HASH
+        value: {{.ConfigHash}}
       # TODO: figure out good resource request values.
       resources:
         requests:
           memory: 50Mi
           cpu: 5m
+      volumeMounts:
+        - name: kms-plugin-socket
+          mountPath: /var/run/kmsplugin
+  volumes:
+    - name: kms-plugin-socket
+      emptyDir: {}
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -1,0 +1,90 @@
+package preflight
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+const (
+	preflightPodName = "kms-preflight"
+)
+
+type KMSPreflightDeployer interface {
+	// Deploy creates the preflight checker with the given config hash
+	Deploy(ctx context.Context, configHash string) error
+	Status(ctx context.Context) (corev1.PodStatus, error)
+	// Cleanup cleans up anything previously created.
+	Cleanup(ctx context.Context) error
+}
+
+// PodPreflightDeployer creates a one-shot pod to run the preflight checker as a command,
+// with the plugin injected as a sidecar container.
+type PodPreflightDeployer struct {
+	namespace string
+	// It uses direct API client instead of a lister to be consistent with what
+	// encryption controllers/components actually use.
+	// In addition to that running the preflight check is not a very frequent operation.
+	coreClient      corev1client.CoreV1Interface
+	operatorImage   string
+	operatorCommand []string
+	kmsCallTimeout  time.Duration
+}
+
+func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) error {
+	pod, err := generatePodTemplate(
+		preflightPodName,
+		d.namespace,
+		configHash,
+		d.operatorImage,
+		d.operatorCommand,
+		d.kmsCallTimeout,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to generate preflight pod template: %w", err)
+	}
+
+	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+
+	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
+	return err
+}
+
+func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, error) {
+	// preflight status checks are not very frequent, so we use the live client instead of a cached lister
+	pod, err := d.coreClient.Pods(d.namespace).Get(ctx, preflightPodName, metav1.GetOptions{})
+	if err != nil {
+		return corev1.PodStatus{}, fmt.Errorf("failed to get pod for preflight %s/%s: %w", d.namespace, preflightPodName, err)
+	}
+
+	return pod.Status, nil
+}
+
+func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
+	err := d.coreClient.Pods(d.namespace).Delete(ctx, preflightPodName, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err)
+	}
+	return nil
+}
+
+func NewPodPreflightDeployer(
+	namespace string,
+	coreClient corev1client.CoreV1Interface,
+	operatorImage string,
+	operatorCommand []string,
+	kmsCallTimeout time.Duration,
+) *PodPreflightDeployer {
+	return &PodPreflightDeployer{
+		namespace:       namespace,
+		coreClient:      coreClient,
+		operatorImage:   operatorImage,
+		operatorCommand: operatorCommand,
+		kmsCallTimeout:  kmsCallTimeout,
+	}
+}
--- a/pkg/operator/encryption/kms/preflight/deployer_test.go
+++ b/pkg/operator/encryption/kms/preflight/deployer_test.go
@@ -1,0 +1,270 @@
+package preflight
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	clienttesting "k8s.io/client-go/testing"
+)
+
+const (
+	testNamespace           = "openshift-kube-apiserver"
+	testConfigHash          = "abc123"
+	testOperatorImage       = "quay.io/openshift-release-dev/ocp-v5.0-art-dev@sha256:test"
+	configHashAnnotationKey = "encryption.apiserver.operator.openshift.io/kms-preflight-config-hash"
+	testDeployerTimeout     = 10 * time.Second
+)
+
+var testOperatorCommand = []string{"cluster-kube-apiserver-operator", "kms-preflight"}
+
+var expectedDeployPodYAML = `
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kms-preflight
+  namespace: openshift-kube-apiserver
+  annotations:
+    encryption.apiserver.operator.openshift.io/kms-preflight-config-hash: abc123
+spec:
+  restartPolicy: Never
+  priorityClassName: system-cluster-critical
+  nodeSelector:
+    node-role.kubernetes.io/master: ""
+  tolerations:
+    - key: node-role.kubernetes.io/master
+      operator: Exists
+      effect: NoSchedule
+    - key: node-role.kubernetes.io/master
+      operator: Exists
+      effect: NoExecute
+  containers:
+    - name: kms-preflight-check
+      image: quay.io/openshift-release-dev/ocp-v5.0-art-dev@sha256:test
+      command: ["cluster-kube-apiserver-operator","kms-preflight"]
+      args:
+        - --kms-call-timeout=10s
+      env:
+      - name: POD_NAME
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.name
+      - name: POD_NAMESPACE
+        valueFrom:
+          fieldRef:
+            fieldPath: metadata.namespace
+      - name: CONFIG_HASH
+        value: abc123
+      resources:
+        requests:
+          memory: 50Mi
+          cpu: 5m
+      volumeMounts:
+        - name: kms-plugin-socket
+          mountPath: /var/run/kmsplugin
+  volumes:
+    - name: kms-plugin-socket
+      emptyDir: {}
+`
+
+func newTestDeployer(t *testing.T, objects ...runtime.Object) (*PodPreflightDeployer, *fake.Clientset) {
+	t.Helper()
+
+	kubeClient := fake.NewSimpleClientset(objects...)
+	deployer := NewPodPreflightDeployer(
+		testNamespace,
+		kubeClient.CoreV1(),
+		testOperatorImage,
+		testOperatorCommand,
+		testDeployerTimeout,
+	)
+	return deployer, kubeClient
+}
+
+func TestPodPreflightDeployer_Deploy(t *testing.T) {
+	ctx := context.Background()
+	deployer, kubeClient := newTestDeployer(t)
+
+	expectedPod, err := resourceread.ReadPodV1([]byte(expectedDeployPodYAML))
+	if err != nil {
+		t.Fatalf("failed to parse expected pod YAML: %v", err)
+	}
+
+	if err := deployer.Deploy(ctx, testConfigHash); err != nil {
+		t.Fatalf("Deploy() error = %v", err)
+	}
+
+	actions := kubeClient.Actions()
+	if len(actions) != 1 {
+		t.Fatalf("expected 1 action, got %d: %#v", len(actions), actions)
+	}
+	if !actions[0].Matches("create", "pods") {
+		t.Fatalf("unexpected action: %#v", actions[0])
+	}
+
+	createAction, ok := actions[0].(clienttesting.CreateAction)
+	if !ok {
+		t.Fatalf("expected CreateAction, got %T", actions[0])
+	}
+	if createAction.GetNamespace() != testNamespace {
+		t.Fatalf("unexpected namespace %q", createAction.GetNamespace())
+	}
+
+	actualPod, ok := createAction.GetObject().(*corev1.Pod)
+	if !ok {
+		t.Fatalf("expected *corev1.Pod, got %T", createAction.GetObject())
+	}
+	if !equality.Semantic.DeepEqual(expectedPod, actualPod) {
+		t.Fatalf("pod does not match expected:\ngot:  %+v\nwant: %+v", actualPod, expectedPod)
+	}
+}
+
+func TestPodPreflightDeployer_Status(t *testing.T) {
+	scenarios := []struct {
+		name        string
+		objects     []runtime.Object
+		expectPhase corev1.PodPhase
+		expectErr   string
+	}{
+		{
+			name:      "missing pod returns error",
+			expectErr: "failed to get pod",
+		},
+		{
+			name: "pod returns pod status",
+			objects: []runtime.Object{
+				&corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      preflightPodName,
+						Namespace: testNamespace,
+					},
+					Status: corev1.PodStatus{Phase: corev1.PodRunning},
+				},
+			},
+			expectPhase: corev1.PodRunning,
+		},
+	}
+
+	for _, scenario := range scenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			deployer, _ := newTestDeployer(t, scenario.objects...)
+
+			status, err := deployer.Status(context.Background())
+			if scenario.expectErr != "" {
+				if err == nil || !strings.Contains(err.Error(), scenario.expectErr) {
+					t.Fatalf("expected error containing %q, got %v", scenario.expectErr, err)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("Status() error = %v", err)
+			}
+			if status.Phase != scenario.expectPhase {
+				t.Fatalf("expected phase %q, got %q", scenario.expectPhase, status.Phase)
+			}
+		})
+	}
+}
+
+func TestPodPreflightDeployer_Cleanup(t *testing.T) {
+	existingPod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      preflightPodName,
+			Namespace: testNamespace,
+			Annotations: map[string]string{
+				configHashAnnotationKey: testConfigHash,
+			},
+		},
+	}
+
+	scenarios := []struct {
+		name          string
+		objects       []runtime.Object
+		setupClient   func(*fake.Clientset)
+		expectErr     string
+		verifyActions func(t *testing.T, actions []clienttesting.Action)
+	}{
+		{
+			name:    "deletes existing pod",
+			objects: []runtime.Object{existingPod},
+			verifyActions: func(t *testing.T, actions []clienttesting.Action) {
+				if len(actions) != 1 {
+					t.Fatalf("expected 1 action, got %d: %#v", len(actions), actions)
+				}
+				if !actions[0].Matches("delete", "pods") {
+					t.Fatalf("unexpected action: %#v", actions[0])
+				}
+				deleteAction, ok := actions[0].(clienttesting.DeleteAction)
+				if !ok {
+					t.Fatalf("expected DeleteAction, got %T", actions[0])
+				}
+				if deleteAction.GetName() != preflightPodName {
+					t.Fatalf("unexpected pod name %q", deleteAction.GetName())
+				}
+				if deleteAction.GetNamespace() != testNamespace {
+					t.Fatalf("unexpected namespace %q", deleteAction.GetNamespace())
+				}
+			},
+		},
+		{
+			name: "missing pod is not an error",
+			verifyActions: func(t *testing.T, actions []clienttesting.Action) {
+				if len(actions) != 1 {
+					t.Fatalf("expected 1 action, got %d: %#v", len(actions), actions)
+				}
+				if !actions[0].Matches("delete", "pods") {
+					t.Fatalf("unexpected action: %#v", actions[0])
+				}
+			},
+		},
+		{
+			name:    "delete error is returned",
+			objects: []runtime.Object{existingPod},
+			setupClient: func(kubeClient *fake.Clientset) {
+				kubeClient.PrependReactor("delete", "pods", func(action clienttesting.Action) (bool, runtime.Object, error) {
+					return true, nil, apierrors.NewForbidden(corev1.Resource("pods"), preflightPodName, nil)
+				})
+			},
+			expectErr: "failed to delete pod",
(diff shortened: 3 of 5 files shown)


pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,128 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type KMSPreflightDeployer interface {
+	// Deploy creates the preflight checker with the given identifier (e.g. the hash of the kms config)
+	Deploy(ctx context.Context, identifier string, pluginConfig configv1.KMSPluginConfig) error
```
p0lyn0mial: I prefer smaller PRs. Could the first iteration be to just deploy the pod without the sidecar/configuration ?

tjungblu: it does that.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-deployment.gotpl
```diff
@@ -0,0 +1,38 @@
...
+  selector:
+    matchLabels:
+      app: {{.LabelApp}}
+  template:
+    metadata:
+      labels:
+        app: {{.LabelApp}}
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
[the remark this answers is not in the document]
tjungblu: just used master now, because this is what clusterbot uses.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+		operatorServiceAccount: operatorServiceAccount,
+	}
+}
+
+// StaticPodPreflightDeployer will create a real pod to install a static pod in the host filesystem, which contains
+// the preflight checker as a command and the plugin injected as a sidecar container.
+type StaticPodPreflightDeployer struct {
+}
+
+func NewStaticPodPreflightDeployer() *StaticPodPreflightDeployer {
```
p0lyn0mial: could we remove it for now ? (unused during the first iteration)

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+		podClient:              podClient,
+		operatorImage:          operatorImage,
+		operatorBinaryName:     operatorBinaryName,
+		operatorServiceAccount: operatorServiceAccount,
+	}
+}
+
+// StaticPodPreflightDeployer will create a real pod to install a static pod in the host filesystem, which contains
+// the preflight checker as a command and the plugin injected as a sidecar container.
+type StaticPodPreflightDeployer struct {
```
p0lyn0mial: could we remove it for now ? (unused during the first iteration)

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+		d.operatorBinaryName,
+		d.operatorServiceAccount,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to generate deployment template: %w", err)
+	}
+
+	// TODO(thomas): inject KMS plugin sidecar container into deployment.Spec.Template.Spec
+
+	err = d.appsClient.Deployments(d.namespace).Delete(ctx, deployment.Name, metav1.DeleteOptions{})
```
p0lyn0mial: shouldn't the Deploy only create the deployment ?

shouldn't the cleanup logic be provided by the Cleanup function ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	appsclientv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type KMSPreflightDeployer interface {
+	// Deploy creates the preflight checker with the given identifier (e.g. the hash of the kms config)
+	Deploy(ctx context.Context, identifier string) error
+	// Status returns the pod status using the given identifier (e.g. the hash of the kms config)
+	Status(ctx context.Context, identifier string) (corev1.PodStatus, error)
```
p0lyn0mial: could we change the syntax to `Status(ctx context.Context) (corev1.PodStatus, error)` ?

the reason is that i was planing to use [readiness gates](https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/controllers/kms_preflight_controller.go#L192C17-L192C32) for reporting checks and the hash/generation.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+	// Cleanup cleans up anything previously created with the given identifier (e.g. the hash of the kms config)
+	Cleanup(ctx context.Context, identifier string) error
```
p0lyn0mial: since the pod name is stable we could change the syntax to `Cleanup(ctx context.Context) error`

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	_, err = d.appsClient.Deployments(d.namespace).Create(ctx, deployment, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create deployment %s/%s: %w", d.namespace, deployment.Name, err)
+	}
+	return nil
+}
+
+func (d *DeploymentPreflightDeployer) Status(ctx context.Context, identifier string) (corev1.PodStatus, error) {
+	pods, err := d.podClient.List(ctx, metav1.ListOptions{
```
p0lyn0mial: since the pod name is stable we don't have to List we could Get. WDYT?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if len(pods.Items) > 1 {
+		return corev1.PodStatus{}, fmt.Errorf("expected 1 pod for preflight %s/%s, got %d", d.namespace, deploymentName, len(pods.Items))
+	}
+
+	return pods.Items[0].Status, nil
+}
+
+func (d *DeploymentPreflightDeployer) Cleanup(ctx context.Context, identifier string) error {
+	deployment, err := d.appsClient.Deployments(d.namespace).Get(ctx, deploymentName, metav1.GetOptions{})
```
p0lyn0mial: since the controller owns the state i think we can skip and just call Delete. WDYT ?

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: also, please justify the usage of the live client.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+	namespace          string
+	appsClient         appsclientv1.AppsV1Interface
+	podClient          corev1client.PodInterface
+	operatorImage      string
+	operatorBinaryName string
+	// TODO(thomas): we need to figure out which service account to use, probably the same as the health monitor for now
+	operatorServiceAccount string
+}
+
+func (d *DeploymentPreflightDeployer) Deploy(
```
p0lyn0mial: I think that the convention is to keep the signature on one line.

pkg/operator/encryption/kms/preflight/deployment_template.go
```diff
@@ -0,0 +1,67 @@
...
+	return map[string]string{
+		"app":              labelApp,
+		identifierLabelKey: identifier,
+	}
+}
+
+// GenerateDeploymentTemplate renders the KMS preflight deployment YAML template.
+// The deployment runs a single replica on control plane nodes with the preflight
+// checker as the main container command. The KMS plugin sidecar is injected separately.
+func GenerateDeploymentTemplate(namespace, identifier, operatorImage, operatorBinaryName, operatorServiceAccount string) (*appsv1.Deployment, error) {
```
p0lyn0mial: does this method have to be exported ?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-deployment.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
```
p0lyn0mial: nit: could this be `.yaml` file ? I think that  @every other templated manifest in this repo uses .yaml.
we could stay consistent.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+		LabelSelector: labels.Set(preflightLabelSelector(identifier)).String(),
+	})
+	if err != nil {
+		return corev1.PodStatus{}, fmt.Errorf("failed to list pods for preflight %s/%s: %w", d.namespace, deploymentName, err)
+	}
+	if len(pods.Items) == 0 {
+		return corev1.PodStatus{Phase: corev1.PodPending}, nil
+	}
+	if len(pods.Items) > 1 {
+		return corev1.PodStatus{}, fmt.Errorf("expected 1 pod for preflight %s/%s, got %d", d.namespace, deploymentName, len(pods.Items))
```
p0lyn0mial: can deployment have more than one pod during rollout ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (d *DeploymentPreflightDeployer) Status(ctx context.Context, identifier string) (corev1.PodStatus, error) {
+	pods, err := d.podClient.List(ctx, metav1.ListOptions{
+		LabelSelector: labels.Set(preflightLabelSelector(identifier)).String(),
+	})
+	if err != nil {
+		return corev1.PodStatus{}, fmt.Errorf("failed to list pods for preflight %s/%s: %w", d.namespace, deploymentName, err)
+	}
+	if len(pods.Items) == 0 {
+		return corev1.PodStatus{Phase: corev1.PodPending}, nil
```
p0lyn0mial: mhm, what if a pod exists and it is pending ? how do we distinguish those ? do we have to ?

p0lyn0mial (commented):
did a first pass. let me know what you think.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: yes

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: yes

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: what if it already exists because a previous cleanup failed? should it fail subsequent deploys?

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: I don't really see how this relates to readiness gates, but I need some identifier to get the pod by the hash. So I don't think this will work.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: removed

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: the pod name is not stable, it will have a random suffix generated by the deployment controller

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: that's fair, just called delete now.

tjungblu: that's fair, just called delete now.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: not aware of any convention, but made it a single line again

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: do you want me to use an informer instead?

pkg/operator/encryption/kms/preflight/deployment_template.go
tjungblu: no, made it private

pkg/operator/encryption/kms/preflight/assets/kms-preflight-deployment.yaml
tjungblu: made it a yaml

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: depends on how you configure them, but do you want a status of multiple pods?

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: you tell me how you want to deal with this in the controller. I can also just return nil and an error for that case

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: move to a pod lister instead

pkg/operator/encryption/kms/preflight/assets/kms-preflight-deployment.yaml
```diff
@@ -0,0 +1,49 @@
...
+kind: Deployment
```
p0lyn0mial: I'm wondering why we would want to run a deployment. Are there any deployment level fields that affect pod runtime behavior? 

It seems that these fields are more about orchestration. In the future, we could simply copy `oasDeployment.Spec.Template.Spec`. Maybe we could start simple and learn what we actually need?

tjungblu: The deployment works fine, what other workload do you suggest to run?
We can't just run a pod, because it will disappear and may not actually run to completion.

pkg/operator/resource/resourceread/apps.go
```diff
@@ -18,11 +20,23 @@ func init() {
...
-	requiredObj, err := runtime.Decode(appsCodecs.UniversalDecoder(appsv1.SchemeGroupVersion), objBytes)
+	requiredObj, err := ReadDeploymentV1(objBytes)
 	if err != nil {
 		panic(err)
 	}
-	return requiredObj.(*appsv1.Deployment)
+	return requiredObj
+}
+
+func ReadDeploymentV1(objBytes []byte) (*appsv1.Deployment, error) {
```
p0lyn0mial: no longer needed. we can remove it.

pkg/operator/encryption/kms/preflight/pod_template.go
```diff
@@ -9,37 +9,69 @@ import (
...
+) (*corev1.Pod, error) {
 	rawManifest := mustAsset("assets/kms-preflight-pod.yaml")
 
 	operatorCommandQuoted := make([]string, len(operatorCommand))
 	for i, cmd := range operatorCommand {
 		operatorCommandQuoted[i] = fmt.Sprintf("%q", cmd)
 	}
 
+	if kmsCallTimeout <= 0 {
+		kmsCallTimeout = defaultKMSCallTimeout
```
p0lyn0mial: this must come from the controller and match what we will use in production. 
the key controller is responsible for assigning a default value to it.

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: i think that the controller will decide when to cleanup.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -3,18 +3,43 @@ kind: Pod
...
+      env:
+      - name: POD_NAME
+        valueFrom: { fieldRef: { fieldPath: metadata.name } }
+      - name: POD_NAMESPACE
+        valueFrom: { fieldRef: { fieldPath: metadata.namespace } }
       resources:
         requests:
           memory: 50Mi
           cpu: 5m
+      volumeMounts:
```
p0lyn0mial, line 41: could we delay adding the mounts once we have the lifecycle function ?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -3,18 +3,43 @@ kind: Pod
...
+      operator: "Exists"
+      effect: "NoExecute"
   containers:
     - name: kms-preflight-check
       image: {{.OperatorImage}}
       command: [{{.Command}}]
       args:
         - --kms-call-timeout={{.KMSCallTimeout}}
-      # TODO: figure out good resource request values.
+      env:
```
p0lyn0mial, line 29: why is this required ?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -3,18 +3,43 @@ kind: Pod
 metadata:
   name: {{.PodName}}
   namespace: {{.Namespace}}
+  labels:
```
p0lyn0mial: The pod already has a static name so we don't need a label selector to find it.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func NewPodPreflightDeployer(
+	namespace string,
+	coreClient corev1client.CoreV1Interface,
+	podLister corev1listers.PodNamespaceLister,
+	operatorImage string,
+	operatorBinaryName string,
+	operatorServiceAccount string,
```
p0lyn0mial: I think that the deployer should be responsible of creating the SA, RBAC and the pod.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+
+	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create pod %s/%s: %w", d.namespace, pod.Name, err)
+	}
+	return nil
+}
+
+func (d *PodPreflightDeployer) Status(ctx context.Context, identifier string) (corev1.PodStatus, error) {
+	pods, err := d.podLister.List(labels.Set(preflightLabelSelector(identifier)).AsSelector())
```
p0lyn0mial: The pod already has a static name so we don't need a label selector to find it.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return nil
+}
+
+func NewPodPreflightDeployer(
+	namespace string,
+	coreClient corev1client.CoreV1Interface,
+	podLister corev1listers.PodNamespaceLister,
+	operatorImage string,
+	operatorBinaryName string,
```
p0lyn0mial: could this be `operatorCommand []string` as in `GeneratePodTemplate` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (d *PodPreflightDeployer) Status(ctx context.Context, identifier string) (corev1.PodStatus, error) {
+	pods, err := d.podLister.List(labels.Set(preflightLabelSelector(identifier)).AsSelector())
+	if err != nil {
+		return corev1.PodStatus{}, fmt.Errorf("failed to list pods for preflight %s/%s: %w", d.namespace, preflightPodName, err)
+	}
+	if len(pods) == 0 {
+		return corev1.PodStatus{Phase: corev1.PodPending}, nil
+	}
+	if len(pods) > 1 {
```
p0lyn0mial: since there will be a single pod we don't need that check anymore.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create pod %s/%s: %w", d.namespace, pod.Name, err)
+	}
+	return nil
+}
+
+func (d *PodPreflightDeployer) Status(ctx context.Context, identifier string) (corev1.PodStatus, error) {
```
p0lyn0mial: I think we could simplify:

```
func (d *PodPreflightDeployer) Status(ctx context.Context) (*corev1.PodStatus, error) {
    pod, err := d.coreClient.Pods(d.namespace).Get(ctx, preflightPodName, metav1.GetOptions{})
    if err != nil {
          return nil, err
     }
    return &pod.Status, nil
  }
```


then the controller could do:

```
status, err := deployer.Status(ctx)
    return deployer.Deploy(ctx, identifier)
}
...
```

p0lyn0mial (commented):
did another pass.
haven't looked at the tests.

pkg/operator/resource/resourceread/apps.go
tjungblu: removed

pkg/operator/encryption/kms/preflight/pod_template.go
tjungblu: okay, I remove the default and rely on the parameter

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
tjungblu: how would the pod know where to put the status?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
tjungblu: why? what's the matter with adding this now

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: okay but we decided to use the same as the health monitor, so it must be injected

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: that's right, removed it from the interface

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: updated

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: fixed

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: fixed

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
tjungblu: removed

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: removed

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: removed the SA

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+		t.Fatalf("Deploy() error = %v", err)
+	}
+
+	pod, err := deployer.coreClient.Pods(testNamespace).Get(ctx, preflightPodName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get pod: %v", err)
+	}
+
+	if pod.Annotations[configHashAnnotationKey] != testConfigHash {
+		t.Fatalf("unexpected pod config hash annotation %q=%q", configHashAnnotationKey, pod.Labels[configHashAnnotationKey])
```
p0lyn0mial: we should read `pod.Annotations` not `pod.Labels`

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func TestPodPreflightDeployer_Cleanup(t *testing.T) {
+	ctx := context.Background()
+	existing := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      preflightPodName,
+			Namespace: testNamespace,
+			Labels: map[string]string{
+				configHashAnnotationKey: testConfigHash,
```
p0lyn0mial, line 183: we should put the hash into Annotations not Labels

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+}
+
+func TestPodPreflightDeployer_Cleanup_deleteError(t *testing.T) {
+	ctx := context.Background()
+	existing := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      preflightPodName,
+			Namespace: testNamespace,
+			Labels: map[string]string{
```
p0lyn0mial: we should put the hash into Annotations not Labels

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+func TestPodPreflightDeployer_Deploy(t *testing.T) {
+	ctx := context.Background()
+	deployer := newTestDeployer(t)
+
+	if err := deployer.Deploy(ctx, testConfigHash); err != nil {
+		t.Fatalf("Deploy() error = %v", err)
+	}
+
+	pod, err := deployer.coreClient.Pods(testNamespace).Get(ctx, preflightPodName, metav1.GetOptions{})
```
p0lyn0mial: is this how a fake client should be used ?
i think the more idiomatic way is to check the recorded Actions.

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		t.Fatalf("failed to get pod: %v", err)
+	}
+
+	if pod.Annotations[configHashAnnotationKey] != testConfigHash {
```
p0lyn0mial: instead of checking every field could we define an expected pod and compare against it ?

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,199 @@
...
+				t.Fatalf("Status() error = %v", err)
+			}
+			if status.Phase != scenario.expectPhase {
+				t.Fatalf("expected phase %q, got %q", scenario.expectPhase, status.Phase)
+			}
+		})
+	}
+}
+
+func TestPodPreflightDeployer_Cleanup(t *testing.T) {
```
p0lyn0mial, line 177: could we create a single table test for three Cleanup tests we have now ?

p0lyn0mial (commented):
did a pass on the unit tests, left a few comments.

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -0,0 +1,227 @@
...
+		testDeployerTimeout,
+	)
+	return deployer, kubeClient
+}
+
+func TestPodPreflightDeployer_Deploy(t *testing.T) {
+	ctx := context.Background()
+	deployer, kubeClient := newTestDeployer(t)
+
+	expectedPod, err := generatePodTemplate(
```
p0lyn0mial: should we use a production function to generate an expected pod or should we just define an expected pod ?

if there is a bug in the generatePodTemplate then we will not catch it.

tjungblu: TestGeneratePodTemplate already does that, doesn't it?

pkg/operator/encryption/kms/preflight/deployer_test.go
tjungblu: fixed

pkg/operator/encryption/kms/preflight/deployer_test.go
tjungblu: fixed

pkg/operator/encryption/kms/preflight/deployer_test.go
p0lyn0mial: `TestGeneratePodTemplate` tests `generatePodTemplate ` and compares against the expected 
https://github.com/openshift/library-go/blob/b41e1d0a8abb801ce6a51768dc6ff03da9c2efed/pkg/operator/encryption/kms/preflight/pod_template_test.go#L73

tjungblu: sounds like unnecessary redundancy, but added a static yaml to compare against.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -0,0 +1,90 @@
...
+
+// PodPreflightDeployer creates a one-shot pod to run the preflight checker as a command,
+// with the plugin injected as a sidecar container.
+type PodPreflightDeployer struct {
+	namespace string
+	// It uses direct API client instead of a lister to be consistent with what
+	// encryption controllers/components actually use.
+	// In addition to that running the preflight check is not a very frequent operation.
+	coreClient      corev1client.CoreV1Interface
+	operatorImage   string
```
ardaguclu, line 34: Just for awareness: In order to make this working in cluster-authentication-operator, we need this PR https://github.com/openshift/cluster-authentication-operator/pull/927

PR #2310, opened 2026-06-17
NO-JIRA: configure kms socket endpoint in cmd

tjungblu:
This adds a new required arg to the preflight checker cmd.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/cmd.go
+++ b/pkg/operator/encryption/kms/preflight/cmd.go
@@ -12,10 +12,9 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const kmsSocketEndpoint = "unix:///var/run/kmsplugin/kms.sock"
-
 type options struct {
-	kmsCallTimeout time.Duration
+	kmsCallTimeout    time.Duration
+	kmsSocketEndpoint string
 }
 
 // NewCommand creates the kms-preflight cobra command.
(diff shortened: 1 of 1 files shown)


pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -39,21 +38,27 @@ func NewCommand(ctx context.Context) *cobra.Command {
...
 
 func (o *options) validate() error {
 	if o.kmsCallTimeout <= 0 {
 		return fmt.Errorf("--kms-call-timeout must be greater than 0")
 	}
+
+	// parsing the endpoint is taken care of in k8senvelopekmsv2.NewGRPCService below
+	if o.kmsSocketEndpoint == "" {
+		return fmt.Errorf("--kms-socket-endpoint is required")
+	}
```
p0lyn0mial, line 52: let's also check the format of the endpoint.

please have a look at https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/kms/health/cmd.go#L96

pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -39,21 +38,27 @@ func NewCommand(ctx context.Context) *cobra.Command {
 
 func (o *options) addFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.kmsCallTimeout, "kms-call-timeout", 0, "timeout for each gRPC call to the KMS plugin")
+	fs.StringVarP(&o.kmsSocketEndpoint, "endpoint", "e", "", "kms socket endpoint of KMS plugin")
```
p0lyn0mial, line 41: `kms-socket`

p0lyn0mial, line 41: nit: let's drop `"e"`

pkg/operator/encryption/kms/preflight/cmd.go
tjungblu: mind if I copy&paste the regex? I wouldn't want to introduce a dependency to the health package

pkg/operator/encryption/kms/preflight/cmd.go
p0lyn0mial, line 41: do we need a slice - there might be more than one plugin. not sure though.

tjungblu: changed

tjungblu: done

tjungblu: good point, when we have old and new config side-by-side: would we need to check both? then we can also just copy the entire command or run the command twice?

tjungblu:
/close

wasn't actually needed, we use emptyDir everywhere to mount

PR #2311, opened 2026-06-17
NO-JIRA: Using image as env variable

gangwgr:
Using image as env variable

## Summary by CodeRabbit

* **Tests**
  * Updated Vault KMS test setup to resolve the Vault KMS plugin image at runtime.
  * If a shared directory setting is available, the image is loaded from the shared location with whitespace trimmed, and empty/unavailable values are treated as not set.
  * Ensures consistent test behavior when no valid image is found.

### DIFF
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -22,6 +23,34 @@ import (
 	library "github.com/openshift/library-go/test/library/encryption"
 )
 
+// resolveVaultKMSPluginImage determines the vault-kube-kms plugin image to use.
+// It checks SHARED_DIR because the openshift-e2e-test step ref is a widely-used
+// shared ref that does not declare VAULT_KMS_PLUGIN_IMAGE in its env list.
+// The vault-install step writes the image reference to a file in SHARED_DIR,
+// allowing subsequent steps to pick it up without modifying the shared ref.
+func resolveVaultKMSPluginImage(t testing.TB) string {
+	t.Helper()
+	if img := os.Getenv("VAULT_KMS_PLUGIN_IMAGE"); img != "" {
+		t.Logf("Using vault KMS plugin image from VAULT_KMS_PLUGIN_IMAGE env: %s", img)
+		return img
+	}
+	sharedDir := os.Getenv("SHARED_DIR")
+	if sharedDir == "" {
+		t.Fatal("SHARED_DIR environment variable is not set; cannot resolve vault KMS plugin image")
+	}
+	imagePath := sharedDir + "/vault-kms-plugin-image"
+	data, err := os.ReadFile(imagePath)
+	if err != nil {
+		t.Fatalf("failed to read vault KMS plugin image from %s: %v", imagePath, err)
+	}
+	img := strings.TrimSpace(string(data))
+	if img == "" {
+		t.Fatalf("vault KMS plugin image file %s is empty", imagePath)
+	}
+	t.Logf("Resolved vault KMS plugin image from %s: %s", imagePath, img)
+	return img
+}
+
 const (
 	defaultVaultNamespace          = "vault-kms"
 	defaultVaultServiceName        = "vault"
@@ -30,7 +59,6 @@ const (
 	defaultVaultAppRoleSecretName  = "vault-approle-secret"
 	defaultVaultConfigMapName      = "vault-ca-bundle"
 	defaultFAKEVaultKMSPluginImage = "quay.io/openshifttest/mock-kms-plugin@sha256:958a2f8276037468aa47dc2137d3c30dfcd96489455eddb2fe655f8168a57622"
-	defaultVaultKMSPluginImage     = "registry.ci.openshift.org/control-plane-custom-builds/vault-kube-kms@sha256:33599dd6eee61dcf9a60138759fafda3d88593a3c2072585156882c6b5bd3fa5"
 	defaultVaultAddress            = "https://vault.vault-kms.svc:8200"
 	defaultVaultEnterpriseNS       = "admin"
 	defaultVaultTransitMount       = "transit"
@@ -53,6 +81,7 @@ const (
 // and bundles the AppRole secret setup.
 func DefaultVaultEncryptionProvider(ctx context.Context, t testing.TB) library.EncryptionProvider {
 	cfg := DefaultVaultKMSPluginConfig
+	cfg.KMS.Vault.KMSPluginImage = resolveVaultKMSPluginImage(t)
 	// Use the Service ClusterIP instead of DNS name because kube-apiserver pods
 	// cannot resolve cluster-local Service names (they use host network DNS).
 	cfg.KMS.Vault.VaultAddress = getVaultServiceAddress(ctx, t, defaultVaultNamespace, defaultVaultServiceName)
@@ -74,7 +103,6 @@ var DefaultVaultKMSPluginConfig = configv1.APIServerEncryption{
 	KMS: configv1.KMSPluginConfig{
 		Type: configv1.VaultKMSProvider,
 		Vault: configv1.VaultKMSPluginConfig{
-			KMSPluginImage: defaultVaultKMSPluginImage,
 			VaultAddress:   defaultVaultAddress,
 			VaultNamespace: defaultVaultEnterpriseNS,
 			TransitMount:   defaultVaultTransitMount,
@@ -125,7 +153,6 @@ var SecondaryVaultKMSPluginConfig = configv1.APIServerEncryption{
 	KMS: configv1.KMSPluginConfig{
 		Type: configv1.VaultKMSProvider,
 		Vault: configv1.VaultKMSPluginConfig{
-			KMSPluginImage: defaultVaultKMSPluginImage,
 			VaultAddress:   secondaryVaultAddress,
 			VaultNamespace: defaultVaultEnterpriseNS,
 			TransitMount:   defaultVaultTransitMount,
@@ -150,6 +177,7 @@ var SecondaryVaultKMSPluginConfig = configv1.APIServerEncryption{
 // for the secondary Vault instance, used in KMS-to-KMS migration e2e tests.
 func SecondaryVaultEncryptionProvider(ctx context.Context, t testing.TB) library.EncryptionProvider {
 	cfg := SecondaryVaultKMSPluginConfig
+	cfg.KMS.Vault.KMSPluginImage = resolveVaultKMSPluginImage(t)
 	cfg.KMS.Vault.VaultAddress = getVaultServiceAddress(ctx, t, secondaryVaultNamespace, secondaryVaultServiceName)
 	return library.EncryptionProvider{
 		APIServerEncryption: cfg,


gangwgr:
/hold
testing in ci

gangwgr:
test run - https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2202/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-aws-operator-encryption-kms/2067279587556986880

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,22 @@ import (
 	library "github.com/openshift/library-go/test/library/encryption"
 )
 
+var defaultVaultKMSPluginImage = resolveVaultKMSPluginImage()
+
+func resolveVaultKMSPluginImage() string {
+	if img := os.Getenv("VAULT_KMS_PLUGIN_IMAGE"); img != "" {
+		return img
+	}
+	if sharedDir := os.Getenv("SHARED_DIR"); sharedDir != "" {
```
ardaguclu: Why we are trying to get image from this dir?. Can't we have one place to get?. How this extraction will work on metal instances that we mirror image?

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,22 @@ import (
 	library "github.com/openshift/library-go/test/library/encryption"
 )
 
+var defaultVaultKMSPluginImage = resolveVaultKMSPluginImage()
+
+func resolveVaultKMSPluginImage() string {
+	if img := os.Getenv("VAULT_KMS_PLUGIN_IMAGE"); img != "" {
```
ardaguclu, line 33: Why don't we have only this?

gangwgr: we can't pass env in openshift-e2e-test it is widely used in release repo, not want to change anything in that step. This avoids modifying the widely-used ref and uses an established CI pattern.

test/library/encryption/kms/vault.go
gangwgr: For disconnected/metal, the image reference stays the same — ImageDigestMirrorSet on the cluster handles redirecting pulls to the local mirror transparently. The extraction logic doesn't need to change.
reason to use https://github.com/openshift/library-go/pull/2311#discussion_r3433135064

test/library/encryption/kms/vault.go
ardaguclu: So why don't we always extract it SHARED_DIR?, who is using VAULT_KMS_PLUGIN_IMAGE?

gangwgr: I added for safe check, removed VAULT_KMS_PLUGIN_IMAGE var

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,19 @@ import (
 	library "github.com/openshift/library-go/test/library/encryption"
 )
 
+var defaultVaultKMSPluginImage = resolveVaultKMSPluginImage()
+
+func resolveVaultKMSPluginImage() string {
+	if sharedDir := os.Getenv("SHARED_DIR"); sharedDir != "" {
```
ardaguclu: Can you please add documentation about why we read from SHARED_DIR?

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,19 @@ import (
...
+		if data, err := os.ReadFile(sharedDir + "/vault-kms-plugin-image"); err == nil {
```
ardaguclu: Does all the jobs write image to this file?

gangwgr: yes all our KMS jobs

ardaguclu: Makes sense. Did we test the changes here?. Otherwise, we'll break all the jobs?

ardaguclu: Is there any similar usage in openshift code base?. I'm wondering this is the generic approach to overcome the issue

gangwgr: > Makes sense. Did we test the changes here?. Otherwise, we'll break all the jobs?

https://github.com/openshift/library-go/pull/2311#issuecomment-4738094397

gangwgr: > Is there any similar usage in openshift code base?. I'm wondering this is the generic approach to overcome the issue

As checked with @chaibot 
```
Yes — using SHARED_DIR to pass data between CI test steps is a very common and well-established pattern across the OpenShift codebase. Here are real examples:

---

1. Image-Based Upgrade — passing image pullspec and version

bash
# Writing step (seed-create):
echo "${target_version}" > "${SHARED_DIR}/target_version"
echo "${target_image}" > "${SHARED_DIR}/target_image"

# Reading step (target):
TARGET_VERSION=$(cat ${SHARED_DIR}/target_version)
TARGET_IMAGE=$(cat ${SHARED_DIR}/target_image)




2. Upgrade tests — passing release image overrides

yaml
# set-releases-release step writes override images:
commands: echo "${RELEASE_IMAGE_INTERMEDIATE},${RELEASE_IMAGE_LATEST}" > ${SHARED_DIR}/override-upgrade
Downstream steps check for this file and use its content to override default TARGET_RELEASES.

3. HyperShift — passing registry config with image references

bash
cat >>"$SHARED_DIR"/image_registry.ini <<EOF
route: $REG_ROUTE
```
src_image: $SRC_IMG
dst_image: $DST_IMG
EOF




4. MCE/Assisted — extracting config from container images

The mce-conf-os-images step moves rhcos.json from the assisted-services base image to SHARED_DIR for subsequent steps.

5. Defensive file-exists checks (best practice)

bash
if [ -f "${SHARED_DIR}/image-registry-pvc.yaml" ]; then
  # use the file
else
  echo "File does not exist, skipping."
fi

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,19 @@ import (
...
+			if img := strings.TrimSpace(string(data)); img != "" {
+				return img
+			}
+		}
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
ardaguclu: @gangwgr  Can we also apply the suggestions here?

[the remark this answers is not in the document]
gangwgr: it is fixed

gangwgr:
testing one more time https://github.com/openshift/cluster-kube-apiserver-operator/pull/2202 in fake bump

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,28 @@ import (
...
 )
 
+var defaultVaultKMSPluginImage = resolveVaultKMSPluginImage()
+
+// resolveVaultKMSPluginImage determines the vault-kube-kms plugin image to use.
+// It checks SHARED_DIR because the openshift-e2e-test step ref is a widely-used
+// shared ref that does not declare VAULT_KMS_PLUGIN_IMAGE in its env list.
+// The vault-install step writes the image reference to a file in SHARED_DIR,
+// allowing subsequent steps to pick it up without modifying the shared ref.
+func resolveVaultKMSPluginImage() string {
```
ardaguclu: We need non-empty image for all tests. So we should return error whenever we get anything that may lead to empty image. Otherwise, there is no point of moving foward.

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,31 @@ import (
...
+// resolveVaultKMSPluginImage determines the vault-kube-kms plugin image to use.
+// It checks SHARED_DIR because the openshift-e2e-test step ref is a widely-used
+// shared ref that does not declare VAULT_KMS_PLUGIN_IMAGE in its env list.
+// The vault-install step writes the image reference to a file in SHARED_DIR,
+// allowing subsequent steps to pick it up without modifying the shared ref.
+// Panics if the image cannot be resolved — all KMS tests require a valid image.
+func resolveVaultKMSPluginImage() string {
+	sharedDir := os.Getenv("SHARED_DIR")
+	if sharedDir == "" {
+		panic("SHARED_DIR environment variable is not set; cannot resolve vault KMS plugin image")
```
ardaguclu: return err? and the caller will call t.Fatal?

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,31 @@ import (
...
+	}
+	imagePath := sharedDir + "/vault-kms-plugin-image"
+	data, err := os.ReadFile(imagePath)
+	if err != nil {
+		panic(fmt.Sprintf("failed to read vault KMS plugin image from %s: %v", imagePath, err))
```
ardaguclu: same

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,31 @@ import (
...
+	}
+	img := strings.TrimSpace(string(data))
+	if img == "" {
+		panic(fmt.Sprintf("vault KMS plugin image file %s is empty", imagePath))
```
ardaguclu: same

test/library/encryption/kms/vault.go
```diff
@@ -52,6 +76,7 @@ const (
 // It resolves the Vault Service ClusterIP at call time to avoid DNS resolution issues,
 // and bundles the AppRole secret setup.
 func DefaultVaultEncryptionProvider(ctx context.Context, t testing.TB) library.EncryptionProvider {
+	require.NoError(t, defaultVaultKMSPluginImageErr, "failed to resolve vault KMS plugin image")
```
ardaguclu: Can we call resolveVaultKMSPluginImage() here, so that we don't need `var defaultVaultKMSPluginImage, defaultVaultKMSPluginImageErr = resolveVaultKMSPluginImage()
`

ardaguclu:
/lgtm
/hold
Feel free to unhold when you see CI green

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,29 @@ import (
 	library "github.com/openshift/library-go/test/library/encryption"
 )
 
+// resolveVaultKMSPluginImage determines the vault-kube-kms plugin image to use.
```
tjungblu, line 26: can we still have an env variable for running this locally? it's a bit awkward to create a folder with a file in it

test/library/encryption/kms/vault.go
```diff
@@ -22,6 +23,29 @@ import (
...
+	imagePath := sharedDir + "/vault-kms-plugin-image"
+	data, err := os.ReadFile(imagePath)
+	if err != nil {
+		t.Fatalf("failed to read vault KMS plugin image from %s: %v", imagePath, err)
+	}
+	img := strings.TrimSpace(string(data))
+	if img == "" {
+		t.Fatalf("vault KMS plugin image file %s is empty", imagePath)
+	}
+	return img
```
tjungblu, line 51: do you mind to add a `t.Log`, so we can see what it resolves to at runtime?

test/library/encryption/kms/vault.go
ardaguclu: I asked to remove env variable. If we run these tests on local, it makes sense to add it back.

gangwgr: @tjungblu is it ok set env on local we can add this for local
```
	if img := os.Getenv("VAULT_KMS_PLUGIN_IMAGE"); img != "" {
		t.Logf("Using vault KMS plugin image from VAULT_KMS_PLUGIN_IMAGE env: %s", img)
		return img
	}
```

tjungblu: yeah I frequently run those from local as well, would be glad to just have the option to fallback to it.

gangwgr: added

gangwgr:
/unhold
As @tjungblu testing in rebase

PR #2313, opened 2026-06-17
CNTRLPLANE-3237: KMS preflight checker part3

p0lyn0mial:
## Summary by CodeRabbit

* **Refactor**
  * Improved KMS provider configuration handling to consistently retrieve provider-specific source settings through a shared abstraction.
* **New Features**
  * Added deterministic hashing for KMS preflight checks, combining the provider configuration with explicitly referenced Secret/ConfigMap data using stable ordering for consistent change detection.
* **Tests**
  * Expanded automated coverage to verify deterministic hash outputs across configuration variations and to confirm clear errors when referenced resources or required keys are missing.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -2,8 +2,16 @@ package controllers
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
+	"hash"
+	"hash/fnv"
+	"sort"
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -14,6 +22,101 @@ import (
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
+
+type kmsConfigHasher struct {
+	provider   kmsProviderConfig
+	coreClient corev1client.CoreV1Interface
+	namespace  string
+}
+
+// newKMSConfigHasher creates a hasher for a KMS provider config and its referenced resources.
+// namespace is the namespace where the referenced Secrets and ConfigMaps are stored (e.g., openshift-config).
+func newKMSConfigHasher(provider kmsProviderConfig, coreClient corev1client.CoreV1Interface, namespace string) *kmsConfigHasher {
+	return &kmsConfigHasher{provider: provider, coreClient: coreClient, namespace: namespace}
+}
+
+// hash computes a deterministic hash over the provider config and the specific data keys
+// from its referenced Secret and ConfigMap. Uses FNV-32, JSON encoding, and base64 URL
+// encoding, consistent with resourcehash.GetSecretHash and resourcehash.GetConfigMapHash.
+func (h *kmsConfigHasher) hash(ctx context.Context) (string, error) {
+	hasher := fnv.New32()
+
+	if err := json.NewEncoder(hasher).Encode(h.provider.sourceConfig()); err != nil {
+		return "", fmt.Errorf("failed to hash provider config: %w", err)
+	}
+
+	if err := h.hashReferencedSecret(ctx, hasher); err != nil {
+		return "", err
+	}
+	if err := h.hashReferencedConfigMap(ctx, hasher); err != nil {
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+func (h *kmsConfigHasher) hashReferencedSecret(ctx context.Context, hasher hash.Hash) error {
+	name, keys, err := h.provider.referencedSecretName()
+	if err != nil {
+		return fmt.Errorf("failed to get referenced secret name: %w", err)
+	}
+	if name == "" {
+		return nil
+	}
+
+	secret, err := h.coreClient.Secrets(h.namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s/%s: %w", h.namespace, name, err)
+	}
+
+	// Write each key name before its value to prevent collisions when bytes
+	// shift between adjacent values (e.g. role-id="ab",secret-id="cd" vs
+	// role-id="abc",secret-id="d" would otherwise both hash as "abcd").
+	sort.Strings(keys)
+	for _, k := range keys {
+		v, ok := secret.Data[k]
+		if !ok {
+			return fmt.Errorf("key %q not found in secret %s/%s", k, h.namespace, name)
+		}
+		if _, err := hasher.Write([]byte(k)); err != nil {
+			return fmt.Errorf("failed to hash key %q: %w", k, err)
+		}
+		if _, err := hasher.Write(v); err != nil {
+			return fmt.Errorf("failed to hash key %q: %w", k, err)
+		}
+	}
+	return nil
+}
+
+func (h *kmsConfigHasher) hashReferencedConfigMap(ctx context.Context, hasher hash.Hash) error {
+	name, keys, err := h.provider.referencedConfigMapName()
+	if err != nil {
+		return fmt.Errorf("failed to get referenced configmap name: %w", err)
+	}
+	if name == "" {
+		return nil
+	}
+
+	cm, err := h.coreClient.ConfigMaps(h.namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get configmap %s/%s: %w", h.namespace, name, err)
+	}
+
+	sort.Strings(keys)
+	for _, k := range keys {
+		v, ok := cm.Data[k]
+		if !ok {
+			return fmt.Errorf("key %q not found in configmap %s/%s", k, h.namespace, name)
+		}
+		if _, err := hasher.Write([]byte(k)); err != nil {
+			return fmt.Errorf("failed to hash key %q: %w", k, err)
+		}
+		if _, err := hasher.Write([]byte(v)); err != nil {
+			return fmt.Errorf("failed to hash key %q: %w", k, err)
+		}
+	}
+	return nil
+}
 
 type kmsPreflightController struct {
 	controllerInstanceName string
--- a/pkg/operator/encryption/controllers/kms_preflight_controller_test.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller_test.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"strings"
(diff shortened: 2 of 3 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
 
+type kmsConfigHasher struct {
```
p0lyn0mial, line 26: planning to add unit tests.

p0lyn0mial, line 26: this type would be also used by the key controller.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
...
+	}
+
+	if err := h.hashReferencedSecret(ctx, hasher); err != nil {
+		return "", err
+	}
+	if err := h.hashReferencedConfigMap(ctx, hasher); err != nil {
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
```
tjungblu, line 55: how long is that going to be? 32 bit of FNV as base64 is 6 characters?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
...
+	return &kmsConfigHasher{provider: provider, coreClient: coreClient, namespace: namespace}
+}
+
+// hash computes a deterministic hash over the provider config and the specific data keys
+// from its referenced Secret and ConfigMap. Uses FNV-32, JSON encoding, and base64 URL
+// encoding, consistent with resourcehash.GetSecretHash and resourcehash.GetConfigMapHash.
+func (h *kmsConfigHasher) hash(ctx context.Context) (string, error) {
+	hasher := fnv.New32()
+
+	if err := json.NewEncoder(hasher).Encode(h.provider.sourceConfig()); err != nil {
```
tjungblu, line 44: that's a nice trick I haven't thought of, nice!

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: looks good, let me know if you have the tests ready

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
...
+// newKMSConfigHasher creates a hasher for a KMS provider config and its referenced resources.
+// namespace is the namespace where the referenced Secrets and ConfigMaps are stored (e.g., openshift-config).
+func newKMSConfigHasher(provider kmsProviderConfig, coreClient corev1client.CoreV1Interface, namespace string) *kmsConfigHasher {
+	return &kmsConfigHasher{provider: provider, coreClient: coreClient, namespace: namespace}
+}
+
+// hash computes a deterministic hash over the provider config and the specific data keys
+// from its referenced Secret and ConfigMap. Uses FNV-32, JSON encoding, and base64 URL
+// encoding, consistent with resourcehash.GetSecretHash and resourcehash.GetConfigMapHash.
+func (h *kmsConfigHasher) hash(ctx context.Context) (string, error) {
```
ardaguclu, line 41: Shouldn't key controller calculate this hash in `sourceConfig`, so that it is able to check the results of preflight checker in order to create the key?. If we add hash calculation in preflight controller, how can key-controller query the results?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
...
+	}
+	if err := h.hashReferencedConfigMap(ctx, hasher); err != nil {
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil)), nil
+}
+
+func (h *kmsConfigHasher) hashReferencedSecret(ctx context.Context, hasher hash.Hash) error {
+	name, keys, err := h.provider.referencedSecretName()
```
ardaguclu, line 59: Key controller accesses all these already. Maybe we should unify the core logic to be shared by both of the controllers.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: yes, 6 chars + padding (check the unit tests)

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: tests added PTAL.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: the key controller will use `kmsConfigHasher` to calculate the hash. 

`kmsConfigHasher` will be shared between these two controllers.

does it make sense ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: `kmsConfigHasher` will be shared between these two controllers.
`kmsConfigHasher` uses `kmsProviderConfig` which is generic and provider agnostic.

does it make sense ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -15,6 +23,92 @@ import (
...
+
+	sort.Strings(keys)
+	for _, k := range keys {
+		v, ok := secret.Data[k]
+		if !ok {
+			return fmt.Errorf("key %q not found in secret %s/%s", k, h.namespace, name)
+		}
+		if _, err := hasher.Write(v); err != nil {
+			return fmt.Errorf("failed to hash key %q: %w", k, err)
+		}
```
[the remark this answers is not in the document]
p0lyn0mial: i am looking into this ...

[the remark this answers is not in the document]
tjungblu: I thought about this yesterday as well, but given this is a secret and it holds some sort of UUID, the chances of this happening is very slim. Especially since we're sorting the keys upfront.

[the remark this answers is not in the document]
tjungblu: nope, this was a response to you @coderabbitai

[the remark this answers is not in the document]
p0lyn0mial: @coderabbitai fixed. PTAL.

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
```diff
@@ -23,6 +24,281 @@ import (
...
+		vaultConfig   configv1.VaultKMSPluginConfig
+		resources     []runtime.Object
+		expectedHash  string
+		expectedError string
+	}{
+		{
+			name:         "same config and resources produce the same hash",
+			vaultConfig:  baseVaultConfig,
+			resources:    []runtime.Object{baseSecret, baseConfigMap},
+			expectedHash: "k6dSVA==",
```
tjungblu, line 70: this is a maintenance nitpick for me, but imagine we're changing the structure in the future and the hashes change. Maybe we don't want to hardcode the hashes on every test instead of a "match/mismatch" flag.

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: can we use this as a suffix in a deployment name? 
> DC20hA==

"==" wouldn't work. Maybe just hex the string instead?

Alternatively we can just generate a random name and have this as a label, that's also cool for me. Just adds another filtering step to the queries.

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
p0lyn0mial: in an AI era the cost of updating the hashes is very low. I did that many time while I was developing the tests.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: yes, it makes sense. Thank you

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
p0lyn0mial: also i wanted to avoid precomputing the hash in the test because i wanted to avoid using the same production code path or build a new function for the tests

p0lyn0mial: does it make sense ?

tjungblu: not really, how does the AI generate the hashes when the struct changes? Let's say we add a new provider in KMSPluginConfig. 

> avoid precomputing the hash in the test 

so how did you get the hashes in the first place?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: yeah, I think we should have a stable pod name. There will be a single instance of the preflight pod per namespace at any given time. It will make cleanup easier. 

We could add the hash as an annotation to keep track of the version/generation/configuration that the preflight pod is checking.

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
p0lyn0mial: > so how did you get the hashes in the first place?

directly from the `hash` method.

p0lyn0mial: > not really, how does the AI generate the hashes when the struct changes?

this is a similar technique to 

- https://github.com/openshift/library-go/blob/29d1794ce0aa8e9d45c944c8a399b387dfccdcad/pkg/manifestclienttest/client_write_test.go#L290
- https://github.com/openshift/openshift-apiserver/blob/main/pkg/bootstrappolicy/policy_test.go#L100
- https://github.com/openshift/machine-config-operator/blob/main/devex/cmd/mco-sanitize/sanitize_e2e_test.go#L150

just now you can ask ai to replace the hardcoded values.

p0lyn0mial: I added the following comment as a hint for agents:

```
// Each scenario changes exactly one field from the baseline and verifies the hash changes.
// If the hash algorithm or encoding changes, update the expectedHash values by running
// the test and copying the actual values from the error output.
```

ardaguclu:
LGTM

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
tjungblu: > just now you can ask ai to replace the hardcoded values.

yeah, but the agent does the following (try it out and check the commands):
* runs the tests
* input the failed actual values into the expected values

You might as well just save yourself the tokens and directly generate the expected values.

PR #2315, opened 2026-06-18
NO-JIRA: report preflight pod status in status condition

tjungblu:
Uses the downward API to post the status of the check in dedicated pod conditions.
This also adds the kekId that is important for the rotation later.

## Release Notes

* **New Features**
  * KMS preflight now publishes additional Pod readiness gate conditions for the preflight result, config hash, and detected KMS key ID.

* **Improvements**
  * The `kms-preflight` command now supports config-hash and pod identity inputs, polls plugin health before running checks, and records both checker and reporting outcomes.
  * Pod status updates for preflight conditions are now applied via a patch.

* **Tests**
  * Expanded unit tests to verify returned key details and Pod condition generation/update behavior.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/cmd.go
+++ b/pkg/operator/encryption/kms/preflight/cmd.go
@@ -2,6 +2,7 @@ package preflight
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -9,6 +10,8 @@ import (
 	"github.com/spf13/pflag"
 
 	k8senvelopekmsv2 "k8s.io/apiserver/pkg/storage/value/encrypt/envelope/kmsv2"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
@@ -16,6 +19,10 @@ const kmsSocketEndpoint = "unix:///var/run/kmsplugin/kms.sock"
 
 type options struct {
 	kmsCallTimeout time.Duration
+	podName        string
+	podNamespace   string
+	configHash     string
+	kubeconfig     string
 }
 
 // NewCommand creates the kms-preflight cobra command.
@@ -39,12 +46,26 @@ func NewCommand(ctx context.Context) *cobra.Command {
 
 func (o *options) addFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.kmsCallTimeout, "kms-call-timeout", 0, "timeout for each gRPC call to the KMS plugin")
+	fs.StringVar(&o.configHash, "config-hash", o.configHash, "hash of config to use for encryption")
+	fs.StringVar(&o.podName, "pod-name", o.podName, "name of pod to use to report checker status")
+	fs.StringVar(&o.podNamespace, "pod-namespace", o.podNamespace, "namespace of pod to report checker status")
+	fs.StringVar(&o.kubeconfig, "kubeconfig", o.kubeconfig, "path to a kubeconfig; empty uses in-cluster config")
 }
 
 func (o *options) validate() error {
 	if o.kmsCallTimeout <= 0 {
 		return fmt.Errorf("--kms-call-timeout must be greater than 0")
 	}
+	if o.configHash == "" {
+		return fmt.Errorf("--config-hash is required")
+	}
+	if o.podName == "" {
+		return fmt.Errorf("--pod-name is required")
+	}
+	if o.podNamespace == "" {
+		return fmt.Errorf("--pod-namespace is required")
+	}
+
 	return nil
 }
 
@@ -58,10 +79,24 @@ func (o *options) run(ctx context.Context) error {
 		return fmt.Errorf("failed to create KMS gRPC client: %w", err)
 	}
 
+	kubeConfig, err := clientcmd.BuildConfigFromFlags("", o.kubeconfig)
+	if err != nil {
+		return fmt.Errorf("failed to build kubeconfig: %w", err)
+	}
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create kube client: %w", err)
+	}
+
 	checker := newChecker(service)
 	start := time.Now()
-	if err = checker.check(ctx); err != nil {
-		return fmt.Errorf("kms preflight check failed: %w", err)
+	status, checkErr := checker.check(ctx)
+
+	podClient := kubeClient.CoreV1().Pods(o.podNamespace)
+	reportErr := setPodCheckCondition(ctx, podClient, o.podName, o.configHash, status, checkErr)
+	// join the errors to not lose the original error message
+	if err := errors.Join(checkErr, reportErr); err != nil {
+		return err
 	}
 
 	klog.Infof("KMS preflight check passed, total latency=%v", time.Since(start))
--- a/pkg/operator/encryption/kms/preflight/pod_status.go
+++ b/pkg/operator/encryption/kms/preflight/pod_status.go
@@ -1,0 +1,83 @@
+package preflight
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/controllers"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
+	kmsservice "k8s.io/kms/pkg/service"
+)
+
+func setPodCheckCondition(ctx context.Context, podClient corev1client.PodInterface, podName string, configHash string,
+	status *kmsservice.StatusResponse, checkErr error) error {
+	conditions := podCheckConditions(configHash, status, checkErr)
+	return updatePodCheckConditions(ctx, podClient, podName, conditions)
+}
+
+func updatePodCheckConditions(ctx context.Context, podClient corev1client.PodInterface, name string, conditions []corev1.PodCondition) error {
+	err := retry.OnError(retry.DefaultRetry, func(err error) bool { return true }, func() error {
+		pod, err := podClient.Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		for _, newCond := range conditions {
+			found := false
+			for i, existingCond := range pod.Status.Conditions {
+				if existingCond.Type == newCond.Type {
+					pod.Status.Conditions[i] = newCond
+					found = true
+					break
+				}
+			}
+			if !found {
+				pod.Status.Conditions = append(pod.Status.Conditions, newCond)
+			}
+		}
+		_, err = podClient.UpdateStatus(ctx, pod, metav1.UpdateOptions{})
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("failed to update pod status for %s: %w", name, err)
+	}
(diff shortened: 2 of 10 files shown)


pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
+
+	"github.com/openshift/library-go/pkg/config/client"
+	kmsservice "k8s.io/kms/pkg/service"
+)
+
+func setPodCheckCondition(ctx context.Context, configHash string, podNamespace string, podName string,
+	status *kmsservice.StatusResponse, checkErr error) error {
+	kubeConfig, err := client.GetKubeConfigOrInClusterConfig("", nil)
```
ardaguclu: I think, for static pod, we need to accept kubeconfig as option like this https://github.com/openshift/library-go/blob/257053230f0be3e7325f38dc00d98147c7f77e16/pkg/operator/encryption/kms/health/cmd.go#L87

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func applyPodCheckConditions(pod *corev1.Pod, conditions []corev1.PodCondition) {
+	for _, condition := range conditions {
+		upsertPodCondition(&pod.Status.Conditions, condition)
+	}
+}
+
+func updatePodCheckConditions(ctx context.Context, podClient corev1client.PodInterface, name string, conditions []corev1.PodCondition) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
ardaguclu: How can we get conflict error?

pkg/operator/encryption/kms/preflight/pod_status.go
ardaguclu: Then pass something like this https://github.com/openshift/library-go/blob/257053230f0be3e7325f38dc00d98147c7f77e16/pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go#L16

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+	return updatePodCheckConditions(ctx, podClient, podName, podCheckConditions(configHash, status, checkErr))
+}
+
+func applyPodCheckConditions(pod *corev1.Pod, conditions []corev1.PodCondition) {
+	for _, condition := range conditions {
+		upsertPodCondition(&pod.Status.Conditions, condition)
+	}
+}
+
+func updatePodCheckConditions(ctx context.Context, podClient corev1client.PodInterface, name string, conditions []corev1.PodCondition) error {
```
p0lyn0mial, line 21: shouldn't we also specify the conditions in the pod' spec ?

xref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate

`To use this, set readinessGates in the Pod's spec to specify a list of additional conditions that the kubelet evaluates for Pod readiness.....`

pkg/operator/encryption/kms/preflight/pod_status.go
p0lyn0mial: also should we create the client in cmd.go ?

pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -39,12 +43,25 @@ func NewCommand(ctx context.Context) *cobra.Command {
 
 func (o *options) addFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.kmsCallTimeout, "kms-call-timeout", 0, "timeout for each gRPC call to the KMS plugin")
+	fs.StringVarP(&o.configHash, "config-hash", "", "", "hash of config to use for encryption")
```
p0lyn0mial: sine we don't provide short flags use `StringVar` instead .

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+
+	conditions = append(conditions,
+		corev1.PodCondition{
+			Type:               controllers.PodConditionKMSPreflightKeyID,
+			Status:             corev1.ConditionTrue,
+			Message:            status.KeyID,
+			LastTransitionTime: now,
+		},
+		corev1.PodCondition{
+			Type:               controllers.PodConditionKMSPreflightConfigHash,
```
p0lyn0mial: does this condition depend on the status ? if not it seems we should always set it.

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create kube client: %w", err)
+	}
+
+	podClient := kubeClient.CoreV1().Pods(podNamespace)
+	return updatePodCheckConditions(ctx, podClient, podName, podCheckConditions(configHash, status, checkErr))
+}
+
+func applyPodCheckConditions(pod *corev1.Pod, conditions []corev1.PodCondition) {
```
p0lyn0mial: nit: could we inline this function ? it is short and called only from one place.

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func updatePodCheckConditions(ctx context.Context, podClient corev1client.PodInterface, name string, conditions []corev1.PodCondition) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		pod, err := podClient.Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		applyPodCheckConditions(pod, conditions)
```
p0lyn0mial: since this pod runs only once could this be simplified to:

```
pod, err := podClient.Get(ctx, name, metav1.GetOptions{})
		if err != nil {
			return err
		}
		pod.Status.Conditions = append(pod.Status.Conditions, conditions...)
		_, err = podClient.UpdateStatus(ctx, pod, metav1.UpdateOptions{})
		return err
```

?

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,96 @@
...
+		_, err = podClient.UpdateStatus(ctx, pod, metav1.UpdateOptions{})
+		return err
+	})
+}
+
+func podCheckConditions(configHash string, status *kmsservice.StatusResponse, checkErr error) []corev1.PodCondition {
```
p0lyn0mial, line 49: I think this could be:
```
func podCheckConditions(configHash string, status *kmsservice.StatusResponse, checkErr error) []corev1.PodCondition {
	now := metav1.Now()

	checkStatus, checkReason, checkMessage := corev1.ConditionTrue, "Succeeded", ""
	if checkErr != nil {
		checkStatus, checkReason, checkMessage = corev1.ConditionFalse, "Failed", checkErr.Error()
	}

	conditions := []corev1.PodCondition{
		{
			Type:               controllers.PodConditionKMSPreflightResult,
			Status:             checkStatus,
			Reason:             checkReason,
			Message:            checkMessage,
			LastTransitionTime: now,
		},
		{
			Type:               controllers.PodConditionKMSPreflightConfigHash,
			Status:             corev1.ConditionTrue,
			Message:            configHash,
			LastTransitionTime: now,
		},
	}

	if status != nil {
		conditions = append(conditions, corev1.PodCondition{
			Type:               controllers.PodConditionKMSPreflightKeyID,
			Status:             corev1.ConditionTrue,
			Message:            status.KeyID,
			LastTransitionTime: now,
		})
	}

	return conditions
}
```

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,65 @@
...
+	return updatePodCheckConditions(ctx, podClient, podName, podCheckConditions(configHash, status, checkErr))
+}
+
+func updatePodCheckConditions(ctx context.Context, podClient corev1client.PodInterface, name string, conditions []corev1.PodCondition) error {
+	pod, err := podClient.Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get pod %s: %w", name, err)
+	}
+	pod.Status.Conditions = append(pod.Status.Conditions, conditions...)
+	if _, err = podClient.UpdateStatus(ctx, pod, metav1.UpdateOptions{}); err != nil {
```
p0lyn0mial: ah, also, shouldn't we use PATCH here ?

pkg/operator/encryption/kms/preflight/pod_status.go
```diff
@@ -0,0 +1,71 @@
...
+	patch := corev1.Pod{
+		Status: corev1.PodStatus{
+			Conditions: conditions,
+		},
+	}
+	patchBytes, err := json.Marshal(patch)
+	if err != nil {
+		return fmt.Errorf("failed to marshal pod status patch: %w", err)
+	}
+	if _, err = podClient.Patch(ctx, name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{}, "status"); err != nil {
```
ardaguclu: I think this is the correct action without overriding the other statuses.

ardaguclu:
LGTM

pkg/operator/encryption/kms/preflight/pod_status.go
tjungblu: updated

pkg/operator/encryption/kms/preflight/pod_status.go
tjungblu: always set

pkg/operator/encryption/kms/preflight/pod_status.go
tjungblu: not needed to update the pod status

pkg/operator/encryption/kms/preflight/cmd.go
```diff
@@ -39,12 +46,26 @@ func NewCommand(ctx context.Context) *cobra.Command {
 
 func (o *options) addFlags(fs *pflag.FlagSet) {
 	fs.DurationVar(&o.kmsCallTimeout, "kms-call-timeout", 0, "timeout for each gRPC call to the KMS plugin")
+	fs.StringVar(&o.configHash, "config-hash", "", "hash of config to use for encryption")
```
ardaguclu:
```suggestion
	fs.StringVar(&o.configHash, "config-hash", o.configHash, "hash of config to use for encryption")
```
same for the others

tjungblu: applied

pkg/operator/encryption/kms/preflight/pod_status_test.go
```diff
@@ -0,0 +1,123 @@
...
+	updated, err := podClient.Get(context.Background(), pod.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Get() error = %v", err)
+	}
+	if len(updated.Status.Conditions) != 3 {
+		t.Fatalf("expected 3 conditions, got %d", len(updated.Status.Conditions))
+	}
+}
+
+func TestUpdatePodCheckConditions_retriesOnConflict(t *testing.T) {
```
tjungblu, line 93: that's a bit specific, we retry everything now  - happy to change it to something else if it is too confusing

PR #2316, opened 2026-06-18
kms: add readiness check hook to revision controller

bertinatto:
## Summary by CodeRabbit

* **New Features**
  * Added an optional “revision readiness” callback during revision creation.
  * The callback runs after the revision ConfigMap is created and before the revision is marked ready; it can prevent readiness on error or return value `false` (default: always ready).
  * Exposed this callback in the static pod controller builder.
* **Bug Fixes**
  * Updated controller wiring to support the expanded initialization.
* **Tests**
  * Added coverage for readiness success and failure, including latest-available revision updates and the revision-ready annotation behavior.

### DIFF
--- a/pkg/operator/revisioncontroller/revision_controller.go
+++ b/pkg/operator/revisioncontroller/revision_controller.go
@@ -39,7 +39,8 @@ type RevisionController struct {
 	configMapGetter corev1client.ConfigMapsGetter
 	secretGetter    corev1client.SecretsGetter
 
-	revisionPrecondition PreconditionFunc
+	revisionPrecondition   PreconditionFunc
+	revisionReadinessCheck ReadinessCheckFunc
 }
 
 type RevisionResource struct {
@@ -49,6 +50,9 @@ type RevisionResource struct {
 
 // PreconditionFunc checks if revision precondition is met (is true) and then proceeeds with the creation of new revision
 type PreconditionFunc func(ctx context.Context) (bool, error)
+
+// ReadinessCheckFunc validates the assembled revision data after all resources have been copied but before the revision is marked as ready.
+type ReadinessCheckFunc func(ctx context.Context, revision int32) (bool, error)
 
 // NewRevisionController create a new revision controller.
 func NewRevisionController(
@@ -62,9 +66,15 @@ func NewRevisionController(
 	secretGetter corev1client.SecretsGetter,
 	eventRecorder events.Recorder,
 	revisionPrecondition PreconditionFunc,
+	revisionReadinessCheck ReadinessCheckFunc,
 ) factory.Controller {
 	if revisionPrecondition == nil {
 		revisionPrecondition = func(ctx context.Context) (bool, error) {
+			return true, nil
+		}
+	}
+	if revisionReadinessCheck == nil {
+		revisionReadinessCheck = func(_ context.Context, _ int32) (bool, error) {
 			return true, nil
 		}
 	}
@@ -75,10 +85,11 @@ func NewRevisionController(
 		configMaps:             configMaps,
 		secrets:                secrets,
 
-		operatorClient:       operatorClient,
-		configMapGetter:      configMapGetter,
-		secretGetter:         secretGetter,
-		revisionPrecondition: revisionPrecondition,
+		operatorClient:         operatorClient,
+		configMapGetter:        configMapGetter,
+		secretGetter:           secretGetter,
+		revisionPrecondition:   revisionPrecondition,
+		revisionReadinessCheck: revisionReadinessCheck,
 	}
 
 	return factory.New().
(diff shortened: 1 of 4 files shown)


bertinatto:
Proof PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2121

ardaguclu:
This PR looks good to me. But I think we need review from @p0lyn0mial

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -50,6 +51,9 @@ type RevisionResource struct {
 // PreconditionFunc checks if revision precondition is met (is true) and then proceeeds with the creation of new revision
 type PreconditionFunc func(ctx context.Context) (bool, error)
 
+// PostconditionFunc validates the assembled revision data after all resources have been copied but before the revision is marked as ready.
+type PostconditionFunc func(ctx context.Context, revision int32) (bool, error)
+
```
bertinatto, line 56: I just realized that `Postcondition` might not be the best name because the check is done before the operator conditions are set, right before the revision is marked as ready, not after the condition is set. Maybe `RevisionReadinessCheck` is a better name?

bertinatto: I went ahead and renamed it

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -62,23 +66,30 @@ func NewRevisionController(
...
 	revisionPrecondition PreconditionFunc,
+	revisionReadinessCheck ReadinessCheckFunc,
 ) factory.Controller {
 	if revisionPrecondition == nil {
 		revisionPrecondition = func(ctx context.Context) (bool, error) {
 			return true, nil
 		}
 	}
+	if revisionReadinessCheck == nil {
+		revisionReadinessCheck = func(_ context.Context, _ int32) (bool, error) {
```
ardaguclu, line 77: Instead of this mock function, why don't we pass nil and invoke the function only when it is not nil?

bertinatto: I prefer that approach too (initially implemented it like that), but I changed it to keep it consistent with the precondition check from line 71

ardaguclu:
LGTM

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -62,23 +66,30 @@ func NewRevisionController(
...
 
-		operatorClient:       operatorClient,
-		configMapGetter:      configMapGetter,
-		secretGetter:         secretGetter,
-		revisionPrecondition: revisionPrecondition,
+		operatorClient:         operatorClient,
+		configMapGetter:        configMapGetter,
+		secretGetter:           secretGetter,
+		revisionPrecondition:   revisionPrecondition,
+		revisionReadinessCheck: revisionReadinessCheck,
```
p0lyn0mial, line 92: could we rename `revisionReadinessCheck` to `revisionPostcondition` so that it is symmetric with `revisionPrecondition` ?
(applies to other places in the PR)

pkg/operator/revisioncontroller/revision_controller.go
```diff
@@ -309,6 +320,15 @@ func (c RevisionController) createNewRevision(ctx context.Context, recorder even
 		}
 	}
 
+	// We run the revision readiness check function here because at this point in time, the revision
+	// ConfigMap is already created, but it doesn't yet have the "revision-ready: true" annotation.
+	// This means that the installer pod has not yet started deploying the revision.
+	if ready, err := c.revisionReadinessCheck(ctx, revision); err != nil {
+		return false, fmt.Errorf("revision readiness check failed for revision %d: %w", revision, err)
```
p0lyn0mial, line 333: let's add a comment explaining why it is safe to fail the revision.
let's add a comment explaining how the platform converges.

pkg/operator/revisioncontroller/revision_controller.go
bertinatto: In my last push, I renamed it from `revisionPostcondition` to `revisionRedinessCheck` :) 

IMO "postcondition" is wrong because technically the check is done **before** the condition is set, not after. 

Rougly, this is the timeline:

1. pre condition check
2. create the revision artifacts
3. mark the revision artifacts as `ready`
4. update the operator status with the new condition

The check we're introducing in this PR is done between steps 2 and 3. It feels like `revisionPostcondition` would be something after step 3 or 4. WDYT?

bertinatto: xref: https://github.com/openshift/library-go/pull/2316#discussion_r3454986435

pkg/operator/revisioncontroller/revision_controller.go
bertinatto: added

pkg/operator/revisioncontroller/revision_controller.go
p0lyn0mial: ok, thanks, it make sense.

PR #2317, opened 2026-06-19
CNTRLPLANE-3363: Inject health reporter as sidecar container

ardaguclu:
This PR injects health reporter container into apiserver pods, if any kms plugin container is added. 

## Summary by CodeRabbit

* **New Features**
  * Added health reporting support for KMS plugin pods. A health-reporter sidecar container can now be configured and injected into KMS plugin deployments to enable health monitoring and reporting capabilities.

* **Tests**
  * Added comprehensive test coverage for health reporter injection and configuration in KMS plugin pod specs.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/builder.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/builder.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/health"
 )
 
 // KMSPluginBuilder constructs KMS plugin pod spec contributions for injection
@@ -20,6 +21,8 @@ type KMSPluginBuilder struct {
 	encryptionConfig           *encryptiondata.Config
 	encryptionConfigSecretName string
 	staticPod                  bool
+
+	healthReporter *healthReporter
 }
 
 // NewKMSPluginBuilder creates a builder that defaults to deployment mode.
@@ -41,6 +44,19 @@ func (b *KMSPluginBuilder) FromEncryptionConfig(encryptionConfigSecretName strin
 // data from the resource-dir volume and run as root (UID 0).
 func (b *KMSPluginBuilder) AsStaticPod() *KMSPluginBuilder {
 	b.staticPod = true
+	return b
+}
+
+// WithHealthReporter enables injection of a health-reporter sidecar.
+// operatorBinary is the parent binary (e.g. "cluster-kube-apiserver-operator");
+// when empty, the subcommand is invoked directly via the image ENTRYPOINT.
+func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage string) *KMSPluginBuilder {
+	b.healthReporter = &healthReporter{
+		name:           "kms-health-reporter",
+		operatorBinary: operatorBinary,
+		subcommand:     health.Subcommand,
+		image:          operatorImage,
+	}
 	return b
 }
 
@@ -83,9 +99,11 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string) 
 	socketVolumeMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: false}
 	refDataVolumeMount := corev1.VolumeMount{Name: refDataVolumeName, MountPath: refDataMountPath, ReadOnly: true}
 
+	sockets := make([]string, 0, len(kmsConfigurations))
 	for _, kmsConfiguration := range kmsConfigurations {
 		// ExtractUniqueAndSortedKMSConfigurations function rewrites the .Name field to include only the key ID
 		keyID := kmsConfiguration.Name
+		sockets = append(sockets, kmsConfiguration.Endpoint)
 
 		pluginConfig, ok := b.encryptionConfig.KMSPlugins[keyID]
 		if !ok {
@@ -135,6 +153,10 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string) 
 		if err := ensureReferenceDataVolume(podSpec, b.encryptionConfigSecretName); err != nil {
 			return err
 		}
+	}
+
+	if err := b.applyHealthReporter(podSpec, sockets); err != nil {
+		return err
 	}
 
 	return nil
--- a/pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
@@ -1,0 +1,124 @@
+package pluginlifecycle
+
+import (
+	"fmt"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/utils/ptr"
+)
+
+// defaultStaticPodKubeconfig is reused from the cert-syncer kubeconfig because
+// in-cluster config does not work on host-network static pods (kubernetes.default.svc
+// does not resolve). This matches the pattern used by the startup monitor.
+// TODO: move to a dedicated least-privilege kubeconfig once available.
+const defaultStaticPodKubeconfig = "/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig"
+
+// applyHealthReporter injects the health reporter sidecar. It is opt-in via
+// WithHealthReporter. Plugin lifecycle callers enable it; preflight callers
+// do not, since they only verify plugin reachability.
+func (b *KMSPluginBuilder) applyHealthReporter(podSpec *corev1.PodSpec, sockets []string) error {
+	if b.healthReporter == nil {
+		return nil
+	}
+
+	if len(sockets) == 0 {
+		return nil
+	}
+
+	if b.healthReporter.name == "" || b.healthReporter.operatorBinary == "" || b.healthReporter.image == "" || b.healthReporter.subcommand == "" {
+		return fmt.Errorf("health reporter name, operatorBinary, image and subcommand are required when WithHealthReporter is used")
+	}
+
+	b.healthReporter.sockets = sockets
+	if b.staticPod {
+		b.healthReporter.kubeconfig = defaultStaticPodKubeconfig
+	}
+
+	if err := ensureSidecarContainer(podSpec, b.healthReporter); err != nil {
+		return err
+	}
+
+	if err := ensureSocketVolume(podSpec); err != nil {
+		return err
+	}
+
+	socketMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: true}
+	if err := ensureVolumeMountInContainer(podSpec.InitContainers, b.healthReporter.Name(), socketMount); err != nil {
+		return err
+	}
+
+	if b.staticPod {
+		resourceDirMount := corev1.VolumeMount{Name: resourceDirVolumeName, MountPath: resourcesDir, ReadOnly: true}
+		if err := ensureVolumeMountInContainer(podSpec.InitContainers, b.healthReporter.Name(), resourceDirMount); err != nil {
+			return err
+		}
+		if err := setRunAsRoot(podSpec.InitContainers, b.healthReporter.Name()); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+type healthReporter struct {
+	name           string
+	operatorBinary string
+	subcommand     string
+	image          string
+	sockets        []string
+	kubeconfig     string
+}
+
+func (h *healthReporter) Name() string {
+	return h.name
+}
+
+func (h *healthReporter) BuildSidecarContainer() (corev1.Container, error) {
+	if len(h.sockets) == 0 {
+		return corev1.Container{}, fmt.Errorf("at least one KMS socket is required")
+	}
+
+	args := []string{
+		fmt.Sprintf("--kms-sockets=%s", strings.Join(h.sockets, ",")),
+		"--node-name=$(NODE_NAME)",
+	}
+	if h.kubeconfig != "" {
+		args = append(args, fmt.Sprintf("--kubeconfig=%s", h.kubeconfig))
+	}
+
+	return corev1.Container{
+		Name:                     h.Name(),
+		Image:                    h.image,
+		Command:                  []string{h.operatorBinary, h.subcommand},
+		Args:                     args,
+		ImagePullPolicy:          corev1.PullIfNotPresent,
+		RestartPolicy:            ptr.To(corev1.ContainerRestartPolicyAlways),
+		TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+		// NODE_NAME is expanded in --node-name=$(NODE_NAME) above so the
+		// health reporter can identify which node's health it is reporting.
+		Env: []corev1.EnvVar{
+			{
+				Name: "NODE_NAME",
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "spec.nodeName",
+					},
+				},
+			},
+		},
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("32Mi"),
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			ReadOnlyRootFilesystem:   ptr.To(true),
+			AllowPrivilegeEscalation: ptr.To(false),
+			Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+			SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+		},
+	}, nil
+}
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -52,7 +52,7 @@ func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSP
 //
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
 // The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
-func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
+func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, operatorBinary string, operatorImage string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
 	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
 		return nil
 	}
@@ -75,6 +75,7 @@ func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.Pod
 	return NewKMSPluginBuilder().
 		FromEncryptionConfig(encryptionConfigSecretName, cfg).
 		AsStaticPod().
+		WithHealthReporter(operatorBinary, operatorImage).
 		Apply(podSpec, containerName)
 }
 
@@ -82,7 +83,7 @@ func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.Pod
 //
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
 // The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
-func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
+func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, operatorBinary string, operatorImage string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
 	if !featureGateAccessor.AreInitialFeatureGatesObserved() {
 		return nil
 	}
@@ -104,6 +105,7 @@ func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, 
 
 	return NewKMSPluginBuilder().
 		FromEncryptionConfig(encryptionConfigSecretName, cfg).
+		WithHealthReporter(operatorBinary, operatorImage).
 		Apply(podSpec, containerName)
 }
 
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
@@ -126,6 +126,39 @@ func TestAddKMSPluginSidecarToPodSpec(t *testing.T) {
 		"-metrics-port=0",
 	}
 
+	expectedHealthReporter := func(sockets string) corev1.Container {
+		return corev1.Container{
+			Name:                     "kms-health-reporter",
+			Image:                    "quay.io/test/operator:latest",
+			Command:                  []string{"cluster-kube-apiserver-operator", "kms-health-reporter"},
+			Args:                     []string{fmt.Sprintf("--kms-sockets=%s", sockets), "--node-name=$(NODE_NAME)"},
+			ImagePullPolicy:          corev1.PullIfNotPresent,
+			RestartPolicy:            ptr.To(corev1.ContainerRestartPolicyAlways),
+			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+			Env: []corev1.EnvVar{
+				{
+					Name: "NODE_NAME",
+					ValueFrom: &corev1.EnvVarSource{
+						FieldRef: &corev1.ObjectFieldSelector{FieldPath: "spec.nodeName"},
+					},
+				},
+			},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceMemory: resource.MustParse("32Mi"),
+					corev1.ResourceCPU:    resource.MustParse("10m"),
+				},
+			},
+			SecurityContext: &corev1.SecurityContext{
+				ReadOnlyRootFilesystem:   ptr.To(true),
+				AllowPrivilegeEscalation: ptr.To(false),
+				Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+				SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+			},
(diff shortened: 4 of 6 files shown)


pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -137,9 +158,89 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string)
...
+
+		reporter := &healthReporter{name: b.healthReporterContainerName, image: b.healthReporterImage, sockets: sockets, kubeconfig: b.healthReporterKubeconfig}
+		if err := ensureSidecarContainer(podSpec, reporter); err != nil {
+			return err
+		}
+
+		healthReporterSocketMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: true}
+		if err := ensureVolumeMountInContainer(podSpec.InitContainers, reporter.Name(), healthReporterSocketMount); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
ardaguclu: Please check latest status, it should be resolved.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -137,9 +156,104 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string)
...
+
+		sockets := make([]string, 0, len(kmsConfigurations))
+		for _, kmsConfiguration := range kmsConfigurations {
+			sockets = append(sockets, kmsConfiguration.Endpoint)
+		}
+
+		var kubeconfig string
+		if b.staticPod {
+			kubeconfig = defaultStaticPodKubeconfig
```
ibihim: Nit: I know that it is idiomatic in K8s code, but the apply function is very long. We could split it up?

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -137,9 +156,104 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string)
...
 	return nil
 }
 
+// defaultStaticPodKubeconfig is reused from the cert-syncer kubeconfig because
+// in-cluster config does not work on host-network static pods (kubernetes.default.svc
+// does not resolve). This matches the pattern used by the startup monitor.
+// TODO: move to a dedicated least-privilege kubeconfig once available.
+const defaultStaticPodKubeconfig = "/etc/kubernetes/static-pod-resources/configmaps/kube-apiserver-cert-syncer-kubeconfig/kubeconfig"
+
+type healthReporter struct {
```
ibihim: Nit: I know that long files are also idiomatic in K8s, but maybe it would deserve its own file?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -52,7 +53,7 @@ func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSP
 //
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
 // The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
-func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
+func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess, operatorImage string) error {
```
ibihim: Nit: there are more than 4 args, worth considering functional arguments or a config instead. I know it is idiomatic to have plenty of args in Kubernetes.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -137,9 +156,104 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string)
...
+				},
+			},
+		},
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("32Mi"),
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
```
ibihim: I reviewed a PR from Fabio, where this was being set. Is this the same? Does it make sense to have one specific one that can be referenced or called?

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -20,6 +23,11 @@ type KMSPluginBuilder struct {
 	encryptionConfig           *encryptiondata.Config
 	encryptionConfigSecretName string
 	staticPod                  bool
+
+	enableHealthReporter        bool
```
ibihim: Is there a scenario, where we don't want to enable the health reporter?

ardaguclu: yes, preflight checker won't use health reporter.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: Good idea. I've moved the logic inside `applyHealthReporter`.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: Good idea. I've moved this logic inside `health_reporter.go`.

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: It makes sense. But I'd prefer keeping this PR only focusing on health reporter changes without refactoring the other bits.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: They are identical. But is there a guarantee that kms plugins and health reporter use the same security context?. I think, they should be configured separately?

ardaguclu:
/retitle [CNTRLPLANE-3363](https://redhat.atlassian.net/browse/CNTRLPLANE-3363): Inject health reporter as sidecar container

ardaguclu:
/hold
until CI is green

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ibihim: Now that you say it, it is obvious. And we won't need a comment, because `sidecar.go` will contain a function for the preflight scenario, right?

ardaguclu: That makes sense. I've updated with documentation.

ardaguclu:
CI jobs passed;
* cluster-authentication-operator: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/924/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-kms/2068942752682348544
* cluster-openshift-apiserver-operator: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/717/pull-ci-openshift-cluster-openshift-apiserver-operator-main-e2e-aws-operator-encryption-kms/2068942750169960448
* cluster-kube-apiserver-operator: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2206/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-aws-operator-encryption-kms/2068942745975656448

ardaguclu:
Unrelated
/retest

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	socketMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: true}
+	if err := ensureVolumeMountInContainer(podSpec.InitContainers, reporter.Name(), socketMount); err != nil {
+		return err
+	}
+
+	if b.staticPod {
+		resourceDirMount := corev1.VolumeMount{Name: resourceDirVolumeName, MountPath: resourcesDir, ReadOnly: true}
+		if err := ensureVolumeMountInContainer(podSpec.InitContainers, reporter.Name(), resourceDirMount); err != nil {
```
bertinatto: I find it very elegant that you reused `ensureSidecarContaienr`, `ensureVolumeMountInContainer`

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -20,6 +20,11 @@ type KMSPluginBuilder struct {
...
+	healthReporterContainerName string
+	healthReporterOperatorCmd   string
+	healthReporterImage         string
```
bertinatto: would it make sense to create the `healthReport` here instead? Also, if we make it a pointer, we wouldn't need the boolean

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+		Env: []corev1.EnvVar{
+			{
+				Name: "NODE_NAME",
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "spec.nodeName",
+					},
+				},
+			},
+		},
```
bertinatto, line 110: what do we need this env for? Can we add a comment?

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+		"--node-name=$(NODE_NAME)",
+	}
+	if h.kubeconfig != "" {
+		args = append(args, fmt.Sprintf("--kubeconfig=%s", h.kubeconfig))
+	}
+
+	return corev1.Container{
+		Name:                     h.name,
+		Image:                    h.image,
+		Command:                  []string{h.operatorCmd, h.name},
```
bertinatto: do you need to specify `Command` because KASO doesn't have an entrypoint?

in your KASO proof PR I see:

```
if err := kmspluginlifecycle.AddKMSPluginSidecarToStaticPodSpec(ctx, &required.Spec, "kube-apiserver", operatorclient.TargetNamespace, "encryption-config", secretClient, featureGateAccessor, "cluster-kube-apiserver-operator", operatorImagePullSpec); err != nil {
```

so I think `Command` here will be `cluster-kube-apiserver-operator kms-health-reporter`.  This makes me wonder if we need to make sure the PATH is correctly set or if we need to use the whole path (`/usr/bin/cluster-kube-apiserver-operator`).

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+type healthReporter struct {
+	name        string
+	operatorCmd string
+	image       string
+	sockets     []string
+	kubeconfig  string
+}
+
+func (h *healthReporter) Name() string {
+	return h.name
```
bertinatto, line 75: it seems like `h.name` is both the container name (for which it's correctly used in this method), and also the operator cli command. It'd be nice to make that distinction visible somehow

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (h *healthReporter) BuildSidecarContainer() (corev1.Container, error) {
+	args := []string{
+		fmt.Sprintf("--kms-sockets=%s", strings.Join(h.sockets, ",")),
+		"--node-name=$(NODE_NAME)",
+	}
+	if h.kubeconfig != "" {
+		args = append(args, fmt.Sprintf("--kubeconfig=%s", h.kubeconfig))
+	}
```
bertinatto, line 89: isn't the kubeconfig required for all health reporters?

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+	kubeconfig  string
+}
+
+func (h *healthReporter) Name() string {
+	return h.name
+}
+
+func (h *healthReporter) BuildSidecarContainer() (corev1.Container, error) {
+	args := []string{
+		fmt.Sprintf("--kms-sockets=%s", strings.Join(h.sockets, ",")),
```
bertinatto, line 84: do we need to validate for an empty slice?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -75,14 +76,15 @@ func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.Pod
...
+		WithHealthReporter(healthReporterContainerName, operatorCmd, operatorImage).
 		Apply(podSpec, containerName)
 }
 
 // AddKMSPluginSidecarToPodSpec injects KMS plugin sidecar containers into an aggregated API server pod spec (e.g., openshift-apiserver, oauth-apiserver).
 //
 // It is a no-op when the KMSEncryption feature gate is not enabled or the encryption-config secret does not exist.
 // The secretClient should be uncached to avoid injecting sidecars based on a stale encryption configuration.
-func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess) error {
+func AddKMSPluginSidecarToPodSpec(ctx context.Context, podSpec *corev1.PodSpec, containerName string, encryptionConfigNamespace string, encryptionConfigSecretName string, secretClient corev1client.SecretsGetter, featureGateAccessor featuregates.FeatureGateAccess, operatorCmd string, operatorImage string) error {
```
bertinatto: interesting, I see that the authentication operator has the entreypoint correctly set to `/usr/bin/authentication-operator`, so we wouldn't need `operatorCmd` here, right?

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+				corev1.ResourceMemory: resource.MustParse("32Mi"),
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			ReadOnlyRootFilesystem:   ptr.To(true),
+			AllowPrivilegeEscalation: ptr.To(false),
+			Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+			SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+		},
```
bertinatto, line 122: can we be more strict with the health reporter?

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: nit: the params ordering seems a bit off. IMO the feature gate acessor fits better at the end. What about something like this?

```
  func AddKMSPluginSidecarToStaticPodSpec(
      ctx context.Context,
      podSpec *corev1.PodSpec,
      containerName string,
      encryptionConfigNamespace string,
      encryptionConfigSecretName string,
      operatorCmd string,
      operatorImage string,
      secretClient corev1client.SecretsGetter, // this is not necessarily tied to the encryptionConfig*
      featureGateAccessor featuregates.FeatureGateAccess,
  ) error
```

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: Yes, it makes sense. Updated.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: I agree that it is confusing. I've updated by using subcommand (which refers to the command name in health/cmd.go). Please let me know your thoughts.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: Yes, it is better to validate it. Updated.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: cluster-kube-apiserver-operator requires kubeconfig but others won't need it (they will use in-cluster kubeconfig)

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: I've updated the code to also support empty commands, so that entrpoint will be automatically used (if command is passed as empty). But it is caller's responsibility to verify the presence of the binary IMO.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: Necessity is documented.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: There is nothing left apart from `RunAsRoot: false`?. KAS-o needs to run as root (uid: 0)

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
ardaguclu: I agree, if we pass operatorCmd empty (to fallback to entrypoint), mechanism should work. So I've updated the code to handle that case.

ardaguclu: Updated.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var kubeconfig string
+	if b.staticPod {
+		kubeconfig = defaultStaticPodKubeconfig
+	}
+
+	reporter := &healthReporter{subcommand: b.healthReporter.subcommand, operatorCmd: b.healthReporter.operatorCmd, image: b.healthReporter.image, sockets: sockets, kubeconfig: kubeconfig}
+	if err := ensureSidecarContainer(podSpec, reporter); err != nil {
+		return err
+	}
```
bertinatto, line 45: nit: we could get rid of the `kubeconfig` variable if we checked for static pod after creating the `reporter` var, something like:

```
	reporter := &healthReporter{
		subcommand:  b.healthReporter.subcommand,
		operatorCmd: b.healthReporter.operatorCmd,
		image:       b.healthReporter.image,
		sockets:     sockets,
	}

	if b.staticPod {
		reporter.kubeconfig = defaultStaticPodKubeconfig
	}
```

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,127 @@
...
+
+	socketMount := corev1.VolumeMount{Name: kmsPluginSocketVolumeName, MountPath: kmsPluginSocketMountPath, ReadOnly: true}
+	if err := ensureVolumeMountInContainer(podSpec.InitContainers, reporter.Name(), socketMount); err != nil {
+		return err
+	}
```
bertinatto, line 50: do we also need to add the socket volume to `podSpec.Volumes` (`ensureSocketVolume()`), otherwise the pod fails to start, right? I assume it's not failing in your proof PR because the volume is being added to the podSpec by the KMS plugin injection code path

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,127 @@
...
+
+	return nil
+}
+
+type healthReporter struct {
+	// subcommand is the operator CLI subcommand registered via health.NewCommand
+	// (see pkg/operator/encryption/kms/health/cmd.go). Also used as the init container name.
+	subcommand  string
+	operatorCmd string
```
bertinatto: nit: I find there's some asymmetry in `operatorCmd` and `subcommand`, so I'll suggest some alternatives, feel free to ignore if you don't like:

* `entrypoint/subcommand`
* `mainCommand/subcommand`
* `rootCommand/command` (in cobra's terminology)
* `operatorBinary/subcommand`

also, you might want to switch the oder (command before subcommand)

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -44,6 +46,19 @@ func (b *KMSPluginBuilder) AsStaticPod() *KMSPluginBuilder {
 	return b
 }
 
+// WithHealthReporter enables injection of a health-reporter sidecar.
+// subcommand is the operator CLI subcommand (e.g. "kms-health-reporter").
+// operatorCmd is the parent binary (e.g. "cluster-kube-apiserver-operator");
+// when empty, the subcommand is invoked directly via the image ENTRYPOINT.
+func (b *KMSPluginBuilder) WithHealthReporter(subcommand, operatorCmd, operatorImage string) *KMSPluginBuilder {
```
bertinatto: nit: should command come before the subcommand?

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
bertinatto: part of me feels like we should split the concerns and have 2 fields (`name` and `subcommand`) and have different values for them, but I can also go with what we have. Maybe we need one more opinion? :)

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -75,14 +76,15 @@ func AddKMSPluginSidecarToStaticPodSpec(ctx context.Context, podSpec *corev1.Pod
 	return NewKMSPluginBuilder().
 		FromEncryptionConfig(encryptionConfigSecretName, cfg).
 		AsStaticPod().
+		WithHealthReporter(health.Subcommand, operatorCmd, operatorImage).
```
bertinatto: if subcommand is always the same, could we drop it from this function?

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: You are right. Updated.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: Nice catch. Updated to call `ensureSocketVolume()`. This function is idempotent.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: Fully agree. Updated. Now we have 2 separate notions -- subcommand and name.

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: Agree. All your suggestions make sense. I've picked `operatorBinary/subcommand`. Please let me know your thoughts.

ardaguclu:
Fake bumps to test the latest version;
* [cluster-kube-apiserver-operator](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2210)
* [cluster-openshift-apiserver-operator](https://github.com/openshift/cluster-openshift-apiserver-operator/pull/721)
* [cluster-authentication-operator](https://github.com/openshift/cluster-authentication-operator/pull/930)

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,130 @@
...
+		"--node-name=$(NODE_NAME)",
+	}
+	if h.kubeconfig != "" {
+		args = append(args, fmt.Sprintf("--kubeconfig=%s", h.kubeconfig))
+	}
+
+	return corev1.Container{
+		Name:                     h.Name(),
+		Image:                    h.image,
+		Command:                  []string{h.operatorBinary, h.Subcommand()},
```
ardaguclu: @bertinatto If entrypoint is defined in the Containerfile, I think we have to pass subcommand into the args rather than Command. But for the simplicity maybe we can enforce passing the non-empty operatorBinary?.

bertinatto: yeah, totally agree

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (h *healthReporter) Name() string {
+	return h.name
+}
+
+func (h *healthReporter) Subcommand() string {
+	// subcommand is the operator CLI subcommand registered via health.NewCommand
+	// (see pkg/operator/encryption/kms/health/cmd.go). Also used as the init container name.
+	return h.subcommand
```
bertinatto: this method is not part of the sidecarProvider interface, and I don't think it's used anywhere else except this file. If that's correct, can we drop this method and just use `h.subcommand` directly?

pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
```diff
@@ -126,6 +126,39 @@ func TestAddKMSPluginSidecarToPodSpec(t *testing.T) {
...
+			SecurityContext: &corev1.SecurityContext{
+				ReadOnlyRootFilesystem:   ptr.To(true),
+				AllowPrivilegeEscalation: ptr.To(false),
+				Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
+				SeccompProfile:           &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault},
+			},
+			VolumeMounts: []corev1.VolumeMount{{Name: "kms-plugin-socket", MountPath: "/var/run/kmsplugin", ReadOnly: true}},
+		}
+	}
+
```
bertinatto, line 161: nit: now that we introduce more than one sidecar, I wonder if we should have a similar function for the other sidecar

pkg/operator/encryption/kms/pluginlifecycle/sidecar_test.go
```diff
@@ -513,6 +548,7 @@ func TestAddKMSPluginSidecarToPodSpec(t *testing.T) {
 						},
 						VolumeMounts: []corev1.VolumeMount{socketMount, refDataMount},
 					},
+					expectedHealthReporter("unix:///var/run/kmsplugin/kms-555.sock"),
```
bertinatto, line 551: it seems like we're only testing the health injection in `AddKMSPluginSidecarPodSpec`, which I think is fine

bertinatto (commented):
Just added a few optional nits that could be addressed in a different PR if they make sense.

/lgtm

Holding for tests

/hold

pkg/operator/encryption/kms/pluginlifecycle/health_reporter.go
ardaguclu: @bertinatto I've updated this PR to drop this unnecessary function. Does it make sense?

ardaguclu:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-openshift-apiserver-operator/721/pull-ci-openshift-cluster-openshift-apiserver-operator-main-e2e-aws-operator-encryption-kms/2069677056613945344 passed
I'll wait the others

ardaguclu:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/930/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-kms/2069759934207102976 passed

ardaguclu:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-kube-apiserver-operator/2210/pull-ci-openshift-cluster-kube-apiserver-operator-main-e2e-aws-operator-encryption-kms/2069760042168487936 passed
/hold cancel

PR #2318, opened 2026-06-19
CNTRLPLANE-3234: health reporter writer

ibihim:
## What

Add health report status writer.

## Why

We need to write the status into the status of each individual (aggregated) apiserver's operator status for access to KMSv2 controllers.

## Notes

Check usage pattern at https://github.com/openshift/cluster-authentication-operator/pull/929.

## Summary by CodeRabbit

* **New Features**
  * KMS health checks now publish encryption status updates instead of only logging probe results.
  * Health reporting now works consistently across the supported control plane status views.

* **Bug Fixes**
  * Improved status mapping so unexpected health values are reported safely as errors.
  * Added clearer handling for missing or empty report details to avoid invalid status updates.

* **Tests**
  * Added coverage for end-to-end health report generation and status publishing behavior.


pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -35,21 +41,21 @@ type options struct {
 	NodeName     string
 	Kubeconfig   string
 
-	newOperatorClient func(*rest.Config) (v1helpers.OperatorClient, error)
+	newWriter NewEncryptionStatusWriterFunc
```
ardaguclu: Before discussing the details, I think it would be very useful to have a PoC PR in respective operators to see how callers will call this (https://github.com/openshift/cluster-kube-apiserver-operator/blob/4e7619777cb6e5bb7853836d07656da3e1eb14df/cmd/cluster-kube-apiserver-operator/main.go#L70-L74)

pkg/operator/encryption/kms/health/writers/writers.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/client-go/rest"
+
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/health"
+)
+
+// applyOpts ensures that we are the dominant fieldManager.
+func applyOpts(fieldManager string) metav1.ApplyOptions {
```
ardaguclu: Do we really need this just for Force?

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -114,6 +116,9 @@ func (o *options) validate() error {
 	if o.NodeName == "" {
 		return fmt.Errorf("--node-name is required")
 	}
+	if fieldManager := fieldManagerPrefix + o.NodeName; len(fieldManager) > metav1validation.FieldManagerMaxLength {
+		return fmt.Errorf("--node-name too long: reporter identity %q is %d chars, exceeds the %d-char fieldManager limit", fieldManager, len(fieldManager), metav1validation.FieldManagerMaxLength)
```
ardaguclu: Can we really hit this error?.

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -168,7 +179,7 @@ func buildPlugins(ctx context.Context, sockets []string, timeout time.Duration)
 
 		// Unique name per plugin so the gRPC client's KMS operation metrics
 		// don't merge both plugins into one series.
-		service, err := k8senvelopekmsv2.NewGRPCService(ctx, socket, Subcommand+"-"+keyID, timeout)
+		service, err := k8senvelopekmsv2.NewGRPCService(ctx, socket, fieldManagerPrefix+keyID, timeout)
```
ardaguclu:
```suggestion
		service, err := k8senvelopekmsv2.NewGRPCService(ctx, socket, fmt.Sprintf("%s-%s", Subcommand, keyID), timeout)
```

pkg/operator/encryption/kms/health/writers/writers.go
```diff
@@ -0,0 +1,90 @@
...
+	return metav1.ApplyOptions{FieldManager: fieldManager, Force: true}
+}
+
+// NewAuthenticationWriter writes into Authentication/cluster at
+// .status.oauthAPIServer.encryptionStatus. The auth operator manages the
+// oauth-apiserver. There is an extra hop the other two operators don't have.
+func NewAuthenticationWriter(restConfig *rest.Config, fieldManager string) (health.EncryptionStatusWriter, error) {
```
ardaguclu: I really like this mechanism

ardaguclu (commented):
Overall looks good to me

pkg/operator/encryption/kms/health/writer_test.go
```diff
@@ -0,0 +1,67 @@
...
+			WithKeyId("3").
+			WithStatus(operatorv1.KMSPluginHealthStatusError).
+			WithLastCheckedTime(metav1.NewTime(checked)).
+			WithDetail("DeadlineExceeded"),
+	)
+
+	require.Equal(t, want, have)
+}
+
+func TestMapStatus(t *testing.T) {
```
ardaguclu: Do we need this test?

pkg/operator/encryption/kms/health/writer_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+	applyoperatorv1 "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestBuildEncryptionStatus(t *testing.T) {
+	// Fixed UTC time dodges Go's monotonic clock and timezone drift.
+	checked := time.Unix(0, 0).UTC()
```
ardaguclu: There is a fakeclock, or something like this. Maybe we can use it

pkg/operator/encryption/kms/health/cmd.go
```diff
@@ -126,9 +131,11 @@ func (o *options) Config(ctx context.Context) (*Config, error) {
 		return nil, fmt.Errorf("build rest config: %w", err)
 	}
 
-	operatorClient, err := o.newOperatorClient(restCfg)
+	// fieldManager is the per-node ownership identity.
+	fieldManager := fieldManagerPrefix + o.NodeName
```
ardaguclu: In another place we use `fieldManagerPrefix + keyID` but here we use `fieldManagerPrefix + o.NodeName`. Isn't this a discrepancy?

pkg/operator/encryption/kms/health/cmd.go
tjungblu: how would you even get out of this situation as an admin? rename your nodes?

tjungblu: with the prefix we have like 100 characters left, I think it's OK. We can also make the prefix a bit shorter?

ardaguclu: Maybe we can just continue as is and allow kube-apiserver errors out?

pkg/operator/encryption/kms/health/writer.go
```diff
@@ -0,0 +1,58 @@
...
+func buildEncryptionStatus(nodeName string, reports []pluginHealthReport) *applyoperatorv1.KMSEncryptionStatusApplyConfiguration {
+	healthReports := make([]*applyoperatorv1.KMSPluginHealthReportApplyConfiguration, 0, len(reports))
+	for _, r := range reports {
+		hr := applyoperatorv1.KMSPluginHealthReport().
+			WithNodeName(nodeName).
+			WithKeyId(r.KeyID).
+			WithStatus(mapStatus(r.Status)).
+			WithLastCheckedTime(metav1.NewTime(r.LastChecked))
+
+		// kekId/detail have MinLength=1; setting "" would fail validation.
```
tjungblu: for kekId I think the behavior is correct, because we want to retain the existing value, even when we transiently can't connect to the plugin.

The Detail, I'm not so sure. What do you think?

pkg/operator/encryption/kms/health/cmd.go
tjungblu: you can probably even leave this empty, nobody is going to look at the metric this string generates

pkg/operator/encryption/kms/health/cmd.go
tjungblu: I think he has a case with this, but we probably need a different solution here 

checkout the max length we can have per node:
https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validate/content/dns.go#L56-L57

pkg/operator/encryption/kms/health/cmd.go
tjungblu: OR just use the Subcommand on its own, I guess the metrics can then be sliced by node/podip or so which you get for free by prometheus scraper

pkg/operator/encryption/kms/health/writer.go
```diff
@@ -0,0 +1,58 @@
...
+
+		healthReports = append(healthReports, hr)
+	}
+
+	return applyoperatorv1.KMSEncryptionStatus().WithHealthReports(healthReports...)
+}
+
+// mapStatus defaults to Error so an unknown value never becomes an empty,
+// invalid enum.
+func mapStatus(s string) operatorv1.KMSPluginHealthStatus {
```
tjungblu: any reason not to directly return the operator constant from pluginHealthReport?

as in, you probably can just avoid that switch and put it directly into the prober logic

pkg/operator/encryption/kms/health/writers/writers.go
ibihim: no, we could remove it.

pkg/operator/encryption/kms/health/writer_test.go
ibihim: I can remove it.

pkg/operator/encryption/kms/health/cmd.go
ibihim: we could invert the string, to have the uniqueness in the front :rofl:

pkg/operator/encryption/kms/health/cmd.go
ibihim: `+ o.NodeName` distinguishes the 3 health reporters in the KAS setup. So this will be unique per writer.

`+ keyID` is to distinguish individual sockets within a node. This was the case in creating the service, but I dropped it to just the `Subcommand` according to https://github.com/openshift/library-go/pull/2318#discussion_r3473500601.

pkg/operator/encryption/kms/health/writer.go
ibihim: AFAIU this is an alternating logic...

`KekID` would be empty on an error.
`Detail` would be empty without an error.

pkg/operator/encryption/kms/health/writer.go
ibihim: lol. This code goes through its third iteration at this point... I seem to have lost the eye for detail.

ibihim: Do we need the "pluginHealthReport" struct at all, we could use the API directly, no?

pkg/operator/encryption/kms/health/writer.go
```diff
@@ -0,0 +1,18 @@
...
+
+// NewEncryptionStatusWriterFunc builds the EncryptionStatusWriter for a target
+// apiserver operator status CR. fieldManager sets the owner in the
+// managedFields when doing SSA.
+type NewEncryptionStatusWriterFunc func(restConfig *rest.Config, fieldManager string) (EncryptionStatusWriter, error)
+
+// EncryptionStatusWriter is capable of applying the
+// KMSEncryptionStatusApplyConfiguration at the correct place in the operator's
+// status.
+type EncryptionStatusWriter func(ctx context.Context, status *applyoperatorv1.KMSEncryptionStatusApplyConfiguration) error
```
ibihim: Add this point, we could move this to cmd.go. It is small enough.

If cmd.go is considered too big, we could move out the options logic into its own file.

pkg/operator/encryption/kms/health/cmd.go
tjungblu: I was thinking about using a hash or the node id? how about that? a uuid fits nicely

pkg/operator/encryption/kms/health/writer.go
ardaguclu: Just an idea; Maybe we can move functions under `writers/writers.go` to here?

pkg/operator/encryption/kms/health/cmd.go
ardaguclu: I agree with Thomas. If it is possible, maybe we can use something else (uuid, hash, etc.) to remove that check. I think, other than this thread, this PR is ready to merge.

tjungblu: we can do that as a follow-up. This is probably a bit bigger because we have to rename the cmdline and add this in the lifecycle

ardaguclu: What about cutting the static length from node name, if it exceeds?

tjungblu: you'd have to cut the unique bits from the back and hope they don't have a subdomain per node, as Krzys says

ardaguclu: That makes sense. Thank you

ardaguclu:
@ibihim the task that wiring of health reporter to the operators has been finished. Now you can test this PR on operators via fake bump. So that we can verify that changes here work.

pkg/operator/encryption/kms/health/writer.go
ibihim: I liked the idea to have them as something distinct, like they would be "already on the operator side". But if you like I could move them here.

ardaguclu: Not needed. That works for me. Thank you

ardaguclu:
[CI Job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-authentication-operator/937/pull-ci-openshift-cluster-authentication-operator-master-e2e-aws-operator-encryption-kms/2071617585568288768) in [fake PR](https://github.com/openshift/cluster-authentication-operator/pull/937) proves that health of kms plugins are reported;

```
$ omc get authentications.operator.openshift.io cluster -oyaml
apiVersion: operator.openshift.io/v1
kind: Authentication
metadata:
  annotations:
    include.release.openshift.io/self-managed-high-availability: "true"
    include.release.openshift.io/single-node-developer: "true"
    release.openshift.io/create-only: "true"
  creationTimestamp: "2026-06-29T16:02:49Z"
  generation: 11
  name: cluster
....
oauthAPIServer:
    encryptionStatus:
      healthReports:
      - kekId: CgQKAnYxGjwKJgoFYWRtaW4aHQobaHR0cHM6Ly8xNzIuMzAuMjM4LjE2Nzo4MjAwGgd0cmFuc2l0IgdrbXMta2V5KA
        keyId: "6"
        lastCheckedTime: "2026-06-29T17:49:06Z"
        nodeName: ip-10-0-42-231.ec2.internal
        status: Healthy
      - kekId: CgQKAnYxGjwKJgoFYWRtaW4aHQobaHR0cHM6Ly8xNzIuMzAuMjM4LjE2Nzo4MjAwGgd0cmFuc2l0IgdrbXMta2V5KA
        keyId: "6"
        lastCheckedTime: "2026-06-29T17:49:09Z"
        nodeName: ip-10-0-20-198.ec2.internal
        status: Healthy
      - kekId: CgQKAnYxGjwKJgoFYWRtaW4aHQobaHR0cHM6Ly8xNzIuMzAuMjM4LjE2Nzo4MjAwGgd0cmFuc2l0IgdrbXMta2V5KA
        keyId: "6"
        lastCheckedTime: "2026-06-29T17:49:04Z"
        nodeName: ip-10-0-2-175.ec2.internal
        status: Healthy
```

pkg/operator/encryption/kms/health/options.go
```diff
@@ -0,0 +1,97 @@
...
+		return fmt.Errorf("--read-timeout must be positive")
+	}
+	if o.WriteTimeout <= 0 {
+		return fmt.Errorf("--write-timeout must be positive")
+	}
+	if o.NodeName == "" {
+		return fmt.Errorf("--node-name is required")
+	}
+	if fieldManager := fmt.Sprintf("%s-%s", Subcommand, o.NodeName); len(fieldManager) > metav1validation.FieldManagerMaxLength {
+		return fmt.Errorf("--node-name too long: reporter identity %q is %d chars, exceeds the %d-char fieldManager limit", fieldManager, len(fieldManager), metav1validation.FieldManagerMaxLength)
```
ardaguclu: This one is probably the only part that requires a change.

pkg/operator/encryption/kms/health/writers/writers.go
```diff
@@ -0,0 +1,110 @@
...
+			// is enough and always fits the apiserver's length limit.
+			metav1.ApplyOptions{FieldManager: health.Subcommand, Force: true},
+		)
+		return err
+	}, nil
+}
+
+// NewKubeAPIServerWriter writes into KubeAPIServer/cluster at
+// .status.encryptionStatus.
+func NewKubeAPIServerWriter(restConfig *rest.Config, nodeName string) (health.EncryptionStatusWriter, error) {
```
ardaguclu: In the previous version, writers were consistent. They just accept the field manager and set it accordingly. Now we leak the field manager extraction into the writer and this makes them inconsistent (some hardcodes, some send request to API Server). In my opinion, it would be better to return back to previous version by passing field manager.

pkg/operator/encryption/kms/health/options.go
```diff
@@ -0,0 +1,94 @@
...
+	// token + KUBERNETES_SERVICE_HOST), which is the deployed path.
+	restCfg, err := clientcmd.BuildConfigFromFlags("", o.Kubeconfig)
+	if err != nil {
+		return nil, fmt.Errorf("build rest config: %w", err)
+	}
+
+	// The writer owns its fieldManager: it turns the node name into whatever
+	// ownership identity its status needs (a per-node UID for the kube-apiserver,
+	// a constant for the single-reporter operators).
+	writeStatus, err := o.newWriter(restCfg, o.NodeName)
```
ardaguclu: Can't we get UID here and pass it as field manager to `newWriter`?. Why shouldn't we use UID for aggregated apiservers?.

tjungblu: what if we do a simpler step and just set the `--node-name` argument as the node.UUID instead of the node.Name from the lifecycle. 

We can rename the argument later. I don't really think we should be querying node objects from here just to resolve the name.

ibihim: Yeah, I must admit, that querying node-objects is not optimal, but I didn't want to offload it to somewhere else, but we can do so. Maybe the lifecycle has it at hand already.

pkg/operator/encryption/kms/health/writers/writers.go
ibihim: Honestly, I thought this is my master piece :D The operators themselves know best what to put in there, but based on the discussion in the other comment, we can keep it as is.

pkg/operator/encryption/kms/health/options.go
ibihim: I can drop the comment and we "interpret" the node-name flag creatively as something unique within bounds like an UID?

I might adjust the description for the flag, but not sure if it is worth changing the flag itself, maybe? Because of tests? WDYT?

tjungblu:
/approve
/lgtm

I have a bit of time at my hands to figure out the nodename hashing

p0lyn0mial:
approving without a code review. 

/approve

PR #2321, opened 2026-06-22
NO-JIRA: inject kms sidecar in preflight deployer

tjungblu:
## Summary by CodeRabbit

* **New Features**
  * Preflight deployment now provisions KMS plugin components (including a plugin socket/init container) with the required encryption configuration.
  * Added support for creating and cleaning up the encryption configuration secret used by the preflight workflow.
* **Bug Fixes**
  * Deployment now validates required inputs and safely replaces stale preflight resources using a consistent delete-then-create order.
  * Cleanup performs best-effort deletion of both the preflight pod and its secret, tolerating missing resources and reporting combined failures.
* **Tests**
  * Expanded unit tests to cover secret/plugin injection and detailed cleanup/error ordering scenarios.
* **Improvements**
  * Set explicit CPU/memory resource requests for the preflight check container.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -2,9 +2,11 @@ package preflight
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/pluginlifecycle"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -12,7 +14,9 @@ import (
 )
 
 const (
-	preflightPodName = "kms-preflight"
+	preflightPodName                    = "kms-preflight"
+	preflightCheckContainerName         = "kms-preflight-check"
+	preflightEncryptionConfigSecretName = "kms-preflight-encryption-config"
 )
 
 // PodPreflightDeployer creates a one-shot pod to run the preflight checker as a command,
@@ -28,7 +32,27 @@ type PodPreflightDeployer struct {
 	kmsCallTimeout  time.Duration
 }
 
-func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) error {
+func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, encryptionConfigSecret *corev1.Secret) error {
+	if configHash == "" {
+		return fmt.Errorf("configHash is empty")
+	}
+	if encryptionConfigSecret == nil {
+		return fmt.Errorf("encryptionConfigSecret is nil")
+	}
+
+	// ensure that there is nothing left over from previous runs
+	if err := d.Cleanup(ctx); err != nil {
+		return fmt.Errorf("failed to clean up existing preflight resources: %w", err)
+	}
+
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	// rewrite the entire ObjectMeta to avoid copying resource versions or managed fields
+	encryptionConfigSecret.ObjectMeta = metav1.ObjectMeta{Namespace: d.namespace, Name: preflightEncryptionConfigSecretName}
+	_, err := d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create preflight encryption config secret: %w", err)
+	}
+
 	pod, err := generatePodTemplate(
 		preflightPodName,
 		d.namespace,
@@ -41,10 +65,20 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
 		return fmt.Errorf("failed to generate preflight pod template: %w", err)
 	}
 
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	err = pluginlifecycle.NewKMSPluginBuilder().
+		WithSecretRequired().
+		FromEncryptionConfigSecret(d.namespace, preflightEncryptionConfigSecretName, d.coreClient).
+		Apply(ctx, &pod.Spec, preflightCheckContainerName)
+	if err != nil {
+		return fmt.Errorf("failed to apply preflight plugin: %w", err)
+	}
 
 	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
-	return err
+	if err != nil {
+		return fmt.Errorf("failed to create preflight pod: %w", err)
+	}
+
+	return nil
 }
 
 func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, error) {
@@ -58,11 +92,19 @@ func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, er
 }
 
 func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
+	var errs []error
+
 	err := d.coreClient.Pods(d.namespace).Delete(ctx, preflightPodName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err)
+		errs = append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err))
 	}
-	return nil
+
+	err = d.coreClient.Secrets(d.namespace).Delete(ctx, preflightEncryptionConfigSecretName, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		errs = append(errs, fmt.Errorf("failed to delete secret %s/%s: %w", d.namespace, preflightEncryptionConfigSecretName, err))
+	}
+
+	return errors.Join(errs...)
 }
 
 func NewPodPreflightDeployer(
--- a/pkg/operator/encryption/kms/preflight/deployer_test.go
+++ b/pkg/operator/encryption/kms/preflight/deployer_test.go
@@ -6,12 +6,16 @@ import (
 	"testing"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+	encryptiontesting "github.com/openshift/library-go/pkg/operator/encryption/testing"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	clienttesting "k8s.io/client-go/testing"
 )
@@ -22,6 +26,7 @@ const (
 	testOperatorImage       = "quay.io/openshift-release-dev/ocp-v5.0-art-dev@sha256:test"
 	configHashAnnotationKey = "encryption.apiserver.operator.openshift.io/kms-preflight-config-hash"
 	testDeployerTimeout     = 10 * time.Second
+	testReferenceDataNS     = "openshift-config"
 )
 
 var testOperatorCommand = []string{"cluster-kube-apiserver-operator", "kms-preflight"}
@@ -48,6 +53,39 @@ spec:
     - key: node-role.kubernetes.io/master
       operator: Exists
       effect: NoExecute
+  initContainers:
+    - name: vault-kms-plugin-1
+      image: registry.example.com/kms-plugin@sha256:abcdef1234567890abcdef1234567890abcdef1234567890abcdef1234567890
+      args:
+        - -listen-address=unix:///var/run/kmsplugin/kms.sock
+        - -vault-address=https://vault.example.com
+        - -transit-mount=
+        - -transit-key=test-transit-key
+        - -approle-role-id=test-role-id
+        - -approle-secret-id-path=/var/run/secrets/kms-plugin/kms-plugin-secret-vault-approle-secret_secret-id-1
+        - -tls-ca-file=/var/run/secrets/kms-plugin/kms-plugin-configmap-vault-ca-bundle_ca-bundle.crt-1
+        - -metrics-port=0
+      imagePullPolicy: IfNotPresent
+      restartPolicy: Always
+      terminationMessagePolicy: FallbackToLogsOnError
+      resources:
+        requests:
+          memory: 64Mi
+          cpu: 10m
+      securityContext:
+        allowPrivilegeEscalation: false
+        capabilities:
+          drop:
+            - ALL
+        readOnlyRootFilesystem: true
+        seccompProfile:
+          type: RuntimeDefault
+      volumeMounts:
+        - name: kms-plugin-socket
+          mountPath: /var/run/kmsplugin
+        - name: kms-plugins-data
+          mountPath: /var/run/secrets/kms-plugin
+          readOnly: true
   containers:
     - name: kms-preflight-check
       image: quay.io/openshift-release-dev/ocp-v5.0-art-dev@sha256:test
@@ -75,12 +113,101 @@ spec:
   volumes:
     - name: kms-plugin-socket
       emptyDir: {}
+    - name: kms-plugins-data
+      secret:
+        secretName: kms-preflight-encryption-config
 `
+
+func testReferenceDataObjects(t *testing.T) []runtime.Object {
+	t.Helper()
+
+	return []runtime.Object{
+		&corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "vault-approle-secret",
+				Namespace: testReferenceDataNS,
+			},
+			Data: map[string][]byte{
+				"role-id":   []byte("test-role-id"),
+				"secret-id": []byte("test-secret-id"),
+			},
+		},
+		&corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "vault-ca-bundle",
+				Namespace: testReferenceDataNS,
+			},
+			Data: map[string]string{
+				"ca-bundle.crt": "test-ca-cert",
+			},
+		},
+	}
+}
+
+func testPreflightEncryptionConfigSecret(t *testing.T) *corev1.Secret {
+	t.Helper()
+
+	var secretData encryptiondata.KMSPluginsReferenceData
+	if err := secretData.SetFromRawKey("1", "vault-approle-secret_role-id", []byte("test-role-id")); err != nil {
+		t.Fatalf("failed to set secret reference data: %v", err)
+	}
+	if err := secretData.SetFromRawKey("1", "vault-approle-secret_secret-id", []byte("test-secret-id")); err != nil {
+		t.Fatalf("failed to set secret reference data: %v", err)
+	}
+
+	var configMapData encryptiondata.KMSPluginsReferenceData
+	if err := configMapData.SetFromRawKey("1", "vault-ca-bundle_ca-bundle.crt", []byte("test-ca-cert")); err != nil {
+		t.Fatalf("failed to set configmap reference data: %v", err)
+	}
+
+	config := testPreflightEncryptionConfigFromData(t, secretData, configMapData)
+	secret, err := encryptiondata.ToSecret(testNamespace, preflightEncryptionConfigSecretName, config)
+	if err != nil {
+		t.Fatalf("failed to build encryption config secret: %v", err)
+	}
+	return secret
+}
+
+func testPreflightEncryptionConfigFromData(
+	t *testing.T,
+	secretData encryptiondata.KMSPluginsReferenceData,
+	configMapData encryptiondata.KMSPluginsReferenceData,
+) *encryptiondata.Config {
+	t.Helper()
+
+	return &encryptiondata.Config{
+		Encryption: &apiserverconfigv1.EncryptionConfiguration{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "EncryptionConfiguration",
+				APIVersion: "apiserver.config.k8s.io/v1",
+			},
+			Resources: []apiserverconfigv1.ResourceConfiguration{{
+				Resources: []string{"secrets"},
+				Providers: []apiserverconfigv1.ProviderConfiguration{{
+					KMS: &apiserverconfigv1.KMSConfiguration{
+						APIVersion: "v2",
+						Name:       "1_secrets",
+						Endpoint:   kmsSocketEndpoint,
+						Timeout:    &metav1.Duration{Duration: testDeployerTimeout},
+					},
+				}, {
+					Identity: &apiserverconfigv1.IdentityConfiguration{},
+				}},
+			}},
+		},
+		KMSPlugins: map[string]configv1.KMSPluginConfig{
+			"1": encryptiontesting.DefaultKMSPluginConfig,
+		},
+		KMSPluginsSecretData:    secretData,
+		KMSPluginsConfigMapData: configMapData,
+	}
+}
 
 func newTestDeployer(t *testing.T, objects ...runtime.Object) (*PodPreflightDeployer, *fake.Clientset) {
 	t.Helper()
 
-	kubeClient := fake.NewSimpleClientset(objects...)
+	allObjects := append(testReferenceDataObjects(t), objects...)
+	kubeClient := fake.NewSimpleClientset(allObjects...)
 	deployer := NewPodPreflightDeployer(
 		testNamespace,
 		kubeClient.CoreV1(),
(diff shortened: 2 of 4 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -24,14 +24,14 @@ import (
 )
 
 type kmsConfigHasher struct {
-	provider   kmsProviderConfig
+	provider   KMSProviderConfig
```
ardaguclu: Instead of exposing this, can't we move the buildephemeralencryptionconfig function here?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -5,19 +5,27 @@ import (
...
+	preflightPodName            = "kms-preflight"
+	preflightCheckContainerName = "kms-preflight-check"
+	kmsSocketEndpoint           = "unix:///var/run/kmsplugin/kms.sock"
 )
 
 type KMSPreflightDeployer interface {
-	// Deploy creates the preflight checker with the given config hash
-	Deploy(ctx context.Context, configHash string) error
+	// Deploy creates the preflight checker with the given config hash and KMS plugin configuration.
+	Deploy(ctx context.Context, configHash string, pluginConfig configv1.KMSPluginConfig) error
```
ardaguclu: and pass encryptionconfig name in the namespace here?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,7 +57,27 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
+		return fmt.Errorf("failed to build preflight encryption config secret: %w", err)
+	}
+
+	recorder := events.NewInMemoryRecorder("kms-preflight-deployer", clock.RealClock{})
+	if _, _, err := resourceapply.ApplySecret(ctx, d.coreClient, recorder, encryptionConfigSecret); err != nil {
+		return fmt.Errorf("failed to apply preflight encryption config secret: %w", err)
+	}
+
+	err = pluginlifecycle.NewKMSPluginBuilder().
+		FromEncryptionConfig(preflightEncryptionConfigSecretName, encryptionConfig).
```
ardaguclu: builder only needs an existed secret name

pkg/operator/encryption/kms/preflight/encryption_config.go
```diff
@@ -0,0 +1,98 @@
...
+		KMSPlugins: map[string]configv1.KMSPluginConfig{
+			preflightEncryptionKeyID: pluginConfig,
+		},
+		KMSPluginsSecretData:    secretData,
+		KMSPluginsConfigMapData: configMapData,
+	}, nil
+}
+
+func ephemeralEncryptionConfigSecret(namespace string, cfg *encryptiondata.Config) (*corev1.Secret, error) {
+	return encryptiondata.ToSecret(namespace, preflightEncryptionConfigSecretName, cfg)
```
tjungblu: This may also need a flag to omit the finalizer, otherwise we won't be able to delete it easily

ardaguclu: Manually deleting the finalizer from the value returned from ToSecret might work?

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: that's sensible

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: the other comment I understood, I've moved the ephemeral builder to the other package. What should I replace this with?

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: that's fair, but the pods needs a real secret to run and I need to create it. What should I change here?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -31,12 +42,13 @@ type PodPreflightDeployer struct {
...
 	// In addition to that running the preflight check is not a very frequent operation.
 	coreClient      corev1client.CoreV1Interface
+	eventRecorder   events.Recorder
 	operatorImage   string
 	operatorCommand []string
 	kmsCallTimeout  time.Duration
 }
 
-func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) error {
+func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, pluginConfig configv1.KMSPluginConfig) error {
```
ardaguclu: Logical flow in that function makes sense to me. I think, we just may need to agree upon the how we align built-in functions.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +62,24 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
 	}
 
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	secret, err := encryptiondata.ToSecret(d.namespace, preflightEncryptionConfigSecretName, encryptionConfig)
+	if err != nil {
+		return fmt.Errorf("failed to generate proposed encryption config secret: %w", err)
+	}
+
+	err = pluginlifecycle.NewKMSPluginBuilder().
+		FromEncryptionConfig(secret.Name, encryptionConfig).
```
ardaguclu: fyi: I'm planning to simplify this to `FromEncryptionConfig(secret.Name)` by internally calling fetchEncryptionConfig. If this happens, `Deploy` can get Secret rather than encryptiondata.Config.

ardaguclu: For now, this makes sense though.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -70,6 +97,34 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
 	if err != nil && !apierrors.IsNotFound(err) {
 		return fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err)
 	}
+
+	if err := d.cleanupEncryptionConfigSecret(ctx); err != nil {
```
p0lyn0mial: `return d.cleanupEncryptionConfigSecret(ctx)` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -3,21 +3,27 @@ package preflight
...
+	preflightPodName                    = "kms-preflight"
+	preflightCheckContainerName         = "kms-preflight-check"
+	preflightEncryptionConfigSecretName = "kms-preflight-encryption-config"
 )
 
 type KMSPreflightDeployer interface {
-	// Deploy creates the preflight checker with the given config hash
-	Deploy(ctx context.Context, configHash string) error
+	// Deploy creates the preflight checker with the given config hash and the encryption config.
+	Deploy(ctx context.Context, configHash string, encryptionConfig *encryptiondata.Config) error
```
p0lyn0mial: didn't we want to pass `encryptionConfig *core.Secret` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +62,24 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
 		return fmt.Errorf("failed to generate preflight pod template: %w", err)
 	}
 
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	secret, err := encryptiondata.ToSecret(d.namespace, preflightEncryptionConfigSecretName, encryptionConfig)
```
p0lyn0mial: could we create a copy of the passed config which wouldn't need finalizer stripping and namespace/name changes ? something like ?

```
preflightSecret := &corev1.Secret{
              Name:      preflightEncryptionConfigSecretName,
              Namespace: d.namespace,
          },
          Data: encryptionConfigSecret.Data,
}
```

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +62,24 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
+	err = pluginlifecycle.NewKMSPluginBuilder().
+		FromEncryptionConfig(secret.Name, encryptionConfig).
+		// TODO(thomas): add a "WithPreflight" option to fill the socket names and other arguments (like health cmd)
+		Apply(&pod.Spec, preflightCheckContainerName)
+	if err != nil {
+		return fmt.Errorf("failed to apply preflight plugin: %w", err)
+	}
 
 	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
-	return err
```
p0lyn0mial, line 47: `return err` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -70,6 +97,34 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
...
+func (d *PodPreflightDeployer) cleanupEncryptionConfigSecret(ctx context.Context) error {
+	secret, err := d.coreClient.Secrets(d.namespace).Get(ctx, preflightEncryptionConfigSecretName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s/%s: %w", d.namespace, preflightEncryptionConfigSecretName, err)
+	}
+
+	if idx := slices.Index(secret.Finalizers, secrets.EncryptionSecretFinalizer); idx > -1 {
```
p0lyn0mial: if we do https://github.com/openshift/library-go/pull/2321/changes#r3465827847 then this method could be simplified.

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: if we temporary need `encryptionConfig` then we could use `encryptiondata.FromSecret` for now.

p0lyn0mial (commented):
did a first pass. 
haven't looked at the tests.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: applied

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: applied

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: used the deep copy

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: applied

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: yep, removed

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,7 +59,26 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	// strip any finalizers, because we also want to delete it later
+	encryptionConfigSecret.Finalizers = []string{}
+	_, err = d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create preflight config secret: %w", err)
+	}
+
+	encryptionConfig, err := encryptiondata.FromSecret(encryptionConfigSecret)
```
p0lyn0mial: ok, this is a temp situation, right ?

ardaguclu: I think, yes

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,7 +59,26 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
 		return fmt.Errorf("failed to generate preflight pod template: %w", err)
 	}
 
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	// strip any finalizers, because we also want to delete it later
+	encryptionConfigSecret.Finalizers = []string{}
+	_, err = d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
```
p0lyn0mial: If the secret is created but pod creation fails the caller would need to know to call Cleanup even though Deploy failed.

We could create the pod first (it'll stay Pending until the secret is created), then create the secret with an ownerReference for the pod. 

This would also simplify the cleanup logic. GC deletes the secret automatically when the pod is deleted.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: I guess so

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +59,43 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
+	encryptionConfigSecret.Finalizers = []string{}
+	encryptionConfigSecret.OwnerReferences = []metav1.OwnerReference{{
+		APIVersion: "v1",
+		Kind:       "Pod",
+		Name:       createdPod.Name,
+		UID:        createdPod.UID,
+	}}
+	_, err = d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
+	if err != nil {
+		deleteErr := d.coreClient.Pods(d.namespace).Delete(ctx, preflightPodName, metav1.DeleteOptions{})
```
tjungblu: that could also be `Cleanup()` - wdyt @p0lyn0mial ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +59,43 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
-	return err
+	createdPod, err := d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create preflight pod: %w", err)
+	}
+
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	// strip any finalizers, because garbage collection should be able to delete it with the pod
+	encryptionConfigSecret.Finalizers = []string{}
+	encryptionConfigSecret.OwnerReferences = []metav1.OwnerReference{{
```
p0lyn0mial: What happens if the Pod is deleted but the Secret isn't? The controller starts again and calls deploy. The Pod is created, but Secret creation fails. Will the Pod still start and use the old/outdated Secret?

Would BlockOwnerDeletion + foreground deletion help ?

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: yeah, we don't need it.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +59,43 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
+		FromEncryptionConfig(encryptionConfigSecret.Name, encryptionConfig).
+		// TODO(thomas): add a "WithPreflight" option to fill the socket names and other arguments (like health cmd)
+		Apply(&pod.Spec, preflightCheckContainerName)
+	if err != nil {
+		return fmt.Errorf("failed to apply preflight plugin: %w", err)
+	}
 
-	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
-	return err
+	createdPod, err := d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
```
p0lyn0mial: In the future we will need an SA for the pod. Can a pod be created without an SA ? If not could we use a default SA ?

tjungblu: I'm using the default SA for the namespace at the minute.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: called cleanup directly

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: added foreground deletion

tjungblu:
@coderabbitai resume

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -2,24 +2,37 @@ package preflight
...
-	// Deploy creates the preflight checker with the given config hash
-	Deploy(ctx context.Context, configHash string) error
+	// Deploy creates the preflight pod and encryption config secret.
+	//
+	// Deploy is idempotent with respect to the preflight pod and secret: it calls Cleanup
+	// before creating resources, so repeated calls or controller retries replace any
+	// previous preflight run.
+	//
+	// Deploy is fail-safe on partial failure: if secret creation fails after the pod was
+	// created, Cleanup is attempted before the error is returned.
```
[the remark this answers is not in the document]
tjungblu: that's why Deploy is idempotent, you can just retry that error

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -49,10 +73,35 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	encryptionConfig, err := encryptiondata.FromSecret(encryptionConfigSecret)
+	if err != nil {
+		return fmt.Errorf("failed to generate proposed encryption config secret: %w", err)
+	}
+
+	err = pluginlifecycle.NewKMSPluginBuilder().
+		FromEncryptionConfig(preflightEncryptionConfigSecretName, encryptionConfig).
+		// TODO(thomas): add a "WithPreflight" option to fill the socket names and other arguments (like health cmd)
+		Apply(&pod.Spec, preflightCheckContainerName)
```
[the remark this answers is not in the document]
tjungblu: added a nil check and a unit test

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -28,7 +33,37 @@ type PodPreflightDeployer struct {
...
-func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) error {
+func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, encryptionConfigSecret *corev1.Secret) error {
+	if configHash == "" {
+		return fmt.Errorf("configHash is empty")
+	}
+	if encryptionConfigSecret == nil {
+		return fmt.Errorf("encryptionConfigSecret is nil")
+	}
+
+	encryptionConfig, err := encryptiondata.FromSecret(encryptionConfigSecret)
```
ardaguclu: We don't need this anymore?. Just modify the encryptionConfigSecret and create it?

tjungblu: yes Arda, sorry I'm still rebasing this correctly :D

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -28,7 +32,29 @@ type PodPreflightDeployer struct {
...
+	if err := d.Cleanup(ctx); err != nil {
+		return fmt.Errorf("failed to clean up existing preflight resources: %w", err)
+	}
+
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	encryptionConfigSecret.Name = preflightEncryptionConfigSecretName
+	encryptionConfigSecret.Namespace = d.namespace
+	// strip any finalizers, because we also want to delete it later
+	encryptionConfigSecret.Finalizers = []string{}
+	_, err := d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
ardaguclu: Nice catch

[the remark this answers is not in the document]
tjungblu: fixed

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -91,30 +223,106 @@ func newTestDeployer(t *testing.T, objects ...runtime.Object) (*PodPreflightDepl
...
+	t.Helper()
+
+	for _, action := range actions {
+		if action.Matches(verb, resource) {
+			t.Fatalf("unexpected %s %s action: %#v", verb, resource, action)
+		}
+	}
+}
+
+func assertSecretExists(t *testing.T, kubeClient *fake.Clientset, namespace, name string) {
```
ardaguclu: nit: Instead of having one liner function, would it be better to GET every time?

pkg/operator/encryption/kms/preflight/deployer_test.go
```diff
@@ -75,12 +113,106 @@ spec:
...
+				Namespace: testReferenceDataNS,
+			},
+			Data: map[string]string{
+				"ca-bundle.crt": "test-ca-cert",
+			},
+		},
+	}
+}
+
+func testPreflightEncryptionConfig(t *testing.T) *encryptiondata.Config {
```
ardaguclu: Do we need this and its utility functions?. Can't we directly return Secret?

ardaguclu:
Dropped a few comments about the test changes. But core logic looks good to me.
/lgtm
/hold
I think we will need another round of review from @p0lyn0mial

pkg/operator/encryption/kms/preflight/deployer_test.go
tjungblu: I'm merging this with its caller, there is no reuse here

pkg/operator/encryption/kms/preflight/deployer_test.go
tjungblu: that makes sense, there is not much here. It's also redundant to the mock action recording

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -41,10 +67,20 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
...
+		WithSecretRequired().
+		FromEncryptionConfigSecret(d.namespace, preflightEncryptionConfigSecretName, d.coreClient).
+		Apply(ctx, &pod.Spec, preflightCheckContainerName)
+	if err != nil {
+		return fmt.Errorf("failed to apply preflight plugin: %w", err)
+	}
 
 	_, err = d.coreClient.Pods(d.namespace).Create(ctx, pod, metav1.CreateOptions{})
-	return err
+	if err != nil {
```
p0lyn0mial: could we simplify to `return err` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -41,10 +67,20 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string) er
 		return fmt.Errorf("failed to generate preflight pod template: %w", err)
 	}
 
-	// TODO(thomas): inject KMS plugin sidecar container into pod.Spec
+	err = pluginlifecycle.NewKMSPluginBuilder().
```
p0lyn0mial, line 68: I though we were planning to expose a function instead of using the builder. I guess we could to that in a new PR.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -28,7 +32,29 @@ type PodPreflightDeployer struct {
...
+	// ensure that there is nothing left over from previous runs
+	if err := d.Cleanup(ctx); err != nil {
+		return fmt.Errorf("failed to clean up existing preflight resources: %w", err)
+	}
+
+	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
+	// rewrite the entire ObjectMeta to avoid copying resource versions or managed fields
+	encryptionConfigSecret.ObjectMeta = metav1.ObjectMeta{Namespace: d.namespace, Name: preflightEncryptionConfigSecretName}
+	// strip any finalizers, because we also want to delete it later
+	encryptionConfigSecret.Finalizers = []string{}
```
p0lyn0mial: is this required if we already created a new `ObjectMeta` (line 50) ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -58,11 +94,19 @@ func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, er
 }
 
 func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
+	var errs []error
+
 	err := d.coreClient.Pods(d.namespace).Delete(ctx, preflightPodName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
```
p0lyn0mial, line 98: could it be simplified to `!apierrors.IsNotFound(err)` ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -58,11 +94,19 @@ func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, er
...
-		return fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err)
+		errs = append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err))
 	}
-	return nil
+
+	err = d.coreClient.Secrets(d.namespace).Delete(ctx, preflightEncryptionConfigSecretName, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
```
p0lyn0mial, line 103: could it be simplified to `!apierrors.IsNotFound(err)` ?

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: nope, let's leave this out

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: applied

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: what's the point of using a builder if we're not using it?

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: you mean skipping the nil check? I don't think that IsNotFound is nil safe

p0lyn0mial: yes, ok.

PR #2322, opened 2026-06-22
encryption/controllers/kms_preflight_controller: read EncryptionKMSPreflightRequired condition and compare config hash

p0lyn0mial:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved KMS configuration preflight behavior with a deterministic hash-based protocol to coordinate readiness and verify configuration consistency before preflight actions run.

* **Tests**
  * Expanded and refactored KMS preflight and hashing test coverage using shared fixtures, clearer scenario expectations, and stronger assertions for operator condition outcomes and error messaging.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -12,6 +12,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
@@ -123,6 +124,7 @@ type kmsPreflightController struct {
 
 	operatorClient  operatorv1helpers.OperatorClient
 	apiServerClient configv1client.APIServerInterface
+	coreClient      corev1client.CoreV1Interface
 
 	provider                 Provider
 	preconditionsFulfilledFn preconditionsFulfilled
@@ -204,6 +206,11 @@ func NewKMSPreflightController(
 	operatorClient operatorv1helpers.OperatorClient,
 	apiServerClient configv1client.APIServerInterface,
 	apiServerInformer configv1informers.APIServerInformer,
+	// coreClient reads referenced Secrets and ConfigMaps in openshift-config for hash
+	// computation. No informer is needed: the key-controller detects config changes and
+	// posts a new EncryptionKMSPreflightRequired condition, which triggers this controller
+	// via the operatorClient informer. The minute-based resync covers the rest.
+	coreClient corev1client.CoreV1Interface,
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &kmsPreflightController{
@@ -211,6 +218,7 @@ func NewKMSPreflightController(
 
 		operatorClient:  operatorClient,
 		apiServerClient: apiServerClient,
+		coreClient:      coreClient,
 
 		provider:                 provider,
 		preconditionsFulfilledFn: preconditionsFulfilledFn,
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	requiredCondition := operatorv1helpers.FindOperatorCondition(operatorStatus.Conditions, "EncryptionKMSPreflightRequired")
+	if requiredCondition == nil || requiredCondition.Status != operatorv1.ConditionTrue {
+		return "", nil
+	}
+	apiServer, err := c.apiServerClient.Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get apiserver config: %w", err)
+	}
+
+	providerCfg, err := newKMSProviderConfig(apiServer.Spec.Encryption.KMS)
```
ardaguclu, line 303: @tjungblu we have accessed providerCfg. This can be used for encryption-config.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	if err != nil {
+		return "", fmt.Errorf("failed to compute KMS config hash: %w", err)
+	}
+
+	requiredHash := requiredCondition.Message
+	// No requeue needed: the key-controller will post an updated condition when it
+	// picks up the config change (via apiServerInformer), which triggers us through
+	// operatorClient.Informer(). The minute-based resync is a backstop.
+	if currentHash != requiredHash {
+		klog.V(4).Infof("KMS config hash changed: required=%s, current=%s; waiting for the key-controller to post an updated condition", requiredHash, currentHash)
```
ardaguclu, line 317: It might make sense adding a TODO about we should cleanup the running pods. They are obsolete now.

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: maybe @p0lyn0mial can make up his mind on what he wants to pass down to the Deployer. I'll take whatever works.

p0lyn0mial: not sure if i understand your comment.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 }
 
 func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) error {
-	return fmt.Errorf("implement me")
+	requiredHash, err := c.preflightRequired(ctx)
```
tjungblu: don't you need to pass the config along that matches this hash here? the deployer can't deploy from the hash itself

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu, line 317: can we record an event for this?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	if requiredHash == "" {
+		return nil
+	}
+
+	return fmt.Errorf("preflight checks not yet implemented for hash %s", requiredHash)
+}
+
+// preflightRequired returns the config hash that needs preflight validation,
+// or an empty string when no preflight is needed.
+func (c *kmsPreflightController) preflightRequired(ctx context.Context) (string, error) {
```
tjungblu: this reads like it will return a bool, `requiredPreflightHash` maybe? 

depends on whether you intend to return the config along

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	_, operatorStatus, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
```
tjungblu, line 291: you mention 
> which triggers this controller ... via the operatorClient informer.

earlier, why use a live client here?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: I think our plan for the first iteration was to "just" deploy the preflight pod. The next iteration will focus on wiring the configuration.

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: I guess we can resolve, since we had the meeting earlier today

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: sounds good

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: it is a standard client used in the encryption controllers. 
have a look at https://github.com/openshift/library-go/blob/3fc9e9752e67859639680f340eb8c000a3d5012d/pkg/operator/encryption/controllers.go#L32

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: i will rename to `requiredPreflightHash`

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: I don't want to record an event here. In general, recording events is problematic for operations or functions that can be executed more than once. There is no guarantee that this function will only be executed once, which means it could produce multiple events. We could add internal state to the controller, but that state would be lost if the operator restarts.

p0lyn0mial: the cleanup logic will be implemented in the runPreflightChecks function.
please have a look at https://github.com/openshift/library-go/pull/2327/changes#diff-dc1a42a37fd3b2abff10ac35964a76a818a7556d30736d5e12099b3494beef04R463

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -265,5 +273,50 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+		return "", fmt.Errorf("failed to compute KMS config hash: %w", err)
+	}
+
+	requiredHash := requiredCondition.Message
+	// No requeue needed: the key-controller will post an updated condition when it
+	// picks up the config change (via apiServerInformer), which triggers us through
+	// operatorClient.Informer(). The minute-based resync is a backstop.
+	if currentHash != requiredHash {
+		klog.V(4).Infof("KMS config hash changed: required=%s, current=%s; waiting for the key-controller to post an updated condition", requiredHash, currentHash)
```
ardaguclu, line 317: nit: This is pretty useful log, it might make sense decreasing the log level to 2 from 4.

ardaguclu (commented):
/lgtm

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: seems reasonable. i will update the code in a followup. thx.

ardaguclu:
@tjungblu I think you can override this one too

tjungblu:
I wish I can, @p0lyn0mial can you?

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: I think it would be useful for debugging in CI timeline. I also don't see how event duplication is a concern here.

ardaguclu:
let's leave it running then :)

PR #2330, opened 2026-06-25
CNTRLPLANE-3363: Internalize encryption config Secret retrieval in Builder

ardaguclu:
After this refactoring, preflight checker simply call this;

```go
_, err := d.secretClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})

err = pluginlifecycle.NewKMSPluginBuilder(d.secretClient).
      FromEncryptionConfig(d.namespace, preflightEncryptionConfigSecretName).
      Apply(ctx, podSpec, preflightCheckContainerName)
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved KMS sidecar injection to load the encryption-config from the cluster at execution time rather than using pre-parsed data.
  * If the encryption-config Secret is missing or resolves to no usable configuration, injection now safely performs a no-op instead of failing.
  * Maintains consistent, idempotent results regardless of builder call order, including for static pod handling.
* **Tests**
  * Updated KMS plugin builder and health reporter tests to match the new secret-loading and `Apply` control flow.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/builder.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/builder.go
@@ -18,25 +18,28 @@ import (
 // KMSPluginBuilder constructs KMS plugin pod spec contributions for injection
 // into API server pods.
 type KMSPluginBuilder struct {
-	encryptionConfig           *encryptiondata.Config
+	secretClient               corev1client.SecretsGetter
+	encryptionConfigNamespace  string
 	encryptionConfigSecretName string
 	staticPod                  bool
+	errorIfNotFound            bool
 
 	healthReporter *healthReporter
 }
 
 // NewKMSPluginBuilder creates a builder that defaults to deployment mode.
-func NewKMSPluginBuilder() *KMSPluginBuilder {
-	return &KMSPluginBuilder{}
+// The secretClient is used to fetch the encryption-config Secret at Apply time.
+func NewKMSPluginBuilder(secretClient corev1client.SecretsGetter) *KMSPluginBuilder {
+	return &KMSPluginBuilder{
+		secretClient: secretClient,
+	}
 }
 
-// FromEncryptionConfig loads all KMS plugins from a parsed encryption config.
-// The encryptionConfigSecretName identifies the Secret the config was parsed
-// from; it is used for volume configuration in both deployment and static pod
-// modes.
-func (b *KMSPluginBuilder) FromEncryptionConfig(encryptionConfigSecretName string, cfg *encryptiondata.Config) *KMSPluginBuilder {
-	b.encryptionConfigSecretName = encryptionConfigSecretName
-	b.encryptionConfig = cfg
+// FromEncryptionConfig configures the builder to load KMS plugins from the
+// named encryption-config Secret. The Secret is fetched at Apply time.
+func (b *KMSPluginBuilder) FromEncryptionConfig(namespace, secretName string) *KMSPluginBuilder {
+	b.encryptionConfigNamespace = namespace
+	b.encryptionConfigSecretName = secretName
 	return b
 }
 
@@ -44,6 +47,14 @@ func (b *KMSPluginBuilder) FromEncryptionConfig(encryptionConfigSecretName strin
 // data from the resource-dir volume and run as root (UID 0).
 func (b *KMSPluginBuilder) AsStaticPod() *KMSPluginBuilder {
 	b.staticPod = true
+	return b
+}
+
+// ErrorIfNotFound makes Apply return an error when the encryption-config Secret
+// does not exist, instead of silently treating it as a no-op. This is useful for
+// callers like the preflight checker that expect the Secret to be present.
+func (b *KMSPluginBuilder) ErrorIfNotFound() *KMSPluginBuilder {
+	b.errorIfNotFound = true
 	return b
 }
 
@@ -66,7 +77,7 @@ func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage stri
 //
 // It is a no-op (returns nil error) when no KMS plugins are found.
 // It is idempotent.
-func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string) error {
+func (b *KMSPluginBuilder) Apply(ctx context.Context, podSpec *corev1.PodSpec, containerName string) error {
 	if podSpec == nil {
 		return fmt.Errorf("pod spec cannot be nil")
 	}
@@ -74,7 +85,19 @@ func (b *KMSPluginBuilder) Apply(podSpec *corev1.PodSpec, containerName string) 
 		return fmt.Errorf("container name cannot be empty")
 	}
 
-	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(b.encryptionConfig)
+	if b.secretClient == nil {
+		return fmt.Errorf("secret client cannot be nil")
+	}
+
+	cfg, err := fetchEncryptionConfig(ctx, b.encryptionConfigNamespace, b.encryptionConfigSecretName, b.secretClient, b.errorIfNotFound)
+	if err != nil {
+		return err
+	}
+	if cfg == nil {
+		return nil
+	}
+
+	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(cfg)
 	if err != nil {
 		return fmt.Errorf("failed to get KMS configurations: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -66,15 +68,23 @@ func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage stri
...
 	}
 
-	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(b.encryptionConfig)
+	cfg, err := fetchEncryptionConfig(ctx, b.encryptionConfigNamespace, b.encryptionConfigSecretName, b.secretClient)
+	if err != nil {
+		return err
+	}
+	if cfg == nil {
+		return nil
+	}
```
tjungblu, line 98: I know this is the current behavior (more or less) - but is it desired? 

For example when I call apply and I pass a malformed secret that contains no EncryptionConfig, I would start a preflight that contains no plugin. Of course that will eventually fail, but it would fail with some connectivity errors.

ardaguclu: [This](https://github.com/openshift/library-go/blob/87f06ace10716d2ffe2c4e3b47179b303b311633/pkg/operator/encryption/kms/pluginlifecycle/builder.go#L168) can only happen if encryption is turned off in the cluster (or day 0). If Secret is malformed, it returns error.

tjungblu: That still sounds wacky to me because I can't really discern whether a pod was applied or not in this case, but okay. fetchEncryptionConfig should return the NotFound instead of logging...

I'll add additional validation that expects the initContainer length to change to workaround.

tjungblu: there's also another path that returns (nil,nil) in encryptiondata.FromSecret

ardaguclu: Actually you hit a crucial point. This is obvious from plugin lifecycle. But I would like think this for preflight.

ardaguclu: > there's also another path that returns (nil,nil) in encryptiondata.FromSecret

This is captured here https://github.com/openshift/library-go/blob/87f06ace10716d2ffe2c4e3b47179b303b311633/pkg/operator/encryption/kms/pluginlifecycle/builder.go#L181 and propagated as error

ardaguclu: @tjungblu I've introduced new ErrorIfNotFound builder function. Please let me know your thoughts.

tjungblu: :+1:

pkg/operator/encryption/kms/pluginlifecycle/builder.go
tjungblu, line 200: this also has a path where it could return (nil, nil), e.g. when the expected map key isn't found, but we don't have to solve this right now.

ardaguclu: Is this different than https://github.com/openshift/library-go/pull/2330#discussion_r3472706315?

tjungblu: oh you're right, sorry. Please disregard.

tjungblu:
/lgtm

in case anyone else wants to review it, otherwise feel free to cancel @ardaguclu 
/hold

ardaguclu:
Thank you. It is better to get another review from @bertinatto too

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -66,15 +68,23 @@ func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage stri
...
 	if podSpec == nil {
 		return fmt.Errorf("pod spec cannot be nil")
 	}
 	if containerName == "" {
 		return fmt.Errorf("container name cannot be empty")
 	}
 
-	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(b.encryptionConfig)
+	cfg, err := fetchEncryptionConfig(ctx, b.encryptionConfigNamespace, b.encryptionConfigSecretName, b.secretClient)
+	if err != nil {
```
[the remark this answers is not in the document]
ardaguclu: Latest version checks the emptiness of secretClient?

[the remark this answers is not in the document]
ardaguclu: Doesn't kube-apiserver handle this?

ardaguclu:
/hold cancel
I'll fix any requested changes in followup PR

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -18,25 +18,28 @@ import (
...
-// from; it is used for volume configuration in both deployment and static pod
-// modes.
-func (b *KMSPluginBuilder) FromEncryptionConfig(encryptionConfigSecretName string, cfg *encryptiondata.Config) *KMSPluginBuilder {
-	b.encryptionConfigSecretName = encryptionConfigSecretName
-	b.encryptionConfig = cfg
+// FromEncryptionConfig configures the builder to load KMS plugins from the
+// named encryption-config Secret. The Secret is fetched at Apply time.
+func (b *KMSPluginBuilder) FromEncryptionConfig(namespace, secretName string) *KMSPluginBuilder {
+	b.encryptionConfigNamespace = namespace
+	b.encryptionConfigSecretName = secretName
```
bertinatto, line 42: this no longer is `FromEncryptionConfig`, but more like `FromEncryptionConfigSecret`?

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -66,15 +77,27 @@ func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage stri
...
 	if containerName == "" {
 		return fmt.Errorf("container name cannot be empty")
 	}
 
-	kmsConfigurations, err := encryptiondata.ExtractUniqueAndSortedKMSConfigurations(b.encryptionConfig)
+	if b.secretClient == nil {
+		return fmt.Errorf("secret client cannot be nil")
+	}
+
+	cfg, err := fetchEncryptionConfig(ctx, b.encryptionConfigNamespace, b.encryptionConfigSecretName, b.secretClient, b.errorIfNotFound)
```
bertinatto, line 92: there seems to be a disconnect because I need a `secretClient` to create a builder, and `encryptionConfigNamespace` and `encryptionConfigSecretName` are not required; however, if when I use `Apply()` all of them are required.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
bertinatto, line 42: maybe secretClient should be passed here?

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -47,6 +50,14 @@ func (b *KMSPluginBuilder) AsStaticPod() *KMSPluginBuilder {
 	return b
 }
 
+// ErrorIfNotFound makes Apply return an error when the encryption-config Secret
+// does not exist, instead of silently treating it as a no-op. This is useful for
+// callers like the preflight checker that expect the Secret to be present.
+func (b *KMSPluginBuilder) ErrorIfNotFound() *KMSPluginBuilder {
```
bertinatto, line 56: nit: `WithSecretRequired`?

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: https://github.com/openshift/library-go/pull/2338

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: Yeah, I think currently Apply indeed requires those fields.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: https://github.com/openshift/library-go/pull/2338

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: https://github.com/openshift/library-go/pull/2338

PR #2331, opened 2026-06-25
CNTRLPLANE-3237: encryption/controllers: move KMSPreflightDeployer interface to the consumer package and document the deployer contract

p0lyn0mial:
https://go.dev/wiki/CodeReviewComments#interfaces

`Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values`

In addition to that the PR documents the contract between the preflight controller and the deployers

## Summary by CodeRabbit

* **Refactor**
  * Updated internal KMS preflight deployment handling for improved structure and maintainability.
  * Preserved existing deployment, status, and cleanup behavior without changing the user experience.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 
@@ -116,6 +117,28 @@ func (h *kmsConfigHasher) hashReferencedConfigMap(ctx context.Context, hasher ha
 		}
 	}
 	return nil
+}
+
+// KMSPreflightDeployer abstracts the lifecycle of a preflight workload that
+// validates KMS plugin configuration before an encryption key is created.
+// All methods are idempotent.
+//
+// Deploy creates all resources the workload needs (ServiceAccount, Secret,
+// RBAC, Pod). The controller only calls Deploy when Status reports no pod
+// exists, so no running pod can be affected.
+//
+// Cleanup removes all resources created by Deploy.
+type KMSPreflightDeployer interface {
+	// Deploy creates the preflight workload with the given config hash
+	// and encryption configuration. It is idempotent.
+	Deploy(ctx context.Context, configHash string, encryptionConfiguration *corev1.Secret) error
+
+	// Status returns the current pod status of the preflight workload.
+	// It returns an apierrors.IsNotFound error when no preflight pod exists.
+	Status(ctx context.Context) (corev1.PodStatus, error)
+
+	// Cleanup removes all resources created by Deploy.
+	Cleanup(ctx context.Context) error
 }
 
 type kmsPreflightController struct {
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -15,14 +15,6 @@ const (
 	preflightPodName = "kms-preflight"
 )
 
-type KMSPreflightDeployer interface {
-	// Deploy creates the preflight checker with the given config hash
-	Deploy(ctx context.Context, configHash string) error
-	Status(ctx context.Context) (corev1.PodStatus, error)
-	// Cleanup cleans up anything previously created.
-	Cleanup(ctx context.Context) error
-}
-
 // PodPreflightDeployer creates a one-shot pod to run the preflight checker as a command,
 // with the plugin injected as a sidecar container.
 type PodPreflightDeployer struct {


ardaguclu:
/lgtm
/hold
in case @tjungblu has any reservations

pkg/operator/encryption/controllers/kms_preflight_deployer.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// If any resource from a previous run still exists, Deploy must return
+// AlreadyExists rather than reuse it — stale resources (e.g. old credentials,
+// wrong CA bundle) are not safe to run against. This can happen when a
+// previous Cleanup partially failed: the pod was removed but a supporting
+// resource was not. The controller handles AlreadyExists by calling Cleanup
+// to remove the leftover resources and retrying Deploy on the next sync.
+type KMSPreflightDeployer interface {
+	// Deploy creates the preflight workload with the given config hash.
+	// It returns AlreadyExists if any resource from a previous run still
```
tjungblu: can you define AlreadyExists here as well?

pkg/operator/encryption/controllers/kms_preflight_deployer.go
```diff
@@ -0,0 +1,44 @@
...
+	// exists, signalling to the controller that Cleanup is needed first.
+	Deploy(ctx context.Context, configHash string) error
```
tjungblu: please add the corev1.Secret as a parameter and document this accordingly

pkg/operator/encryption/controllers/kms_preflight_deployer.go
```diff
@@ -0,0 +1,44 @@
...
+// missing credentials. Cleanup must handle partial state — for example, if
+// a previous Cleanup deleted the pod but failed to remove the Secret, the
+// next Cleanup must remove the remaining Secret.
+//
+// If any resource from a previous run still exists, Deploy must return
+// AlreadyExists rather than reuse it — stale resources (e.g. old credentials,
+// wrong CA bundle) are not safe to run against. This can happen when a
+// previous Cleanup partially failed: the pod was removed but a supporting
+// resource was not. The controller handles AlreadyExists by calling Cleanup
+// to remove the leftover resources and retrying Deploy on the next sync.
```
tjungblu: I still don't get why we can't make Deploy() idempotent instead. It is much more work for you in the controller to manage Cleanup.

p0lyn0mial: Because the controller will have to call Cleanup independently from Deploy, say after some time or when KMS has been disabled.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
p0lyn0mial: I moved the existing interface. I will add a new param in the next PR.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
p0lyn0mial: I don't understand your comment here.

tjungblu: please add the exported error definition in this file as well

pkg/operator/encryption/controllers/kms_preflight_deployer.go
tjungblu: you can call Cleanup whenever you want, that doesn't hinder you from calling Deploy. It just makes your controller logic easier, because you don't have to cleanup before calling deploy.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
tjungblu: why always the additional roundtrips? it doesn't make sense to merge 20 line PRs.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
p0lyn0mial: It is a K8S error. Already defined. The controller will use `apierrors.IsAlreadyExists(err) to check it`

p0lyn0mial: updated the doc to reflect that.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
ibihim: What's the case where erroring out beats converging?

Idempotency is a core k8s idiom. Deviating from it should have a use-case.

p0lyn0mial: > It just makes your controller logic easier, because you don't have to cleanup before calling deploy.

I think in practice that would mean removing the following block:

```
if apierrors.IsAlreadyExists(deployErr) {
  // Stale resources from a previous iteration that did not clean up.
  return true, c.deployer.Cleanup(ctx)
}
```

p0lyn0mial: > I still don't get why we can't make Deploy() idempotent instead.

ok, we can try making the Deploy idempotent. I'm fine with removing the block I showed you in the prev comment.

p0lyn0mial: ok, updated the PR. PTAL.

pkg/operator/encryption/controllers/kms_preflight_deployer.go
p0lyn0mial: updated. I can pass `nil` from my controller for now.

ardaguclu:
All comments seem to be addressed
/lgtm

tjungblu:
/hold cancel

Thanks.

tjungblu:
/override ci/prow/e2e-aws-encryption

there's no need to wait another 3h for the test to finish, this PR has no impact on this

PR #2336, opened 2026-06-26
CNTRLPLANE-2654: added the library functions for workloads

sandeepknd:
Added the library helper functions for workloads. 
Kindly refer this https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030#discussion_r3469132035 for the actual context.

## Summary by CodeRabbit

* **New Features**
  * Added helper support for cordoning/uncordoning Kubernetes nodes by toggling schedulability.
  * Added OLM workload helpers to create/delete OperatorGroups and Subscriptions, including automatic install plan and optional starting CSV.
  * Added utilities to verify CatalogSource availability and to fetch CSV details (name, related images, and phase).
  * Added helpers to read package metadata and select the current CSV from the default channel.
* **Bug Fixes**
  * Avoid redundant cordon/uncordon operations when nodes are already in the requested state.
  * Improved CSV polling to treat “Succeeded” as success and “Failed” as an immediate error.

### DIFF
--- a/test/library/olm/olm.go
+++ b/test/library/olm/olm.go
@@ -1,0 +1,352 @@
+package olm
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	semver "github.com/blang/semver/v4"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+)
+
+// RelatedImage represents an image referenced in a CSV
+type RelatedImage struct {
+	Name  string
+	Image string
+}
+
+// OperatorGroupGVR returns the GroupVersionResource for OperatorGroup
+func OperatorGroupGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1",
+		Resource: "operatorgroups",
+	}
+}
+
+// SubscriptionGVR returns the GroupVersionResource for Subscription
+func SubscriptionGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "subscriptions",
+	}
+}
+
+// CSVGVR returns the GroupVersionResource for ClusterServiceVersion
+func CSVGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "clusterserviceversions",
+	}
+}
+
+// PackageManifestGVR returns the GroupVersionResource for PackageManifest
+func PackageManifestGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "packages.operators.coreos.com",
+		Version:  "v1",
+		Resource: "packagemanifests",
+	}
+}
+
+// CatalogSourceGVR returns the GroupVersionResource for CatalogSource
+func CatalogSourceGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "catalogsources",
+	}
+}
+
+// CreateOperatorGroup creates an OperatorGroup for an operator.
+//
+// This helper accepts unstructured objects to avoid vendoring operator-framework/api
+// types in library-go. Callers in individual operator repositories should:
+// 1. Vendor the operator-framework/api types they need
+// 2. Construct typed OLM objects (e.g., operatorsv1.OperatorGroup)
+// 3. Convert to unstructured using runtime.DefaultUnstructuredConverter.ToUnstructured
+// 4. Pass the unstructured object to this helper
+//
+// Example:
+//
+//	og := &operatorsv1.OperatorGroup{
+//	    ObjectMeta: metav1.ObjectMeta{Name: "my-og", Namespace: "my-ns"},
+//	    Spec: operatorsv1.OperatorGroupSpec{TargetNamespaces: []string{"my-ns"}},
+//	}
+//	unstructuredOG, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(og)
+//	err := CreateOperatorGroup(ctx, client, &unstructured.Unstructured{Object: unstructuredOG})
+func CreateOperatorGroup(ctx context.Context, dynamicClient dynamic.Interface, og *unstructured.Unstructured) error {
+	_, err := dynamicClient.Resource(OperatorGroupGVR()).Namespace(og.GetNamespace()).Create(ctx, og, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create OperatorGroup %s: %w", og.GetName(), err)
+	}
+	return nil
+}
+
+// DeleteOperatorGroup deletes the OperatorGroup and waits for it to be fully removed.
+// See CreateOperatorGroup for details on using unstructured objects.
+func DeleteOperatorGroup(ctx context.Context, dynamicClient dynamic.Interface, og *unstructured.Unstructured) error {
+	err := dynamicClient.Resource(OperatorGroupGVR()).Namespace(og.GetNamespace()).Delete(ctx, og.GetName(), metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete OperatorGroup %s: %w", og.GetName(), err)
+	}
+
+	// Wait for the OperatorGroup to be fully deleted
+	return wait.PollUntilContextCancel(ctx, 1*time.Second, true, func(ctx context.Context) (bool, error) {
+		_, err := dynamicClient.Resource(OperatorGroupGVR()).Namespace(og.GetNamespace()).Get(ctx, og.GetName(), metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+// CreateSubscription creates a Subscription for an operator.
+// See CreateOperatorGroup for details on using unstructured objects.
+func CreateSubscription(ctx context.Context, dynamicClient dynamic.Interface, sub *unstructured.Unstructured) error {
+	_, err := dynamicClient.Resource(SubscriptionGVR()).Namespace(sub.GetNamespace()).Create(ctx, sub, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create Subscription %s: %w", sub.GetName(), err)
+	}
+	return nil
+}
+
+// DeleteSubscription deletes the Subscription and waits for it to be fully removed.
+// See CreateOperatorGroup for details on using unstructured objects.
+func DeleteSubscription(ctx context.Context, dynamicClient dynamic.Interface, sub *unstructured.Unstructured) error {
+	err := dynamicClient.Resource(SubscriptionGVR()).Namespace(sub.GetNamespace()).Delete(ctx, sub.GetName(), metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete Subscription %s: %w", sub.GetName(), err)
+	}
+
+	// Wait for the Subscription to be fully deleted
+	return wait.PollUntilContextCancel(ctx, 1*time.Second, true, func(ctx context.Context) (bool, error) {
+		_, err := dynamicClient.Resource(SubscriptionGVR()).Namespace(sub.GetNamespace()).Get(ctx, sub.GetName(), metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+// CatalogSourceExists checks if required catalog source is available.
+// The catalogSource and catalogSourceNamespace parameters should come from
+// the Subscription's spec.catalogSource and spec.catalogSourceNamespace fields.
+func CatalogSourceExists(ctx context.Context, dynamicClient dynamic.Interface, catalogSource, catalogSourceNamespace string) error {
+	_, err := dynamicClient.Resource(CatalogSourceGVR()).Namespace(catalogSourceNamespace).Get(ctx, catalogSource, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("catalog source %s not found in %s: %w", catalogSource, catalogSourceNamespace, err)
+	}
+
+	return nil
+}
+
+// GetTheLatestCSVName gets the CSV name for the operator using label selector.
+// When multiple CSVs are found, it returns the one with the highest semver version.
+func GetTheLatestCSVName(ctx context.Context, dynamicClient dynamic.Interface, namespace, labelSelector string) (string, error) {
+	csvList, err := dynamicClient.Resource(CSVGVR()).Namespace(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labelSelector,
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to list CSVs: %w", err)
+	}
+
+	if len(csvList.Items) == 0 {
+		return "", fmt.Errorf("no CSV found with label selector: %s", labelSelector)
+	}
+
+	// If only one CSV, return it directly
+	if len(csvList.Items) == 1 {
+		return csvList.Items[0].GetName(), nil
+	}
+
+	// Multiple CSVs found - select the one with the highest semver version
+	var latestCSV *unstructured.Unstructured
+	var latestVersion semver.Version
+
+	for i := range csvList.Items {
+		csv := &csvList.Items[i]
+		versionStr, found, err := unstructured.NestedString(csv.Object, "spec", "version")
+		if err != nil || !found || versionStr == "" {
+			// Skip CSVs without a valid version field
+			continue
+		}
+
+		version, err := semver.Parse(versionStr)
+		if err != nil {
+			// Skip CSVs with invalid semver format
+			continue
+		}
+
+		if latestCSV == nil || version.GT(latestVersion) {
+			latestVersion = version
+			latestCSV = csv
+		}
+	}
+
+	if latestCSV == nil {
+		// Fallback to first CSV if none have valid versions
+		return csvList.Items[0].GetName(), nil
+	}
+
+	return latestCSV.GetName(), nil
+}
+
+// GetCSVRelatedImages gets the relatedImages from a CSV.
+// See CreateOperatorGroup for details on using unstructured objects.
+func GetCSVRelatedImages(csv *unstructured.Unstructured) ([]RelatedImage, error) {
+	// Extract relatedImages from spec using k8s unstructured helper
+	relatedImages, found, err := unstructured.NestedSlice(csv.Object, "spec", "relatedImages")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get relatedImages: %w", err)
+	}
+	if !found {
+		// Empty relatedImages is valid - return empty slice
+		return []RelatedImage{}, nil
+	}
+
+	// Convert to RelatedImage structs using k8s unstructured helpers
+	var images []RelatedImage
+	for _, item := range relatedImages {
+		imgMap, ok := item.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		// Skip entries where name field is missing, has wrong type (err != nil),
+		// doesn't exist (!found), or is empty
+		name, found, err := unstructured.NestedString(imgMap, "name")
+		if err != nil || !found || name == "" {
+			continue
+		}
+
+		// Skip entries where image field is missing, has wrong type (err != nil),
+		// doesn't exist (!found), or is empty
+		image, found, err := unstructured.NestedString(imgMap, "image")
+		if err != nil || !found || image == "" {
+			continue
+		}
+
+		images = append(images, RelatedImage{
+			Name:  name,
+			Image: image,
+		})
+	}
+
+	return images, nil
+}
+
+// BuildSubscriptionFromPackageManifest fetches a packagemanifest for a given package
+// and builds an unstructured Subscription object populated with the default channel, catalog source, and starting CSV information.
+// See CreateOperatorGroup for details on using unstructured objects.
+//
+// The returned unstructured Subscription can be converted to a typed object in the caller:
+//
+//	sub := &operatorsv1alpha1.Subscription{}
+//	runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredSub.Object, sub)
+func BuildSubscriptionFromPackageManifest(ctx context.Context, dynamicClient dynamic.Interface, packageName, namespace string) (*unstructured.Unstructured, error) {
+	if packageName == "" {
+		return nil, fmt.Errorf("packageName cannot be empty")
+	}
+	if namespace == "" {
+		return nil, fmt.Errorf("namespace cannot be empty")
+	}
+
+	pm, err := dynamicClient.Resource(PackageManifestGVR()).Namespace(namespace).Get(ctx, packageName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get packagemanifest %s: %w", packageName, err)
+	}
+
+	// Extract catalog source name
+	catalogSource, found, err := unstructured.NestedString(pm.Object, "status", "catalogSource")
+	if err != nil {
+		return nil, fmt.Errorf("error reading catalogSource: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("catalogSource not found in packagemanifest")
+	}
+
+	// Extract catalog source namespace
+	catalogSourceNamespace, found, err := unstructured.NestedString(pm.Object, "status", "catalogSourceNamespace")
+	if err != nil {
+		return nil, fmt.Errorf("error reading catalogSourceNamespace: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("catalogSourceNamespace not found in packagemanifest")
+	}
+
+	// Extract default channel
+	defaultChannel, found, err := unstructured.NestedString(pm.Object, "status", "defaultChannel")
+	if err != nil {
+		return nil, fmt.Errorf("error reading defaultChannel: %w", err)
+	}
+	if !found {
+		return nil, fmt.Errorf("defaultChannel not found in packagemanifest")
+	}
+
+	// Extract channels to find the latest CSV
+	channels, found, err := unstructured.NestedSlice(pm.Object, "status", "channels")
+	if err != nil || !found {
+		return nil, fmt.Errorf("could not find channels in packagemanifest")
+	}
+
+	var startingCSV string
+	for _, ch := range channels {
+		channel, ok := ch.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		name, found, err := unstructured.NestedString(channel, "name")
+		if err != nil || !found {
+			continue
+		}
+		if name == defaultChannel {
+			startingCSV, found, err = unstructured.NestedString(channel, "currentCSV")
+			if err != nil {
+				return nil, fmt.Errorf("error reading currentCSV from channel %s: %w", name, err)
+			}
+			if !found {
+				return nil, fmt.Errorf("currentCSV not found in default channel %s", name)
+			}
+			break
+		}
+	}
+
+	if startingCSV == "" {
+		return nil, fmt.Errorf("default channel %s not found in packagemanifest channels", defaultChannel)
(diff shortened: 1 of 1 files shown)


test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
+package workloads
```
ingvagabund: `workloads` is too generic. Better to have each set of helpers under its own name. E.g. `test/library/olm` or similar.

test/library/workloads/olm.go
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// RelatedImage represents an image referenced in a CSV
+type RelatedImage struct {
+	Name  string
+	Image string
+}
+
+// GetOperatorGroupGVR returns the GroupVersionResource for OperatorGroup
+func GetOperatorGroupGVR() schema.GroupVersionResource {
```
ingvagabund: `GetOperatorGroupGVR` -> `OperatorGroupGVR` as there's really no "Get" operation performed. The same for other `GetXXX` functions.

test/library/workloads/olm.go
```diff
@@ -0,0 +1,366 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete Subscription %s: %w", sub.Name, err)
+	}
+
+	klog.Infof("Successfully deleted Subscription %s", sub.Name)
+	return nil
+}
+
+// SkipMissingCatalogSources checks if required catalog sources are available
+func (sub *Subscription) SkipMissingCatalogSources(ctx context.Context, dynamicClient dynamic.Interface) error {
```
ingvagabund: Where the "Skip" keyword comes from? The method checks for an existence, it does not skip.

test/library/workloads/node.go
sandeepknd: Is it like 
test/library/olm/olm.go
test/library/node/node.go ?

test/library/olm/olm.go
```diff
@@ -0,0 +1,366 @@
...
+// GetCSVName gets the CSV name for the operator using label selector
+func GetCSVName(ctx context.Context, dynamicClient dynamic.Interface, namespace, labelSelector string) (string, error) {
+	csvList, err := dynamicClient.Resource(GetCSVGVR()).Namespace(namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labelSelector,
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to list CSVs: %w", err)
+	}
+
+	if len(csvList.Items) == 0 {
```
ingvagabund, line 167: What is there are two or more? Is that a valid case?

test/library/workloads/olm.go
```diff
@@ -0,0 +1,366 @@
...
+
+	// Convert to RelatedImage structs using k8s unstructured helpers
+	var images []RelatedImage
+	for _, item := range relatedImages {
+		imgMap, ok := item.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		// Use k8s built-in helpers instead of manual map access
+		name, _, _ := unstructured.NestedString(imgMap, "name")
```
ingvagabund: what if the field does not exists or an error is returned? The same for the other field.

test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
...
+			return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		}
+
+		if node.Spec.Unschedulable {
+			klog.Infof("Node %s is already cordoned", nodeName)
+			return nil
+		}
+
+		node.Spec.Unschedulable = true
+		_, err = client.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
```
ingvagabund: What about `Patch` instead of `Update`?

test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
...
+			return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		}
+
+		if !node.Spec.Unschedulable {
+			klog.Infof("Node %s is already uncordoned", nodeName)
+			return nil
+		}
+
+		node.Spec.Unschedulable = false
+		_, err = client.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
```
ingvagabund: ditto

test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
...
+// Uses retry.RetryOnConflict to handle concurrent node updates by kubelet/controllers.
+func CordonNode(ctx context.Context, client kubernetes.Interface, nodeName string) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node, err := client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		}
+
+		if node.Spec.Unschedulable {
+			klog.Infof("Node %s is already cordoned", nodeName)
```
ingvagabund: Is logging this information useful?

test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
...
+			return nil
+		}
+
+		node.Spec.Unschedulable = true
+		_, err = client.CoreV1().Nodes().Update(ctx, node, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+
+		klog.Infof("Successfully cordoned node %s", nodeName)
```
ingvagabund: Is logging this information useful? Instead, the invoker might check for err == nil and log this if needed.

test/library/workloads/olm.go
```diff
@@ -0,0 +1,366 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
+)
+
+// OperatorGroup represents an OLM OperatorGroup resource configuration
+type OperatorGroup struct {
```
ingvagabund: Are these new types duplicates of already existing definitions? E.g.:
- https://github.com/operator-framework/api/blob/492d6ba7263f792b5a3dcf24569fcd4f7d1c498d/pkg/operators/v1/operatorgroup_types.go#L118
- https://github.com/operator-framework/api/blob/492d6ba7263f792b5a3dcf24569fcd4f7d1c498d/pkg/operators/v1alpha1/subscription_types.go#L31
- etc.

Any way of reusing these? To avoid drifting from them.

test/library/olm/olm.go
```diff
@@ -0,0 +1,366 @@
...
+		Version:  "v1alpha1",
+		Resource: "catalogsources",
+	}
+}
+
+// CreateOperatorGroup creates an OperatorGroup for an operator
+func (og *OperatorGroup) CreateOperatorGroup(ctx context.Context, dynamicClient dynamic.Interface) error {
+	klog.Infof("Creating OperatorGroup %s in namespace %s", og.Name, og.Namespace)
+
+	operatorGroup := &unstructured.Unstructured{
```
ingvagabund: Why not import https://github.com/operator-framework/api/blob/492d6ba7263f792b5a3dcf24569fcd4f7d1c498d/pkg/operators/v1/operatorgroup_types.go#L118 type directly and populate it? There might even be the corresponding clients available? The same for other places.

There's a chance most of the code here can be replaced by the github.com/operator-framework/*** code.

test/library/workloads/node.go
```diff
@@ -0,0 +1,63 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+)
+
+// CordonNode marks a node as unschedulable, preventing new pods from being scheduled on it.
+// This is commonly used during maintenance operations or when testing pod eviction scenarios.
+// Uses retry.RetryOnConflict to handle concurrent node updates by kubelet/controllers.
+func CordonNode(ctx context.Context, client kubernetes.Interface, nodeName string) error {
```
ingvagabund: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/drain/cordon.go#L35.

test/library/workloads/olm.go
sandeepknd: Re: Using github.com/operator-framework/api/pkg/operators/v1 OperatorGroup

 This dependency is currently unavailable in the project. Tests confirm it's not present and would require adding a new dependency:

```
  # Check if dependency exists
  $ grep "operator-framework/api" go.mod go.sum
  # (no output)

  $ go mod why github.com/operator-framework/api
  (main module does not need package github.com/operator-framework/api)

  $ go list -m github.com/operator-framework/api
  go: module github.com/operator-framework/api: not a known dependency

  $ find vendor -type d -name "operator-framework"
  # (no output)
```

  Using the official v1.OperatorGroup type would require:
  - go get github.com/operator-framework/api@<version>
  - go mod tidy && go mod vendor

Are we okay to import dependencies in test directory ?

test/library/olm/olm.go
sandeepknd: same as [here](https://github.com/openshift/library-go/pull/2336#discussion_r3489275623)

test/library/workloads/node.go
sandeepknd: This would need adding new dependencies. 
k8s.io/kubectl/pkg/drain

test/library/olm/olm.go
```diff
@@ -0,0 +1,417 @@
...
+	klog.Infof("Deleting OperatorGroup %s in namespace %s", og.Name, og.Namespace)
+
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 20*time.Second, true, func(ctx context.Context) (bool, error) {
+		err := dynamicClient.Resource(OperatorGroupGVR()).Namespace(og.Namespace).Delete(ctx, og.Name, metav1.DeleteOptions{})
+		if err != nil {
+			klog.Warningf("Failed to delete OperatorGroup, retrying: %v", err)
+			return false, nil
+		}
+		return true, nil
+	})
```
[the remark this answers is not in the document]
sandeepknd: I think its better to keep it simple and avoid unnecessary overhead and complexity. This deletion is mainly going to be invoked in the test cleanup and this cleanup portion gets called after every single individual test, despite placing it in AfterAll(), which would unnecessarily add to the wait time. 

Anyway, once the operator namespace is deleted successfully, it would eventually cleanup all the resources associated with it.

test/library/workloads/node.go
ingvagabund: Actually, you should be able to use `CordonHelper` directly in the KDO repository. Without creating any helpers here.

test/library/workloads/olm.go
ingvagabund: That is ok. The dependency will be vendored only when `test/library/olm` is imported.

test/library/workloads/node.go
sandeepknd: Available k8s.io packages in kdo vendor:
  - k8s.io/api
  - k8s.io/client-go
  - k8s.io/kubernetes
  Unavailable:
  - k8s.io/kubectl 
So, CordonHelper usage will cause new dependency to be downloaded in kdo repo.
```
$ go list -m k8s.io/kubectl
go: module k8s.io/kubectl: can't resolve module using the vendor directory
	(Use -mod=mod or -mod=readonly to bypass.)

```

sandeepknd: Hi @ingvagabund ,
Currently in kdo repo, k8s.io/kubectl is not used anywhere.

ingvagabund: What is your concern?

sandeepknd: This is my current concern. 
https://github.com/openshift/library-go/pull/2336#discussion_r3491209829
If I need to use CordonHelper directly in kdo, we need import the dependency k8s.io/kubectl. Are we good to go with the import of this dep ?

ingvagabund: Yes, it's ok to vendor the extra dependency.

sandeepknd:
The current failure encountered in make verify is due to go version being reverted back to 1.26.3! The go mod tidy keeps upgrading it. This means we cannot use operator-framework/api without upgrading Go, which will cause make verify to fail.

  This is a fundamental blocker. 

  1. operator-framework/api requires Go 1.25.3+ (even v0.40.0)
  2. Upgrading Go causes stricter go vet checks
  3. The stricter checks fail on pre-existing code in pkg/operator/*

  We cannot proceed without either:
  - Fixing the pre-existing go vet issues.
  - Or not using operator-framework/api typed structs at all

ingvagabund:
> This means we cannot use operator-framework/api without upgrading Go

Is it not possible to use less recent version of operator-framework/api?

sandeepknd:
> Is it not possible to use less recent version of operator-framework/api?

After downgrading to operator-framework/api v0.37.0 (requires only Go 1.24.6), Go stayed at 1.25.0 (same as master).
make verify passed.

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+		Group:    "operators.coreos.com",
+		Version:  "v1alpha1",
+		Resource: "catalogsources",
+	}
+}
+
+// CreateOperatorGroup creates an OperatorGroup for an operator
+func CreateOperatorGroup(ctx context.Context, dynamicClient dynamic.Interface, og *OperatorGroup) error {
+	// Convert typed OperatorGroup to unstructured
+	unstructuredOG, err := runtime.DefaultUnstructuredConverter.ToUnstructured(og)
```
sandeepknd: The operator-framework/api package (v0.37.0) provides only type definitions, not typed clients. It's a pure API definition library similar to k8s.io/api. The actual OLM controllers and clients live in the separate operator-lifecycle-manager repository that would require substantial (bulk) dependencies download.

sandeepknd:
Hi @ingvagabund ,
Could you PTAL.

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+)
+
+// OperatorGroup type alias for the actual OLM OperatorGroup type
+type OperatorGroup = operatorsv1.OperatorGroup
```
ingvagabund: Is there a need for the alias?

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+
+// Subscription type alias for the actual OLM Subscription type
+type Subscription = operatorsv1alpha1.Subscription
```
ingvagabund: Is there a need for the alias?

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+
+// RelatedImage represents an image referenced in a CSV
+type RelatedImage struct {
+	Name  string
+	Image string
+}
+
+// OperatorGroupGVR returns the GroupVersionResource for OperatorGroup
+func OperatorGroupGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
```
ingvagabund, line 25: Any way to reuse https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/register.go#L13-L15 or https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/register.go#L19 or other exported symbols from the package?

test/library/olm/olm.go
ingvagabund: What does this comment refer to? Can you share more details?

test/library/olm/olm.go
sandeepknd: No, these type aliases are not strictly necessary. They could be removed and the functions could use the full types directly:
  - *operatorsv1.OperatorGroup instead of *OperatorGroup
  - *operatorsv1alpha1.Subscription instead of *Subscription
However, the type aliases do provide some benefits like - abstraction - if the underlying OLM types change packages, only the alias needs updating. Since its a test library, it does provide the export convenience - callers of this library can use olm.OperatorGroup instead of importing the operator-framework package directly.

test/library/olm/olm.go
sandeepknd: same [here](https://github.com/openshift/library-go/pull/2336#discussion_r3531321723)

test/library/olm/olm.go
sandeepknd: Should it throw an error if multiple CSVs are found 
or
return the newest/latest one ?

sandeepknd:
Hi @ingvagabund ,
Could you PTAL.

test/library/olm/olm.go
sandeepknd: addressed.

test/library/olm/olm.go
ingvagabund: > if the underlying OLM types change packages

Hiding types like this is not a common pattern. Also if the OLM types changes their location it's a one line change that can be easily performed via AI tooling. Please drop the aliases.

test/library/olm/olm.go
ingvagabund: I would assume there can be more than one depending on which operator versions are available. At which case the intent would be to select the most recent version?

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+			return false, fmt.Errorf("CSV %s failed", csvName)
+		}
+
+		return false, nil
+	})
+}
+
+// GetPackageManifest fetches packagemanifest values dynamically for a given package
+// Returns a Subscription struct populated with channel, source, and starting CSV information
+func GetPackageManifest(ctx context.Context, dynamicClient dynamic.Interface, packageName, namespace string) (*Subscription, error) {
```
ingvagabund: The description does not correspond to the output. The signature says get package manifest but returns a subscription

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to delete Subscription %s: %w", sub.Name, err)
+	}
+
+	return nil
+}
+
+// VerifyCatalogSourceExists checks if required catalog source is available
+func VerifyCatalogSourceExists(ctx context.Context, dynamicClient dynamic.Interface, sub *Subscription) error {
```
ingvagabund: VerifyCatalogSourceExists -> CatalogSourceExists

ingvagabund:
@sandeepknd it's also important to keep https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030 updated. It might happen the simplification in this PR might turn the PR void. For the moment can you please focus on duplicating the simplification in https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030 as well? So it's visible what is left and what is useful to reuse.

sandeepknd:
> @sandeepknd it's also important to keep [openshift/cluster-kube-descheduler-operator#2030](https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030) updated. It might happen the simplification in this PR might turn the PR void. For the moment can you please focus on duplicating the simplification in [openshift/cluster-kube-descheduler-operator#2030](https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030) as well? So it's visible what is left and what is useful to reuse.

Updated the test [PR](https://github.com/openshift/cluster-kube-descheduler-operator/pull/2030/changes#diff-8b15de8ee67749892b5cc39a31e1fad911b51f85fbe3a27622ad16f87f568a44R23).

sandeepknd:
Hi @ingvagabund ,
Could you PTAL.

test/library/olm/olm.go
ingvagabund: Still a valid task to remove the aliases

test/library/olm/olm.go
ingvagabund: Group name and version as still used as literals. Not imported from https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/register.go#L13-L15.

test/library/olm/olm.go
ingvagabund: Still not addressed

test/library/olm/olm.go
ingvagabund: Still not addressed

test/library/olm/olm.go
```diff
@@ -0,0 +1,344 @@
...
+// RelatedImage represents an image referenced in a CSV
+type RelatedImage struct {
+	Name  string
+	Image string
+}
+
+// OperatorGroupGVR returns the GroupVersionResource for OperatorGroup
+func OperatorGroupGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "operators.coreos.com",
```
ingvagabund, line 26: Group name and version as still used as literals. Not imported from https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/register.go#L13-L15.

ingvagabund:
@sandeepknd looks like some of my review comments do not show up right away. Please also check all unresolved threads.

test/library/olm/olm.go
sandeepknd: addressed.

test/library/olm/olm.go
sandeepknd: missed out at first.
resolved now.

test/library/olm/olm.go
sandeepknd: resolved now. 
used 
```
Group:    operators.GroupName,
```

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+		Group:    operators.GroupName,
+		Version:  operatorsv1alpha1.GroupVersion,
+		Resource: "clusterserviceversions",
+	}
+}
+
+// PackageManifestGVR returns the GroupVersionResource for PackageManifest
+func PackageManifestGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "packages.operators.coreos.com",
```
sandeepknd, line 53: The reason behind the hard-coded Group is that PackageManifest belongs to a different API group (packages.operators.coreos.com) that doesn't have constants defined in the operator-framework/api package we're using.

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return fmt.Errorf("catalog source %s not found in %s: %w", sub.Spec.CatalogSource, sub.Spec.CatalogSourceNamespace, err)
+	}
+
+	return nil
+}
+
+// GetCSVName gets the CSV name for the operator using label selector.
+// When multiple CSVs are found, it returns the one with the highest semver version.
+func GetCSVName(ctx context.Context, dynamicClient dynamic.Interface, namespace, labelSelector string) (string, error) {
```
ingvagabund: Given the function could be invoked to get the latest CSV it's better to return the whole CSV object.

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+			return false, fmt.Errorf("CSV %s failed", csvName)
+		}
+
+		return false, nil
+	})
+}
+
+// BuildSubscriptionFromPackageManifest fetches a packagemanifest for a given package
+// and builds a Subscription object populated with the default channel, catalog source, and starting CSV information.
+func BuildSubscriptionFromPackageManifest(ctx context.Context, dynamicClient dynamic.Interface, packageName, namespace string) (*operatorsv1alpha1.Subscription, error) {
```
ingvagabund: The function will get more usable if you pass the whole package manifest object. The type is defined at https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/package-server/apis/operators/v1/packagemanifest_types.go#L22.

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+	if latestCSV == nil {
+		// Fallback to first CSV if none have valid versions
+		return csvList.Items[0].GetName(), nil
+	}
+
+	return latestCSV.GetName(), nil
+}
+
+// GetCSVRelatedImages gets the relatedImages from a CSV
+func GetCSVRelatedImages(ctx context.Context, dynamicClient dynamic.Interface, namespace, csvName string) ([]RelatedImage, error) {
```
ingvagabund: ditto. More useful to pass the whole csv object

test/library/olm/olm.go
ingvagabund: s/GetCSVName/GetTheLatestCSVName

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to create Subscription %s: %w", sub.Name, err)
+	}
+
+	return nil
+}
+
+// DeleteSubscription deletes the Subscription
+func DeleteSubscription(ctx context.Context, dynamicClient dynamic.Interface, sub *operatorsv1alpha1.Subscription) error {
```
ingvagabund: Is deleting a subscription waiting for something? Or, how often does it happen the deletion errors?

test/library/olm/olm.go
```diff
@@ -0,0 +1,381 @@
...
+
+	if err != nil {
+		return fmt.Errorf("failed to create OperatorGroup %s: %w", og.Name, err)
+	}
+
+	return nil
+}
+
+// DeleteOperatorGroup deletes the OperatorGroup
+func DeleteOperatorGroup(ctx context.Context, dynamicClient dynamic.Interface, og *operatorsv1.OperatorGroup) error {
```
ingvagabund: Is deleting a group waiting for something? Or, how often does it happen the deletion errors?

test/library/olm/olm.go
ingvagabund: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/package-server/apis/operators/v1/register.go#L12

ingvagabund (commented):
Given the helpers receives either a name or an object and then convert it into unstructured object via `runtime.DefaultUnstructuredConverter.ToUnstructured` it's worth considering to pass only unstructured objects and perform the conversion in individual operator repositories. With a detailed comment about the intention. This way the OLM specific types vendoring can be moved to the operator repositories.

The current suggestion:
- updating the helper signatures to accept and return unstructured objects (with proper comment)
- vendoring the code changes into the KDO repository
- (de)converting objects

test/library/olm/olm.go
sandeepknd: It introduces dependency conflicts and once again updates the go version to Go 1.26.x that creates the same issue which we encountered in "make verify:.

test/library/olm/olm.go
sandeepknd: The same issue is encountered here on importing the types from 
packagesv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1"

```
go mod tidy 2>&1 | head -50)
  ⎿  go: finding module for package github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1
     go: toolchain upgrade needed to resolve github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1
     go: github.com/operator-framework/operator-lifecycle-manager@v0.45.0 requires go >= 1.26.3; switching to go1.26.5

```
which will cause 
```
make verify
```
to fail

sandeepknd:
> Given the helpers receives either a name or an object and then convert it into unstructured object via `runtime.DefaultUnstructuredConverter.ToUnstructured` it's worth considering to pass only unstructured objects and perform the conversion in individual operator repositories. With a detailed comment about the intention. This way the OLM specific types vendoring can be moved to the operator repositories.
> 
> The current suggestion:
> 
> * updating the helper signatures to accept and return unstructured objects (with proper comment)
> * vendoring the code changes into the KDO repository
> * (de)converting objects

Does it mean reverting to the original changes without any of these imports
```
 -  "github.com/operator-framework/api/pkg/operators"                                                                                                                          
     operatorsv1 "github.com/operator-framework/api/pkg/operators/v1"                                                                                                           
     operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1" 
```
?

sandeepknd:
> Given the helpers receives either a name or an object and then convert it into unstructured object via `runtime.DefaultUnstructuredConverter.ToUnstructured` it's worth considering to pass only unstructured objects and perform the conversion in individual operator repositories. With a detailed comment about the intention. This way the OLM specific types vendoring can be moved to the operator repositories.
> 
> The current suggestion:
> 
> * updating the helper signatures to accept and return unstructured objects (with proper comment)
> * vendoring the code changes into the KDO repository
> * (de)converting objects

addressed.
No vendor dependencies needed now, which is the major advantage of leveraging the unstructured k8s client object. I think this is best approach here as it avoids unnecessary dep vendor-ing in library-go which is used in many other core repos.

sandeepknd:
Hi @ingvagabund ,
Could you PTAL.

sandeepknd:
Hi @ingvagabund ,
Could you PTAL.

mrda:
Hi @ingvagabund, can you take a look at this PR?

ropatil010:
Relatable PR: https://github.com/openshift/cluster-kube-descheduler-operator/blob/main/go.mod#L71 got merged.  
/assign @p0lyn0mial 
PTAL look on this PR and help to merge

ingvagabund:
> Relatable PR: https://github.com/openshift/cluster-kube-descheduler-operator/blob/main/go.mod#L71 got merged.

I merged the PR to unblock some of the improvements in the e2e. Yet, still the actual validation of installation of the operator is not finished. Which currently puts reviewing of this PR on hold until fundamental questions are answered. I already shared some of my concerns with @sandeepknd on the Slack.

PR #2337, opened 2026-06-26
Fix: OCPBUGS-93148: installer skip NotReady nodes during static pod revision rollout

dpateriya:
## Summary

- Deprioritizes master nodes that have been Kubernetes-NotReady for more than 10 minutes during static pod revision rollout
- Times out installer pods stuck on NotReady nodes after 15 minutes, marking them as failed so rollout proceeds on healthy masters
- Wires a node informer into the installer controller factory so node readiness changes trigger resync

## Problem

When a master node enters a prolonged NotReady state (e.g., hardware failure), the InstallerController's `nodeToStartRevisionWith` function selects it first (because its static pod isn't reporting Ready). This blocks the entire revision rollout ring, preventing healthy masters from receiving updated configurations.

This is critical during certificate rotation: new certificates are written to etcd but operators cannot roll out new revisions to disk. Static pods on healthy masters continue using old (expired) certificate files, causing cascading authentication failures (401 Unauthorized for all users).

## Root Cause

The InstallerController had no awareness of the Kubernetes `NodeReady` condition. It only checked static pod readiness (via mirror pod status), which is unreportable when a node is down. Combined with `tolerations: [{operator: Exists}]` on installer pods, installations were attempted on unreachable nodes indefinitely.

## Fix Design

**Layer 1: Deprioritization in `nodeToStartRevisionWith`**
All selection loops (TargetRevision, LastFailedRevision, not-ready static pod, wrong-revision, oldest-revision) now skip nodes where `isNodeNotReadyForTooLong` returns true.

**Layer 2: Timeout in `newNodeStateForInstallInProgress`**
If an installer pod has been in a non-terminal phase on a NotReady node for more than 15 minutes, it is deleted and the installation is marked as failed. The normal back-off retry mechanism then applies.

**Backward Compatibility:**
- `WithNodeLister` is opt-in; when `nodeLister` is nil, all checks return false (legacy behavior preserved)
- Existing tests pass with `nil` for the new parameter

## Test Plan

- [x] Unit tests for `isNodeNotReadyForTooLong` (Ready, below threshold, above threshold, Unknown status, no condition, nil lister)
- [x] Unit tests for `nodeToStartRevisionWith` with NotReady skip (single NotReady, multiple NotReady, all NotReady fallback)
- [x] All existing `installer_controller_test.go` tests pass unchanged
- [x] Full `go build ./...` passes
- [ ] Manual validation: deploy on a 3-master cluster, cordon+drain one master, verify revision rollout proceeds on remaining 2 masters

## References

- Bug: https://redhat.atlassian.net/browse/OCPBUGS-93148
- Incident: NotReady master for 11+ days → cert rotation blocked → cascading auth failures on healthy masters

## Summary by CodeRabbit

* **New Features**
  * Enhanced installer rollout ordering to skip nodes that have been unready longer than a configurable threshold.
* **Bug Fixes**
  * Added automatic handling for installer pods stuck on long-unready nodes: the pod is deleted and the installation is marked as failed.
  * Proactively clears stale targets for long-unready nodes and records detailed failure information to avoid blocking healthy rollout.
* **Tests**
  * Added unit tests covering node-readiness thresholds, skip behavior, and stuck-node timeout handling.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 	"k8s.io/utils/ptr"
@@ -44,6 +45,16 @@ const (
 	hostPodManifestDir = "/etc/kubernetes/manifests"
 
 	revisionLabel = "revision"
+
+	// nodeNotReadyThreshold is how long a node must be Kubernetes-NotReady before
+	// the installer controller deprioritizes it in rollout ordering. This prevents
+	// a prolonged hardware failure from blocking cert rotation on healthy masters.
+	nodeNotReadyThreshold = 10 * time.Minute
+
+	// installerPodStuckOnNotReadyNodeTimeout is the maximum duration an installer
+	// pod can remain in a non-terminal phase on a NotReady node before being
+	// force-failed so that the rollout can proceed on healthy nodes.
+	installerPodStuckOnNotReadyNodeTimeout = 15 * time.Minute
 
 	nodeStatusOperandFailedReason         = "OperandFailed"
 	nodeStatusInstalledFailedReason       = "InstallerFailed"
@@ -97,6 +108,11 @@ type InstallerController struct {
 	eventRecorder    events.Recorder
 	now              func() time.Time // for test plumbing
 
+	// nodeLister provides access to node objects for checking the NodeReady condition.
+	// When set, the installer deprioritizes nodes that have been NotReady longer than
+	// nodeNotReadyThreshold and times out installer pods stuck on such nodes.
+	nodeLister corev1listers.NodeLister
+
 	// installerPodImageFn returns the image name for the installer pod
 	installerPodImageFn func() string
 	// ownerRefsFn sets the ownerrefs on the pruner pod
@@ -130,6 +146,19 @@ func (c *InstallerController) WithInstallerPodMutationFn(installerPodMutationFn 
 
 func (c *InstallerController) WithMinReadyDuration(minReadyDuration time.Duration) *InstallerController {
 	c.minReadyDuration = minReadyDuration
+	return c
+}
+
+// WithNodeLister enables NodeReady-aware rollout ordering. Nodes that have been
+// NotReady for longer than nodeNotReadyThreshold are deprioritized in the rollout
+// ring, and installer pods stuck on such nodes are timed out. This prevents a
+// single failed master from blocking revision rollout (e.g., cert rotation) on
+// healthy masters. The nodeInformer triggers controller resync on node status changes.
+func (c *InstallerController) WithNodeLister(nodeLister corev1listers.NodeLister, nodeInformer factory.Informer) *InstallerController {
+	c.nodeLister = nodeLister
+	if nodeInformer != nil {
+		c.factory.WithInformers(nodeInformer)
+	}
 	return c
 }
 
@@ -322,11 +351,44 @@ func (c *InstallerController) getStaticPodState(ctx context.Context, nodeName st
 
 type staticPodStateFunc func(ctx context.Context, nodeName string) (state staticPodState, revision, reason string, errors []string, ts time.Time, err error)
 
+// isNodeNotReadyForTooLong returns true if the named node has been in a
+// non-Ready state for longer than nodeNotReadyThreshold. Returns false when
+// the node lister is not configured, the node cannot be found, or the node
+// has been NotReady for less than the threshold.
+func (c *InstallerController) isNodeNotReadyForTooLong(nodeName string) bool {
+	if c.nodeLister == nil {
+		return false
+	}
+	node, err := c.nodeLister.Get(nodeName)
+	if err != nil {
+		klog.V(4).Infof("Cannot get node %s to check readiness, assuming ready: %v", nodeName, err)
+		return false
+	}
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == corev1.NodeReady {
+			if cond.Status != corev1.ConditionTrue {
+				notReadyDuration := c.clock.Now().Sub(cond.LastTransitionTime.Time)
+				if notReadyDuration > nodeNotReadyThreshold {
+					klog.Infof("Node %s has been NotReady for %v (threshold %v), deprioritizing for rollout",
+						nodeName, notReadyDuration.Round(time.Second), nodeNotReadyThreshold)
+					return true
+				}
+			}
+			return false
+		}
+	}
+	// No NodeReady condition found — the node may be freshly bootstrapping;
+	// do not penalize it.
+	return false
+}
+
 // nodeToStartRevisionWith returns a node index i and guarantees for every node < i that it is
 // - not updating
 // - ready
 // - at the revision claimed in CurrentRevision.
-func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodStateFunc, nodes []operatorv1.NodeStatus) (int, string, error) {
+// isNodeNotReadyTooLong, when non-nil, is used to skip nodes that have been Kubernetes-NotReady
+// for too long, so that healthy nodes are served first during revision rollout.
+func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodStateFunc, nodes []operatorv1.NodeStatus, isNodeNotReadyTooLong func(string) bool) (int, string, error) {
 	if len(nodes) == 0 {
 		return 0, "", fmt.Errorf("nodes array cannot be empty")
 	}
@@ -334,6 +396,9 @@ func nodeToStartRevisionWith(ctx context.Context, getStaticPodStateFn staticPodS
 	// find upgrading node as this will be the first to start new revision (to minimize number of down nodes)
 	for i := range nodes {
 		if nodes[i].TargetRevision != 0 {
+			if isNodeNotReadyTooLong != nil && isNodeNotReadyTooLong(nodes[i].NodeName) {
+				continue
+			}
 			reason := fmt.Sprintf("node %s is progressing towards %d", nodes[i].NodeName, nodes[i].TargetRevision)
 			return i, reason, nil
 		}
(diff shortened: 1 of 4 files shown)


tjungblu:
Aside from the technicalities involved in the static pod machinery that makes this approach infeasible, the cert rotation has a 9 month (!) buffer for you to fix your node:
https://github.com/openshift/cluster-etcd-operator/blob/main/pkg/tlshelpers/tlshelpers.go#L28-L34

are you telling me that there is a node that was not ready for 9 consecutive months?

dpateriya:
Hi @tjungblu, it is not about the etcd certificate.

It is about the aggregator-client, aggregator-client-signer, and csr-signer certificates, which have a shorter validity period compared to etcd certs.

When these certificates rotate, the new cert is written to etcd and a new revision is triggered. The installer controller then targets the NotReady node first because nodeToStartRevisionWith prioritizes nodes whose static pod is not reporting Ready — which includes nodes that are Kubernetes-NotReady (since kubelet is down, the mirror pod status is stale/absent).

The customer had a master node in NotReady state for 12 days due to hardware failure. They were aware of this and the fix (hardware replacement) was estimated to take another 2-3 weeks. The cluster was expected to continue operating with 2 healthy masters during this period.

Due to the aggregator-client and aggregator-client-signer certificate rotation triggering a new revision that got stuck on the NotReady node, the customer was unable to login to the OCP cluster via console and CLI (401 Unauthorized for all users).

dpateriya:
Also, this fix preserves the one-at-a-time rollout invariant; it only changes which node is picked first in the ring.

dpateriya:
@tjungblu, can you please do the needful once you have time?

p0lyn0mial:
> It is about the aggregator-client, aggregator-client-signer, and csr-signer certificates.

@dpateriya please check but I think these certificates are unrevisioned.

https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/pkg/operator/starter.go#L675C1-L675C29
https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/pkg/operator/starter.go#L661
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/main/pkg/operator/starter.go#L350

Unrevisioned resources are delivered by a sidecar and don't go through the revision/installer mechanism. The static pods reload them dynamically from disk without restart.

If that the case then it could be that something else broke the cluster.

dpateriya:
@p0lyn0mial , I did confirm the same from the must-gather report.

While the leaf aggregator-client cert is indeed delivered by cert-syncer (unrevisioned), the aggregator-client-ca ConfigMap IS a revisioned resource (visible in CertConfigMapNamePrefixes). When the signer rotated, it updated this CA bundle, triggering revision 301.
~~~
  - currentRevision: 299
    lastFailedCount: 1
    lastFailedReason: InstallerFailed
    lastFailedRevision: 5
    lastFailedRevisionErrors:
    - |
      installer: elet-client",
        (string) (len=16) "node-kubeconfigs"
       },
       OptionalCertSecretNamePrefixes: ([]string) (len=11 cap=16) {
        (string) (len=17) "user-serving-cert",
        (string) (len=21) "user-serving-cert-000",
        (string) (len=21) "user-serving-cert-001",
        (string) (len=21) "user-serving-cert-002",
        (string) (len=21) "user-serving-cert-003",
        (string) (len=21) "user-serving-cert-004",
        (string) (len=21) "user-serving-cert-005",
        (string) (len=21) "user-serving-cert-006",
        (string) (len=21) "user-serving-cert-007",
        (string) (len=21) "user-serving-cert-008",
        (string) (len=21) "user-serving-cert-009"
       },
       CertConfigMapNamePrefixes: ([]string) (len=4 cap=4) {
        (string) (len=20) "aggregator-client-ca",
        (string) (len=9) "client-ca",
        (string) (len=29) "control-plane-node-kubeconfig",
        (string) (len=26) "check-endpoints-kubeconfig"
       },
       OptionalCertConfigMapNamePrefixes: ([]string) (len=1 cap=1) {
        (string) (len=17) "trusted-ca-bundle"
       },
       CertDir: (string) (len=57) "/etc/kubernetes/static-pod-resources/kube-apiserver-certs",
       ResourceDir: (string) (len=36) "/etc/kubernetes/static-pod-resources",
       PodManifestDir: (string) (len=25) "/etc/kubernetes/manifests",
       Timeout: (time.Duration) 2m0s,
       StaticPodManifestsLockFile: (string) "",
       PodMutationFns: ([]installerpod.PodMutationFunc) <nil>,
       KubeletVersion: (string) ""
      })
      I0416 07:49:27.372032       1 cmd.go:410] Getting controller reference for node master1.example.com
      I0416 07:49:27.379776       1 cmd.go:423] Waiting for installer revisions to settle for node master1.example.com
      I0416 07:49:27.381713       1 cmd.go:515] Waiting additional period after revisions have settled for node master1.example.com
      I0416 07:49:57.382465       1 cmd.go:521] Getting installer pods for node master1.example.com
      F0416 07:50:11.386170       1 cmd.go:106] Get "https://172.30.0.1:443/api/v1/namespaces/openshift-kube-apiserver/pods?labelSelector=app%3Dinstaller": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
    lastFailedTime: "2025-04-16T07:53:56Z"
    lastFallbackCount: 0
    nodeName: master1.example.com
    targetRevision: 301
  ~~~

The InstallerController selected the NotReady node first for revision 301, got stuck, and never proceeded to the two healthy nodes. The healthy nodes remained on revision 299 with the old CA trust bundle, so they could not validate the newly rotated leaf certs delivered by cert-syncer — resulting in x509: certificate has expired errors and authentication failures.

Once the NotReady node was removed (unblocking the operator), the revision rolled out to the healthy nodes and login was immediately restored — confirming the revision contained the critical aggregator-client-ca update.

The fix ensures the InstallerController deprioritizes NotReady nodes so that healthy nodes receive the new revision first, preventing this class of outage.

p0lyn0mial:
> the aggregator-client-ca ConfigMap IS a revisioned resource

@dpateriya please double check but it seems that `aggregator-client-ca` is not revisioned.

have a look at: https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/pkg/operator/starter.go#L661

```
var CertConfigMaps = []installer.UnrevisionedResource{
	{Name: "aggregator-client-ca"},
	...
}
```

dpateriya:
@p0lyn0mial you are correct, `aggregator-client-ca` is unrevisioned resource. Thanks for the correction.

However, the must-gather evidence clearly shows that revision 301 was created (likely triggered by a revisioned resource such as kube-apiserver-cert-syncer-kubeconfig, sa-token-signing-certs, or bound-sa-token-signing-certs) and got stuck on the NotReady node:
~~~
nodeStatuses:
- nodeName: master1.example.com (NotReady since June 14)
  currentRevision: 299
  targetRevision: 301   # stuck here

- nodeName: master2.example.com (Ready)
  currentRevision: 299
  targetRevision: 0     # never attempted

- nodeName: master3.example.com (Ready)
  currentRevision: 299
  targetRevision: 0     # never attempted
  ~~~

The InstallerController selected the NotReady node first for revision 301. The two healthy nodes with targetRevision: 0 were never started. Once the NotReady node was removed and the revision completed on the healthy nodes, login was immediately restored.

Regardless of which specific revisioned resource triggered revision 301,  the core issue is that any new revision gets stuck when the InstallerController targets a NotReady node first, blocking healthy nodes from progressing. This fix ensures healthy nodes are prioritized so the cluster remains functional even with a prolonged NotReady master.

I'd value your perspective on whether deprioritizing NotReady nodes in `nodeToStartRevisionWith` is the right approach here, or if there's an existing mechanism I may have missed that handles this case.

tjungblu:
> Once the NotReady node was removed and the revision completed on the healthy nodes, login was immediately restored.

are you sure, Claude? because in the support case this is not the case, the customer wanted to keep the not-ready node around at all cost. 


The procedure applied was:

>Static Pod Force-Rollout: We manually forced a rollout of the kube-apiserver, kube-scheduler, and kube-controller-manager static pods by temporarily moving their manifest files out of /etc/kubernetes/manifests and added back to same path.

> Component Restarts: We restarted the running pods within the openshift-oauth-apiserver, openshift-authentication, and openshift-apiserver namespaces to force them to pick up new tokens/secrets, though initial oc login attempts continued to fail due to the pending cluster configuration updates.




But it is not clear which of those restarts were actually fixing the issue. Maybe just restarting oauth-apiserver would've fixed it. They have experienced login issues and they're using IDP. I think somebody from auth should take a look at this case first and analyse the must-gather before we start plumbing the static pod machinery here.

dpateriya:
@tjungblu, claude was not on the call with the customer. I was on the call with the customer.

When I said I removed the master node, it meant that the master role (node-role.kubernetes.io/master=) from the not ready master node was removed, and then the installercontroller targeted the rest of available 2 master nodes.

Manual restart of oauth-openshift pods, openshift-apiserver pods, openshift-oauth-apiserver pods and even static pods like KCM, KAS were restarted.

The static pods were restarted by moving their pod yaml from /etc/kubernetes/manifests to /home/core and then again moved back to the original location.


All these were the efforts made during the remote call.  But this does not change the fact that the installer controller prioritizes the not ready master node first.

pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -45,6 +46,16 @@ const (
 
 	revisionLabel = "revision"
 
+	// nodeNotReadyThreshold is how long a node must be Kubernetes-NotReady before
+	// the installer controller deprioritizes it in rollout ordering. This prevents
+	// a prolonged hardware failure from blocking cert rotation on healthy masters.
+	nodeNotReadyThreshold = 10 * time.Minute
```
tjungblu, line 52: a bare metal node can take about an hour to reboot, so the node will be not ready for 50 minutes. Will this cause etcd downtime when a static pod revision rolls out?

dpateriya: AFAIK, etcd can suffer a loss/unavailability of one master node.

So even if a bare-metal node takes hours to reboot, this will have no downtime for etcd.

Also, the KAS revision rollout does not trigger etcd pod rollout. 

Taking your feedback into consideration, we can increase the `nodeNotReadyThreshold`  to 60 minutes to avoid unnecessary skipping during normal reboots for bare-metal nodes while still catching prolonged outages (the incident involved 12+ days NotReady).

Please confirm.

tjungblu: You understand that the code is shared between all control plane operators, right?

> So even if a bare-metal node takes hours to reboot, this will have no downtime for etcd.

of course it does, because your code doesn't stop the rollout.

dpateriya: Agree, so a bare-metal node reboot + etcd revision rollout on the other master node will briefly trigger quorum loss and a cluster outage.

We can exclude etcd without adding any new methods or interfaces. The etcd-operator is the only consumer that sets `WithRevisionControllerPrecondition` (for quorum safety). KAS, KCM, and scheduler don't use it.

So the operators like etcd, which are quorum-sensitive, will not get the NotReady skip and all others like KAS, KCM, and scheduler will do.

dpateriya: One condition change in controllers.go line 300:
```
// Before:
if clusterInformers != nil {

// After:
if clusterInformers != nil && b.revisionControllerPrecondition == nil {
```

I think this would help then.

dpateriya: Changes have been made.

You can verify the same.

dpateriya:
Hi Team, can someone please re-review this?

PR #2339, opened 2026-06-29
CNTRLPLANE-3237: encryption/controllers/kms_preflight_controller: manage preflight pod lifecycle

p0lyn0mial:
## Summary by CodeRabbit

* **New Features**
  * Implemented KMS preflight workflow end-to-end: workload deployment, readiness/result signal interpretation, operator degraded updates, and controlled cleanup/retention.
  * Added new readiness/result condition signals to more accurately reflect preflight progress and outcomes.

* **Bug Fixes**
  * Improved handling for stale pods, crashed pods, missing readiness reporting (startup timeout), and failure propagation with clearer error messaging.

* **Tests**
  * Refreshed and expanded preflight controller tests with an injectable deployer to validate lifecycle, requeue behavior, retention vs. cleanup, and degraded conditions across many scenarios.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
@@ -120,6 +121,27 @@ func (h *kmsConfigHasher) hashReferencedConfigMap(ctx context.Context, hasher ha
 	return nil
 }
 
+// Pod readiness gate condition types set by the preflight checker running inside
+// the pod. The checker PATCHes its own pod status with these conditions.
+const (
+	// KMSPreflightConfigHashPodCondition carries the config hash the pod was
+	// deployed for. The controller compares this against the required hash to
+	// detect stale pods from a previous config.
+	KMSPreflightConfigHashPodCondition corev1.PodConditionType = "KMSPreflightConfigHash"
+
+	// KMSPreflightResultPodCondition carries the outcome of the preflight check.
+	// Status True means the check passed; False means it failed, with details
+	// in the condition message.
+	KMSPreflightResultPodCondition corev1.PodConditionType = "KMSPreflightResult"
+
+	// KMSPreflightKEKIDPodCondition carries the current key encryption key (KEK)
+	// ID reported by the KMS plugin during the preflight evaluation.
+	KMSPreflightKEKIDPodCondition corev1.PodConditionType = "KMSPreflightKekID"
+
+	preflightPodRetention      = 1 * time.Hour
+	preflightPodStartupTimeout = 3 * time.Minute
+)
+
 // KMSPreflightDeployer abstracts the lifecycle of a preflight workload that
 // validates KMS plugin configuration before an encryption key is created.
 // All methods are idempotent.
@@ -149,6 +171,7 @@ type kmsPreflightController struct {
 	apiServerClient configv1client.APIServerInterface
 	coreClient      corev1client.CoreV1Interface
 
+	deployer                 KMSPreflightDeployer
 	provider                 Provider
 	preconditionsFulfilledFn preconditionsFulfilled
 }
@@ -226,6 +249,7 @@ func NewKMSPreflightController(
 	instanceName string,
 	provider Provider,
 	preconditionsFulfilledFn preconditionsFulfilled,
+	deployer KMSPreflightDeployer,
 	operatorClient operatorv1helpers.OperatorClient,
 	apiServerClient configv1client.APIServerInterface,
 	apiServerInformer configv1informers.APIServerInformer,
@@ -243,6 +267,7 @@ func NewKMSPreflightController(
 		apiServerClient: apiServerClient,
 		coreClient:      coreClient,
 
+		deployer:                 deployer,
 		provider:                 provider,
 		preconditionsFulfilledFn: preconditionsFulfilledFn,
 	}
@@ -260,6 +285,7 @@ func NewKMSPreflightController(
 	)
 }
 
+// TODO: in the future report "progress" and "success" conditions.
 func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncContext) (err error) {
 	degradedCondition := applyoperatorv1.OperatorCondition().WithType("EncryptionKMSPreflightControllerDegraded")
 
@@ -282,12 +308,19 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 		return err // we will get re-kicked when the operator status updates
 	}
 
-	preflightErr := c.runPreflightChecks(ctx)
+	requeue, preflightErr := c.runPreflightChecks(ctx)
+	if requeue {
+		syncCtx.Queue().AddAfter(syncCtx.QueueKey(), 30*time.Second)
+	}
 	if preflightErr != nil {
+		reason, msg := "Error", preflightErr.Error()
+		if pe, ok := preflightErr.(*preflightError); ok {
+			reason = pe.reason
+		}
 		degradedCondition = degradedCondition.
 			WithStatus(operatorv1.ConditionTrue).
-			WithReason("Error").
-			WithMessage(preflightErr.Error())
+			WithReason(reason).
+			WithMessage(msg)
 	} else {
 		degradedCondition = degradedCondition.
 			WithStatus(operatorv1.ConditionFalse)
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -295,21 +302,215 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 	return preflightErr
 }
 
-func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) error {
-	requiredHash, err := c.requiredPreflightHash(ctx)
+// Pod readiness gate condition types set by the preflight checker running inside
+// the pod. The checker PATCHes its own pod status with these conditions.
+const (
```
tjungblu: nitpick: any reason you're defining the constants in the middle somewhere instead at the top?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -295,21 +302,215 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+		return false, pe
+	}
+
+	// Scenario 3b: pod has not reported its config hash yet.
+	hashCondition := findPodCondition(podStatus.Conditions, podConditionKMSPreflightConfigHash)
+	if hashCondition == nil {
+		if podStatus.Phase == corev1.PodSucceeded {
+			return false, &preflightError{reason: "PodCompletedWithoutResult", message: fmt.Sprintf("preflight pod completed without reporting result for hash %s", requiredHash)}
+		}
+		if pe := podStartupTimeoutError(podStatus, "preflight pod has not reported config hash"); pe != nil {
```
tjungblu, line 411: I think we can also solve this with a deadline in the pod config, so it just goes through the PodFailed path

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -295,21 +302,215 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	if resultCondition == nil {
+		if podStatus.Phase == corev1.PodSucceeded {
+			return false, &preflightError{reason: "PodCompletedWithoutResult", message: fmt.Sprintf("preflight pod completed without reporting result for hash %s", requiredHash)}
+		}
+		if pe := podStartupTimeoutError(podStatus, "preflight pod has not reported result"); pe != nil {
+			return true, pe
+		}
+		return true, nil
+	}
+
```
tjungblu, line 433: will we ever get here? because at some point the hash will vanish and we're just cleaning it up at the top with... ?

``` 
	if requiredHash == "" {
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -295,21 +302,215 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	podConditionKMSPreflightConfigHash corev1.PodConditionType = "KMSPreflightConfigHash"
+
+	// podConditionKMSPreflightResult carries the outcome of the preflight check.
+	// Status True means the check passed; False means it failed, with details
+	// in the condition message.
+	podConditionKMSPreflightResult corev1.PodConditionType = "KMSPreflightResult"
+)
+
+const preflightPodRetention = 1 * time.Hour
+const preflightPodStartupTimeout = 2 * time.Minute
```
ardaguclu: Image will be pulled, likely communicated with external service, etc. Isn't 2 minute too short?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -295,21 +302,215 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+
+// runPreflightChecks manages the preflight pod lifecycle across sync iterations.
+// It returns whether to requeue and any error.
+//
+// Scenarios:
```
ardaguclu, line 334: States look reasonable to me

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: would it also terminate pods stuck in Pending ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: the key controller will be responsible for filling the required hash. i think that the controller doesn't even have to clear the hash or clear it only when a new provider was configured.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: what timeout should we set then ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: i can move to the top if that is cleaner. let me know.

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: nope, for pods it does not. For Jobs and Deployments, it would.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: Maybe 5 minutes would be better?

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: that would be good. In the PR #2315 I have the following constants:

```
const (
	// KMSPreflightCheckConditionType is the pod condition the checker sets when it
	// finishes. Deployers should reference this type in readiness gates so the
	// operator controller can observe the result via pod status.
	KMSPreflightCheckConditionType corev1.PodConditionType = "EncryptionKMSPreflightCheck"
	// KMSPreflightKeyIDConditionType reports the KMS plugin key ID in Message.
	KMSPreflightKeyIDConditionType corev1.PodConditionType = "EncryptionKMSPreflightKeyID"
	// KMSPreflightVersionConditionType reports the KMS plugin version in Message.
	KMSPreflightVersionConditionType corev1.PodConditionType = "EncryptionKMSPreflightVersion"
)
``` 

I think yours is better and I'll rebase over it, do you mind to add EncryptionKMSPreflightKeyID along? Then we can record the KEK with it.

tjungblu: oh and please export them...

tjungblu:
```suggestion
// Pod readiness gate condition types set by the preflight checker running inside
// the pod. The checker PATCHes its own pod status with these conditions.
const (
	// PodConditionKMSPreflightConfigHash carries the config hash the pod was
	// deployed for. The controller compares this against the required hash to
	// detect stale pods from a previous config.
	PodConditionKMSPreflightConfigHash corev1.PodConditionType = "KMSPreflightConfigHash"

	// PodConditionKMSPreflightResult carries the outcome of the preflight check.
	// Status True means the check passed; False means it failed, with details
	// in the condition message.
	PodConditionKMSPreflightResult corev1.PodConditionType = "KMSPreflightResult"

	// PodConditionKMSPreflightKeyID reports the KMS plugin key ID in Message.
	PodConditionKMSPreflightKeyID corev1.PodConditionType = "KMSPreflightKeyID"
)
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: is there any state we miss ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: then i think we would keep `podStartupTimeoutError`

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: I don't think so

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: ok, will do

p0lyn0mial: `KMSPreflightKeyID` or `KMSPreflightKekID` ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: decided to bump to 3 min for now.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -120,6 +121,27 @@ func (h *kmsConfigHasher) hashReferencedConfigMap(ctx context.Context, hasher ha
 	return nil
 }
 
+// Pod readiness gate condition types set by the preflight checker running inside
```
ardaguclu, line 124: Do we set readiness gate condition types?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -282,34 +308,226 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+//
+//     a) Pod phase is Failed — the pod crashed before or after posting
+//     conditions. Report degraded and keep the pod for inspection. The
+//     admin fixes the config, which triggers a new hash and cleanup
+//     via scenario (c).
+//
+//     b) No KMSPreflightConfigHash condition yet — the checker has not
+//     started reporting. If the pod phase is Succeeded, it exited
+//     without reporting; return an error. Otherwise requeue and wait.
+//     If the pod has exceeded the startup timeout (2m) without
```
ardaguclu:
```suggestion
//     If the pod has exceeded the startup timeout (3m) without
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -282,34 +308,226 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+		return true, c.deployer.Deploy(ctx, requiredHash, nil)
+	}
+	if err != nil {
+		return false, fmt.Errorf("failed to get preflight pod status: %w", err)
+	}
+
+	// Scenario 3a: pod crashed. Keep for inspection; the admin will update
+	// the config which triggers a new hash and cleanup via scenario 3c.
+	if podStatus.Phase == corev1.PodFailed {
+		pe := podFailureError(podStatus)
```
ardaguclu, line 400:
```suggestion
		pe, ok := podFailureError(podStatus)
		if !ok {
		  return false, ...
		}
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -282,34 +308,226 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+		pe.message = fmt.Sprintf("preflight pod failed for hash %s: %s", requiredHash, pe.message)
+		return false, pe
+	}
+
+	// Scenario 3b: pod has not reported its config hash yet.
+	hashCondition := findPodCondition(podStatus.Conditions, KMSPreflightConfigHashPodCondition)
+	if hashCondition == nil {
+		if podStatus.Phase == corev1.PodSucceeded {
+			return false, &preflightError{reason: "PodCompletedWithoutResult", message: fmt.Sprintf("preflight pod completed without reporting result for hash %s", requiredHash)}
```
ardaguclu, line 409: nit: Would it make sense adding a documentation about how this can occur?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -120,6 +121,27 @@ func (h *kmsConfigHasher) hashReferencedConfigMap(ctx context.Context, hasher ha
...
+	KMSPreflightConfigHashPodCondition corev1.PodConditionType = "KMSPreflightConfigHash"
+
+	// KMSPreflightResultPodCondition carries the outcome of the preflight check.
+	// Status True means the check passed; False means it failed, with details
+	// in the condition message.
+	KMSPreflightResultPodCondition corev1.PodConditionType = "KMSPreflightResult"
+
+	// KMSPreflightKEKIDPodCondition carries the current key encryption key (KEK)
+	// ID reported by the KMS plugin during the preflight evaluation.
+	KMSPreflightKEKIDPodCondition corev1.PodConditionType = "KMSPreflightKekID"
```
ardaguclu, line 139: Is this used?. Who is going to use it?

p0lyn0mial: requested by @tjungblu

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: sure, if that helps i can document.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: not sure if i understand this comment.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: That makes sense

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: already updated.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: @tjungblu have we checked the readiness conditions are not needed ?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -282,34 +308,226 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
...
+	}
+	if podStatus.Message != "" {
+		return &preflightError{reason: reason, message: podStatus.Message}
+	}
+	return &preflightError{reason: reason, message: "unknown"}
+}
+
+func podStartupTimeoutError(podStatus corev1.PodStatus, msgPrefix string) *preflightError {
+	if podStatus.StartTime == nil || time.Since(podStatus.StartTime.Time) <= preflightPodStartupTimeout {
+		return nil
```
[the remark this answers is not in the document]
p0lyn0mial: i'm looking into this.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: ah, my bad. I thought there was a type casting, but instead it is a function call. Ignore my comment.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: actually, have a look at the doc for the `runPreflightChecks` method. 
each scenario listed here is documented. for `3b` we have:


```
//     b) No KMSPreflightConfigHash condition yet — the checker has not
//     started reporting. If the pod phase is Succeeded, it exited
//     without reporting; return an error. Otherwise requeue and wait.
//     If the pod has exceeded the startup timeout (3m) without
//     reporting, return an error with the reason the pod is stuck
//     (e.g. ImagePullBackOff, Pending).

```

ardaguclu: > If the pod phase is Succeeded, it exited without reporting

This is more of a question rather than documentation update: How can pod succeed without reporting?

ardaguclu: Does that mean Pod has run but it detected an issue in kms plugin?

p0lyn0mial: This could be a bug in the preflight binary. 
This could be a failure when updating the status from the preflight binary to the server (a network err).
This could be a different actor in the system that removed the conditions. 
...

ardaguclu: That makes sense. Thank you.

ardaguclu:
/lgtm
/hold
for @tjungblu

tjungblu:
all good, let's merge
/hold cancel

PR #2341, opened 2026-06-30
Assert also provider name in encrypted resource

ardaguclu:
Currently we only compare the prefixes of the encrypted resources in etcd. This works until now because we only exercise migrations from one encryption mode to another. However, in kms to kms migration scenarios, all the prefixes are made up of `KMS:`. That's why, we need to inspect the encrypted resources with more granularity. 

This PR appends provider name in KMS modes rather than plain `KMS` prefix. So that operators will have to update their assertion functions like this;

```go
 func AssertSecretOfLifeEncryptedByKMSProvider(t testing.TB, clientSet encryption.ClientSet, resource runtime.Object) {
      secret := resource.(*corev1.Secret)

      lastKeyMeta, err := encryption.GetLastKeyMeta(t, clientSet.Kube, namespace, labelSelector)
      require.NoError(t, err)

      parts := strings.Split(lastKeyMeta.Name, "-")
      expectedMode := "KMS:" + parts[len(parts)-1] + "_"

      _, err = encryption.VerifyResources(t, clientSet.Etcd,
          "/kubernetes.io/secrets/"+secret.Namespace+"/"+secret.Name,
          expectedMode, false)
      require.NoError(t, err)
  }
```

Thanks to that we'll verify that resources are migration from 1_secret to 2_secret.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption mode checks so KMS v2 values are recognized more accurately, including provider-specific modes.
  * Mode matching is now more flexible, reducing false failures when encrypted/decrypted values include a compatible prefix.

### DIFF
--- a/test/library/encryption/assertion.go
+++ b/test/library/encryption/assertion.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ardaguclu:
/close
I'm not happy about the approach here.

PR #2343, opened 2026-07-01
NO-JIRA: apply RBAC on preflight deploys

tjungblu:
This adds an deploy-only RBAC to ensure we can update the pod status conditions. 

## Summary by CodeRabbit

* **New Features**
  * Preflight checks now run with a dedicated service account.
  * Required RBAC access for the preflight pod is created automatically before the pod starts.
* **Bug Fixes**
  * Improved preflight reliability by granting permissions to read pods and update pod status.
  * Updated validation so generated pod definitions match the new service account setup.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -7,16 +7,29 @@ import (
 	"time"
 
 	"github.com/openshift/library-go/pkg/operator/encryption/kms/pluginlifecycle"
+	"github.com/openshift/library-go/pkg/operator/events"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	rbacclientv1 "k8s.io/client-go/kubernetes/typed/rbac/v1"
+
+	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	rbacv1 "k8s.io/api/rbac/v1"
 )
 
 const (
 	preflightPodName                    = "kms-preflight"
 	preflightCheckContainerName         = "kms-preflight-check"
 	preflightEncryptionConfigSecretName = "kms-preflight-encryption-config"
+	preflightRBACName                   = "kms-preflight"
+	preflightAppLabel                   = "openshift-kms-preflight"
+)
+
+var (
+	labels = map[string]string{
+		"app": preflightAppLabel,
+	}
 )
 
 // PodPreflightDeployer creates a one-shot pod to run the preflight checker as a command,
@@ -27,6 +40,8 @@ type PodPreflightDeployer struct {
 	// encryption controllers/components actually use.
 	// In addition to that running the preflight check is not a very frequent operation.
 	coreClient      corev1client.CoreV1Interface
+	rbacClient      rbacclientv1.RbacV1Interface
+	eventRecorder   events.Recorder
 	operatorImage   string
 	operatorCommand []string
 	kmsCallTimeout  time.Duration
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -107,16 +124,85 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
...
+			Name:      preflightRBACName,
+			Namespace: d.namespace,
+			Labels:    labels,
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to apply preflight service account: %w", err)
+	}
+
+	_, _, err = resourceapply.ApplyRole(ctx, d.rbacClient, d.eventRecorder, &rbacv1.Role{
```
ardaguclu, line 146: Just an idea: Would it make sense templating these resources under `assets`?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -107,16 +124,85 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
 	return errors.Join(errs...)
 }
 
+func (d *PodPreflightDeployer) applyPreflightRBAC(ctx context.Context) error {
+	labels := map[string]string{
+		"app": preflightAppLabel,
```
ardaguclu: I assume that we add this label for all the resource we created here?

tjungblu: yep

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: I initially had it that way, but then how do I transform this into an apply config? and it's not that much code either to directly encode it

tjungblu: oh and there is also nothing to template, besides the namespace

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: actually forgot to set it on the encryption secret. Added it now.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -107,16 +130,82 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
 	return errors.Join(errs...)
 }
 
+func (d *PodPreflightDeployer) applyPreflightRBAC(ctx context.Context) error {
+
+	_, _, err := resourceapply.ApplyServiceAccount(ctx, d.coreClient, d.eventRecorder, &corev1.ServiceAccount{
```
ardaguclu, line 135: There will always be one instance of resources and this won't create any race conditions?. Probably this `ApplyServiceAccount` is not a problem because we always apply the same manifest.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -107,16 +130,82 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      preflightRBACName,
```
ardaguclu, line 137: Will this ServiceAccount work on static pod?

ardaguclu:
/hold
Feel free to unhold if the answer of this question https://github.com/openshift/library-go/pull/2343#discussion_r3506681066 is yes

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: yep it's a singleton, I mean the service account will just stay forever, there's nothing we could patch here besides the labels.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: that I do not know yet, I have my doubts that we can patch conditions on a mirror pod and also on kubelet's guarantees on actually reporting the mirror pod accurately. Oh and there might be races here as well. Imagine the static pod runs to completion, but then mirror pod is not yet created in the API. That means we can't update it either.

tjungblu:
> Feel free to unhold if the answer of this question https://github.com/openshift/library-go/pull/2343#discussion_r3506681066 is yes

it's "I don't know yet" :) hope that counts too. We can adjust later, if necessary. The static pod deployer will be a different struct entirely.

/hold cancel

PR #2347, opened 2026-07-03
feat(crypto): add TLS group/curve mapping support

damdo:
Map OpenShift API `TLSGroup` constants to Go's `tls.CurveID` values, following the same pattern used for cipher suite mapping. This enables downstream consumers (observer, operators) to resolve curve preferences from TLS security profiles.

All seven API-defined groups are mapped, including the three post-quantum hybrid key exchanges (`X25519MLKEM768`, `SecP256r1MLKEM768`, `SecP384r1MLKEM1024`) added in Go 1.26.

### DIFF
--- a/pkg/crypto/crypto.go
+++ b/pkg/crypto/crypto.go
@@ -20,8 +20,10 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -211,6 +213,17 @@ var ciphersUnsupportedByGo = map[string]string{
 	"AES256-SHA256":             "TLS_RSA_WITH_AES_256_CBC_SHA256",
 }
 
+// tlsGroupToCurveID maps OpenShift API TLSGroup values to Go's tls.CurveID.
+var tlsGroupToCurveID = map[configv1.TLSGroup]tls.CurveID{
+	configv1.TLSGroupX25519:             tls.X25519,
+	configv1.TLSGroupSecP256r1:          tls.CurveP256,
+	configv1.TLSGroupSecP384r1:          tls.CurveP384,
+	configv1.TLSGroupSecP521r1:          tls.CurveP521,
+	configv1.TLSGroupX25519MLKEM768:     tls.X25519MLKEM768,
+	configv1.TLSGroupSecP256r1MLKEM768:  tls.SecP256r1MLKEM768,
+	configv1.TLSGroupSecP384r1MLKEM1024: tls.SecP384r1MLKEM1024,
+}
+
 // CipherSuitesToNamesOrDie given a list of cipher suites as ints, return their readable names
 func CipherSuitesToNamesOrDie(intVals []uint16) []string {
 	ret := []string{}
@@ -354,6 +367,44 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
 	}
 
 	return ianaCiphers
+}
+
+// TLSGroupToCurveID returns the Go tls.CurveID for an OpenShift API TLSGroup
+// constant, or (0, false) if the group is not recognized.
+func TLSGroupToCurveID(group configv1.TLSGroup) (tls.CurveID, bool) {
+	id, ok := tlsGroupToCurveID[group]
+	return id, ok
+}
+
+// TLSGroupsToCurveIDs converts a slice of TLSGroup values to their tls.CurveID
+// codes. Returns the mapped curve IDs and a list of unrecognized groups.
+// Unrecognized groups are silently filtered — callers should log warnings.
+func TLSGroupsToCurveIDs(groups []configv1.TLSGroup) ([]tls.CurveID, []configv1.TLSGroup) {
+	var curves []tls.CurveID
+	var unrecognized []configv1.TLSGroup
+
+	for _, group := range groups {
+		id, ok := TLSGroupToCurveID(group)
+		if !ok {
+			unrecognized = append(unrecognized, group)
+			continue
+		}
+		curves = append(curves, id)
+	}
+
+	return curves, unrecognized
+}
+
+// ValidTLSGroups returns the recognized TLS group names, sorted alphabetically.
+func ValidTLSGroups() []configv1.TLSGroup {
+	groups := make([]configv1.TLSGroup, 0, len(tlsGroupToCurveID))
+	for g := range tlsGroupToCurveID {
+		groups = append(groups, g)
+	}
+	slices.SortFunc(groups, func(a, b configv1.TLSGroup) int {
+		return strings.Compare(string(a), string(b))
+	})
+	return groups
 }
 
 type TLSCertificateConfig struct {
--- a/pkg/crypto/crypto_test.go
+++ b/pkg/crypto/crypto_test.go
@@ -2,10 +2,12 @@ package crypto
 
 import (
 	"crypto"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"fmt"
 	"go/importer"
+	"go/types"
 	"os"
 	"path/filepath"
 	"sort"
@@ -22,11 +24,73 @@ import (
 
 const certificateLifetime = time.Hour * 24 * 365 * 2
 
+// goTLSGroups lists all TLS supported groups (key exchange mechanisms) recognized
+// by Go's crypto/tls, keyed by Go constant name. Kept in sync with the crypto/tls
+// package by TestConstantMaps.
+var goTLSGroups = map[string]tls.CurveID{
+	"CurveP256":          tls.CurveP256,          // IANA 23
+	"CurveP384":          tls.CurveP384,          // IANA 24
+	"CurveP521":          tls.CurveP521,          // IANA 25
+	"X25519":             tls.X25519,             // IANA 29
+	"X25519MLKEM768":     tls.X25519MLKEM768,     // IANA 4588
+	"SecP256r1MLKEM768":  tls.SecP256r1MLKEM768,  // IANA 4587
+	"SecP384r1MLKEM1024": tls.SecP384r1MLKEM1024, // IANA 4589
+}
+
 func TestDefaultCipherSuite(t *testing.T) {
 	// Ensure that conversion of the default cipher suite to names
 	// completes without panic.
 	_ = CipherSuitesToNamesOrDie(DefaultCiphers())
 }
+
+// ignoredTLSConstants lists crypto/tls constants that are not versions, ciphers,
+// or groups. If Go adds a new constant, TestConstantMaps will fail until it is
+// placed in the right map or added here.
+var ignoredTLSConstants = sets.New[string](
+	// SSLv3 — not supported by OpenShift
+	"VersionSSL30",
+	// Sentinel cipher
+	"TLS_FALLBACK_SCSV",
+	// Client auth types
+	"NoClientCert",
+	"RequestClientCert",
+	"RequireAnyClientCert",
+	"RequireAndVerifyClientCert",
+	"VerifyClientCertIfGiven",
+	// Renegotiation policies
+	"RenegotiateNever",
+	"RenegotiateOnceAsClient",
+	"RenegotiateFreelyAsClient",
+	// Signature schemes
+	"PKCS1WithSHA1",
+	"PKCS1WithSHA256",
+	"PKCS1WithSHA384",
+	"PKCS1WithSHA512",
+	"PSSWithSHA256",
+	"PSSWithSHA384",
+	"PSSWithSHA512",
+	"ECDSAWithSHA1",
+	"ECDSAWithP256AndSHA256",
+	"ECDSAWithP384AndSHA384",
+	"ECDSAWithP521AndSHA512",
+	"Ed25519",
+	// QUIC events and encryption levels
+	"QUICEncryptionLevelInitial",
+	"QUICEncryptionLevelEarly",
+	"QUICEncryptionLevelHandshake",
+	"QUICEncryptionLevelApplication",
+	"QUICNoEvent",
+	"QUICSetReadSecret",
+	"QUICSetWriteSecret",
+	"QUICWriteData",
+	"QUICTransportParameters",
+	"QUICTransportParametersRequired",
+	"QUICRejectedEarlyData",
+	"QUICHandshakeDone",
+	"QUICResumeSession",
+	"QUICStoreSession",
+	"QUICErrorEvent",
+)
 
 func TestConstantMaps(t *testing.T) {
 	pkg, err := importer.Default().Import("crypto/tls")
@@ -36,12 +100,22 @@ func TestConstantMaps(t *testing.T) {
 	}
 	discoveredVersions := map[string]bool{}
 	discoveredCiphers := map[string]bool{}
+	discoveredGroups := map[string]bool{}
 	for _, declName := range pkg.Scope().Names() {
-		if strings.HasPrefix(declName, "VersionTLS") {
-			discoveredVersions[declName] = true
+		if _, ok := pkg.Scope().Lookup(declName).(*types.Const); !ok {
+			continue
 		}
-		if strings.HasPrefix(declName, "TLS_RSA_") || strings.HasPrefix(declName, "TLS_ECDHE_") || strings.HasPrefix(declName, "TLS_AES_") || strings.HasPrefix(declName, "TLS_CHACHA20_") {
+		switch {
+		case strings.HasPrefix(declName, "VersionTLS"):
+			discoveredVersions[declName] = true
+		case strings.HasPrefix(declName, "TLS_RSA_") || strings.HasPrefix(declName, "TLS_ECDHE_") || strings.HasPrefix(declName, "TLS_AES_") || strings.HasPrefix(declName, "TLS_CHACHA20_"):
 			discoveredCiphers[declName] = true
+		case strings.HasPrefix(declName, "CurveP") || strings.HasPrefix(declName, "X25519") || strings.HasPrefix(declName, "SecP"):
+			discoveredGroups[declName] = true
+		case ignoredTLSConstants.Has(declName):
+			// known unrelated constant
+		default:
+			t.Errorf("unclassified tls constant %q — add it to the appropriate map (goTLSVersions, goCipherSuites, goTLSGroups) or to ignoredTLSConstants", declName)
 		}
 	}
 
@@ -70,6 +144,17 @@ func TestConstantMaps(t *testing.T) {
 	for k := range enabledTLSVersions {
 		if _, ok := discoveredVersions[k]; !ok {
 			t.Errorf("enabledTLSVersions map has %s not in tls package", k)
+		}
+	}
+
+	for k := range discoveredGroups {
+		if _, ok := goTLSGroups[k]; !ok {
+			t.Errorf("discovered group tls.%s not in goTLSGroups map", k)
+		}
+	}
+	for k := range goTLSGroups {
+		if _, ok := discoveredGroups[k]; !ok {
+			t.Errorf("goTLSGroups map has %s not in tls package", k)
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/crypto/crypto.go
```diff
@@ -211,6 +211,39 @@ var ciphersUnsupportedByGo = map[string]string{
...
+// package by TestConstantMaps.
+// Ref: https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+var goTLSGroups = map[string]tls.CurveID{
+	"CurveP256":      tls.CurveP256,      // IANA 23
+	"CurveP384":      tls.CurveP384,      // IANA 24
+	"CurveP521":      tls.CurveP521,      // IANA 25
+	"X25519":         tls.X25519,         // IANA 29
+	"X25519MLKEM768": tls.X25519MLKEM768, // IANA 4588
+	// TODO: add "SecP256r1MLKEM768": tls.SecP256r1MLKEM768 (IANA 4587) when Go 1.26 is the minimum version
+	// TODO: add "SecP384r1MLKEM1024": tls.SecP384r1MLKEM1024 (IANA 4590) when Go 1.26 is the minimum version
```
smith-xyz:
```suggestion
	// TODO: add "SecP384r1MLKEM1024": tls.SecP384r1MLKEM1024 (IANA 4589) when Go 1.26 is the minimum version
```

https://www.ietf.org/archive/id/draft-ietf-tls-ecdhe-mlkem-01.html#name-secp384r1mlkem1024

probably safer to document as 4589 - unless I'm incorrect and this is something specific to Go

pkg/crypto/crypto_test.go
```diff
@@ -36,13 +37,17 @@ func TestConstantMaps(t *testing.T) {
...
 	discoveredCiphers := map[string]bool{}
+	discoveredGroups := map[string]bool{}
 	for _, declName := range pkg.Scope().Names() {
 		if strings.HasPrefix(declName, "VersionTLS") {
 			discoveredVersions[declName] = true
 		}
 		if strings.HasPrefix(declName, "TLS_RSA_") || strings.HasPrefix(declName, "TLS_ECDHE_") || strings.HasPrefix(declName, "TLS_AES_") || strings.HasPrefix(declName, "TLS_CHACHA20_") {
 			discoveredCiphers[declName] = true
 		}
+		if strings.HasPrefix(declName, "CurveP") || strings.HasPrefix(declName, "X25519") {
```
smith-xyz: Does this filter need SecP and MLKEM prefixes? Or will that be updated when addressing the TODOs for go 1.26 >= ?

pkg/crypto/crypto.go
```diff
@@ -211,6 +211,39 @@ var ciphersUnsupportedByGo = map[string]string{
...
+// by Go's crypto/tls, keyed by Go constant name. Kept in sync with the crypto/tls
+// package by TestConstantMaps.
+// Ref: https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+var goTLSGroups = map[string]tls.CurveID{
+	"CurveP256":      tls.CurveP256,      // IANA 23
+	"CurveP384":      tls.CurveP384,      // IANA 24
+	"CurveP521":      tls.CurveP521,      // IANA 25
+	"X25519":         tls.X25519,         // IANA 29
+	"X25519MLKEM768": tls.X25519MLKEM768, // IANA 4588
+	// TODO: add "SecP256r1MLKEM768": tls.SecP256r1MLKEM768 (IANA 4587) when Go 1.26 is the minimum version
```
smith-xyz: Might be antipattern, but we could reduce all these TODOs by creating a separate file with something like: 

```go
//go:build go1.26

package crypto

import (
	"crypto/tls"

)

func init() {
	goTLSGroups["SecP256r1MLKEM768"] = tls.SecP256r1MLKEM768
	goTLSGroups["SecP384r1MLKEM1024"] = tls.SecP384r1MLKEM1024

	tlsGroupToCurveID[configv1.TLSGroupSecP256r1MLKEM768] = tls.SecP256r1MLKEM768
	tlsGroupToCurveID[configv1.TLSGroupSecP384r1MLKEM1024] = tls.SecP384r1MLKEM1024
}
```

Not sure if we normally leverage this in the project, just trying to think of ways to not have to come back to the TODOs 😄

joelanford: Seems like a good idea to me. This would make it possible for projects using older go to continue using newer library-go.

pkg/crypto/crypto_test.go
```diff
@@ -36,13 +37,17 @@ func TestConstantMaps(t *testing.T) {
 	}
 	discoveredVersions := map[string]bool{}
 	discoveredCiphers := map[string]bool{}
+	discoveredGroups := map[string]bool{}
 	for _, declName := range pkg.Scope().Names() {
```
joelanford, line 104: I wonder if we should have a `default` kind of case here, where we register all other known unrelated constants in the package. That way if a new constant is added that doesn't match one of our existing patterns, this test fails and we are forced to decide:
- is this a new version?
- is this a new cipher?
- is this a new group?
- is it unrelated

And add it to the respective set.

pkg/crypto/crypto.go
```diff
@@ -356,6 +389,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
 	return ianaCiphers
 }
 
+// CurveIDForTLSGroup maps an OpenShift API TLSGroup constant to Go's tls.CurveID.
+// Returns (0, false) if the group is not supported by this Go version.
```
davidesalerno: I'll change this comment since the map `tlsGroupToCurveID` includes raw IDs for groups that are not actually implemented in Go 1.25.
So mapped/recognized by library-go is more accurate than “supported by this Go version.”

pkg/crypto/crypto.go
```diff
@@ -356,6 +389,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
...
+			unsupported = append(unsupported, string(group))
+			continue
+		}
+		curves = append(curves, id)
+	}
+
+	return curves, unsupported
+}
+
+// ValidTLSGroups returns the TLS group names that this Go version supports,
```
davidesalerno: Same concern shared for `CurveIDForTLSGroup` related to the fact that `tlsGroupToCurveID` includes `tls.CurveID(4587)` and `tls.CurveID(4590)`.

pkg/crypto/crypto.go
```diff
@@ -211,6 +211,39 @@ var ciphersUnsupportedByGo = map[string]string{
 	"AES256-SHA256":             "TLS_RSA_WITH_AES_256_CBC_SHA256",
 }
 
+// goTLSGroups lists all TLS supported groups (key exchange mechanisms) recognized
+// by Go's crypto/tls, keyed by Go constant name. Kept in sync with the crypto/tls
+// package by TestConstantMaps.
+// Ref: https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
+var goTLSGroups = map[string]tls.CurveID{
```
davidesalerno: `goTLSGroups` looks test-only in this change. I think it would be cleaner to move it into `crypto_test.go` so production code only contains runtime mapping logic, while the completeness mirror of `crypto/tls` stays with the test that uses it.

pkg/crypto/crypto_test.go
davidesalerno: When Go 1.26 adds named constants like `tls.SecP256r1MLKEM768 `/ `tls.SecP384r1MLKEM1024`, this test will miss them and won’t force `goTLSGroups` to be updated. 

Is it this intentional?

tmshort:
This doesn't add curve changes to the TLS observer

damdo:
> This doesn't add curve changes to the TLS observer

@tmshort This is intentional. I want to keep them separate, we'll merge this one first and follow up with just the observer addition in your PR.

tmshort:
> This is intentional. I want to keep them separate, we'll merge this one first and follow up with just the observer addition in your PR.

@damdo Have you addressed all the comments on this PR?

No changes have been made for over a month.

pkg/crypto/crypto.go
tmshort: The golang 1.26.5 source has this value as 4589

damdo: It is, good catch! Thanks @smith-xyz @tmshort

pkg/crypto/crypto.go
damdo: TODOs are going away now that I'm rebasing onto 1.26, so I think we don't need to do this, but thanks 👍

tmshort: golang 1.26.5 is now the default builder for openshift-5.0

pkg/crypto/crypto_test.go
damdo: Yes I addressed it with the 1.26 TODOs, added `strings.HasPrefix(declName, "MLKEM")`. `MLKEM` prefix doesn't exist per-se, it is part of `MLKEM`.

damdo: see: https://github.com/openshift/library-go/pull/2347/changes#r3714038258

pkg/crypto/crypto.go
damdo: Good call 👍

pkg/crypto/crypto.go
damdo: Now that I dropped the raw IDs given the 1.26 upgrade I think we can keep the comment unchanged. TY

pkg/crypto/crypto.go
damdo: https://github.com/openshift/library-go/pull/2347/changes#r3714055549

pkg/crypto/crypto.go
```diff
@@ -211,6 +211,39 @@ var ciphersUnsupportedByGo = map[string]string{
...
+// tls.SecP384r1MLKEM1024 once Go 1.26 is the minimum version, and add
+// the corresponding entries to goTLSGroups.
+var tlsGroupToCurveID = map[configv1.TLSGroup]tls.CurveID{
+	configv1.TLSGroupX25519:             tls.X25519,
+	configv1.TLSGroupSecP256r1:          tls.CurveP256,
+	configv1.TLSGroupSecP384r1:          tls.CurveP384,
+	configv1.TLSGroupSecP521r1:          tls.CurveP521,
+	configv1.TLSGroupX25519MLKEM768:     tls.X25519MLKEM768,
+	configv1.TLSGroupSecP256r1MLKEM768:  tls.CurveID(4587), // TODO: replace with tls.SecP256r1MLKEM768 when Go 1.26 is the minimum version
+	configv1.TLSGroupSecP384r1MLKEM1024: tls.CurveID(4590), // TODO: replace with tls.SecP384r1MLKEM1024 when Go 1.26 is the minimum version
```
tmshort:
```suggestion
	configv1.TLSGroupSecP384r1MLKEM1024: tls.CurveID(4589), // TODO: replace with tls.SecP384r1MLKEM1024 when Go 1.26 is the minimum version
```
4589 is the correct value

tmshort: (But this will be replaced with the Go 1.26 value)

pkg/crypto/crypto_test.go
damdo: Done, LMK if it looks reasonable to you, thanks!

pkg/crypto/crypto_test.go
```diff
@@ -598,3 +614,113 @@ func TestCiphersUnsupportedByGoAreActuallyUnsupported(t *testing.T) {
...
+		wantID tls.CurveID
+		wantOK bool
+	}{
+		{configv1.TLSGroupX25519, tls.X25519, true},
+		{configv1.TLSGroupSecP256r1, tls.CurveP256, true},
+		{configv1.TLSGroupSecP384r1, tls.CurveP384, true},
+		{configv1.TLSGroupSecP521r1, tls.CurveP521, true},
+		{configv1.TLSGroupX25519MLKEM768, tls.X25519MLKEM768, true},
+		{configv1.TLSGroupSecP256r1MLKEM768, tls.CurveID(4587), true},  // TODO: replace with tls.SecP256r1MLKEM768 when Go 1.26 is the minimum version
+		{configv1.TLSGroupSecP384r1MLKEM1024, tls.CurveID(4590), true}, // TODO: replace with tls.SecP384r1MLKEM1024 when Go 1.26 is the minimum version
```
tmshort: This should be 4589, but will be changed to a proper constant when 1.26 is used.
```suggestion
		{configv1.TLSGroupSecP384r1MLKEM1024, tls.CurveID(4589, true}, // TODO: replace with tls.SecP384r1MLKEM1024 when Go 1.26 is the minimum version
```

damdo: Done, thanks

pkg/crypto/crypto.go
damdo: Done, thanks

pkg/crypto/crypto_test.go
```diff
@@ -36,13 +100,26 @@ func TestConstantMaps(t *testing.T) {
 	}
 	discoveredVersions := map[string]bool{}
 	discoveredCiphers := map[string]bool{}
+	discoveredGroups := map[string]bool{}
+	classified := sets.New[string]()
```
tmshort: `classified` is defined here, and appended to on line 122; but it's not used anywhere else. This can probably be removed; or the code that uses it needs to be added.

pkg/crypto/crypto.go
tmshort: I agree with @davidesalerno here.

tmshort (commented):
Some additional comments.

pkg/crypto/crypto_test.go
tmshort: And this was resolved while I was going over the rest of the code.

tmshort:
I have one comment above, but otherwise LGTM

jp8x7rgh6j-cmd (approved):
image

pkg/crypto/crypto.go
damdo: Done

pkg/crypto/crypto.go
damdo: Done

pkg/crypto/crypto.go
damdo: Done

pkg/crypto/crypto_test.go
damdo: Done

pkg/crypto/crypto.go
damdo: Done

damdo:
Hey all, I've addressed the feedback, PR is ready for labelling, thanks!

/assign @p0lyn0mial @bertinatto

pkg/crypto/crypto.go
```diff
@@ -356,6 +367,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
...
+// Returns (0, false) if the group is not recognized by library-go.
+func CurveIDForTLSGroup(group configv1.TLSGroup) (tls.CurveID, bool) {
+	id, ok := tlsGroupToCurveID[group]
+	return id, ok
+}
+
+// CurveIDsForTLSGroups converts a slice of TLSGroup values to their tls.CurveID
+// codes. Returns the supported curve IDs and a list of unsupported group names.
+// Unsupported groups are silently filtered — callers should log warnings.
+func CurveIDsForTLSGroups(groups []configv1.TLSGroup) ([]tls.CurveID, []string) {
```
davidesalerno: I think that we should evaluate to return` ([]tls.CurveID, []configv1.TLSGroup) `so that we can remove the type cast from T`LSGroup` to `string` at line 387, in this way if `TLSGroup` will became a struct in future in OpenShift API that won't support automatic type cast we won't need any change here.

pkg/crypto/crypto.go
```diff
@@ -356,6 +367,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
...
+		}
+		curves = append(curves, id)
+	}
+
+	return curves, unsupported
+}
+
+// ValidTLSGroups returns the TLS group names that this Go version supports,
+// sorted alphabetically.
+func ValidTLSGroups() []string {
```
davidesalerno: Same concern here to returning an array of `string` instead of an array of `TLSGroup`

pkg/crypto/crypto.go
```diff
@@ -356,6 +367,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
 	return ianaCiphers
 }
 
+// CurveIDForTLSGroup maps an OpenShift API TLSGroup constant to Go's tls.CurveID.
+// Returns (0, false) if the group is not recognized by library-go.
+func CurveIDForTLSGroup(group configv1.TLSGroup) (tls.CurveID, bool) {
```
sanchezl:
```suggestion
// TLSGroupToCurveID returns the Go tls.CurveID for the OpenShift API TLSGroup, or false if not recognized.
func TLSGroupToCurveID(group configv1.TLSGroup) (tls.CurveID, bool) {
```
Can we flip the name to give a better sense of the mapping aspect of this? (bonus: matches underlying map name)

pkg/crypto/crypto.go
sanchezl:
```suggestion
func TLSGroupsToCurveIDs(groups []configv1.TLSGroup) ([]tls.CurveID, []configv1.TLSGroup) {
```
I also agree that we should not perform the lossy string cast.

pkg/crypto/crypto.go
```diff
@@ -356,6 +367,43 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
...
+// ValidTLSGroups returns the TLS group names that this Go version supports,
+// sorted alphabetically.
+func ValidTLSGroups() []string {
+	groups := make([]string, 0, len(tlsGroupToCurveID))
+	for g := range tlsGroupToCurveID {
+		groups = append(groups, string(g))
+	}
+	sort.Strings(groups)
+	return groups
+}
```
sanchezl: I see where this mirrors the `ValidCipherSuites()` func but, as it is only used for logging, I don't see why we need to export either of them. If we do export, it should not cast away the type.

sanchezl (changes requested):

pkg/crypto/crypto.go
damdo: Sounds like a good suggestion, thanks! will do

pkg/crypto/crypto.go
damdo: Ok yes, let's uncast the type. I'll keep it exported for consistency with `ValidCipherSuites()`

damdo: Done

pkg/crypto/crypto.go
damdo: Done

pkg/crypto/crypto.go
damdo: Done

pkg/crypto/crypto.go
damdo: Done

sanchezl (approved):
/lgtm

tmshort:
/approve
(Not that I have the "rights")

pkg/crypto/crypto.go
```diff
@@ -356,6 +369,44 @@ func OpenSSLToIANACipherSuites(ciphers []string) []string {
...
+
+// ValidTLSGroups returns the recognized TLS group names, sorted alphabetically.
+func ValidTLSGroups() []configv1.TLSGroup {
+	groups := make([]configv1.TLSGroup, 0, len(tlsGroupToCurveID))
+	for g := range tlsGroupToCurveID {
+		groups = append(groups, g)
+	}
+	slices.SortFunc(groups, func(a, b configv1.TLSGroup) int {
+		return strings.Compare(string(a), string(b))
+	})
```
bertinatto, line 406: nit: I think you could use `slices.Sort(groups)` here because `configv1.TLSGroup` is an "ordered" type

bertinatto (commented):
/approve

I reviewed these changes and, to the best of my knowledge, this PR LGTM.

I checked the enhancement created in https://github.com/openshift/enhancements/pull/1894 and I understand why we need a mechanism to map the values we have in o/api to the go's std lib values. TLS groups are specified in the api, and that will be deserialized to our o/api types. Now we need a way to convert those types to their golang counterparts so that we can use them with the go std lib.

PR #2348, opened 2026-07-03
NO-JIRA: KMS Health Prober - hash nodename

tjungblu:
This adds a simple SHA-256 hash of the nodename, with additional reasoning as to why we couldn't use anything else at this point.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of long node names so configuration no longer fails because of field manager length limits.
  * Made resource update identities consistent and deterministic across runs, including for long names.

### DIFF
--- a/pkg/operator/encryption/kms/health/options.go
+++ b/pkg/operator/encryption/kms/health/options.go
@@ -2,12 +2,12 @@ package health
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"time"
 
 	"github.com/spf13/pflag"
 
-	metav1validation "k8s.io/apimachinery/pkg/apis/meta/v1/validation"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/clientcmd"
 )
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/kms/health/options.go
```diff
@@ -61,13 +61,20 @@ func (o *options) validate() error {
...
 	return nil
 }
 
+// fieldManagerForNode returns the SSA field manager for a reporter on nodeName.
+// Node names can exceed the fieldManager length limit, so the name is hashed to
+// a fixed-size SHA-256 hex digest while the prober still reports the real name.
+// The node UID would also fit, but it is not exposed through the downward API in
+// the sidecar; querying the Node API for it would add unnecessary complexity.
+func fieldManagerForNode(nodeName string) string {
+	sum := sha256.Sum256([]byte(nodeName))
```
ibihim, line 74: +1
We strive for uniqueness and don't want to handle collision-cases, neither does the fieldManager -> no fnv.

ibihim:
We can merge as is. The only objection I have is that we don't need to hash for 2 out of 5 cases (non-kube-apiserver-health-reporters).
But this should be fine as it is 2 out of 5 cases AND we don't rely on the assumption that there will always be just 1 instance of those apiservers in the future.

Right?

pkg/operator/encryption/kms/health/options.go
```diff
@@ -61,13 +61,20 @@ func (o *options) validate() error {
...
+	return fmt.Sprintf("%s-%x", Subcommand, sum)
```
ardaguclu, line 75: Did we check that what characters are allowed for field manager and this aligns with the generated sum here?

tjungblu: yes, hexadecimal values are allowed.

PR #2345, opened 2026-07-02
DO-NOT-MERGE: static pod deployer

tjungblu:
This is just an e2e, not destined for review, we pull individual refactorings out.

## Summary by CodeRabbit

* **New Features**
  * Added support for running KMS preflight on control-plane nodes via a static-pod workflow.
  * Introduced a one-shot installer Pod to prepare and render the preflight pod manifest and encryption configuration resources.
  * Improved static-pod behavior with conditional spec changes (host networking, kubeconfig wiring, and static pod resource mounting).

* **Refactor**
  * Streamlined KMS preflight pod template rendering to support both regular and static-pod variants, plus installer-pod templating.

* **Tests**
  * Expanded unit tests for template generation, deploy/status/cleanup flows, kubeconfig sourcing (cluster vs fallback), and installer creation failure handling.

PR #2350, opened 2026-07-07
NO-JIRA: update preflight pod manifest

tjungblu:
to facilitate static pods, this is a spin-off from #2345.

## Summary by CodeRabbit

* **New Features**
  * Added support for running KMS preflight checks in both standard and static pod modes.
  * Static pod mode now uses host networking and mounts the Kubernetes manifests directory automatically.

* **Bug Fixes**
  * Improved pod configuration so the correct service account and kubeconfig are applied based on the selected mode.
  * Updated pod setup to better match the required runtime permissions for static pod environments.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
+++ b/pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
@@ -12,7 +12,9 @@ spec:
   # This is a one-shot preflight check, not a long-running pod.
   # The operator creates it, waits for completion, and inspects the result.
   restartPolicy: Never
+{{- if not .StaticPod}}
   serviceAccountName: kms-preflight
+{{- end}}
   priorityClassName: system-cluster-critical
   nodeSelector:
     node-role.kubernetes.io/master: ""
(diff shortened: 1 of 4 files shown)


pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -12,7 +12,9 @@ spec:
   # This is a one-shot preflight check, not a long-running pod.
   # The operator creates it, waits for completion, and inspects the result.
   restartPolicy: Never
+{{- if not .StaticPod}}
   serviceAccountName: kms-preflight
+{{- end}}
```
tjungblu, line 17: not possible to use a service account anymore due to:
https://github.com/kubernetes/kubernetes/blob/cbe5e3096ec29950198ed0b77ed9c611f3183c14/pkg/kubelet/config/common.go#L151-L159

pkg/operator/encryption/kms/preflight/pod_template.go
```diff
@@ -19,38 +19,78 @@ type kmsPreflightTemplate struct {
...
+	staticPod bool,
+) (*corev1.Pod, error) {
+	rawManifest := mustAsset("assets/kms-preflight-pod.yaml")
+	tmplVal := newKmsPreflightTemplate(
+		podName, namespace, configHash, operatorImage, operatorCommand, kmsCallTimeout, staticPod,
+	)
+	return renderPodTemplate("kms-preflight", string(rawManifest), tmplVal)
+}
+
+func renderPodTemplate(name, rawManifest string, tmplVal any) (*corev1.Pod, error) {
```
tjungblu, line 92: just fyi, this is going to be reused for the installer pod, before you ask why there are three different functions now doing the same

PR #2351, opened 2026-07-08
NO-JIRA: add kms preflight installer pod

tjungblu:
to facilitate static pods in preflight, this is the second spin-off from #2345.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
+++ b/pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
@@ -1,0 +1,92 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: kms-preflight-installer
+  namespace: {{.Namespace}}
+  labels:
+    app: openshift-kms-preflight-installer
+spec:
+  automountServiceAccountToken: false
+  nodeSelector:
+    node-role.kubernetes.io/master: ""
+  tolerations:
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoExecute"
+  initContainers:
+    - name: cleanup
+      image: {{.InstallerImage}}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          rm -f /etc/kubernetes/manifests/kms-preflight-pod.yaml
+          rm -rf /etc/kubernetes/manifests/secrets/kms-preflight-encryption-config
+      securityContext:
+        privileged: true
+        runAsUser: 0
+      resources:
+        requests:
+          memory: 10Mi
+          cpu: 10m
+      volumeMounts:
+        - mountPath: /etc/kubernetes/manifests
+          name: manifests-dir
+  containers:
+    - name: installer
+      image: {{.InstallerImage}}
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          target_secret_dir="/etc/kubernetes/manifests/secrets/kms-preflight-encryption-config"
+          mkdir -p "${target_secret_dir}"
+          cp -f /install/secret/* "${target_secret_dir}/"
+          cp -f /install/pod/lb-int.kubeconfig "${target_secret_dir}/lb-int.kubeconfig"
+
+          manifest_path="/etc/kubernetes/manifests/kms-preflight-pod.yaml"
+          mkdir -p /etc/kubernetes/manifests
+          cp -f /install/pod/pod.yaml "${manifest_path}"
+          chmod 0600 "${manifest_path}"
+      securityContext:
+        privileged: true
+        runAsUser: 0
+      resources:
+        requests:
+          memory: 10Mi
+          cpu: 10m
+      volumeMounts:
+        - mountPath: /etc/kubernetes/manifests
+          name: manifests-dir
+        - mountPath: /install/pod
+          name: pod-manifest
+        - mountPath: /install/secret
+          name: encryption-config
+          readOnly: true
+  restartPolicy: Never
+  priorityClassName: system-node-critical
+  securityContext:
+    runAsUser: 0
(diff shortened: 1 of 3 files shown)


pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: v1
```
p0lyn0mial, line 1: how will it work ? which component will create the installer pod ?

tjungblu: check out 
https://github.com/openshift/library-go/pull/2345/changes#diff-e046906f51ab47b80bbb9ffebf90e9e9ec32038742331616ee94f61d45880e11R28

for KASO we configure the StaticPodPreflightDeployer instead of the existing pod deployer. That will spawn the installer on Deploy.

p0lyn0mial: do we have to schedule the installer always on the same node ?

tjungblu: we do not, where do you see that we schedule it on the same node?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
```diff
@@ -0,0 +1,92 @@
...
+      imagePullPolicy: IfNotPresent
+      terminationMessagePolicy: FallbackToLogsOnError
+      command:
+        - /bin/sh
+        - -c
+        - |
+          #!/bin/sh
+          set -euo pipefail
+
+          rm -f /etc/kubernetes/manifests/kms-preflight-pod.yaml
```
p0lyn0mial, line 31: so, after a successful run, nothing will remove the preflight pod from the node, right?
maybe the pod should terminate itself?

tjungblu: the pod terminates itself already. I assume we have to either build an "uninstaller" pod or have the preflight pod clean up after itself (somehow)

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
p0lyn0mial, line 1: what about reusing the existing machinery for installing the static pods ? not feasible ?

tjungblu: the most we could reuse is the static pod from here:
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/controller/installer/manifests/installer-pod.yaml


the static pod installer controller is unfortunately very coupled with the operator status and we don't really want to use a full controller reconciliation loop running for this, do we?

I briefly considered to reuse the installer cmd (we already have those in the operator image):
https://github.com/openshift/library-go/blob/master/pkg/operator/staticpod/installerpod/cmd.go

IMHO it's not really worth the hassle of integration, you see this is only a handful of lines in bash compared to fitting this in there. 

If you have a better idea, let me know. Happy to try out another approach.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
bertinatto: I think adding a clean-up `trap` in the `installer` container script would be enough, right? IIUC it'd clean-up after itself (after a successful run or not). This init container would still be useful for cases where the pod is forcefully killed

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
bertinatto: please correct me if I'm wrong, but I think the question was more like: is it OK if this runs on alternate nodes every time?

bertinatto: I wonder if we need to run an installer per node... I don't think we need, right? Does any of the checks we do are node-dependent?

tjungblu: > is it OK if this runs on alternate nodes every time?

sure, why not?

> I wonder if we need to run an installer per node... I don't think we need, right?

what's the value on installing it on every control plane node? do we  need to preflight check the same thing five times (worst case in a 5 node CP)

We also don't do that with the "normal" pod

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
tjungblu: I'll try the trap approach, but we need to change the arguments into a bash script.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
p0lyn0mial: > is it OK if this runs on alternate nodes every time?

yes. i think we don't have to run the installer per node. only one instance is enough. 

do we risk anything by running the pods on random nodes? For example, what if our changes cause a broader outage?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-installer-pod.yaml
p0lyn0mial: yeah, i would like to avoid another "uninstaller" pod. 

could the preflight pod remove its own manifest from disk ?

tjungblu: hah there's a catch-22 here: when we delete it from within the pod, the pod condition and our preflight result will be gone...

p0lyn0mial: what happens to a preflight pod that has completed or exited? Will it be restarted by kubelet? 

what happens to a preflight pod that completed on a machine that was rebooted? will it run again?

tjungblu: > what happens to a preflight pod that has completed or exited? Will it be restarted by kubelet?

it just completes, it will not be restarted.

> what happens to a preflight pod that completed on a machine that was rebooted? will it run again?

it will run again, yes. Also when you restart kubelet, as it reconciles the static pods and the completed status does not persist (afaik). I'll verify this once the cluster is running.

PR #2349, opened 2026-07-03
Adding KAS-O helpers and assertions

gangwgr:
Adding KAS-O helpers and assertions

## Summary by CodeRabbit

* **New Features**
  * Expanded encryption provider migration testing to support multiple scenarios per run, applying providers in the correct order and verifying results after each step.
  * Added KAS-focused test utilities for creating a consistent “secret-of-life” and reading its raw etcd content.
  * Added helper assertions to confirm expected encryption/decryption for Secrets and ConfigMaps, including last-migrated key verification.
* **Bug Fixes**
  * Strengthened migration scenario validation and verification logic, improving correctness of encrypted vs. unencrypted checks across provider and identity transitions.

PR #2352, opened 2026-07-08
CNTRLPLANE-3237: Add inParallel helper for running encryption test steps concurrently

p0lyn0mial:
could be used as a more generic mechanism to run arbitrary scenarios for https://github.com/openshift/library-go/pull/2349

## Summary by CodeRabbit

* **Tests**
  * Added new unit tests covering the parallel execution behavior of encryption test steps.
  * Improved validation of composed step naming and failure handling (including panics and immediate failure).
  * Added a concurrency timing check to confirm multiple steps run concurrently rather than sequentially.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/library/encryption/helpers_test.go
```diff
@@ -0,0 +1,133 @@
+package encryption
```
gangwgr, line 1: do we need this file ?

p0lyn0mial: yes. I would prefer having tests that ensure errors from individual runs will be reported and propagated to the caller.

test/library/encryption/helpers_test.go
```diff
@@ -0,0 +1,133 @@
...
+	runtime.Goexit()
+}
+
+func (r *recordingTB) Failed() bool { return r.failed.Load() }
+
+func countCompletions(steps []testStep, completed *atomic.Int32) []testStep {
+	wrapped := make([]testStep, len(steps))
+	for i, s := range steps {
+		orig := s.testFunc
+		wrapped[i] = testStep{name: s.name, testFunc: func(t testing.TB) {
```
gangwgr: instead of closing over the loop variable, can we consider passing s.testFunc into the closure.

test/library/encryption/helpers_test.go
```diff
@@ -0,0 +1,133 @@
...
+			combined := inParallel(countCompletions(tt.steps, &completed)...)
+			combined.testFunc(rec)
+
+			if combined.name != tt.expectName {
+				t.Errorf("expected name %q, actual %q", tt.expectName, combined.name)
+			}
+			if rec.Failed() != tt.expectFailed {
+				t.Errorf("expected Failed()=%v, actual %v", tt.expectFailed, rec.Failed())
+			}
+			if actual := completed.Load(); actual != tt.expectCompleted {
```
gangwgr, line 115: checking completed.Load() immediately may not give goroutines enough time to complete. 
can we add small grace time here?

p0lyn0mial: I think that wg.Wait() guarantees all goroutines will finish before inParallel returns.
Does it make sense ?

test/library/encryption/helpers.go
```diff
@@ -475,3 +476,35 @@ func WaitForCurrentKeyMigrated(t testing.TB, kubeClient kubernetes.Interface, pr
...
+						if r := recover(); r != nil {
+							t.Errorf("step %q panicked: %v", s.name, r)
+						}
+					}()
+					s.testFunc(t)
+				})
+			}
+			wg.Wait()
+		},
+	}
```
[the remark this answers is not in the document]
p0lyn0mial: @coderabbitai the sequential runner already checks t.Failed() after each step and stops execution. Adding t.FailNow() inside inParallel would be  problematic since FailNow must be called from the test goroutine, which is exactly what we're already doing in the runner loop. does it make sense ?

test/library/encryption/helpers_test.go
p0lyn0mial: pushed, see if you like it.

PR #2357, opened 2026-07-09
kms: remove TODO comment about resource requests in kms-preflight pod manifest

bertinatto:
The current values should be good enough for a one shot pod. These values are also used for pods that use the same binary, but different commands in KASO.

## Summary by CodeRabbit

* **Chores**
  * Removed an outdated note from the preflight pod template. No functional behavior or user-facing configuration changed.


p0lyn0mial:
@bertinatto  the pod will also have plugins attached. I think you've already evaluated the resource requirements for each plugin. Do you think the resource requirements we've defined today are good enough?

PR #2361, opened 2026-07-13
NO-JIRA: kms preflight virtualize a static pod deployment

tjungblu:
This PR configures only hostNetwork=true in the pod, to preflight check how it would behave as a static pod but without additional installer complexity.

## Summary by CodeRabbit

* **New Features**
  * Added support for a virtual static pod mode for KMS preflight checks.
  * Virtual static pod mode deploys preflight pods with host networking enabled.
* **Bug Fixes**
  * Static/virtual static pod configuration is simplified: consistently uses the designated service account, and removes static-pod-specific kubeconfig, privileged execution settings, and host-mounted resource access.
* **Tests**
  * Updated preflight deployment and pod template tests to cover both regular and virtual static pod modes.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -45,6 +45,8 @@ type PodPreflightDeployer struct {
 	operatorImage   string
 	operatorCommand []string
 	kmsCallTimeout  time.Duration
+	// virtualStaticPod will use settings like a static pod, e.g. hostNetwork=True, but run as a normal pod
+	virtualStaticPod bool
 }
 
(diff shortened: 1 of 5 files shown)


p0lyn0mial:
is this pr ready for review ?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -200,14 +203,16 @@ func NewPodPreflightDeployer(
 	operatorImage string,
 	operatorCommand []string,
 	kmsCallTimeout time.Duration,
+	virtualStaticPod bool,
```
p0lyn0mial: could we create a new function that would set that field to `true` ?
`NewStaticPodPreflightDeployer` ?

tjungblu: yes, updated the PR

PR #2363, opened 2026-07-14
kms: introduce WithDiskSecretName() to decouple encryption-config secret and mount point

bertinatto:
## Summary by CodeRabbit

* **New Features**
  * Static KMS plugin pods can now use a separate disk secret name for their reference data directory.
  * Added support for customizing the disk secret name during static pod configuration.

* **Bug Fixes**
  * Corrected static pod volume and mount paths when the disk secret differs from the encryption configuration secret.

* **Tests**
  * Added coverage for default and custom disk secret naming scenarios.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/builder.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/builder.go
@@ -21,6 +21,7 @@ type KMSPluginBuilder struct {
 	secretClient               corev1client.SecretsGetter
 	encryptionConfigNamespace  string
 	encryptionConfigSecretName string
+	diskSecretName             string
 	staticPod                  bool
 	errorIfNotFound            bool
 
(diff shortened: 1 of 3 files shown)


pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -48,6 +49,12 @@ func (b *KMSPluginBuilder) AsStaticPod() *KMSPluginBuilder {
 	return b
 }
 
+// WithDiskSecretName overrides the name used to compute the on-disk referenceDataDir path in static pod mode.
```
ardaguclu, line 52: Should we mention about if this is not called, encryptionConfigSecretName will be used as pathName?

ardaguclu:
Asked about a documentation update. Overall looks good to me. But once we merge this, operators will require to be updated which may negatively impact rebase process.

ardaguclu:
It looks like library-go and operators are on 1.36 already.
/lgtm

PR #2367, opened 2026-07-16
kms: add EnsureKMSPluginSidecar* to prune stale sidecars

bertinatto:
The existing Add* functions use upsert semantics, which leaves orphaned KMS sidecar containers, health reporters, and volumes when a provider is removed from the encryption configuration.

This PR adds EnsureKMSPluginSidecarInStaticPodSpec, which removes all KMS-managed resources before re-adding exactly what the current config requires.

Proofs:

* https://github.com/openshift/cluster-kube-apiserver-operator/pull/2231
* https://github.com/openshift/cluster-openshift-apiserver-operator/pull/733

## Summary by CodeRabbit

- **Bug Fixes**
  - KMS encryption resources are now reconciled more reliably when encryption settings change.
  - Stale KMS sidecars, initialization containers, volume mounts, and related volumes are automatically removed.
  - All KMS-managed resources are cleaned up when no KMS plugins are configured.
  - Conflicting socket mounts are corrected instead of causing pod specification errors.
- **Improvements**
  - KMS sidecar detection is more consistent across deployment and static-pod configurations.
  - Existing resources are updated to match current encryption requirements.

### DIFF
--- a/pkg/operator/encryption/kms/pluginlifecycle/builder.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/builder.go
@@ -107,6 +107,10 @@ func (b *KMSPluginBuilder) Apply(ctx context.Context, podSpec *corev1.PodSpec, c
 	if err != nil {
 		return fmt.Errorf("failed to get KMS configurations: %w", err)
 	}
+
+	// Strip all existing KMS-related resources before re-adding exactly what the current encryption config requires.
+	removeAllKMSManagedResources(podSpec, containerName)
+
 	if len(kmsConfigurations) == 0 {
 		klog.V(4).Infof("skipping KMS sidecar injection: no KMS plugins found in EncryptionConfiguration")
 		return nil
--- a/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
+++ b/pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
@@ -3,10 +3,12 @@ package pluginlifecycle
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/health"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -25,6 +27,11 @@ const (
 	kmsPluginSocketMountPath  = "/var/run/kmsplugin"
 )
 
+const (
+	vaultSidecarPrefix = "vault-kms-plugin"
+	healthReporterName = "kms-health-reporter"
+)
+
 // sidecarProvider abstracts the construction of a KMS plugin sidecar container for a specific provider (e.g. Vault).
 type sidecarProvider interface {
 	// Name returns the identifier used to name the sidecar container and locate its volume mounts.
@@ -38,13 +45,15 @@ type sidecarProvider interface {
 func newSidecarProvider(keyID string, udsPath string, pluginConfig configv1.KMSPluginConfig, refData *referenceDataResolver) (sidecarProvider, error) {
 	switch pluginConfig.Type {
 	case configv1.VaultKMSProvider:
-		return newVaultSidecarProvider("vault-kms-plugin", keyID, udsPath, pluginConfig.Vault, refData)
+		return newVaultSidecarProvider(vaultSidecarPrefix, keyID, udsPath, pluginConfig.Vault, refData)
 	default:
 		return nil, fmt.Errorf("unsupported KMS plugin configuration")
 	}
 }
 
-// AddKMSPluginSidecarToStaticPodSpec injects KMS plugin sidecar containers into a kube-apiserver static pod spec.
+// EnsureKMSPluginSidecarInStaticPodSpec reconciles KMS plugin sidecar containers in a kube-apiserver static pod spec.
+// It removes all KMS-managed resources (sidecars, volumes, volume mounts) and then re-adds exactly what the
+// current encryption config requires, ensuring stale resources from a previous configuration are pruned.
 //
 // Static pods access KMS plugin data through the resource-dir volume, which the static pod revision controller
 // populates on disk from the encryption-config Secret. Because those files are owned by root, each sidecar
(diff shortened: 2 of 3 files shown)


pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -172,6 +180,53 @@ func ensureReferenceDataVolume(podSpec *corev1.PodSpec, secretName string) error
...
+	for i, c := range podSpec.Containers {
+		if c.Name == containerName {
+			mounts := c.VolumeMounts[:0]
+			for _, m := range c.VolumeMounts {
+				if m.Name != kmsPluginSocketVolumeName {
+					mounts = append(mounts, m)
+				}
+			}
+			podSpec.Containers[i].VolumeMounts = mounts
+			break
```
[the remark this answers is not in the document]
bertinatto: we expect containers that have this mount volume to be added by the Ensure* function from this file — which means that if we add it, we need to remove it as well. This function has the expectation documented.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -69,7 +69,7 @@ func (b *KMSPluginBuilder) WithSecretRequired() *KMSPluginBuilder {
 // when empty, the subcommand is invoked directly via the image ENTRYPOINT.
 func (b *KMSPluginBuilder) WithHealthReporter(operatorBinary, operatorImage string) *KMSPluginBuilder {
 	b.healthReporter = &healthReporter{
-		name:           "kms-health-reporter",
+		name:           health.Subcommand,
```
tjungblu: we just have exactly one per pod, why change the name here? it's a bit strange to have the subcommand as the name too, also while checking it below it seems better to compare a constant name

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -172,6 +180,57 @@ func ensureReferenceDataVolume(podSpec *corev1.PodSpec, secretName string) error
...
+	}
+	return false
+}
+
+func isKMSManagedVolume(name string) bool {
+	return name == kmsPluginSocketVolumeName || name == referenceDataVolumeName
+}
+
+// removeAllKMSManagedResources removes all KMS-managed initContainers, volumes, and volume mounts from the pod spec.
+// IMPORTANT: When adding new KMS resource types (e.g., ephemeral containers, config maps), update this function
```
tjungblu, line 203: can you come up with a drift test for that?

e.g. creating an empty pod spec, running the create and then remove, ensure it's empty again

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
```diff
@@ -172,6 +180,57 @@ func ensureReferenceDataVolume(podSpec *corev1.PodSpec, secretName string) error
...
+// and the corresponding isKMSManaged* helper functions. Failure to do so breaks the pre-ready revision checks
+// that rely on this Ensure function to achieve the desired state.
+func removeAllKMSManagedResources(podSpec *corev1.PodSpec, containerName string) {
+	filtered := podSpec.InitContainers[:0]
```
tjungblu, line 207: that's a nifty trick I didn't know yet! thanks :)

pkg/operator/encryption/kms/pluginlifecycle/builder.go
```diff
@@ -107,6 +107,10 @@ func (b *KMSPluginBuilder) Apply(ctx context.Context, podSpec *corev1.PodSpec, c
 	if err != nil {
 		return fmt.Errorf("failed to get KMS configurations: %w", err)
 	}
+
+	// Strip all existing KMS-related resources before re-adding exactly what the current encryption config requires.
+	removeAllKMSManagedResources(podSpec, containerName)
```
tjungblu, line 112: could there be an ordering interaction with other operator controllers that add sidecars?
Let's say the targetconfigcontroller adds a cert rotation sidecar AFTER the kms containers:

Containers:
- KMS Plugin
- KMS Health
- Cert Rotation

Then removing will create
Containers:
- Cert Rotation

And ensuring will:
Containers:
- Cert Rotation
- KMS Plugin
- KMS Health


-> causing a new rollout, even though none of those technically changed
I know it's unlikely, but it depends on what the operator does with the other sidecars.

tjungblu: in a sense, it would be better to calculate a diff of the sidecars first? just pitching some alternatives

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: I agree with Thomas here. We can have a separate const for the name or leave it as is.

pkg/operator/encryption/kms/pluginlifecycle/builder.go
ardaguclu: Since we only care about init containers, I think order can not matter.

tjungblu: I was more after the potential of a revision flapping, but I just noticed we have this init container here:

https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/bindata/assets/kube-apiserver/pod.yaml#L15-L70

so it may happen that there is an old KAS pod that is still terminating (and handling requests?) when a new pod init container is running?

maybe I'm overthinking this

pkg/operator/encryption/kms/pluginlifecycle/builder.go
bertinatto: my bad, fixed

pkg/operator/encryption/kms/pluginlifecycle/sidecar.go
bertinatto: makes sense, added

pkg/operator/encryption/kms/pluginlifecycle/builder.go
bertinatto: after @tjungblu brought it up in Slack, I investigated this ordering issue, but ultimately considered it to be a non-issue because the only kube native sidecars that we have today are KMS-related.

do you think the Apply function should sort the init containers?

\* those that keep running after the kube-apiserver starts, unlike the `setup` init container linked above

ardaguclu: > do you think the Apply function should sort the init containers?

I think, this would be the correct way, even though it has no impact on functionality.

tjungblu: cool, then let's skip this line of thought. It makes no sense to sort here because the operator can add more sidecars on top, which then would change the order.

PR #2371, opened 2026-07-21
NO-JIRA: add foundation for a basic KMS preflight drift test

tjungblu:
This PR exposes constants for the test to consume, adds an e2e test for the preflight deployer, which in turn will assert the drift between the operand pod and the preflight pod it deployed.

Fake bump PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2233

## Release Notes

* **Bug Fixes / Improvements**
  * Standardized KMS encryption preflight resource identifiers (pod name, check container, encryption-config secret) so deploy, status, and cleanup consistently target the correct objects.

* **Tests**
  * Added end-to-end validation for KMS encryption preflight deployments, including secret/pod configuration checks and controller-reported results.
  * Updated and strengthened preflight deployment test utilities and pod/secret assertions to match canonical naming.

### DIFF
--- a/pkg/operator/encryption/kms/preflight/deployer.go
+++ b/pkg/operator/encryption/kms/preflight/deployer.go
@@ -19,11 +19,15 @@ import (
 )
 
 const (
-	preflightPodName                    = "kms-preflight"
-	preflightCheckContainerName         = "kms-preflight-check"
-	preflightEncryptionConfigSecretName = "kms-preflight-encryption-config"
-	preflightRBACName                   = "kms-preflight"
-	preflightAppLabel                   = "openshift-kms-preflight"
+	// PodName is the fixed name of the preflight pod created by Deploy.
+	PodName = "kms-preflight"
+	// CheckContainerName is the container that runs the preflight checker command.
+	CheckContainerName = "kms-preflight-check"
+	// EncryptionConfigSecretName is the secret Deploy creates for the preflight pod.
+	EncryptionConfigSecretName = "kms-preflight-encryption-config"
+
+	preflightRBACName = "kms-preflight"
+	preflightAppLabel = "openshift-kms-preflight"
 )
 
 var (
@@ -64,14 +68,14 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, en
 
 	encryptionConfigSecret = encryptionConfigSecret.DeepCopy()
 	// rewrite the entire ObjectMeta to avoid copying resource versions or managed fields
-	encryptionConfigSecret.ObjectMeta = metav1.ObjectMeta{Namespace: d.namespace, Name: preflightEncryptionConfigSecretName, Labels: labels}
+	encryptionConfigSecret.ObjectMeta = metav1.ObjectMeta{Namespace: d.namespace, Name: EncryptionConfigSecretName, Labels: labels}
 	_, err := d.coreClient.Secrets(d.namespace).Create(ctx, encryptionConfigSecret, metav1.CreateOptions{})
 	if err != nil {
 		return fmt.Errorf("failed to create preflight encryption config secret: %w", err)
 	}
 
 	pod, err := renderPreflightPodTemplate(
-		preflightPodName,
+		PodName,
 		d.namespace,
 		configHash,
 		d.operatorImage,
@@ -85,8 +89,8 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, en
 
 	err = pluginlifecycle.NewKMSPluginBuilder().
 		WithSecretRequired().
-		FromEncryptionConfigSecret(d.namespace, preflightEncryptionConfigSecretName, d.coreClient).
-		Apply(ctx, &pod.Spec, preflightCheckContainerName)
+		FromEncryptionConfigSecret(d.namespace, EncryptionConfigSecretName, d.coreClient).
+		Apply(ctx, &pod.Spec, CheckContainerName)
 	if err != nil {
 		return fmt.Errorf("failed to apply preflight plugin: %w", err)
 	}
@@ -108,9 +112,9 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, en
 
 func (d *PodPreflightDeployer) Status(ctx context.Context) (corev1.PodStatus, error) {
 	// preflight status checks are not very frequent, so we use the live client instead of a cached lister
-	pod, err := d.coreClient.Pods(d.namespace).Get(ctx, preflightPodName, metav1.GetOptions{})
+	pod, err := d.coreClient.Pods(d.namespace).Get(ctx, PodName, metav1.GetOptions{})
 	if err != nil {
-		return corev1.PodStatus{}, fmt.Errorf("failed to get pod for preflight %s/%s: %w", d.namespace, preflightPodName, err)
+		return corev1.PodStatus{}, fmt.Errorf("failed to get pod for preflight %s/%s: %w", d.namespace, PodName, err)
 	}
 
 	return pod.Status, nil
@@ -120,14 +124,14 @@ func (d *PodPreflightDeployer) Cleanup(ctx context.Context) error {
 	// See Deploy: preflight RBAC is intentionally not deleted here.
 	var errs []error
 
-	err := d.coreClient.Pods(d.namespace).Delete(ctx, preflightPodName, metav1.DeleteOptions{})
+	err := d.coreClient.Pods(d.namespace).Delete(ctx, PodName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		errs = append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, preflightPodName, err))
+		errs = append(errs, fmt.Errorf("failed to delete pod %s/%s: %w", d.namespace, PodName, err))
 	}
 
-	err = d.coreClient.Secrets(d.namespace).Delete(ctx, preflightEncryptionConfigSecretName, metav1.DeleteOptions{})
+	err = d.coreClient.Secrets(d.namespace).Delete(ctx, EncryptionConfigSecretName, metav1.DeleteOptions{})
 	if err != nil && !apierrors.IsNotFound(err) {
-		errs = append(errs, fmt.Errorf("failed to delete secret %s/%s: %w", d.namespace, preflightEncryptionConfigSecretName, err))
+		errs = append(errs, fmt.Errorf("failed to delete secret %s/%s: %w", d.namespace, EncryptionConfigSecretName, err))
 	}
 
 	return errors.Join(errs...)
--- a/test/library/encryption/preflight_deploy.go
+++ b/test/library/encryption/preflight_deploy.go
@@ -1,0 +1,198 @@
+package encryption
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	apiserverconfigv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/operator/encryption/controllers"
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/preflight"
+)
+
+const (
+	preflightDeployConfigHash = "e2e-preflight-drift"
+	// PreflightDeployCallTimeout is the KMS gRPC call timeout used by preflight e2e Deploy.
+	PreflightDeployCallTimeout  = 10 * time.Second
+	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms.sock"
+	preflightStatusPollInterval = 5 * time.Second
+	preflightStatusPollTimeout  = 5 * time.Minute
+	openshiftConfigNS           = "openshift-config"
+)
+
+// PreflightDeployScenario configures a real-cluster Deploy() check against a live
+// operand namespace. The operand namespace must already have preflight RBAC.
+//
+// BasicScenario.Namespace is the operand namespace (Deploy target).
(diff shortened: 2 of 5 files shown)


pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   # The operator creates it, waits for completion, and inspects the result.
   restartPolicy: Never
   serviceAccountName: kms-preflight
-  priorityClassName: system-cluster-critical
+  priorityClassName: system-node-critical
```
ardaguclu: Key controller will not take any action until getting a signal from preflight checker. Is the preflight checker pod so critical to have this className?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -19,11 +19,15 @@ import (
 )
 
 const (
-	preflightPodName                    = "kms-preflight"
-	preflightCheckContainerName         = "kms-preflight-check"
-	preflightEncryptionConfigSecretName = "kms-preflight-encryption-config"
-	preflightRBACName                   = "kms-preflight"
-	preflightAppLabel                   = "openshift-kms-preflight"
+	// PodName is the fixed name of the preflight pod created by Deploy.
```
ardaguclu, line 22: Nit: That would be awesome if those changes are in separate commit.

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,231 @@
...
+		if err := deployer.Cleanup(context.Background()); err != nil {
+			t.Errorf("preflight Cleanup: %v", err)
+		}
+	})
+	require.NoError(t, deployer.Deploy(ctx, preflightDeployConfigHash, secret))
+
+	preflightPod, err := clientSet.Kube.CoreV1().Pods(scenario.Namespace).Get(ctx, preflight.PodName, metav1.GetOptions{})
+	require.NoError(t, err, "preflight pod %s/%s", scenario.Namespace, preflight.PodName)
+
+	require.Equal(t, scenario.StaticPod, preflightPod.Spec.HostNetwork, "hostNetwork")
```
ardaguclu: Would it be better to compare it with json instead of one by one?

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,231 @@
...
+	require.Equal(t, corev1.ConditionTrue, resultCond.Status, "preflight check failed: %s", resultCond.Message)
+	require.Equal(t, "Succeeded", resultCond.Reason)
+
+	kekCond := findPodCondition(status.Conditions, controllers.KMSPreflightKEKIDPodCondition)
+	require.NotNil(t, kekCond, "missing %s condition", controllers.KMSPreflightKEKIDPodCondition)
+	require.Equal(t, corev1.ConditionTrue, kekCond.Status)
+	require.NotEmpty(t, kekCond.Message, "KEK ID")
+}
+
+func findPodCondition(conditions []corev1.PodCondition, condType corev1.PodConditionType) *corev1.PodCondition {
```
ardaguclu: I remember we have this already, or it was something different.

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,231 @@
...
+			return &conditions[i]
+		}
+	}
+	return nil
+}
+
+// OperatorImageFromDeployment returns the image of containerName from the named
+// Deployment. Used by preflight e2e to run the same operator binary the cluster
+// is already running, without requiring OPERATOR_IMAGE.
+func OperatorImageFromDeployment(ctx context.Context, t testing.TB, namespace, deploymentName, containerName string) string {
```
ardaguclu, line 131: Can't operators pass their images via argument instead of extracting in library-go?

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
tjungblu: it's not critical, but we wanted to keep the pod configurations _similar_, hence the same classname.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: I can create a separate PR as well, if you like

ardaguclu: No problem in this PR

test/library/encryption/preflight_deploy.go
tjungblu: the binary name, I would get, but how would the release image of the operand find its way into the test binary?

test/library/encryption/preflight_deploy.go
tjungblu: let me try to export something we already have :)

test/library/encryption/preflight_deploy.go
tjungblu: the host network you mean? the diffing logic isn't in this PR

ardaguclu: ah, now I saw your comment on Slack about diffing. Sounds good.

pkg/operator/encryption/kms/preflight/deployer.go
tjungblu: separated it into two, removed the priorityClassName change for now

test/library/encryption/preflight_deploy.go
tjungblu: no worries, I also left a todo in there :)

test/library/encryption/preflight_deploy.go
tjungblu: exported the one we have in preflight controller

test/library/encryption/preflight_deploy.go
ardaguclu: If I'm not mistaken, this test will be triggered by operator and we'll pass operator image like this https://github.com/openshift/cluster-kube-apiserver-operator/pull/2220/changes#diff-0d623dfd885adb20f991bda4c2453aebd732ca6dbb4d1d4be6e79805c3b48de6R360.

tjungblu: the code above runs inside the operator starter, which has the operator image defined in
https://github.com/openshift/cluster-kube-apiserver-operator/blob/main/manifests/0000_20_kube-apiserver-operator_06_deployment.yaml#L71-L72

and then injected via CVO at runtime during upgrades. 

The test you're reviewing will be executed inside an external cluster container that will only have kubeconfig access to the cluster you're testing against

ardaguclu: You are right. I thought this would be triggered by operator like unit test. But e2e tests are executed externally.

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,222 @@
...
+	openshiftConfigNS           = "openshift-config"
+)
+
+// PreflightDeployScenario configures a real-cluster Deploy() + drift check against a
+// live target operand pod. The operand namespace must already have preflight RBAC.
+//
+// BasicScenario.Namespace is the operand namespace (Deploy target and pod lookup).
+// BasicScenario.LabelSelector selects the live target operand pod to compare against.
+// Other BasicScenario fields are unused.
+type PreflightDeployScenario struct {
```
ardaguclu, line 36: What about if we update this to;

```go
 type PreflightDeployScenario struct {
      BasicScenario
      CreateDeployerFunc func(ctx context.Context, t testing.TB, clientSet ClientSet) *preflight.PodPreflightDeployer
      EncryptionProvider EncryptionProvider
  }
```

and operators call like this;

```go
PreflightDeployScenario{
      BasicScenario: basicScenario,
      CreateDeployerFunc: func(ctx context.Context, t testing.TB, cs ClientSet) *preflight.PodPreflightDeployer {
          image := OperatorImageFromDeployment(ctx, t, "openshift-kube-apiserver-operator", "kube-apiserver-operator", "kube-apiserver-operator")
          return preflight.NewStaticPodPreflightDeployer(ns, cs.Kube.CoreV1(), cs.Kube.RbacV1(), recorder, image, []string{"cluster-kube-apiserver-operator", "kms-preflight"}, timeout)
      },
      EncryptionProvider: kms.DefaultVaultEncryptionProvider(ctx, t),
  }
```
So that we don't need to have operatorimage, operatorcommand, staticpod. Everything will be created in that function.

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,222 @@
...
+
+	preflightPod, err := clientSet.Kube.CoreV1().Pods(scenario.Namespace).Get(ctx, preflight.PodName, metav1.GetOptions{})
+	require.NoError(t, err, "preflight pod %s/%s", scenario.Namespace, preflight.PodName)
+
+	require.Equal(t, scenario.StaticPod, preflightPod.Spec.HostNetwork, "hostNetwork")
+	require.Len(t, preflightPod.Spec.Containers, 1)
+	require.Equal(t, preflight.CheckContainerName, preflightPod.Spec.Containers[0].Name)
+	require.NotEmpty(t, preflightPod.Spec.InitContainers, "expected KMS plugin init container")
+
+	assertPreflightCheckConditions(ctx, t, deployer)
```
ardaguclu: Move all the assertions in to this function;

```go
 func AssertPreflightDeploy(ctx context.Context, t testing.TB, clientSet ClientSet, deployer *preflight.PodPreflightDeployer)
```
Update Preflight scenario like this;

```go
 type PreflightDeployScenario struct {
      BasicScenario
      CreateDeployerFunc func(ctx context.Context, t testing.TB, clientSet ClientSet) *preflight.PodPreflightDeployer
      AssertDeployFunc   func(ctx context.Context, t testing.TB, clientSet ClientSet, deployer *preflight.PodPreflightDeployer)
      EncryptionProvider EncryptionProvider
  }
```

and operators will call it like this;

```go
 PreflightDeployScenario{
      BasicScenario:      basicScenario,
      CreateDeployerFunc: func(...) { ... },
      AssertDeployFunc:   encryption.AssertPreflightDeploy,
      EncryptionProvider: provider,
  }
```

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,222 @@
...
+		if pods.Items[i].Status.Phase == corev1.PodRunning {
+			return &pods.Items[i]
+		}
+	}
+	require.FailNow(t, fmt.Sprintf("no Running pod in %s matching %q (found %d)", namespace, labelSelector, len(pods.Items)))
+	return nil
+}
+
+// TODO(thomas): this should go away once we have an official mapping function between plugin config and transient encryption config
+func mustPreflightDeployEncryptionConfigSecret(ctx context.Context, t testing.TB, clientSet ClientSet, namespace string, plugin configv1.KMSPluginConfig) *corev1.Secret {
```
ardaguclu: This function is Vault specific. It is better to rename it with Vault. I think, whenever new plugin comes in, there will be another similar function for it.

tjungblu: I'm actually banking on @bertinatto to save the day here ;) I'll rename it in a bit

test/library/encryption/preflight_deploy.go
tjungblu: will do!

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,206 @@
...
+		if c.Name == containerName {
+			require.NotEmpty(t, c.Image, "operator container %q image must not be empty", containerName)
+			return c.Image
+		}
+	}
+	require.FailNow(t, fmt.Sprintf("container %q not found in deployment %s/%s", containerName, namespace, deploymentName))
+	return ""
+}
+
+func mustGetRunningPod(ctx context.Context, t testing.TB, clientSet ClientSet, namespace, labelSelector string) *corev1.Pod {
```
ardaguclu: This is unused currently. Would it make sense removing this now, and add it back when it is needed?

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if scenario.EncryptionProvider.Setup != nil {
+		scenario.EncryptionProvider.Setup(ctx, t)
+	}
+
+	clientSet := GetClients(t)
+	deployer := scenario.CreateDeployerFunc(ctx, t, clientSet)
+	require.NotNil(t, deployer)
+
+	secret := mustVaultPreflightEncryptionConfigSecret(ctx, t, clientSet, scenario.Namespace, scenario.EncryptionProvider.KMS)
```
ardaguclu: Can we move this under AssertDeployFunction?

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,206 @@
...
+			return false, statusErr
+		}
+		if status.Phase == corev1.PodFailed {
+			return false, fmt.Errorf("preflight pod failed: reason=%q message=%q", status.Reason, status.Message)
+		}
+		return controllers.FindPodCondition(status.Conditions, controllers.KMSPreflightResultPodCondition) != nil, nil
+	})
+	require.NoError(t, err, "waiting for KMSPreflightResult condition")
+
+	hashCond := controllers.FindPodCondition(status.Conditions, controllers.KMSPreflightConfigHashPodCondition)
```
ardaguclu: Can we have utility function and reuse for all condition types?

tjungblu: applied

test/library/encryption/preflight_deploy.go
tjungblu: yes, applied

test/library/encryption/preflight_deploy.go
tjungblu: yep, removed it for now

test/library/encryption/preflight_deploy.go
```diff
@@ -0,0 +1,196 @@
...
+
+	if scenario.EncryptionProvider.Setup != nil {
+		scenario.EncryptionProvider.Setup(ctx, t)
+	}
+
+	clientSet := GetClients(t)
+	deployer := scenario.CreateDeployerFunc(ctx, t, clientSet)
+	require.NotNil(t, deployer)
+
+	secret := scenario.CreateEncryptionConfigFunc(ctx, t, clientSet, scenario.Namespace, scenario.EncryptionProvider.KMS)
```
ardaguclu, line 69: Assert notnil of secret

tjungblu: added

PR #2375, opened 2026-07-24
CNTRLPLANE-3237: kms: report preflight results via KMSEncryptionStatus

p0lyn0mial:
## Summary by CodeRabbit

* **New Features**
  * KMS preflight checks now persist the remote key ID in encryption status.
  * Completed preflight checks are reused when the KMS configuration hash hasn’t changed, reducing unnecessary pod activity.
* **Bug Fixes**
  * Preflight results are now more authoritative across retries, failures, and configuration changes.
  * Preflight pods are cleaned up immediately after successful completion; stale or mismatched preflight states are handled more reliably.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -134,9 +134,9 @@ const (
 	// in the condition message.
 	KMSPreflightResultPodCondition corev1.PodConditionType = "KMSPreflightResult"
 
-	// KMSPreflightKEKIDPodCondition carries the current key encryption key (KEK)
-	// ID reported by the KMS plugin during the preflight evaluation.
-	KMSPreflightKEKIDPodCondition corev1.PodConditionType = "KMSPreflightKekID"
+	// KMSPreflightRemoteKeyIDPodCondition carries the remote key ID reported by
+	// the KMS plugin during the preflight evaluation.
+	KMSPreflightRemoteKeyIDPodCondition corev1.PodConditionType = "KMSPreflightRemoteKeyID"
 
 	preflightPodRetention      = 1 * time.Hour
 	preflightPodStartupTimeout = 3 * time.Minute
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -139,7 +139,6 @@ const (
 	// the KMS plugin during the preflight evaluation.
 	KMSPreflightRemoteKeyIDPodCondition corev1.PodConditionType = "KMSPreflightRemoteKeyID"
 
-	preflightPodRetention      = 1 * time.Hour
 	preflightPodStartupTimeout = 3 * time.Minute
 )
 
@@ -369,7 +368,7 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 //     the startup timeout, return an error with the stuck reason.
 //
 //     e) Hash matches, KMSPreflightResult is True — check passed.
-//     Keep the pod for retention (1h), then clean up.
+//     Clean up the pod immediately.
 //
 //     f) Hash matches, KMSPreflightResult is False — check failed. Report
 //     degraded with the failure message. Keep the pod for inspection.
@@ -458,7 +457,7 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
 		}); err != nil {
 			return false, err
 		}
-		return false, c.cleanupAfterRetention(ctx, resultCondition.LastTransitionTime.Time)
+		return false, c.deployer.Cleanup(ctx)
 	}
 
 	// Scenario 3f: check failed. Keep pod for inspection; the admin will
@@ -490,15 +489,6 @@ func (c *kmsPreflightController) ensurePreflightResult(ctx context.Context, exis
 	})
 }
 
-func (c *kmsPreflightController) cleanupAfterRetention(ctx context.Context, completedAt time.Time) error {
-	age := time.Since(completedAt)
-	if age < preflightPodRetention {
-		klog.V(4).Infof("Preflight pod completed %s ago, keeping for inspection (retention %s)", age.Truncate(time.Second), preflightPodRetention)
-		return nil
-	}
-	klog.V(2).Infof("Preflight pod retention period elapsed (%s), cleaning up", preflightPodRetention)
-	return c.deployer.Cleanup(ctx)
-}
 
 type preflightError struct {
 	reason  string
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -386,7 +386,7 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 //
 // TODO: in the future we might want to add retries for failed preflights.
 func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeue bool, err error) {
-	requiredHash, err := c.preflightRequired(ctx)
+	requiredHash, existingResult, err := c.preflightRequired(ctx)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller_test.go
```diff
@@ -866,6 +943,10 @@ func TestKMSPreflightController(t *testing.T) {
 				t.Fatalf("unexpected error: %v", err)
 			}
 
+			if !reflect.DeepEqual(scenario.encryptionStatusProvider.writtenStatus, scenario.expectedKMSPreflightResult) {
```
ardaguclu: Can equality.semantics be used?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -581,14 +582,15 @@ func FindPodCondition(conditions []corev1.PodCondition, condType corev1.PodCondi
...
 func (c *kmsPreflightController) preflightRequired(ctx context.Context) (string, error) {
-	_, operatorStatus, _, err := c.operatorClient.GetOperatorState()
+	encryptionStatus, err := c.encryptionStatusProvider.GetKMSEncryptionStatus(ctx)
 	if err != nil {
-		return "", fmt.Errorf("failed to get operator state: %w", err)
+		return "", fmt.Errorf("failed to get KMS encryption status: %w", err)
 	}
-	requiredCondition := operatorv1helpers.FindOperatorCondition(operatorStatus.Conditions, "EncryptionKMSPreflightRequired")
-	if requiredCondition == nil || requiredCondition.Status != operatorv1.ConditionTrue {
+	requiredHash := encryptionStatus.Preflight.ObservedConfigHash
```
ardaguclu, line 620: Can encryptionStatus.Preflight be nil?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -458,7 +457,7 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
 		}); err != nil {
 			return false, err
 		}
-		return false, c.cleanupAfterRetention(ctx, resultCondition.LastTransitionTime.Time)
+		return false, c.deployer.Cleanup(ctx)
```
ardaguclu, line 480: I'm just thinking about the benefits of keeping this pod 1 hour, even in success cases. But probably there is no scenario cluster admin checks it. So this change makes sense to me.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -490,15 +489,6 @@ func (c *kmsPreflightController) ensurePreflightResult(ctx context.Context, exis
 	})
 }
 
-func (c *kmsPreflightController) cleanupAfterRetention(ctx context.Context, completedAt time.Time) error {
```
ardaguclu, line 447: Why do we remove this. Don't we retain the pod 1 hour in failed cases?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -431,27 +463,54 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
...
-	age := time.Since(completedAt)
-	if age < preflightPodRetention {
-		klog.V(4).Infof("Preflight pod completed %s ago, keeping for inspection (retention %s)", age.Truncate(time.Second), preflightPodRetention)
+// ensurePreflightResult writes result to KMSEncryptionStatus.Preflight.Result
+// if it has not already been written for this hash. It is a no-op when
+// existingResult is non-nil and its ConfigHash already matches result.ConfigHash,
+// making it safe to call on every sync without overwriting a previously stored outcome.
+func (c *kmsPreflightController) ensurePreflightResult(ctx context.Context, existingResult *operatorv1.KMSPreflightResult, result operatorv1.KMSPreflightResult) error {
+	if existingResult != nil && existingResult.ConfigHash == result.ConfigHash {
```
ardaguclu, line 510: Idempotency is only for configHash. So status field can be updated with the given hash?. Should we document this or should we add status in the idempotency?

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
```diff
@@ -329,6 +332,40 @@ func (f *fakeDeployer) Cleanup(_ context.Context) error {
 	return f.cleanupErr
 }
 
+type fakeEncryptionStatusProvider struct {
+	observedConfigHash string
+	writtenStatus      *operatorv1.KMSPreflightResult
+	updateCallCount    int
+	updateErr          error
```
ardaguclu, line 339: We don't assert this, do we?

pkg/operator/encryption/kms/preflight/pod_status_test.go
```diff
@@ -39,8 +39,8 @@ func TestPodCheckConditions(t *testing.T) {
 	}
 
 	kekId := conditions[2]
```
ardaguclu:
```suggestion
	remoteKeyID := conditions[2]
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -540,39 +599,43 @@ func FindPodCondition(conditions []corev1.PodCondition, condType corev1.PodCondi
...
 	if currentHash != requiredHash {
-		klog.V(4).Infof("KMS config hash changed: required=%s, current=%s; waiting for the key-controller to post an updated condition", requiredHash, currentHash)
-		return "", nil
+		klog.V(4).Infof("KMS config hash changed: required=%s, current=%s; waiting for the key-controller to update ObservedConfigHash", requiredHash, currentHash)
+		return "", nil, nil
 	}
 
-	return requiredHash, nil
+	if encryptionStatus.Preflight.Result.ConfigHash == requiredHash {
+		return requiredHash, &encryptionStatus.Preflight.Result, nil
```
ardaguclu: Do we have to return as reference type?

ardaguclu (commented):
With respect to states, only a comment about the removal of cleanup retention. The rest is about code.

pkg/operator/encryption/controllers/kms_preflight_controller.go
tjungblu: must-gathers really, but for that we should maybe keep them around for 12h or so

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -383,10 +400,25 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
...
+
 	// Check whether a preflight pod already exists.
 	podStatus, err := c.deployer.Status(ctx)
-	// Scenario 2: no pod exists, deploy a new one.
+	// Scenario 2: no pod exists.
 	if apierrors.IsNotFound(err) {
+		// Result already recorded as Failed and the pod is gone: surface the error
+		// without re-deploying. The admin must fix the config (new hash) before a
+		// new check can run.
+		if isPreflightResult(existingResult, operatorv1.KMSPreflightResultFailed) {
```
tjungblu: can you add two functions helpers where this is more explicit in the name `PreflightResultSuceeded`  /  `PreflightResultFailed`

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: `encryptionStatus.Preflight` is a value type

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: The failed pod is retained forever this was about pods that passed. I think it didn't make sense. Thus I decided to simply the code. The failure case is more important for which the pod will be kept until the "next" preflight.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: the logic here was that for a given hash status should not "flap"(on/off). does it make sense ?

```
  // Idempotency is keyed on ConfigHash only, not Status, because for a given hash
  // the result status is stable: a pod posts its result condition once and does not
  // change it. Success pods are cleaned up immediately; failure pods are retained
  // until the config changes (new hash), so the same hash cannot produce both
  // Succeeded and Failed.
  ```

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: we don't have to. I can change it. 

in general it was ok because the client returns a new struct on each invocation.

pkg/operator/encryption/controllers/kms_preflight_controller_test.go
p0lyn0mial: sure, we can.

ardaguclu:
Thank you
/lgtm
/hold
for @tjungblu looks at

tjungblu:
/hold cancel

All good! Thanks @p0lyn0mial

PR #2377, opened 2026-07-24
[WIP] kms: compute dry-run encryption config secret for preflight pod

bertinatto:
When deploying a new KMS preflight pod, the preflight controller now computes the required encryption configuration Secret using a dry-run sandbox.

## Summary by CodeRabbit

* **New Features**
  * KMS preflight now computes the managed `encryption-config` via a dry-run and deploys the preflight workload using the computed, validated result.
  * Preflight reconciles until encryption state converges while continuing to rewrite KMS provider endpoints to the preflight socket.
* **Bug Fixes**
  * Prevents deploying the preflight workload with a missing/`nil` `encryption-config` when no preflight pod exists.
* **Tests**
  * Added unit tests for dry-run `encryption-config` generation and for the dry-run secrets overlay.
  * Updated controller tests to verify the generated `encryption-config` is passed to the deployer.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/encryption/kms"
+	"github.com/openshift/library-go/pkg/operator/encryption/statemachine"
 	"github.com/openshift/library-go/pkg/operator/events"
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 )
@@ -165,13 +166,17 @@ type KMSPreflightDeployer interface {
 }
 
 type kmsPreflightController struct {
-	controllerInstanceName string
+	controllerInstanceName   string
+	instanceName             string
+	unsupportedConfigPrefix  []string
+	encryptionSecretSelector metav1.ListOptions
 
 	operatorClient  operatorv1helpers.OperatorClient
 	apiServerClient configv1client.APIServerInterface
 	coreClient      corev1client.CoreV1Interface
 
 	deployer                 KMSPreflightDeployer
+	encryptionDeployer       statemachine.Deployer
 	provider                 Provider
 	preconditionsFulfilledFn preconditionsFulfilled
 	encryptionStatusProvider kms.EncryptionStatusProvider
@@ -249,28 +254,39 @@ type kmsPreflightController struct {
 // so that its logs can be inspected, then cleaned up by a subsequent sync.
 func NewKMSPreflightController(
 	instanceName string,
+	unsupportedConfigPrefix []string,
 	provider Provider,
 	preconditionsFulfilledFn preconditionsFulfilled,
 	deployer KMSPreflightDeployer,
+	// encryptionDeployer is the same statemachine.Deployer used by the key and state
+	// controllers. It is consulted once when seeding the dry-run that produces the
+	// encryption-config Secret for the preflight pod.
+	encryptionDeployer statemachine.Deployer,
 	operatorClient operatorv1helpers.OperatorClient,
 	apiServerClient configv1client.APIServerInterface,
 	apiServerInformer configv1informers.APIServerInformer,
 	// coreClient reads referenced Secrets and ConfigMaps in openshift-config for hash
-	// computation. No informer is needed: the key-controller detects config changes and
-	// updates ObservedConfigHash, which triggers this controller via the operatorClient
+	// computation and seeds the key/state dry-run sandbox. No informer is needed for
+	// openshift-config: the key-controller detects config changes and updates
+	// ObservedConfigHash, which triggers this controller via the operatorClient
 	// informer. The minute-based resync covers the rest.
 	coreClient corev1client.CoreV1Interface,
+	encryptionSecretSelector metav1.ListOptions,
 	encryptionStatusProvider kms.EncryptionStatusProvider,
 	eventRecorder events.Recorder,
 ) factory.Controller {
 	c := &kmsPreflightController{
-		controllerInstanceName: factory.ControllerInstanceName(instanceName, "EncryptionKMSPreflight"),
+		controllerInstanceName:   factory.ControllerInstanceName(instanceName, "EncryptionKMSPreflight"),
+		instanceName:             instanceName,
+		unsupportedConfigPrefix:  unsupportedConfigPrefix,
+		encryptionSecretSelector: encryptionSecretSelector,
 
 		operatorClient:  operatorClient,
 		apiServerClient: apiServerClient,
 		coreClient:      coreClient,
 
 		deployer:                 deployer,
+		encryptionDeployer:       encryptionDeployer,
 		provider:                 provider,
 		preconditionsFulfilledFn: preconditionsFulfilledFn,
 		encryptionStatusProvider: encryptionStatusProvider,
@@ -356,7 +372,9 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 //     2a. Result already recorded as Failed and pod is gone: surface the error
 //     without re-deploying. The admin must fix the config (new hash) before
 //     a new check can run.
-//     2b. No result yet: call Deploy. On success, requeue and wait for the pod to report results.
+//     2b. No result yet: compute the encryption-config Secret via dry-run, then
+//     call Deploy. On success, requeue and wait for the pod to report results.
+//     If the encryption deployer has not converged, requeue and wait before Deploy.
 //
 //  3. Preflight required, pod exists (Status returns a PodStatus).
 //     Evaluate the pod state via conditions and phase:
@@ -408,8 +426,8 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 //	1         No preflight required — cleanup                   false    nil   False     False        No
 //	1a        Already Succeeded — cleanup, no pod work          false    nil   False     False        No
 //	2a        No pod, already Failed — surface error            false    *pe   True      False        Yes
-//	2b        No pod, Deploy success                            true     nil   False     True         No
-//	2b        No pod, Deploy error                              true     err   True      False        No
+//	2b        No pod, dry-run / Deploy success                  true     nil   False     True         No
+//	2b        No pod, dry-run / Deploy error                    true     err   True      False        No
 //	3a        Pod Failed — keep for inspection                  false    *pe   True      False        Yes
 //	3b        No hash, pod Running, no timeout                  true     nil   False     True         No
 //	3b        No hash, timeout exceeded                         true     *pe   True      False        Yes
@@ -454,8 +472,28 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
 				message: fmt.Sprintf("preflight check failed for hash %s: pod was removed but failure is recorded in status", requiredHash),
 			}
 		}
-		// TODO: compute the encryption configuration and pass it to the deployer
-		if err := c.deployer.Deploy(ctx, requiredHash, nil); err != nil {
+		requeueForConvergence, encryptionSecret, err := computeEncryptionConfigSecretDryRun(
+			ctx,
+			c.instanceName,
+			c.unsupportedConfigPrefix,
+			c.provider,
+			c.encryptionDeployer,
+			c.operatorClient,
+			c.apiServerClient,
+			c.coreClient,
+			c.encryptionSecretSelector,
+		)
+		if err != nil {
+			return true, "", "", fmt.Errorf("failed to compute encryption config for preflight: %w", err)
+		}
+		// The preflight pod must test the actual encryption config that will be deployed
+		// to the API server. If the deployer hasn't converged (e.g., a new revision is
+		// still rolling out), the dry-run won't produce the correct config. Wait for
+		// convergence before deploying the preflight workload.
+		if requeueForConvergence {
+			return true, "RunningPreflightCheck", "Waiting for encryption deployer to converge before computing preflight encryption config", nil
+		}
+		if err := c.deployer.Deploy(ctx, requiredHash, encryptionSecret); err != nil {
 			return true, "", "", err
 		}
 		return true, "RunningPreflightCheck", fmt.Sprintf("Deploying preflight pod for hash %s", requiredHash), nil
--- a/pkg/operator/encryption/controllers/kms_preflight_overlay.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_overlay.go
@@ -1,0 +1,153 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	applyconfigurationscorev1 "k8s.io/client-go/applyconfigurations/core/v1"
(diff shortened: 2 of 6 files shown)


pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,422 @@
...
+	operatorClient operatorv1helpers.OperatorClient,
+	apiServerClient configv1client.APIServerInterface,
+	coreClient corev1client.CoreV1Interface,
+	encryptionSecretSelector metav1.ListOptions,
+) (requeue bool, secret *corev1.Secret, err error) {
+	liveDeployed, converged, err := encryptionDeployer.DeployedEncryptionConfigSecret(ctx)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed to get deployed encryption config: %w", err)
+	}
+	if !converged {
```
ardaguclu, line 49: Controllers do not reconcile, if apiservers are not converged OR migration is in progress. However, prelight should still work even in these conditions to fix the possible stuck issue.

ardaguclu: But that can be managed via new flags in each controller, i.e. continue when migration is in progress.

ardaguclu (commented):
This sandboxing looks brilliant to me. I just commented about a few items that logically differ between preflight and key controller.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -247,27 +252,39 @@ type kmsPreflightController struct {
...
 	preconditionsFulfilledFn preconditionsFulfilled,
 	deployer KMSPreflightDeployer,
+	// encryptionDeployer is the same statemachine.Deployer used by the key and state
+	// controllers. It is consulted once when seeding the dry-run that produces the
+	// encryption-config Secret for the preflight pod.
+	encryptionDeployer statemachine.Deployer,
 	operatorClient operatorv1helpers.OperatorClient,
 	apiServerClient configv1client.APIServerInterface,
 	apiServerInformer configv1informers.APIServerInformer,
+	kubeInformersForNamespaces operatorv1helpers.KubeInformersForNamespaces,
```
ardaguclu: Why do we have informers?. Wouldn't it be better to just reconcile periodically?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -387,8 +406,24 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
...
+			c.encryptionDeployer,
+			c.operatorClient,
+			c.apiServerClient,
+			c.coreClient,
+			c.encryptionSecretSelector,
+		)
+		if err != nil {
+			return false, fmt.Errorf("failed to compute encryption config for preflight: %w", err)
+		}
+		if requeueForConvergence {
```
ardaguclu, line 493: It is better to document why we requeue when there is no convergence for future readers.

pkg/operator/encryption/controllers/kms_preflight_overlay.go
```diff
@@ -0,0 +1,147 @@
...
+	return &dryRunSecretClient{
+		real:    g.real.Secrets(ns),
+		ns:      ns,
+		written: g.written,
+	}
+}
+
+// getWritten returns the captured secret for namespace/name, or an error if none exists.
+func (g *dryRunSecretsGetter) getWritten(namespace, name string) (*corev1.Secret, error) {
+	s, ok := g.written[namespace+"/"+name]
```
ardaguclu: Probably not needed but should we use mutex?

pkg/operator/encryption/controllers/kms_preflight_overlay.go
```diff
@@ -0,0 +1,147 @@
...
+func (c *dryRunSecretClient) Apply(_ context.Context, _ *applyconfigurationscorev1.SecretApplyConfiguration, _ metav1.ApplyOptions) (*corev1.Secret, error) {
+	return nil, fmt.Errorf("dry-run secrets client does not support Apply")
+}
+
+func (c *dryRunSecretClient) capture(secret *corev1.Secret) *corev1.Secret {
+	cp := secret.DeepCopy()
+	if cp.Namespace == "" {
+		cp.Namespace = c.ns
+	}
+	c.written[cp.Namespace+"/"+cp.Name] = cp
```
ardaguclu: should we use mutex?

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,228 @@
...
+
+	expectedName := fmt.Sprintf("%s-%s", encryptiondata.EncryptionConfSecretName, instanceName)
+	ecSecret, err := overlay.getWritten(openshiftConfigManagedNS, expectedName)
+	if err != nil {
+		return nil, fmt.Errorf("state controller dry-run did not create or update encryption config secret %q", expectedName)
+	}
+	return ecSecret, nil
+}
+
+func alwaysTruePreconditions() (bool, error) {
```
ardaguclu: nit: Do we really need this?

ardaguclu (commented):
Since we agree upon pursuing this mechanism, I dropped some comments about the code.

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
bertinatto: it's used to make sure the key/state controllers to always run; but I can also inline it when I create those controllers, PTAL

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
bertinatto: I see, so the first step would be to remove this early-return, and then patch the controllers to continue without convergence, is that correct?

pkg/operator/encryption/controllers/kms_preflight_overlay.go
bertinatto: as far as I investigated, it's not necessary because this is not shared between multiple goroutines. If you prefer, I could add a mutex, but currently it shouldn't be necessary

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: I think it'd be bad if the key controller would have to wait for 1 minute (the current resync time for this controller) to generate a key. But I think you're right that this particular informer shouldn't be needed.

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: added comment

pkg/operator/encryption/controllers/kms_preflight_overlay.go
bertinatto: see my previous comment

pkg/operator/encryption/controllers/kms_preflight_overlay.go
ardaguclu: I agree, we wouldn't need this.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -249,28 +254,39 @@ type kmsPreflightController struct {
...
 func NewKMSPreflightController(
 	instanceName string,
+	unsupportedConfigPrefix []string,
 	provider Provider,
 	preconditionsFulfilledFn preconditionsFulfilled,
 	deployer KMSPreflightDeployer,
+	// encryptionDeployer is the same statemachine.Deployer used by the key and state
+	// controllers. It is consulted once when seeding the dry-run that produces the
+	// encryption-config Secret for the preflight pod.
+	encryptionDeployer statemachine.Deployer,
```
ardaguclu, line 264: nit: Naming is confusing for me especially looking with the one line above.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: @p0lyn0mial since in-place field updates work even if there is no convergence, preflight should work too.

pkg/operator/encryption/controllers/kms_preflight_overlay.go
```diff
@@ -0,0 +1,162 @@
...
+	ns     string
+}
+
+var _ corev1client.SecretInterface = (*dryRunSecretClient)(nil)
+
+func (c *dryRunSecretClient) Get(ctx context.Context, name string, opts metav1.GetOptions) (*corev1.Secret, error) {
+	if s, ok := c.parent.written[c.ns][name]; ok {
+		return s.DeepCopy(), nil
+	}
+	return c.parent.real.Secrets(c.ns).Get(ctx, name, opts)
```
ardaguclu, line 68: result should not be written to c.parent.written back, because this does not work as cache

pkg/operator/encryption/controllers/kms_preflight_overlay.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (c *dryRunSecretClient) List(ctx context.Context, opts metav1.ListOptions) (*corev1.SecretList, error) {
+	selector, err := labels.Parse(opts.LabelSelector)
+	if err != nil {
+		return nil, err
+	}
+	list, err := c.parent.real.Secrets(c.ns).List(ctx, opts)
```
ardaguclu, line 76: Would make sense adding nil checks for c.parent.real and return error

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,233 @@
...
+	if cfg == nil || cfg.Encryption == nil {
+		return nil, fmt.Errorf("encryption configuration is empty")
+	}
+
+	wantOld := fmt.Sprintf("unix:///var/run/kmsplugin/kms-%d.sock", keyID)
+	rewrote := false
+	for i := range cfg.Encryption.Resources {
+		for j := range cfg.Encryption.Resources[i].Providers {
+			kms := cfg.Encryption.Resources[i].Providers[j].KMS
+			if kms == nil || kms.Endpoint != wantOld {
```
ardaguclu: it is better to compare against providerName instead of endpoint.

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,233 @@
...
+
+	created, err := recordedKeySecret(overlay, instanceName)
+	if err != nil {
+		return "", err
+	}
+	return created.Name, nil
+}
+
+// recordedKeySecret returns the single encryption-key secret recorded for instanceName.
+func recordedKeySecret(overlay *dryRunSecretsGetter, instanceName string) (*corev1.Secret, error) {
```
ardaguclu: This sounds like leaking the knowledge of encryption controllers. Instead of this function, can't we get keyID like this;

```go
keySecrets, err := secrets.ListKeySecrets(ctx, overlay, encryptionSecretSelector)
  if err != nil {
      return false, nil, err
  }

  var latestKeyID uint64
  for _, s := range keySecrets {
      if id, ok := state.NameToKeyID(s.Name); ok && id > latestKeyID {
          latestKeyID = id
      }
  }
```

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,233 @@
...
+
+// rewritePreflightWriteKeyEndpoint rewrites the write-key KMS endpoint from the
+// per-key production socket (kms-{id}.sock) to the fixed preflight socket
+// (kms.sock). Today's preflight checker dials that fixed path, while the plugin
+// builder uses each provider's Endpoint as -listen-address, so the dry-run
+// secret must agree with the checker.
+//
+// TODO: once preflight dials the write-key socket (kms-{id}.sock) directly,
+// this rewrite can be removed and the dry-run secret can be used as-is.
+func rewritePreflightWriteKeyEndpoint(secret *corev1.Secret, keyID uint64) (*corev1.Secret, error) {
```
ardaguclu, line 201: Instead of having this function, can't we call this in the caller;

```go
target := fmt.Sprintf("%d_", latestKeyID)
  for i := range cfg.Encryption.Resources {
      for j := range cfg.Encryption.Resources[i].Providers {
          if kms := cfg.Encryption.Resources[i].Providers[j].KMS; kms != nil && strings.HasPrefix(kms.Name, target) {
              kms.Endpoint = preflightKMSSocketEndpoint
          }
      }
  }
```

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: would it be better if we renamed `deployer` to `preflightDeployet` in a separate PR?

pkg/operator/encryption/controllers/kms_preflight_overlay.go
bertinatto: agreed; the object we get from the apiserver should not be stored in `written` (which should only contain objects that were tried to be 'written' [created/updated]). At least that's the idea, are you suggesting it's not currently doing that?

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: That works for me

pkg/operator/encryption/controllers/kms_preflight_overlay.go
ardaguclu: Current mechanism looks good to me

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: yes, i think so.

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
bertinatto: done

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
bertinatto: done, PTAL

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
bertinatto: if I understand correctly, that's just part of what this function do, the caller still needs to call `encryptiondata.FromSecret`. Inlining this in `computeEncryptionConfigSecretDryRun()` would make it even bigger. Also, the idea is that this would be dropped in the future, so a helpers makes that easier. WDYT? Am I misunderstanding it?

bertinatto: oh, and the caller would also need `encryptiondata.ToSecret`

pkg/operator/encryption/controllers/kms_preflight_overlay.go
bertinatto: I can add a nil check in the constructor (rather than checking it on every Get/List call. PTAL

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
ardaguclu: Yeah, it would be better to leave this func as is

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: so this needs to be removed

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
ardaguclu: yes, that is correct. I have 2 PRs for that https://github.com/openshift/library-go/pull/2393, https://github.com/openshift/library-go/pull/2391

ardaguclu (commented):
Overall looks good to me

ardaguclu:
https://github.com/openshift/library-go/pull/2410 supersedes this PR
/close

PR #2386, opened 2026-07-29
NO-JIRA: kms preflight: add reflection based drift detection

tjungblu:
This adds a basic diff between the preflight pod and the respective apiserver pod. We do this via reflection to avoid constantly changing the logic when new fields are added. There's also an ignore list where we can tag things we want to allow to diverge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption preflight checks by comparing the preflight pod spec against the matching currently running operand pod, selected via an explicit label selector.
  * Strengthened PodSpec drift detection with semantic comparison that tolerates known, intentional differences through a configurable ignore list.

* **Tests**
  * Added test coverage for the default ignored fields, custom ignore overrides, detection of unexpected PodSpec changes (such as host networking), and validation behavior when pods/specs don’t match.

### DIFF
--- a/test/library/encryption/preflight_deploy.go
+++ b/test/library/encryption/preflight_deploy.go
@@ -31,8 +31,8 @@ const (
 // PreflightDeployScenario configures a real-cluster Deploy() check against a live
 // operand namespace. The operand namespace must already have preflight RBAC.
 //
-// BasicScenario.Namespace is the operand namespace (Deploy target).
-// BasicScenario.LabelSelector is reserved for a follow-up operand pod drift check.
+// BasicScenario.Namespace is the operand namespace (Deploy target and pod lookup).
+// BasicScenario.LabelSelector selects an operand pod for AssertNoPreflightConfigDrift.
 type PreflightDeployScenario struct {
 	BasicScenario
 	// CreateDeployerFunc builds the deployer (image, command, static-pod mode).
@@ -48,10 +48,11 @@ type PreflightDeployScenario struct {
 }
 
 // TestPreflightDeployAndPodMatchesOperand deploys a preflight pod into the operand
-// namespace and runs AssertDeployFunc against it.
+// namespace, runs AssertDeployFunc, then AssertNoPreflightConfigDrift.
 func TestPreflightDeployAndPodMatchesOperand(ctx context.Context, t testing.TB, scenario PreflightDeployScenario) {
 	t.Helper()
 	require.NotEmpty(t, scenario.Namespace)
+	require.NotEmpty(t, scenario.LabelSelector)
 	require.NotNil(t, scenario.CreateDeployerFunc)
 	require.NotNil(t, scenario.CreateEncryptionConfigFunc)
 	require.NotNil(t, scenario.AssertDeployFunc)
@@ -77,10 +78,12 @@ func TestPreflightDeployAndPodMatchesOperand(ctx context.Context, t testing.TB, 
 
 	require.NoError(t, deployer.Deploy(ctx, preflightDeployConfigHash, secret))
 	scenario.AssertDeployFunc(ctx, t, clientSet, scenario.Namespace, deployer)
+	AssertNoPreflightConfigDrift(ctx, t, clientSet, scenario.Namespace, scenario.LabelSelector)
 }
 
 // AssertPreflightDeploy waits for checker pod conditions and validates Deploy side effects
-// (check container, plugin init container, config-hash / result / KEK-ID conditions).
+// (check container, plugin init container, config-hash / result / remote-key-ID conditions).
+// PodSpec drift vs the operand is AssertNoPreflightConfigDrift — call that separately.
 func AssertPreflightDeploy(ctx context.Context, t testing.TB, clientSet ClientSet, namespace string, deployer *preflight.PodPreflightDeployer) {
 	t.Helper()
 	require.NotNil(t, deployer)
--- a/test/library/encryption/preflight_drift.go
+++ b/test/library/encryption/preflight_drift.go
@@ -1,0 +1,91 @@
+package encryption
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/kms/preflight"
+)
+
+// DefaultPreflightPodSpecDriftIgnore lists PodSpec field names skipped by
+// AssertNoPreflightConfigDrift via cmpopts.IgnoreFields. Unknown diffs outside
+// this list fail the drift check.
+//
+// Keep entries as exported Go field names (e.g. "HostNetwork"), not JSON tags.
+// Prefer a short list: fields that are nil/zero on both sides already compare equal, so
+// only ignore fields we know differ by design between preflight and the operand.
+var DefaultPreflightPodSpecDriftIgnore = []string{
+	// --- Workload-owned: preflight is a one-shot checker, not the apiserver ---
+	// Preflight runs kms-preflight-check (+ KMS plugin init/sidecar); operands run apiserver containers.
+	"Containers",
+	"InitContainers",
+	"EphemeralContainers",
+	// Volume mounts follow the containers above (encryption-config, sockets, etc.).
+	"Volumes",
+	// Preflight uses RestartPolicy=Never; operands typically Always.
+	"RestartPolicy",
+	// Preflight uses its own SA (kms-preflight) vs the operand service account.
+	"ServiceAccountName",
+	"DeprecatedServiceAccount",
+	"AutomountServiceAccountToken",
+	// OpenShift injects the SA dockercfg onto the preflight pod; static KAS pods typically have none.
+	"ImagePullSecrets",
+	// One-shot job semantics; operands are long-running.
+	"ActiveDeadlineSeconds",
+	"TerminationGracePeriodSeconds",
+
+	// Filled by the scheduler on Running pods; preflight and operand will always share a control plane node one way or another.
+	"NodeName",
+	// Preflight sets master nodeSelector; static KAS pods are placed differently.
+	"NodeSelector",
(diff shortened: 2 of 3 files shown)


test/library/encryption/preflight_deploy.go
```diff
@@ -115,6 +117,8 @@ func AssertPreflightDeploy(ctx context.Context, t testing.TB, clientSet ClientSe
 
 	kekCond := requirePodCondition(t, status.Conditions, controllers.KMSPreflightRemoteKeyIDPodCondition, corev1.ConditionTrue)
 	require.NotEmpty(t, kekCond.Message, "remote key ID")
+
+	AssertPreflightPodSpecMatchesOperand(ctx, t, clientSet, namespace, labelSelector)
```
ardaguclu: AssertPreflightDeploy and AssertPreflightPodSpecMatchesOperand are different assertions. Would it make sense call it separately from the operators?. AssertPreflightDeploy focuses on the deployment and the other focuses on the drift.

ardaguclu: Can we rename the function shorter like `AssertNoPreflightConfigDrift`?

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+	// Preflight template still uses system-cluster-critical; operands use system-node-critical.
+	"PriorityClassName",
+	// KAS sets an explicit numeric priority; preflight leaves it to the PriorityClass.
+	"Priority",
+}
+
+// ComparePodSpecIgnoringDrift walks every exported PodSpec field via reflection and
+// reports fields where expected and actual differ, skipping ignoreFields.
+// Returns an empty string when there is no unexpected drift.
+func ComparePodSpecIgnoringDrift(expected, actual *corev1.PodSpec, ignoreFields []string) string {
```
ardaguclu: Can't we use this as simpler;

```go
  func ComparePodSpecIgnoringDrift(expected, actual *corev1.PodSpec, ignoreFields []string) string {
      return cmp.Diff(expected, actual,
          cmpopts.IgnoreFields(corev1.PodSpec{}, ignoreFields...),
          cmpopts.EquateEmpty(),
      )
  }
```

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+	"RestartPolicy",
+	// Preflight uses its own SA (kms-preflight) vs the operand service account.
+	"ServiceAccountName",
+	"DeprecatedServiceAccount",
+	"AutomountServiceAccountToken",
+	// One-shot job semantics; operands are long-running.
+	"ActiveDeadlineSeconds",
+	"TerminationGracePeriodSeconds",
+
+	// Filled by the scheduler on Running pods; preflight and operand rarely share a node.
```
tjungblu:
```suggestion
	// Filled by the scheduler on Running pods; preflight and operand will always share a control plane node one way or another.
```

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+		preflightPod.Namespace, preflightPod.Name, targetPod.Namespace, targetPod.Name, diff)
+}
+
+func mustGetRunningOperandPod(ctx context.Context, t testing.TB, clientSet ClientSet, namespace, labelSelector string) *corev1.Pod {
+	t.Helper()
+	pods, err := clientSet.Kube.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	require.NoError(t, err)
+	for i := range pods.Items {
+		if pods.Items[i].Status.Phase == corev1.PodRunning {
+			return &pods.Items[i]
```
ardaguclu, line 87: We are picking the first running pod, because there is only one pod with that label?

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+		require.FailNow(t, fmt.Sprintf("no Running pod in %s matching %q", namespace, labelSelector))
+		return
+	}
+
+	diff := ComparePodSpecIgnoringDrift(&targetPod.Spec, &preflightPod.Spec, DefaultPreflightPodSpecDriftIgnore)
+	require.Empty(t, diff, "preflight pod %s/%s drifted from target %s/%s:\n%s",
+		preflightPod.Namespace, preflightPod.Name, targetPod.Namespace, targetPod.Name, diff)
+}
+
+func mustGetRunningOperandPod(ctx context.Context, t testing.TB, clientSet ClientSet, namespace, labelSelector string) *corev1.Pod {
```
ardaguclu:
```suggestion
func findRunningOperandPod(ctx context.Context, t testing.TB, clientSet ClientSet, namespace, labelSelector string) *corev1.Pod {
```

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+	preflightPod, err := clientSet.Kube.CoreV1().Pods(namespace).Get(ctx, preflight.PodName, metav1.GetOptions{})
+	require.NoError(t, err, "preflight pod %s/%s", namespace, preflight.PodName)
+
+	targetPod := mustGetRunningOperandPod(ctx, t, clientSet, namespace, labelSelector)
+	if targetPod == nil {
+		require.FailNow(t, fmt.Sprintf("no Running pod in %s matching %q", namespace, labelSelector))
+		return
+	}
+
+	diff := ComparePodSpecIgnoringDrift(&targetPod.Spec, &preflightPod.Spec, DefaultPreflightPodSpecDriftIgnore)
```
ardaguclu: Why do we export this function?. Can't we have it unexported?

test/library/encryption/preflight_drift.go
```diff
@@ -0,0 +1,128 @@
...
+	require.NotEmpty(t, namespace)
+	require.NotEmpty(t, labelSelector)
+
+	preflightPod, err := clientSet.Kube.CoreV1().Pods(namespace).Get(ctx, preflight.PodName, metav1.GetOptions{})
+	require.NoError(t, err, "preflight pod %s/%s", namespace, preflight.PodName)
+
+	targetPod := mustGetRunningOperandPod(ctx, t, clientSet, namespace, labelSelector)
+	if targetPod == nil {
+		require.FailNow(t, fmt.Sprintf("no Running pod in %s matching %q", namespace, labelSelector))
+		return
```
ardaguclu: dead code?

test/library/encryption/preflight_drift.go
tjungblu: does it matter which operand pod we pick?

test/library/encryption/preflight_deploy.go
tjungblu: sure.

test/library/encryption/preflight_drift.go
ardaguclu: yes, it doesn't matter

test/library/encryption/preflight_drift.go
tjungblu: it is, but the linter is unhappy because the nil deref on the targetPod. I guess we rather make this a require.NotNil....

test/library/encryption/preflight_drift.go
tjungblu: making it private

test/library/encryption/preflight_deploy.go
tjungblu: Done — \`AssertPreflightDeploy\` only checks deploy/conditions now. \`TestPreflightDeployAndPodMatchesOperand\` calls \`AssertNoPreflightConfigDrift\` afterward (and operators can call it the same way).

tjungblu: Renamed to \`AssertNoPreflightConfigDrift\`.

test/library/encryption/preflight_drift.go
tjungblu: Good call — switched to \`cmp.Diff\` + \`cmpopts.IgnoreFields\` / \`EquateEmpty\`.

test/library/encryption/preflight_drift.go
tjungblu: Label can match multiple replicas (e.g. 3 KAS static pods). We still pick any Running one — PodSpecs should match for the fields we compare. Renamed helper to \`findRunningOperandPod\`.

test/library/encryption/preflight_drift.go
tjungblu: Done — \`findRunningOperandPod\`.

test/library/encryption/preflight_drift.go
tjungblu: Kept the assert exported (now \`AssertNoPreflightConfigDrift\`) so the scenario / operators can call it outside \`AssertPreflightDeploy\`. The pod lookup helper is unexported.

test/library/encryption/preflight_drift.go
tjungblu: Removed — \`require.NotNil\` on the finder result instead of the unreachable \`return\` after \`FailNow\`.

test/library/encryption/preflight_deploy.go
tjungblu: on your initial topic, I didn't want to change the operator test logic (just a simple vendor bump), but I can add another assertion function if you fancy @ardaguclu

test/library/encryption/preflight_drift.go
tjungblu: renamed this "findAnyOperandPod" now

ardaguclu:
@tjungblu changes look good to me. I just have a question: Should we run this test after turning on the KMS and check the drift?, would it be better?

tjungblu:
@ardaguclu thanks!

Re running after KMS is on: I don't think we need that for this assert. `AssertNoPreflightConfigDrift` only compares PodSpec *shell* fields (HostNetwork, DNS, etc.) — containers/init/volumes (where the KMS plugin actually shows up) are on the ignore list by design. Enabling encryption first wouldn't change what we compare today, and would couple this check to a full migration.

Happy to add a follow-up later if we want a separate assert that runs post-migration (e.g. that the live operand actually carries a plugin sidecar), but I'd keep this one independent of mutating APIServer encryption.

tjungblu:
> Should we run this test after turning on the KMS and check the drift?, would it be better?

I think we should cover both cases, there's definitely value in checking preflight with an already existing sidecar. I'm not sure what the best way would be. We could do it as part of KMS-2-KMS migration (we will implicitly once we enable preflight controller anyway)?

Need to think about this for a while. Thanks for the quick review :)

PR #2387, opened 2026-07-29
OCPBUGS-100060: staticpod: add installer precondition hook

mkowalski:
## Summary

First of two PRs for [OCPBUGS-100060](https://issues.redhat.com/browse/OCPBUGS-100060): etcd quorum loss during upgrades when the etcd-operator's revision installer restarts an etcd member while MCO is simultaneously rebooting another master (2/3 members down, ~2min leaderless, cluster-wide API outage returning `429 storage is (re)initializing`).

- The installer controller creates installer pods unconditionally once a node has a pending target revision; the installer pod replaces the operand static-pod manifest, restarting the operand. The etcd-operator's `QuorumChecker` gates only **revision creation** (`WithRevisionControllerPrecondition`), so per-node installs of an existing revision proceed with no safety check. Evidence: in both incident runs the installer killed master-1's etcd 150–156s before master-0 finished its MCO reboot ([run 2075907197388197888](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-upgrade/2075907197388197888), [run 2077192709163978752](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-main-nightly-5.0-e2e-metal-ipi-ovn-upgrade-runc/2077192709163978752)).
- Adds `WithInstallerPrecondition(func(ctx, nodeName) (safe bool, reason string, err error))` to `InstallerController` and the static-pod controllers `Builder`. Consulted immediately before `ensureInstallerPod`; when unmet, emits `InstallerPreconditionNotMet` and requeues (15s) instead of restarting the operand. `nil` precondition preserves existing behavior for all other operators.
- Companion PR in cluster-etcd-operator wires this to a quorum/cordon safety check (`IsSafeToRestartMember`).

## Test plan

- [x] `gofmt`, `go vet`, `go build ./pkg/operator/staticpod/...`
- [x] `go test ./pkg/operator/staticpod/controller/installer/` — new `TestCreateInstallerPodPrecondition` (unmet delays pod + consults correct node; met allows; error fails sync); existing tests pass (`internal/atomicdir TestSwap` fails identically on pristine master in my environment — pre-existing, unrelated)

---
This PR was generated using AI. Please verify before acting on it.

## Summary by CodeRabbit

* **New Features**
  * Added an optional precondition check before creating installer pods.
  * Installer pod creation can now be delayed by a requested duration when conditions are not ready.
  * Added warning events and automatic retries using the specified delay.
  * A zero delay allows pod creation to proceed immediately.
  * Errors from the precondition check stop synchronization and prevent pod creation.
  * Added support for configuring the installer precondition through the static pod operator builder.

### DIFF
--- a/pkg/operator/staticpod/controller/installer/installer_controller.go
+++ b/pkg/operator/staticpod/controller/installer/installer_controller.go
@@ -106,6 +106,11 @@ type InstallerController struct {
 
 	installerPodMutationFns []InstallerPodMutationFunc
 
+	// installerPrecondition, when set, is consulted immediately before an installer pod is created for a node.
+	// When it returns false the installer controller emits an InstallerPreconditionNotMet event and requeues
+	// instead of creating the installer pod (which would restart the operand static pod on that node).
+	installerPrecondition InstallerPreconditionFunc
+
 	startupMonitorEnabled func() (bool, error)
 
 	factory          *factory.Factory
@@ -130,6 +135,24 @@ func (c *InstallerController) WithInstallerPodMutationFn(installerPodMutationFn 
 
 func (c *InstallerController) WithMinReadyDuration(minReadyDuration time.Duration) *InstallerController {
 	c.minReadyDuration = minReadyDuration
+	return c
+}
+
+// installerPreconditionRequeueDuration is how long the installer controller waits before rechecking
+// an unmet installer precondition.
+const installerPreconditionRequeueDuration = 15 * time.Second
+
+// InstallerPreconditionFunc returns true when it is safe to create an installer pod (which will
+// restart the operand static pod) on the given node.  When it returns false with a reason, the
+// installer controller requeues and retries later.  An error makes the sync fail.
+type InstallerPreconditionFunc func(ctx context.Context, nodeName string) (safe bool, reason string, err error)
+
+// WithInstallerPrecondition sets a precondition consulted immediately before creating an installer
+// pod for a node.  Operators can use it to delay operand restarts when doing so would be unsafe,
+// for example when restarting an etcd member while another control plane node is being rebooted
+// would lose quorum.
+func (c *InstallerController) WithInstallerPrecondition(precondition InstallerPreconditionFunc) *InstallerController {
+	c.installerPrecondition = precondition
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


pkg/operator/staticpod/controller/installer/installer_controller.go
```diff
@@ -133,6 +138,24 @@ func (c *InstallerController) WithMinReadyDuration(minReadyDuration time.Duratio
...
 }
 
+// installerPreconditionRequeueDuration is how long the installer controller waits before rechecking
+// an unmet installer precondition.
+const installerPreconditionRequeueDuration = 15 * time.Second
+
+// InstallerPreconditionFunc returns true when it is safe to create an installer pod (which will
+// restart the operand static pod) on the given node.  When it returns false with a reason, the
+// installer controller requeues and retries later.  An error makes the sync fail.
+type InstallerPreconditionFunc func(ctx context.Context, nodeName string) (safe bool, reason string, err error)
```
tchap: I think that it would be more flexible to return the backoff duration as a return value, so we could turn `safe` into a `time.Duration` and wait when that duration is `> 0`. We can then remove `installerPreconditionRequeueDuration`, which is not configurable.

tchap (commented):
This improves the situation when MCO is already rebooting a node. In that case we can postpone. But in the other direction, MCO can still start rebooting a node when we are installing a new revision. So this is certainly not bullet-proof, but makes the situation much better, because installation does not take that long while the MCO reboot window can be pretty large.

I guess to really solve this, we would need to use a lease or something.

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -2769,3 +2770,152 @@ func TestWaitToObserveWrites(t *testing.T) {
...
+		if preconditionEvent.Type != corev1.EventTypeWarning {
+			t.Fatalf("expected warning event, got type %q", preconditionEvent.Type)
+		}
+	})
+
+	t.Run("zero delay allows installer pod", func(t *testing.T) {
+		c, eventRecorder, getPod := newController(func(ctx context.Context, nodeName string) (time.Duration, string, error) {
+			return 0, "", nil
+		})
+		for i := 0; i < 3; i++ {
```
tchap, line 2894: 1. If you set `TargetRevision: 1` on the node status, you can just call `Sync` once IMO.
2. In case you need a loop, use `for range 3 {}`

The same goes for the next test.

pkg/operator/staticpod/controller/installer/installer_controller_test.go
```diff
@@ -2769,3 +2770,152 @@ func TestWaitToObserveWrites(t *testing.T) {
...
+		checkedNode := ""
+		c, eventRecorder, getPod := newController(func(ctx context.Context, nodeName string) (time.Duration, string, error) {
+			checkedNode = nodeName
+			return requestedDelay, "another master is rebooting", nil
+		})
+		queue := &recordingRateLimitingQueue{RateLimitingInterface: workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())}
+		defer queue.ShutDown()
+		syncCtx := recordingSyncContext{recorder: eventRecorder, queue: queue, queueKey: "test-key"}
+		for i := 0; i < 3 && len(queue.addAfterCalls) == 0; i++ {
+			if err := c.Sync(context.TODO(), syncCtx); err != nil {
```
tchap, line 2859: I would honestly just call `manageInstallationPods` in these tests. That would allow us to mock much less while still testing everything. You could just check the return value matches the requested delay, for example.

PR #2388, opened 2026-07-29
NO-JIRA: Add documentation about we don't add informers for openshift-config namespace

ardaguclu:
This PR adds documentation about why we decided to not add informers for openshift-config namespace.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that Secrets and ConfigMaps in the configuration namespace are intentionally not continuously watched, and that updates are picked up during the next periodic resync cycle.
* **Bug Fixes**
  * No functional behavior changes were introduced in this update.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -116,12 +116,19 @@ func NewKeyController(
 	return factory.New().
 		WithSync(c.sync).
 		WithControllerInstanceName(c.controllerInstanceName).
+		// Keep the resync interval short: openshift-config secrets/configmaps are not
+		// watched directly, so credential and CA bundle rotations are only detected on
+		// resync. Increasing this significantly delays in-place config propagation.
 		ResyncEvery(time.Minute).
 		WithInformers(
 			apiServerInformer.Informer(),
 			operatorClient.Informer(),
 			kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
-			// TODO: add informers for openshift-config namespace to watch referenced Secrets and ConfigMaps for KMS plugin data changes
+			// openshift-config secrets/configmaps are not watched directly. While we could
+			// build a mechanism to watch only the referenced resources, creating and
+			// maintaining it is not free, and watching all resources in the namespace
+			// would create excessive API server load. Instead, changes are picked up
+			// on the next periodic resync.
 			deployer,
 		).ToController(
 		c.controllerInstanceName,


bertinatto (commented):
/lgtm

adding a hold to give the others a chance to take a look:

/hold

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -121,7 +121,8 @@ func NewKeyController(
 			apiServerInformer.Informer(),
 			operatorClient.Informer(),
 			kubeInformersForNamespaces.InformersFor("openshift-config-managed").Core().V1().Secrets().Informer(),
-			// TODO: add informers for openshift-config namespace to watch referenced Secrets and ConfigMaps for KMS plugin data changes
+			kubeInformersForNamespaces.InformersFor("openshift-config").Core().V1().Secrets().Informer(),
```
p0lyn0mial: We need to evaluate whether it makes sense to filter events from these resources so that we only process the ones we care about. 

The encryption controllers use live clients, so on every reconciliation they hammer the server. Note that this is not how the majority of controllers work. 

I think we have at least two cases: 
1. encryption is enabled and uses 'static' encryption. In this case, we could ignore all events from these two informers. 
2. encryption is enabled and uses 'kms' mode. In this case, we could react only to updates for the resources used by the configured kms.

ardaguclu: These are good points. We need to discuss.

tjungblu: openshift-config is a pretty calm namespace. Don't ever do this with node informers, they update ping every 1s on the keep alive 🤣

ardaguclu: After @p0lyn0mial's comment, I think we should never watch this namespace. After just 1 minute, key controller will reconcile anyway. What do we gain if we watch it?

tjungblu: initial cache sync gating: when you start the sync loop you're guaranteed that the watch cache is populated with _something_. Otherwise you may run into a race condition (eg on startup) where you may expect a secret that previously was there, but now isn't anymore. Sometimes this causes funny issues like a revision rollout. 

For example, that's actually an issue I've fixed yesterday:
https://github.com/openshift/cluster-etcd-operator/pull/1659

ardaguclu: But we always use live client

ardaguclu:
/retitle NO-JIRA: Add documentation about we don't add informers for openshift-config namespace

PR #2389, opened 2026-07-29
kms: don't resync preflight controller when apiserver changes

bertinatto:
When the user changes `apiserver/cluster`, the preflight controller currently resyncs. However, the controller will only be able to run the checks if the `ObservedConfigHash` field was set accordingly by the key controller. In that case, it's better to just rely operator client for that status update instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved KMS preflight synchronization behavior by preventing unnecessary syncs from API server events.
  * KMS preflight checks continue to update based on operator status changes and periodic refreshes.


p0lyn0mial:
thanks

/lgtm
/retest

PR #2390, opened 2026-07-30
NO-JIRA: Standartize and document the returned latest key id from needsNewKey func

ardaguclu:
When new key is not needed, returned latest Key ID is not used. However, with in-place field updates, this latest Key ID will be used to update the kms provider config, unless it is 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the most recent encryption key ID when rotation isn’t required (including non-migrated and KMS “no time-based rotation” scenarios), improving correctness for consumers that read the current key.
* **Documentation**
  * Clarified the meaning of `latestKeyID`, `reason`, `needed`, and `err` to make key-rotation outcomes and returned values easier to interpret.


ardaguclu:
/close

PR #2391, opened 2026-07-30
NO-JIRA: Move inProgress check lower levels to allow in-place updates

ardaguclu:
This PR moves the location of convergence check to allow in-place field updates will be effective, even if there is new rollout in progress. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption key management while an encryption update is in progress.
  * The system now completes key requirement checks before deferring further processing, allowing configuration issues to be detected earlier.
  * New encryption keys are still not created until the active update has finished.


ardaguclu:
/close

PR #2393, opened 2026-07-30
NO-JIRA: Runs state controller even if there is a transition

ardaguclu:
Key controller may propagate data, even if there isn't any convergence. So from that point, state controller should run even if there is an active transition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption configuration reconciliation by removing unnecessary delays during transitions.
  * Encryption configuration secrets are now generated and applied directly when reconciliation occurs.
  * Updated event reporting to reflect changes to the applied encryption configuration.


ardaguclu:
/close

PR #2395, opened 2026-08-03
NO-JIRA: Update in-place field update scenario to align other scenarios

ardaguclu:
In-place field update scenario diverged from other scenarios. We need to align it with the others. This PR achieves this.

## Summary by CodeRabbit

* **Tests**
  * Expanded encryption validation coverage for in-place resource updates.
  * Added checks confirming resources remain encrypted after provider configuration changes.
  * Improved verification that updates do not trigger unnecessary encryption key creation.
  * Added reusable validation for confirming resource-specific in-place update behavior.

### DIFF
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -337,62 +337,70 @@ func ApplyEncryption(ctx context.Context, t testing.TB, encryption configv1.APIS
 	t.Logf("Applied encryption config (type=%s)", encryption.Type)
 }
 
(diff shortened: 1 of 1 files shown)


gangwgr:
@ardaguclu can we have a test run?

ardaguclu:
> @ardaguclu can we have a test run?

I have to run this with https://github.com/openshift/library-go/pull/2241. Would it make sense we merge this PR and after that I'll run the test with https://github.com/openshift/library-go/pull/2241. If we find any bug, I can fix it.

PR #2399, opened 2026-08-03
WIP: compute encryption-config secret (isolated namespace approach)

bertinatto:
## Summary by CodeRabbit

* **New Features**
  * KMS preflight checks now generate and validate encryption configuration in an isolated temporary environment.
  * Preflight processing supports existing encryption keys, provider changes, and deployments that are not yet fully converged.
  * Encryption keys and configuration can be read from configurable namespaces.
  * Temporary environments are automatically cleaned up, with cleanup issues reported as warnings.
* **Bug Fixes**
  * KMS endpoints are correctly rewritten for preflight validation.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_sandbox.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_sandbox.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	preflightKMSSocketEndpoint = "unix:///var/run/kmsplugin/kms.sock"
+	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms.sock"
 	preflightTempNamespaceLabel = "encryption.apiserver.operator.openshift.io/kms-preflight"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,321 @@
...
+	if err := ensurePreflightTempNamespace(ctx, coreClient, tempNamespace); err != nil {
+		return nil, err
+	}
+	defer func() {
+		if cleanupErr := deletePreflightTempNamespace(ctx, coreClient, tempNamespace); cleanupErr != nil {
+			klog.Warningf("failed to delete preflight temp namespace %q: %v", tempNamespace, cleanupErr)
+		}
+	}()
+
+	if err := seedEncryptionKeySecrets(ctx, coreClient, secrets.EncryptionKeysNamespace, tempNamespace, encryptionSecretSelector); err != nil {
```
bertinatto, line 68: this is necessary to copy *existing* key secrets. We need to copy those for cases when we need an in-place upgrade for instance

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,321 @@
...
+	effectiveDeployer := forceConvergedDeployer{delegate: encryptionDeployer}
+
+	tempNamespace := preflightTempNamespaceName(configHash)
+	if err := ensurePreflightTempNamespace(ctx, coreClient, tempNamespace); err != nil {
+		return nil, err
+	}
+	defer func() {
+		if cleanupErr := deletePreflightTempNamespace(ctx, coreClient, tempNamespace); cleanupErr != nil {
+			klog.Warningf("failed to delete preflight temp namespace %q: %v", tempNamespace, cleanupErr)
+		}
```
bertinatto, line 65: should we worry about namespace churn (too many namespaces being created/deleted)?

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,321 @@
...
+		if !foundKey || id > latestKeyID {
+			latestKeyID = id
+			foundKey = true
+		}
+	}
+	if !foundKey {
+		return nil, fmt.Errorf("no encryption key secrets found after key controller run in temp namespace %q", tempNamespace)
+	}
+
+	rewritten, err := rewritePreflightWriteKeyEndpoint(encryptionSecret, latestKeyID)
```
bertinatto, line 101: need to confirm with @p0lyn0mial, but IIUC we'll be able to remove this workaround

pkg/operator/encryption/controllers/kms_preflight_sandbox.go
```diff
@@ -0,0 +1,318 @@
+package controllers
```
bertinatto, line 1: this isn't really a "sandbox" because it's not possible to control where the key/state controllers will store data (i.e., an change in one of those controllers might introduce a resource somewhere outside the designated namespace and the preflight controller won't know it)

ardaguclu:
https://github.com/openshift/library-go/pull/2410 supersedes this PR
/close

PR #2410, opened 2026-08-06
CNTRLPLANE-3237: kms: add EncryptionPlanner and migrate key, state, and preflight controllers

bertinatto:
## Summary by CodeRabbit

- **New Features**
  - Improved encryption key planning for creation, rotation, migration, and configuration changes.
  - Added candidate encryption configuration generation and validation during KMS preflight checks.
  - Added support for KMS endpoint overrides during preflight validation.

- **Bug Fixes**
  - Prevented key creation when KMS preflight validation fails.
  - Improved handling of missing, invalid, or incomplete KMS credentials.
  - Ensured encryption state and candidate configurations remain consistent.
  - Automatically cleaned up successful KMS preflight workloads.
  - Improved progress reporting while encryption changes are being applied.

### DIFF
--- a/pkg/operator/encryption/controllers/encryption_planner.go
+++ b/pkg/operator/encryption/controllers/encryption_planner.go
@@ -1,0 +1,268 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/encryptiondata"
+	"github.com/openshift/library-go/pkg/operator/encryption/secrets"
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+	"github.com/openshift/library-go/pkg/operator/encryption/statemachine"
+	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+)
+
+const (
+	encryptionConfigManagedNS = "openshift-config-managed"
+)
+
+// EncryptionPlanner provides shared key planning and encryption-config computation for the key, state, and KMS preflight controllers without drift.
+type EncryptionPlanner struct {
+	instanceName             string
+	unsupportedConfigPrefix  []string
+	deployer                 statemachine.Deployer
+	secretClient             corev1client.SecretsGetter
+	configMapClient          corev1client.ConfigMapsGetter
+	apiServerClient          configv1client.APIServerInterface
+	operatorClient           operatorv1helpers.OperatorClient
+	encryptionSecretSelector metav1.ListOptions
+}
+
+func NewEncryptionPlanner(instanceName string, unsupportedConfigPrefix []string, deployer statemachine.Deployer, secretClient corev1client.SecretsGetter, configMapClient corev1client.ConfigMapsGetter, apiServerClient configv1client.APIServerInterface, operatorClient operatorv1helpers.OperatorClient, encryptionSecretSelector metav1.ListOptions) *EncryptionPlanner {
+	return &EncryptionPlanner{
+		instanceName:             instanceName,
+		unsupportedConfigPrefix:  unsupportedConfigPrefix,
+		deployer:                 deployer,
+		secretClient:             secretClient,
+		configMapClient:          configMapClient,
+		apiServerClient:          apiServerClient,
+		operatorClient:           operatorClient,
+		encryptionSecretSelector: encryptionSecretSelector,
+	}
+}
+
+// EncryptionStateSnapshot is deployer convergence and persisted key secrets.
+// Produced by LoadState. Input to ComputeConfig.
+type EncryptionStateSnapshot struct {
+	ProgressingReason string
+	CurrentConfig     *encryptiondata.Config
+	DesiredBeforePlan map[schema.GroupResource]state.GroupResourceState
+	KeySecrets        []*corev1.Secret
+	EncryptedGRs      []schema.GroupResource
+}
+
+// KeyPlanningSnapshot adds resolved encryption mode for key planning.
+// Produced by Load. Required by PlanNextKey and MaterializeKey.
+type KeyPlanningSnapshot struct {
+	State              EncryptionStateSnapshot
+	CurrentMode        state.Mode
+	ExternalReason     string
+	APIEncryption      configv1.APIServerEncryption
+	desiredProviderCfg kmsProviderConfig
+}
+
+// KeyPlan is the deterministic result of deciding whether a new key is needed.
+type KeyPlan struct {
+	Needed         bool
+	KeyID          uint64
+	Reasons        []string
+	InternalReason string
+}
+
+// PlannedEncryptionKey is a materialized in-memory key secret ready to persist or include in a candidate config.
+type PlannedEncryptionKey struct {
+	Secret  *corev1.Secret
+	KeyID   uint64
+	Reasons []string
+}
+
+// LoadOptions controls how Load reads key-planning state.
+type LoadOptions struct {
+	// ListKeysWhileProgressing lists persisted key secrets even when the
+	// deployer has not converged. Default false so the key controller does
+	// not List during API-server rollouts. Preflight sets this so PlanNextKey
+	// can run. ProgressingReason is still populated in either case.
+	ListKeysWhileProgressing bool
+}
+
+// EncryptionPlanResult is returned by ComputeConfig.
+type EncryptionPlanResult struct {
+	ProgressingReason string
+	PlannedKey        *PlannedEncryptionKey
+	CurrentConfig     *encryptiondata.Config
+	DesiredState      map[schema.GroupResource]state.GroupResourceState
+	EncryptionConfig  *encryptiondata.Config
+	EncryptionSecret  *corev1.Secret
+	KeySecrets        []*corev1.Secret
+}
+
+// plannedKeyBuildError marks failures while materializing an in-memory planned key (missing referenced Secrets/ConfigMaps, etc.). The key controller wraps these as "failed to create key" for historical degraded messages.
+type plannedKeyBuildError struct {
+	err error
+}
+
+func (e plannedKeyBuildError) Error() string { return e.err.Error() }
+func (e plannedKeyBuildError) Unwrap() error { return e.err }
+
+// LoadState reads deployer convergence and key secrets.
+func (p *EncryptionPlanner) LoadState(ctx context.Context, encryptedGRs []schema.GroupResource) (*EncryptionStateSnapshot, error) {
+	currentConfig, desiredBeforePlan, keySecrets, progressingReason, err := statemachine.GetEncryptionConfigAndState(ctx, p.deployer, p.secretClient, p.encryptionSecretSelector, encryptedGRs)
+	if err != nil {
+		return nil, err
+	}
+	return &EncryptionStateSnapshot{
+		ProgressingReason: progressingReason,
+		CurrentConfig:     currentConfig,
+		DesiredBeforePlan: desiredBeforePlan,
+		KeySecrets:        keySecrets,
+		EncryptedGRs:      encryptedGRs,
+	}, nil
+}
+
+// Load resolves encryption mode and reads deployer/key state for key planning.
+// The returned snapshot is non-nil when err is nil.
+func (p *EncryptionPlanner) Load(ctx context.Context, encryptedGRs []schema.GroupResource, opts LoadOptions) (*KeyPlanningSnapshot, error) {
+	if p.apiServerClient == nil || p.operatorClient == nil {
+		return nil, fmt.Errorf("Load requires apiServerClient and operatorClient")
+	}
+
+	// Resolve mode before reading deployer/key state so callers fail fast on APIServer/operator errors.
+	currentMode, externalReason, apiEncryption, err := getCurrentModeReasonAndEncryptionConfig(ctx, p.apiServerClient, p.operatorClient, p.unsupportedConfigPrefix)
+	if err != nil {
+		return nil, err
+	}
+
+	stateSnap, err := p.LoadState(ctx, encryptedGRs)
+	if err != nil {
+		return nil, err
+	}
+
+	// LoadState calls GetEncryptionConfigAndState, which returns before listing keys when the deployer has not converged.
+	// Preflight loads keys and desired state here via ListKeysWhileProgressing; the key controller skips this to avoid Lists during rollout.
+	if len(stateSnap.ProgressingReason) > 0 && opts.ListKeysWhileProgressing {
+		keySecrets, err := secrets.ListKeySecrets(ctx, p.secretClient, p.encryptionSecretSelector)
+		if err != nil {
+			return nil, err
+		}
+		stateSnap.KeySecrets = keySecrets
+		stateSnap.DesiredBeforePlan = statemachine.GetDesiredEncryptionState(stateSnap.CurrentConfig, keySecrets, encryptedGRs)
+	}
+
+	snap := &KeyPlanningSnapshot{
+		State:              *stateSnap,
+		CurrentMode:        currentMode,
+		ExternalReason:     externalReason,
+		APIEncryption:      apiEncryption,
+		desiredProviderCfg: noopKMSProviderConfig{},
+	}
+
+	if currentMode == state.KMS {
+		desiredProviderCfg, err := newKMSProviderConfig(apiEncryption.KMS)
+		if err != nil {
+			return nil, err
+		}
+		snap.desiredProviderCfg = desiredProviderCfg
+	}
+
+	return snap, nil
+}
+
+// PlanNextKey deterministically decides whether a new key is needed. It does not generate key material.
+func (p *EncryptionPlanner) PlanNextKey(snap *KeyPlanningSnapshot) (*KeyPlan, error) {
+	if snap == nil {
+		return nil, fmt.Errorf("snapshot is required")
+	}
+
+	hasBeenOnBefore := snap.State.CurrentConfig != nil || len(snap.State.KeySecrets) > 0
+	if snap.CurrentMode == state.Identity && !hasBeenOnBefore {
+		return &KeyPlan{}, nil
+	}
+
+	keyPlan, err := planNextEncryptionKey(snap.State.DesiredBeforePlan, snap.CurrentMode, snap.ExternalReason, snap.State.EncryptedGRs, snap.desiredProviderCfg)
+	if err != nil {
+		return nil, err
+	}
+	return &KeyPlan{
+		Needed:         keyPlan.needed,
+		KeyID:          keyPlan.keyID,
+		Reasons:        keyPlan.reasons,
+		InternalReason: keyPlan.internalReason,
+	}, nil
+}
+
+// MaterializeKey builds an in-memory key secret for plan. Non-deterministic (fresh key bytes). Requires plan.Needed == true.
+func (p *EncryptionPlanner) MaterializeKey(ctx context.Context, snap *KeyPlanningSnapshot, plan *KeyPlan) (*PlannedEncryptionKey, error) {
+	if snap == nil {
+		return nil, fmt.Errorf("snapshot is required")
+	}
+	if plan == nil || !plan.Needed {
+		return nil, fmt.Errorf("MaterializeKey requires a needed key plan")
+	}
+	if p.configMapClient == nil {
+		return nil, fmt.Errorf("configMapClient is required for MaterializeKey")
+	}
+
+	ks, _, _, err := buildEncryptionKeyState(ctx, plan.KeyID, snap.CurrentMode, snap.APIEncryption, snap.desiredProviderCfg, p.secretClient, p.configMapClient, plan.InternalReason, snap.ExternalReason)
+	if err != nil {
+		return nil, plannedKeyBuildError{err: err}
+	}
+	keySecret, err := secrets.FromKeyState(p.instanceName, ks)
+	if err != nil {
+		return nil, plannedKeyBuildError{err: err}
+	}
+	return &PlannedEncryptionKey{
+		Secret:  keySecret,
+		KeyID:   plan.KeyID,
+		Reasons: plan.Reasons,
+	}, nil
+}
+
+// ComputeConfig builds the desired encryption-config secret from persisted keys, optionally including plannedKey.
+// When plannedKey is nil it reuses DesiredBeforePlan instead of recomputing desired state.
+func (p *EncryptionPlanner) ComputeConfig(state *EncryptionStateSnapshot, plannedKey *PlannedEncryptionKey) (*EncryptionPlanResult, error) {
+	if state == nil {
+		return nil, fmt.Errorf("snapshot is required")
+	}
+
+	keySecrets := state.KeySecrets
+	if plannedKey != nil && plannedKey.Secret != nil {
+		keySecrets = append(append([]*corev1.Secret{}, state.KeySecrets...), plannedKey.Secret)
+	}
+
+	out := &EncryptionPlanResult{
+		CurrentConfig: state.CurrentConfig,
+		KeySecrets:    keySecrets,
+		PlannedKey:    plannedKey,
+	}
+
+	if state.CurrentConfig == nil && len(keySecrets) == 0 {
+		return out, nil
+	}
+
+	// DesiredBeforePlan already reflects persisted keys. Recompute only when a planned key was appended
+	out.DesiredState = state.DesiredBeforePlan
+	if plannedKey != nil && plannedKey.Secret != nil {
+		out.DesiredState = statemachine.GetDesiredEncryptionState(state.CurrentConfig, keySecrets, state.EncryptedGRs)
+	}
+
+	cfg, err := encryptiondata.FromEncryptionState(out.DesiredState)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build encryption config: %w", err)
+	}
+	out.EncryptionConfig = cfg
+
+	secretName := fmt.Sprintf("%s-%s", encryptiondata.EncryptionConfSecretName, p.instanceName)
+	secret, err := encryptiondata.ToSecret(encryptionConfigManagedNS, secretName, cfg)
+	if err != nil {
+		return nil, err
+	}
+	out.EncryptionSecret = secret
+
+	return out, nil
+}
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -183,47 +183,25 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
 }
 
 func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext factory.SyncContext, encryptedGRs []schema.GroupResource) error {
-	currentMode, externalReason, apiEncryptionConfiguration, err := getCurrentModeReasonAndEncryptionConfig(ctx, c.apiServerClient, c.operatorClient, c.unsupportedConfigPrefix)
+	planner := NewEncryptionPlanner(c.instanceName, c.unsupportedConfigPrefix, c.deployer, c.secretClient, c.configMapClient, c.apiServerClient, c.operatorClient, c.encryptionSecretSelector)
+	snap, err := planner.Load(ctx, encryptedGRs, LoadOptions{})
 	if err != nil {
 		return err
 	}
-
-	currentConfig, desiredEncryptionState, secrets, isProgressingReason, err := statemachine.GetEncryptionConfigAndState(ctx, c.deployer, c.secretClient, c.encryptionSecretSelector, encryptedGRs)
-	if err != nil {
-		return err
-	}
-	if len(isProgressingReason) > 0 {
+	if len(snap.State.ProgressingReason) > 0 {
 		syncContext.Queue().AddAfter(syncContext.QueueKey(), 2*time.Minute)
 		return nil
 	}
 
-	// avoid intended start of encryption
-	hasBeenOnBefore := currentConfig != nil || len(secrets) > 0
-	if currentMode == state.Identity && !hasBeenOnBefore {
-		return nil
-	}
-
-	// note here that desiredEncryptionState is never empty because getDesiredEncryptionState
-	// fills up the state with all resources and set identity write key if write key secrets
-	// are missing.
-
-	var desiredProviderCfg kmsProviderConfig = noopKMSProviderConfig{}
-	if currentMode == state.KMS {
-		var err error
-		desiredProviderCfg, err = newKMSProviderConfig(apiEncryptionConfiguration.KMS)
-		if err != nil {
-			return err
-		}
-	}
-
-	keyPlan, err := planNextEncryptionKey(desiredEncryptionState, currentMode, externalReason, encryptedGRs, desiredProviderCfg)
+	plan, err := planner.PlanNextKey(snap)
 	if err != nil {
 		return err
 	}
-	if !keyPlan.needed {
+	if !plan.Needed {
 		return nil
 	}
-	keySecret, preconditionMet, err := c.generateKeySecret(ctx, keyPlan.keyID, currentMode, apiEncryptionConfiguration, desiredProviderCfg, keyPlan.internalReason, externalReason)
+
+	keySecret, preconditionMet, err := c.generateKeySecret(ctx, plan.KeyID, snap.CurrentMode, snap.APIEncryption, snap.desiredProviderCfg, plan.InternalReason, snap.ExternalReason)
 	if err != nil {
 		return fmt.Errorf("failed to create key: %v", err)
 	}
@@ -231,16 +209,17 @@ func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext fact
 		syncContext.Queue().AddAfter(syncContext.QueueKey(), 30*time.Second)
 		return nil
 	}
+
 	_, createErr := c.secretClient.Secrets("openshift-config-managed").Create(ctx, keySecret, metav1.CreateOptions{})
 	if errors.IsAlreadyExists(createErr) {
-		return c.validateExistingSecret(ctx, keySecret, keyPlan.keyID)
+		return c.validateExistingSecret(ctx, keySecret, plan.KeyID)
 	}
 	if createErr != nil {
 		syncContext.Recorder().Warningf("EncryptionKeyCreateFailed", "Secret %q failed to create: %v", keySecret.Name, createErr)
 		return createErr
 	}
 
-	syncContext.Recorder().Eventf("EncryptionKeyCreated", "Secret %q successfully created: %q", keySecret.Name, keyPlan.reasons)
+	syncContext.Recorder().Eventf("EncryptionKeyCreated", "Secret %q successfully created: %q", keySecret.Name, plan.Reasons)
 
 	return nil
 }
@@ -263,6 +242,7 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 
 	return nil // we made this key earlier
 }
+
 // generateKeySecret builds the key secret for the given key ID and mode.
 //
 // For KMS mode it also computes the config hash (from the referenced Secret and
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -18,7 +18,6 @@ import (
 	apiserverv1 "k8s.io/apiserver/pkg/apis/apiserver/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/v2"
-	"k8s.io/utils/ptr"
 
 	configv1 "github.com/openshift/api/config/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
@@ -204,12 +203,6 @@ func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext fact
 		return nil
 	}
 
-	var (
-		newKeyRequired bool
-		newKeyID       uint64
-		reasons        []string
-	)
-
 	// note here that desiredEncryptionState is never empty because getDesiredEncryptionState
 	// fills up the state with all resources and set identity write key if write key secrets
 	// are missing.
@@ -223,40 +216,14 @@ func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext fact
 		}
 	}
 
-	var commonReason *string
-	for gr, grKeys := range desiredEncryptionState {
-		latestKeyID, internalReason, needed, err := needsNewKey(grKeys, currentMode, externalReason, encryptedGRs, desiredProviderCfg)
-		if err != nil {
-			return err
-		}
-		if !needed {
-			continue
-		}
-
-		if commonReason == nil {
-			commonReason = &internalReason
-		} else if *commonReason != internalReason {
-			commonReason = ptr.To("") // this means we have no common reason
-		}
-
-		newKeyRequired = true
-		nextKeyID := latestKeyID + 1
-		if newKeyID < nextKeyID {
-			newKeyID = nextKeyID
-		}
-		reasons = append(reasons, fmt.Sprintf("%s-%s", gr.Resource, internalReason))
+	keyPlan, err := planNextEncryptionKey(desiredEncryptionState, currentMode, externalReason, encryptedGRs, desiredProviderCfg)
+	if err != nil {
+		return err
 	}
-	if !newKeyRequired {
+	if !keyPlan.needed {
 		return nil
 	}
-
-	if commonReason != nil && len(*commonReason) > 0 && len(reasons) > 1 {
-		reasons = []string{*commonReason} // don't repeat reasons
-	}
-
-	sort.Sort(sort.StringSlice(reasons))
-	internalReason := strings.Join(reasons, ", ")
-	keySecret, preconditionMet, err := c.generateKeySecret(ctx, newKeyID, currentMode, apiEncryptionConfiguration, desiredProviderCfg, internalReason, externalReason)
+	keySecret, preconditionMet, err := c.generateKeySecret(ctx, keyPlan.keyID, currentMode, apiEncryptionConfiguration, desiredProviderCfg, keyPlan.internalReason, externalReason)
 	if err != nil {
 		return fmt.Errorf("failed to create key: %v", err)
 	}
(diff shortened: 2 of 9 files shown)


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -299,14 +266,73 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
...
+	}
+	if currentMode != state.KMS {
+		return secret, true, nil
+	}
+
+	ks, err := secrets.ToKeyState(secret)
+	if err != nil {
+		return nil, false, err
+	}
+	refSecret, refCM, err := referencedResourcesFromKeyState(ks, desiredProviderCfg)
```
ardaguclu: This could be a separate refactoring that we can merge it now.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -317,95 +343,158 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
...
+	}
 
-		resources := &prefetchedKMSConfigHasherResourceProvider{secret: refSecret, configMap: refCM}
-		hasher, err := newKMSConfigHasher(desiredProviderCfg, resources, openshiftConfigNS)
-		if err != nil {
-			return nil, false, fmt.Errorf("failed to create KMS config hasher: %w", err)
+	return ks, nil
+}
+
+func buildEncryptionKeySecret(
```
ardaguclu: Do we really need this function?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -317,95 +343,158 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	return secrets.FromKeyState(instanceName, ks)
+}
+
+// referencedResourcesFromKeyState rebuilds the Secret/ConfigMap objects the hasher
+// expects from credentials already embedded in the planned key state.
+func referencedResourcesFromKeyState(ks state.KeyState, desiredProviderCfg kmsProviderConfig) (*corev1.Secret, *corev1.ConfigMap, error) {
```
ardaguclu: I think this function is better move independent from the refactoring (like we did exporting GetDesiredState.)

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -317,95 +343,158 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
...
-	return secret, true, nil
+	return refSecret, refCM, nil
 }
 
 func (c *keyController) getCurrentModeReasonAndEncryptionConfig(ctx context.Context) (state.Mode, string, configv1.APIServerEncryption, error) {
-	apiServer, err := c.apiServerClient.Get(ctx, "cluster", metav1.GetOptions{})
+	return resolveEncryptionModeAndConfig(ctx, c.apiServerClient, c.operatorClient, c.unsupportedConfigPrefix)
+}
+
+func resolveEncryptionModeAndConfig(
```
ardaguclu: Do we really need this function?. I think we can directly make `getCurrentModeReasonAndEncryptionConfig` standalone in a separate PR (and we can merge it now)

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -264,61 +262,49 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
...
-		c.configMapClient,
-		internalReason,
-		externalReason,
-		"",
-	)
+// Return values mirror the former generateKeySecret precondition:
+//   - (true, nil)  — proceed with key creation
+//   - (false, nil) — preflight in progress; caller should back off
+//   - (false, err) — preflight failed or transient error
+func (c *keyController) ensureKMSPreflightBeforeKeyCreate(ctx context.Context, currentMode state.Mode, plannedKeySecret *corev1.Secret) (bool, error) {
```
ardaguclu: This is likely coming from another PR whose already has merged.

pkg/operator/encryption/controllers/state_controller.go
```diff
@@ -163,13 +177,8 @@ func (c *stateController) generateAndApplyCurrentEncryptionConfigSecret(ctx cont
...
 }
 
-func (c *stateController) applyEncryptionConfigSecret(ctx context.Context, secretData *encryptiondata.Config, recorder events.Recorder) (bool, error) {
-	s, err := encryptiondata.ToSecret("openshift-config-managed", fmt.Sprintf("%s-%s", encryptiondata.EncryptionConfSecretName, c.instanceName), secretData)
-	if err != nil {
-		return false, err
-	}
-
-	_, changed, applyErr := resourceapply.ApplySecret(ctx, c.secretClient, recorder, s)
+func (c *stateController) applyEncryptionConfigSecret(ctx context.Context, secret *corev1.Secret, recorder events.Recorder) (bool, error) {
```
ardaguclu: nit: Do we have to have this function anymore?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -763,3 +863,63 @@ func (c *kmsPreflightController) preflightRequired(ctx context.Context) (string,
 	}
 	return requiredHash, nil, nil
 }
+
+// Temporary: fixed preflight KMS socket. Safe to delete with latestKeyIDFromSecrets,
+// rewriteWriteKeyKMSEndpoint, and the PlannedKey==nil rewrite in computeEncryptionConfigSecret.
+const preflightKMSSocketEndpoint = "unix:///var/run/kmsplugin/kms.sock"
+
+func latestKeyIDFromSecrets(keySecrets []*corev1.Secret) (uint64, error) {
```
ardaguclu: Why do we have this function?. Doesn't plannedKey give this us for free?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -317,95 +329,154 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
...
+		Encryption: &apiserverv1.KMSConfiguration{
+			APIVersion: "v2",
+			Name:       fmt.Sprintf("%d", keyID),
+			Endpoint:   endpoint,
+			Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
+		},
+		Plugin: apiServerEncryption.KMS,
+	}
+
+	if secretName, expectedKeys, err := desiredProviderCfg.referencedSecretName(); err != nil {
```
ardaguclu: Can't we use `referencedResourcesFromKeyState` automatically instead of duplicating the code?

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,289 @@
...
+			return nil, err
+		}
+		snap.CurrentMode = currentMode
+		snap.ExternalReason = externalReason
+		snap.APIEncryption = apiEncryption
+		snap.modeResolved = true
+		snap.desiredProviderCfg = noopKMSProviderConfig{}
+
+		if currentMode == state.KMS {
+			desiredProviderCfg, err := newKMSProviderConfig(apiEncryption.KMS)
```
ardaguclu: Previously `newKMSProviderConfig` is called after the `progressingReason` check. I think, if `progressingReason` is not empty, we need to short cut first to not have any behavioral changes.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: I meant we can have a generic function that fetches the configmap and secrets from the given provider config. This function can be used by everywhere automatically.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -599,6 +632,73 @@ func (c *kmsPreflightController) cleanupDeployer(ctx context.Context) error {
...
+	}
+	if result.EncryptionSecret == nil {
+		return false, nil, fmt.Errorf("no encryption key secrets available to compute preflight encryption config")
+	}
+
+	secret = result.EncryptionSecret
+	// When reusing an existing write key, rewrite its endpoint so the preflight
+	// checker dials the fixed socket. New planned keys already use the override.
+	if plannedKey == nil {
+		writeKeyID, err := latestKeyIDFromSecrets(result.KeySecrets)
```
ardaguclu:
```suggestion
 for _, grState := range snap.DesiredBeforePlan {
      if grState.HasWriteKey() {
          writeKeyID, _ := state.NameToKeyID(grState.WriteKey.Key.Name)
          // use writeKeyID
          break
      }
  }

```

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We can use referenced Secret/Configmap fetcher function instead of duplicating.

ardaguclu:
Changes, with this planner and computer approach, make encryption controllers more elegant. I dropped a few comments for some refactorings that will be even useful without the changes here. So I think, we can merge those refactorings first. So this will reduce the number of changed lines in this PR.

pkg/operator/encryption/controllers/key_controller.go
bertinatto: inlined

pkg/operator/encryption/controllers/key_controller.go
bertinatto: this is gone

pkg/operator/encryption/controllers/state_controller.go
bertinatto: inlined

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: this was supposed to be temporary because the preflight controller currently only works with a fixed UDS patch (`unix:///var/run/kmsplugin/kms.sock`)

pkg/operator/encryption/controllers/state_controller.go
bertinatto: inlined

pkg/operator/encryption/controllers/key_controller.go
bertinatto: I'm not sure I understand well; but if we extract this now, it wouldn't have any use right now, correct? Something like this? referencedResourcesFromKeyState

ardaguclu: Ignore my comment. I missed that we rebuild from KeyState.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -297,16 +253,50 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
 	return nil // we made this key earlier
 }
 
-// generateKeySecret builds the key secret for the given key ID and mode.
+// ensureKMSPreflightBeforeKeyCreate gates KMS key persistence on a successful preflight check for the current KMS config hash. Non-KMS modes always pass.
 //
-// For KMS mode it also computes the config hash (from the referenced Secret and
-// ConfigMap it already reads) and gates on the KMS preflight check. The boolean
-// return value signals the outcome:
+// plannedKeySecret is the in-memory key secret from MaterializeKey; its embedded plugin credentials are reused for hashing so we do not re-fetch openshift-config resources.
```
ardaguclu: I think, instead of rebuilding the credentials from KeyState (which is done https://github.com/openshift/library-go/pull/2420/), we should always fetch them from API Server to work with up to date data. Because if the content of referenced Secret/Configmap changes, hash must mismatch in preflight (i.e. `ensureKMSPrelightPassed`). So that process needs to restart with the new hashes. 

Rebuilding referenced data from KeyState may work on stale data which is not ideal. 

So I think, this https://github.com/openshift/library-go/pull/2418/ is the right refactoring. This function can call `fetchReferencedResources`. Am I missing something?. Please let me know your thoughts.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: As far as I understand, this function is not a replacement of `generateKeySecret`?. Why do we need this?

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: We just need like this;

```go
 resources := newCoreClientKMSConfigHasherResourceProvider(c.secretsClient, c.configMapsClient)
  hasher, _ := newKMSConfigHasher(snap.desiredProviderCfg, resources, openshiftConfigNS)
  configHash, _ := hasher.hash(ctx)
  return c.ensureKMSPreflightPassed(ctx, configHash)
```

If hash mismatches, process will restart.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: xref: https://github.com/openshift/library-go/pull/2410#discussion_r3756239755

I think, we don't need this function.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: xref: https://github.com/openshift/library-go/pull/2410#discussion_r3756239755

I think, we don't need this function

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: yes, but we should already have latestKeyID. We don't need to iterate over the keySecrets again.

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: moved

pkg/operator/encryption/controllers/key_controller.go
bertinatto: done

pkg/operator/encryption/controllers/key_controller.go
bertinatto: the problem of removing this function and using `fetchReferencedResources` in `ensureKMSPreflightBeforeKeyCreate` is that we end up calling the api twice for the same ref cm and secret

ardaguclu: Why is that?. Don't we need to call generateKeySecret once (which is the only function that calls ensureKMSPreflightpassed)?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -521,6 +548,60 @@ func (c *keyController) ensureKMSPreflightPassed(ctx context.Context, configHash
...
 }
 
+type encryptionKeyPlan struct {
+	needed         bool
+	keyID          uint64
+	reasons        []string
+	internalReason string
+}
+
+func planNextEncryptionKey(desiredEncryptionState map[schema.GroupResource]state.GroupResourceState, currentMode state.Mode, externalReason string, encryptedGRs []schema.GroupResource, desiredProviderCfg kmsProviderConfig) (*encryptionKeyPlan, error) {
```
ardaguclu, line 486: That looks identical and looks good to me.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -299,14 +266,50 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
...
-// ConfigMap it already reads) and gates on the KMS preflight check. The boolean
-// return value signals the outcome:
+// For KMS mode it also computes the config hash (from credentials embedded in the
+// key state) and gates on the KMS preflight check. The boolean return value signals
+// the outcome:
 //
 //   - (secret, true,  nil) — preflight passed; caller should persist the key.
 //   - (nil,   false, nil) — preflight still in progress; caller should back off.
 //   - (nil,   false, err) — preflight failed or transient error; caller should surface it.
 func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, currentMode state.Mode, apiServerEncryption configv1.APIServerEncryption, desiredProviderCfg kmsProviderConfig, internalReason, externalReason string) (*corev1.Secret, bool, error) {
```
ardaguclu, line 255: `buildEncryptionKeyState` accesses refSecret, refCM already. If it returns those values as well, this function will be simplified (and we don't need round-tripping);

```go
 func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, currentMode state.Mode, apiServerEncryption configv1.APIServerEncryption, desiredProviderCfg kmsProviderConfig, internalReason, externalReason
  string) (*corev1.Secret, bool, error) {
        ks, refSecret, refCM, err := buildEncryptionKeyState(ctx, keyID, currentMode, apiServerEncryption, desiredProviderCfg, c.secretClient, c.configMapClient, internalReason, externalReason, "")
        if err != nil {
                return nil, false, err
        }
        if currentMode == state.KMS {
                resources := &prefetchedKMSConfigHasherResourceProvider{secret: refSecret, configMap: refCM}
                hasher, err := newKMSConfigHasher(desiredProviderCfg, resources, openshiftConfigNS)
                if err != nil {
                        return nil, false, fmt.Errorf("failed to create KMS config hasher: %w", err)
                }
                configHash, err := hasher.hash(ctx)
                if err != nil {
                        return nil, false, fmt.Errorf("failed to compute KMS config hash: %w", err)
                }
                preflightPassed, err := c.ensureKMSPreflightPassed(ctx, configHash)
                if err != nil {
                        return nil, false, err
                }
                if !preflightPassed {
                        return nil, false, nil
                }
        }
        secret, err := secrets.FromKeyState(c.instanceName, ks)
        if err != nil {
                return nil, false, err
        }
        return secret, true, nil
  }
```

ardaguclu: preflight can ignore refCM and refSecret.

bertinatto: I think that's a great suggestion

pkg/operator/encryption/controllers/key_controller.go
bertinatto: you're right, it was no longer necessary after the refactoring you suggested

pkg/operator/encryption/controllers/key_controller.go
bertinatto: done, PTAL

ardaguclu: Would it make sense having this second commit in separate pr, and if it looks good to me, merging it?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -296,68 +263,21 @@ func (c *keyController) validateExistingSecret(ctx context.Context, keySecret *c
...
-	bs := crypto.ModeToNewKeyFunc[currentMode]()
-	ks := state.KeyState{
-		Key: apiserverv1.Key{
-			Name:   fmt.Sprintf("%d", keyID),
-			Secret: base64.StdEncoding.EncodeToString(bs),
-		},
-		Mode:           currentMode,
-		InternalReason: internalReason,
-		ExternalReason: externalReason,
+	ks, refSecret, refCM, err := buildEncryptionKeyState(ctx, keyID, currentMode, apiServerEncryption, desiredProviderCfg, c.secretClient, c.configMapClient, internalReason, externalReason, "")
```
ardaguclu: Would it make sense this function accepts `encryptionKeyPlan`?

bertinatto: a `encryptionKeyPlan` arg would replace `keyID` and `internalReason` (so, 1 fewer arg). However, the other fields of the `encryptionKeyPlan` would be empty. IMO it's better to leave it as is, but I don't have a strong opinion

ardaguclu: Actually that makes sense, we don't need to carry all the data in it. So let's leave it as is.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -183,64 +183,43 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
...
-	if currentMode == state.KMS {
-		var err error
-		desiredProviderCfg, err = newKMSProviderConfig(apiEncryptionConfiguration.KMS)
-		if err != nil {
-			return err
-		}
-	}
-
-	keyPlan, err := planNextEncryptionKey(desiredEncryptionState, currentMode, externalReason, encryptedGRs, desiredProviderCfg)
+	plan, err := planner.PlanNextKey(snap)
```
bertinatto, line 196: this is essentially the same thing: `planner.PlanNextKey(snap) == planNextEncryptionKey(desiredEncryptionState, currentMode, externalReason, encryptedGRs, desiredProviderCfg)`. The only difference is that the former returns an exported type

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,253 @@
...
+// When apiServerClient and operatorClient are non-nil, mode is resolved and modeResolved is set so PlanNextKey / MaterializeKey may be used. When those clients are nil (state-controller path), mode is left unresolved and only ComputeConfig is supported.
+func (p *EncryptionPlanner) Load(ctx context.Context, encryptedGRs []schema.GroupResource) (*EncryptionSnapshot, error) {
+	snap := &EncryptionSnapshot{
+		EncryptedGRs: encryptedGRs,
+	}
+
+	// Resolve mode before reading deployer/key state when mode clients are present.
+	// Callers that lack an APIServer (and unit tests that assert action order) fail
+	// here without listing pods/secrets first.
+	if p.apiServerClient != nil && p.operatorClient != nil {
```
ardaguclu: As far as I understand, we have this logic. Because key controller, preflight controller need mode resolution but state controller does not (state controller has nothing to do with the resolution). 

Just an idea: Would it be better to divide this function into 2:

```go
 // LoadState reads deployer convergence and key secrets. Used by all consumers.
  func (p *EncryptionPlanner) LoadState(ctx context.Context, encryptedGRs []schema.GroupResource) (*EncryptionSnapshot, error) {
      currentConfig, desiredBeforePlan, keySecrets, progressingReason, err := statemachine.GetEncryptionConfigAndState(ctx, p.deployer, p.secretClient, p.encryptionSecretSelector, encryptedGRs)
      // ... populate snap ...
      return snap, nil
  }

func (p *EncryptionPlanner) Load(ctx context.Context, encryptedGRs []schema.GroupResource) (*EncryptionSnapshot, error) {
      // Mode first (same order as old key controller)
      currentMode, externalReason, apiEncryption, err := getCurrentModeReasonAndEncryptionConfig(...)
      if err != nil {
          return nil, err
      }
  
      snap, err := p.LoadState(ctx, encryptedGRs)
      if err != nil {
          return nil, err
      }

      snap.CurrentMode = currentMode
      snap.ExternalReason = externalReason
      snap.APIEncryption = apiEncryption

      if len(snap.ProgressingReason) == 0 && currentMode == state.KMS {
          snap.desiredProviderCfg, err = newKMSProviderConfig(apiEncryption.KMS)
          if err != nil {
              return nil, err
          }
      }
      return snap, nil
  }
```
State controller calls LoadState -> ComputeConfig. Key controller and preflight call Load -> PlanNextKey -> MaterializeKey -> ComputeConfig.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,253 @@
...
+		Reasons: plan.Reasons,
+	}, nil
+}
+
+// ComputeConfig builds the desired encryption-config secret from the snapshot's persisted keys, optionally including opts.PlannedKey.
+func (p *EncryptionPlanner) ComputeConfig(snap *EncryptionSnapshot, opts ComputeConfigOptions) (*EncryptionPlanResult, error) {
+	if snap == nil {
+		return nil, fmt.Errorf("snapshot is required")
+	}
+	if len(snap.ProgressingReason) > 0 {
```
ardaguclu: We have decided that preflight controller should continue progress even if there isn't any convergence between apiservers (i.e. progressingreason > 0). So we should remove this. Key controller and state controller should continue requeueing separately in their logic..

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,253 @@
...
+
+// PlanNextKey deterministically decides whether a new key is needed. It does not generate key material. Requires a snapshot with mode resolved.
+func (p *EncryptionPlanner) PlanNextKey(snap *EncryptionSnapshot) (*KeyPlan, error) {
+	if snap == nil {
+		return nil, fmt.Errorf("snapshot is required")
+	}
+	if !snap.modeResolved {
+		return nil, fmt.Errorf("PlanNextKey requires a snapshot with mode resolved (apiServerClient and operatorClient must be set on the planner)")
+	}
+	if len(snap.ProgressingReason) > 0 {
```
ardaguclu: same https://github.com/openshift/library-go/pull/2410/changes/db91d5faec5ecee133162d6a0a2c00ed3be35357#r3801767288

pkg/operator/encryption/controllers/encryption_planner.go
ardaguclu: After this separation, we can remove `modeResolved` field.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,253 @@
...
+	currentConfig, desiredBeforePlan, keySecrets, progressingReason, err := statemachine.GetEncryptionConfigAndState(ctx, p.deployer, p.secretClient, p.encryptionSecretSelector, encryptedGRs)
+	if err != nil {
+		return nil, err
+	}
+	snap.CurrentConfig = currentConfig
+	snap.DesiredBeforePlan = desiredBeforePlan
+	snap.KeySecrets = keySecrets
+	snap.ProgressingReason = progressingReason
+
+	if snap.modeResolved && len(progressingReason) == 0 && snap.CurrentMode == state.KMS {
```
ardaguclu: Logic about progressionReason varies on each caller. So we can update this;

```go
if snap.CurrentMode == state.KMS {
      snap.desiredProviderCfg, err = newKMSProviderConfig(snap.APIEncryption.KMS)
      if err != nil {
          return nil, err
      }
  }
```

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,253 @@
...
+
+// PlannedEncryptionKey is a materialized in-memory key secret ready to persist or include in a candidate config.
+type PlannedEncryptionKey struct {
+	Secret  *corev1.Secret
+	KeyID   uint64
+	Reasons []string
+}
+
+// ComputeConfigOptions controls how ComputeConfig builds the encryption-config secret.
+type ComputeConfigOptions struct {
```
ardaguclu: Do we really need this struct?. Can't we just pass PlannedKey?

ardaguclu: Just one note for that review session: Today I only reviewed second commit.

pkg/operator/encryption/controllers/encryption_computer.go
```diff
@@ -0,0 +1,114 @@
+package controllers
```
ardaguclu: Do we need this abstraction (encryption_computer)?. Preflight and key controller can directly call: planner.Load -> PlanNextKey -> MaterializeKey -> ComputeConfig. It would be easier to read the logic from each controller. 
Moreover, we wouldn't maintain another layer which may not be needed.

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: this sounds like a good idea, but doing that introduces room for misusing the  planner's api, like calling `LoadState() + PlanNextKey()`. What if we split `EncryptionSnapshot` into 2 different types, `EncryptionStateSnapshot` and `KeyPlanningSnapshot`, so that `LoadState()` and `Load()` would return them, respectively? That way we'd use the type system to prevent the planner misuse. I experimented with the idea in a new commit to make it easier to review and maybe delete. PTAL ff5df94880425d446c52f38351f14f08d77e060b

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: good point, I attempted to fix this in a separate commit: 88188ee9e6c90c63d94c1993b0bdaaf0091fb613. If we're happy with it, I can incorporate the changes into the second commit

pkg/operator/encryption/controllers/encryption_computer.go
bertinatto: I agree, I think we don't need it; this can be simplified into a method in `kmsPreflightController`. I'll leave it as is for now and until we're happy with the planner

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -161,14 +161,30 @@ func (p *EncryptionPlanner) Load(ctx context.Context, encryptedGRs []schema.Grou
 	return snap, nil
 }
 
+// LoadForPreflight resolves encryption mode and key-planning state for KMS preflight.
+// Unlike Load, when deployer convergence is still in progress it still loads persisted
+// key secrets and computes DesiredBeforePlan so PlanNextKey and ComputeConfig can proceed.
+func (p *EncryptionPlanner) LoadForPreflight(ctx context.Context, encryptedGRs []schema.GroupResource) (*KeyPlanningSnapshot, error) {
```
ardaguclu: I think, this may contradict our goal that key controller and preflight use the exact same flow to generate the same outcome. Now preflight diverges from the key controller. 

Maybe we can have a flag in Snapshot like ignore progress reason, etc.. This flag can be used to what to do in `Load` function (short cut or proceed) instead of extending it to newer LoadForPreflight.

pkg/operator/encryption/controllers/encryption_planner.go
ardaguclu: I think this separation looks nice to me.

pkg/operator/encryption/controllers/encryption_planner.go
ardaguclu: ref: https://github.com/openshift/library-go/pull/2410#discussion_r3810851881

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -184,96 +183,43 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
 }
 
 func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext factory.SyncContext, encryptedGRs []schema.GroupResource) error {
-	currentMode, externalReason, apiEncryptionConfiguration, err := getCurrentModeReasonAndEncryptionConfig(ctx, c.apiServerClient, c.operatorClient, c.unsupportedConfigPrefix)
+	planner := NewEncryptionPlanner(c.instanceName, c.unsupportedConfigPrefix, c.deployer, c.secretClient, c.configMapClient, c.apiServerClient, c.operatorClient, c.encryptionSecretSelector)
+	snap, err := planner.Load(ctx, encryptedGRs)
```
ardaguclu: nit: Would it make sense NewEncryptionPlanner returns an interface with 5 functions `Load`, `PlanNextKey`, `ComputeConfig`, `LoadState`, `MaterializeKey` instead of concrete type?. It is totally up to your taste.

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: this is a good point. I dropped that additional commit and implemented your suggestion of adding a boolean in `KeyPlanningSnapshot`, PTAL.

as we dicussed offline, after implementing it, I found it a bit clumsy because that type was supposed to be an output (snapshot), but now it contains a field that's not part of the snapshot and is used as an input.

I experimented with adding a `ListOptions` to the `Load()` method and I think it looks OK. I left it in a separate commit to make it easier to delete, PTAL: https://github.com/openshift/library-go/pull/2410/changes/e4817cd60c02e73ef70ac545353985a1d1a9cb74

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: you're right, we don't need this type. Removed.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,271 @@
...
+
+	snap := &KeyPlanningSnapshot{
+		State:              *stateSnap,
+		CurrentMode:        currentMode,
+		ExternalReason:     externalReason,
+		APIEncryption:      apiEncryption,
+		desiredProviderCfg: noopKMSProviderConfig{},
+	}
+
+	if currentMode == state.KMS {
```
ardaguclu, line 165: Current code does not populate `desiredProviderCfg`, if progressingReason is non-empty. But this is benign in-memory operation. So we should be fine. Also, we will likely use these populated data in in-place field updates.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,271 @@
...
+		return nil, err
+	}
+
+	stateSnap, err := p.LoadState(ctx, encryptedGRs)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(stateSnap.ProgressingReason) > 0 && opts.ListKeysWhileProgressing {
+		keySecrets, err := secrets.ListKeySecrets(ctx, p.secretClient, p.encryptionSecretSelector)
```
ardaguclu, line 149: `GetEncryptionConfigAndState` function in `LoadState` short cuts, when there is no convergence. Therefore, we are doing this extra operation for preflight controller to have the necessary data. I think, it would be great if we document the reasons why we do this.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,271 @@
...
+	CurrentMode state.Mode
+}
+
+// LoadOptions controls how Load reads key-planning state.
+type LoadOptions struct {
+	// ListKeysWhileProgressing lists persisted key secrets even when the
+	// deployer has not converged. Default false so the key controller does
+	// not List during API-server rollouts. Preflight sets this so PlanNextKey
+	// can run. ProgressingReason is still populated in either case.
+	ListKeysWhileProgressing bool
```
ardaguclu, line 91: I think, this is the way of idiomatic k8s, just as we use from client-go. So, this looks good to me.

pkg/operator/encryption/controllers/encryption_planner.go
```diff
@@ -0,0 +1,271 @@
...
+		KeySecrets:    keySecrets,
+		CurrentMode:   opts.CurrentMode,
+		PlannedKey:    opts.PlannedKey,
+	}
+
+	if state.CurrentConfig == nil && len(keySecrets) == 0 {
+		return out, nil
+	}
+
+	desiredState := statemachine.GetDesiredEncryptionState(state.CurrentConfig, keySecrets, state.EncryptedGRs)
```
ardaguclu: We don't need to call that function, if there is no plannedKey. What about something like this:

```go
out.DesiredState = snap.DesiredBeforePlan

if opts.PlannedKey != nil && opts.PlannedKey.Secret != nil {
desiredState := statemachine.GetDesiredEncryptionState(state.CurrentConfig, keySecrets, state.EncryptedGRs)
out.DesiredState = desiredState
}

cfg, err := encryptiondata.FromEncryptionState(out.DesiredState)
if err != nil {
	return nil, fmt.Errorf("failed to build encryption config: %w", err)
}
out.EncryptionConfig = cfg
```

pkg/operator/encryption/controllers/encryption_planner.go
ardaguclu: > I experimented with adding a ListOptions to the Load() method and I think it looks OK.

I think, that looks good.

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -183,64 +183,43 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
...
-
-	currentConfig, desiredEncryptionState, secrets, isProgressingReason, err := statemachine.GetEncryptionConfigAndState(ctx, c.deployer, c.secretClient, c.encryptionSecretSelector, encryptedGRs)
-	if err != nil {
+	planner := NewEncryptionPlanner(c.instanceName, c.unsupportedConfigPrefix, c.deployer, c.secretClient, c.configMapClient, c.apiServerClient, c.operatorClient, c.encryptionSecretSelector)
+	snap := &KeyPlanningSnapshot{}
+	if err := planner.Load(ctx, encryptedGRs, snap); err != nil {
 		return err
 	}
-	if len(isProgressingReason) > 0 {
+	if len(snap.State.ProgressingReason) > 0 {
```
ardaguclu, line 191: nit: Short documentation about `Load` should populate `snap`, if err is nil.

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: agreed; addded  a comment

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: ack, integrated

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: ack, thanks for pointing that out

pkg/operator/encryption/controllers/encryption_planner.go
bertinatto: took the suggestion, `ComputeConfig` now reuses `snap.DesiredBeforePlan` that was computed before

pkg/operator/encryption/controllers/key_controller.go
bertinatto: done

pkg/operator/encryption/controllers/key_controller_test.go
```diff
@@ -623,6 +630,32 @@ func TestKeyController(t *testing.T) {
 			},
 		},
 
+		{
+			name: "degraded when KMS preflight check failed",
```
ardaguclu, line 634: just comment: Unrelated but useful test.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -354,6 +367,45 @@ func NewKMSPreflightController(
...
 }
 
+func (c *kmsPreflightController) computeEncryptionConfiguration(ctx context.Context) (*corev1.Secret, error) {
+	// TODO: remove this fallback once EncryptionConfigurationComputer is deleted.
+	if c.encryptionConfigurationComputer != nil {
+		return c.encryptionConfigurationComputer.ComputeEncryptionConfiguration(ctx)
+	}
+
+	planner := NewEncryptionPlanner(c.instanceName, c.unsupportedConfigPrefix, c.encryptionDeployer, c.secretsClient, c.configMapsClient, c.apiServerClient, c.operatorClient, c.encryptionSecretSelector)
+	snap, err := planner.Load(ctx, c.provider.EncryptedGRs(), LoadOptions{ListKeysWhileProgressing: true})
```
ardaguclu, line 379: nit: Document why we set `ListKeysWhileProgressing`

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -354,6 +367,45 @@ func NewKMSPreflightController(
...
+	if err != nil {
+		return nil, err
+	}
+	if snap.CurrentMode != "" && snap.CurrentMode != state.KMS {
+		return nil, fmt.Errorf("preflight encryption config computation requires KMS mode, got %q", snap.CurrentMode)
```
ardaguclu, line 384: Is the returning error correct action here?. We work with up to date data. Cluster admin may enable KMS, key controller asks preflight to check it. But before preflight runs, cluster admin may switch to another encryption mode.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -354,6 +367,45 @@ func NewKMSPreflightController(
...
+	}
+
+	plan, err := planner.PlanNextKey(snap)
+	if err != nil {
+		return nil, err
+	}
+
+	var plannedKey *PlannedEncryptionKey
+	if plan.Needed {
+		plannedKey, err = planner.MaterializeKey(ctx, snap, plan, "")
```
ardaguclu: Shouldn't we pass kmsEndpointOverride?

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -354,6 +367,45 @@ func NewKMSPreflightController(
...
+			return nil, err
+		}
+	}
+
+	result, err := planner.ComputeConfig(&snap.State, plannedKey)
+	if err != nil {
+		return nil, err
+	}
+	if result.EncryptionSecret == nil {
+		return nil, fmt.Errorf("no encryption key secrets available to compute preflight encryption config")
```
ardaguclu, line 405: I think, this is the right action (returning error). But I will still think about this.

pkg/operator/encryption/controllers/kms_preflight_compute_test.go
```diff
@@ -0,0 +1,633 @@
+package controllers
```
ardaguclu, line 1: Overall, this test file requires some refactoring for idiomatic Go.

ardaguclu (commented):
I have 2 comments in preflight, the rest is mostly about cosmetic.

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: added a comment

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: that's a good point; the race you pointed out will result in this piece returning the error, which make the cluster degraded until the next resync. Would you rather log a warning instead?

ardaguclu: I think, we should log and `return nil, nil` (or another variant that triggers the clean up.)

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: good catch; `kmsEndpointOverride` was introduced before https://github.com/openshift/library-go/pull/2430, when the preflight had a hardcoded endoint `unix:///var/run/kmsplugin/kms.sock`. Now this is no longer true, so I need to go back to the first commit and remove that param because it's no longer needed. This means that an empty value is correct here (until the param exists). Do you agree?

ardaguclu: I agree, we don't need to override. I think this will simplify the first commit slightly.

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: I'll check this after the meeting.

bertinatto: ack

pkg/operator/encryption/controllers/kms_preflight_controller.go
bertinatto: ack

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -354,6 +367,47 @@ func NewKMSPreflightController(
...
+	}
+
+	plan, err := planner.PlanNextKey(snap)
+	if err != nil {
+		return nil, err
+	}
+
+	var plannedKey *PlannedEncryptionKey
+	if plan.Needed {
+		plannedKey, err = planner.MaterializeKey(ctx, snap, plan)
```
bertinatto, line 394: removed the `kmsEndpointOverride` param

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: https://github.com/openshift/library-go/pull/2435

pkg/operator/encryption/controllers/kms_preflight_controller.go
ardaguclu: Returning error is the right action.

ardaguclu:
I've opened this https://github.com/openshift/library-go/pull/2435 which is an independent PR touching the preflight states. 

This PR looks good to me (other than unit test refactoring comment). We just need to see `e2e-aws-operator-encryption-kms`, `e2e-aws-encryption-kms-single-node` green.

ardaguclu:
ref: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2276#issuecomment-5406035973
/lgtm
/approve
/hold
I think, we may need to run this also in authentication and openshift-apiserver.

bertinatto:
> ref: [openshift/cluster-kube-apiserver-operator#2276 (comment)](https://github.com/openshift/cluster-kube-apiserver-operator/pull/2276#issuecomment-5406035973) /lgtm /approve /hold I think, we may need to run this also in authentication and openshift-apiserver.

Created the following PRs and CI is green:

- https://github.com/openshift/cluster-kube-apiserver-operator/pull/2276
- https://github.com/openshift/cluster-openshift-apiserver-operator/pull/758
- https://github.com/openshift/cluster-authentication-operator/pull/975

During KMS standup, we agreed to merge this PR.

/hold cancel

PR #2414, opened 2026-08-10
OPRUN-4711: configobserver: add ObserveTLSSecurityProfileWithGroupPaths

tmshort:
https://redhat.atlassian.net/browse/OPRUN-4711

Implements the library-go portion of the [TLS curves configuration enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/security/api-tls-curves-config.md): observe the `TLSSecurityProfile.groups` field so operators can propagate curve (group) preferences to their operands alongside `minTLSVersion` and `cipherSuites`.

## What this adds

**`pkg/operator/configobserver/apiserver`**
- `ObserveTLSSecurityProfileWithGroupPaths` — like `ObserveTLSSecurityProfileWithPaths`, but also observes group preferences at a caller-supplied path. The version/cipher/group observations share one resolved profile spec so they cannot drift.
- When the Go runtime is in FIPS mode, non-FIPS-approved groups are dropped before storing, so consuming components receive an already-safe list. The filtering is in `tls_fips.go` (same package, unexported) and can be removed once Go 1.27 without OpenSSL FIPS is supported — at that point `crypto/tls` will filter non-approved curves at negotiation time without crashing, making the pre-filtering redundant.
- The filtered group list is propagated verbatim, including an empty list. Per the enhancement, if the resulting groups cannot negotiate with a peer the handshake is expected to fail rather than fall back to unrequested curves.

## Design notes (addressing earlier review)

An earlier revision had the FIPS filtering split across `pkg/crypto` as exported helpers (`IsFIPSEnabled`, `FIPSApprovedTLSGroups`). After review feedback those are now unexported and co-located with the observer in `pkg/operator/configobserver/apiserver/tls_fips.go` — the only place they're needed.

## Test plan

- [x] `go test ./pkg/operator/configobserver/apiserver/...`
- [x] same under `GODEBUG=fips140=on`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
+++ b/pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
@@ -3,6 +3,7 @@ package apiserver
 import (
 	"fmt"
 	"reflect"
+	"slices"
 
 	"k8s.io/klog/v2"
 
@@ -17,24 +18,38 @@ import (
 // ObserveTLSSecurityProfile observes APIServer.Spec.TLSSecurityProfile field and sets
 // the ServingInfo.MinTLSVersion, ServingInfo.CipherSuites fields of observed config
 func ObserveTLSSecurityProfile(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
-	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, []string{"servingInfo", "minTLSVersion"}, []string{"servingInfo", "cipherSuites"})
+	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, []string{"servingInfo", "minTLSVersion"}, []string{"servingInfo", "cipherSuites"}, nil)
 }
 
 // ObserveTLSSecurityProfileWithPaths is like ObserveTLSSecurityProfile, but accepts
 // custom paths for ServingInfo.MinTLSVersion and ServingInfo.CipherSuites fields of observed config.
 func ObserveTLSSecurityProfileWithPaths(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, minTLSVersionPath, cipherSuitesPath []string) (map[string]interface{}, []error) {
-	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, minTLSVersionPath, cipherSuitesPath)
+	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, minTLSVersionPath, cipherSuitesPath, nil)
+}
+
+// ObserveTLSSecurityProfileWithGroupPaths is like ObserveTLSSecurityProfileWithPaths but also
+// observes the TLS group (curve) preferences at groupsPath. Group names are stored as []string
+// matching the TLSGroup constants from openshift/api and can be passed directly to operands
+// that accept group names as CLI arguments.
+func ObserveTLSSecurityProfileWithGroupPaths(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, minTLSVersionPath, cipherSuitesPath, groupsPath []string) (map[string]interface{}, []error) {
+	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, minTLSVersionPath, cipherSuitesPath, groupsPath)
 }
 
 // ObserveTLSSecurityProfileToArguments observes APIServer.Spec.TLSSecurityProfile field and sets
 // the tls-min-version and tls-cipher-suites fileds of observedConfig.apiServerArguments
 func ObserveTLSSecurityProfileToArguments(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}) (map[string]interface{}, []error) {
-	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, []string{"apiServerArguments", "tls-min-version"}, []string{"apiServerArguments", "tls-cipher-suites"})
+	return innerTLSSecurityProfileObservations(genericListers, recorder, existingConfig, []string{"apiServerArguments", "tls-min-version"}, []string{"apiServerArguments", "tls-cipher-suites"}, nil)
 }
 
-func innerTLSSecurityProfileObservations(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, minTLSVersionPath, cipherSuitesPath []string) (ret map[string]interface{}, _ []error) {
+// innerTLSSecurityProfileObservations is the shared implementation for all Observe* functions.
+// When groupsPath is non-nil, TLS group preferences are also observed and stored at that path.
+func innerTLSSecurityProfileObservations(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, minTLSVersionPath, cipherSuitesPath, groupsPath []string) (ret map[string]interface{}, _ []error) {
 	defer func() {
-		ret = configobserver.Pruned(ret, minTLSVersionPath, cipherSuitesPath)
+		paths := [][]string{minTLSVersionPath, cipherSuitesPath}
+		if len(groupsPath) > 0 {
+			paths = append(paths, groupsPath)
+		}
+		ret = configobserver.Pruned(ret, paths...)
 	}()
 
 	listers := genericListers.(APIServerLister)
@@ -52,6 +67,15 @@ func innerTLSSecurityProfileObservations(genericListers configobserver.Listers, 
 		// keep going on read error from existing config
 	}
 
+	var currentGroups []string
+	if len(groupsPath) > 0 {
+		var groupsErr error
+		currentGroups, _, groupsErr = unstructured.NestedStringSlice(existingConfig, groupsPath...)
+		if groupsErr != nil {
+			errs = append(errs, fmt.Errorf("failed to retrieve groups: %v", groupsErr))
+		}
+	}
+
 	apiServer, err := listers.APIServerLister().Get("cluster")
 	if errors.IsNotFound(err) {
 		klog.Warningf("apiserver.config.openshift.io/cluster: not found")
@@ -60,8 +84,12 @@ func innerTLSSecurityProfileObservations(genericListers configobserver.Listers, 
 		return existingConfig, append(errs, err)
 	}
 
+	// Resolve the effective profile spec once and share it between the cipher and
+	// group extractors so the two cannot resolve different profiles.
+	profileSpec := resolveProfileSpec(apiServer.Spec.TLSSecurityProfile)
+
 	observedConfig := map[string]interface{}{}
-	observedMinTLSVersion, observedCipherSuites := getSecurityProfileCiphers(apiServer.Spec.TLSSecurityProfile)
+	observedMinTLSVersion, observedCipherSuites := getSecurityProfileCiphers(profileSpec)
 	if err = unstructured.SetNestedField(observedConfig, observedMinTLSVersion, minTLSVersionPath...); err != nil {
 		return existingConfig, append(errs, err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/crypto/fips.go
```diff
@@ -0,0 +1,32 @@
...
+var fipsTLSGroups = map[configv1.TLSGroup]struct{}{
+	configv1.TLSGroupSecP256r1: {},
+	configv1.TLSGroupSecP384r1: {},
+	configv1.TLSGroupSecP521r1: {},
+}
+
+// IsFIPSEnabled reports whether the Go runtime is operating in FIPS 140 mode.
+func IsFIPSEnabled() bool {
+	return fips140.Enabled()
+}
```
tmshort: This is an alternative method used in an operator to determine whether FIPS mode is enabled
https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/deployment.go#L57

tmshort: But it appears that's an old approach, this method was introduced recent.y

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -107,3 +115,103 @@ func getSecurityProfileCiphers(profile *configv1.TLSSecurityProfile) (string, []
...
+
+// getSecurityProfileGroups returns the TLS group preference names for the given
+// TLSSecurityProfile as []string. The strings match the TLSGroup constants from
+// openshift/api and can be passed directly to operands that accept group names as
+// CLI arguments. Returns nil when no groups are configured (e.g. a Custom profile
+// with an empty groups list).
+func getSecurityProfileGroups(profile *configv1.TLSSecurityProfile) []string {
+	var profileType configv1.TLSProfileType
+	if profile == nil {
+		profileType = crypto.DefaultTLSProfileType
```
smith-xyz: is this the proper default? seems like it could be - but any reason why someone should even call this if profile is nil?

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -107,3 +115,103 @@ func getSecurityProfileCiphers(profile *configv1.TLSSecurityProfile) (string, []
...
+		if fips && !crypto.IsFIPSApprovedTLSGroup(g) {
+			klog.V(4).Infof("Dropping TLS group %q: not FIPS-approved", g)
+			continue
+		}
+		groups = append(groups, string(g))
+	}
+	return groups
+}
+
+func innerTLSSecurityProfileObservationsWithGroups(genericListers configobserver.Listers, recorder events.Recorder, existingConfig map[string]interface{}, minTLSVersionPath, cipherSuitesPath, groupsPath []string) (ret map[string]interface{}, _ []error) {
```
smith-xyz: need to check more closely but this looks very similar to `innerTLSSecurityProfileObservations`

can we extend the existing inner function with groupsPath []string (empty = skip) to collapse two functions into one?

tmshort: Yeah, the `inner` functions, not being exposed API, can be consolidated.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
tmshort: If the APIServer/cluster resource is not found (too early in the startup/install), then it could be that this is called with `nil`, and that the default is used. When the APIserver/cluster is created, then the watches will find it, and a real profile will be found.

tmshort:
Looking for review: @deads2k @benluddy @p0lyn0mial

tmshort:
Rebased onto main, and flattened @sanchezl's commits on top of mine.
See: https://github.com/tmshort/library-go/pull/1

tmshort:
@coderabbitai review

tmshort:
Rebased, and added some stuff from @sanchezl

tmshort:
@coderabbitai review

pkg/crypto/fips.go
```diff
@@ -0,0 +1,66 @@
...
+	"crypto/fips140"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// fipsTLSGroups is the allowlist of TLSGroup values treated as FIPS-approved.
+// Only the NIST P-curves qualify: their domain parameters are specified in
+// NIST SP 800-186 and their use for ECDH key establishment in SP 800-56A.
+// (The earlier "FIPS 186-5" citation was inaccurate: that is the Digital
+// Signature Standard, which governs signatures, not TLS key-exchange groups.)
```
jsafrane: What earlier citation? This is a brand new file + comment.

pkg/crypto/tls_handshake_failsafe.go
```diff
@@ -0,0 +1,123 @@
...
+//   - It cannot be rejected at admission time: the config is legitimately
+//     correct in a fleet where every peer speaks TLS 1.3. minTLSVersion is a
+//     floor, not a guarantee that 1.2 is ever used.
+//   - It is more likely under FIPS + post-quantum intent, where the approved PQC
+//     options ARE the 1.3-only hybrids.
+//
+// So instead of rejecting, we fail SAFE: when minVersion permits TLS 1.2 but no
+// requested group can service a 1.2 handshake, we APPEND (never replace) the
+// classical fallback groups. Appending preserves the administrator's
+// post-quantum preference for 1.3 peers while restoring 1.2 capability.
```
jsafrane: I admit I got lost in the reasoning here. 

User **explicitly** asks for only specific TLS 1.3 curves and at the same time allows minTLSVersion=1.2.

Why does the code assume the user wants **any** 1.2 curves? They explicitly listed curves they want, didn't they? What is the list good for. if OCP does not respect it?

The [KEP](https://github.com/openshift/enhancements/blob/master/enhancements/security/api-tls-curves-config.md) says:

> Runtime behavior: If no mutually supported groups (or ciphers) remain after filtering, TLS handshakes will fail with errors like "handshake failure" (for cipher suites) or "no shared group" (for groups). This is the expected and desired behavior — it ensures only supported cryptographic parameters are used.

If you want a different behavior, then please update the KEP. Not all components use library-go and they must append the same 1.2 curves.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -76,17 +103,87 @@ func innerTLSSecurityProfileObservations(genericListers configobserver.Listers,
...
+		// every reconcile of a steady-state augmented profile would spam warnings.
+		if groupsAugmented && groupsChanged {
+			recorder.Warningf("ObserveTLSSecurityProfile",
+				"TLS profile groups were all TLS 1.3-only but minTLSVersion permits TLS 1.2; "+
+					"appended classical fallback groups so TLS 1.2 handshakes can complete: %q", observedGroups)
+		}
+		// Symmetry with the augmentation warning: if the admin configured groups
+		// but every one was dropped (unsupported by this Go build, or not
+		// FIPS-approved under FIPS), the operand silently falls back to its own
+		// defaults — most notably an all-PQ Custom profile on a FIPS cluster. Warn
```
jsafrane: Again, user explicitly listed some curves. The operator should not fall back to its own list. IMO, if the final list of curves is empty, then let it be empty and let TLS fail hard.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -76,17 +103,87 @@ func innerTLSSecurityProfileObservations(genericListers configobserver.Listers,
...
-		profileType = crypto.DefaultTLSProfileType
-	} else {
+// groupsEqual compares two TLS group-name slices treating nil and empty as
+// equal. This prevents a first reconcile (currentGroups nil) against an empty
+// observation from being mistaken for a change.
+func groupsEqual(a, b []string) bool {
+	if len(a) == 0 && len(b) == 0 {
+		return true
+	}
+	return reflect.DeepEqual(a, b)
```
jsafrane: `slices.Equal()`?

jsafrane (commented):
If you plement a KEP, then please link it next time!

I did not read the full code, the design direction needs to be clarified first.

pkg/crypto/fips.go
tmshort: This was an edit as part of this PR. Being a new file, this history of edits is not necessary. @sanchezl

tmshort: Good catch — removed. There was no earlier citation; the parenthetical was a leftover. The comment now just states the NIST SP 800-186 / SP 800-56A basis for the P-curve allowlist.

pkg/crypto/tls_handshake_failsafe.go
tmshort: Agreed, and you're right per the KEP. I've removed the entire fail-safe mechanism (`tls_handshake_failsafe.go` and its use in the observer). The observer no longer appends any TLS 1.2 curves: the administrator's group list is honored verbatim, and a legacy-client handshake failure is the intended, documented outcome. This also avoids the divergence you noted — components that consume the profile directly no longer have to replicate an augmentation step that only library-go performed.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
tmshort: Done. The observer now stores the filtered list verbatim, including an empty list — it no longer omits the path or substitutes defaults. An empty result means no mutually-supported groups remain and TLS fails hard, matching the KEP's "expected and desired" behavior.

Related: FIPS omission is no longer done in the observer at all. Per the KEP that filtering is the consuming component's responsibility, so it's now an opt-in `pkg/crypto` helper (`IsFIPSEnabled` + `FIPSApprovedTLSGroups`) rather than something baked into the observation.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
tmshort: Done — replaced the bespoke `groupsEqual` helper with `slices.Equal` (it already treats nil and empty as equal, which is the nil-vs-empty first-reconcile case).

tmshort:
> If you implement a KEP, then please link it next time!
> 
> I did not read the full code, the design direction needs to be clarified first.

Thanks for the info; I was unaware of the KEP. A lot of the testing and the failsafe behavior (which is apparently not needed) was implemented by @sanchezl. I've simplified this PR to match the KEP's behavior, so we can have the minimum necessary to allow library-go users to proceed. 

I did keep an API to allow components to filter for FIPS, but it's not part of the observer (which returns the curves unmodified). This allows components to filter for FIPS without having to maintain the filter themselves. It's not part of the KEP, but it's also a non-required/optional part that still puts the FIPS filtering on components.

@sanchezl, if you want to enhance this, I would propose another PR be made after this merges.

tmshort:
@coderabbitai review

pkg/crypto/fips.go
```diff
@@ -0,0 +1,76 @@
...
+//
+//	groups := profileSpec.Groups
+//	if crypto.IsFIPSEnabled() {
+//		groups = crypto.FIPSApprovedTLSGroups(groups)
+//	}
+//
+// This is a pure classification against the allowlist above; it does not itself
+// check IsFIPSEnabled, so it stays unit-testable without a FIPS runtime and
+// callers keep explicit control over when filtering applies.
+func FIPSApprovedTLSGroups(groups []configv1.TLSGroup) []configv1.TLSGroup {
```
jsafrane: Do you plan to use this function in your operator? Since operators should not filter, I personally think it's not very useful. Would you mind removing `fips.go` from the PR?

The observer below is good to go.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
```diff
@@ -99,11 +146,37 @@ func getSecurityProfileCiphers(profile *configv1.TLSSecurityProfile) (string, []
...
+// openshift/api and can be passed directly to operands that accept group names as
+// CLI arguments.
+//
+// The groups are returned verbatim: no groups are dropped and none are injected.
+// FIPS-mode omission of non-approved groups is the consuming component's
+// responsibility (crypto.IsFIPSEnabled + crypto.FIPSApprovedTLSGroups), matching
+// the api-tls-curves-config enhancement which places that filtering on the
+// component, not on this observer. If the resulting groups cannot negotiate with a
+// peer the handshake is expected to fail rather than fall back to unrequested
+// curves.
```
jsafrane: > FIPS-mode omission of non-approved groups is the consuming component's responsibility 

Is it? I understood the KEP as we should pass it down to a TLS sever / client code and let it fail if  the list of curves (or algorighms) is not FIPS compliant.

pkg/crypto/fips.go
tmshort: The KEP states:

>FIPS mode constraint: X25519 and X25519MLKEM768 are not FIPS-approved. Components running in FIPS mode must omit these groups. The FIPS-approved post-quantum alternatives (SecP256r1MLKEM768, SecP384r1MLKEM1024) require Go 1.26+ and are not currently supported.

That clearly states that operators must omit these groups, which is contrary to what you're saying.

pkg/operator/configobserver/apiserver/observe_tlssecurityprofile.go
tmshort: Yes, it's in the KEP. See above.

pkg/crypto/fips.go
candita: It's not an old method, it is a method we created in collaboration with Hypershift, where the FIPS setting may not exist on the node, but on the management interface.

pkg/crypto/fips.go
candita: We are running Go 1.26+ in OCP 5.0, and the library reflects this in https://github.com/openshift/library-go/blob/master/go.mod#L3

tmshort:
With the Hypershift angle (requiring additional checks to know if FIPS is enabled) I'd suggest keeping the filtering API available for that use case. It could still be pulled into the observer, but keeping the API available separately is important.

tmshort:
@jsafrane where are we with this PR?

jsafrane:
/lgtm
/approve
Thanks for your patience!

PR #2415, opened 2026-08-10
kms: make getCurrentModeReasonAndEncryptionConfig standalone

bertinatto:
## Summary
- Promote `getCurrentModeReasonAndEncryptionConfig` to a package-level helper with explicit APIServer/operator clients and unsupported-config prefix.
- Keep a thin `keyController` method wrapper so existing call sites are unchanged.
- Point the unit test at the standalone function directly.

This is intentionally a small, independently mergeable precursor (same idea as exporting `GetDesiredEncryptionState`) so later EncryptionPlanner work can reuse mode resolution without inventing a parallel helper.

## Test plan
- [x] `go test ./pkg/operator/encryption/controllers/ -run TestGetCurrentModeReasonAndEncryptionConfig`

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal encryption configuration state retrieval while preserving existing behavior.
  * Improved flexibility in how encryption configuration details are obtained and evaluated.

* **Tests**
  * Updated automated coverage to validate the revised encryption configuration retrieval flow.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -184,7 +184,7 @@ func (c *keyController) sync(ctx context.Context, syncCtx factory.SyncContext) (
 }
 
 func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext factory.SyncContext, encryptedGRs []schema.GroupResource) error {
-	currentMode, externalReason, apiEncryptionConfiguration, err := c.getCurrentModeReasonAndEncryptionConfig(ctx)
+	currentMode, externalReason, apiEncryptionConfiguration, err := getCurrentModeReasonAndEncryptionConfig(ctx, c.apiServerClient, c.operatorClient, c.unsupportedConfigPrefix)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ardaguclu:
This is correct move in any scenario. 
/lgtm
/approve

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -395,17 +395,28 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
 }
 
 func (c *keyController) getCurrentModeReasonAndEncryptionConfig(ctx context.Context) (state.Mode, string, configv1.APIServerEncryption, error) {
-	apiServer, err := c.apiServerClient.Get(ctx, "cluster", metav1.GetOptions{})
+	return getCurrentModeReasonAndEncryptionConfig(ctx, c.apiServerClient, c.operatorClient, c.unsupportedConfigPrefix)
```
ardaguclu: nit: What do we lose if key controller directly use the standalone, instead of this?

PR #2420, opened 2026-08-10
kms: extract referencedResourcesFromKeyState helper

bertinatto:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved encryption configuration handling when restoring referenced credentials and configuration data.
  - Added validation for missing or incomplete referenced resources, with clearer error reporting.
  - Ensured recovered resources retain the expected namespace, names, and stored values.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -394,6 +394,47 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
 	return secret, true, nil
 }
 
+// referencedResourcesFromKeyState rebuilds the Secret/ConfigMap objects the hasher expects from credentials already embedded in the planned key state.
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -394,6 +394,47 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
 	return secret, true, nil
 }
 
+// referencedResourcesFromKeyState rebuilds the Secret/ConfigMap objects the hasher expects from credentials already embedded in the planned key state.
+func referencedResourcesFromKeyState(ks state.KeyState, desiredProviderCfg kmsProviderConfig) (*corev1.Secret, *corev1.ConfigMap, error) {
```
ardaguclu, line 398: Ref: https://github.com/openshift/library-go/pull/2410/changes#r3756239755

I think we don't need to rebuild Secret from KeyState. Current `generateKeySecret` function works nicely https://github.com/openshift/library-go/blob/2bf145c31cad5aea84b2056f709aa616a414597e/pkg/operator/encryption/controllers/key_controller.go#L309.

PR #2424, opened 2026-08-13
add retries when apiserver down in sno parallel run

gangwgr:
In SNO, when kube-apiserver is unavailable, operators can not be accessible. so adding retries when apiserver down in sno parallel run

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when creating Secrets, Routes, and OAuth access tokens by waiting for API services to stabilize.
  * Temporary service-unavailable errors are now treated as retryable during resource creation and migration progress checks.
  * Encryption configuration changes now wait for required cluster operators to become available before continuing.
  * Improved migration progress monitoring by continuing to poll when temporary API errors occur.
  * Improved test behavior across single-node and multi-node environments.

### DIFF
--- a/test/library/encryption/helpers.go
+++ b/test/library/encryption/helpers.go
@@ -102,7 +102,7 @@ func SetAndWaitForEncryptionType(ctx context.Context, t testing.TB, provider Enc
 
 	var previousEncryption configv1.APIServerEncryption
 	var needsUpdate bool
-	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	err = onErrorWithTimeout(waitPollTimeout, retry.DefaultBackoff, orError(errors.IsConflict, transientAPIError), func() error {
 		apiServer, err := clientSet.ApiServerConfig.Get(ctx, "cluster", metav1.GetOptions{})
 		if err != nil {
 			return err
@@ -200,7 +200,7 @@ func WaitForNoNewEncryptionKey(t testing.TB, kubeClient kubernetes.Interface, pr
 	if err := wait.Poll(waitNoKeyPollInterval, waitNoKeyPollTimeout, func() (bool, error) {
 		currentKeyMeta, err := GetLastKeyMeta(t, kubeClient, namespace, labelSelector)
 		if err != nil {
-			return false, err
+			return false, nil
 		}
 
 		if currentKeyMeta.Name != prevKeyMeta.Name {
@@ -241,7 +241,7 @@ func WaitForNextMigratedKey(t testing.TB, kubeClient kubernetes.Interface, prevK
 	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
 		currentKeyMeta, err := GetLastKeyMeta(t, kubeClient, namespace, labelSelector)
 		if err != nil {
-			return false, err
+			return false, nil
 		}
 
 		if currentKeyMeta.Name != observedKeyName {
@@ -483,7 +483,7 @@ func WaitForCurrentKeyMigrated(t testing.TB, kubeClient kubernetes.Interface, pr
 	if err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
 		currentKeyMeta, err := GetLastKeyMeta(t, kubeClient, namespace, labelSelector)
 		if err != nil {
-			return false, err
+			return false, nil
 		}
 		if currentKeyMeta.Name != prevKeyMeta.Name {
 			return false, fmt.Errorf("unexpected key observed %q, expected no new key", currentKeyMeta.Name)


test/library/encryption/helpers.go
```diff
@@ -605,7 +614,12 @@ func CreateAndStoreWellKnownRouteOfLife(ctx context.Context, t testing.TB, cs Cl
 	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(route)
 	require.NoError(t, err)
 
-	created, err := cs.DynamicClient.Resource(wellKnownRouteGVR).Namespace(ns).Create(ctx, &unstructured.Unstructured{Object: obj}, metav1.CreateOptions{})
+	var created *unstructured.Unstructured
+	err = onErrorWithTimeout(createResourceRetryTimeout, retry.DefaultBackoff, transientAPIError, func() error {
```
ardaguclu: This may not be sufficient. We should adjust it based on how long it takes, on average, for apiservers to start serving again.

ardaguclu: or instead of this retry, we can wait until the apiservers signal healthy.  After that we can try to create the resource.

ardaguclu: You just need to query and check the condition and search for success

test/library/encryption/errors.go
```diff
@@ -46,7 +46,7 @@ func transientAPIError(err error) bool {
 	switch {
 	case err == nil:
 		return false
-	case errors.IsServerTimeout(err), errors.IsTooManyRequests(err), net.IsProbableEOF(err), net.IsConnectionReset(err), net.IsNoRoutesError(err), isConnectionRefusedError(err), isGolangNetTimeout(err):
+	case errors.IsServiceUnavailable(err), errors.IsServerTimeout(err), errors.IsTooManyRequests(err), net.IsProbableEOF(err), net.IsConnectionReset(err), net.IsNoRoutesError(err), isConnectionRefusedError(err), isGolangNetTimeout(err):
```
ardaguclu, line 49: Let's open a separate PR for this change. I'll review it and preferably merge it. If this is applicable for all scenarios.

test/library/encryption/perf_helpers.go
```diff
@@ -26,6 +26,10 @@ func watchForMigrationControllerProgressingCondition(t testing.TB, getOperatorCo
 	err := wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
 		conditions, err := getOperatorConditionsFn(t)
 		if err != nil {
+			if transientAPIError(err) {
```
ardaguclu, line 29: Let's open a separate PR for this change. I'll review it and preferably merge it. Is this used by other scenarios as well?

gangwgr: it is for perf cases

ardaguclu: So anywhere else does not use `transientAPIError` function?

gangwgr: helpers.go -> GetLastKeyMeta | Retries secret list on transient errors (503, EOF, timeout, etc.)
helpers.go -> ForceKeyRotation | Retries unsupported-config update on conflict or transient errors
perf_helpers.go .> watchForMigrationControllerProgressingCondition

gangwgr: new pr https://github.com/openshift/library-go/pull/2425

test/library/encryption/helpers.go
ardaguclu: I think we should change this to `onErrorWithTimeout`. So that it will continue retry, if kube-apiserver is down for sometime. 

```go
err = onErrorWithTimeout(waitPollTimeout, retry.DefaultBackoff, orError(errors.IsConflict, transientAPIError), func() error {
```

ardaguclu:
```suggestion
			return false, nil
```

ardaguclu:
```suggestion
			return false, nil
```

ardaguclu:
```suggestion
			return false, nil
```

ardaguclu (commented):
I think, new function `waitForAPIEncryptionClusterOperatorsAvailable` won't fix our problem entirely. I proposed alternative approach. Please let me know your thoughts.

gangwgr:
> I think, new function `waitForAPIEncryptionClusterOperatorsAvailable` won't fix our problem entirely. I proposed alternative approach. Please let me know your thoughts.

Are we adding both approaches, or I can remove mine
because you approach also good we are keep retrying for sometime when KAS down

ardaguclu:
> > I think, new function `waitForAPIEncryptionClusterOperatorsAvailable` won't fix our problem entirely. I proposed alternative approach. Please let me know your thoughts.
> 
> Are we adding both approaches, or I can remove mine because you approach also good we are keep retrying for sometime when KAS down

We can remove this waitForAPIEncryptionClusterOperatorsAvailable

test/library/encryption/helpers.go
ardaguclu: @gangwgr what about this one?

gangwgr:
> > > I think, new function `waitForAPIEncryptionClusterOperatorsAvailable` won't fix our problem entirely. I proposed alternative approach. Please let me know your thoughts.
> > 
> > 
> > Are we adding both approaches, or I can remove mine because you approach also good we are keep retrying for sometime when KAS down
> 
> We can remove this waitForAPIEncryptionClusterOperatorsAvailable

updated

ardaguclu:
That looks good. Thank you. We need to run this with fake bumps to see all the jobs pass.

test/library/encryption/helpers.go
gangwgr: done

gangwgr:
@ardaguclu test passed here https://github.com/openshift/cluster-kube-apiserver-operator/pull/2270
kms job is failing due to monitor case issue not our test failed

ardaguclu:
@gangwgr don't we need to result from SNO KMS run?

gangwgr:
> @gangwgr don't we need to result from SNO KMS run?

job passed - https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-cluster-kube-apiserver-operator-2270-periodics-e2e-aws-encryption-kms-single-node/2089953184444321792

PR #2426, opened 2026-08-14
OCPBUGS-105876: add availableInertia support to StatusSyncer

fgiudici:
### Motivation
clusteroperator/olm briefly flips Available=False (sub-second) during OCP upgrades: catalogd's soft anti-affinity occasionally co-schedules both replicas on one master, and when MCO drains that node both pods fail readiness and get evicted. 
Full scenario explanation at [OCPBUGS-105876](https://redhat.atlassian.net/browse/OCPBUGS-105876?focusedCommentId=18013374).

### Idea
Add availableInertia, mirroring the existing degradedInertia: a new WithAvailableInertia builder lets a consumer opt in to a grace period before a brief Available=False is propagated to the ClusterOperator. Defaults to nil (unchanged instant-propagation behavior) for any consumer that doesn't call it, so this is a no-op for every existing StatusSyncer user.

### Notes
TestAvailableInertia exercises the new behavior through Sync, mirroring TestDegraded's structure (including per-condition-type duration overrides via InertiaCondition).

**Proof PR still missing, will add here in place of this line as soon as done.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status updates by applying configurable inertia to availability conditions.
  * Helps prevent transient availability changes from causing immediate status flips.
  * Supports custom timing for availability conditions while preserving clear status reasons and messages.

### DIFF
--- a/pkg/operator/status/status_controller_test.go
+++ b/pkg/operator/status/status_controller_test.go
@@ -3,12 +3,13 @@ package status
 import (
 	"context"
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"reflect"
 	"regexp"
 	"strings"
 	"testing"
 	"time"
+
+	clocktesting "k8s.io/utils/clock/testing"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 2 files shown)


pkg/operator/status/status_controller_test.go
```diff
@@ -3,13 +3,14 @@ package status
...
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"reflect"
 	"regexp"
 	"strings"
 	"testing"
 	"time"
 
+	clocktesting "k8s.io/utils/clock/testing"
+
```
tchap, line 13: Could we just move this to the k8s import block? Thanks.

pkg/operator/status/status_controller_test.go
```diff
@@ -368,6 +369,157 @@ func TestDegraded(t *testing.T) {
 	}
 }
 
+func TestAvailableInertia(t *testing.T) {
+	fakeClock := clocktesting.NewFakePassiveClock(time.Now())
+	threeMinutesAgo := metav1.NewTime(fakeClock.Now().Add(-3 * time.Minute))
+	fiveSecondsAgo := metav1.NewTime(fakeClock.Now().Add(-2 * time.Second))
+	yesterday := metav1.NewTime(fakeClock.Now().Add(-24 * time.Hour))
```
tchap, line 376: Could we add a comment here explaining which are covered by the inertia? Or perhaps even name them such that it's obvious, like `withinInertiaTime` or something. Would be nice to do that with Type as well, like `customInertiaType = "TypeDAvailable"` or something. I can always scroll down and remember which types are special, but this is just making it easier to understand. Just feed this to an agent and it's done 🙂

pkg/operator/status/status_controller_test.go
```diff
@@ -368,6 +369,157 @@ func TestDegraded(t *testing.T) {
...
+			expectedCondition := &configv1.ClusterOperatorStatusCondition{
+				Type:   configv1.OperatorAvailable,
+				Status: tc.expectedStatus,
+			}
+			if len(tc.expectedMessages) > 0 {
+				expectedCondition.Message = strings.Join(tc.expectedMessages, "\n")
+			}
+			if len(tc.expectedReason) > 0 {
+				expectedCondition.Reason = tc.expectedReason
+			}
```
tchap, line 509: You don't really need these ifs there, it will work just fine without them.

tchap (commented):
Sorry you turned it to a draft while I was reviewing, but I am still gonna post the comments 🙂

pkg/operator/status/status_controller_test.go
```diff
@@ -368,6 +369,157 @@ func TestDegraded(t *testing.T) {
...
+			controller = controller.WithAvailableInertia(MustNewInertia(
+				2*time.Minute,
+				InertiaCondition{
+					ConditionTypeMatcher: regexp.MustCompile("^TypeCAvailable$"),
+					Duration:             5 * time.Minute,
+				},
+				InertiaCondition{
+					ConditionTypeMatcher: regexp.MustCompile("^TypeDAvailable$"),
+					Duration:             time.Minute,
+				},
```
tchap, line 492: One more comment: I think that we only need to test the custom inertia takes effect, we don't need to test the custom inertia implementation, e.g. custom condition matchers, so I would just make this test as simple as possible.

tchap:
I actually wanted to get this merged this week since it seemed I would also needed it, but it's not really the case yet. So I created https://github.com/openshift/library-go/pull/2429 and it's just hanging there. Honestly it looks a bit better from my perspective, so if you need this change and there are no review issues, feel free to tag that PR and get it merged. Otherwise feel free to ask @dusk125 for review as he volunteered to review that PR of mine. If you decide to continue with this PR, feel free to close mine any time.

fgiudici:
@tchap , thanks for the comments.
Fine with me to go with #2429 where you fixed the tests: closing this in favor of #2429 .

PR #2428, opened 2026-08-18
OCPBUGS-38662: Add GuardControllerDegradedConditionType constant

tchap:
Introduce a new condition type to track errors in the guard controller.

This will be then used to set custom Degraded inertia in related operators.

## Summary by CodeRabbit

- **New Features**
  - Added a standardized status condition for indicating guard-controller degradation.
  - The condition covers errors involving guard pods or PodDisruptionBudgets.

### DIFF
--- a/pkg/operator/condition/condition.go
+++ b/pkg/operator/condition/condition.go
@@ -69,4 +69,7 @@ const (
 	// NodeControllerDegradedConditionType is true when the operator observed a master node that is not ready.
 	// Note that a node is not ready when its Condition.NodeReady wasn't set to true
 	NodeControllerDegradedConditionType = "NodeControllerDegraded"
+
+	// GuardControllerDegradedConditionType is true when the guard controller observed an error while managing guard pods or PodDisruptionBudgets.
+	GuardControllerDegradedConditionType = "GuardControllerDegraded"
 )


pkg/operator/condition/condition.go
```diff
@@ -69,4 +69,7 @@ const (
 	// NodeControllerDegradedConditionType is true when the operator observed a master node that is not ready.
 	// Note that a node is not ready when its Condition.NodeReady wasn't set to true
 	NodeControllerDegradedConditionType = "NodeControllerDegraded"
+
+	// GuardControllerDegradedConditionType is true when the guard controller observed an error while managing guard pods or PodDisruptionBudgets.
+	GuardControllerDegradedConditionType = "GuardControllerDegraded"
```
p0lyn0mial, line 74: shouldn't you also change the guard controller to make use of the new const ?

tchap: The condition name is assembled dynamically, so it's not used within library-go. It basically does `GuardController` + `Degraded` 🙂 

The constant is then gonna be used here as other constants already there:
https://github.com/openshift/cluster-kube-controller-manager-operator/blob/8db74e3fe8793043d942ef1f59cce3b0a0bcc548/pkg/operator/starter.go#L318

p0lyn0mial: will the condition be used only by the KCM-O?

tchap: KCM-O, KAS-O and KS-O at least, I would say. They all use the same Degraded inertia for static pod stuff.

PR #2057, opened 2025-11-21
OCPBUGS-23746: Add availableInertia support to prevent transient APIServicesAvailable flapping

sdodson:
## Summary
- Add `availableInertia` field and `WithAvailableInertia()` method to StatusSyncer to support inertia for Available conditions
- Add `WithStatusControllerAPIServicesAvailableInertia()` helper that sets 5-second inertia for APIServicesAvailable conditions
- Apply availableInertia when setting OperatorAvailable condition in StatusSyncer.Sync()

This prevents brief transient errors (like temporary network issues or missing HTTP headers) from causing APIServicesAvailable conditions to flap between Available=True and Available=False. The 5-second inertia allows these transient issues to self-resolve before affecting the operator's Available status.

## Test plan
- [ ] Verify that temporary APIServicesAvailable errors (< 5 seconds) do not cause Available=False
- [ ] Verify that persistent APIServicesAvailable errors (> 5 seconds) correctly set Available=False
- [ ] Run existing unit and integration tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2429, opened 2026-08-18
OCPBUGS-23746: Add availableInertia support to operator status controller

tchap:
The OperatorAvailable condition now respects an inertia window, similar to the existing degradedInertia behavior. This allows operators to suppress transient availability fluctuations within a configurable time window. A new WithAvailableInertia method enables operators to set custom inertia thresholds, and comprehensive tests validate the inertia behavior for various condition scenarios.

This basically replaces https://github.com/openshift/library-go/pull/2426 . I am pushing this now since the other author is on vacation and I would like to proceed.

https://github.com/openshift/library-go/pull/2057 can also be closed once this is merged IMO.

## Summary by CodeRabbit

* **New Features**
  * Added configurable inertia for operator availability status updates.
  * Availability can now remain active briefly after a temporary unavailable condition, reducing status fluctuation.
  * Multiple unavailable conditions are aggregated when determining the final availability status.

### DIFF
--- a/pkg/operator/status/status_controller.go
+++ b/pkg/operator/status/status_controller.go
@@ -52,6 +52,7 @@ type StatusSyncer struct {
 	controllerFactory *factory.Factory
 	recorder          events.Recorder
 	degradedInertia   Inertia
+	availableInertia  Inertia
 
 	removeUnusedVersions bool
 }
@@ -120,6 +121,14 @@ func (c *StatusSyncer) Run(ctx context.Context, workers int) {
 func (c *StatusSyncer) WithDegradedInertia(inertia Inertia) *StatusSyncer {
 	output := *c
 	output.degradedInertia = inertia
+	return &output
+}
+
+// WithAvailableInertia returns a copy of the StatusSyncer with the
+// requested inertia function for available conditions.
+func (c *StatusSyncer) WithAvailableInertia(inertia Inertia) *StatusSyncer {
+	output := *c
+	output.availableInertia = inertia
 	return &output
 }
 
@@ -217,7 +226,7 @@ func (c StatusSyncer) Sync(ctx context.Context, syncCtx factory.SyncContext) err
 
 	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.OperatorDegraded, operatorv1.ConditionFalse, c.degradedInertia, currentDetailedStatus.Conditions...), c.clock)
 	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.OperatorProgressing, operatorv1.ConditionFalse, nil, currentDetailedStatus.Conditions...), c.clock)
-	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.OperatorAvailable, operatorv1.ConditionTrue, nil, currentDetailedStatus.Conditions...), c.clock)
+	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.OperatorAvailable, operatorv1.ConditionTrue, c.availableInertia, currentDetailedStatus.Conditions...), c.clock)
 	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.OperatorUpgradeable, operatorv1.ConditionTrue, nil, currentDetailedStatus.Conditions...), c.clock)
 	configv1helpers.SetStatusCondition(&clusterOperatorObj.Status.Conditions, UnionClusterCondition(configv1.EvaluationConditionsDetected, operatorv1.ConditionFalse, nil, currentDetailedStatus.Conditions...), c.clock)
 
--- a/pkg/operator/status/status_controller_test.go
+++ b/pkg/operator/status/status_controller_test.go
@@ -3,7 +3,6 @@ package status
 import (
 	"context"
 	"fmt"
-	clocktesting "k8s.io/utils/clock/testing"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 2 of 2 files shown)


fgiudici (commented):
/lgtm

pkg/operator/status/status_controller.go
```diff
@@ -123,6 +124,14 @@ func (c *StatusSyncer) WithDegradedInertia(inertia Inertia) *StatusSyncer {
 	return &output
 }
 
+// WithAvailableInertia returns a copy of the StatusSyncer with the
+// requested inertia function for available conditions.
+func (c *StatusSyncer) WithAvailableInertia(inertia Inertia) *StatusSyncer {
+	output := *c
+	output.availableInertia = inertia
+	return &output
+}
```
bertinatto, line 133: Since this was considered and [nacked before](https://github.com/openshift/library-go/pull/632#discussion_r356249018), I think we should talk about what are the implications of this and then make an informed decision. See this comment in the PR where the inertia framework was introduced: https://github.com/openshift/library-go/pull/632#discussion_r356273639. The statement, `If someone thinks up a case, they'll need to argue it out in a library-go PR` from @wking is what need to happen now :)

I understand this is intended to fix the flipping of the Available condition, however, I believe the lack of support for inertia here was intentional. From my understanding, the current rationale is that it's acceptable to delay reporting Degraded because nothing breaks if it's slow to flip. However, delaying moving out from a `Available=false` state can trigger more strong reactions, like blocking upgrades, alerts, etc. The other way around is also problematic: delaying moving out from a `Available=true` is lying about availability (which seems to be a bigger sin than lying about degradation :) ).

I'd like to hear @wking thoughts on this.

bertinatto: Also, if we decide to allow this, would it make sense define a default inertia value for Available, so that operators can't overdo it?

wking: There's a lot of diversity of view around how this kind of thing should be handled, and I've personally given up on trying to convince folks to consolidate around some kind of consistent policy here ;).  Trying to describe the main approaches I've heard:

* Some folks feel that if an component is not available for even a short time (e.g. while a leader lease is being handed off between outgoing and incoming instances of a controller), then the controller is not available during that lease-handoff period, and the ClusterOperator should show `Available=False`.  For these folks, `Available` inertia doesn't seem to be very useful.
* Some folks feel like there are expected, blippy things that happen in the life of a healthy cluster, and even though [`ClusterOperatorDown` cooks `for: 10m` before sounding its `critical` alert][1], someone watching the ClusterOperator directly (e.g. in `oc get clusteroperators ...` or the web-console `/settings/cluster/clusteroperators` page) might see an `Available=False` claim and panic, without realizing that the issue was regular business and expected to resolve itself shortly.  These folks might find `Available` inertia useful, but can also apply their own inertia internally before feeding into library-go helpers to provide as much buffering as they think "regular business" deserves.  For example, retrying a failed Kube API request a few times before setting an initial `Available*` to `False` or `Unknown`.  So while this pull would help folks with this approach to denoise their ClusterOperator claims vs. expected disruption, I don't see this pull being a _requirement_ for that kind of denoising.
* Some folks feel like a highly-available cluster should have no disruption, in which case there's no need to have any inertia _or_ set `Available=False`.  I feel like we do a pretty solid job at zero-disruption in general, but we're certainly not perfect.

[The current OpenShift origin test-policy][2] says that the "someone watching the ClusterOperator directly might see an `Available=False` claim and panic" concerns have weight.  If the folks setting policy decide to soften that approach, or grant exceptions that last forever and which we don't plan on ever fixing, that's up to them (e.g. openshift/origin#30775 adjusted the policy to stop caring what happened on single-node clusters).  If the policies require denoising, both the inertial/retry option (whether via this pull or other tooling) and the "reduce the actual disruption" options are on the table, and I'm fine leaving it up to the impacted component's maintainers to pick the approach that seems most appropriate to them for that particular disruption.  If folks overdo it on the "decide not to bubble up disruption because it might be expected/acceptable" direction, anyone who notices (e.g. a bitten cluster admin) can file a bug reporting that.  And if folks underdo it and pester admins about issues that shortly afterward resolve on their own, anyone who notices can file bugs reporting that too (which is what this pull's [OCPBUGS-23746](https://redhat.atlassian.net/browse/OCPBUGS-23746) is).  So there are forcing functions in both directions to keep the component maintainers honest.

[1]: https://github.com/openshift/cluster-version-operator/blob/97473b9587f3be4d14c33e0bd8bb38ae1dafea6f/install/0000_90_cluster-version-operator_02_servicemonitor.yaml#L97-L106
[2]: https://github.com/openshift/origin/blob/3068afd5a49966e71e6b1e4bb1979844604cdbf7/pkg/monitortests/clusterversionoperator/legacycvomonitortests/operators.go

sdodson: @bertinatto Do you know what amount of inertia would be required to get to a place where there are no Avaialble=False blips in CI runs for HA clusters? I'm imagining that's in the sub 10 second range? If so I can't see delaying alerting an admin by 10 seconds to be problematic.

sdodson: In the bug this links to it looks like 4 seconds is what's currently observed, I would be fine with 10 seconds.

tchap: Yeah, regarding all the blipping jobs, I always only spotted this for a few seconds.

Regarding the overall discussion, I don't think we need to decide whether you can or can't set an Available inertia. This is a library function. IMO the discussion should happen for each operator that wants to set it, unless we want a general 10-second inertia.

This was open to replace https://github.com/openshift/library-go/pull/2426 also, which points to a real use case, I guess, which is not just about silencing things. Last CI job I checked related to Available inertia made me send a PR into the etcd operator. So I don't think adding a general inertial initially is really needed, unless we just don't care about those minor issues, which can turn out to be real issues. But going to investigate your operator and deciding that inertia is the right call, there should be an option for people to enable that, I think.

fgiudici: Yep, I completely agree with tchap: this is a library and why not allow the caller to decide if and how much intertia to have in their own operator?. This PR adds a way to *optionally* add some inertia via the library. The default is nil, so no intertia at all.

sdodson: That sounds fine to me.

PR #2430, opened 2026-08-18
CNTRLPLANE-3237:  preflight simplify socket flags

p0lyn0mial:
## Summary by CodeRabbit

* **New Features**
  * Added KMS plugin preflight checking with support for multiple Unix socket endpoints.
  * Automatically provides configured KMS socket endpoints to the preflight checker.
  * Additional sockets receive reachability checks, while the primary socket undergoes full verification.
  * KMS socket paths now support multiple encryption keys.

* **Bug Fixes**
  * Added validation for missing, malformed, or duplicate socket endpoints.
  * Improved handling of invalid timeouts and missing configuration values.
  * Reports an error when the configured target container is unavailable.

PR #2431, opened 2026-08-20
CNTRLPLANE-3237: test/preflight_deploy.go: fix preflight e2e socket endpoint to match validation pattern

p0lyn0mial:
fixes the socket endpoint to use `kms-1.sock` to match the validation pattern.

this should unblock https://github.com/openshift/cluster-kube-apiserver-operator/pull/2271

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the preflight deployment configuration to use the correct KMS socket endpoint.

### DIFF
--- a/test/library/encryption/preflight_deploy.go
+++ b/test/library/encryption/preflight_deploy.go
@@ -22,7 +22,7 @@ const (
 	preflightDeployConfigHash = "e2e-preflight-drift"
 	// PreflightDeployCallTimeout is the KMS gRPC call timeout used by preflight e2e Deploy.
 	PreflightDeployCallTimeout  = 10 * time.Second
-	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms.sock"
+	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms-1.sock"
 	preflightStatusPollInterval = 5 * time.Second
 	preflightStatusPollTimeout  = 5 * time.Minute
 	openshiftConfigNS           = "openshift-config"


test/library/encryption/preflight_deploy.go
```diff
@@ -22,7 +22,7 @@ const (
 	preflightDeployConfigHash = "e2e-preflight-drift"
 	// PreflightDeployCallTimeout is the KMS gRPC call timeout used by preflight e2e Deploy.
 	PreflightDeployCallTimeout  = 10 * time.Second
-	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms.sock"
+	preflightKMSSocketEndpoint  = "unix:///var/run/kmsplugin/kms-1.sock"
```
ardaguclu, line 25: How doesn't this collide with the uds path generated for key_id: 1?

p0lyn0mial: https://github.com/openshift/cluster-kube-apiserver-operator/blob/c0c59b9e09854161e382491b563e2ab6bab09aa6/vendor/github.com/openshift/library-go/test/library/encryption/preflight_deploy.go#L185

and then

https://github.com/openshift/cluster-kube-apiserver-operator/blob/bef6dbf0ca0906bf25adc470cb8b6df704250547/test/e2e-encryption-kms/encryption_kms_2.go#L137

p0lyn0mial: in the prev pr we added validation (https://github.com/openshift/library-go/pull/2430/changes#diff-8f18967f1605d1c1cf23f0d9e910c90dc8863453820ce827c226284ef65648d7R22)

p0lyn0mial: setting `unix:///var/run/kmsplugin/kms.sock` makes the validation to fail.

p0lyn0mial: But TBH why do we even have an e2e test for the preflight checker? In my opinion, the preflight mechanism should be tested by providing an incorrect configuration to the API and then verifying a degraded condition on the operator CR (set by the preflight controller). This feels more like an e2e approach.

ardaguclu: If preexisting key exists in this format https://github.com/openshift/library-go/blob/0387a1950a9621c42e0c46fba668554fddbc54bb/pkg/operator/encryption/controllers/key_controller.go#L325, new key will directly collide with this in preflight.

p0lyn0mial: the test controls the key, please take a look at https://github.com/openshift/cluster-kube-apiserver-operator/blob/c0c59b9e09854161e382491b563e2ab6bab09aa6/vendor/github.com/openshift/library-go/test/library/encryption/preflight_deploy.go#L176

ardaguclu: Maybe we should delete that test.

p0lyn0mial: I am going to rewrite it. Let's merge the fix which will unblock pulling the library-go to the operator repos.

PR #2433, opened 2026-08-21
NO-JIRA: Extract key id dynamically instead of hardcoding

ardaguclu:
It is better to extract potential key id and run preflight deployer e2e test with that rather than hardcoding. This is just an approximation, correctness of key id is not important anyone. 

## Summary by CodeRabbit

* **New Features**
  * Encryption preflight deployments now automatically select the next available key ID when existing keys are detected.
  * Related Vault, KMS, secret, ConfigMap, socket, and plugin settings are updated consistently for the selected key ID.
* **Bug Fixes**
  * Prevented conflicts caused by always reusing key ID `1` during encryption setup.

### DIFF
--- a/test/library/encryption/preflight_deploy.go
+++ b/test/library/encryption/preflight_deploy.go
@@ -3,6 +3,7 @@ package encryption
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ardaguclu:
Opened this https://github.com/openshift/cluster-kube-apiserver-operator/pull/2274

gangwgr:
lgtm

PR #2435, opened 2026-08-24
NO-JIRA: preflight: Gracefully handle if mode is changed from KMS

ardaguclu:
Preflight controller queries APIServer config CRD to ensure that encryption mode is KMS (to continue the process). However, as discussed [here](https://github.com/openshift/library-go/pull/2410/changes/20d847fa4a1a70808a0519f81738d0d2670c1f33#r3827646703),  encryption computation also queries the APIServer config CRD. Therefore, there is a very small window that cluster admin may change the encryption mode moving away from KMS during that time.

This PR handles this gracefully without degrading by introducing a new sub-state for 2b.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved KMS preflight handling when encryption mode changes away from KMS.
  - Skips deployment cleanly when no encryption configuration is required, without unnecessary retries.
  - Preserves validation of encryption configuration during deployment errors.

- **Documentation**
  - Updated deployment scenarios to describe the non-KMS transition path.

### DIFF
--- a/pkg/operator/encryption/controllers/kms_preflight_controller.go
+++ b/pkg/operator/encryption/controllers/kms_preflight_controller.go
@@ -422,7 +422,10 @@ func (c *kmsPreflightController) sync(ctx context.Context, syncCtx factory.SyncC
 //     2a. Result already recorded as Failed and pod is gone: surface the error
 //     without re-deploying. The admin must fix the config (new hash) before
 //     a new check can run.
-//     2b. No result yet: call Deploy. On success, requeue and wait for the pod to report results.
+//     2b. No result yet: compute the encryption-config Secret, then call Deploy.
+//     If the encryption mode is no longer KMS, skip deployment — next sync
+//     will re-evaluate whether preflight is required.
+//     On success, requeue and wait for the pod to report results.
 //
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -524,6 +528,11 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
 		if err != nil {
 			return true, "", "", fmt.Errorf("failed to compute encryption configuration: %w", err)
 		}
+		if encryptionConfig == nil {
+			// Mode switched away from KMS — nothing to deploy.
```
p0lyn0mial, line 532: could we extend the computer to accept `apiServer.Spec.Encryption.KMS` (and any other dep)?

then the computer is coupled with the controller's view of the world.

ardaguclu: I think, we can. But I'll leave the decision up to @bertinatto

ardaguclu:
We discussed this offline. Probably we will need to make changes in encryption computation to prevent the drift.
/close

PR #2436, opened 2026-08-24
NO-JIRA: Move to internal plugin config derived from openshift/api

ardaguclu:
We are planning to remove image field from KMSPluginConfiguration in openshift/api (ref: https://github.com/openshift/enhancements/pull/2082), since it will be populated by the OLM operators and stored in ConfigMap. 

However, encryption controllers still need to carry this image field along to plugin lifecycle. This PR introduces new encodable/decodable internal API type which is derived from KMSPluginConfiguration but extended with image field. 

## Summary by CodeRabbit

- **New Features**
  - KMS encryption state now preserves both provider settings and the associated plugin image.
  - Vault-based encryption sidecars use the configured KMS plugin image.
  - KMS configuration secrets support the updated internal format for reliable storage and retrieval.

- **Bug Fixes**
  - Improved provider-instance comparison and KMS migration handling by using complete stored configuration data.
  - Expanded validation for Vault settings, authentication, plugin images, and encryption configuration round trips.

### DIFF
--- a/pkg/operator/encryption/controllers/key_controller.go
+++ b/pkg/operator/encryption/controllers/key_controller.go
@@ -325,7 +325,10 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
 				Endpoint:   fmt.Sprintf(kmsEndpointFormat, keyID),
 				Timeout:    &metav1.Duration{Duration: defaultKMSTimeout},
 			},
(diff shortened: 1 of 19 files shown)


ardaguclu:
/hold
this is currently an experimentation

ardaguclu:
This was added just for demonstrating the idea. I think, we'll discuss this more
/close

PR #2438, opened 2026-08-25
OCPBUGS-109578: fix oc darwin client hcp login when insecure tls flag is used

ehearne-redhat:
This fix fixes an issue seen when oc client is used to login to a HCP cluster on macOS. It falls back to kubeconfig CA when a string based x509 error is observed on the macOS platform.

See https://github.com/golang/go/blob/master/src/crypto/x509/root_darwin.go#L74 for reference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS authentication reliability by falling back to the configured certificate authority when system certificate validation encounters an unexpected error.

### DIFF
--- a/pkg/oauth/tokenrequest/request_token.go
+++ b/pkg/oauth/tokenrequest/request_token.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"runtime"
 	"slices"
 	"strings"
 
@@ -602,8 +603,23 @@ func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509
 		intermediates.AddCert(c)
 	}
 
-	return chain[0].Verify(x509.VerifyOptions{
+	certChainList, err := chain[0].Verify(x509.VerifyOptions{
 		Intermediates: intermediates,
 		DNSName:       dnsName,
 	})
+
+	if runtime.GOOS == "darwin" && strings.HasPrefix(err.Error(), "x509:") {
+		// this check fills in the gap where root_darwin.go has insufficient
+		// typed errors for macOS platform, leading to a generic string based
+		// x509 error. This will return a proper x509 error which can be used
+		// to return kubeconfig CA client. x509.UnknownAuthorityError was
+		// chosen because this is the fall back also used on Linux/Windows,
+		// keeping the approach consistent across platforms.
+		// https://github.com/golang/go/blob/5a6340ff28c87e099f33c941e3d73e50d715ddf7/src/crypto/x509/root_darwin.go#L74
+		return nil, x509.UnknownAuthorityError{
+			Cert: chain[0],
+		}
+	}
+
+	return certChainList, err
 }


pkg/oauth/tokenrequest/request_token.go
```diff
@@ -584,6 +585,13 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 			klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
 			return restclient.TransportFor(clientConfig)
 		}
+		if runtime.GOOS == "darwin" && strings.HasPrefix(err.Error(), "x509:") {
```
ardaguclu: There is a precedent discussion here https://github.com/openshift/oc/pull/1207. In my opinion, we should avoid adding workarounds in this file in library-go.

ardaguclu: Actually I was wrong. This code was in oc at the time we discussed that PR. However, it has been moved in library-go. Thus, change should be here.

ricardomaraschini (commented):
This change matches what has been described in the function doc: 

```
// transportWithSystemRoots tries to retrieve the serving certificate from the
// issuer, validates it against the system roots and if the validation passes,
// returns transport using just system roots, otherwise it returns a transport
// that uses the CA from kubeconfig
```

/lgtm

pkg/oauth/tokenrequest/request_token.go
ibihim, line 591: Nit:

We could flatten the whole structure.

```suggestion
	switch err.(type) {
	case nil:
		// copy the config so we can freely mutate it
		configWithSystemRoots := restclient.CopyConfig(clientConfig)

		// explicitly unset CA cert information
		// this will make the transport use the system roots or OS specific verification
		// this is required to have reasonable behavior on windows (cannot get system roots)
		// in general there is no good with to say "I want system roots plus this CA bundle"
		// so we just try system roots first before using the kubeconfig CA bundle
		configWithSystemRoots.CAFile = ""
		configWithSystemRoots.CAData = nil

		// no error meaning the system roots work with the OAuth server
		klog.V(4).Info("using system roots as no error was encountered")
		systemRootsRT, err := restclient.TransportFor(configWithSystemRoots)
		if err != nil {
			return nil, err
		}
		return systemRootsRT, nil
	case x509.UnknownAuthorityError, x509.HostnameError, x509.CertificateInvalidError, x509.SystemRootsError,
		tls.RecordHeaderError, *net.OpError:
		// fallback to the CA in the kubeconfig since the system roots did not work
		// we are very broad on the errors here to avoid failing when we should fallback
		klog.V(4).Infof("falling back to kubeconfig CA due to possible x509 error: %v", err)
		return restclient.TransportFor(clientConfig)
	}

	switch err {
	case io.EOF, io.ErrUnexpectedEOF, io.ErrNoProgress:
		// also fallback on various io errors
		klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
		return restclient.TransportFor(clientConfig)
	}

	if runtime.GOOS == "darwin" && strings.HasPrefix(err.Error(), "x509:") {
		// also fallback on string based x509 error on darwin systems
		// this mirrors the Linux x509.UnknownAuthorityError path
		// https://github.com/golang/go/blob/master/src/crypto/x509/root_darwin.go#L74
		klog.V(4).Infof("falling back to kubeconfig CA due to possible darwin generic x509 error: %v", err)
		return restclient.TransportFor(clientConfig)
	}

	// unknown error, fail (ideally should never occur)
	klog.V(4).Infof("unexpected error during system roots probe: %v", err)
	return nil, err
}

```

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -584,6 +585,13 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
...
+			// also fallback on string based x509 error on darwin systems
+			// this mirrors the Linux x509.UnknownAuthorityError path
+			// https://github.com/golang/go/blob/master/src/crypto/x509/root_darwin.go#L74
```
ibihim: I really like the reference to the Go codebase, but we should not anchor it to the master branch as the line number doesn't age well then. It should be a permalink anchored on a commit SHA.

And I think the comment is a little confusing. `x509.UnknownAuthorityError` is handled above. This branch specifically catches unmapped Apple codes that come back untyped.

pkg/oauth/tokenrequest/request_token.go
ibihim: If we then move the err cases into its own function, we could write unit tests against it. Not a requirement for this PR.

ardaguclu:
We had finalized this issue after this comment https://github.com/openshift/oc/pull/1207#issuecomment-1428032909. What has changed that causes the revival of the bug?. If there is a typed error we can use, why can't we use it?

ehearne-redhat:
Hey @ardaguclu - so this is where things lie currently. 

TLDR: The error code checks in `root_darwin.go` doesn't provide sufficient typed errors for us to use. Standalone vs HCP operate differently on the oauth cert chain level. Feel free to read on if you would like a more concrete answer. :) 

At its core, the main reason this change is needed is the difference in standalone OpenShift versus hypershift. 

Logging in with insecure flag works fine in standalone OpenShift because the OAuth cert chain is managed by service-ca and ingress operators, and is therefore trusted. 

In HyperShift, the cert-ca for OAuth is self-signed by root-ca locally, and has no backing from service-ca and ingress operators as they don't exist, making it untrusted by the system. This gives us the string-based x509 error seen when trying to log in to HCP with insecure flag on macOS. 

The mentioned issue that is believed to resolve the issue, does not. 

https://github.com/golang/go/commit/c8104a1b90d3b09ed6ec080e8ecba98c8805c137#diff-bd06d21bb2959af9c46b4ba18865be1d48c1c9de3240687eee045f9e9c324c58R66-R68

The actual code for checking the macos errors contains just three out of several error codes that could lead to the untrusted error message seen. 

https://github.com/golang/go/blob/master/src/crypto/x509/internal/macos/security.go#L23-L30

Because of this, we are seeing the check fail, and resolve to string based x509 error where this change is justified as getting a change in to the go library would likely take a long time to do. This change resolves a major bottleneck in that the error code checks in `root_darwin.go` don't cover enough scope for an appropriate error type to be returned.

Apple does actually post all of its error codes online in this regard. After doing a good bit of research, I think the error code could be -66996, but I have not tested this. There are of course several candidates. :)

https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/header_symlinks/Security/CSCommon.h#L142

Should you wish to take a look at the other extensive list where the three error codes were chosen for `root_darwin.go` please take a look. 

https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/header_symlinks/Security/SecBase.h#L394

Please let me know if this helps with the question at hand. :)

ardaguclu:
Would it be better to handle this error here https://github.com/ehearne-redhat/library-go/blob/9cc4cf448a81c04f00f1b36be0458a63a9e5948a/pkg/oauth/tokenrequest/request_token.go#L554 for darwin. If error begins with "x509:", we can convert it to `x509.UnknownAuthorityError` (or something like that). So that switch https://github.com/ehearne-redhat/library-go/blob/9cc4cf448a81c04f00f1b36be0458a63a9e5948a/pkg/oauth/tokenrequest/request_token.go#L555 will take care accordingly. Does this make sense?

ehearne-redhat:
It does make sense to me - I am just making the required changes now and suggested changes from ibihim. :)

pkg/oauth/tokenrequest/request_token.go
ehearne-redhat: OK - understood. If we moved the switch case into its own function, it can specifically be unit tested. I do think this might slightly be out of scope for this PR, as Arda actually [made a suggestion](https://github.com/openshift/library-go/pull/2438#issuecomment-5494344317) which moves the onus for the check into a different function - `verifyServerCertChain()` .

ehearne-redhat:
So that should be done now - tested oc login on hcp using macos and problem is resolved even after moving the logic to `verifyServerCertChain()` . 

Also updated the comment/commit message too. 

@ardaguclu @ibihim PTAL when you get a chance. If the switch case refactoring is desired please let me know and I will look to do that. :)

ardaguclu:
/lgtm
/cc @bertinatto 
for approval

ardaguclu:
@ehearne-redhat once this PR merges, would you mind opening a library-go bump PR in oc?

ehearne-redhat:
@ardaguclu Yes absolutely - will open a PR in oc for library-go bump. :)

bertinatto (commented):
/approve

PR #2439, opened 2026-08-26
CNTRLPLANE-3237: KMS preflight positive assertion

p0lyn0mial:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved KMS encryption validation to confirm successful, non-degraded preflight checks with matching configuration and current key information.
  - Strengthened handling of existing Kubernetes resources and transient creation conflicts during encryption setup.
  - Updated encryption operations to use consistent default retry and timeout behavior.

- **Tests**
  - End-to-end encryption tests now use the real encryption configuration processing and verify that KMS preflight completes successfully after setup.

### DIFF
--- a/test/library/encryption/scenarios.go
+++ b/test/library/encryption/scenarios.go
@@ -75,9 +75,14 @@ func TestEncryptionTypeAESGCM(ctx context.Context, t testing.TB, scenario BasicS
 func TestEncryptionTypeKMS(ctx context.Context, t testing.TB, scenario BasicScenario, providers ...EncryptionProvider) {
 	provider := resolveProvider(t, configv1.EncryptionTypeKMS, providers)
 	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	// Snapshot preflight before applying the new config so the assertion can confirm a fresh
+	// preflight ran for it (the remote key id advances when the config genuinely changes).
+	previousPreflight, err := ReadKMSPreflightForOperator(ctx, e, GetClients(e), scenario.OperatorNamespace)
+	require.NoError(e, err)
 	clientSet := SetAndWaitForEncryptionType(ctx, e, provider, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
 	scenario.AssertFunc(e, clientSet, provider.Type, scenario.Namespace, scenario.LabelSelector)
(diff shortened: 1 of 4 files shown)


test/library/encryption/scenarios.go
```diff
@@ -75,9 +75,14 @@ func TestEncryptionTypeAESGCM(ctx context.Context, t testing.TB, scenario BasicS
...
 	provider := resolveProvider(t, configv1.EncryptionTypeKMS, providers)
 	e := NewE(t, PrintEventsOnFailure(scenario.OperatorNamespace))
+	// Snapshot preflight before applying the new config so the assertion can confirm a fresh
+	// preflight ran for it (the remote key id advances when the config genuinely changes).
+	previousPreflight, err := ReadKMSPreflightForOperator(ctx, e, GetClients(e), scenario.OperatorNamespace)
+	require.NoError(e, err)
 	clientSet := SetAndWaitForEncryptionType(ctx, e, provider, scenario.TargetGRs, scenario.Namespace, scenario.LabelSelector)
 	scenario.AssertFunc(e, clientSet, provider.Type, scenario.Namespace, scenario.LabelSelector)
 	AssertEncryptionConfig(e, clientSet, scenario.EncryptionConfigSecretName, scenario.EncryptionConfigSecretNamespace, scenario.TargetGRs)
+	AssertKMSPreflightSucceededForOperator(ctx, e, clientSet, scenario.OperatorNamespace, previousPreflight)
```
ardaguclu, line 85: This is probably not important but wouldn't it be better to assert preflight before the AssertEncryptionConfig?

ardaguclu (commented):
I have one non-blocker comment. Other than that changes look good to me.

test/library/encryption/scenarios.go
p0lyn0mial: we need to make sure the preflight ran after setting the encryption mode/cfg which happens in the SetAndWaitForEncryptionType function. does it make sense ?

ardaguclu: That is correct. But conceptually this should be in this order;

```go
AssertKMSPreflightSucceededForOperator(ctx, e, clientSet, scenario.OperatorNamespace, previousPreflight)
AssertEncryptionConfig(e, clientSet, scenario.EncryptionConfigSecretName, scenario.EncryptionConfigSecretNamespace, scenario.TargetGRs)
```

p0lyn0mial: I don't think it matters but I can change the order.

ardaguclu: No, I agree. It is not important.

PR #2440, opened 2026-08-26
CNTRLPLANE-3237: controllers/kms_preflight_controller: return KMSPluginConfig from preflightRequired

p0lyn0mial:
the cfg can be passed to the EC in a new PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved KMS preflight handling by consistently identifying and returning the active encryption configuration.
  * Ensured no configuration is reported when preflight checks exit early or no matching configuration is found.


bertinatto (commented):
This makes sense to me. The returned KMSPluginConfig can be used to avoid re-fetching the `apiserver` object in `EncryptionPlanner.Load()` (which could incur a race between hash verification and config creation).

/lgtm

PR #2448, opened 2026-08-31
NO-JIRA: add KMS rotation key convergence helper

tjungblu:
simple loop to detect whether all remote key ids are the same non-empty string.

## Summary by CodeRabbit

- **New Features**
  - Added KMS health convergence reporting to identify whether plugin reports agree on a shared remote key ID.
  - Handles missing, empty, and conflicting key IDs with clear convergence results.

- **Tests**
  - Added coverage for empty, single-report, unanimous, conflicting, and mixed key ID scenarios.

### DIFF
--- a/pkg/operator/encryption/kms/health/convergence.go
+++ b/pkg/operator/encryption/kms/health/convergence.go
@@ -1,0 +1,39 @@
+package health
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// ConvergenceResult reports whether a set of KMS plugin health reports agree on
+// the same remote key ID.
+type ConvergenceResult struct {
+	Converged   bool
+	RemoteKeyID string
+}
+
+// ConvergedRemoteKeyID returns whether all reports in the slice report the same
+// non-empty RemoteKeyID.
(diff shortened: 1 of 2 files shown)


pkg/operator/encryption/kms/health/convergence.go
```diff
@@ -0,0 +1,39 @@
...
+// ConvergenceResult reports whether a set of KMS plugin health reports agree on
+// the same remote key ID.
+type ConvergenceResult struct {
+	Converged   bool
+	RemoteKeyID string
+}
+
+// ConvergedRemoteKeyID returns whether all reports in the slice report the same
+// non-empty RemoteKeyID.
+func ConvergedRemoteKeyID(reports []operatorv1.KMSPluginHealthReport) ConvergenceResult {
```
ardaguclu, line 16: I think, this functionality should filter out based on the given keyID. Otherwise, how do we know that for which key id we are checking the convergence, if there are multiple kms plugins configured.

tjungblu: good call, will adjust

ardaguclu:
Lastly, it would be awesome if there is a PoC PR demonstrating the overall goal. So that it would be easier to understand and review these smaller PRs.

tjungblu:
/close

We'll spin this out of the e2e PR.

PR #2449, opened 2026-08-31
CNTRLPLANE-3237: kms preflight deployer report hash

p0lyn0mial:
## Summary by CodeRabbit

- **Changes**
  - KMS encryption preflight now determines deployment configuration directly from the KMS preflight result.
  - Preflight completion tracking now relies on the result condition rather than a pod configuration-hash condition.
  - Configuration hashes remain available through pod metadata annotations; they are no longer passed to the preflight container as command-line or environment settings.
  - Removed the legacy encryption-configuration fallback and always-successful preflight behavior.
  - Updated encryption integration scenarios and status reporting accordingly.


p0lyn0mial:
/hold

until the test pr in kas-o passes.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -588,15 +594,28 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
...
 
-	// Scenario 3a: pod crashed. Keep for inspection; the admin will update
-	// the config which triggers a new hash and cleanup via scenario 3c.
+	// Scenario 3a: stale pod from a previous config. The deployed hash is read
+	// from the pod itself (its annotation), so a mismatch is detected regardless
+	// of the pod phase or whether it ever ran (ImagePullBackOff, Pending, or a
+	// crash before it could report a condition). Reap it; the next sync deploys a
+	// fresh pod. This is what lets a failed pod from an old config make way for
+	// the corrected one, so it is checked before the phase/condition scenarios.
+	if deployedHash != requiredHash {
```
ardaguclu, line 520: can deployedHash be empty or is its emptiness important?

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -111,14 +117,17 @@ func (d *PodPreflightDeployer) Deploy(ctx context.Context, configHash string, en
...
+func (d *PodPreflightDeployer) Status(ctx context.Context) (string, corev1.PodStatus, error) {
 	// preflight status checks are not very frequent, so we use the live client instead of a cached lister
 	pod, err := d.coreClient.Pods(d.namespace).Get(ctx, PodName, metav1.GetOptions{})
 	if err != nil {
-		return corev1.PodStatus{}, fmt.Errorf("failed to get pod for preflight %s/%s: %w", d.namespace, PodName, err)
+		return "", corev1.PodStatus{}, fmt.Errorf("failed to get pod for preflight %s/%s: %w", d.namespace, PodName, err)
 	}
 
-	return pod.Status, nil
+	return pod.Annotations[ConfigHashAnnotation], pod.Status, nil
```
ardaguclu: Shouldn't we check the existence of this annotation and cleanup if it is absent?

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: a good question. i think a malicious user could remove the annotation.
however in that case the condition will be true and the controller will cleanup the existing pod, right?

ardaguclu: So due to the mismatch, pod will be cleaned up. Key controller triggers preflight again to get a signal without creating any new key.

pkg/operator/encryption/kms/preflight/deployer.go
ardaguclu: It is better to check the emptiness of annotations and the existence of this annotation.

pkg/operator/encryption/controllers/kms_preflight_controller.go
p0lyn0mial: the key controller doesn't have to trigger the preflight. the key controller triggers the preflight only when the config changes. 

please describe the issue/situation you have in mind. it would be easier for me to explain.

ardaguclu: I'm trying to understand the expectation from cluster admin, if he/she accidentally deletes the annotation from the pod. 

Key controller expects success from the preflight status. However, preflight cleans up that pod. I assume that preflight will create new pod with the correct annotation in next reconciliation loop. So there is no intervention needed from cluster admin?

p0lyn0mial: it also depends when the annotation will be removed - please specify.

p0lyn0mial: if the annotation is removed before the preflight is reported then the current pod will be removed and a new pod will be created by the controller.

p0lyn0mial: if the annotation is removed after the preflight was reported but before the preflight removes the pod nothing happens.

p0lyn0mial: the annotation cannot be removed after the preflight was reported and the controller removed the pod.

ardaguclu: That makes sense. Thanks

pkg/operator/encryption/kms/preflight/deployer.go
ardaguclu: We may not need to check the existence of annotation. But to prevent any panic, it is better to check the emptiness of Annotations.

p0lyn0mial: in go reading a key that is not present in a map is safe and gives "".

returning "" from the status function will match the condition because `"" != requiredHash` which will cause the controller to remove the pod and create a new one.

ardaguclu: You are right. Sorry for the noise :)

ardaguclu:
This PR looks good to me. But I think it is better to ask opinions from @tjungblu and @bertinatto

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: np, thanks for taking the time to review the pr.

pkg/operator/encryption/kms/preflight/deployer.go
```diff
@@ -26,6 +26,12 @@ const (
 	// EncryptionConfigSecretName is the secret Deploy creates for the preflight pod.
 	EncryptionConfigSecretName = "kms-preflight-encryption-config"
 
+	// ConfigHashAnnotation records, on the preflight pod, the config hash it was
+	// deployed for. It is set at render time (see assets/kms-preflight-pod.yaml)
+	// and read back by Status so the controller can detect a stale pod even when
+	// the pod never ran and therefore never reported any status condition.
+	ConfigHashAnnotation = "encryption.apiserver.operator.openshift.io/kms-preflight-config-hash"
```
tjungblu: would you mind to update the template to inject this constant? otherwise we have no means to detect the drift after changing this.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
```diff
@@ -33,16 +33,13 @@ spec:
       command: [{{.Command}}]
       args:
         - --kms-call-timeout={{.KMSCallTimeout}}
-        - --config-hash=$(CONFIG_HASH)
```
tjungblu, line 36: this PR is quite large already, do you mind to do the removal in a separate PR? I think that will remove a lot of the diff

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -588,35 +581,30 @@ func (c *kmsPreflightController) runPreflightChecks(ctx context.Context) (requeu
...
+	if podStatus.Phase == corev1.PodFailed {
+		pe := podFailureError(podStatus)
+		pe.message = fmt.Sprintf("preflight pod failed for hash %s: %s", requiredHash, pe.message)
+		return false, "", "", pe
+	}
+
+	// Scenario 3c: pod has not reported its result yet. The checker writes its
+	// result condition once the check completes, so its absence means the check
+	// is still running (or the pod is stuck before it could run).
 	resultCondition := FindPodCondition(podStatus.Conditions, KMSPreflightResultPodCondition)
```
tjungblu, line 538: I wonder if this case makes sense with what we discovered? 

How will we be able to detect 
> podStartupTimeoutError

when the pod got lost?

p0lyn0mial: when the pod is lost / deleted then a new one will be created.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
p0lyn0mial: sure, I can split into smaller ones.

pkg/operator/encryption/kms/preflight/deployer.go
p0lyn0mial: sure, i can add that.

pkg/operator/encryption/kms/preflight/assets/kms-preflight-pod.yaml
p0lyn0mial: opened https://github.com/openshift/library-go/pull/2451

tjungblu: thx

p0lyn0mial:
I rebased the PR on the master branch and now the PR contains only changes that remove code. PTAL.

ardaguclu:
/lgtm
/approve
/hold
for @tjungblu looks at

p0lyn0mial:
/hold

I will put a hold until https://github.com/openshift/cluster-kube-apiserver-operator/pull/2287 gets green.

thanks for the reviews.

PR #2452, opened 2026-09-01
NO-JIRA: KMS Rotation End2End

tjungblu:
Enhancement: https://github.com/openshift/enhancements/pull/2041
E2E PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2291

## Summary by CodeRabbit

- **New Features**
  - Added coordinated remote KMS key rotation with convergence tracking and guarded promotion of stable keys.
  - Encryption state now preserves remote-key rotation metadata.
  - Key creation is deferred while remote-key migration is in progress.

- **Bug Fixes**
  - Annotation updates now preserve concurrent changes and unrelated metadata.
  - Improved handling of unavailable, mismatched, or incomplete KMS key reports.

- **Tests**
  - Expanded coverage for rotation, migration delays, convergence detection, and annotation updates.
  - Updated rotation checks to verify annotation convergence.


pkg/operator/encryption/controllers/remote_key_reconciler.go
```diff
@@ -0,0 +1,180 @@
...
+) (time.Duration, error) {
+	if !convergence.Converged {
+		changed, err := clearRemoteKeyConvergence(ctx, secretClient, secretName, rk)
+		return 0, errIfChanged(changed, err)
+	}
+
+	if convergence.RemoteKeyID == rk.TargetRemoteKeyID {
+		changed, err := clearRemoteKeyConvergence(ctx, secretClient, secretName, rk)
+		return 0, errIfChanged(changed, err)
+	}
```
tjungblu, line 138: this can be reworked a bit by filtering for the target remote key up front

tjungblu:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/2291

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -220,10 +236,41 @@ func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext fact
...
+
+	encryptionStatus, err := c.encryptionStatusProvider.GetKMSEncryptionStatus(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get KMS encryption status for remote key rotation: %w", err)
+	}
+
+	return reconcileRemoteKeyRotation(ctx, c.secretClient, c.instanceName, snap.State.EncryptedGRs, snap.State.DesiredBeforePlan, encryptionStatus, systemRemoteKeyClock{})
+}
+
+func (c *keyController) checkAndCreateKeys(ctx context.Context, syncContext factory.SyncContext, encryptedGRs []schema.GroupResource) error {
```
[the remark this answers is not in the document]
tjungblu: removed

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -216,25 +216,40 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
...
-		finished, result, when, err := c.migrator.EnsureMigration(gr, grActualKeys.WriteKey.Key.Name)
+		finished, result, when, err := c.migrator.EnsureMigration(gr, migrationWriteKey)
 		if err == nil && finished && result != nil && time.Since(when) > migrationRetryDuration {
 			// last migration error is far enough ago. Prune and retry.
 			if err := c.migrator.PruneMigration(gr); err != nil {
 				errs = append(errs, err)
 				continue
 			}
-			finished, result, when, err = c.migrator.EnsureMigration(gr, grActualKeys.WriteKey.Key.Name)
+			finished, result, when, err = c.migrator.EnsureMigration(gr, migrationWriteKey)
```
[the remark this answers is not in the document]
tjungblu: replaced

test/library/encryption/helpers.go
```diff
@@ -90,6 +92,46 @@ func WaitForNextEncryptionKeyRotation() WaitForRotationCompleteFunc {
...
+		require.NoError(t, err)
+		before, err := secrets.ReadRemoteKeyAnnotations(secret)
+		require.NoError(t, err)
+		previousMigrated := before.MigratedRemoteKeyID
+
+		t.Logf("Waiting for KMS remote key rotation to complete on secret %q (previous migrated-remote-key-id=%q)", prevKeyMeta.Name, previousMigrated)
+		err = wait.Poll(waitPollInterval, waitPollTimeout, func() (bool, error) {
+			secret, err := clientSet.Kube.CoreV1().Secrets(scenario.Namespace).Get(context.Background(), prevKeyMeta.Name, metav1.GetOptions{})
+			if err != nil {
+				return false, nil
```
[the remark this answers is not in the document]
tjungblu: applied

pkg/operator/encryption/state/types.go
```diff
@@ -50,6 +50,16 @@ type KeyState struct {
...
+	// RemoteKey tracks KMS remote key rotation annotations on the backing secret.
+	RemoteKey RemoteKeyState
+}
+
+// RemoteKeyState is the in-memory view of remote key rotation annotations.
+type RemoteKeyState struct {
+	TargetRemoteKeyID   string
+	MigratedRemoteKeyID string
+	ConvergedAt         time.Time
+	ConvergedID         string
```
bertinatto: would it make sense to move this inside `KMSState`?

pkg/operator/encryption/state/types.go
```diff
@@ -50,6 +50,16 @@ type KeyState struct {
 	ExternalReason string
 	// stores all the KMS encryption mode related configurations
 	KMS *KMSState
+	// RemoteKey tracks KMS remote key rotation annotations on the backing secret.
+	RemoteKey RemoteKeyState
```
bertinatto, line 54: every key (non-kms) would have a `RemoveKey`, would it make sense to move this inside `KMSState`?

pkg/operator/encryption/controllers/remote_key_reconciler.go
```diff
@@ -0,0 +1,204 @@
...
+		rk.ConvergedID = ""
+		rk.ConvergedAt = time.Time{}
+		return true, nil
+	})
+	return true, err
+}
+
+func errIfChanged(changed bool, err error) error {
+	return err
+}
```
bertinatto, line 204: is this a leftover?

pkg/operator/encryption/state/types.go
tjungblu: yeah good suggestion, that makes sense

pkg/operator/encryption/controllers/remote_key_reconciler.go
tjungblu: removed

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -216,26 +216,50 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
...
+		if remoteKeyMigration {
+			targetRemoteKeyID = remoteKeyAnnotations.TargetRemoteKeyID
+			writeKeyGRs = append(writeKeyGRs, gr)
+		}
+
+		if !remoteKeyMigration {
+			if alreadyMigrated, _, _ := state.MigratedFor([]schema.GroupResource{gr}, grActualKeys.WriteKey); alreadyMigrated {
+				continue
+			}
+		}
```
tjungblu, line 250: I'm open for a more elegant solution here, we need to get the target key that has triggered the migration

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -283,6 +329,40 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
 		}
 	}
 
+	if hadRemoteKeyMigration && len(writeKeySecretName) > 0 && len(writeKeyGRs) > 0 {
```
tjungblu: this sucks

pkg/operator/encryption/controllers/migrators/kubestorageversionmigrator.go
```diff
@@ -99,6 +99,26 @@ func (m *KubeStorageVersionMigrator) EnsureMigration(gr schema.GroupResource, wr
...
+func (m *KubeStorageVersionMigrator) MigrationSucceeded(gr schema.GroupResource, writeKey string) (bool, error) {
+	name := migrationResourceName(gr)
+	migration, err := m.informer.StorageVersionMigrations().Lister().Get(name)
+	if errors.IsNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	if migration.Annotations[writeKeyAnnotationKey] != writeKey {
```
tjungblu: there's still a mismatch here with the writeKey not containing the remote key id

ardaguclu:
This rotation topic introduces some intricacies into the state machine, and it is inherently difficult to implement and review (without introducing any non-obvious regressions). I'd propose following this flow;

* EP has merged, which is great news. However, since it would be better for all of us to collaborate on this topic, I think we may need to wait inputs from @bertinatto @p0lyn0mial @ibihim by reading the EP.
* Regarding the implementation, we could adopt a mechanism similar to the one @bertinatto used in https://github.com/openshift/library-go/pull/2410. He organized the changes into logically separated commits per each controller. During each review round, I reviewed one commit at a time, and we moved on to the next commit once we are done. This allowed us to iterate step by step.

That said, rotation is more difficult than the encryption computation refactoring. So we all may need to agree on the design.

tjungblu:
@ardaguclu I think the actual change to the state machine is pretty small. I also don't think some of the pieces in keycontroller/migration controller are good/salvagable in this PR and they are more complicated than they need to be. 

If we can get the low-risk data structure and helper stuff out of the way in:
* https://github.com/openshift/library-go/pull/2461
* https://github.com/openshift/library-go/pull/2460

That cuts the diff already by almost 2/3 and we can do a controller-wise split easily and iterate from there.

pkg/operator/encryption/secrets/types.go
```diff
@@ -67,6 +67,15 @@ const (
...
 	encryptionSecretKMSConfigMapDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-configmap-"
+
+	// EncryptionSecretTargetRemoteKeyID is the target remote KMS key ID to migrate toward.
+	EncryptionSecretTargetRemoteKeyID = "encryption.apiserver.operator.openshift.io/target-remote-key-id"
+	// EncryptionSecretMigratedRemoteKeyID is the last fully migrated remote KMS key ID.
+	EncryptionSecretMigratedRemoteKeyID = "encryption.apiserver.operator.openshift.io/migrated-remote-key-id"
+	// EncryptionSecretRemoteKeyConvergedAt records when a candidate remote key ID first achieved cluster convergence.
+	EncryptionSecretRemoteKeyConvergedAt = "encryption.apiserver.operator.openshift.io/remote-key-converged-at"
+	// EncryptionSecretRemoteKeyConvergedID is the candidate remote key ID the converged-at timestamp belongs to.
+	EncryptionSecretRemoteKeyConvergedID = "encryption.apiserver.operator.openshift.io/remote-key-converged-id"
```
bertinatto: asking here just for my understanding (not trying to change the design): this sounds like a small state machine embedded in the key secret. I was wondering how the cluster admin is expected to know the state of the migration. For instance, let's say the key rotated remotely and the admin wants to know if the system is done with the migration; is that possible without checking the annotations?

tjungblu: sure, no worries. It's indeed a mini state-machine mechanism we introduce here, with two new pieces:
* the 5 minute grace 
* the target != current

The 5 minute grace is not observed in this PR from the operator status, we need to add this. I had a bit of code to do this, but it blew up this diff by like 300 lines so I dropped it as it is not functionally important right now.

The target!=current is tracked with the current progressing states of the existing state machine, so you get feedback through there.

pkg/operator/encryption/controllers/migration_controller.go
```diff
@@ -283,6 +313,23 @@ func (c *migrationController) migrateKeysIfNeededAndRevisionStable(ctx context.C
 		}
 	}
 
+	if hadRemoteKeyMigration && len(writeKeySecretName) > 0 && len(remoteKeyMigratedGRs) == len(writeKeyGRs) {
```
tjungblu, line 316: // TODO(thomas): this probably should be reconciled in the above loop, so we can keep track of the migrated GRs in our own annotation

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -451,33 +483,33 @@ func (p *prefetchedKMSConfigHasherResourceProvider) getConfigMap(_ context.Conte
 //     back off.
 //
 // Callers are responsible for requeuing when this returns (false, nil).
-func (c *keyController) ensureKMSPreflightPassed(ctx context.Context, configHash string) (bool, error) {
+func (c *keyController) ensureKMSPreflightPassed(ctx context.Context, configHash string) (bool, string, error) {
```
ardaguclu, line 486: This can be a separate PR.

tjungblu: will create once #2466 merges

tjungblu:
three continuous bumps for this PR:
https://github.com/openshift/cluster-kube-apiserver-operator/pull/2291 <- including the test
https://github.com/openshift/cluster-authentication-operator/pull/1012
https://github.com/openshift/cluster-openshift-apiserver-operator/pull/769

PR #2454, opened 2026-09-02
NO-JIRA: add vault namespace to rotation helpers

tjungblu:
The vault enterprise NS was missing, otherwise we would receive a 404 on the response.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vault Enterprise compatibility for key rotation and key-version retrieval by targeting the correct administrative namespace.

### DIFF
--- a/test/library/encryption/kms/vault.go
+++ b/test/library/encryption/kms/vault.go
@@ -213,7 +213,7 @@ func rotateKey(ctx context.Context, t testing.TB) {
 	// Command: vault write -f transit/keys/<key-name>/rotate
 	// Reference: https://developer.hashicorp.com/vault/api-docs/secret/transit#rotate-key
 	cmd := exec.CommandContext(commandCtx, "oc", "exec", defaultVaultPodName, "-n", defaultVaultNamespace, "--",
-		"vault", "write", "-f", fmt.Sprintf("%s/rotate", defaultVaultKeyPath))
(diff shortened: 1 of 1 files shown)


test/library/encryption/kms/vault.go
```diff
@@ -213,7 +213,7 @@ func rotateKey(ctx context.Context, t testing.TB) {
 	// Command: vault write -f transit/keys/<key-name>/rotate
 	// Reference: https://developer.hashicorp.com/vault/api-docs/secret/transit#rotate-key
 	cmd := exec.CommandContext(commandCtx, "oc", "exec", defaultVaultPodName, "-n", defaultVaultNamespace, "--",
-		"vault", "write", "-f", fmt.Sprintf("%s/rotate", defaultVaultKeyPath))
+		"vault", "write", "-n", defaultVaultEnterpriseNS, "-f", fmt.Sprintf("%s/rotate", defaultVaultKeyPath))
```
ardaguclu, line 216: We run 2 instances of Vault to test kms to kms migration. I assume that it doesn't matter?

tjungblu: we hardcode the defaultVaultNamespace, which is the primary deployment. This helper is only used for rotation. I assume the other one is too.

gangwgr: It shouldn't matter. Both KMS instances are configured to use the same underlying Vault setup/namespace for this test, so triggering the rotation on the default pod is enough to rotate the underlying key that both instances might be looking at. Unless we are not rotating on both.

PR #2455, opened 2026-09-03
OCPBUGS-109578: fix oc darwin client hcp login when insecure tls flag is used

ehearne-redhat:
This fix resolves a major issue introduced in a recent change where logging in using oc on macos to a hcp client with insecure flag was used.

That PR was reverted - https://github.com/openshift/library-go/pull/2456 . 

On other clusters where there was no error, the
check now panics on strings.HasPrefix(err.Error(), ...) as there was no check to see if err != nil.

This was flagged by coderabbitai in library-go bump pr in oc.

https://github.com/openshift/oc/pull/2391#discussion_r3922929300

This fix fixes an issue seen when oc client is used to login to a HCP cluster on macOS. It falls back to kubeconfig CA when a string based x509 error is observed on the macOS platform.

See https://github.com/golang/go/blob/master/src/crypto/x509/root_darwin.go#L74 for reference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate verification error handling on macOS.
  * Untrusted certificate errors are now reported consistently, while successful verification and unrelated errors remain unaffected.

* **Tests**
  * Added coverage for macOS certificate-error conversion, including successful verification, untrusted certificates, and unrelated errors.

### DIFF
--- a/pkg/oauth/tokenrequest/request_token.go
+++ b/pkg/oauth/tokenrequest/request_token.go
@@ -5,11 +5,13 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
 	"net/url"
+	"runtime"
 	"slices"
 	"strings"
 
@@ -551,6 +553,7 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 	resp.Body.Close()
 
 	_, err = verifyServerCertChain(issuerURL.Hostname(), resp.TLS.PeerCertificates)
+
 	switch err.(type) {
 	case nil:
 		// copy the config so we can freely mutate it
@@ -584,6 +587,13 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 			klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
 			return restclient.TransportFor(clientConfig)
 		}
+		// could be string based x509 error...
+		err = convertErrorIfUnknownX509(runtime.GOOS, err)
+		var target unknownX509VerificationError
+		if errors.As(err, &target) {
+			klog.V(4).Infof("falling back to kubeconfig CA due to possible unknown x509 error: %v", err)
+			return restclient.TransportFor(clientConfig)
+		}
 		// unknown error, fail (ideally should never occur)
 		klog.V(4).Infof("unexpected error during system roots probe: %v", err)
 		return nil, err
@@ -607,3 +617,28 @@ func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509
 		DNSName:       dnsName,
 	})
 }
+
+// convertErrorIfUnknownX509 normalizes certificate verification errors on macOS.
+// root_darwin.go in the Go standard library has insufficient typed errors for the
+// macOS platform, leading to a generic string based "x509:" error rather than a
+// typed one. This wraps such an error in an unknownX509VerificationError so callers
+// can react to it the same way they do on Linux/Windows (e.g. falling back to the
+// kubeconfig CA), keeping the approach consistent across platforms. The original
+// error is preserved and remains recoverable via errors.Unwrap/errors.Is.
+//
+// goos is passed in (rather than read from runtime.GOOS) so the darwin branch can
+// be exercised in tests regardless of the platform they run on.
+// https://github.com/golang/go/blob/5a6340ff28c87e099f33c941e3d73e50d715ddf7/src/crypto/x509/root_darwin.go#L74
+func convertErrorIfUnknownX509(goos string, err error) error {
+	if goos == "darwin" && err != nil && strings.HasPrefix(err.Error(), "x509:") {
+		return unknownX509VerificationError{err}
+	}
+	return err
+}
+
+// unknownX509VerificationError wraps an opaque, string based x509 verification
+// error (see convertErrorIfUnknownX509) so it can be matched by type while still
+// preserving the underlying error.
+type unknownX509VerificationError struct{ error }
+
+func (e unknownX509VerificationError) Unwrap() error { return e.error }
--- a/pkg/oauth/tokenrequest/request_token_test.go
+++ b/pkg/oauth/tokenrequest/request_token_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -755,6 +756,67 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
 					t.Errorf(`expected error "%s", got error "%s"`, tc.expectError, err.Error())
 					return
 				}
+			}
+		})
+	}
+}
+
+func TestConvertErrorIfUnknownX509(t *testing.T) {
+	x509Err := errors.New("x509: certificate signed by unknown authority")
+	otherErr := errors.New("some other error")
+
+	for _, tc := range []struct {
+		name string
+		goos string
+		err  error
+		// wrapExpected is true when the error is expected to be wrapped in an
+		// unknownX509VerificationError; otherwise the original error must be
+		// returned unchanged.
+		wrapExpected bool
+	}{
+		{
+			name:         "nil error stays nil on darwin",
+			goos:         "darwin",
+			err:          nil,
+			wrapExpected: false,
+		},
+		{
+			name:         "x509-prefixed error is wrapped on darwin",
+			goos:         "darwin",
+			err:          x509Err,
+			wrapExpected: true,
+		},
+		{
+			name:         "non-x509 error is returned unchanged on darwin",
+			goos:         "darwin",
+			err:          otherErr,
+			wrapExpected: false,
+		},
+		{
+			name:         "x509-prefixed error is left unchanged on linux",
+			goos:         "linux",
+			err:          x509Err,
+			wrapExpected: false,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got := convertErrorIfUnknownX509(tc.goos, tc.err)
+
+			if tc.wrapExpected {
+				wrapped, ok := got.(unknownX509VerificationError)
+				if !ok {
+					t.Fatalf("expected unknownX509VerificationError, got %T: %v", got, got)
+				}
+				if !errors.Is(wrapped, tc.err) {
+					t.Fatalf("expected the original error to be unwrappable, got %v", errors.Unwrap(wrapped))
+				}
+				return
+			}
+
+			// nil passes through as nil; on non-darwin (or non-x509 errors)
+			// the original error is returned unchanged.
+			if got != tc.err {
+				t.Fatalf("expected the original error %v to be returned unchanged, got %T: %v", tc.err, got, got)
 			}
 		})
 	}


pkg/oauth/tokenrequest/request_token.go
```diff
@@ -608,7 +608,7 @@ func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509
 		DNSName:       dnsName,
 	})
 
-	if runtime.GOOS == "darwin" && strings.HasPrefix(err.Error(), "x509:") {
+	if runtime.GOOS == "darwin" && err != nil && strings.HasPrefix(err.Error(), "x509:") {
```
tchap: I know this is a minor fix, but I don't like that this is essentially impossible to test.

Could we make `verifyServerCertChain` accept `runtime.GOOS` as an argument and add a unit test that ensures the error is handled properly when it's `darwin` pls?

ehearne-redhat: That's now done - tested locally. :)

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +766,100 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
 	}
 }
 
+func TestVerifyServerCertChain_RuntimeOS(t *testing.T) {
+	// selfSignedCert builds a self-signed certificate for the given DNS name.
+	// A self-signed cert is never chained to a system trust root, so verifying
+	// it always fails with an x509 error, which is exactly the condition the
+	// darwin-specific branch in verifyServerCertChain rewrites.
+	selfSignedCert := func(t *testing.T, dnsName string) *x509.Certificate {
```
tchap: You can use a function from `library-go` here:

```golang
caConfig, err := libcrypto.MakeSelfSignedCAConfigForDuration(dnsName, time.Hour)
if err != nil {
    t.Fatalf("failed to generate cert: %v", err)
}
untrustedChain := caConfig.Certs
```

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -593,7 +593,7 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 
 // verifyCertChain uses the system trust bundle in order to perform validation
 // of a certificate chain
-func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509.Certificate, error) {
+func verifyServerCertChain(runtimeOS string, dnsName string, chain []*x509.Certificate) ([][]*x509.Certificate, error) {
```
tchap: Could we make `runtimeOS` the last argument actually pls?

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +766,100 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
...
+			} else if err != nil {
+				t.Fatalf("expected no error, got %v", err)
+			}
+
+			if tc.expectUnknownAuthority {
+				if _, ok := err.(x509.UnknownAuthorityError); !ok {
+					t.Fatalf("expected error of type x509.UnknownAuthorityError, got %T: %v", err, err)
+				}
+			}
+		})
```
tchap, line 821: Honestly I would prefer having `checkError` function in the table that would encapsulate whatever is needed, because this is just not pretty to read. It can accept `t` and fail as needed.

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +762,69 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
...
+			runtimeOS:  "linux",
+			dnsName:    dnsName,
+			chain:      nil,
+			checkError: requireError,
+		},
+		{
+			name:       "darwin rewrites an x509 error into a typed UnknownAuthorityError",
+			runtimeOS:  "darwin",
+			dnsName:    dnsName,
+			chain:      untrustedChain,
```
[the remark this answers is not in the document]
tchap: @ehearne-redhat Actually this is not a bad idea. Just create a small helper for error handling and test that helper, something like:

```golang
func darwinCertVerifyErr(err error, cert *x509.Certificate) error {
  if err == nil {
    return nil
  }
  if strings.HasPrefix(err.Error(), "x509:") {
    return x509.UnknownAuthorityError{Cert: cert}
  }
  return err
}
```

and then in the function itself we can do

```golang
 if runtimeOS == "darwin" {
    err = darwinCertVerifyErr(err, chain[0])
  }
```

ehearne-redhat:
@coderabbitai what do you make of the latest change?

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +761,56 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
...
+					t.Fatalf("expected x509.UnknownAuthorityError, got %T: %v", err, err)
+				}
+				if unknownAuthErr.Cert != cert {
+					t.Fatalf("expected the offending cert to be preserved on the error")
+				}
+			},
+		},
+		{
+			name: "non-x509 error is returned unchanged",
+			err:  fmt.Errorf("some other error"),
```
tchap: Could we use `errors.New` here pls since there is no formatting?

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +761,56 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
...
+			checkError: func(t *testing.T, err error) {
+				t.Helper()
+				if err != nil {
+					t.Fatalf("expected nil error, got %T: %v", err, err)
+				}
+			},
+		},
+		{
+			name: "x509-prefixed error is rewritten to UnknownAuthorityError",
+			err:  fmt.Errorf("x509: certificate signed by unknown authority"),
```
tchap: Same here, could we use errors.New here pls since there is no formatting?

tchap:
/lgtm

Thanks for the patience 😅

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -608,18 +608,28 @@ func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509
...
+// same way they do on Linux/Windows (e.g. falling back to the kubeconfig CA),
+// keeping the approach consistent across platforms.
+// https://github.com/golang/go/blob/5a6340ff28c87e099f33c941e3d73e50d715ddf7/src/crypto/x509/root_darwin.go#L74
+func darwinCertVerifyErr(err error, cert *x509.Certificate) error {
+	if err == nil {
+		return nil
+	}
+	if strings.HasPrefix(err.Error(), "x509:") {
+		return x509.UnknownAuthorityError{Cert: cert}
+	}
```
everettraven: Do we actually want every x509 error to be rolled up into an unknown authority error?

Do we know which errors are equivalent that Go doesn't have sufficient coverage for?

I know this was pre-existing logic, but this makes _every_ error (even typed ones) returned from the x509 verification on macOS considered an unknown authority error. This seems potentially problematic to me.

We should, at the very least, ensure that we are only doing this for unknown errors. Even then, re-using the `x509.UnknownAuthorityError` seems problematic because it obscures the actual error that was returned. Maybe we should instead introduce a net-new local error type that we wrap unknown errors in if we are confident they are certificate verification errors? Then callers can add that local error type to their determination to fall back to kubeconfig CA?

ehearne-redhat: I agree - and I wish that I had a better mechanism to do this. 

As mentioned in #2438 the actual crypto implementation is not effective enough to do so. It literally has three out of many error codes that macOS could throw around this issue. 

https://github.com/golang/go/blob/master/src/crypto/x509/internal/macos/security.go#L23-L30

I also believe that how it throws a string-based x509 error isn't ideal when it could have explored at least 10 or so similar codes. 

There is a very long list of codes that Apple actually does provide in two files that you're more than welcome to look at :)

https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/header_symlinks/Security/CSCommon.h#L142

https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/header_symlinks/Security/SecBase.h#L394

Getting the exact error code thrown here when this issue occurs is what would make it more concrete.

From what I can understand, x509.UnknownAuthorityError is the error that makes the most sense (though it does seem like a blanket error in this case) as the system actually doesn't know the authority of the cert because it is self-signed. 

What I'm getting from this comment is we would like to know what the error code is that macOS throws, which is technically possible with a few hacky modifications to this library in oc. If we can get that code, we can use this as part of the check. Once we have that, we can create our own error type to use as part of the switch statement.

ehearne-redhat: Also, I do want to note that we need the fix in as soon as possible so oc does not fail on standalone clusters or hcp clusters on non-darwin platforms when using the insecure tls verify flag. It would not be good if the state of library-go now was part of a bump in other components that may use this code other than oc. 

Would you be happy with getting this in as a safety patch as-is, and I can absolutely get the specific code and create new error type in meantime? You do raise good points, but my main concern is how this could affect components that use this code and they bump to the current state.

everettraven: I don't actually care about the exact error code. I'm more concerned with us overwriting the error that is returned from the verification process to always be "x509: unknown certificate authority".

We might actually trust the certificate authority, but the certificate might have some other problem with it. We overwrite any of that with a generic error that could be entirely incorrect with the current approach.

everettraven: FWIW, I do think there is a better way to do this.

Something like:
```go
...
switch err.(type) {
  case x509.UnknownAuthorityError, x509.HostnameError, x509.CertificateInvalidError:
      return err // verification already mapped to a typed error
  default:
      if strings.HasPrefix(err.Error(), "x509:") {
          return &UnknownX509VerificationError{err} // we now return a custom typed error that signals an unknown x509 verification error
      }
      
      return err // any other error (i.e a non-x509 verification error)
}
...

type UnknownX509VerificationError struct {
  error
}
```

is along the lines of what I am thinking. It preserves any errors that don't meet our criteria and ensures we don't overwrite anything we don't intend to.

ehearne-redhat: OK - that makes sense, and helps a bit with getting the implementation done. :) 

I think that sounds quite reasonable to me - I'll get that done. Thanks for the example implementation!

tchap: Actually based on https://github.com/golang/go/blob/master/src/crypto/x509/root_darwin.go#L65 it seems to me we are doing pretty bad conversion, basically dropping the error types that make sense. We may only want to do additional wrapping when the default case is hit...

(This is references in the original code, I am guessing that is what is casing the returned error.)

tchap: Yeah, so I agree with Bryce.

ehearne-redhat: Yeah but you see this default case is what is always being hit because go implementation doesn't give us enough codes. So that's why we have the generic string error. I am already in the process of implementing. :D

bertinatto (commented):
Created https://github.com/openshift/library-go/pull/2456 to buy us time.

Please revert the revert from 2456 and create a [proof PR](https://github.com/openshift/library-go#proof-prs) for OC. Let's also manually test the change to make sure we're not missing anything.

ardaguclu:
@ehearne-redhat can you please give me some details about why did we revert the previous one?

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -607,3 +610,28 @@ func verifyServerCertChain(dnsName string, chain []*x509.Certificate) ([][]*x509
...
+// kubeconfig CA), keeping the approach consistent across platforms. The original
+// error is preserved and remains recoverable via errors.Unwrap/errors.Is.
+//
+// goos is passed in (rather than read from runtime.GOOS) so the darwin branch can
+// be exercised in tests regardless of the platform they run on.
+// https://github.com/golang/go/blob/5a6340ff28c87e099f33c941e3d73e50d715ddf7/src/crypto/x509/root_darwin.go#L74
+func convertErrorIfUnknownX509(goos string, err error) error {
+	if goos == "darwin" && err != nil && strings.HasPrefix(err.Error(), "x509:") {
+		return unknownX509VerificationError{err}
+	}
```
everettraven, line 635: While I can understand the simplification of just wrapping every error in this error type, is this actually what we want?

_IF_ the x509 verification process returns a known concretely typed error, we probably want to just return that typed error instead of wrapping it in a different type.

ehearne-redhat:
> @ehearne-redhat can you please give me some details about why did we revert the previous one?

Hey @ardaguclu - the reason why I asked for the revert was because there was no nil error check in the implementation which would cause a panic when we call `err.Error()` . 

@bertinatto explained it would be a good idea to revert so it would give us more time on coming up with a better solution so components e.g. oc bumping library-go would not begin to get failures. 

https://github.com/openshift/oc/pull/2391#discussion_r3922929300

pkg/oauth/tokenrequest/request_token.go
ardaguclu: In my opinion, instead of calling this function for every error type, we should convert to our custom error only if goos is darwin and error has prefix of `x509`.

ehearne-redhat: OK - makes sense. I have been thinking about where exactly to move it. 

I'm thinking where the default switch case is probably best. I'll push it up and hopefully this suits. :)

ehearne-redhat:
Hey @ardaguclu @everettraven - apologies for the delay in pushing. Feel free to take a look when you get a chance and let me know your thoughts. :)

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -584,6 +586,13 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 			klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
 			return restclient.TransportFor(clientConfig)
 		}
+		// could be string based x509 error...
+		err = convertErrorIfUnknownX509(runtime.GOOS, err)
```
ardaguclu, line 591: I was thinking something like this;

```go
default:
        err = convertErrorIfUnknownX509(runtime.GOOS, err)
		switch err {
		case io.EOF, io.ErrUnexpectedEOF, io.ErrNoProgress, unknownX509VerificationError:
			// also fallback on various io errors
			klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
			return restclient.TransportFor(clientConfig)
		}
		// unknown error, fail (ideally should never occur)
		klog.V(4).Infof("unexpected error during system roots probe: %v", err)
		return nil, err
	}
```

or is this suggestion contradicting with the other reviews :)?

ehearne-redhat: It may contradict as I believe the point @everettraven was making, was that if the error happens to meet the three criteria and it was actually not the error we intended it to be, then we are wrapping the error in the wrong error type. 

The reason why I moved it below all known type checks was so we can do one last check to see if the error is the string based x509 before failing. It mitigates that concern. This way we don't erroneously (pun intended) wrap the wrong error.

If it is a fear of nesting, I believe the effect should still be the same as before as the switch statement is not further nested but placed below a switch statement.

Let me know your thoughts - there could be something I am missing.

ardaguclu: That makes sense. Would it make sense if we use errors.is instead of another nested switch statement to compare with our custom error type?

ehearne-redhat: Sure - I'll update it now. :)

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -584,6 +587,12 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
...
+		if unknownX509Err := (unknownX509VerificationError{}); errors.As(err, &unknownX509Err) {
```
ardaguclu:
```suggestion
		err = convertErrorIfUnknownX509(runtime.GOOS, err)

var target unknownX509VerificationError
if errors.As(err, &target) {
	klog.V(4).Infof("falling back to kubeconfig CA due to possible unknown x509 error: %v", err)
	return restclient.TransportFor(clientConfig)
}
```

ehearne-redhat: IDE suggested `errors.AsType` . Let me know if that's ok, or if you prefer this way. :)

pkg/oauth/tokenrequest/request_token.go
```diff
@@ -584,6 +587,12 @@ func transportWithSystemRoots(issuer string, clientConfig *restclient.Config) (h
 			klog.V(4).Infof("falling back to kubeconfig CA due to possible IO error: %v", err)
 			return restclient.TransportFor(clientConfig)
 		}
+		// could be string based x509 error...
+		err = convertErrorIfUnknownX509(runtime.GOOS, err)
+		if _, ok := errors.AsType[unknownX509VerificationError](err); ok {
```
ardaguclu: Does this mechanism work well if err is nil?

ehearne-redhat: It seems to, here is the code for `errors.AsType()`

```golang
func AsType[E error](err error) (E, bool) {
	if err == nil {
		var zero E
		return zero, false
	}
	var pe *E // lazily initialized
	return asType(err, &pe)
}
```

ardaguclu:
@ehearne-redhat it would be better to open a fake bump in oc to see everything works properly.

ehearne-redhat:
> @ehearne-redhat it would be better to open a fake bump in oc to see everything works properly.

Yes - I do already have one set up in https://github.com/openshift/oc/pull/2391 so I will go and update it shortly.

pkg/oauth/tokenrequest/request_token_test.go
```diff
@@ -760,6 +761,67 @@ func TestRequestToken_BrowserFlow(t *testing.T) {
 	}
 }
 
+func TestConvertErrorIfUnknownX509(t *testing.T) {
```
everettraven, line 764: Are there any additional tests cases it would make sense to add for the method that calls this new conversion function to ensure that we are now correctly falling back on the error type this new conversion function returns?

ehearne-redhat: There could be additional ones - for example, looking at it again, `"nil error stays nil on darwin"` should also exist for linux i.e.`"nil error stays nil on linux"`. 

Also we could add a linux equivalent of `"non-x509 error is returned unchanged on darwin"` i.e. `"non-x509 error is returned unchanged on linux"`

That way, the three scenarios are tested fully on each platform. So 6 test cases in total.

How does this sound? I feel adding more platforms proves pointless because in my mind there is darwin and non darwin scenarios. Then the type of errors the function could receive would be `nil`, non-nil, and x509 string based errors.

everettraven: I guess what I'm hoping to see is that we have explicit tests in place that test when the verification process returns an unknown error case that meets the darwin + "x509:" prefix criteria that we are correctly falling back instead of returning the error.

ehearne-redhat: I think I need more clarification because I feel that those cases are accounted for? Unless you are looking for the mechanism itself to change... I feel I may be missing something.

pkg/oauth/tokenrequest/request_token.go
bertinatto: just keep in mind that `errors.AsType` is only available in go 1.26 (?), so if you want to backport this patch to OCP 4.22 (go 1.25), it'll likely not be possible

ehearne-redhat: I see - in that case, I will use `errors.As` for this reason in case someone does want to backport.

pkg/oauth/tokenrequest/request_token_test.go
everettraven: This looks like it correctly tests the net-new function that you've added.

What I'm getting at is that it seems to be the case that there is no explicit testing for the `transportWithSystemRoots` method that calls it to ensure that the fallback behaviors are working as expected.

Since we have run into this issue, I think it would be helpful to ensure we have some additional test coverage here to make sure that the behaviors of that `transportWithSystemRoots` method aligns with what our expectations are (i.e if the server certificate chain verification returns an error, what do we expect to happen?).

It doesn't look like there is an existing test for this so I'll leave it up to @bertinatto to decide whether or not we should block this PR on explicit testing for this path of the code base to be added.

I just noticed that this looks like a gap that should be filled, and since we are already modifying this part of the code we might as well fill that gap.

ehearne-redhat: Yeah - I guess it did exist before in this PR and then through the review process we tried to minimise. I don't mind adding it back but I'll wait for Fabio's reply. 

Sorry I didn't make the connection - I think it makes perfect sense to add if it is desired.

bertinatto: I agree with Bryce it'd be great to have additional higher level tests. However, given the history of this PR (and its predecessor), I feel like we don't need to block what we have so far. 

@ehearne-redhat when/if you'd have some time, I'd greatly appreciate if you could add those tests and help improve these corner of library-go ;)

bertinatto (commented):
/approve

pkg/oauth/tokenrequest/request_token_test.go
ehearne-redhat: Absolutely - I will add it on the to-do list. Thanks for the approve!

PR #2461, opened 2026-09-08
NO-JIRA: add KMS remote-key health-report convergence helpers

tjungblu:
## Summary
- Add health-report convergence helpers to detect a unanimous remote key ID and to filter reports to the current write key.

Enhancement: https://github.com/openshift/enhancements/pull/2041
E2E PR: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2291
Tracking PR: https://github.com/openshift/library-go/pull/2452

## Summary by CodeRabbit

* **New Features**
  * Added KMS health reporting support to determine when all reports consistently identify the same non-empty remote encryption key.
  * Added filtering of health reports by encryption key ID for more targeted status analysis.

* **Tests**
  * Added coverage for consistent, conflicting, empty, and incomplete remote key reports.
  * Added validation that key-based filtering returns the complete matching report set.

### DIFF
--- a/pkg/operator/encryption/kms/health/convergence.go
+++ b/pkg/operator/encryption/kms/health/convergence.go
@@ -1,0 +1,39 @@
+package health
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// ConvergedRemoteKeyID returns the unanimous RemoteKeyID across reports when every
+// report carries the same non-empty RemoteKeyID. An empty return value indicates
+// no convergence (empty input, conflicting IDs, or any empty RemoteKeyID).
+// Callers must filter reports to the relevant key ID and ensure the slice
+// includes every expected node before treating a non-empty result as converged.
+func ConvergedRemoteKeyID(reports []operatorv1.KMSPluginHealthReport) string {
+	if len(reports) == 0 {
+		return ""
+	}
+	remoteKeyID := reports[0].RemoteKeyID
+	if remoteKeyID == "" {
+		return ""
+	}
+	for _, report := range reports[1:] {
+		if report.RemoteKeyID != remoteKeyID {
+			return ""
+		}
+	}
+	return remoteKeyID
+}
+
+// ReportsForKeyID returns health reports for the given encryption key ID (kms-{keyID}.sock).
+// During KMS-to-KMS provider migration multiple plugins may report per node; callers use
+// the current write key's keyID so backup/read-only keys are excluded from rotation logic.
+func ReportsForKeyID(reports []operatorv1.KMSPluginHealthReport, keyID string) []operatorv1.KMSPluginHealthReport {
+	filtered := make([]operatorv1.KMSPluginHealthReport, 0, len(reports))
+	for _, report := range reports {
+		if report.KeyID == keyID {
+			filtered = append(filtered, report)
+		}
+	}
+	return filtered
+}
--- a/pkg/operator/encryption/kms/health/convergence_test.go
+++ b/pkg/operator/encryption/kms/health/convergence_test.go
@@ -1,0 +1,77 @@
+package health
+
+import (
+	"testing"
+
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+func TestConvergedRemoteKeyID(t *testing.T) {
+	scenarios := []struct {
+		name            string
+		reports         []operatorv1.KMSPluginHealthReport
+		wantRemoteKeyID string
+	}{
+		{
+			name: "unanimous",
+			reports: []operatorv1.KMSPluginHealthReport{
+				{KeyID: "3", RemoteKeyID: "remote-a"},
+				{KeyID: "3", RemoteKeyID: "remote-a"},
+			},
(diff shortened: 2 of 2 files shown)


pkg/operator/encryption/kms/health/convergence.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	operatorv1 "github.com/openshift/api/operator/v1"
+)
+
+// ConvergenceResult reports whether health reports agree on a single remote key ID.
+type ConvergenceResult struct {
+	Converged   bool
+	RemoteKeyID string
+}
```
bertinatto: nit: maybe this would negatively affect how the consumers of this would look like, but I noticed we could get rid of this type and return multiple values here

bertinatto (commented):
/lgtm

adding a hold so that @ardaguclu has a chance to review as well:

/hold

pkg/operator/encryption/kms/health/convergence.go
ardaguclu: I think, we don't need neither `ConvergenceResult` nor `Converged` field. Emptiness of RemoteKeyID is already an indicator of convergence.

So that functions can only return RemoteKeyID?.

pkg/operator/encryption/kms/health/convergence.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ConvergedRemoteKeyID returns Converged=true when every report in the slice carries
+// the same RemoteKeyID. The caller is responsible for filtering reports and handling
+// missing or empty entries.
+func ConvergedRemoteKeyID(reports []operatorv1.KMSPluginHealthReport) ConvergenceResult {
+	if len(reports) == 0 {
+		return ConvergenceResult{}
+	}
+	remoteKeyID := reports[0].RemoteKeyID
```
ardaguclu, line 16: Empty remoteKeyIDs should not lead to the convergence?.

Do we need something like this;

```go
remoteKeyID := reports[0].RemoteKeyID
        if remoteKeyID == "" {
                return ""
        }
        for _, report := range reports[1:] {
                if report.RemoteKeyID != remoteKeyID {
                        return ""
                }
        }

```

pkg/operator/encryption/kms/health/convergence.go
tjungblu: all good, we can do that. @ardaguclu do you also like to see multiple returns here?

pkg/operator/encryption/kms/health/convergence.go
```diff
@@ -0,0 +1,43 @@
...
+		Converged:   true,
+		RemoteKeyID: remoteKeyID,
+	}
+}
+
+// ReportsForKeyID returns health reports for the given encryption key ID (kms-{keyID}.sock).
+// During KMS-to-KMS provider migration multiple plugins may report per node; callers use
+// the current write key's keyID so backup/read-only keys are excluded from rotation logic.
+func ReportsForKeyID(reports []operatorv1.KMSPluginHealthReport, keyID string) []operatorv1.KMSPluginHealthReport {
+	filtered := make([]operatorv1.KMSPluginHealthReport, 0, len(reports))
```
ardaguclu, line 32: Is it possible the health reporter has not yet reported the health status of one of the nodes?. If it is possible, isn't there a theoretical chance that this new report may differ from the others?

pkg/operator/encryption/kms/health/convergence.go
ardaguclu: I think, we can just return RemoteKeyID?

pkg/operator/encryption/kms/health/convergence.go
ardaguclu: I think we should short cut with non-convergence, if health reporter misses a status of one of the nodes.

tjungblu: the controller can do the check, if there's an unexpected amount of health checks for the keyId we need. There's also the ready/unready node information

pkg/operator/encryption/kms/health/convergence.go
tjungblu: the way the API is designed, there can't be an empty RemoteKeyId. But I'll address this case.

pkg/operator/encryption/kms/health/convergence.go
tjungblu: fixed

pkg/operator/encryption/kms/health/convergence.go
ardaguclu: I think, key controller and health reporter do not know the number of nodes. So we have to rely on whatever reported by health reporter.

ardaguclu:
/lgtm
/approve
/hold
for @bertinatto

pkg/operator/encryption/kms/health/convergence.go
tjungblu: The keycontroller has the operator status and ReadyReplicas (or the operand?), so I think we can at least get a hint.

bertinatto (commented):
/hold cancel
LGTM

PR #2462, opened 2026-09-08
NO-JIRA: expose the specific KMS migration scenarios

tjungblu:
to be able to configure them per operator, avoiding to have a cross-operator dependency on migration resources.

## Summary by CodeRabbit

- **Tests**
  - Improved the reuse and organization of encryption key management validation scenarios.
  - Existing scenario behavior, configuration, callbacks, and migration checks remain unchanged.
  - No user-facing functionality or product behavior was changed.


tjungblu:
/close

we can use this with a test command

PR #2465, opened 2026-09-10
NO-JIRA: add RemoteKeyState to KMSState

tjungblu:
## Summary by CodeRabbit

- **New Features**
  - Added visibility into remote KMS key rotation progress, including target keys, migrated keys, convergence timestamps, and converged key identifiers.
  - Improved encryption state reporting for KMS-managed keys, making migration progress easier to monitor.
  - Non-KMS keys continue to be handled safely without exposing remote-key migration details.

### DIFF
--- a/pkg/operator/encryption/secrets/types.go
+++ b/pkg/operator/encryption/secrets/types.go
@@ -67,6 +67,18 @@ const (
 	// values fetched from the referenced configmap in openshift-config. The full data key is
 	// constructed as prefix + configMapName + separator + dataKey.
 	encryptionSecretKMSConfigMapDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-configmap-"
+
+	// encryptionSecretTargetRemoteKeyID is the target remote KMS key ID to migrate toward.
+	encryptionSecretTargetRemoteKeyID = "encryption.apiserver.operator.openshift.io/target-remote-key-id"
+
+	// encryptionSecretMigratedRemoteKeyID is the last fully migrated remote KMS key ID.
+	encryptionSecretMigratedRemoteKeyID = "encryption.apiserver.operator.openshift.io/migrated-remote-key-id"
+
+	// encryptionSecretRemoteKeyConvergedAt records when a candidate remote key ID first achieved cluster convergence.
+	encryptionSecretRemoteKeyConvergedAt = "encryption.apiserver.operator.openshift.io/remote-key-converged-at"
+
+	// encryptionSecretRemoteKeyConvergedID is the candidate remote key ID the converged-at timestamp belongs to.
+	encryptionSecretRemoteKeyConvergedID = "encryption.apiserver.operator.openshift.io/remote-key-converged-id"
 )
 
 // MigratedGroupResources is the data structured stored in the
--- a/pkg/operator/encryption/state/types.go
+++ b/pkg/operator/encryption/state/types.go
@@ -68,6 +68,32 @@ func (k *KeyState) HasKMSConfigMapData() bool {
 	return k != nil && k.KMS != nil && len(k.KMS.PluginConfigMapData.entries) > 0
 }
 
+// RemoteKey returns the remote key rotation state. Non-KMS keys have no remote key.
+func (k *KeyState) RemoteKey() RemoteKeyState {
+	if k == nil || k.KMS == nil {
+		return RemoteKeyState{}
+	}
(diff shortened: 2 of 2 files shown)


pkg/operator/encryption/state/types.go
```diff
@@ -81,6 +97,9 @@ type KMSState struct {
 
 	// PluginConfigMapData stores data key-value pairs fetched from referenced configmaps.
 	PluginConfigMapData KMSReferenceData
+
+	// RemoteKey tracks KMS remote key rotation annotations on the backing secret.
+	RemoteKey RemoteKeyState
```
ardaguclu, line 112: Should we expect that this data will always be non-empty, if the encryption mode is KMS?

pkg/operator/encryption/secrets/types.go
```diff
@@ -67,6 +67,18 @@ const (
 	// values fetched from the referenced configmap in openshift-config. The full data key is
 	// constructed as prefix + configMapName + separator + dataKey.
 	encryptionSecretKMSConfigMapDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-configmap-"
+
+	// EncryptionSecretTargetRemoteKeyID is the target remote KMS key ID to migrate toward.
+	EncryptionSecretTargetRemoteKeyID = "encryption.apiserver.operator.openshift.io/target-remote-key-id"
```
ardaguclu: Do we really need to export these annotations?. Maybe we can add them as unexported and only export them when it is needed?

pkg/operator/encryption/state/types.go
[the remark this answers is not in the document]
tjungblu: it's okay, the wiring comes later

pkg/operator/encryption/state/types.go
```diff
@@ -68,6 +68,22 @@ func (k *KeyState) HasKMSConfigMapData() bool {
...
+func (k *KeyState) RemoteKey() RemoteKeyState {
+	if k == nil || k.KMS == nil {
+		return RemoteKeyState{}
+	}
+	return k.KMS.RemoteKey
+}
+
+// RemoteKeyState is the in-memory view of remote key rotation annotations.
+type RemoteKeyState struct {
+	TargetRemoteKeyID   string
```
ardaguclu: Can you please add short documentation for each field by also mentioning which ones we should always expect non-nil and which ones can be nil?

pkg/operator/encryption/secrets/types.go
ardaguclu: If these are exported for unit tests, we don't need this. We can hardcode them in unit tests.

tjungblu: depends on where you want to read them from. I can make them private for now.

pkg/operator/encryption/state/types.go
tjungblu: after a successful preflight it is non-empty, yes

pkg/operator/encryption/state/types.go
tjungblu: none of them can be nil, because string default value is "".  Time is an empty 0 struct:
> January 1, year 1, 00:00:00 UTC.

tjungblu: left some comments to explain in what scenario they are used, set and otherwise default.

pkg/operator/encryption/secrets/types.go
```diff
@@ -67,6 +67,18 @@ const (
 	// values fetched from the referenced configmap in openshift-config. The full data key is
 	// constructed as prefix + configMapName + separator + dataKey.
 	encryptionSecretKMSConfigMapDataPrefix = "encryption.apiserver.operator.openshift.io-kms-plugin-configmap-"
+
+	// encryptionSecretTargetRemoteKeyID is the target remote KMS key ID to migrate toward.
+	encryptionSecretTargetRemoteKeyID = "encryption.apiserver.operator.openshift.io/target-remote-key-id"
```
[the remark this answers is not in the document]
tjungblu: I don't get it either, but my reviewers seem to like 50 line PRs that only add strings.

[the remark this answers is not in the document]
tjungblu: heh I don't make the rules

ardaguclu:
/lgtm
/approve
/hold
in case @bertinatto may want to look at

bertinatto (commented):
/lgtm

PR #2466, opened 2026-09-10
NO-JIRA: kms rotation: add read/write helpers for remote key secret annotations

tjungblu:
## Summary by CodeRabbit

- **New Features**
  - Remote encryption key rotation state is now preserved with KMS secrets.
  - Remote key identifiers and convergence timestamps are restored when reading secret configuration and saved when generating secrets.

- **Bug Fixes**
  - Invalid rotation timestamps and inconsistent convergence details are now reported as errors instead of being silently ignored.
  - Empty remote key values now correctly clear previously stored rotation annotations.

### DIFF
--- a/pkg/operator/encryption/secrets/remote_key.go
+++ b/pkg/operator/encryption/secrets/remote_key.go
@@ -1,0 +1,65 @@
+package secrets
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ReadRemoteKeyStateFromSecret reads remote key rotation annotations from a key secret.
+func ReadRemoteKeyStateFromSecret(s *corev1.Secret) (state.RemoteKeyState, error) {
+	if s == nil {
+		return state.RemoteKeyState{}, nil
+	}
+	annotations := s.Annotations
+	rk := state.RemoteKeyState{
+		TargetRemoteKeyID:   annotations[encryptionSecretTargetRemoteKeyID],
+		MigratedRemoteKeyID: annotations[encryptionSecretMigratedRemoteKeyID],
+		ConvergedID:         annotations[encryptionSecretRemoteKeyConvergedID],
+	}
+	if v, ok := annotations[encryptionSecretRemoteKeyConvergedAt]; ok && len(v) > 0 {
+		ts, err := time.Parse(time.RFC3339, v)
+		if err != nil {
+			return state.RemoteKeyState{}, fmt.Errorf("secret %s/%s has invalid %s annotation: %v", s.Namespace, s.Name, encryptionSecretRemoteKeyConvergedAt, err)
+		}
+		rk.ConvergedAt = ts
+	}
+
+	if err := rk.Validate(); err != nil {
+		return state.RemoteKeyState{}, fmt.Errorf("secret %s/%s is invalid %s", s.Namespace, s.Name, err)
+	}
+
+	return rk, nil
+}
+
+// applyRemoteKeyAnnotations writes remote key rotation annotations into the given map.
+// Empty values remove the corresponding annotation keys. Returns an error when the remote key state is invalid
+func applyRemoteKeyAnnotations(annotations map[string]string, rk state.RemoteKeyState) error {
+	if err := rk.Validate(); err != nil {
+		return err
+	}
+
+	delete(annotations, encryptionSecretTargetRemoteKeyID)
+	if rk.TargetRemoteKeyID != "" {
+		annotations[encryptionSecretTargetRemoteKeyID] = rk.TargetRemoteKeyID
+	}
+
+	delete(annotations, encryptionSecretMigratedRemoteKeyID)
+	if rk.MigratedRemoteKeyID != "" {
+		annotations[encryptionSecretMigratedRemoteKeyID] = rk.MigratedRemoteKeyID
+	}
+
+	delete(annotations, encryptionSecretRemoteKeyConvergedID)
+	if rk.ConvergedID != "" {
+		annotations[encryptionSecretRemoteKeyConvergedID] = rk.ConvergedID
+	}
+
+	delete(annotations, encryptionSecretRemoteKeyConvergedAt)
+	if !rk.ConvergedAt.IsZero() {
+		annotations[encryptionSecretRemoteKeyConvergedAt] = rk.ConvergedAt.Format(time.RFC3339)
+	}
+
+	return nil
+}
--- a/pkg/operator/encryption/secrets/remote_key_test.go
+++ b/pkg/operator/encryption/secrets/remote_key_test.go
@@ -1,0 +1,117 @@
+package secrets
+
+import (
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+)
+
+func TestReadRemoteKeyAnnotations(t *testing.T) {
+	ts := time.Date(2026, 8, 31, 10, 0, 0, 0, time.UTC)
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-config-managed",
+			Name:      "encryption-key-test-1",
+			Annotations: map[string]string{
+				encryptionSecretTargetRemoteKeyID:    "remote-old",
+				encryptionSecretMigratedRemoteKeyID:  "remote-old",
+				encryptionSecretRemoteKeyConvergedID: "remote-new",
+				encryptionSecretRemoteKeyConvergedAt: ts.Format(time.RFC3339),
+			},
+		},
+	}
+
+	got, err := ReadRemoteKeyStateFromSecret(secret)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if got.TargetRemoteKeyID != "remote-old" || got.MigratedRemoteKeyID != "remote-old" {
+		t.Fatalf("unexpected ids: %#v", got)
+	}
+	if got.ConvergedID != "remote-new" || !got.ConvergedAt.Equal(ts) {
+		t.Fatalf("unexpected convergence: %#v", got)
+	}
+}
+
+func TestReadRemoteKeyAnnotationsInvalidConvergence(t *testing.T) {
+	tests := []struct {
+		name      string
+		id        string
+		at        time.Time
+		expectErr bool
+	}{
+		{name: "both default", id: "", at: time.Time{}, expectErr: false},
+		{name: "only id", id: "id", at: time.Time{}, expectErr: true},
+		{name: "only time", id: "", at: time.Now(), expectErr: true},
+		{name: "both set", id: "id", at: time.Now(), expectErr: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			secret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "openshift-config-managed",
+					Name:      "encryption-key-test-1",
+					Annotations: map[string]string{
+						encryptionSecretRemoteKeyConvergedID: tt.id,
+						encryptionSecretRemoteKeyConvergedAt: tt.at.Format(time.RFC3339),
+					},
+				},
+			}
+
+			_, err := ReadRemoteKeyStateFromSecret(secret)
+			if tt.expectErr && err == nil {
+				t.Errorf("expected error but got nil")
+			} else if !tt.expectErr && err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+		})
+	}
+}
+
+func TestInvalidRemoteKeyAnnotationBeforeSetting(t *testing.T) {
+	err := applyRemoteKeyAnnotations(make(map[string]string), state.RemoteKeyState{
+		ConvergedAt: time.Time{},
+		ConvergedID: "1337",
+	})
+	if err == nil {
+		t.Errorf("expected error but got nil")
+	}
+}
+
+func TestApplyRemoteKeyAnnotations(t *testing.T) {
+	now := time.Now()
+	nowStr := now.Format(time.RFC3339)
+	annotations := map[string]string{
+		encryptionSecretTargetRemoteKeyID:    "remove",
+		encryptionSecretMigratedRemoteKeyID:  "keep",
+		encryptionSecretRemoteKeyConvergedID: "old",
+		encryptionSecretRemoteKeyConvergedAt: nowStr,
+	}
+	err := applyRemoteKeyAnnotations(annotations, state.RemoteKeyState{
+		TargetRemoteKeyID:   "", // should remove
+		MigratedRemoteKeyID: "keep",
+		ConvergedAt:         now,
+		// updated with a new key
+		ConvergedID: "new-key",
+	})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if _, ok := annotations[encryptionSecretTargetRemoteKeyID]; ok {
+		t.Fatal("expected target annotation to be removed")
+	}
+	if annotations[encryptionSecretMigratedRemoteKeyID] != "keep" {
+		t.Fatalf("expected annotation %s to stay", encryptionSecretMigratedRemoteKeyID)
+	}
+	if annotations[encryptionSecretRemoteKeyConvergedID] != "new-key" {
+		t.Fatalf("expected annotation %s to update", encryptionSecretRemoteKeyConvergedID)
+	}
+	if annotations[encryptionSecretRemoteKeyConvergedAt] != nowStr {
+		t.Fatalf("expected annotation %s to stay", encryptionSecretRemoteKeyConvergedAt)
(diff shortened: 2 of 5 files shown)


pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/client-go/util/retry"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+)
+
+// RemoteKeyAnnotations is the annotation-backed view of state.RemoteKeyState.
+type RemoteKeyAnnotations = state.RemoteKeyState
+
+// ReadRemoteKeyAnnotations reads remote key rotation annotations from a key secret.
+func ReadRemoteKeyAnnotations(s *corev1.Secret) (RemoteKeyAnnotations, error) {
```
tjungblu: I need this exported for the e2e test helpers, which I would like to get out as the next step to avoid having to constantly juggle with them.

ardaguclu: I think, at this stage, keeping the changes minimal is more important than the logical code sharing between core functionality and unit tests. So, we can drop all the functions in that file aside from `readRemoteKeyAnnotations`, `applyRemoteKeyAnnotations`. Unit tests can have their own read, apply functions.

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,125 @@
...
+	setOrDeleteAnnotation(annotations, EncryptionSecretMigratedRemoteKeyID, rk.MigratedRemoteKeyID)
+	setOrDeleteAnnotation(annotations, EncryptionSecretRemoteKeyConvergedID, rk.ConvergedID)
+	if rk.ConvergedAt.IsZero() {
+		delete(annotations, EncryptionSecretRemoteKeyConvergedAt)
+	} else {
+		annotations[EncryptionSecretRemoteKeyConvergedAt] = rk.ConvergedAt.Format(time.RFC3339)
+	}
+}
+
+func setOrDeleteAnnotation(annotations map[string]string, key, value string) {
```
ardaguclu: We can do annotation operations inside the `applyRemoteKeyAnnotations`, so we can remove this function.

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -66,6 +66,11 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 		key.Mode = keyMode
 	case state.KMS:
 		key.KMS = &state.KMSState{}
+		remoteKey, err := readRemoteKeyAnnotations(s.Annotations, s.Namespace, s.Name)
+		if err != nil {
+			return state.KeyState{}, err
```
ardaguclu: Any error causes the degradation of the apiservers. 

As far as I can see, error can only occur, if the timestamp field is manually corrupted in the Secret. In that case, degradation of apiserver is correct approach.

Is there any other possibility that error can occur?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -162,6 +167,10 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Annotations[EncryptionSecretMigratedResources] = string(bs)
 	}
 
+	if ks.Mode == state.KMS {
```
ardaguclu: We shouldn't have this if statement. If annotation is added, we should convert it. Annotation must not be added to the Secret, if encryption mode is not KMS anyway.

pkg/operator/encryption/secrets/remote_key.go
tjungblu: I don't really get the risk here. We're reading annotations into a struct and we're not using them for unit tests, but for the only end2end test that can verify this feature even works. 

Are you saying we should copy&paste this function (including the private annotation keys) into the test/library/encryption/kms package?

ardaguclu: Actually you are right. Let's keep ReadRemoteKeyAnnotations and ApplyRemoteKeyAnnotations as exported. This is no risk.

pkg/operator/encryption/secrets/secrets.go
tjungblu: so you're saying we should just swallow the error and return time.Zero() or what's the resolution here?

ardaguclu: No, I'm saying that returning an error is correct resolution. But we need to be careful for each error we return due to the consequences.

pkg/operator/encryption/secrets/remote_key.go
tjungblu: moved the deletes into ApplyRemoteKeyAnnotations

pkg/operator/encryption/secrets/secrets.go
tjungblu: removed the if condition

pkg/operator/encryption/secrets/remote_key.go
tjungblu: removed the other helpers that are not needed for this PR

tjungblu:
@ardaguclu / @bertinatto please take a look when you find time

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	delete(annotations, encryptionSecretRemoteKeyConvergedID)
+	if rk.ConvergedID != "" {
+		annotations[encryptionSecretRemoteKeyConvergedID] = rk.ConvergedID
+	}
+
+	delete(annotations, encryptionSecretRemoteKeyConvergedAt)
+	if !rk.ConvergedAt.IsZero() {
+		annotations[encryptionSecretRemoteKeyConvergedAt] = rk.ConvergedAt.Format(time.RFC3339)
```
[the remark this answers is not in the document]
tjungblu: fractional seconds are simply not relevant for a 5 minute window

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if v, ok := annotations[encryptionSecretRemoteKeyConvergedAt]; ok && len(v) > 0 {
+		ts, err := time.Parse(time.RFC3339, v)
+		if err != nil {
+			return RemoteKeyAnnotations{}, fmt.Errorf("secret %s/%s has invalid %s annotation: %v", namespace, name, encryptionSecretRemoteKeyConvergedAt, err)
+		}
+		rk.ConvergedAt = ts
+	}
+	return rk, nil
+}
```
bertinatto, line 35: do we need  a mechanism to check whether the state stored in the annotations is valid (like converged ID is absent, but converted at is not)? I'm thinking about an invalid combo...

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+	delete(annotations, encryptionSecretRemoteKeyConvergedID)
+	if rk.ConvergedID != "" {
+		annotations[encryptionSecretRemoteKeyConvergedID] = rk.ConvergedID
+	}
+
+	delete(annotations, encryptionSecretRemoteKeyConvergedAt)
+	if !rk.ConvergedAt.IsZero() {
+		annotations[encryptionSecretRemoteKeyConvergedAt] = rk.ConvergedAt.Format(time.RFC3339)
+	}
+}
```
bertinatto, line 65: if we do need to validate the state in the annotations, we might want to do the same validation in `rk` (which could get in a bad state because of a bug)

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/util/retry"
+
+	"github.com/openshift/library-go/pkg/operator/encryption/state"
+)
+
+// RemoteKeyAnnotations is the annotation-backed view of state.RemoteKeyState.
+type RemoteKeyAnnotations = state.RemoteKeyState
```
ardaguclu: There are already many concrete objects in encryption controllers difficult to follow. Aliasing hurts readability more. Can we remove this and instead using `state.RemoteKeyState`?

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+
+// ReadRemoteKeyAnnotations reads remote key rotation annotations from a key secret.
+func ReadRemoteKeyAnnotations(s *corev1.Secret) (RemoteKeyAnnotations, error) {
+	if s == nil {
+		return RemoteKeyAnnotations{}, nil
+	}
+	return readRemoteKeyAnnotations(s.Annotations, s.Namespace, s.Name)
+}
+
+func readRemoteKeyAnnotations(annotations map[string]string, namespace, name string) (RemoteKeyAnnotations, error) {
```
ardaguclu: If we have to have `ReadRemoteKeyAnnotations`, why do we have this unexported function. We can move the logic inside `ReadRemoteKeyAnnotations` and delete this?

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -66,6 +66,11 @@ func ToKeyState(s *corev1.Secret) (state.KeyState, error) {
 		key.Mode = keyMode
 	case state.KMS:
 		key.KMS = &state.KMSState{}
+		remoteKey, err := readRemoteKeyAnnotations(s.Annotations, s.Namespace, s.Name)
```
ardaguclu: nit: This is totally cosmetic: Can you please move this logic just above https://github.com/openshift/library-go/blob/bca39fdc68cfafa961b17a15d2b811f1bb8b0e7a/pkg/operator/encryption/secrets/secrets.go#L91

So that most important annotation `EncryptionSecretKMSEncryptionConfig` will stay first

pkg/operator/encryption/secrets/secrets.go
```diff
@@ -162,6 +167,8 @@ func FromKeyState(component string, ks state.KeyState) (*corev1.Secret, error) {
 		s.Annotations[EncryptionSecretMigratedResources] = string(bs)
 	}
 
+	applyRemoteKeyAnnotations(s.Annotations, ks.RemoteKey())
```
ardaguclu: nit: Could you please move this just above https://github.com/openshift/library-go/blob/bca39fdc68cfafa961b17a15d2b811f1bb8b0e7a/pkg/operator/encryption/secrets/secrets.go#L181, so that most important data (i.e. `encryptionConfigurationData`) will stay first.

pkg/operator/encryption/secrets/secrets.go
ardaguclu: We can use exported function (ReadRemoteKeyAnnotations) just like everyone else?

pkg/operator/encryption/secrets/remote_key.go
```diff
@@ -0,0 +1,115 @@
...
+	if !rk.ConvergedAt.IsZero() {
+		annotations[encryptionSecretRemoteKeyConvergedAt] = rk.ConvergedAt.Format(time.RFC3339)
+	}
+}
+
+// MigrationWriteKeyName returns the StorageVersionMigration write-key annotation value.
+// When target-remote-key-id is set, the write-key is always suffixed with that ID
+// (first enablement and remote-key rotation). Plain keyName is used only when
+// target-remote-key-id is unset.
+func MigrationWriteKeyName(keyName string, rk RemoteKeyAnnotations) string {
```
ardaguclu: I think this PR can solely focus on roundtripping between key Secret <-> KeyState (initiated by key controller).

In my opinion, we can ignore migration controller for now. Can you please remove `MigrationWriteKeyName`, `RemoteKeyIDFromMigrationWriteKey`, `PatchRemoteKeyAnnotations` from this PR?

pkg/operator/encryption/secrets/remote_key.go
tjungblu: removed

pkg/operator/encryption/secrets/remote_key.go
tjungblu: moved

pkg/operator/encryption/secrets/remote_key.go
tjungblu: removed

pkg/operator/encryption/secrets/secrets.go
tjungblu: moved

tjungblu: used the exported function

pkg/operator/encryption/secrets/secrets.go
tjungblu: moved

pkg/operator/encryption/secrets/remote_key.go
tjungblu: good idea, right now the validations are part of the patch assertions here:

https://github.com/openshift/library-go/pull/2452/changes#diff-b866b1c67f36ca3a0ee669f60eca68344fcc475aafa6a015898bf382aaa29fe3R92-R163


ConvergedId and ConvergedAt should move in lockstep, so I've added an additional validation here when either is non-default

pkg/operator/encryption/secrets/remote_key.go
tjungblu: yeah good idea, let's add a validation function to the struct, so we can verify before writing and after reading

pkg/operator/encryption/secrets/remote_key_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	if got.TargetRemoteKeyID != "remote-old" || got.MigratedRemoteKeyID != "remote-old" {
+		t.Fatalf("unexpected ids: %#v", got)
+	}
+	if got.ConvergedID != "remote-new" || !got.ConvergedAt.Equal(ts) {
+		t.Fatalf("unexpected convergence: %#v", got)
+	}
+}
+
+func TestApplyRemoteKeyAnnotationsClearsEmptyValues(t *testing.T) {
```
ardaguclu: Would it make sense enriching this function with more cases such as annotation is deleted, annotation is updated, etc.?

pkg/operator/encryption/secrets/remote_key.go
ardaguclu: In my opinion, it makes sense adding a validation in this function for the ones that are certainly expected to be not empty. However, currently they are all empty. We may need to add validations later on.

ardaguclu (commented):
Would you please run the changes in operators with fake bumps just to be in the safe side? (I would not expect any regressions though).

pkg/operator/encryption/secrets/remote_key_test.go
tjungblu: yep, adding a few more cases

tjungblu:
@ardaguclu  there you go:

- cluster-authentication-operator: https://github.com/openshift/cluster-authentication-operator/pull/996
- cluster-kube-apiserver-operator: https://github.com/openshift/cluster-kube-apiserver-operator/pull/2308
- cluster-openshift-apiserver-operator: https://github.com/openshift/cluster-openshift-apiserver-operator/pull/769

pkg/operator/encryption/state/types.go
```diff
@@ -94,6 +95,14 @@ type RemoteKeyState struct {
 	ConvergedID string
 }
 
+// Validate checks whether the remote key state is coherent. Returns a descriptive error when it is not.
+func (rk RemoteKeyState) Validate() error {
```
ardaguclu, line 99: Would it be better to have this referenced type (`rk *RemoteKeyState`) and store as referenced type inside KMSKeyState?

tjungblu: why? are you worried about extensive stack copies or what is the matter?

ardaguclu: Not about extensive stack copies but more about following the convention here https://github.com/openshift/library-go/blob/master/pkg/operator/encryption/state/types.go. All the types are carried as referenced type?

ardaguclu: But this is not a blocker for me, so I'll leave the decision to you

tjungblu: they're all value types here:

``` 
// KMSState stores all KMS encryption mode related configurations
type KMSState struct {

	// Plugin stores KMS plugin specific configurations
	Plugin configv1.KMSPluginConfig

	// PluginSecretData stores data key-value pairs fetched from referenced secrets.
	PluginSecretData KMSReferenceData

	// PluginConfigMapData stores data key-value pairs fetched from referenced configmaps.
	PluginConfigMapData KMSReferenceData

	// RemoteKey tracks KMS remote key rotation annotations on the backing secret.
	RemoteKey RemoteKeyState
}
```  

You mean the function declaration uses pointers? 
As with https://go.dev/doc/faq#methods_on_values_or_pointers - we don't mutate and there is no other function on that struct yet with a pointer, so we don't _need_ a pointer here.

ardaguclu: > You mean the function declaration uses pointers?

Yeah, I meant that. We can leave as is then. Thanks

ardaguclu:
/approve
Roundtripping looks good to me. I'll leave the tag to @bertinatto

bertinatto (commented):
/lgtm

PR #2472, opened 2026-09-14
NO-JIRA: add kms rotation wait e2e test helper

tjungblu:
This test helper waits through the entire annotation based states with comprehensive log statements for easier debugging in case of failures.

## Summary by CodeRabbit

- **Tests**
  - Added automated validation for remote encryption key rotation completion.
  - Rotation checks now confirm that the new key is fully active and differs from the previously migrated key.
  - Temporary secret retrieval failures are retried, while invalid key state and timeout conditions are reported clearly.


test/library/encryption/helpers.go
```diff
@@ -91,6 +92,63 @@ func WaitForNextEncryptionKeyRotation() WaitForRotationCompleteFunc {
...
+		t.Helper()
+		require.NotEmpty(t, prevKeyMeta.Name, "previous key name is required")
+
+		t.Logf("Waiting for KMS remote key rotation to complete on secret %q", prevKeyMeta.Name)
+		var initialRemoteKeyID string
+		// we can't use the test context here, because ginkgo doesn't populate it
+		err := wait.PollUntilContextTimeout(context.Background(), waitPollInterval, waitPollTimeout, true, func(ctx context.Context) (bool, error) {
+			secret, err := clientSet.Kube.CoreV1().Secrets(scenario.Namespace).Get(ctx, prevKeyMeta.Name, metav1.GetOptions{})
+			if err != nil {
+				return false, nil
```
bertinatto, line 111: is it intentional that you're not returning any GET errors here?

tjungblu: of course, why would I return an error here? when a GET fails I want to retry. Returning an error will immediately stop the loop.

bertinatto (commented):
/lgtm

PR #2473, opened 2026-09-14
NO-JIRA: record first remote key id after preflight succeeded

tjungblu:
## Summary by CodeRabbit

- **Enhancements**
  - KMS preflight results now include the associated remote KMS key ID.
  - Generated key configurations retain the remote KMS key ID after successful validation, supporting subsequent key management operations.
  - Existing handling for preflight errors and pending states remains unchanged.
  - Key secrets are created only after successful KMS preflight validation.


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -267,13 +267,14 @@ func (c *keyController) generateKeySecret(ctx context.Context, keyID uint64, cur
...
 		}
-		preflightPassed, err := c.ensureKMSPreflightPassed(ctx, configHash)
+		preflightPassed, remoteKeyID, err := c.ensureKMSPreflightPassed(ctx, configHash)
 		if err != nil {
 			return nil, false, err
 		}
 		if !preflightPassed {
 			return nil, false, nil
 		}
+		ks.KMS.RemoteKey.TargetRemoteKeyID = remoteKeyID
```
bertinatto: can there be a situation where the preflight passed and `remoteKeyID` is empty? Do we have a problem when that happens?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -451,33 +452,33 @@ func (p *prefetchedKMSConfigHasherResourceProvider) getConfigMap(_ context.Conte
...
 		}
 		// Scenario 1: back off: wait for the preflight controller to pick up the new hash.
-		return false, nil
+		return false, "", nil
 	}
 
 	// Scenario 2: preflight passed — proceed.
 	if isPreflightResultSucceeded(&encryptionStatus.Preflight.Result, configHash) {
-		return true, nil
+		return true, encryptionStatus.Preflight.Result.RemoteKeyID, nil
```
bertinatto: it feels like the remote key doesn't belong in `ensureKMSPreflightPassed`; perhaps it'd be better to have a specific function for that? To avoid fetching the status twice, both functions (`ensureKMSPreflightPassed` and the new one) could accept a status param? Does that makse sen?

tjungblu: sure, will refactor this

pkg/operator/encryption/controllers/key_controller.go
tjungblu: as per 
https://github.com/openshift/api/blob/9edb1b2245e261685a93d77f526b8c578bcdd864/operator/v1/types_kmsencryption.go#L170

it can't be empty, no.

bertinatto (commented):
/lgtm

PR #2463, opened 2026-09-09
CNTRLPLANE-3237: kms preflight drift positive path

p0lyn0mial:
## Summary by CodeRabbit

- **Tests**
  - Improved KMS encryption scenario validation by capturing preflight pod information during configuration.
  - Added checks for configuration drift between the KMS preflight pod and the running operand.
  - Validation now covers operator-specific status paths, preflight status transitions, pod details, namespace, and label selection.
  - Added safeguards to ensure captured preflight information remains current and is validated consistently.

PR #2474, opened 2026-09-14
CNTRLPLANE-3237: reland kms preflight drift positive path

p0lyn0mial:
brings back https://github.com/openshift/library-go/pull/2463

and adds the second commit on top of it which properly finds the preflight pod.


p0lyn0mial:
/hold 

for testing in https://github.com/openshift/cluster-kube-apiserver-operator/pull/2310

PR #2476, opened 2026-09-16
WIP: CNTRLPLANE-3237: Move to internal KMS API type

ardaguclu:
We have decided to take KMSPluginConfiguration out of openshift/api to be managed by external OLM operator (ref: https://github.com/openshift/api/pull/3028). As complementary next step, this PR updates the encryption controllers to use internalized KMS API to work with external OLM operator API contract. 

This is intermediate step to minimize the changes to not delay the delivery of this feature. When we start adopting plugin-agnostic generic approach, we can get rid of internal KMS API and directly use `unstructured`. 

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Vault-based encryption through referenced KMS configuration resources.
  * KMS settings can now be resolved dynamically, including provider details, plugin images, authentication, TLS, and referenced Secrets or ConfigMaps.
  * Added validation for missing, invalid, unsupported, or incomplete KMS configurations.

* **Bug Fixes**
  * Encryption preflight checks now wait for configuration and key state to match before proceeding.
  * Invalid or stale KMS configurations prevent key creation and report degraded status.
  * Improved handling of KMS configuration and plugin image updates.


pkg/operator/encryption/kms/types.go
```diff
@@ -0,0 +1,93 @@
...
+	scheme.AddKnownTypes(SchemeGroupVersion, &KMSPluginConfig{})
+	return nil
+}
+
+// KMSPluginConfig holds resolved plugin configuration.
+// Remove this type when the encryption lifecycle uses unstructured configuration.
+type KMSPluginConfig struct {
+	metav1.TypeMeta `json:",inline"`
+	Type            KMSProviderType      `json:"type"`
+	Vault           VaultKMSPluginConfig `json:"vault,omitempty,omitzero"`
```
bertinatto, line 22: in the openshift/api PR, we haven't removed `configv1.VaultKMSPluginConfig` yet — we just tombstoned the `KMSPluginConfig.Vault` field. As a result, we could re-use those types here, so this struct could look like:

```
type KMSPluginConfig struct {
	metav1.TypeMeta `json:",inline"`
	Type            configv1.KMSProviderType      `json:"type"`
	Vault           configv1.VaultKMSPluginConfig `json:"vault,omitempty,omitzero"`
}
```

I'm not sure this is a good idea, though. I experimented with it [here](https://github.com/openshift/library-go/compare/master...bertinatto:library-go:new-kms-api-2?expand=1) and the diff got smaller, but not substantially smaller. Anyway, just wanted to share this idea.

pkg/operator/encryption/controllers/kms_config.go
```diff
@@ -50,6 +55,21 @@ func ResolveKMSConfig(ctx context.Context, client dynamic.Interface, reference c
 			return kms.KMSPluginConfig{}, fmt.Errorf("failed to read KMS plugin image: %w", err)
 		}
 		config.Vault.KMSPluginImage = image
+		if config.Vault.KMSPluginImage == "" {
+			return kms.KMSPluginConfig{}, fmt.Errorf("KMS plugin configuration status.kmsPluginImage must not be empty")
+		}
```
bertinatto, line 60: do we need additional validation to the image field? For example, do we need to make sure the operator correctly specified a digest rather then just a tag?

pkg/operator/encryption/controllers/kms_config.go
bertinatto, line 41: nit: I started to have second thoughts about the `type` field. Since we only support Vault for now, we could assume the reference object is vault-related. If not, we could fail. When we add a new provider, we'll support the generic interface, so we won't need the `type` field anyway.

The problem with removing the `type` field now is that once we introduce a new provider we'll **have** to implement the generic interface. If we keep the field, we'd still have the option to use the same approach as we do for Vault (internal type - although we probably don't want to do that).

pkg/operator/encryption/controllers/kms_config.go
```diff
@@ -50,6 +55,21 @@ func ResolveKMSConfig(ctx context.Context, client dynamic.Interface, reference c
...
+		if config.Vault.VaultKeyPath == "" {
+			return kms.KMSPluginConfig{}, fmt.Errorf("KMS plugin configuration spec.vaultKeyPath must not be empty")
+		}
+		if config.Vault.Authentication.Type == "" {
+			return kms.KMSPluginConfig{}, fmt.Errorf("KMS plugin configuration spec.authentication.type must not be empty")
+		}
+		if config.Vault.Authentication.Type == kms.VaultAuthenticationTypeAppRole && config.Vault.Authentication.AppRole.Secret.Name == "" {
+			return kms.KMSPluginConfig{}, fmt.Errorf("KMS plugin configuration spec.authentication.appRole.secret.name must not be empty")
+		}
```
bertinatto, line 72: I wonder if we should implement all validations we currently have at the api admission level today. For instance, the user may specify a random resource whose validations were not properly done)

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -409,7 +409,11 @@ func modeAndExternalReasonFromAPIServer(ctx context.Context, apiServerClient con
 		return "", "", configv1.APIServerEncryption{}, err
 	}
 
-	return modeAndExternalReasonFromAPIServerEncryption(apiServer.Spec.Encryption, operatorClient, unsupportedConfigPrefix)
+	mode, reason, err := modeAndExternalReasonFromAPIServerEncryption(apiServer.Spec.Encryption, operatorClient, unsupportedConfigPrefix)
+	if err != nil {
+		return "", "", configv1.APIServerEncryption{}, err
+	}
+	return mode, reason, apiServer.Spec.Encryption, nil
```
bertinatto, line 416: good catch

bertinatto (commented):
overall these changes look good to me. Before merging this, we'll need to coordinate:

* openshift/api PR
* this PR
* operators bump
* openshift/release PR (that will install the OLM operator or just create a fake CR)

pkg/operator/encryption/controllers/kms_config.go
ardaguclu: Since this approach is still deep integration, I think we should validate to accept only digest and error out if it is tag.

pkg/operator/encryption/controllers/kms_config.go
ardaguclu: I see some value keeping `type` field. But it is not strong enough to convince the API reviewers :) 

> once we introduce a new provider we'll have to implement the generic interface

That is correct

pkg/operator/encryption/controllers/kms_config.go
ardaguclu: That is a good point. My motivation behind adding these validations was to prevent any panic. I think, we can assume that there should be a validation in API level.

test/e2e-encryption/encryption_test.go
```diff
@@ -541,6 +543,84 @@ func TestEncryptionIntegration(tt *testing.T) {
 	)
 	waitForConditionStatus("Encrypted", operatorv1.ConditionTrue)
 
+	t.Logf("Install VaultKMSConfig CRD and prepare provider CR helpers")
```
tjungblu, line 546: I think we can remove this section now with https://github.com/openshift/release/pull/85287

ardaguclu:
Closing this one in favor of https://github.com/openshift/library-go/pull/2486
/close

PR #2477, opened 2026-09-16
CNTRLPLANE-3237: clear any Preflight left from a prior KMS episode so a later re-enable re-runs the reachability check.

p0lyn0mial:
the bug was uncovered by https://github.com/openshift/cluster-kube-apiserver-operator/pull/2310 and https://github.com/openshift/library-go/pull/2474

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption status handling when KMS encryption is disabled by clearing stale preflight results.
  * Ensured preflight status cleanup can be retried if an update fails.
  * Prevented outdated KMS preflight data from persisting when encryption modes change.
  * Kept deployment status accurate after encryption settings are updated, without unnecessary degradation.


p0lyn0mial:
/hold

for a test PR in kas-o

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -214,7 +214,10 @@ type kmsPreflightController struct {
 	// An alternative would be to check a cached lister before issuing Delete
 	// calls, but the encryption controllers do not use informers and have no
 	// shared cache available.
-	dirtyDeployer            bool
+	dirtyDeployer bool
+	// dirtyPreflight is true when Preflight may be set and needs clearing once
+	// encryption is no longer KMS. Mirrors dirtyDeployer; initialized to true.
+	dirtyPreflight           bool
```
p0lyn0mial, line 220: i'm still experimenting. 

i think the better idea would be to add the generation from the api obj to the hash.

ardaguclu: Generation can be changed in any minor change in the APIServer config CRD.

pkg/operator/encryption/controllers/kms_preflight_controller.go
```diff
@@ -719,11 +723,21 @@ func (c *kmsPreflightController) preflightRequired(ctx context.Context) (string,
...
 	}
 	if apiServer.Spec.Encryption.Type != configv1.EncryptionTypeKMS {
-		// Encryption is not KMS — nothing to preflight. A stale ObservedConfigHash
-		// (written when KMS was active) is irrelevant; the key controller will
-		// overwrite it when/if KMS is re-enabled.
+		// Not KMS: clear any Preflight left from a prior KMS episode so a later
+		// re-enable (even with an unchanged config) re-runs the reachability check.
+		// Safe under a transient outage: dirtyPreflight stays set until the write
+		// lands, and a re-enable can't proceed until the off migration completes.
+		if c.dirtyPreflight {
```
ardaguclu, line 730: My idea may not be ideal as well: Can key controller clean up before calling the preflight instead of having dirtyPreflight flag?

p0lyn0mial: yeah, this pr and adding cleanup to the key controller is not ideal. 

it is edge triggered and edges can be missed and we would clear the status which might be useful for debugging. 

i'd prefer https://github.com/openshift/library-go/pull/2478, ptal and let me know what you think

p0lyn0mial:
/close

we went with https://github.com/openshift/library-go/pull/2478

PR #2481, opened 2026-09-17
OPRUN-4768: Support TLS curve preferences in serving configuration

tmshort:
Updates `github.com/openshift/api` to the revision that adds `ServingInfo.CurvePreferences` and refreshes the vendored dependency.

Propagates curve preferences through `config/serving.ToServingOptions` into Kubernetes `SecureServingOptions`.

Adds `crypto.TLSGroupsToCurvePreferences`, which accepts both API `TLSGroup` values and observed string values, converts them to numeric curve IDs, and returns unsupported groups for the caller to handle.

Jira: https://issues.redhat.com/browse/OPRUN-4768
Usage Example: https://github.com/openshift/cluster-olm-operator/pull/248

Tested:
- `make verify`
- `make test-unit`

## Summary by CodeRabbit

* **Bug Fixes**
  * TLS curve preferences configured for secure serving are now preserved and applied consistently.
  * Recognized TLS groups are converted to the corresponding curve preferences, while unrecognized groups are reported for clearer configuration handling.

* **Tests**
  * Added coverage for TLS curve conversion, unsupported values, empty results, and preservation of configured serving preferences.


tmshort:
The **downstream-kms-images** CI is failing because
  - cluster-authentication-operator
  - cluster-kube-apiserver-operator
  - cluster-openshift-apiserver-operator

have not updated to the version of openshift/api that's been pulled into here.

ricardomaraschini (commented):
/lgtm

ingvagabund:
The wiring validated via https://github.com/openshift/cluster-kube-scheduler-operator/pull/679.

tmshort:
> The wiring validated via [openshift/cluster-kube-scheduler-operator#679](https://github.com/openshift/cluster-kube-scheduler-operator/pull/679).

And also: https://github.com/openshift/cluster-olm-operator/pull/248

pkg/crypto/crypto.go
```diff
@@ -395,6 +395,29 @@ func TLSGroupsToCurveIDs(groups []configv1.TLSGroup) ([]tls.CurveID, []configv1.
...
+
+	curveIDs, unrecognizedTLSGroups := TLSGroupsToCurveIDs(tlsGroups)
+	curvePreferences := make([]int32, len(curveIDs))
+	for i, curveID := range curveIDs {
+		curvePreferences[i] = int32(curveID)
+	}
+	var unrecognizedGroups []T
+	for _, group := range unrecognizedTLSGroups {
+		unrecognizedGroups = append(unrecognizedGroups, T(group))
+	}
```
sanchezl, line 417: Nit: Why is this loop not consistent with the other which pre-allocate? Is returning a nil vs empty slice a consideration? Hopefully not, but If so, let's add a comment saying so.

tmshort: No particular reason. Being a nit, I'm inclined to only fix it if there are other substantial changes to be made (this PR already has LGTM).

ricardomaraschini (commented):
Sorry for arriving a little bit late for the party but I was wondering if this requirement (from the [Enhancement Proposal](https://github.com/openshift/enhancements/blob/master/enhancements/security/api-tls-curves-config.md)) affects this PR somehow:

```
FIPS mode constraint: X25519 and X25519MLKEM768 are not FIPS-approved. Components running in FIPS mode
must omit these groups. The FIPS-approved post-quantum alternatives (SecP256r1MLKEM768, SecP384r1MLKEM1024)
require Go 1.26+ and are not currently supported.
```

tmshort:
> Sorry for arriving a little bit late for the party but I was wondering if this requirement (from the [Enhancement Proposal](https://github.com/openshift/enhancements/blob/master/enhancements/security/api-tls-curves-config.md)) affects this PR somehow:
> 
> ```
> FIPS mode constraint: X25519 and X25519MLKEM768 are not FIPS-approved. Components running in FIPS mode
> must omit these groups. The FIPS-approved post-quantum alternatives (SecP256r1MLKEM768, SecP384r1MLKEM1024)
> require Go 1.26+ and are not currently supported.
> ```

@ricardomaraschini A filtering API was added in https://github.com/openshift/library-go/pull/2414 for those components that want to use it (alternatively, components can use an alternative method)

bertinatto (commented):
/approve

tmshort:
There was a merge conflict; someone else already updated the openshift/api dependency.

pkg/crypto/crypto.go
tmshort: This was actually intentional behavior, to return `nil` via `unrecognizedGroups` if there are no unrecognized groups. But the code has been updated to preallocate if there are unrecognized groups.

PR #2484, opened 2026-09-21
NO-JIRA: block encryption key minting during KMS remote-key migration

tjungblu:
## Summary
- Add `secrets.NeedsRemoteKeyMigration` to detect when target and migrated remote key IDs differ.
- Use it in `needsNewKey` so the key controller does not mint a new encryption-key secret while a KMS remote-key migration is in flight.

This is the first vertical slice of KMS remote-key rotation: helper lands with its call site (no unused API).

## Test plan
- [x] `go test ./pkg/operator/encryption/secrets/ ./pkg/operator/encryption/controllers/ -run 'NeedsRemoteKey|KeyController|needsNew'`
- [ ] CI green on this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Pauses encryption key creation when a remote key migration is pending.
  * Prevents new key plans until the migrated remote key matches the configured target key.
* **Tests**
  * Added coverage for remote key migration status, including unset, matching, and mismatched key scenarios.


pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -598,6 +598,11 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
 			return latestKeyID, "kms-provider-changed", true, nil
 		}
 
+		if latestKey.RemoteKey().NeedsRemoteKeyMigration() {
+			// Block new encryption key minting while target and migrated remote key IDs differ.
+			return 0, "", false, nil
```
ardaguclu: This is a dead code. https://github.com/openshift/library-go/blob/975a098b90da415b75324434033ef8171131c8c3/pkg/operator/encryption/controllers/key_controller.go#L612 already returns the same value.

Additionally, key controller pauses if there is any active migration. So I don't think we should add this check in key controller, probably we should add in migration controller.

tjungblu: sorry, that should've been up top. 

how can we stop provider migration in the migration controller?

pkg/operator/encryption/controllers/key_controller.go
```diff
@@ -592,12 +592,18 @@ func needsNewKey(grKeys state.GroupResourceState, currentMode state.Mode, extern
...
-		// We are here because Encryption Mode is not changed
-		// However, we need to create a new key if migration-triggering fields
-		// in the KMS provider configuration have changed.
 		if latestKey.KMS == nil {
 			return 0, "", false, fmt.Errorf("KMS-mode key %q has nil KMS state, possibly corrupted key secret", latestKey.Key.Name)
 		}
+		// Block all new encryption-key minting while remote-key migration is in flight,
+		// including kms-provider-changed. Remote-key rotation must finish on the current
+		// write key before a provider migration can mint another secret.
+		if latestKey.RemoteKey().NeedsRemoteKeyMigration() {
```
ardaguclu, line 601: Do we allow encryption mode changes but not kms configuration changes?. Is the expected mechanism described in EP?. 

I think, first we need to agree upon what is the priority list between encryption mode changes, key rotation and kms configuration change.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: > how can we stop provider migration in the migration controller?

My comment about migration controller does not make sense. Ignore it.

pkg/operator/encryption/controllers/key_controller.go
ardaguclu: In my opinion, key rotation should only be initiated, iff needsnewkey is false. Not reverse

tjungblu: This is described in case 2:
https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/kms-encryption-foundations.md#2-rotation-when-remotekeyid-changes-happy-path

Bullet 6: "stateController is unchanged. keyController blocks new encryption key minting while needsMigration."

Otherwise a new encryption-key secret could be minted mid-rotation (e.g. kms-provider-changed), which would abandon the in-flight remote-key annotations on the current write key and pull stateController into a provider migration. Rotation is supposed to finish on that write key first.

That was one of Lukasz's edge case scenarios where we initially found the race condition.


> In my opinion, key rotation should only be initiated, iff needsnewkey is false. Not reverse

that's fair, but nothing here rotates. We just ensure that we can't create a new key.

PR #2486, opened 2026-09-22
WIP: CNTRLPLANE-3237: Move to internal KMS API type

ardaguclu:
We have decided to take KMSPluginConfiguration out of openshift/api to be managed by external OLM operator (ref: https://github.com/openshift/api/pull/3028). As complementary next step, this PR updates the encryption controllers to use internalized KMS API to work with external OLM operator API contract.

This is intermediate step to minimize the changes to not delay the delivery of this feature. When we start adopting plugin-agnostic generic approach, we can get rid of internal KMS API and directly use unstructured.

This PR supersedes https://github.com/openshift/library-go/pull/2476

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving Vault KMS encryption settings from referenced Kubernetes resources.
  * Added validation for required KMS settings, supported fields, image digests, TLS, and authentication.
  * Added KMS configuration types and resource handling for encryption workflows.

* **Bug Fixes**
  * Improved encryption planning and preflight processing when KMS configuration references change.
  * Preserved migration behavior while avoiding unnecessary key creation for non-identity configuration changes.

* **Tests**
  * Expanded coverage for configuration resolution, validation, migration, preflight failures, and resource switching.


ardaguclu:
I've opened fake bumps:
* https://github.com/openshift/cluster-kube-apiserver-operator/pull/2318
* https://github.com/openshift/cluster-openshift-apiserver-operator/pull/784
* https://github.com/openshift/cluster-authentication-operator/pull/1007

ardaguclu:
Since https://github.com/openshift/api/pull/3028 has merged, I've opened these fake bumps to test;
* https://github.com/openshift/cluster-authentication-operator/pull/1010
* https://github.com/openshift/cluster-openshift-apiserver-operator/pull/786
* https://github.com/openshift/cluster-kube-apiserver-operator/pull/2319

test/e2e-encryption/encryption_test.go
```diff
@@ -541,6 +543,84 @@ func TestEncryptionIntegration(tt *testing.T) {
 	)
 	waitForConditionStatus("Encrypted", operatorv1.ConditionTrue)
 
+	t.Logf("Install VaultKMSConfig CRD and prepare provider CR helpers")
```
tjungblu, line 546: this can be removed now that the vault workflow installs it.
We can also hardcode the two vault configs by name here.

``` 
kubectl get vaultkmsconfigs.kms.openshift.io
NAME                  AGE
vault-kms             2m49s
vault-kms-secondary   2m41s
```

ardaguclu: This is integration test which is independent from CI configuration. That's the reason we need to create it here.

ardaguclu:
/retitle [CNTRLPLANE-3237](https://redhat.atlassian.net/browse/CNTRLPLANE-3237): Move to internal KMS API type
/hold

test/e2e-encryption/encryption_test.go
tjungblu: that sucks that we have to copy and keep the CRDs in two places. Is this one running in the e2e-encryption test?

ardaguclu: This is fully isolated test from all the dependencies. This does not even depend on APIServer config CRD. It stores its own custom CRDs.

tjungblu: having said that, we can also run the vault installation/config steps before running `make test-e2e-encryption` in CI because it runs against a cluster regardless. Local integration testing would require a cluster though, neither a good solution.

pkg/operator/encryption/controllers/helpers_test.go
```diff
@@ -144,3 +149,33 @@ func newDeployedKMSEncryptionConfig(t *testing.T, instanceName string, encrypted
 	}
 	return secret
 }
+
+func kmsConfigReference(config kms.KMSPluginConfig) configv1.KMSPluginConfig {
```
tjungblu: config is unused?

pkg/operator/encryption/controllers/kms_config.go
```diff
@@ -0,0 +1,85 @@
...
+func ResolveKMSConfig(ctx context.Context, client dynamic.Interface, reference configv1.KMSPluginConfig) (kms.KMSPluginConfig, error) {
+	if client == nil {
+		return kms.KMSPluginConfig{}, fmt.Errorf("dynamic client is required to resolve KMS plugin configuration")
+	}
+	ref := reference.PluginConfig
+	gv, err := schema.ParseGroupVersion(ref.APIVersion)
+	if err != nil {
+		return kms.KMSPluginConfig{}, fmt.Errorf("invalid KMS plugin configuration apiVersion %q: %w", ref.APIVersion, err)
+	}
+	obj, err := client.Resource(gv.WithResource(ref.Resource)).Get(ctx, ref.Name, metav1.GetOptions{})
```
tjungblu, line 31: just out of curiosity: are there also dynamic informer caches? I don't think doing a live call here is a huge issue though

ardaguclu: We have decided not to use dynamic informer caches. Since we don't control them, we can't be sure whether they create hot loop or not. So it is better to rely on the periodic 1 minute sync and live call.

pkg/operator/encryption/controllers/helpers_test.go
ardaguclu: Nice catch. I've removed this redundant function.

pkg/operator/encryption/controllers/kms_config.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+		image, _, err := unstructured.NestedString(obj.Object, "status", "kmsPluginImage")
+		if err != nil {
+			return kms.KMSPluginConfig{}, fmt.Errorf("failed to read KMS plugin image: %w", err)
+		}
+		config.Vault.KMSPluginImage = image
+		if config.Vault.KMSPluginImage == "" {
+			return kms.KMSPluginConfig{}, fmt.Errorf("KMS plugin configuration status.kmsPluginImage must not be empty")
+		}
+		imageRef, err := imagereference.Parse(image)
```
ardaguclu, line 62: I used this utility function in library-go that verifies image is composed of digest or not (we don't want to allow the usage of `tags`).

ardaguclu:
@tjungblu can correct me, but as far as I understand downstream-kms-images can not pass in this PR

tjungblu:
We can override it. I don't know why it shows as required, for tide it is not.

tjungblu:
I'll try to figure out why it fails, but I assume it is because of the client-go bump

bertinatto:
> We can override it. I don't know why it shows as required, for tide it is not.
> I'll try to figure out why it fails, but I assume it is because of the client-go bump

/override ci/prow/downstream-kms-images

bertinatto (commented):
/lgtm

ardaguclu:
All the CI jobs have passed https://github.com/openshift/cluster-kube-apiserver-operator/pull/2319

